Issue #14, opened 2018-06-11
Support for dynamics randomization


ryanjulian:
https://blog.openai.com/generalizing-from-simulation/

We should probably not attempt this until #4 is done.

For now, only MuJoCo support is necessary. If MuJoCo is too burdensome, we can consider switching engines (e.g. Bullet)

Related issue https://github.com/openai/mujoco-py/issues/148 suggests this may be nontrivial.

Imported from https://github.com/ryanjulian/rllab/issues/61

ryanjulian:
Fixed by #51

PR #51, opened 2018-06-11
Dynamics randomization for MuJoCo

CatherineSue:
Add support for dynamics randomization for mujoco environment.
Includes a simple data structure consisting of a list of variation
objects, a wrapped environment of mujoco to perform dynamics
randomization.
Each variation object is an instance of the Variation class
that works as a container for each of the fields used to randomized a 
dynamic parameter within the simulation environment.
The wrapper class of mujoco performs dynamics randomization on each
reset().
The data structure and the wrapper class are tested in
test_dynamics_rand.py.

Refer to: #14


gonzaiva:
Reviewers, this PR was brought from rllab, so all of it has been reviewed previously. However, if you want to request a change, feel free to do so. Otherwise, please approve to merge this change.

rllab/envs/mujoco/dynamics_randomization/test_dynamics_rand.py
```diff
@@ -0,0 +1,21 @@
+from rllab.envs.mujoco import SwimmerEnv
```
ryanjulian: Please put this file in tests/

rllab/envs/mujoco/dynamics_randomization/__init__.py
```diff
@@ -0,0 +1,4 @@
+from rllab.envs.mujoco.dynamics_randomization.randomized_env import randomize
```
ryanjulian: Please name the package `randomization`, since it's not actually specific to dynamics.

PR #53, opened 2018-06-11
Add PEP8 checks to the CI

gonzaiva:
Introduce new support packages and scripts for Travis CI to detect
format errors in the code based on the PEP8 style guide.


test_pep8/indentation.py
```diff
@@ -0,0 +1,71 @@
+"""
+Identation is nicely verified by pylint, and this file provides examples
```
eric-heiden: typo: "Indentation"

scripts/travisci/check_flake8.sh
```diff
@@ -0,0 +1,19 @@
...
+# Incremental verification
+# E501 - Line too long
+if [[ "${TRAVIS_PULL_REQUEST}" != "false" ]]; then
+  git diff ${TRAVIS_COMMIT_RANGE} | flake8 --diff --select=E501
+  if [ $? -ne 0 ]; then
+    exit_status=1
+  fi
+fi
+
+exit $exit_status
```
ryanjulian: default bash behavior is non-zero exit status if any subcommand had non-zero exit status. does flake8 always exit(0)?

scripts/travisci/check_flake8.sh
```diff
@@ -0,0 +1,19 @@
+exit_status=0
```
ryanjulian: `#!/usr/bin/env bash`

scripts/travisci/check_pylint.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
```
ryanjulian: `#!/usr/bin/env bash`

test_pep8/indentation.py
```diff
@@ -0,0 +1,71 @@
+"""
```
ryanjulian: I don't think we need to test the linter. I trust that it works.

test_pep8/maximum_line_length.py
```diff
@@ -0,0 +1,31 @@
+"""
```
ryanjulian: no need to test the linter. please remove this file.

test_pep8/indentation.py
ryanjulian: please remove this file.

scripts/travisci/check_flake8.sh
```diff
@@ -0,0 +1,19 @@
...
+# E101 - Indentation contains mixed spaces and tabs
+find -iname "*.py" | flake8 --select=E101,W191
+if [ $? -ne 0 ]; then
+  exit_status=1
+fi
+
+# Incremental verification
+# E501 - Line too long
+if [[ "${TRAVIS_PULL_REQUEST}" != "false" ]]; then
+  git diff ${TRAVIS_COMMIT_RANGE} | flake8 --diff --select=E501
```
ryanjulian: what about the `else` clause? You only covered pull request jobs (there are also branch jobs), which unfortunately need a slightly different way of calculating changed files).

See https://github.com/ryanjulian/rllab/commit/4694327875be9557fe1a7f805f23112bf223541a

scripts/travisci/check_flake8.sh
ryanjulian: exit "${exit_status}"

see https://google.github.io/styleguide/shell.xml

scripts/travisci/check_flake8.sh
```diff
@@ -0,0 +1,19 @@
+exit_status=0
+# Absolute verification
+# W191 - Indentation contains tabs
+# E101 - Indentation contains mixed spaces and tabs
+find -iname "*.py" | flake8 --select=E101,W191
+if [ $? -ne 0 ]; then
```
ryanjulian:
```sh
if [[ $? -ne 0 ]]; then
  exit_status=1
fi
```

see https://google.github.io/styleguide/shell.xml

scripts/travisci/check_flake8.sh
```diff
@@ -0,0 +1,19 @@
...
+  exit_status=1
+fi
+
+# Incremental verification
+# E501 - Line too long
+if [[ "${TRAVIS_PULL_REQUEST}" != "false" ]]; then
+  git diff ${TRAVIS_COMMIT_RANGE} | flake8 --diff --select=E501
+  if [ $? -ne 0 ]; then
```
ryanjulian: use ``if [[ ]]``

see https://google.github.io/styleguide/shell.xml

scripts/travisci/check_flake8.sh
gonzaiva: Done

scripts/travisci/check_flake8.sh
gonzaiva: Done

scripts/travisci/check_pylint.sh
gonzaiva: Done

scripts/travisci/check_flake8.sh
gonzaiva: I noticed that in the push build the script for flake8 was returning zero even though the absolute checks did not passed for the test files I added in this PR, and the incremental errors were not verified since TRAVIS_PULL_REQUEST is false for that build.
I will check this again to see what's going on with the return value.

test_pep8/indentation.py
gonzaiva: I won't push this files to the master branch, I'm only using them to test the incremental changes in the feature branch.

scripts/travisci/check_flake8.sh
gonzaiva: Done

scripts/travisci/check_flake8.sh
gonzaiva: Done

scripts/travisci/check_flake8.sh
gonzaiva: Done

scripts/travisci/check_flake8.sh
```diff
@@ -0,0 +1,146 @@
...
+# W291 trailing whitespace
+errors+="W291"
+# E702 multiple statements on one line (semicolon)
+errors+="E702"
+
+# PROGRAMMING RECOMMENDATIONS
+# E714 use is not operator rather than not ... is.
+errors+="E714"
+
+find -iname "*.py" | flake8 --select=${errors}
```
ryanjulian: I think flake8 checks all .py files recursively by default(?)

gonzaiva: You're right. I used find for pylint and probably copied that from there. I will change it.

scripts/travisci/check_flake8.sh
```diff
@@ -0,0 +1,146 @@
...
+
+if [[ "${TRAVIS_PULL_REQUEST}" != "false" ]]; then
+  git diff ${TRAVIS_COMMIT_RANGE} | flake8 --diff ${quotes} --select=${errors} --ignore=${ignore}
+  if [[ $? -ne 0 ]]; then
+    exit_status=1
+  fi
+else
+  git remote set-branches --add origin master
+  git fetch
+  git diff origin/master | flake8 --diff ${quotes} --select=${errors} --ignore=${ignore}
```
ryanjulian: these variables should all be in " "

https://google.github.io/styleguide/shell.xml

scripts/travisci/check_flake8.sh
```diff
@@ -0,0 +1,146 @@
...
+errors+="E711,"
+# E714 use is not operator rather than not ... is.
+errors+="E714,"
+# E731 do not assign a lambda expression, use a def.
+errors+="E731,"
+# E722 do not use bare except, specify exception instead.
+errors+="E722"
+
+if [[ "${TRAVIS_PULL_REQUEST}" != "false" ]]; then
+  git diff ${TRAVIS_COMMIT_RANGE} | flake8 --diff ${quotes} --select=${errors} --ignore=${ignore}
```
ryanjulian: variables in quotes

https://google.github.io/styleguide/shell.xml

scripts/travisci/check_flake8.sh
```diff
@@ -0,0 +1,146 @@
...
+
+
+# INCREMENTAL VERIFICATION:
+# The following errors are analyzed only in the modified code in the pull
+# request branch.
+#
+# INDENTATION
+# E125 continuation line with same indent as next logical line
+# E128 continuation line under-indented for visual indent
+errors="E125,E128,"
```
ryanjulian: IMO this would be much more readable as an array
```sh
errors=(
# INDENTATION
E125  # continuation line with same indent as next logical line
E128  # continuation line under-indented for visual indent
# MAXIMUM LINE LENGTH
E501  # line too long
# etc
)

# stuff

flake8 --diff "${quotes}" --select="${errors[@]}" --ignore="${ignore[@]}"
```

scripts/travisci/check_flake8.sh
ryanjulian: what is the purpose of `"${quotes}"` and `"${ignore}"`? I don't see them defined anywhere.

gonzaiva:
There are some error codes to find for [Programming Recommendations](https://www.python.org/dev/peps/pep-0008/#programming-recommendations), but most verifications can be done with flake8 thanks to the support packages flake8-docstrings and pep8-naming. I think we could merge this script with the script for import-order since it's the same lint tool.
Some verifications that cannot be performed neither by flake8 nor by pylint so far are:
- [Module dunder names](https://www.python.org/dev/peps/pep-0008/#module-level-dunder-names)
- [Spaces around -> in function annotations](https://www.python.org/dev/peps/pep-0008/#other-recommendations)
- [Combining an argument annotation with a default value: spaces sorrunding the
 11   = sign are not verified](https://www.python.org/dev/peps/pep-0008/#other-recommendations)
- [Trailing commas are not verified] (https://www.python.org/dev/peps/pep-0008/#when-to-use-trailing-commas)
@ryanjulian, is it ok if we leave these verifications out?
Also, PEP8 does not enforce the use of double or single quotes except for docstrings, but asks for consistency. However, I understand that certain circumstances may require the use of both as specified in the section "Code Style" of CONTRIBUTING.md, so only the quote verification for docstrings was enabled.

scripts/travisci/check_pylint.sh
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
+pylint --disable=all --enable=C0301,C0330 $(git diff HEAD HEAD~1 --name-only | grep "\.py")
```
ryanjulian: can you add the errors/ignore and incremental vs absolute features to this file just like flake8? that will make them easy to maintain.

scripts/travisci/check_flake8.sh
```diff
@@ -0,0 +1,146 @@
...
+# Exit status of this script
+exit_status=0
+
+# Add an option to run the incremental verification locally to reduce the time
+# spent in travis for error verification. It's required to rebase the feature
+# branch with the master branch to avoid checking on unknown changes.
+while getopts ":l" opt; do
+  case ${opt} in
+    l)
+      TRAVIS_PULL_REQUEST="false"
```
ryanjulian: you can check the TRAVIS or CI variables instead:
https://docs.travis-ci.com/user/environment-variables/#Convenience-Variables

scripts/travisci/check_flake8.sh
```diff
@@ -0,0 +1,146 @@
...
+  esac
+done
+
+# ABSOLUTE VERIFICATION:
+# The following errors are analyzed in all python files across the repository.
+#
+# TABS
+# W191 indentation contains tabs
+# E101 indentation contains mixed spaces and tabs
+errors="E101,W191,"
```
ryanjulian: Please split this into errors_incremental and errors_absolute for ease of maintenance. You can concatenate them when when you call the script.

ryanjulian:
@gonzaiva All of that sounds fine to me. Please leave documentation in the script for what is not checked (including the rules you turned off). That will make it easy to try turning on rules in the future.

ryanjulian:
Alternatively to .pylintrc and .flake8, you can put all the configs for local checks in `setup.cfg`, so we can avoid a nest of config files.

Be sure to include a `[yapf]` section.

See: http://renesd.blogspot.com/2017/02/setupcfg-solution-to-python-config-file.html

scripts/travisci/check_flake8.sh
gonzaiva: Done

scripts/travisci/check_flake8.sh
gonzaiva: Done.

scripts/travisci/check_flake8.sh
gonzaiva: Done.

scripts/travisci/check_flake8.sh
gonzaiva: Done.

scripts/travisci/check_flake8.sh
gonzaiva: Done

gonzaiva: Variables removed.

scripts/travisci/check_flake8.sh
gonzaiva: Done

scripts/travisci/check_pylint.sh
gonzaiva: Since error codes C0301 C0330 are for indentation but the yapf corrections for indentation conflict with pylint, I think it's better to remove it. Other checks now can be done with flake8 thanks to the support packages for docstring and naming conventions.

scripts/travisci/check_flake8.sh
```diff
@@ -0,0 +1,84 @@
...
+)
+
+errors_absolute="${errors_absolute[@]}"
+flake8 --select="${errors_absolute// /,}"
+
+
+# INCREMENTAL VERIFICATION:
+# The following errors are analyzed only in the modified code in the pull
+# request branch.
+errors_indentation=(
```
ryanjulian: `errors_incremental`?

scripts/travisci/check_flake8.sh
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/env bash
+# ABSOLUTE VERIFICATION:
+# The following errors are analyzed in all python files across the repository.
+errors_absolute=(
```
ryanjulian: can you include a list of errors which are ignored? that would make the file self-documenting.

PR #54, opened 2018-06-11
Moved sandbox.rocky.tf to rllab.tf

jonashen:
Upgraded tf to first-class citizenship. One caveat to note is a bug in yapf formatting of a couple files that results in CI failing. References here and here. In short, CI fails when encountering a comma after *args or **kwargs in code.


rllab/tf/launchers/vpg_cartpole.py
```diff
@@ -1,10 +1,10 @@
 from rllab.baselines import LinearFeatureBaseline
 from rllab.envs import normalize
 from rllab.envs.box2d import CartpoleEnv
-from rllab.misc import run_experiment_lite
-from sandbox.rocky.tf.algos import VPG
-from sandbox.rocky.tf.envs import TfEnv
-from sandbox.rocky.tf.policies import GaussianMLPPolicy
+from rllab.misc import stub, run_experiment_lite
```
eric-heiden: `stub` isn't used here

ryanjulian:
You probably want to wait fo #59 and rebase on that. Sorry for taking up your time fixing the imports--it was a much bigger job than I anticipated.

I'm pretty sure #59 fixes your YAPF problems by removing the offending commas. I think we should rewrite that code, because it is in fact a parse error in Python 3.5.2.

Please check for circular imports using my script before submitting. I'm not sure I will be able to get the CI to run it tonight. If not, I will commit the script to `scripts/travisci/check_imports.sh` but not have the CI run it.

Issue #55, opened 2018-06-11
Cross-platform support for TF asynchronous plotting


jonashen:
The current implementation of async plotting uses Multithreading. However, using multiprocessing will throw segmentation for Mac OSX machines, but we want to prioritize multiprocessing for Linux machines. Therefore, write code that will support both implementations.

PR #56, opened 2018-06-11
Added asynchronous plotting to TensorFlow

jonashen:
Unfortunately, async plotting for TensorFlow requires the use of threading.Thread instead of multiprocessing.Process. Thus, on Linux machines, there will be a slight delay when you run the algorithm that seems like nothing is happening, but once the window loads everything works perfectly and asynchronously.

See: #55


rllab/misc/tabulate.py
```diff
@@ -1,8 +1,9 @@
 """Pretty-print tabular data."""
 
 from collections import namedtuple
+import functools
```
gonzaiva: It may be faster to only import reduce like `from itertools import reduce`

ryanjulian:
You may want to wait on #59

sandbox/rocky/tf/algos/batch_polopt.py
```diff
@@ -165,3 +167,7 @@ def get_itr_snapshot(self, itr, samples_data):
 
     def optimize_policy(self, itr, samples_data):
         raise NotImplementedError
+
+    def update_plot(self):
```
gonzaiva: Not used. Remove.

sandbox/rocky/tf/plotter/plotter.py
```diff
@@ -0,0 +1,108 @@
...
+__all__ = ["Plotter"]
+
+
+class Op(Enum):
+    STOP = 0
+    UPDATE = 1
+    DEMO = 2
+
+
+Message = namedtuple("Message", ["op", "args", "info"])
```
ryanjulian: info? the python convention would be kwargs (ignore what gym does)

sandbox/rocky/tf/plotter/plotter.py
```diff
@@ -0,0 +1,108 @@
...
+import platform
+from queue import Empty, Queue
+from threading import Thread
+
+import numpy as np
+import tensorflow as tf
+
+from rllab.sampler.utils import rollout
+
+__all__ = ["Plotter"]
```
gonzaiva: This is not required if  * is not used in the import. See [here](https://docs.python.org/3.5/tutorial/modules.html#importing-from-a-package) for further information.

sandbox/rocky/tf/plotter/plotter.py
ryanjulian: To add more color, the intention of the interface is this.

All python function definitions can be written as
```python
def foo(*args, **kwargs):
    # do some stuff
```
where `args` is an ordered list of unnamed arguments and `kwargs` is a dictionary of keyword arguments. You can even call any function like this, even if it wasn't declared this way.
```
def foo(bar, baz=0):
    return bar + baz

args = [1]
kwargs = {"baz": 1}
print(foo(*args, **kwargs)) # prints 2
```

Now imagine that this interface gets complicated, and we split the implementation of DEMO, UPDATE, and STOP into functions.
If `Message` has op, args, and kwargs, we can implement dispatch for *any function*:
```python
if Op.FOO in msgs:
    msg = msgs[Op.FOO]
    do_foo(*msg.args, **msg.kwargs)
```

ryanjulian:
I apologize for asking you to fix the imports for just this PR--I didn't realize how many files you'd have to touch.

rllab/tf/algos/npo.py
```diff
@@ -28,7 +28,7 @@ def __init__(self,
         self.name = name
         super(NPO, self).__init__(**kwargs)
 
-    @overrides
+    # @overrides
```
ryanjulian: uhh why is this commented out?

rllab/tf/algos/npo.py
```diff
@@ -105,7 +105,7 @@ def init_opt(self):
                 constraint_name="mean_kl")
             return dict()
 
-    @overrides
+    # @overrides
```
ryanjulian: why commented out?

PR #61, opened 2018-06-12
Update .gitignore

ryanjulian:


.gitignore
```diff
@@ -36,4 +36,5 @@ blackbox.zip
 blackbox
 rllab/config_personal.py
 *.swp
-sandbox
+sandbox/
```
jonashen: Once we finish porting sandbox.rocky.tf to rllab.tf, this will not be necessary right?

ryanjulian: as long as garage uses conda instead of pip it will probably be necessary. i keep my private research code in a sandbox checked out into a sandbox (and so do most people).

jonashen: sgtm

ryanjulian: you forgot to approve :P

jonashen: whoops, my bad

Issue #18, opened 2018-06-11
DDPG for TensorFlow


ryanjulian:
Original paper:
https://arxiv.org/abs/1509.02971

Implementation in the Theano tree:
https://github.com/ryanjulian/rllab/blob/integration/rllab/algos/ddpg.py

OpenAI baselines implementation:
https://github.com/openai/baselines/tree/master/baselines/ddpg

blog post (there are many other resources
http://pemami4911.github.io/blog/2016/08/21/ddpg-rl.html

Sketch:
* Provide an implementation in sandbox/rocky/tf/ddpg.py
* Add any needed primitives to rllab/ and sandbox/rocky/tf/
* Provide a regression test (of the reward curve) against the openai/baselines implementation

Imported from https://github.com/ryanjulian/rllab/issues/26

PR #66, opened 2018-06-13
Add DDPG to TensorFlow

CatherineSue:
Deep Deterministic Policy Gradients uses a stochastic behavior policy
for exploration and a deterministic target policy to perform policy
gradients. It uses actor-critic method to improve current policy. 

The DDPG includes:
 - DDPG algorithm, inherits from RLAlgorithm.
 - Replay buffer, the memory class of DDPG. Used to store transitions
   and random sample transitions in a env. This helps to minimize
   correlations between smaples.
 - OU exploration strategy. Used to add exploration noise to selected
   action output by policy network. Also helps to improve the
   performance.
 - ContinuousMLPPolicy. The policy network of DDPG.
 - ContinuousMLPQFunction. The q value network of DDPG.

Refer to: #18


rllab/tf/exploration_strategy/ou_strategy.py
```diff
@@ -0,0 +1,21 @@
+import numpy as np
+
+
+class OUStrategy:
```
zhanpenghe: Please inherit rllab.exploration_strategies.ExplorationStrategy

ryanjulian:
Please verify the results from this algo carefully against some from OpenAI gym. The Theano implementation of DDPG from rllab is known to be flawed.

rllab/tf/algos/ddpg.py
```diff
@@ -0,0 +1,319 @@
...
+            ema_var = ema.average(var)
+            return ema_var if ema_var else var
+
+        # Create target actor and critic network
+        target_actor = pickle.loads(pickle.dumps(self.actor))
+        target_critic = pickle.loads(pickle.dumps(self.critic))
+
+        # Set up the target actor and critic network
+        target_actor.build_net(reuse=True, custom_getter=ema_getter)
+        target_critic.build_net(reuse=True, custom_getter=ema_getter)
```
zhanpenghe: Why is reuse True here?

rllab/tf/algos/ddpg.py
```diff
@@ -0,0 +1,319 @@
...
+
+        replay_buffer = self.opt_info["replay_buffer"]
+        created_session = True if (sess is None) else False
+        if sess is None:
+            sess = tf.Session()
+            sess.__enter__()
+
+        sess.run(tf.global_variables_initializer())
+
+        state = self.env.reset()
```
zhanpenghe: To conform with the naming convention of other algos, it would be better to name this as obs.

rllab/tf/algos/ddpg.py
```diff
@@ -0,0 +1,319 @@
...
+        epoch_critic_losses = []
+        epoch_ys = []
+        epoch_qs = []
+        epoch_episodes = 0
+
+        for epoch in range(self.n_epochs):
+            for step in range(self.n_epoch_cycles):
+                for rollout in range(self.n_rollout_steps):
+                    if self.plot:
+                        self.env.render()
```
zhanpenghe: Please use the tf async plotter if possible because synchronized plotting will slow down the training process.

rllab/tf/algos/ddpg.py
```diff
@@ -0,0 +1,319 @@
...
+        action = self.actor.get_action(np.array(state))
+        noise = self.es.get()
+        action = action + noise
+        action = np.clip(action, self.action_range[0], self.action_range[1])
+        return action
+
+    @overrides
+    def train(self, sess=None):
+        self._initialize()
+
```
zhanpenghe: Shouldn't this be moved to __init__()?

rllab/tf/algos/ddpg.py
```diff
@@ -0,0 +1,319 @@
+from collections import deque
+import pickle as pickle
```
jonashen: `import pickle` should be fine

rllab/tf/algos/ddpg.py
```diff
@@ -0,0 +1,319 @@
...
+        # Initialize replay buffer
+        replay_buffer = ReplayBuffer(self.replay_buffer_size,
+                                     self.observation_dim, self.action_dim)
+
+        # Set up critic training function
+        y = tf.placeholder(tf.float32, name="y")
+        states = tf.placeholder(
+            tf.float32, shape=(None, self.observation_dim), name="state")
+        actions = tf.placeholder(
+            tf.float32, shape=(None, self.action_dim), name='action')
```
jonashen: double quotes?

CatherineSue:
TODO: Add plotter in garage/tf/algos/ddpg.py.

CatherineSue:
## Regression test of reward curve
image

ryanjulian:
@CatherineSue great results! what task was this on?

Can you post comparison results for the "Mujoco1M" suite in openai baselines?
https://github.com/openai/baselines/blob/master/baselines/bench/benchmarks.py

```python
from baselines.bench import benchmarks
import gym

mujoco1m = benchmarks.get_benchmark()

for task in mujoco1m["tasks"]:
    env = gym.make(task["env_id"])
    for i in task["trials"]:
        env.reset()

        # run/log baselines results
        # run/log garage results
```

CatherineSue:
# Benchmarks for Mujoco1M

Here is the plot of benchmarks Mujoco1M. Each plot represents the average reward comparison between ddpg-garage(using primitives from garage) and ddpg-baselines. There are seven tasks in Mujoco1M.

## InvertedDoublePendulum-v2
benchmark

## InvertedPendulum-v2
benchmark

## HalfCheetah-v2
benchmark

## Hopper-v2
benchmark

## Walker2d-v2
benchmark

## Reacher-v2
benchmark

## Swimmer-v2
Running.

garage/tf/algos/ddpg.py
```diff
@@ -0,0 +1,371 @@
...
+from garage.tf.plotter import Plotter
+from garage.tf.replay_buffer import ReplayBuffer
+
+
+class DDPG(RLAlgorithm):
+    """
+    A DDPG model based on https://arxiv.org/pdf/1509.02971.pdf.
+
+    Example:
+        $ python rllab/tf/launchers/ddpg_pendulum.py
```
ryanjulian: garage/tf

garage/tf/algos/ddpg.py
```diff
@@ -0,0 +1,371 @@
...
+            plot(bool): Whether to visualize the policy performance after each
+         eval_interval.
+            pause_for_plot(bool): Whether to pause before continuing when
+         plotting.
+            actor_optimizer(): Optimizer for training policy network.
+            critic_optimizer(): Optimizer for training q function network.
+        """
+        self.env = env
+
+        self.observation_dim = env.observation_space.shape[-1]
```
ryanjulian: i think you want to use `flat_dim` from `garage.envs.utils`

garage/tf/algos/ddpg.py
```diff
@@ -0,0 +1,371 @@
...
+        self.action_dim = env.action_space.shape[-1]
```
ryanjulian: `flat_dim`

garage/tf/algos/ddpg.py
```diff
@@ -0,0 +1,371 @@
...
+            sess.close()
+
+    def _initialize(self):
+        """Set up the actor, critic and target network."""
+        # Set up the actor and critic network
+        self.actor.build_net()
+        self.critic.build_net()
+
+        # Create target actor and critic network
+        target_actor = pickle.loads(pickle.dumps(self.actor))
```
ryanjulian: why `pickle.loads(pickle.dumps())`?

garage/tf/policies/continuous_mlp_policy.py
```diff
@@ -0,0 +1,146 @@
...
+        Set up q network based on class attributes.
+
+        This function uses layers defined in rllab.tf.
+
+        Args:
+            reuse: A bool indicates whether reuse variables in the same scope.
+            custom_getter: A customized getter object used to get variables.
+            trainable: A bool indicates whether variables are trainable.
+        """
+        trainable = True if trainable is None else trainable
```
ryanjulian: just set `trainable=True` in the prototype for `build_net()`

garage/tf/algos/ddpg.py
CatherineSue: To copy actor to target_actor. Same as `target_actor = copy(self.actor)`.

tests/regression_tests/test_benchmark_ddpg.py
```diff
@@ -0,0 +1,268 @@
...
+    Create garage model and training.
+
+    Replace the ddpg with the algorithm you want to run.
+
+    :param env: Environment of the task.
+    :param seed: Random seed for the trail.
+    :param log_dir: Log dir path.
+    :return:
+    """
+    env.seed(seed)
```
ryanjulian:
```python
from garage.misc import ext
ext.set_seed(seed)
```

tests/regression_tests/test_benchmark_ddpg.py
```diff
@@ -0,0 +1,268 @@
...
+
+    Replace the ddpg and its training with the algorithm you want to run.
+
+    :param env: Environment of the task.
+    :param seed: Random seed for the trail.
+    :param log_dir: Log dir path.
+    :return
+    """
+    rank = MPI.COMM_WORLD.Get_rank()
+    env.seed(seed)
```
ryanjulian: you need to set the tensorflow seed as well. i think it's `tf.set_random_seed()`

garage/tf/algos/ddpg.py
ryanjulian: If that is the case, then please use copy() so it is clearer.

CatherineSue:
There are two reasons why baselines curve looks so smooth in the earlier plots. 
 - Baselines normalize observations and rewards with tf.clip_by_value. (This may not have much effect on the variance, but according to my experiments, it varies the performance of baselines.)
 - Baselines doesn't clear epoch_reward, the list contains all the rewards in one epoch, after the epoch finishes. So the average return curve of baselines is actually the average return from epoch 1 to current epoch. That's why it doesn't vary at all.

Here is a benchmark regarding baselines and garage after I manually set the epoch_reward=[] at the end of each epoch. We can see that the baselines also have high variance. It also indicates that the garage performance may not outperform baselines. But I've changed a lot of code in baselines script, so it may be not accurate.

### Benchmark for InvertedDoublePendulum-v2 in 1M steps
screenshot from 2018-06-28 01-18-52

To further run benchmark, the better solution is to set the epoch_reward in garage the same as baselines. I will rerun the test.

CatherineSue:
# Updated Benchmarks for Mujoco1M

Here is the updated plot of benchmarks Mujoco1M. 

## InvertedDoublePendulum-v2
benchmark

## InvertedPendulum-v2
benchmark

## HalfCheetah-v2


## Hopper-v2


## Walker2d-v2


## Reacher-v2


## Swimmer-v2

CatherineSue:
# Updated Benchmarks for Mujoco1M

Here is the updated plot of benchmarks Mujoco1M. 
```
params = {
    "actor_lr": 1e-4,
    "critic_lr": 1e-3,
    "actor_hidden_sizes": [64, 64],
    "critic_hidden_sizes": [64, 64],
    "n_epochs": 500,
    "n_epoch_cycles": 20,
    "n_rollout_steps": 100,
    "n_train_steps": 50,
    "discount": 0.9,
    "tau": 1e-2,
    "replay_buffer_size": int(1e6),
    "sigma": 0.2,
}
```

## InvertedDoublePendulum-v2
benchmark

## InvertedPendulum-v2
benchmark

## HalfCheetah-v2
benchmark

## Hopper-v2
benchmark

## Walker2d-v2
Running

## Reacher-v2
Running

## Swimmer-v2
Running

garage/tf/algos/ddpg.py
```diff
@@ -0,0 +1,372 @@
...
+        episode_actor_losses = []
+        episode_critic_losses = []
+        episodes = 0
+        epoch_ys = []
+        epoch_qs = []
+
+        for epoch in range(self.n_epochs):
+            logger.push_prefix('epoch #%d | ' % epoch)
+            logger.log("Training started")
+            for epoch_cycle in pyprind.prog_bar(range(self.n_epoch_cycles)):
```
ryanjulian: Please make a github issue to change this to use VectorizedSampler

garage/tf/algos/ddpg.py
```diff
@@ -0,0 +1,372 @@
...
+                logger.record_tabular('Epoch', epoch)
+                logger.record_tabular('Episodes', episodes)
+                logger.record_tabular('AverageReturn',
+                                      np.mean(episode_rewards))
+                logger.record_tabular('StdReturn', np.std(episode_rewards))
+                logger.record_tabular('AveragePolicyLoss',
+                                      np.mean(episode_actor_losses))
+                logger.record_tabular('AverageQFunctionLoss',
+                                      np.mean(episode_critic_losses))
+                logger.record_tabular('AverageQ', np.mean(epoch_qs))
```
ryanjulian: Please namespace your logs according to a relevant component.

For instance, some things are a property of the Q function and can be logged under a QFunction namespace (e.g. "QFunction/AverageQLoss"). Other things are a property of the algo as a whole ans can be left at the top level (e.g. AverageReturn").

garage/tf/launchers/ddpg_pendulum.py
```diff
@@ -0,0 +1,69 @@
+"""
```
ryanjulian: can you leave a comment with on the launcher with the expected final AverageReturn and the approximate rise time (in iters) to achieve that return?

garage/tf/policies/continuous_mlp_policy.py
```diff
@@ -0,0 +1,146 @@
...
+    This class implements a policy network.
+
+    The policy network selects action based on the state of the environment.
+    It uses neural nets to fit the function of pi(s).
+    """
+
+    def __init__(self,
+                 env_spec,
+                 hidden_sizes=(64, 64),
+                 name="StochasticMLPPolicy",
```
ryanjulian: ContinuousMLPPolicy

garage/tf/policies/continuous_mlp_policy.py
```diff
@@ -0,0 +1,146 @@
...
+        self._env_spec = env_spec
+        self._obs_dim = flat_dim(env_spec.observation_space)
+        self._action_dim = flat_dim(env_spec.action_space)
+        self._action_bound = env_spec.action_space.high
+        self._hidden_sizes = hidden_sizes
+        self._hidden_nonlinearity = hidden_nonlinearity
+        self._output_nonlinearity = output_nonlinearity
+        self._batch_norm = bn
+
+    def build_net(self, reuse=None, custom_getter=None, trainable=None):
```
ryanjulian: private: _build_net

garage/tf/algos/ddpg.py
```diff
@@ -0,0 +1,376 @@
...
+            pause_for_plot(bool): Whether to pause before continuing when
+         plotting.
+            actor_optimizer(): Optimizer for training policy network.
+            critic_optimizer(): Optimizer for training q function network.
+        """
+        self.env = env
+
+        self.observation_dim = flat_dim(env.observation_space)
+        self.action_dim = flat_dim(env.action_space)
+        self.action_bound = env.action_space.high
```
jonashen: # from garage.envs.util import flat_dim
You can consider calling `_, self.action_bounds = bounds(env.action_space)`

CatherineSue: Thanks! Done.

Issue #79, opened 2018-06-14
Always provide a dtype to gym.spaces.Box


ryanjulian:
Otherwise it will complain endlessly (and slow down your program)

```
WARN: gym.spaces.Box autodetected dtype as <class 'numpy.float32'>. Please provide explicit dtype.
WARN: gym.spaces.Box autodetected dtype as <class 'numpy.float32'>. Please provide explicit dtype.
WARN: gym.spaces.Box autodetected dtype as <class 'numpy.float32'>. Please provide explicit dtype.
WARN: gym.spaces.Box autodetected dtype as <class 'numpy.float32'>. Please provide explicit dtype.
0% [#                             ] 100% | ETA: 00:00:09WARN: gym.spaces.Box autodetected dtype as <class 'numpy.float32'>. Please provide explicit dtype.
WARN: gym.spaces.Box autodetected dtype as <class 'numpy.float32'>. Please provide explicit dtype.
0% [##                            ] 100% | ETA: 00:00:09WARN: gym.spaces.Box autodetected dtype as <class 'numpy.float32'>. Please provide explicit dtype.
WARN: gym.spaces.Box autodetected dtype as <class 'numpy.float32'>. Please provide explicit dtype.
0% [###                           ] 100% | ETA: 00:00:09WARN: gym.spaces.Box autodetected dtype as <class 'numpy.float32'>. Please provide explicit dtype.
WARN: gym.spaces.Box autodetected dtype as <class 'numpy.float32'>. Please provide explicit dtype.
WARN: gym.spaces.Box autodetected dtype as <class 'numpy.float32'>. Please provide explicit dtype.
WARN: gym.spaces.Box autodetected dtype as <class 'numpy.float32'>. Please provide explicit dtype.
0% [####                          ] 100% | ETA: 00:00:08WARN: gym.spaces.Box autodetected dtype as <class 'numpy.float32'>. Please provide explicit dtype.
WARN: gym.spaces.Box autodetected dtype as <class 'numpy.float32'>. Please provide explicit dtype.
0% [#####                         ] 100% | ETA: 00:00:08WARN: gym.spaces.Box autodetected dtype as <class 'numpy.float32'>. Please provide explicit dtype.
WARN: gym.spaces.Box autodetected dtype as <class 'numpy.float32'>. Please provide explicit dtype.
[...]

Issue #74, opened 2018-06-14
TF policies are broken


zhanpenghe:
Some of the policies in tf tree are broken due to #58. Many spaces are converted to gym.spaces but the policies are still checking if the action_space is rllab.tf.spaces.

E.g. `rllab.tf.gaussian_mlp_policy.GaussianMLPPolicy`

ryanjulian:
Fixed in #81

PR #81, opened 2018-06-15
All policies now check for gym.spaces.Box

jonashen:
Per title. All policies check for gym Box and not rllab.spaces.Box classes.

Ref: #74, #79, #82


jonashen:
Yes, and also using VSC's built-in ag equivalent.

PR #83, opened 2018-06-15
Check docstrings only on added files

gonzaiva:
The docstring checks are only applied to added files to avoid
errors when refactoring legacy code.
Also, other checks are only applied to modified files to avoid errors on
copied, deleted or renamed files.
The name rllab was replaced for garage in option --application-import-
names.
Dots were missing in grep expression at check_pylint.

### DIFF
--- a/scripts/travisci/check_flake8.sh
+++ b/scripts/travisci/check_flake8.sh
@@ -31,7 +31,7 @@ flake8 --isolated --select="${errors_absolute// /,}" \
 # INCREMENTAL VERIFICATION:
 # The following errors are analyzed only in the modified code in the pull
 # request branch.
-errors_incremental=(
+errors_change=(
 # INDENTATION
 E125 # continuation line with same indent as next logical line
 E128 # continuation line under-indented for visual indent
@@ -80,7 +80,7 @@ E731 # do not assign a lambda expression, use a def
 
 # Add the codes of the errors to be ignored for the absolute verification in
 # this array.
-ignored_errors_incremental=(
+ignored_errors_change=(
 )
 
 # DOCSTRING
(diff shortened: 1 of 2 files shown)


scripts/travisci/check_flake8.sh
```diff
@@ -89,21 +83,37 @@ E731 # do not assign a lambda expression, use a def
 ignored_errors_incremental=(
 )
 
+# DOCSTRING
+# The following error code enables all the error codes for docstring defined
+# here:
+# http://pep257.readthedocs.io/en/latest/error_codes.html
+# If one of the errors needs to be ignored, just add it to the ignore array.
+errors_incremental_doc="${errors_incremental[@]} D"
```
ryanjulian: let's call these "errors_add" and "errors_change" to be clear

scripts/travisci/check_flake8.sh
```diff
@@ -89,21 +83,37 @@ E731 # do not assign a lambda expression, use a def
...
+
 errors_incremental="${errors_incremental[@]}"
 ignored_errors_incremental="${ignored_errors_incremental[@]}"
 if [[ "${TRAVIS_PULL_REQUEST}" != "false" && "${TRAVIS}" == "true" ]]; then
-  git diff "${TRAVIS_COMMIT_RANGE}" \
+  git diff "${TRAVIS_COMMIT_RANGE}" --diff-filter=M \
```
ryanjulian: This script would be much easier to read if you split it into 2 stages

1) Build lists of files changed or added, e.g. `files_changed` `files_added`
1) Apply different checks to those files

As you have implemented it, you're actually only running flake8 over *lines changed* in both of those cases (that's what `flake8 diff` does).

gonzaiva: Done

scripts/travisci/check_flake8.sh
gonzaiva: Done

PR #89, opened 2018-06-15
Remove quotes in filename variables in pep8 checks

gonzaiva:
Both the scripts for pylint and flake8 were doing the correct filename
subtitution but with added quotes that were passed to corresponding
command for pylint and flake8, not checking any file at all. For
example, the bad command produced was:
pylint 'garage/algos/cma_es_lib.py garage/envs/dm_control_env.py'
Instead of:
pylint garage/algos/cma_es_lib.py garage/envs/dm_control_env.py
This commit fixes the problem. Also the grep command is used to filter
the python files, because flake8 and pylint may check on all the files
passed to them.

### DIFF
--- a/scripts/travisci/check_flake8.sh
+++ b/scripts/travisci/check_flake8.sh
@@ -94,28 +94,30 @@ errors_change="${errors_change[@]}"
 ignored_errors_change="${ignored_errors_change[@]}"
 if [[ "${TRAVIS_PULL_REQUEST}" != "false" && "${TRAVIS}" == "true" ]]; then
   files_changed="$(git diff "${TRAVIS_COMMIT_RANGE}" --diff-filter=M \
-    --name-only)"
+                     --name-only | grep ".*.py")"
   files_added="$(git diff "${TRAVIS_COMMIT_RANGE}" --diff-filter=A \
-    --name-only)"
+                   --name-only | grep ".*.py")"
   flake8 --isolated --import-order-style=google \
     --application-import-names=sandbox,garage,examples,contrib \
     --select="${errors_change// /,}" \
-    --ignore="${ignored_errors_change// /,}" "${files_changed}"
+    --ignore="${ignored_errors_change// /,}" ${files_changed}
   flake8 --isolated --import-order-style=google \
     --application-import-names=sandbox,garage,examples,contrib \
     --select="${erros_add// /,}" \
-    --ignore="${ignored_errors_change// /,}" "${files_added}"
+    --ignore="${ignored_errors_change// /,}" ${files_added}
 else
   git remote set-branches --add origin master
   git fetch
-  files_changed="$(git diff origin/master --diff-filter=M --name-only)"
-  files_added="$(git diff origin/master --diff-filter=A --name-only)"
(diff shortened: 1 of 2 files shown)


ryanjulian:
Related: see https://github.com/rlworkgroup/garage/blob/clarify_flake8/scripts/travisci/check_flake8.sh which I was just working on.

ryanjulian:
Can you add the offending rules to CHANGED or ADDED, to keep the CI green?

scripts/travisci/check_flake8.sh
```diff
@@ -17,21 +23,20 @@ E702 # multiple statements on one line (semicolon)
...
+exclude_all=(
+./tests/'*'
+)
 
 
-# INCREMENTAL VERIFICATION:
-# The following errors are analyzed only in the modified code in the pull
-# request branch.
-errors_change=(
+### CHANGED_FILES ###
```
ryanjulian: Why an underscore?

scripts/travisci/check_flake8.sh
```diff
@@ -78,44 +83,174 @@ E722 # do not use bare except, specify exception instead
...
+
+################################################################################
+
 # DOCSTRING
 # The following error code enables all the error codes for docstring defined
 # here:
 # http://pep257.readthedocs.io/en/latest/error_codes.html
 # If one of the errors needs to be ignored, just add it to the ignore array.
-erros_add="${errors_change[@]} D"
+#erros_add="${errors_changed[@]} D"
```
ryanjulian: dead code/comments here?

scripts/travisci/check_flake8.sh
```diff
@@ -78,44 +83,174 @@ E722 # do not use bare except, specify exception instead
...
 
-errors_change="${errors_change[@]}"
-ignored_errors_change="${ignored_errors_change[@]}"
 if [[ "${TRAVIS_PULL_REQUEST}" != "false" && "${TRAVIS}" == "true" ]]; then
```
ryanjulian: Tell the reader what you're doing here--it's not obvious.

scripts/travisci/check_flake8.sh
```diff
@@ -78,44 +83,174 @@ E722 # do not use bare except, specify exception instead
...
-    --application-import-names=sandbox,garage,examples,contrib \
-    --select="${erros_add// /,}" \
-    --ignore="${ignored_errors_change// /,}" "${files_added}"
+  files_changed="$(git diff origin/master --diff-filter=M --name-only \
+                     | grep ".*\.py")"
+  files_added="$(git diff origin/master --diff-filter=A --name-only \
+                   | grep ".*\.py")"
+fi
+
+test_files_changed="$(echo "${files_changed}" |  grep "tests/*")"
```
ryanjulian: did `flake8 --exclude=tests ...` not work?

gonzaiva: Yes, the first three checks exclude "./tests/*", and the last three only check on "./tests/*". How can use --exclude in the last three checks that are only for "./tests*"? Some inverted regex?

PR #94, opened 2018-06-15
Remove blank config_personal.py

ryanjulian:


ryanjulian:
This is causing launcher errors when Python finds an empty config_personal.py.

Issue #95, opened 2018-06-15
In contrib.ros 'from garage.config_personal' should be 'from garage.config'


hjzh4:

hjzh4:
Addressed

PR #96, opened 2018-06-16
Fix contrib.ros 'from garage.config_personal' should be 'from garage.config'

hjzh4:
- Refer to #95

### DIFF
--- a/contrib/ros/worlds/block_world/world.py
+++ b/contrib/ros/worlds/block_world/world.py
@@ -13,8 +13,9 @@ import garage.misc.logger as logger
 try:
     from garage.config import VICON_TOPICS
 except ImportError:
-    logger.log("If you want to use real robot please set "
-               "VICON_TOPICS in garage.config_personal.py, otherwise skip this")
+    logger.log(
+        "If you want to use real robot please set "
+        "VICON_TOPICS in garage.config_personal.py, otherwise skip this")
 
 
 class Block(object):
--- a/contrib/ros/worlds/block_world/world.py
+++ b/contrib/ros/worlds/block_world/world.py
@@ -13,9 +13,13 @@ import garage.misc.logger as logger
 try:
     from garage.config import VICON_TOPICS
 except ImportError:
-    logger.log(
-        "If you want to use real robot please set "
-        "VICON_TOPICS in garage.config_personal.py, otherwise skip this")
+    raise NotImplementedError(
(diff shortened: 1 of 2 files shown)


contrib/ros/worlds/block_world/world.py
```diff
@@ -9,8 +9,12 @@
 
 from contrib.ros.worlds.gazebo import Gazebo
 from contrib.ros.worlds.world import World
-from garage.config_personal import VICON_TOPICS
 import garage.misc.logger as logger
+try:
+    from garage.config import VICON_TOPICS
+except ImportError:
+    logger.log("If you want to use real robot please set "
```
ryanjulian: "in config_personal.py"

hjzh4: Well, I'm a little confused that are we still using config_personal.py instead of config.py.
Should it be:
```
try:
    from garage.config_personal import VICON_TOPICS
except ImportError:
    logger.log("If you want to use real robot please set "
                    "VICON_TOPICS in config_personal.py, otherwise skip this")
```

ryanjulian: `config.py`:
```python
from garage.config_personal import *
```

`config.py` is a Python file containing configuration information for garage. Any garage module needing configuration information imports config.py and expects all configs to be there.

Some configuration items are either user-specific or need to be overridden by users. Those items are put in `config_personal.py`, which is automatically copied for each user on first run from `config_personal_template.py`. Anything in `config_personal.py` overwrites the default setting in `config.py`

garage maintainers edit `config.py`/`config_personal_template.py` but users should only ever edit their individual `config_personal.py`.

contrib/ros/worlds/block_world/world.py
```diff
@@ -9,8 +9,13 @@
...
 from contrib.ros.worlds.gazebo import Gazebo
 from contrib.ros.worlds.world import World
-from garage.config_personal import VICON_TOPICS
 import garage.misc.logger as logger
+try:
+    from garage.config import VICON_TOPICS
+except ImportError:
+    logger.log(
+        "If you want to use real robot please set "
+        "VICON_TOPICS in garage.config_personal.py, otherwise skip this")
```
ryanjulian: "garage/config_personal.py"

ryanjulian: Also, how can I skip an `ImportError`?

ryanjulian:
Please address final comments before submit

PR #104, opened 2018-06-26
Implement DDQN

gonzaiva:
An optimal policy is derived by selecting the highest-valued action in
each state of the environment, namely, using Q-learning. Although a
policy is not required since it's obtained from the Q-value function,
Q-learning is limited to act in discrete action spaces.
DDQN has two important ingredients:                                             
    - A target network. Although the online network makes the
    predictions, the target network helps in the training of the online
    network. The parameters of the target networks are copied every
    tau steps from the online network and kept fixed on all other steps.
    This reduces the overestimations produced in Q-learning.                                     
    - Experience replay. The observed transitions are stored for some
    time and sampled uniformly from a memory bank to update the
    network.

### DIFF
--- a/garage/algos/ddqn.py
+++ b/garage/algos/ddqn.py
@@ -1,0 +1,339 @@
+"""Double Deep Q Network.
+
+An optimal policy is derived by selecting the highest-valued action in each
+state of the environment, namely, using Q-learning. Although a policy is
+not required since it's obtained from the Q-value function, Q-learning is
+limited to act in discrete action spaces. For continuous spaces see DDPQ.
+DDQN has two important ingredients:
+    - A target network. Although the online network makes the predictions,
+    the target network helps in the training of the online network. The
+    parameters of the target networks are copied every tau steps from the
+    online network and kept fixed on all other steps. This reduces the
+    overestimations produced in Q-learning.
+    - Experience replay. The observed transitions are stored for some time
+    and sampled uniformly from a memory bank to update the network.
+"""
+
+import lasagne.init as li
+import lasagne.layers as ll
+import lasagne.nonlinearities as lnl
+import lasagne.objectives as lo
+import lasagne.updates as lu
+import numpy as np
+import theano
+import theano.tensor as tensor
+
+from garage.algos import RLAlgorithm
+from garage.misc.overrides import overrides
+
+
+class DDQN(RLAlgorithm):
+    """Implementation of the DDQN algorithm."""
+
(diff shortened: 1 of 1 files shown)


ryanjulian:
Please post benchmark results comparing to OpenAI baselines. You can base it Chang's script. See my comments on her script about setting random seeds in garage and TF

garage/algos/ddqn.py
```diff
@@ -0,0 +1,339 @@
...
+            reward_episode += reward
+
+    def _copy_to_target(self):
+        """Copy weights from online network to target network."""
+        ll.set_all_param_values(self._target_nn,
+                                ll.get_all_param_values(self._online_nn))
+
+
+class MemoryBank:
+    """The memory bank is used for experience replay in DDQN."""
```
zhanpenghe: I think this can be moved to garage.replay_buffer

zhanpenghe: Also there is a replay_buffer in `garage.algos.ddpg` implementation already so if you can just move it out as a module and make use fo it, that would be great.

garage/algos/ddqn.py
```diff
@@ -0,0 +1,339 @@
...
+            Number of transitions to put in the memory batch.
+
+        """
+        random_indices = np.random.randint(0, self._num_transitions, n)
+        obs = [self._obs_arr[index] for index in random_indices]
+        obs_next = [self._obs_next_arr[index] for index in random_indices]
+        actions = [self._action_arr[index] for index in random_indices]
+        rewards = [self._reward_arr[index] for index in random_indices]
+        not_done = [self._not_done_arr[index] for index in random_indices]
+        return MemoryBatch(obs, obs_next, actions, rewards, not_done)
```
zhanpenghe: I think a dictionary would be enough for this purpose.

garage/algos/ddqn.py
```diff
@@ -0,0 +1,339 @@
...
+        self._action = tensor.ivector("Action")
+        self._reward = tensor.vector("Reward")
+        self._done = tensor.vector("Done")
+        # Build the networks
+        self._target_nn = self._create_network()
+        self._online_nn = self._create_network()
+        self._build_graph()
+
+    def _create_network(self):
+        """Create a neural network.
```
zhanpenghe: I think this should be moved to something like q_function. Please remember to create some reusable components.

ryanjulian:
Cleaning up. Please reopen when it's rebased and ready for review.

Issue #117, opened 2018-06-29
Use tf.name_scope() in TensorBoardOutput


ryanjulian:
Use a name_scope in TensorBoardOutput so that tf objects created by TensorBoard do not pollute the computation graph.

CatherineSue:
Does this happen when using record_histogram?

ryanjulian:
Yes.

![image](https://user-images.githubusercontent.com/1376219/42108034-395ab6bc-7b8e-11e8-9fe8-18aefcd65808.png)

ryanjulian:
![image](https://user-images.githubusercontent.com/1376219/42247988-3e8d3158-7ed8-11e8-9bf5-03125570567b.png)

This seems to have messed up the key space for calls to `logger.record_histogram()`. `logger.record_histogram_by_type()` is unaffected.

CatherineSue:
<img width="718" alt="screen shot 2018-07-03 at 5 04 59 pm" src="https://user-images.githubusercontent.com/8605658/42250015-51c6c788-7ee3-11e8-8171-63e413a72f69.png">

```
logger.record_histogram("Embedding/samples/i_0", k_val)
logger.record_histogram("Embedding/samples/i_1", k_val)
logger.dump_tensorboard(step=i)
```
Does this look correct?
I'm not sure how you created i_0 and i_1.

ryanjulian:
That looks correct

Code generating the plot:
```python
        # samples
        num_traj = self.batch_size // self.max_path_length
        latents = samples_data["latents"][:num_traj, 0]
        for i in range(self.policy.latent_space.flat_dim):
            logger.record_histogram("Embedding/samples/i={}".format(i),
                                    latents[:, i])
```

PR #118, opened 2018-06-29
Add name_scope in TensorBoardOutput

CatherineSue:
The record_histogram and get_histogram method in TensorBoardOutput
wasn't group together, so all the variables created in them polluted the
computation graph. By adding a name_scope in the constructor, all
the nodes will be grouped under node TensorBoardOutput in the graph.

Also, update examples/example_tensorboard_logger.py.

See: #117

### DIFF
--- a/examples/example_tensorboard_logger.py
+++ b/examples/example_tensorboard_logger.py
@@ -1,10 +1,12 @@
-import numpy as np
+import tensorflow as tf
 
 from garage.misc import logger
 
 logger.set_tensorboard_dir("./histogram_example")
 N = 400
 for i in range(N):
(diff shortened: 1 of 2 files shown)


CatherineSue:
screen shot 2018-06-29 at 12 44 23 pm

ryanjulian:
Fixes #117

CatherineSue:
Before fixing the bug, if we created variables after calling get_histogram_by_type, the computation graph is like this:
screen shot 2018-06-29 at 1 56 34 pm

After fixing, the graph is:
screen shot 2018-06-29 at 1 47 56 pm

PR #119, opened 2018-06-29
Override plot parameter of algorithm constructor

gonzaiva:
When running a training session, there's two places to enable the
plotter: the algorithm constructor or the function run_experiment.
However, the latter must override the parameter set in the former.
The algorithm contains the plotter object, so it would be sufficient to
set the value of its plot parameter to the value set in the
run_experiment, but the constructor of the algorithm is passed within an
anonymous function to run_experiment and it's not possible to modify the
bytecode to change the plot parameter directly or to access the objects
after the anonymous function is called from run_experiment.
Instead, static methods were implemented in both the plotter of the
Theano and TF branch to disable or enable the plotter from
run_experiment.
Also, a cleanup of the plotter in the Theano branch was performed, and
the corresponding algorithm classes were modified to consider the new
guards and cleanup changes.
Finally, there's some legacy PEP8 style errors that where fixed.

### DIFF
--- a/examples/trpo_gym_cartpole.py
+++ b/examples/trpo_gym_cartpole.py
@@ -1,3 +1,5 @@
+import gym
+
 from garage.algos import TRPO
 from garage.baselines import LinearFeatureBaseline
 from garage.envs import normalize
@@ -22,7 +24,6 @@ def run_task(*_):
         policy=policy,
         baseline=baseline,
         batch_size=4000,
-        max_path_length=env.horizon,
         n_itr=50,
         discount=0.99,
         step_size=0.01,
--- a/examples/trpo_gym_cartpole.py
+++ b/examples/trpo_gym_cartpole.py
@@ -5,6 +5,7 @@ from garage.baselines import LinearFeatureBaseline
 from garage.envs import normalize
 from garage.envs.util import spec
 from garage.misc.instrument import run_experiment
+from garage.envs.util import horizon
 from garage.policies import CategoricalMLPPolicy
 
 
@@ -24,6 +25,7 @@ def run_task(*_):
         policy=policy,
         baseline=baseline,
         batch_size=4000,
+        max_path_length=horizon(env),
         n_itr=50,
         discount=0.99,
         step_size=0.01,
--- a/examples/trpo_gym_cartpole.py
+++ b/examples/trpo_gym_cartpole.py
@@ -3,9 +3,9 @@ import gym
 from garage.algos import TRPO
 from garage.baselines import LinearFeatureBaseline
 from garage.envs import normalize
+from garage.envs.util import horizon
 from garage.envs.util import spec
 from garage.misc.instrument import run_experiment
-from garage.envs.util import horizon
 from garage.policies import CategoricalMLPPolicy
 
 
--- a/garage/algos/batch_polopt.py
+++ b/garage/algos/batch_polopt.py
@@ -1,10 +1,8 @@
 from garage.algos import RLAlgorithm
 import garage.misc.logger as logger
-from garage.plotter import plotter
-from garage.policies import Policy
+from garage.plotter import Plotter
 from garage.sampler import parallel_sampler
 from garage.sampler.base import BaseSampler
-from garage.sampler.utils import rollout
 
 
 class BatchSampler(BaseSampler):
@@ -40,7 +38,8 @@ class BatchSampler(BaseSampler):
 class BatchPolopt(RLAlgorithm):
     """
     Base class for batch sampling-based policy optimization methods.
-    This includes various policy gradient methods like vpg, npg, ppo, trpo, etc.
+    This includes various policy gradient methods like vpg, npg, ppo, trpo,
+    etc.
     """
 
     def __init__(self,
@@ -111,13 +110,14 @@ class BatchPolopt(RLAlgorithm):
 
     def start_worker(self):
         self.sampler.start_worker()
-        if self.plot:
-            plotter.init_plot(self.env, self.policy)
 
     def shutdown_worker(self):
         self.sampler.shutdown_worker()
 
     def train(self):
+        plotter = Plotter()
+        if plotter.status(self.plot):
+            plotter.init_plot(self.env, self.policy)
         self.start_worker()
         self.init_opt()
         for itr in range(self.current_itr, self.n_itr):
@@ -135,12 +135,13 @@ class BatchPolopt(RLAlgorithm):
                 logger.save_itr_params(itr, params)
                 logger.log("saved")
                 logger.dump_tabular(with_prefix=False)
-                if self.plot:
-                    self.update_plot()
+                if plotter.status(self.plot):
+                    plotter.update_plot(self.policy, self.max_path_length)
                     if self.pause_for_plot:
                         input("Plotting evaluation run: Press Enter to "
                               "continue...")
 
+        plotter.shutdown()
         self.shutdown_worker()
 
     def log_diagnostics(self, paths):
@@ -163,7 +164,3 @@ class BatchPolopt(RLAlgorithm):
 
     def optimize_policy(self, itr, samples_data):
         raise NotImplementedError
-
-    def update_plot(self):
-        if self.plot:
-            plotter.update_plot(self.policy, self.max_path_length)
--- a/garage/algos/batch_polopt.py
+++ b/garage/algos/batch_polopt.py
@@ -96,7 +96,7 @@ class BatchPolopt(RLAlgorithm):
         self.max_path_length = max_path_length
         self.discount = discount
         self.gae_lambda = gae_lambda
-        self.plot = plot
+        self.plot = plot or Plotter.enable
         self.pause_for_plot = pause_for_plot
         self.center_adv = center_adv
         self.positive_adv = positive_adv
@@ -116,7 +116,7 @@ class BatchPolopt(RLAlgorithm):
 
     def train(self):
         plotter = Plotter()
-        if plotter.status(self.plot):
+        if self.plot:
             plotter.init_plot(self.env, self.policy)
         self.start_worker()
         self.init_opt()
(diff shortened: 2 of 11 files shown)


examples/trpo_gym_cartpole.py
```diff
@@ -22,7 +24,6 @@ def run_task(*_):
         policy=policy,
         baseline=baseline,
         batch_size=4000,
-        max_path_length=env.horizon,
```
CatherineSue: why not add `from garage.misc.util import horizon`, then `max_path_length=horizon(env)`?

gonzaiva: I didn't know that horizon was moved there. Thanks for letting me know Chang, will do.

CatherineSue: Np. Could you also add this in examples/trpo_gym_pendulum.py, please? It has the same problem. Thanks.

gonzaiva: Ok, will do!

garage/algos/cem.py
```diff
@@ -7,7 +7,9 @@
 from garage.core import Serializable
 import garage.misc.logger as logger
 from garage.misc.special import discount_cumsum
-import garage.plotter as plotter
+from garage.plotter import Plotter
+from garage.sampler import parallel_sampler
```
ryanjulian: where are parallel_sampler and stateful_pool used?

garage/plotter/plotter.py
```diff
@@ -1,88 +1,153 @@
...
+                    animated=True,
+                    speedup=5)
 
+            self._queue.put(
+                Message(op=Op.UPDATE, args=(env, policy), kwargs=None))
 
-def update_plot(policy, max_length=np.inf):
-    queue.put(['demo', policy.get_param_values(), max_length])
+    def update_plot(self, policy, max_length=np.inf):
+        if Plotter.enable is None or Plotter.enable:
```
ryanjulian: equivalent to `if Plotter.enable`

garage/plotter/plotter.py
```diff
@@ -1,88 +1,153 @@
...
+            if ('Darwin' in platform.platform()):
+                self._process = Thread(target=self._worker_start)
+            else:
+                self._process = Process(target=self._worker_start)
+            self._process.daemon = True
+            self._process.start()
+            atexit.register(self.shutdown)
+
+    def init_plot(self, env, policy):
+        if Plotter.enable is None or Plotter.enable:
```
ryanjulian: `if Plotter.enable`

garage/plotter/plotter.py
```diff
@@ -1,88 +1,153 @@
...
+            called, and the plot enabling of the run_experiment otherwise.
+
+        """
+        if Plotter.enable is None:
+            return plot_algo_status
+        else:
+            return Plotter.enable
+
+    def init_worker(self):
+        if Plotter.enable is None or Plotter.enable:
```
ryanjulian: `if Plotter.enable`

garage/plotter/plotter.py
```diff
@@ -1,88 +1,153 @@
...
+            pass
+
+    def shutdown(self):
+        if self._process.is_alive():
+            self._queue.put(Message(op=Op.STOP, args=None, kwargs=None))
+            self._queue.close()
+            self._process.join()
+
+    @staticmethod
+    def set_enable(enable):
```
ryanjulian: this function needs no arguments

garage/plotter/plotter.py
```diff
@@ -1,88 +1,153 @@
...
+        plot_algo_status : boolean
+            The plot enabling defined in the algorithm class.
+
+        Returns
+        -------
+            The plot enabling of the algorithm if the run_experiment was not
+            called, and the plot enabling of the run_experiment otherwise.
+
+        """
+        if Plotter.enable is None:
```
ryanjulian: in general you don't want to distinguish between NoneType and False in application logic. You are overloading None to mean "nobody has set this yet." 

I think you can get equivalent behavior by initializing Plotter.enable to true. Then if either the algorithm or run_experiment calls `Plotter.disable()` the plotter is permanently disabled (there would not need to be a Plotter.enable() API, but there could be).

garage/plotter/plotter.py
```diff
@@ -1,88 +1,153 @@
...
+            self._process.join()
+
+    @staticmethod
+    def set_enable(enable):
+        """Set the plot enabling according to the run_experiment function.
+        """
+        Plotter.enable = enable
+
+    @staticmethod
+    def status(plot_algo_status):
```
ryanjulian: This function having an argument is a leaky abstraction. Either the Plotter is in charge of its enabled status, or the caller is. Not both.

garage/algos/cem.py
gonzaiva: parallel_sampler is used in line 113 and 189, while stateful_pool is used in line 134. Theses imports were added since both parallel_sampler and stateful_pool were detected as uninitialized imports by pylint.

ryanjulian: lgtm

garage/plotter/plotter.py
ryanjulian: err no, now i see

`if Plotter.enable is None` is equivalent to `if not Plotter.enable`

See my comment below about why this is a confusing way to do it.

gonzaiva: I changed it as you suggested Ryan, however, there's one tricky case when plot=True in run_experiment and plot=False in the constructor of the algorithms. Since the algorithms check their own plot flag, they're not aware of the static plot flag in the Plotter class unless the flag is overridden directly in the classes of the algorithms.
I did this in the constructors by performing this operation:
`self.plot = self.plot or Plotter.enable`
The guard in the Plotter class will take care of the cases when plot=False in the function run_experiment.

ryanjulian: Hmm this is still evidence then that the abstraction is leaking. The algorithm shouldn't have to check that (and new algorithm authors are unlikely to know to check it).

Perhaps the solution is to remove the `plot` parameter from either algorithm constructors or `run_experiment` altogether? Or maybe it's something more clever. Another solution is for all the Plotter functions to noop when `run_experiment` has `plot=False`.

The reason that there is a `plot` parameter in `run_experiment` is so that people can easily turn off plotting globally when they are running large jobs on headless servers. Not everyone uses run_experiment all the time, though, so moving `plot=` to *only* `run_experiment` also creates friction.

How would you design it? Put yourself in the user's shoes.

garage/tf/plotter/plotter.py
```diff
@@ -93,21 +98,31 @@ def shutdown(self):
             self.queue.join()
             self.worker_thread.join()
 
+    @staticmethod
+    def disable():
+        """Set the plot enabling according to the run_experiment function.
```
ryanjulian: don't mention the caller in the docstring. we don't know who the caller is.

garage/tf/algos/batch_polopt.py
```diff
@@ -73,7 +72,7 @@ def __init__(self,
         self.max_path_length = max_path_length
         self.discount = discount
         self.gae_lambda = gae_lambda
-        self.plot = plot
+        self.plot = plot or Plotter.enable
```
ryanjulian: we can't ask algorithm implementers to do this check

garage/algos/ddpg.py
```diff
@@ -180,7 +179,7 @@ def __init__(self,
         self.n_updates_per_sample = n_updates_per_sample
         self.include_horizon_terminal_transitions = \
             include_horizon_terminal_transitions
-        self.plot = plot
+        self.plot = plot or Plotter.enable
```
ryanjulian: we can't ask algorithm implementers to do this check

garage/algos/cma_es.py
```diff
@@ -51,12 +49,13 @@ def __init__(self,
         Serializable.quick_init(self, locals())
         self.env = env
         self.policy = policy
-        self.plot = plot
+        self.plot = plot or Plotter.enable
```
ryanjulian: please remove.

garage/algos/cem.py
```diff
@@ -95,7 +97,7 @@ def __init__(self,
         self.env = env
         self.policy = policy
         self.batch_size = batch_size
-        self.plot = plot
+        self.plot = plot or Plotter.enable
```
ryanjulian: please remove.

garage/algos/batch_polopt.py
```diff
@@ -97,7 +96,7 @@ def __init__(self,
         self.max_path_length = max_path_length
         self.discount = discount
         self.gae_lambda = gae_lambda
-        self.plot = plot
+        self.plot = plot or Plotter.enable
```
ryanjulian: please remove.

garage/plotter/plotter.py
```diff
@@ -1,88 +1,130 @@
...
-        while True:
-            msgs = {}
-            # Only fetch the last message of each type
+
+Message = namedtuple("Message", ["op", "args", "kwargs"])
+
+
+class Plotter():
+
+    # Static variable used along with function run_experiment to enable or
```
ryanjulian: Don't mention the caller in a docstring. We don't know who the caller is.

garage/plotter/plotter.py
```diff
@@ -1,88 +1,130 @@
...
-    max_length = None
-    try:
-        while True:
-            msgs = {}
-            # Only fetch the last message of each type
+
+Message = namedtuple("Message", ["op", "args", "kwargs"])
+
+
+class Plotter():
```
ryanjulian: () is not necessary AFAIK

garage/tf/plotter/plotter.py
```diff
@@ -93,21 +98,31 @@ def shutdown(self):
...
     def start(self):
-        if not self.worker_thread.is_alive():
-            tf.get_variable_scope().reuse_variables()
-            self.worker_thread.start()
-            self.queue.put(
-                Message(
-                    op=Op.UPDATE, args=(self.env, self.policy), kwargs=None))
-            self.queue.task_done()
-            atexit.register(self.shutdown)
+        if Plotter.enable:
```
ryanjulian: Use the guard pattern:

```python
if not Plotter.enable:
    return
```

ryanjulian: and so forth for all the others

Issue #112, opened 2018-06-28
Support alphabetizing logger.dump_tabular() output


ryanjulian:
Sometimes related output logs are recorded at different times. We would like them to be grouped together during output for easy inspection.

e.g. right now we have
```python
logger.record_tabular("Policy/Entropy", 1)
logger.record_tabular("Optimizer/LearningRate", 2)
logger.record_tabular("Policy/MeanKL", 3)
logger.dump_tabular()
""" output:
Policy/Entropy            1.0
Optimizer/LearningRate    2.0
Policy/MeanKL             3.0
"""
```

but we would like to have
```python
logger.record_tabular("Policy/Entropy", 1)
logger.record_tabular("Optimizer/LearningRate", 2)
logger.record_tabular("Policy/MeanKL", 3)
logger.dump_tabular()
""" output:
Optimizer/LearningRate    2.0
Policy/Entropy            1.0
Policy/MeanKL             3.0
"""
```

CatherineSue:
Fixed in #124.

PR #124, opened 2018-06-30
Add support for alphabetizing logger tabular

CatherineSue:
Group each tabular data together by the alphabetical order of their
keys.

See: #112

### DIFF
--- a/garage/misc/tabulate.py
+++ b/garage/misc/tabulate.py
@@ -3,7 +3,6 @@
 from collections import namedtuple
 from functools import partial, reduce
 from itertools import zip_longest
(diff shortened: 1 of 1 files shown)


CatherineSue:
There is an issue I haven't dealt.
For example:
```
print(tabulate([[1, 2.34], [-56, "8.999"], ["10.2", 832], [9.3, 123]]))
# right now the return is 
-----  -------
-56      8.999
  1      2.34
 10.2  832
  9.3  123
-----  -------
```
I don't know if we will have column names like this in the future. Currently I sorted keys based on string type.

PR #130, opened 2018-07-02
Sawyer reacher mujoco

zhanpenghe:


garage/envs/mujoco/sawyer/__init__.py
```diff
@@ -1,3 +1,4 @@
 from garage.envs.mujoco.sawyer.bin_sorting_env import BinSortingEnv
 from garage.envs.mujoco.sawyer.block_stacking_env import BlockStackingEnv
 from garage.envs.mujoco.sawyer.pick_and_place_env import PickAndPlaceEnv
+from garage.envs.mujoco.sawyer.reacher_env import ReachEnv
```
ryanjulian: `ReacherEnv`

garage/envs/mujoco/mujoco_env.py
```diff
@@ -230,3 +230,9 @@ def print_stats(self):
 
     def action_from_key(self, key):
         raise NotImplementedError
+
+    def _env_setup(self, initial_qpos):
```
ryanjulian: a private API which is commonly called by subclasses is a bad smell. I think an initial_qpos field in the constructor of MujocoEnv could satisfy the same purpose, right?

garage/envs/mujoco/sawyer/reacher_env.py
```diff
@@ -0,0 +1,144 @@
...
+        done = (self._goal_distance(achieved_goal, goal) <
+                self._distance_threshold)
+
+        return Step(next_obs, reward, done)
+
+    def _compute_reward(self, achieved_goal, goal):
+        # Compute distance between goal and the achieved goal.
+        d = self._goal_distance(achieved_goal, goal)
+        if self._sparse_reward:
+            return -(d > self._distance_threshold).astype(np.float32)
```
ryanjulian: Can you add a completion bonus? We find it's usually necessary to guarantee stability.

I usually set the completion bonus large enough to make the AverageReturn guaranteed positive for an agent which has masted the environment (e.g. it wipes out the negative path rewards accrued while moving to the goal).

garage/envs/mujoco/sawyer/reacher_env.py
```diff
@@ -0,0 +1,144 @@
...
+        :return: observation space
+        """
+        return Box(
+            -np.inf, np.inf, shape=self.get_current_obs()['observation'].shape)
+
+    @overrides
+    def close(self):
+        """Close the viewer."""
+        if self.viewer is not None:
+            self.viewer = None
```
ryanjulian: This didn't actually close the viewer. I don't think Python will automatically call the terminate() API on destruction (unless someone overloaded __del__)

garage/envs/mujoco/mujoco_env.py
zhanpenghe: Yeah, the init_qpos is acquired from the simulation initial qpos in __init__() but I actually want to have initial_qpos set by users. This function is used mostly for robotics. I will modify it as part of the mujoco_env initialization and have a parameter with default None. Does it sound good?

garage/envs/mujoco/sawyer/reacher_env.py
zhanpenghe: I don't think mujoco viewer has a terminate so I just set it to None in order to make it restart the viewer next time.

garage/envs/mujoco/sawyer/reacher_env.py
eric-heiden: I think for sparse rewards, this implementation is good.
For dense rewards, it would actually be nice to compute rewards in [0, 1] because most algorithms need this normalized reward. DM Control Suite does this using [sigmoid functions](https://github.com/deepmind/dm_control/blob/master/dm_control/utils/rewards.py). The env normalizer can to do that by batch-normalizing the rewards which results in a constant rescaling which impacts the training.
I agree with Ryan on the completion bonus - it should be something like the absolute sum of rewards that a typical successful policy can accrue until it hits the target.

vendor/mujoco_models/reach.xml
```diff
@@ -0,0 +1,31 @@
...
+  <actuator>
+    <motor joint="right_j0" ctrlrange="-100.0 100.0" ctrllimited="false"/>
+    <motor joint="right_j1" ctrlrange="-100.0 100.0" ctrllimited="false"/>
+    <motor joint="right_j2" ctrlrange="-100.0 100.0" ctrllimited="false"/>
+    <motor joint="right_j3" ctrlrange="-100.0 100.0" ctrllimited="false"/>
+    <motor joint="right_j4" ctrlrange="-100.0 100.0" ctrllimited="false"/>
+    <motor joint="right_j5" ctrlrange="-100.0 100.0" ctrllimited="false"/>
+    <motor joint="right_j6" ctrlrange="-100.0 100.0" ctrllimited="false"/>
+    <motor gear="1" joint="r_gripper_l_finger_joint" name="slide_l_claw"/>
+    <motor gear="1" joint="r_gripper_r_finger_joint" name="slide_r_claw"/>
```
eric-heiden: You can tune this `gear` parameter for the other motors as well to try to match the Gazebo env.

garage/envs/__init__.py
```diff
@@ -1,7 +1,7 @@
 from garage.envs.env_spec import EnvSpec
 from garage.envs.base import Step  # noqa: I100
-from garage.envs.dm_control_env import DmControlEnv
-from garage.envs.dm_control_viewer import DmControlViewer
+# from garage.envs.dm_control_env import DmControlEnv
```
ryanjulian: please don't commit commented-out code

garage/envs/mujoco/sawyer/reacher_env.py
```diff
@@ -0,0 +1,185 @@
...
+    def _compute_reward(self, achieved_goal, goal):
+        # Compute distance between goal and the achieved goal.
+        d = self._goal_distance(achieved_goal, goal)
+        if self._sparse_reward:
+            reward = -(d > self._distance_threshold).astype(np.float32)
+        else:
+            reward = -d
+
+        if d < self._distance_threshold:
+            # reward += (-self._accumulated_reward)  # Zero the reward
```
ryanjulian: please don't commit commented-out code

vendor/mujoco_models/reach.xml
```diff
@@ -0,0 +1,36 @@
+<mujoco>
```
ryanjulian: i think this should be called reacher.xml (since the env is ReacherEnv).

tests/envs/test_sawyer_envs.py
```diff
@@ -69,17 +72,28 @@ def run_pick_and_place(*_):
...
 
 test_env()
-run_experiment_lite(
-    run_bin_sorting,
-    n_parallel=2,
-    plot=True,
-)
+# run_experiment(
+#     run_bin_sorting,
+#     n_parallel=2,
```
ryanjulian: Please don't commit commented-out code

ryanjulian:
Is this ready to commit? we shouldn't be running experiments on random branches.

eric-heiden:
One more thing: please add an object to the scene where the goal is (free floating, no collision). The rendering is not very helpful otherwise.

Issue #93, opened 2018-06-15
Some envs are broken in render()


ryanjulian:
```sh
(garage) rjulian@nyquist:~/code/garage$ nose2 tests.envs
2018-06-15 11:26:49.815393: I tensorflow/core/platform/cpu_feature_guard.cc:140] Your CPU supports instructions that this TensorFlow binary was not compiled to use: AVX2 AVX512F FMA
WARN: gym.spaces.Box autodetected dtype as <class 'numpy.float32'>. Please provide explicit dtype.
.Testing <class 'garage.envs.grid_world_env.GridWorldEnv'>
ETesting <class 'garage.envs.box2d.cartpole_env.CartpoleEnv'>
.Testing <class 'garage.envs.box2d.car_parking_env.CarParkingEnv'>
.Testing <class 'garage.envs.box2d.cartpole_swingup_env.CartpoleSwingupEnv'>
.Testing <class 'garage.envs.box2d.double_pendulum_env.DoublePendulumEnv'>
.Testing <class 'garage.envs.box2d.mountain_car_env.MountainCarEnv'>
.Testing <class 'garage.envs.mujoco.point_env.PointEnv'>
Creating window glfw
.Testing <class 'garage.envs.mujoco.walker2d_env.Walker2DEnv'>
Creating window glfw
.Testing <class 'garage.envs.mujoco.swimmer_env.SwimmerEnv'>
Creating window glfw
.Testing <class 'garage.envs.mujoco.simple_humanoid_env.SimpleHumanoidEnv'>
Creating window glfw
.Testing <class 'garage.envs.mujoco.inverted_double_pendulum_env.InvertedDoublePendulumEnv'>
Creating window glfw
.Testing <class 'garage.envs.mujoco.hopper_env.HopperEnv'>
Creating window glfw
.Testing <class 'garage.envs.mujoco.half_cheetah_env.HalfCheetahEnv'>
Creating window glfw
.Testing <class 'garage.envs.mujoco.gather.point_gather_env.PointGatherEnv'>
ETesting <class 'garage.envs.mujoco.gather.swimmer_gather_env.SwimmerGatherEnv'>
ETesting <class 'garage.envs.mujoco.gather.ant_gather_env.AntGatherEnv'>
ETesting <class 'garage.envs.mujoco.maze.point_maze_env.PointMazeEnv'>
Creating window glfw
.Testing <class 'garage.envs.mujoco.maze.swimmer_maze_env.SwimmerMazeEnv'>
[...]

PR #136, opened 2018-07-02
Fix box2d render issue

jonashen:
```
garage/lib/python3.5/site-packages/gym/core.py", line 284, in render
    return self.env.render(mode)
```
In gym v0.10.5, gym introduced an additional argument called `mode` that defaults to the string `'human'`. However, since it was not called as a kwarg (i.e. mode=mode), garage's box2d_env render incorrectly 
assumed this was input for the `states` argument and raised a NotImplementedError. 

Ref: #93

### DIFF
--- a/garage/envs/box2d/box2d_env.py
+++ b/garage/envs/box2d/box2d_env.py
@@ -351,7 +351,7 @@ class Box2DEnv(gym.Env):
         return self._compute_com_pos_vel(*com)[2:]
 
     @overrides
-    def render(self, mode='human', states=None, actions=None, pause=False):
+    def render(self, mode, states=None, actions=None, pause=False):
         if not self.viewer:
             self.viewer = Box2DViewer(self.world)
         if states or actions or pause:
--- a/garage/envs/box2d/box2d_env.py
+++ b/garage/envs/box2d/box2d_env.py
@@ -258,11 +258,11 @@ class Box2DEnv(gym.Env):
                 if state.local is not None:
                     loc = state.local
                     position = body.GetWorldPoint(loc)
-                    linearVel = body.GetLinearVelocityFromLocalPoint(loc)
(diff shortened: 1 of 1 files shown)


garage/envs/box2d/box2d_env.py
```diff
@@ -351,7 +351,7 @@ def get_com_velocity(self, *com):
         return self._compute_com_pos_vel(*com)[2:]
 
     @overrides
-    def render(self, states=None, actions=None, pause=False):
+    def render(self, mode='human', states=None, actions=None, pause=False):
```
ryanjulian: This violates the [gym.Env interface](https://github.com/openai/gym/blob/master/gym/core.py#L11). Only the [gym.Wrapper interface](https://github.com/openai/gym/blob/master/gym/core.py#L238) (which you can subclass) supports kwargs other than mode.

garage/envs/box2d/box2d_env.py
```diff
@@ -351,7 +351,7 @@ def get_com_velocity(self, *com):
         return self._compute_com_pos_vel(*com)[2:]
 
     @overrides
-    def render(self, states=None, actions=None, pause=False):
+    def render(self, mode, states=None, actions=None, pause=False):
```
ryanjulian: What if someone called env.render() (entirely valid for a gym.Env), where type(env) == Box2DEnv?

garage/tf/q_functions/continuous_mlp_q_function.py
```diff
@@ -1,16 +1,12 @@
 import tensorflow as tf
 
 from garage.core import Serializable
-from garage.misc import ext
```
ryanjulian: are these unused?

jonashen: yes, it was throwing pylint errors.

Issue #82, opened 2018-06-15
Replace `garage.spaces` with akro


eric-heiden:
Action and observation spaces of `garage`'s policies still rely on `garage.spaces`. We should get rid of that and use the counterparts from `gym.spaces` instead.

ryanjulian:
I think this is preliminary.

I'd like to have as little code as possible too, but I don't think it will be clear what to do with Spaces until we rip out the Layers library, so I think we should wait for a design for that before attempting this.

In gym, spaces only represent input/output spaces for environments. In garage, spaces represent input/output spaces for environments *and* how to convert those spaces into neural network inputs/outputs *and* neural network subgraphs communicate *and* how training algorithms and samplers interface with neural network subgraphs.

For instance, the most important functionality in spaces right now is Space.new_tensor_variable(), which does not exist in gym.spaces. Yes, we can add a conversion function, and maybe that is what we'll eventually do, but wouldn't it be nice to instead have an interface that every NN library implements called `new_tensor_variable()` instead? That's what garage.spaces is. So perhaps this is an argument that garage.Space should inherit from gym.Space and add a few extra functions? Sounds like it could work, but I'm not sure, because it hasn't been designed. Etc.

ryanjulian:
The above PR was a great improvement, but I don't think it really resolved Eric's concerns, so I'm reopening this.

ryanjulian:
Fixed by #537

PR #140, opened 2018-07-03
Add top-level base Space class for garage

jonashen:
This PR creates a top-level base Space class for garage that both TensorFlow and Theano spaces inherit from and build off of. 

Ref: #82


garage/tf/policies/deterministic_mlp_policy.py
```diff
@@ -10,7 +10,7 @@
 from garage.tf.distributions import Categorical
 from garage.tf.misc import tensor_utils
 from garage.tf.policies import Policy
-from garage.tf.spaces import Discrete
+from garage.spaces.tf import Discrete
```
CatherineSue: This import is unused.

PR #143, opened 2018-07-05
Cleanly terminate all worker processes on an interrupt

gonzaiva:
When calling function run_experiment and the parameter n_parallel is
bigger than zero, worker processes are created to do parallel sampling
during the training. However, when there's a keyboard interrupt, there
are certain occasions when processes are not completely killed after the
execution has finished.
By checking the PID of these processes using the command "ps -fu | grep
run_experiment", we noticed they had the status code "Sl", where S means
interruptible sleep, so the processes are waiting for a signal to wake
up and continue their execution.
When executing the command "kill -SIGINT <pid>" with the PID of the
sleeping process, we got the python trace back and found that the
processes were sleeping trying to acquire the lock of the inqueue inside
the pool of python multiprocessing.
By default, when a keyboard interruption occurs all processes get
notified, but the notifications are not sent in any order. This could
produce an issue when a process is interrupted and finished without
releasing the lock that other process are waiting for as it's in this
case.
To avoid this problem, the worker processes are forked with SIGINT
blocked, so only their parent will be notified, taking care of
terminating the pool of workers once the keyboard interrupt is caught.

### DIFF
--- a/garage/sampler/parallel_sampler.py
+++ b/garage/sampler/parallel_sampler.py
@@ -69,6 +69,8 @@ def terminate_task(scope=None):
     singleton_pool.run_each(_worker_terminate_task,
                             [(scope, )] * singleton_pool.n_parallel)
 
+def terminate():
+    singleton_pool.terminate()
 
 def _worker_set_seed(_, seed):
     logger.log("Setting seed to %d" % seed)
--- a/garage/sampler/parallel_sampler.py
+++ b/garage/sampler/parallel_sampler.py
@@ -146,7 +146,7 @@ def truncate_paths(paths, max_samples):
     total_n_samples = sum(len(path["rewards"]) for path in paths)
     while paths and total_n_samples - len(paths[-1]["rewards"]) >= max_samples:
         total_n_samples -= len(paths.pop(-1)["rewards"])
-    if paths > 0:
+    if paths:
         last_path = paths.pop(-1)
         truncated_last_path = dict()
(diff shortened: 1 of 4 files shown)


garage/sampler/parallel_sampler.py
```diff
@@ -129,21 +133,20 @@ def sample_paths(policy_params,
...
     # chop samples collected by extra paths
     # make a copy
     paths = list(paths)
     total_n_samples = sum(len(path["rewards"]) for path in paths)
-    while len(paths) > 0 and total_n_samples - len(
-            paths[-1]["rewards"]) >= max_samples:
+    while paths and total_n_samples - len(paths[-1]["rewards"]) >= max_samples:
         total_n_samples -= len(paths.pop(-1)["rewards"])
-    if len(paths) > 0:
+    if paths > 0:
```
jonashen: `if paths:` would be better since paths is a list

scripts/run_experiment.py
```diff
@@ -119,11 +119,15 @@ def run_experiment(argv):
     if args.seed is not None:
         set_seed(args.seed)
 
+    # Fork the work processes with SIGINT blocked so they don't get
+    # finished when they own locks
```
jonashen: Is it supposed to be when they release their own locks?

jonashen: I see, I only asked because I was confused by your comment

PR #147, opened 2018-07-06
Missing changes to cleanly exit worker processes

gonzaiva:
These are changes that are required for the following commit to work:
dcdc6b4e29edf22a1df9a0b4a708e7d9b6d447de
The worker processes pool is only created when n_parallel is bigger than
1, so n_parallel=1 was creating null exceptions when terminate was
called. This is solved by adding a guard to check if the pool was
instantiated.
Another fix that is required is to rename back the member g in
singleton_pool to G. This was done in a previous change to enforce PEP8
style of having non-capital variables/parameters, but the member has to
keep the same name as defined in singleton_pool.

### DIFF
--- a/garage/sampler/stateful_pool.py
+++ b/garage/sampler/stateful_pool.py
@@ -72,8 +72,8 @@ class StatefulPool(object):
         Run the method on each worker process, and collect the result of
         execution.
 
-        The runner method will receive 'G' as its first argument, followed by
-        the arguments in the args_list, if any
+        The runner method will receive 'G' as its first argument, followed
+        by the arguments in the args_list, if any
         :return:
         """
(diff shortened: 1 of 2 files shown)


garage/sampler/stateful_pool.py
```diff
@@ -64,15 +64,16 @@ def initialize(self, n_parallel):
...
+            self.pool.terminate()
 
     def run_each(self, runner, args_list=None):
         """
         Run the method on each worker process, and collect the result of
         execution.
 
-        The runner method will receive 'G' as its first argument, followed by
-        the arguments in the args_list, if any
+        The runner method will receive 'G' as its first argument, followed
```
ryanjulian: 'g'

Issue #144, opened 2018-07-05
Don't double check pre-commits in the CI


ryanjulian:
Some pre-commit checks are also standalone CI checks. We shouldn't check them twice.

We can either
* Remove the individual scripts and use pre-commit for duplicative checks (preferred, if it works for all cases)
* Turn off the duplicative checks when we use pre-commit

ryanjulian:
Also, it looks to me like the check_precommit script should be using the `--source` and `--origin` args, as documented here https://pre-commit.com/#pre-commit-run

ryanjulian:
Why have more code when you can have less?

jonashen:
It will take me more time to learn how to write the less code :(

PR #149, opened 2018-07-06
Run some CI checks using pre-commit

jonashen:
Remove duplicate flake8 check in pre-commit and rewrote check_precommit.sh using built-in pre-commit API.

Ref: #144


scripts/travisci/check_precommit.sh
```diff
@@ -1,15 +1,8 @@
...
-  fi
+  pre-commit run --source ${TRAVIS_COMMIT} --origin HEAD
 else
   git remote set-branches --add origin master
   git fetch
-  files_changed=$(git diff origin/master --name-only | grep ".*\.py")
-  if [[ ! -z "${files_changed}" ]]; then
-    pre-commit run --files ${files_changed}
-  fi
+  pre-commit run --source origin/master --origin HEAD
```
ryanjulian: i think you might have this reversed. source == HEAD and origin == origin/master

ryanjulian: reading the Travis documentation, it would be more idiomatic to use TRAVIS_BRANCH instead of HEAD.

ryanjulian: (if so please update this case in the other CI scripts)

scripts/travisci/check_precommit.sh
```diff
@@ -1,15 +1,8 @@
 #!/usr/bin/env bash
 if [[ "${TRAVIS_PULL_REQUEST}" != "false" && "${TRAVIS}" == "true" ]]; then
-  files_changed=$(git diff "${TRAVIS_COMMIT_RANGE}" --name-only \
-    | grep ".*\.py")
-  if [[ ! -z "${files_changed}" ]]; then
-    pre-commit run --files  ${files_changed}
-  fi
+  pre-commit run --source ${TRAVIS_COMMIT_RANGE%...*} --origin HEAD
```
ryanjulian: source is the newest commit considered
origin is usually assumed to be older than source (HEAD should be newest)

the format of TRAVIS_COMMIT_RANGE is `SOURCE...ORIGIN`

splitting the TRAVIS_COMMIT_RANGE is IMO the right strategy. this only splits out the SOURCE part, but you should split both (i.e. don't assume ORIGIN==HEAD, though that's usually true).

ryanjulian: See https://git-scm.com/docs/gitrevisions#_specifying_ranges for full spec

ryanjulian: (reading the git spec, the pre-commit documentation is a little misguided on how commit ranges work, but we should stick to their naming scheme).

scripts/travisci/check_precommit.sh
jonashen: I reversed it initially, but it would say no files changed. Not sure why.

scripts/travisci/check_precommit.sh
jonashen: sgtm

scripts/travisci/check_precommit.sh
jonashen: i.e.
```
autopep8 wrapper.....................................(no files to check)Skipped
Check for added large files..........................(no files to check)Skipped
Check python ast.....................................(no files to check)Skipped
Check for merge conflicts............................(no files to check)Skipped
Check Xml............................................(no files to check)Skipped
Check Yaml...........................................(no files to check)Skipped
Debug Statements (Python)............................(no files to check)Skipped
Fix End of Files.....................................(no files to check)Skipped
Forbid new submodules................................(no files to check)Skipped
Trim Trailing Whitespace.............................(no files to check)Skipped
```

ryanjulian: okay i think the pre-commit authors are just wrong in their naming scheme. carry on.

.pre-commit-config.yaml
```diff
@@ -32,7 +32,7 @@ repos:
     -   id: end-of-file-fixer           # Makes sure files end in a newline and only a newline.
```
ryanjulian: don't run autopep8 -- sometimes its results conflict with YAPF

https://github.com/google/yapf/blob/master/plugins/pre-commit.sh

.pre-commit-config.yaml
```diff
@@ -32,7 +32,7 @@ repos:
...
     # -   id: fix-encoding-pragma         # Add # -*- coding: utf-8 -*- to the top of python files.
     # -   id: file-contents-sorter        # Sort the lines in specified files (defaults to alphabetical). You must provide list of target files as input to it.
-    -   id: flake8                      # Run flake8 on your python files.
```
ryanjulian: We should run pylint and YAPF here and get rid of the individual scripts

https://github.com/pre-commit/mirrors-pylint
https://github.com/pre-commit/mirrors-yapf

jonashen: I'm not sure about the accuracy of these two, but at least for flake8, I remember that pre-commit's flake8 would pass my code while our flake8 script would fail me. Maybe pre-commit is not always up to date?

ryanjulian: Our flake8 script has special rules about files which are existing/added/changed which is why we won't use pre-commit for it (we could write a custom pre-commit script for it, but probably not worth it right now).

YAPF and pylint are straightforward invocations of the tool using our setup.cfg. It should be fine.

Angel did the original configuration and can tell you all about it.

CONTRIBUTING.md
```diff
@@ -3,6 +3,10 @@ We welcome all contributions to garage.
 
 Use this guide to prepare your contribution.
 
+# Post-setup instructions
+After activating your conda environment (e.g. `conda activate garage`), make sure you run `pre-commit install` to install the pre-commit package on your machine.
```
ryanjulian: install the pre-commit hooks to your repository.

CONTRIBUTING.md
```diff
@@ -3,6 +3,10 @@ We welcome all contributions to garage.
...
+This will streamline your git process and reduce potential CI errors.
```
ryanjulian: This will help streamline the pull request process by catching basic problems locally before they are checked by the CI.

scripts/travisci/check_precommit.sh
```diff
@@ -1,15 +1,8 @@
 #!/usr/bin/env bash
 if [[ "${TRAVIS_PULL_REQUEST}" != "false" && "${TRAVIS}" == "true" ]]; then
-  files_changed=$(git diff "${TRAVIS_COMMIT_RANGE}" --name-only \
-    | grep ".*\.py")
-  if [[ ! -z "${files_changed}" ]]; then
-    pre-commit run --files  ${files_changed}
-  fi
+  pre-commit run --source ${TRAVIS_COMMIT_RANGE%...*} --origin ${TRAVIS_COMMIT_RANGE#*...}
```
ryanjulian: Did you send a few intentionally broken commits to make sure it's still actually catching errors?

You can ignore precommit hooks using `git commit --no-verify`

jonashen: that's helpful, thanks!

Issue #146, opened 2018-07-06
Move garage.spaces.{theano, tf} to garage.{theano, tf}.spaces


ryanjulian:
This is better prep for the Theano refactor.

(sorry I didn't catch it in code review)

PR #150, opened 2018-07-06
Move spaces into respective directories

jonashen:
Move garage.spaces.{theano, tf} to garage.{theano, tf}.spaces.

Ref: #146


garage/tf/policies/gaussian_mlp_policy.py
```diff
@@ -4,14 +4,14 @@
...
 from garage.misc.overrides import overrides
-from garage.spaces.tf import Box as TfBox
 from garage.tf.core import LayersPowered
 from garage.tf.core import MLP
 import garage.tf.core.layers as L
 from garage.tf.distributions import DiagonalGaussian
 from garage.tf.misc import tensor_utils
 from garage.tf.misc.tensor_utils import enclosing_scope
 from garage.tf.policies import StochasticPolicy
+from garage.tf.spaces import Box as TfBox
```
ryanjulian: Not necessary to call this TfBox -- we're in the TF subtree and we're not using any other kind of Box.

Issue #111, opened 2018-06-28
Define Optimizer API or remove optimizers


ryanjulian:
e.g. dicts, lists, tuples.

CatherineSue:
such as all the inputs below should be accepted?
```
opt = ConjugateGradientOptimizer()
# all the inputs below should be accepted
tr_optimizer.update_opt(inputs=[xs_var, ys_var, old_p_var])
tr_optimizer.update_opt(inputs=(xs_var, ys_var, old_p_var))
tr_optimizer.update_opt(inputs={xs_var, ys_var, old_p_var})
```

ryanjulian:
perhaps soon they will only accept models?

ryanjulian:
TF optimizers should accept any input type that TF does

ryanjulian:
If we can achieve it, it would be best if we could use the optimizer API from each framework and eliminate our own

numpy -> scipy.optimize
TensorFlow -> tf.compat.v1.train.Optimizer?
PyTorch -> torch.optim.Optimizer

PR #152, opened 2018-07-06
Add input types support in TF.optimizers

CatherineSue:
TF.optimizers should accept any input types, such as dict, tuple, list.

Refer to: #111

### DIFF
--- a/garage/tf/misc/tensor_utils.py
+++ b/garage/tf/misc/tensor_utils.py
@@ -123,7 +123,15 @@ def pad_tensor_dict(tensor_dict, max_len):
     return ret
 
 
-class enclosing_scope(object):
+def convert_inputs(inputs, instance=()):
+    """Convert inputs to a sequence has the same type as instance."""
+    if isinstance(inputs, dict):
+        return type(instance)(inputs[k] for k in sorted(inputs.keys()))
+    else:
+        return type(instance)(inputs)
+
+
+class enclosing_scope(object):  # noqa: disable=N801
     def __init__(self, enclosing_name, name, **kwargs):
         self.enclosing_scope = None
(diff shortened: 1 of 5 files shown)


garage/tf/misc/tensor_utils.py
```diff
@@ -123,7 +123,14 @@ def pad_tensor_dict(tensor_dict, max_len):
     return ret
 
 
-class enclosing_scope(object):
+def normalized_inputs(inputs, instance=()):
+    if isinstance(inputs, dict):
+        return type(instance)(inputs.values())
```
ryanjulian: values() is not guaranteed to have a stable order. you may need to sort by keys first.

garage/tf/misc/tensor_utils.py
```diff
@@ -123,7 +123,14 @@ def pad_tensor_dict(tensor_dict, max_len):
     return ret
 
 
-class enclosing_scope(object):
+def normalized_inputs(inputs, instance=()):
```
ryanjulian: can you explain the purpose of this function?

CatherineSue: Transform the `inputs` to the same type as `instance`. I found many functions in tf.optimizers have `extra_inputs`,  so calling `inputs = normalized_inputs(inputs, extra_inputs)` transforms two into the same type.

CatherineSue: Transform the `inputs` to the same type as `instance`. I found many functions in tf.optimizers have `extra_inputs` but not in the same type, some have type list, some tuple. Calling `inputs = normalized_inputs(inputs, extra_inputs)` transforms two into the same type.

PR #158, opened 2018-07-09
Add Sawyer ReacherEnv in MuJoCo

ryanjulian:


scripts/travisci/check_imports.py
```diff
@@ -54,6 +54,8 @@ def get_import_statement(node, lines):
     "mujoco_py.builder",
     "mujoco_py.generated",
     "mujoco_py.generated.const",
+    "OpenGL",
+    "OpenGL.GL",
```
eric-heiden: This is addressed by #160

ryanjulian: I was testing this change by rebasing it on #160

Issue #156, opened 2018-07-09
render(mode="rgb_array") broken in MujocoEnv


eric-heiden:
Rendering a MujocoEnv to an RGB array doesn't work:
```
File "test.py", line 123, in main
    env.render(mode='rgb_array')
  File ".../garage/envs/mujoco/mujoco_env.py", line 197, in render
    data, width, height = viewer.get_image()
AttributeError: 'MjViewer' object has no attribute 'get_image'
```

jonashen:
https://github.com/openai/gym/commit/c5b624c6a6334137e7ae7bcc73b134725956c4ea

This should work

eric-heiden:
Thanks, this works! I will make a PR soon to fix it for garage.

ryanjulian:
Please include a test replicating the bug!

ryanjulian:
I think you could replicate this by testing MujocoEnvs with render_mode="rgb_array" in tests/envs/test_envs.py (in addition to the default, which is render_mode="human")

PR #161, opened 2018-07-09
Fix render(mode="rgb_array")

eric-heiden:
#156

### DIFF
--- a/garage/envs/mujoco/gather/gather_env.py
+++ b/garage/envs/mujoco/gather/gather_env.py
@@ -206,6 +206,8 @@ class GatherEnv(ProxyEnv, Serializable):
         self.dying_cost = dying_cost
         self.objects = []
         self.viewer = None
+        self.render_width = 512
+        self.render_height = 512
         # super(GatherEnv, self).__init__(*args, **kwargs)
         model_cls = self.__class__.MODEL_CLASS
         if model_cls is None:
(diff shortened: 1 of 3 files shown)


garage/envs/mujoco/gather/gather_env.py
```diff
@@ -414,10 +418,14 @@ def stop_viewer(self):
...
-            self.get_viewer().render()
-            data, width, height = self.get_viewer().get_image()
-            return np.fromstring(
-                data, dtype='uint8').reshape(height, width, 3)[::-1, :, :]
+            viewer = self.get_viewer()
+            self.wrapped_env.render()
+            img = viewer.read_pixels(width=self.render_width,
+                                     height=self.render_height,
+                                     depth=False)
+            img = img[::-1, :, :]
```
jonashen: Maybe adding a comment specifying the original image was upside-down and this line flips it can be helpful when looking back on this code

tests/envs/test_envs.py
```diff
@@ -87,4 +87,9 @@ def test_env(env):
         print("Skipping rendering test")
     else:
         env.render()
+        if all(not isinstance(env, T) for T in [GridWorldEnv, Box2DEnv]):
```
ryanjulian: does this actually fail in the CI?

jonashen:
Can someone else confirm if this works? Running `nose2 tests.envs.test_envs` on my mac works, but not on my Linux. I get this error:

```
...
 File "/home/jonathon/Desktop/garage/tests/envs/test_envs.py", line 95, in test_env
   assert img is not None
AssertionError

Ran 25 tests in 0.662s

FAILED (failures=5, errors=14)
```

ryanjulian:
Remove guard which stops rendering on the CI in test_envs.py, which is a relic of the past. Rendering should work just fine in the CI.

ryanjulian:
Two bugs for the price of one? I like it!

Issue #133, opened 2018-07-02
spec() wrapper should return garage.spaces types, not gym.spaces types


ryanjulian:
https://github.com/rlworkgroup/garage/blob/master/garage/envs/util.py#L113

ryanjulian:
This has broken the TF tree (and presumably theano too)

ryanjulian:
Well tip-of-tree is broken for all of tf (and probably theano). This should never happen. We will make arbitrarily large changes to make it unbroken, including rewriting all of spec or spaces.

ryanjulian:
```sh
(garage) rjulian@nyquist:~/code/garage$ python garage/tf/launchers/trpo_cartpole.py
2018-07-06 14:45:51.821318: I tensorflow/core/platform/cpu_feature_guard.cc:140] Your CPU supports instructions that this TensorFlow binary was not compiled to use: AVX2 AVX512F FMA
2018-07-06 14:45:52.155930: I tensorflow/core/common_runtime/gpu/gpu_device.cc:1356] Found device 0 with properties: 
name: TITAN Xp major: 6 minor: 1 memoryClockRate(GHz): 1.582
pciBusID: 0000:17:00.0
totalMemory: 11.91GiB freeMemory: 11.52GiB
2018-07-06 14:45:52.468052: I tensorflow/core/common_runtime/gpu/gpu_device.cc:1356] Found device 1 with properties: 
name: TITAN Xp major: 6 minor: 1 memoryClockRate(GHz): 1.582
pciBusID: 0000:65:00.0
totalMemory: 11.91GiB freeMemory: 11.52GiB
2018-07-06 14:45:52.745682: I tensorflow/core/common_runtime/gpu/gpu_device.cc:1356] Found device 2 with properties: 
name: GeForce GTX 1060 6GB major: 6 minor: 1 memoryClockRate(GHz): 1.7085
pciBusID: 0000:b3:00.0
totalMemory: 5.92GiB freeMemory: 4.72GiB
2018-07-06 14:45:52.746130: I tensorflow/core/common_runtime/gpu/gpu_device.cc:1435] Adding visible gpu devices: 0, 1, 2
2018-07-06 14:45:53.330778: I tensorflow/core/common_runtime/gpu/gpu_device.cc:923] Device interconnect StreamExecutor with strength 1 edge matrix:
2018-07-06 14:45:53.330815: I tensorflow/core/common_runtime/gpu/gpu_device.cc:929]      0 1 2 
[...]

PR #164, opened 2018-07-09
Fix spec(env) implementation

jonashen:
spec(env) now returns garage.Spaces, and the algorithms using env_spec are updated accordingly to treat them as garage.Spaces.

Ref: #133

### DIFF
--- a/examples/vpg_1.py
+++ b/examples/vpg_1.py
@@ -5,7 +5,7 @@ import theano.tensor as TT
 
 from garage.envs import normalize
 from garage.envs.box2d import CartpoleEnv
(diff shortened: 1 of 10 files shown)


ryanjulian:
Does this not also affect the tf tree?

ryanjulian:
Why not do the same in the theano tree?

jonashen:
We can do that, but it will require a lot more refactoring. I was under the impression to get all the examples working first, and the refactor will be addressed in another PR.

ryanjulian:
Okay, please file a bug for the refactor.

PR #167, opened 2018-07-10
Centralize all examples in garage

jonashen:
All examples in garage are now centralized into the garage/examples/ folder.

Some renaming: 
examples -> examples/theano
alexbeloi -> issampler
rllab_hyperopt -> hyperopt
bichengcao -> gym
tf/launchers -> tf


ryanjulian:
Let's do this only for tf and theano for now. The fate of the contrib modules is uncertain (but generally the contract is that you might have to do a lot of extra setup to use something in contrib). Examples should work out-of-the-box.

jonashen:
done.

Issue #69, opened 2018-06-13
Run existing unit tests in the CI


ryanjulian:

ryanjulian:
We may need to disable some tests to keep the CI green. In this case, we file a bug for each broken test.

ryanjulian:
Fixed in #191

PR #178, opened 2018-07-11
Run existing unit tests in the CI

hjzh4:
- refer to #69

### DIFF
--- a/scripts/travisci/run_unit_tests.py
+++ b/scripts/travisci/run_unit_tests.py
@@ -1,0 +1,35 @@
+import os.path as osp
+import sys
+
+import nose2
+
+# failed
+FAILED_TESTS = [
+    'test_spaces', 'test_networks', 'test_baselines', 'test_algos',
+    'regression_tests.test_issue_3', 'algos.test_trpo', 'test_stateful_pool'
+]
+
+# slow but successfully
+SLOW_TESTS = [
+    'test_normalized_gym', 'test_dynamics_rand',
+    'regression_tests.test_benchmark_ddpg', 'envs.test_envs',
+    'envs.test_sawyer_envs', 'test_dm_control'
+]
+
+AVAILABLE_TESTS = [
+    'test_serializable', 'test_sampler', 'test_instrument',
+    'envs.test_maze_env'
+]
+
+tests_dir = osp.join(
+    osp.dirname(osp.dirname(osp.dirname(osp.abspath(__file__)))), 'tests')
+
+test_run = nose2.discover(
+    argv=['-v', '-s', tests_dir] + AVAILABLE_TESTS, exit=False)
+
+passed = False
+
+if test_run.result.wasSuccessful():
+    passed = True
+
+sys.exit(not passed)
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,6 +49,7 @@ script:
   - scripts/travisci/check_flake8.sh
   - scripts/travisci/check_imports.sh
   - scripts/travisci/check_precommit.sh
+  - scripts/travisci/run_unit_tests.sh
 
 notifications:
   email: false
--- a/scripts/travisci/run_unit_tests.sh
+++ b/scripts/travisci/run_unit_tests.sh
@@ -1,0 +1,7 @@
+#!/usr/bin/env bash
+export PYTHONPATH=.
+export TF_CPP_MIN_LOG_LEVEL=3      # shut TensorFlow up
+export DISABLE_MUJOCO_RENDERING=1  # silence glfw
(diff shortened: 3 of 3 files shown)


scripts/travisci/run_unit_tests.py
```diff
@@ -0,0 +1,35 @@
+import os.path as osp
+import sys
+
+import nose2
```
ryanjulian: we should be able to do this purely with nose2

ryanjulian:
Why close it?

scripts/travisci/run_unit_tests.py
hjzh4: I think the right way to do it is to run command nose2 tests directly so that every new pushes tests would be ran automatically. The only problem is how to exclude the tests we don’t want to run currently. The only way to select tests is by using test attributes while existing tests all don’t have attributes actually. I can also move unwanted tests out of the current directory but I think it would make project directory structure look dirty.

ryanjulian: The implementation of attributes doesn't seem hard to add (1 line in the test + 1 line in setup.cfg) and would be useful in the future (e.g. separating regression benchmarks from unit tests). I'd prefer this approach to maintaining another error-prone shell script (we already spend a lot of time debugging the existing ones).

Please correct me if it's harder than I imagine.

http://nose2.readthedocs.io/en/latest/plugins/attrib.html

ryanjulian: Okay I tried it--seems fine.

Steps:
1. Add the following to setup.cfg -- easy
    ```ini
    [unittest]
    plugins = nose2.plugins.attrib
              nose2.plugins.layers
    ```
1. Convert tests to actual Python unittest cases (necessary anyway) and then add an attribute -- easy
    test_networks.py - BEFORE
    ```python
    import lasagne.layers as L
    import numpy as np

    from garage.core import GRUNetwork
    from garage.misc import ext


    def test_gru_network():
        network = GRUNetwork(
            input_shape=(2, 3),
            output_dim=5,
            hidden_dim=4,
        )
        f_output = ext.compile_function(
            inputs=[network.input_layer.input_var],
            outputs=L.get_output(network.output_layer))
        assert f_output(np.zeros((6, 8, 2, 3))).shape == (6, 8, 5)
    ```
    test_networks.py -- AFTER
    ```python
    import unittest

    import lasagne.layers as L
    import numpy as np

    from garage.core import GRUNetwork
    from garage.misc import ext


    class TestNetworks(unittest.TestCase):
        def test_gru_network(self):
            network = GRUNetwork(
                input_shape=(2, 3),
                output_dim=5,
                hidden_dim=4,
            )
            f_output = ext.compile_function(
                inputs=[network.input_layer.input_var],
                outputs=L.get_output(network.output_layer))
            assert f_output(np.zeros((6, 8, 2, 3))).shape == (6, 8, 5)

        test_gru_network.broken = True
    ```
1. Run nose2 selecting an attribute (or its negative)
    ```sh
    (garage) rjulian@nyquist:~/code/garage/new_tests$ nose2 -c ../setup.cfg -A '!broken'
    .
    Ran 1 test in 0.024s

    OK
    (garage) rjulian@nyquist:~/code/garage/new_tests$ nose2 -c ../setup.cfg
    E.
    ======================================================================
    ERROR: test_gru_network (test_networks.TestNetworks)
      File "/home/rjulian/code/garage/new_tests/test_networks.py", line 15, in test_gru_network
        hidden_dim=4,
      File "/home/rjulian/code/garage/garage/core/network.py", line 263, in __init__
        b=l_output_flat.b,
      File "/home/rjulian/miniconda2/envs/garage/lib/python3.5/site-packages/lasagne/layers/dense.py", line 78, in __init__
        super(DenseLayer, self).__init__(incoming, **kwargs)
      File "/home/rjulian/miniconda2/envs/garage/lib/python3.5/site-packages/lasagne/layers/base.py", line 44, in __init__
        if any(d is not None and d <= 0 for d in self.input_shape):
      File "/home/rjulian/miniconda2/envs/garage/lib/python3.5/site-packages/lasagne/layers/base.py", line 44, in <genexpr>
        if any(d is not None and d <= 0 for d in self.input_shape):
    TypeError: unorderable types: tuple() <= int()

    Ran 2 tests in 0.046s

    FAILED (errors=1)
    ```

NOTE: according to the nose2 documentation, it should find the config in setup.cfg, but for some reasons ours is not, so I pass it explicitly.

ryanjulian:
Good work so far. I think we should wait on submitting this until
1. We convert the tests to actual unittest.TestCase classes
1. We go down the list of failing tests and determine which ones are broken by shallow bugs and which ones are Really Broken

jonashen:
https://github.com/rlworkgroup/garage/tree/unit_testing for a starting point.

ryanjulian:
superceded by #191

PR #190, opened 2018-07-12
Fix pre-commit checks in CI

ryanjulian:
This commit message intentionally violates the style guideeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,48 +7,86 @@ repos:
         language: script
         entry: scripts/check_commit_message
         stages: [commit-msg]
+
 -   repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v1.3.0
     # https://github.com/pre-commit/pre-commit-hooks/blob/master/README.md
+    stages: [commit, push]
     hooks:
     # -   id: autopep8-wrapper            # Runs autopep8 over python source.
+    #     stages: [commit, push]
     -   id: check-added-large-files     # Prevent giant files from being committed.
+        stages: [commit, push]
     -   id: check-ast                   # Simply check whether files parse as valid python.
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -7,48 +7,84 @@ repos:
...
     # -   id: sort-simple-yaml            # Sorts simple YAML files which consist only of top-level keys, preserving comments and blocks.
+    #     stages: [commit, push]
     -   id: trailing-whitespace         # Trims trailing whitespace.
+        stages: [commit, push]
+
 -   repo: https://github.com/pre-commit/mirrors-yapf
     rev: v0.22.0
     hooks:
     -   id: yapf
-        args: ['-dpr']
```
jonashen: Are you sure you can remove args? I had to keep them in order for both yapf and pylint to run correctly with pre-commit

.pre-commit-config.yaml
```diff
@@ -7,48 +7,84 @@ repos:
...
+        stages: [commit, push]
+
 -   repo: https://github.com/pre-commit/mirrors-pylint
     rev: v1.9.1
+    stages: [commit, push]
     hooks:
     -   id: pylint
-        args: ['--rcfile=setup.cfg']
```
jonashen: I think you need to keep args

jonashen: Or else pre-commit will say `No files to check(skipped)`

ryanjulian: Oops not sure how these got removed

jonashen:
Is `garage/scripts/check_commit_message` extraneous now?

ryanjulian:
No, it's still used by pre-commit

Issue #113, opened 2018-06-28
PPO for TF


ryanjulian:

ryanjulian:
See [1] for a baseline and [2] for an example of how to implement GPU-oriented on-policy algorithms in garage.
 
[1] https://github.com/openai/baselines/tree/master/baselines/ppo2
[2] https://github.com/ryanjulian/embed2learn/tree/master/algos

The implementation in garage should be quite similar to [2], but much simpler.

ryanjulian:
Fixed in #193

PR #193, opened 2018-07-13
Add PPO to garage/tf

CatherineSue:
Add Proximal Policy Optimization algorithm into garage TensorFlow tree.
PPO can be inherited from NPO. The difference among PPO, NPO and TRPO is
the constraint of the surrogate objective. This commit rewrites the
current NPO in garage tf so that it can run in GPU.

The policy, baseline and optimizer that PPO needs are already
implemented in the primitives of garage.

The process_samples function in garage.samplers.BaseSampler only
produces dense trajectories for RNN implementation. But it is faster if
we always produce dense trajs and flatten them in GPU. This commit adds
a process_samples function in garage.tf.samplers.BatchSampler and make
it the default sampler for TF tree. The VectorizedSampler inherites it.

Also rewrites the TRPO in garage/tf since the NPO implementation has
changed.

Adds flatten functions into garage/tf/misc/tensor_urils.py.

Refer to: #113

### DIFF
--- a/examples/tf/ppo_pendulum.py
+++ b/examples/tf/ppo_pendulum.py
@@ -1,0 +1,55 @@
+"""
+This is an example to train a task with PPO algorithm.
+
+Here it creates InvertedDoublePendulum using gym. And uses a PPO with 1M
+steps.
+
+Results:
+    AverageReturn: 526.092
+    RiseTime: itr 488
+"""
+import gym
+
+from garage.envs import normalize
+from garage.misc.instrument import run_experiment
+from garage.tf.algos import PPO
+from garage.tf.baselines import GaussianMLPBaseline
+from garage.tf.envs import TfEnv
+from garage.tf.policies import GaussianMLPPolicy
+
+
+def run_task(*_):
+    """
+    Wrap PPO training task in the run_task function.
+
+    :param _:
+    :return:
+    """
+    env = TfEnv(normalize(gym.make("InvertedDoublePendulum-v2")))
+
+    policy = GaussianMLPPolicy(
+        name="policy", env_spec=env.spec, hidden_sizes=(64, 64))
+
+    baseline = GaussianMLPBaseline(env_spec=env.spec)
+
+    algo = PPO(
+        env=env,
+        policy=policy,
+        baseline=baseline,
+        batch_size=2048,
+        max_path_length=100,
+        n_itr=488,
+        discount=0.99,
+        step_size=0.01,
+        optimizer_args=dict(batch_size=32, max_epochs=10),
+        plot=False)
+    algo.train()
+
+
+run_experiment(
+    run_task,
+    n_parallel=1,
+    snapshot_mode="last",
+    seed=1,
+    plot=False,
+)
--- a/garage/tf/algos/npo.py
+++ b/garage/tf/algos/npo.py
@@ -1,149 +1,101 @@
+from enum import Enum
+from enum import unique
+
+import numpy as np
 import tensorflow as tf
 
-from garage.misc import ext
-import garage.misc.logger as logger
+from garage.misc import logger
+from garage.misc import special
 from garage.misc.overrides import overrides
 from garage.tf.algos import BatchPolopt
 from garage.tf.misc import tensor_utils
-from garage.tf.optimizers import PenaltyLbfgsOptimizer
+from garage.tf.misc.tensor_utils import compute_adv, compute_ret
+from garage.tf.misc.tensor_utils import filter_valids
+from garage.tf.misc.tensor_utils import filter_valids_dict
+from garage.tf.misc.tensor_utils import flatten_batch
+from garage.tf.misc.tensor_utils import flatten_batch_dict
+from garage.tf.misc.tensor_utils import flatten_inputs
+from garage.tf.misc.tensor_utils import graph_inputs
+from garage.tf.optimizers import LbfgsOptimizer
+
+
+@unique
+class PGLoss(Enum):
+    VANILLA = "vanilla"
+    CLIP = "clip"
 
 
 class NPO(BatchPolopt):
-    """
-    Natural Policy Optimization.
-    """
-
     def __init__(self,
+                 pg_loss=PGLoss.VANILLA,
+                 clip_range=0.01,
                  optimizer=None,
                  optimizer_args=None,
-                 step_size=0.01,
                  name=None,
+                 policy=None,
+                 policy_ent_coeff=1e-2,
                  **kwargs):
+        self.name = name
+        self._name_scope = tf.name_scope(self.name)
+
+        self._pg_loss = pg_loss
         if optimizer is None:
             if optimizer_args is None:
                 optimizer_args = dict()
-            optimizer = PenaltyLbfgsOptimizer(**optimizer_args)
-        self.optimizer = optimizer
-        self.step_size = step_size
-        self.name = name
-        super(NPO, self).__init__(**kwargs)
+            optimizer = LbfgsOptimizer
+
+        with self._name_scope:
+            self.optimizer = optimizer(**optimizer_args)
+            self.clip_range = float(clip_range)
+            self.policy_ent_coeff = float(policy_ent_coeff)
+
+        super(NPO, self).__init__(policy=policy, **kwargs)
 
     @overrides
     def init_opt(self):
-        with tf.name_scope(self.name, "NPO"):
-            is_recurrent = int(self.policy.recurrent)
-            with tf.name_scope("inputs"):
-                obs_var = self.env.observation_space.new_tensor_variable(
-                    'obs',
-                    extra_dims=1 + is_recurrent,
-                )
-                action_var = self.env.action_space.new_tensor_variable(
-                    'action',
-                    extra_dims=1 + is_recurrent,
-                )
-                advantage_var = tensor_utils.new_tensor(
-                    'advantage',
-                    ndim=1 + is_recurrent,
-                    dtype=tf.float32,
-                )
-                dist = self.policy.distribution
+        if self.policy.recurrent:
+            raise NotImplementedError
 
-                old_dist_info_vars = {
-                    k: tf.placeholder(
-                        tf.float32,
-                        shape=[None] * (1 + is_recurrent) + list(shape),
-                        name='old_%s' % k)
-                    for k, shape in dist.dist_info_specs
-                }
-                old_dist_info_vars_list = [
-                    old_dist_info_vars[k] for k in dist.dist_info_keys
-                ]
+        pol_loss_inputs, pol_opt_inputs = self._build_inputs()
+        self._policy_opt_inputs = pol_opt_inputs
 
-                state_info_vars = {
-                    k: tf.placeholder(
-                        tf.float32,
-                        shape=[None] * (1 + is_recurrent) + list(shape),
-                        name=k)
-                    for k, shape in self.policy.state_info_specs
-                }
-                state_info_vars_list = [
-                    state_info_vars[k] for k in self.policy.state_info_keys
-                ]
+        pol_loss, pol_kl = self._build_policy_loss(pol_loss_inputs)
+        self.optimizer.update_opt(
+            loss=pol_loss,
+            target=self.policy,
+            leq_constraint=(pol_kl, self.clip_range),
+            inputs=flatten_inputs(self._policy_opt_inputs),
+            constraint_name="mean_kl")
 
-                if is_recurrent:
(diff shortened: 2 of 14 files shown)


garage/tf/algos/ppo.py
```diff
@@ -0,0 +1,22 @@
+from garage.tf.algos import NPO
+from garage.tf.algos import PGLoss
+from garage.tf.optimizers import FirstOrderOptimizer
+
+
+class PPO(NPO):
+    """
+    Proximal Policy Optimization with a Task Embedding
```
ryanjulian: No task embedding :)

garage/tf/misc/tensor_utils.py
```diff
@@ -10,6 +13,50 @@ def run(*input_vals):
     return run
 
 
+def flatten_batch(t, name=None):
+    with tf.name_scope(name, "flatten_batch", [t]):
+        shape = [-1] + list(t.shape[2:])
+        return tf.reshape(t, shape)
+
+
+def flatten_batch_dict(d, name=None):
```
ryanjulian: Can you verify that these don't already exist elsewhere?

CatherineSue: I didn't find it elsewhere but I'm not sure. I asked Jonathon for review because he refactored some utils in garage and garage/tf.

garage/tf/algos/npo.py
```diff
@@ -1,145 +1,411 @@
...
+        if hasattr(self.baseline, 'fit_with_samples'):
+            self.baseline.fit_with_samples(paths, samples_data)
+        else:
+            self.baseline.fit(paths)
+
+        return self.get_itr_snapshot(itr, samples_data)
 
     @overrides
-    def get_itr_snapshot(self, itr, samples_data):
+    def get_itr_snapshot(self, itr, _samples_data):
```
zhanpenghe: samples_data?

garage/tf/misc/tensor_utils.py
jonashen: I don't think we have this. Maybe you can consder placing these in a separate utils function for garage/tf (the only one we have right now is located in garage/envs/util)

examples/tf/ppo_pendulum.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def run_task(*_):
+    """
+    Wrap PPO training task in the run_task function.
+
+    :param _:
+    :return:
+    """
+    env = TfEnv(normalize(gym.make("InvertedDoublePendulum-v2")))
```
jonashen: You mention using HalfCheetah in your description, but use InvertedDoublePendulum here?

CatherineSue:
# Benchmarks 
This is the regression test result of Mujoco1M benchmark over garage ppo and baselines ppo2; each plot represents the AverageDiscountedReturn(the plot label 'AverageReturn' is a typo) over iterations in a particular task. All the tasks all tested under 1M steps.

## Hyperparameters
|Hyperparameter        |Value  |         
|:------------- |:-------------|
| Dicount      | 0.99 |
| GAE parameter | 0.95 |
| Minibatch size | 32 |
| Num. epochs | 10 |
| Horizon | 2048 |
| Policy hidden sizes | (64, 64) |

## InvertedDoublePendulum-v2

## InvertedPendulum-v2
invertedpendulum-v2_benchmark

## HelfCheetah-v2
Baselines ppo2 implementation can't evaluate HalfCheetah-v2 rewards correctly (nan). Here only shows the curve of garage ppo.
halfcheetah-v2_benchmark

## Hooper-v2
hopper-v2_benchmark

## Walker2d-v2
walker2d-v2_benchmark

## Reacher-v2
reacher-v2_benchmark

## Swimmer-v2
Baselines ppo2 also can't evaluate Swimmer-v2 rewards. Here only shows the curve of garage ppo.
swimmer-v2_benchmark

CatherineSue:
TODO: add recurrent policy implementation into NPO.

garage/tf/misc/tensor_utils.py
ryanjulian: I think it might be a feature of spaces?

Overall we need a strong convention for the shape of sampled data, and then it will be easy to produce flattened views in any NN framework.

The shape used here, which is fairly general, is (trajectory, rollout, data_1, data_2, ..., data_n) where the shape of a single data point is (data_1, ..., data_n).

We can file an issue make this uniform across the codebase later.

ryanjulian:
You can file an issue for recurrent policies.

It looks like the garage implementation is having trouble keeping up with baselines in some cases. Do you have an idea why? Is this the max_path_length issue you mentioned before?

CatherineSue:
I'm running multiple experiments trying to find why.
I don't think it's the max_path_length. I already change the baselines code to make sure the env resets after max_path_length.

garage/tf/samplers/batch_sampler.py
```diff
@@ -1,17 +1,20 @@
...
+    g.sess.__enter__()
 
 
-def worker_init_tf_vars(G):
-    G.sess.run(tf.global_variables_initializer())
+def worker_init_tf_vars(g):
+    g.sess.run(tf.global_variables_initializer())
 
 
 class BatchSampler(BaseSampler):
```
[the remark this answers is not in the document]
CatherineSue: No. I let VectorizedSampler inherits `process_samples` from BatchSampler. But we need to change VPG.py now that we flatten inputs in algorithms using tf_gpu.
All the comments above have been addressed.

garage/tf/algos/npo.py
```diff
@@ -153,3 +137,274 @@ def get_itr_snapshot(self, itr, samples_data):
...
+
+        policy_entropy = self._build_entropy_terms(i)
+
+        with tf.name_scope("augmented_rewards"):
+            rewards = i.reward_var \
+                      + self.policy_ent_coeff * policy_entropy
+
+        with tf.name_scope("policy_loss"):
+            with tf.name_scope("advantage"):
+                gamma_lambda = tf.constant(
```
ryanjulian: can you please move the advantage calculation into a function tensor_utils? it's a very subtle formulation which is easy to get wrong (it took me about a day to write). Please also retain my block comment explaining the method.

it should probably also be tested.

garage/tf/algos/npo.py
```diff
@@ -153,3 +137,274 @@ def get_itr_snapshot(self, itr, samples_data):
...
+        self.f_policy_entropy = tensor_utils.compile_function(
+            flatten_inputs(self._policy_opt_inputs),
+            tf.reduce_mean(policy_entropy * i.valid_var),
+            log_name="f_policy_entropy")
+
+        return policy_entropy
+
+    def _build_returns(self, rewards):
+        with tf.name_scope("returns"):
+            gamma = tf.constant(
```
ryanjulian: this is the same subroutine as the advantage calculation.

ryanjulian: this may be the same subroutine as the advantage calculation, but i recall it being slightly different.

returns are a backward-looking cumulative sum of rewards
r_i := reward at time i
R_i := return at time i
R_i = r_0 + r_1 + ... + r_i

advantages are a forward-looking sum of deltas from the baseline
A_i := advantage at time i
V_i := value function at time i
d_i := delta at time i
d_i := r_i - V_i
T := total length of trajectory

A_i = d_i + d_i+1 + ... + d_T-1 + d_T

CatherineSue: Yes. I have a question, why don't use `special.discount_cumsum()` here to calculate advantages and returns?

garage/tf/algos/npo.py
CatherineSue: em, how should i test it?

garage/tf/algos/npo.py
ryanjulian: special.discount_cumsum() is a numpy operation, so it's non-differentiable. this means we can differentiate through the advantage calculation to (for instance) an entropy bonus term.

this version does everything in tensorflow, so the entire loss function is differentiable. this means that the user can freely augment each time step of each reward with other differentiable terms, and tensorflow will automatically differentiate through them.

garage/tf/algos/npo.py
ryanjulian: a simple unit test with known correct input and output would suffice.

ryanjulian:
I think this is ready to submit once
* Minor comments are addressed
* TRPO results show acceptable performance compared to baselines/trpo_mpi
* There is a small unit test training an easy task (e.g. pendulum) which fails if the performance is too low
* There are large benchmarks automatically comparing with baselines/ppo and baselines/trpo_mpi, similar to test_ddpg_benchmark.py. One benchmark suite per algorithm.

CatherineSue:
# Benchmarks 
This is the regression test result of Mujoco1M benchmark over garage trpo and baselines trpo; each plot represents the AverageReturn over iterations in a particular task. All the tasks all tested under 1M steps. Only trained with one trail. Curve smoothed.

## Hyperparameters
|Hyperparameter        |Value  |         
|:------------- |:-------------|
| Dicount      | 0.99 |
| GAE parameter | 0.98 |
| Minibatch size | 32 |
| Num. epochs | 10 |
| Horizon | 1024 |
| Policy hidden sizes | (32, 32) |
| Baseline(value function) hidden sizes | (32, 32) |

## InvertedDoublePendulum-v2

## InvertedPendulum-v2
invertedpendulum-v2_benchmark

## HelfCheetah-v2
halfcheetah-v2_benchmark

## Hooper-v2
hopper-v2_benchmark

## Walker2d-v2
walker2d-v2_benchmark

## Reacher-v2
reacher-v2_benchmark

## Swimmer-v2
swimmer-v2_benchmark

ryanjulian:
I am concerned about the performance on reacher (which should be easy) and HalfCheetah (which is harder). Perhaps it is only a seeding issue though. Can you update with more seeds when you get the chance?

garage/tf/samplers/batch_sampler.py
```diff
@@ -41,3 +44,102 @@ def obtain_samples(self, itr):
             paths_truncated = parallel_sampler.truncate_paths(
                 paths, self.algo.batch_size)
             return paths_truncated
+
+    def process_samples(self, itr, paths):
```
ryanjulian: Does this break VPG or any other TF algo?

tests/tf/algos/test_trpo.py
```diff
@@ -0,0 +1,43 @@
...
+            baseline=baseline,
+            batch_size=1024,
+            max_path_length=100,
+            n_itr=10,
+            discount=0.99,
+            gae_lambda=0.98,
+            policy_ent_coeff=0.0,
+            plot=False,
+        )
+        algo.train()
```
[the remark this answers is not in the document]
ryanjulian: (and also do the same in TestPPO)

tests/benchmarks/test_benchmark_ppo.py
```diff
@@ -0,0 +1,245 @@
...
+from garage.envs import normalize
+from garage.misc import ext
+from garage.misc import logger as garage_logger
+from garage.tf.algos import PPO
+from garage.tf.baselines import GaussianMLPBaseline
+from garage.tf.envs import TfEnv
+from garage.tf.policies import GaussianMLPPolicy
+
+
+class TestBenchmarkPPO(unittest.TestCase):
```
ryanjulian: make sure this is marked huge

tests/benchmarks/test_benchmark_ppo.py
```diff
@@ -0,0 +1,245 @@
...
+    Create baselines model and training.
+
+    Replace the ppo and its training with the algorithm you want to run.
+
+    :param env: Environment of the task.
+    :param seed: Random seed for the trail.
+    :param log_dir: Log dir path.
+    :return
+    """
+    ncpu = 4
```
ryanjulian: Please select this based on the machine

garage/tf/samplers/batch_sampler.py
ryanjulian: Does it break anything else?

CatherineSue: I don't find it break other things.

ryanjulian: who else uses BatchSampler?

CatherineSue: BatchPolopt, NPO, PPO, and TRPO. These algorithms work well.

ryanjulian: i see. okay, please file a bug for VPG and put it on the 2018.08 milestone. it isn't really used much anyway (and it's not a hard fix).

PR #194, opened 2018-07-14
Remove enclosing scope

gonzaiva:
Enclosing scope is completely removed to avoid the unwanted duplication
of TensorFlow variables, and instead variable_scope is used to create
variables that will be further shared, while name_scope is used to group
operations that interact with the shared variables.

### DIFF
--- a/garage/misc/tensorboard_output.py
+++ b/garage/misc/tensorboard_output.py
@@ -13,7 +13,6 @@ from tensorboard.plugins.custom_scalar import metadata
 import tensorflow as tf
 
 from garage.misc.console import mkdir_p
-from garage.tf.misc.tensor_utils import enclosing_scope
 
 
 class TensorBoardOutput:
@@ -29,7 +28,6 @@ class TensorBoardOutput:
             'normal', 'gamma', 'poisson', 'uniform'
         ]
         self._feed = {}
-        self._name = "TensorBoardOutput/"
 
         self._default_step = 0
         self._writer = None
@@ -67,9 +65,9 @@ class TensorBoardOutput:
         self._dump_histogram(run_step)
         self._dump_tensors()
 
-    def record_histogram(self, key, val):
+    def record_histogram(self, key, val, name=None):
         if str(key) not in self._histogram_ds:
-            with enclosing_scope(self._name, "record_hist"):
+            with tf.name_scope(name, "record_hist"):
                 self._histogram_ds[str(key)] = tf.Variable(val)
             self._histogram_summary_op.append(
                 tf.summary.histogram(str(key), self._histogram_ds[str(key)]))
@@ -82,6 +80,7 @@ class TensorBoardOutput:
                                  histogram_type,
                                  key=None,
                                  shape=[1000],
+                                 name=None,
                                  **kwargs):
         '''
         distribution type and args:
@@ -99,7 +98,7 @@ class TensorBoardOutput:
 
         if str(key) not in self._histogram_ds:
             self._histogram_ds[str(key)] = self._get_histogram_var_by_type(
-                histogram_type, shape, **kwargs)
+                histogram_type, shape, name, **kwargs)
             self._histogram_summary_op.append(
                 tf.summary.histogram(
                     str(key), self._histogram_ds[str(key)][0]))
@@ -128,8 +127,12 @@ class TensorBoardOutput:
             self._scalars.value.add(
                 tag=key + '/' + str(idx).strip('()'), simple_value=float(v))
 
-    def _get_histogram_var_by_type(self, histogram_type, shape, **kwargs):
-        with enclosing_scope(self._name, "get_hist_{}".format(histogram_type)):
+    def _get_histogram_var_by_type(self,
+                                   histogram_type,
+                                   shape,
+                                   name=None,
+                                   **kwargs):
+        with tf.name_scope(name, "get_hist_{}".format(histogram_type)):
             if histogram_type == "normal":
                 # Make a normal distribution, with a shifting mean
                 mean = tf.Variable(kwargs['mean'])
--- a/garage/misc/tensorboard_output.py
+++ b/garage/misc/tensorboard_output.py
@@ -66,15 +66,16 @@ class TensorBoardOutput:
         self._dump_tensors()
 
     def record_histogram(self, key, val, name=None):
-        if str(key) not in self._histogram_ds:
-            with tf.name_scope(name, "record_hist"):
+        with tf.name_scope(name, "record_histogram"):
+            if str(key) not in self._histogram_ds:
                 self._histogram_ds[str(key)] = tf.Variable(val)
-            self._histogram_summary_op.append(
-                tf.summary.histogram(str(key), self._histogram_ds[str(key)]))
-            self._histogram_summary_op_merge = tf.summary.merge(
-                self._histogram_summary_op)
+                self._histogram_summary_op.append(
+                    tf.summary.histogram(
+                        str(key), self._histogram_ds[str(key)]))
+                self._histogram_summary_op_merge = tf.summary.merge(
+                    self._histogram_summary_op)
 
-        self._feed[self._histogram_ds[str(key)]] = val
+            self._feed[self._histogram_ds[str(key)]] = val
 
     def record_histogram_by_type(self,
                                  histogram_type,
--- a/garage/tf/algos/ddpg.py
+++ b/garage/tf/algos/ddpg.py
@@ -242,20 +242,21 @@ class DDPG(RLAlgorithm):
 
     def _initialize(self):
         with tf.name_scope(self.name, "DDPG"):
-            """Set up the actor, critic and target network."""
-            # Set up the actor and critic network
-            self.actor._build_net(trainable=True)
-            self.critic._build_net(trainable=True)
+            with tf.name_scope("setup_networks"):
+                """Set up the actor, critic and target network."""
+                # Set up the actor and critic network
+                self.actor._build_net(trainable=True)
+                self.critic._build_net(trainable=True)
 
-            # Create target actor and critic network
-            target_actor = copy(self.actor)
-            target_critic = copy(self.critic)
+                # Create target actor and critic network
+                target_actor = copy(self.actor)
+                target_critic = copy(self.critic)
 
-            # Set up the target network
-            target_actor.name = "TargetActor"
-            target_actor._build_net(trainable=False)
-            target_critic.name = "TargetCritic"
-            target_critic._build_net(trainable=False)
+                # Set up the target network
+                target_actor.name = "TargetActor"
+                target_actor._build_net(trainable=False)
+                target_critic.name = "TargetCritic"
+                target_critic._build_net(trainable=False)
 
             # Initialize replay buffer
             replay_buffer = ReplayBuffer(self.replay_buffer_size,
--- a/garage/tf/algos/ddpg.py
+++ b/garage/tf/algos/ddpg.py
@@ -119,8 +119,8 @@ class DDPG(RLAlgorithm):
         self.pause_for_plot = pause_for_plot
         self.actor_optimizer = actor_optimizer
         self.critic_optimizer = critic_optimizer
-        with tf.name_scope(name, "DDPG"):
-            self._initialize()
+        self.name = name
+        self._initialize()
 
     @overrides
     def train(self, sess=None):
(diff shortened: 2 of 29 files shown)


garage/tf/core/layers.py
```diff
@@ -833,8 +850,8 @@ def __init__(self, incoming, shape, **kwargs):
                 raise NotImplementedError
                 # if s.ndim != 0:
```
CatherineSue: Should here be deleted?

garage/tf/algos/ddpg.py
```diff
@@ -260,56 +261,64 @@ def _initialize(self):
...
+        with tf.name_scope("inputs"):
+            y = tf.placeholder(tf.float32, shape=(None, 1), name="input_y")
+            obs = tf.placeholder(
+                tf.float32,
+                shape=(None, self.observation_dim),
+                name="input_observation")
+            actions = tf.placeholder(
+                tf.float32, shape=(None, self.action_dim), name="input_action")
 
+        #obs = tf.identity(obs, "obs")
```
CatherineSue: Delete these comments?

garage/tf/policies/continuous_mlp_policy.py
```diff
@@ -61,6 +60,7 @@ def __init__(self,
         self._hidden_nonlinearity = hidden_nonlinearity
         self._output_nonlinearity = output_nonlinearity
         self._batch_norm = bn
+        self._q_network_name = "q_network"
```
CatherineSue: Isn't this a policy network?

CatherineSue:
Sorry! I think I pressed `close` button by mistake.

garage/tf/policies/gaussian_gru_policy.py
```diff
@@ -173,8 +184,8 @@ def __init__(
             LayersPowered.__init__(self, out_layers)
 
     @overrides
-    def dist_info_sym(self, obs_var, state_info_vars, name="dist_info_sym"):
-        with enclosing_scope(self.name, name):
+    def dist_info_sym(self, obs_var, state_info_vars, name=None):
+        with tf.name_scope(name, "dist_info", [obs_var, state_info_vars]):
```
jonashen: "dist_info_sym"?

garage/tf/policies/gaussian_mlp_policy.py
```diff
@@ -223,7 +227,8 @@ def get_reparam_action_sym(self,
         :param old_dist_info_vars:
         :return:
         """
-        with enclosing_scope(self.name, name):
+        with tf.name_scope(name, "get_reparam_action",
```
jonashen: get_reparam_action_sym?

garage/tf/core/layers.py
gonzaiva: According to git blame these comments have been here since 2016-08-07, so they don't seem to be useful anymore and I will remove them.

garage/tf/algos/ddpg.py
gonzaiva: My bad, I will remove them.

garage/tf/optimizers/penalty_lbfgs_optimizer.py
```diff
@@ -47,39 +44,41 @@ def update_opt(self,
...
         :param inputs: A list of symbolic variables as inputs
         :return: No return value.
         """
-        with enclosing_scope(self._name, name):
+        self._target = target
+        self._max_constraint_val = constraint_value
+        self._constraint_name = constraint_name
+        params = target.get_params(trainable=True)
+        with tf.name_scope(name, "PenaltyLbfgsOptimizer",
+                           [leq_constraint, loss, params]):
```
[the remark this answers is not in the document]
gonzaiva: This scope name is only applied to the method update_opt, not to the class PenaltyLbfgsOptimizer, so it's clearer to apply the scope name when the method is called by passing the name as a parameter of update_opt.

garage/tf/policies/continuous_mlp_policy.py
gonzaiva: It's mentioned [here](https://github.com/rlworkgroup/garage/blob/99324f0d215631ac09f2c1f89123520b864462ea/garage/tf/policies/continuous_mlp_policy.py#L67) that a Q Network is built, and since this scope is applied to the output of that network, that's why I decided to use that name, but I will change it to "policy_network" if it's more correct.

garage/tf/policies/gaussian_gru_policy.py
gonzaiva: Well, the sym mean symbolic, that is, the function builds the graph of operations that will be executed later. I think adding sym to the scope name is redundant, since the scope basically only encapsulates tensor operations and variables that are symbolic.

garage/tf/policies/gaussian_mlp_policy.py
gonzaiva: Similar explanation as above.

garage/tf/policies/gaussian_gru_policy.py
ryanjulian: We will remove the _sym suffixes later when we make everything a _sym operation. For now, the scope should match the function name so it's easy to figure out which function produced which graph element.

garage/misc/tensorboard_output.py
```diff
@@ -67,9 +65,9 @@ def dump_tensorboard(self, step=None):
         self._dump_histogram(run_step)
         self._dump_tensors()
 
-    def record_histogram(self, key, val):
+    def record_histogram(self, key, val, name=None):
         if str(key) not in self._histogram_ds:
-            with enclosing_scope(self._name, "record_hist"):
+            with tf.name_scope(name, "record_hist"):
                 self._histogram_ds[str(key)] = tf.Variable(val)
             self._histogram_summary_op.append(
```
ryanjulian: shouldn't adding to the summary op also go inside the scope?

garage/tf/algos/npo.py
```diff
@@ -27,12 +26,13 @@ def __init__(self,
         self.optimizer = optimizer
         self.step_size = step_size
         self.name = name
-        super(NPO, self).__init__(**kwargs)
+        with tf.name_scope(name):
```
ryanjulian: this is confusing to read. i think it would be better to edit BatchPolopt to use the scope correctly.

garage/tf/algos/npo.py
```diff
@@ -76,35 +76,46 @@ def init_opt(self):
...
-                                                       state_info_vars)
-            kl = dist.kl_sym(old_dist_info_vars, dist_info_vars)
-            lr = dist.likelihood_ratio_sym(action_var, old_dist_info_vars,
-                                           dist_info_vars)
-            if is_recurrent:
+        dist_info_vars = self.policy.dist_info_sym(obs_var, state_info_vars)
+        kl = dist.kl_sym(old_dist_info_vars, dist_info_vars)
+        lr = dist.likelihood_ratio_sym(action_var, old_dist_info_vars,
+                                       dist_info_vars)
+        mean_kl_scope = tf.name_scope("mean_kl", values=[kl, valid_var])
```
ryanjulian: this could be made much easier to read (eliminating the `mean_kl_scope` and `surr_loss_scope` variables) by splitting the blocks into a mean_kl block and a surr_loss block.

garage/tf/algos/vpg.py
```diff
@@ -87,37 +86,52 @@ def init_opt(self):
...
-
-            # formulate as a minimization problem
-            # The gradient of the surrogate objective is the policy gradient
-            if is_recurrent:
+        dist_info_vars = self.policy.dist_info_sym(obs_var, state_info_vars)
+        logli = dist.log_likelihood_sym(action_var, dist_info_vars)
+        kl = dist.kl_sym(old_dist_info_vars, dist_info_vars)
+        surr_obj_scope = tf.name_scope(
+            "surr_obj", values=[logli, advantage_var, valid_var])
+        mean_kl_scope = tf.name_scope("mean_kl", values=[kl, valid_var])
```
ryanjulian: same comment about splitting the block to eliminate saving scope variables (which is difficult to read)

garage/tf/policies/gaussian_gru_policy.py
gonzaiva: Got it, I completed all the scope names that were under symbolic functions.

garage/misc/tensorboard_output.py
gonzaiva: Yes, I moved the scope one line up to cover both the variables and summary.merge

garage/tf/algos/npo.py
gonzaiva: BatchPolopt only calls the init_opt of NPO and VPG, so I just moved this scope from the constructors to init_opt.

garage/tf/algos/npo.py
gonzaiva: Done

garage/tf/algos/ddpg.py
```diff
@@ -240,89 +241,99 @@ def train(self, sess=None):
...
-        self.opt_info = dict(
-            f_train_actor=f_train_actor,
-            f_train_critic=f_train_critic,
-            f_init_target=f_init_target,
-            f_update_target=f_update_target,
-            replay_buffer=replay_buffer,
-            target_critic=target_critic,
-            target_actor=target_actor)
+        with tf.name_scope(self.name, "DDPG"):
+            """Set up the actor, critic and target network."""
```
jonashen: would a `tf.name_scope(self.name, "setup_networks")` be appropriate here? i.e. from lines 245 (here) to 258 (`target_critic._build_net(trainable=False)`)

garage/tf/algos/npo.py
```diff
@@ -31,64 +30,74 @@ def __init__(self,
...
-                'action',
-                extra_dims=1 + is_recurrent,
-            )
-            advantage_var = tensor_utils.new_tensor(
-                'advantage',
-                ndim=1 + is_recurrent,
-                dtype=tf.float32,
-            )
-            dist = self.policy.distribution
+            with tf.name_scope("inputs"):
```
jonashen: These seem more like "setup_vars" to me

garage/tf/algos/vpg.py
```diff
@@ -41,51 +40,51 @@ def __init__(self,
...
-            state_info_vars_list = [
-                state_info_vars[k] for k in self.policy.state_info_keys
-            ]
-
-            if is_recurrent:
-                valid_var = tf.placeholder(
-                    tf.float32, shape=[None, None], name="valid")
-            else:
-                valid_var = None
+            with tf.name_scope("inputs"):
```
jonashen: "setup_vars"

jonashen: lmk what you think

garage/tf/algos/vpg.py
```diff
@@ -94,16 +93,30 @@ def init_opt(self):
...
+                else:
+                    mean_kl = tf.reduce_mean(kl)
+                tf.identity(mean_kl, name="mean_kl")
+
+            with tf.name_scope("max_kl", values=[kl, valid_var]):
+                if is_recurrent:
+                    max_kl = tf.reduce_max(kl * valid_var)
+                else:
+                    max_kl = tf.reduce_max(kl)
+                tf.identity(max_kl, name="max_kl")
```
jonashen: I like the way you separated the code here!

garage/tf/algos/vpg.py
```diff
@@ -94,16 +93,30 @@ def init_opt(self):
...
-                max_kl = tf.reduce_max(kl)
+            with tf.name_scope(
+                    "surr_obj", values=[logli, advantage_var, valid_var]):
+                if is_recurrent:
+                    surr_obj = (
+                        -tf.reduce_sum(logli * advantage_var * valid_var) /
+                        tf.reduce_sum(valid_var))
+                else:
+                    surr_obj = -tf.reduce_mean(logli * advantage_var)
+                tf.identity(surr_obj, name="surr_obj")
```
jonashen: what does `tf.identity` do?

garage/tf/optimizers/lbfgs_optimizer.py
```diff
@@ -27,29 +25,32 @@ def update_opt(self,
...
+        :param leq_constraint: A constraint provided as a tuple (f, epsilon),
+         of the form f(*inputs) <= epsilon.
         :param inputs: A list of symbolic variables as inputs
         :return: No return value.
         """
-        with enclosing_scope(self._name, name):
-            self._target = target
+        self._target = target
+        params = target.get_params(trainable=True)
+        with tf.name_scope(name, "LbfgsOptimizer",
```
jonashen: I'm a little confused about the naming here. Why call it "LbfgsOptimizer" here and not in `__init__`? Will `update_opt()` only be called once?

garage/tf/optimizers/lbfgs_optimizer.py
```diff
@@ -67,39 +68,40 @@ def loss(self, inputs, extra_inputs=None):
...
-        else:
-            opt_callback = None
-
-        scipy.optimize.fmin_l_bfgs_b(
-            func=f_opt_wrapper,
-            x0=self._target.get_param_values(trainable=True),
-            maxiter=self._max_opt_itr,
-            callback=opt_callback,
-        )
+            def f_opt_wrapper(flat_params):
```
jonashen: do we not need name_scopes here?

garage/tf/optimizers/lbfgs_optimizer.py
```diff
@@ -67,39 +68,40 @@ def loss(self, inputs, extra_inputs=None):
...
+                self._target.set_param_values(flat_params, trainable=True)
+                ret = f_opt(*inputs)
+                return ret
+
+            itr = [0]
+            start_time = time.time()
+
+            if self._callback:
+
+                def opt_callback(params):
```
jonashen: Or here

garage/tf/optimizers/lbfgs_optimizer.py
jonashen: Nvm I think you answered it below, but I was wondering if you could still clarify for me as I'm still a little confused

garage/tf/optimizers/penalty_lbfgs_optimizer.py
```diff
@@ -102,75 +101,77 @@ def loss(self, inputs):
...
 
-            return f
+            penalty_scale_factor = None
+            f_opt = self._opt_fun["f_opt"]
+            f_penalized_loss = self._opt_fun["f_penalized_loss"]
 
-        cur_params = self._target.get_param_values(
-            trainable=True).astype('float64')
-        opt_params = cur_params
+            def gen_f_opt(penalty):
```
jonashen: name_scope?

garage/tf/optimizers/penalty_lbfgs_optimizer.py
```diff
@@ -102,75 +101,77 @@ def loss(self, inputs):
...
+                def f(flat_params):
```
jonashen: hmm, if it's only used internally I guess you don't need it here

garage/tf/policies/categorical_conv_policy.py
```diff
@@ -42,44 +41,47 @@ def __init__(
...
-                      obs_var,
-                      state_info_vars=None,
-                      name="dist_info_sym"):
-        with enclosing_scope(self._name, name):
-            return dict(
-                prob=L.get_output(self._l_prob,
-                                  {self._l_obs: tf.cast(obs_var, tf.float32)}))
+    def dist_info_sym(self, obs_var, state_info_vars=None, name=None):
+        with tf.name_scope(name, "dist_info_sym", [obs_var]):
+            with tf.name_scope(self._prob_network_name, [obs_var]):
```
jonashen: what is the purpose of this second name_scope

garage/tf/algos/npo.py
ryanjulian: they are in fact inputs to the graph

garage/tf/algos/ddpg.py
gonzaiva: I could add a scope, but I think the name "networks" would be more appropiate, since the Actor, Critic and Targets that would be enclosed only contain the operations per layer in each network.
Do you agree?

garage/tf/algos/npo.py
gonzaiva: Although programatically we're setting up variables here, in the graph these represent the inputs of the operations in the algorithm.
For me, the use of the word setup implies operations, like setup_target in DDPG, but in this case we're only defining
the input variables.
Do you agree?

garage/tf/algos/vpg.py
gonzaiva: Thanks!

garage/tf/optimizers/lbfgs_optimizer.py
gonzaiva: Inside f_opt_wrapper there's two calls, one for set_param_values and one for f_opt. The former is defined [here](https://github.com/rlworkgroup/garage/blob/efb8e142206754582e4c83ec130024020e16a21e/garage/tf/core/parameterized.py#L69) and it already has a scope, while the latter is a TF compiled function that does not generate any variables or operations, so I found redundant to add another scope besides "optimize" to enclose f_opt_wrapper.

garage/tf/optimizers/lbfgs_optimizer.py
gonzaiva: No TF operations or variables are generated here.

garage/tf/optimizers/penalty_lbfgs_optimizer.py
gonzaiva: Similar explanation as above.

garage/tf/policies/categorical_conv_policy.py
gonzaiva: When the function [get_output](https://github.com/rlworkgroup/garage/blob/efb8e142206754582e4c83ec130024020e16a21e/garage/tf/core/layers.py#L2065) is called, a new instance of the operations in the neural network defined by the output layers provided as input is generated. Before calling this function it's necessary to add a scope to enclose all the operations that are generated, which is something that I did for the code on the TF branch, so it's clear in the graph that those operations belong to the networks.
Althogh this two scopes may seem redundant, since there can be more than one instance of the nerual network operations, add the scope for dist_info_sym makes more clear where the instances produced by ths method are in the graph.

garage/tf/policies/categorical_gru_policy.py
```diff
@@ -86,19 +86,24 @@ def __init__(
...
-                L.get_output([
-                    prob_network.step_output_layer,
-                    prob_network.step_hidden_layer
-                ], {prob_network.step_input_layer: feature_var}))
+                with tf.name_scope("feature_network", values=[flat_input_var]):
+                    feature_var = L.get_output(
+                        l_flat_feature,
+                        {feature_network.input_layer: flat_input_var})
+
+            with tf.name_scope(self._prob_network_name, values=[feature_var]):
```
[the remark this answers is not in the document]
gonzaiva: In this case I'm reusing the same name to produce different instances of the same network. Each set of network operations under name scope is different since their inputs and output are connected to different places in the graph, but they perform the same symbolic operations as the network they're instantiated from.

Issue #85, opened 2018-06-15
Refactor the Sawyer mujoco envs to GoalEnv


zhanpenghe:

zhanpenghe:
I am thinking about using inplementation like this:
MujocoEnv->GoalEnv->SawyerEnv->AnySpecificSawyerEnv

ryanjulian:
I suppose that's fine as long as all SawyerEnvs are always GoalEnvs. Is the definition of a GoalEnv that task completion can be represented by a sparse {0, 1} reward?

zhanpenghe:
I don't think it is part of the definition. According to gym's documentation, the reward should depend on a desiredGoal. But for our case it is true that our SawyerEnvs are all GoalEnvs.

ryanjulian:
This sounds fine to me as long as it simplifies the code. If someone gets more creative with Sawyer in the future, they will have to possibly undo this.

PR #198, opened 2018-07-16
Use gym.GoalEnv for sawyer envs

zhanpenghe:
#85

### DIFF
--- a/garage/envs/mujoco/sawyer/bin_sorting_env.py
+++ b/garage/envs/mujoco/sawyer/bin_sorting_env.py
@@ -2,11 +2,11 @@ import numpy as np
 
 from garage.core import Serializable
 from garage.envs import Step
-from garage.envs.mujoco import MujocoEnv
+from garage.envs.mujoco.sawyer.sawyer_env import SawyerEnv
 from garage.misc.overrides import overrides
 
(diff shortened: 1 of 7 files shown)


garage/envs/mujoco/sawyer/bin_sorting_env.py
```diff
@@ -2,11 +2,11 @@
 
 from garage.core import Serializable
 from garage.envs import Step
-from garage.envs.mujoco import MujocoEnv
+from garage.envs.mujoco.sawyer.sawyer_env import SawyerEnv
```
ryanjulian: Please use short imports

garage/envs/mujoco/sawyer/sawyer_env.py
```diff
@@ -0,0 +1,51 @@
+"""An interface for all sawyer robot environment."""
+
+import gym
+from gym.envs.robotics.utils import reset_mocap_welds
+
+from garage.envs.mujoco import MujocoEnv
```
ryanjulian: Please use short imports

zhanpenghe:
I did not use short import because of circular imports. I will submit after one more approval.

PR #207, opened 2018-07-20
Add a GitHandler to switch to a target commit 

gonzaiva:
The new class GitHandler is incorporated along with the function
run_experiment to allow the users to run experiments at a desired commit
based on a branch name (remote or local), tag name or SHA. The parameter
passed to run_experiment is a key value, for example:
- branch: feature_branch
- branch: origin/master
- tag: baseline_052318
- sha: 4d93a274bcd2de0

Besides this specific feature, GitHandler provides methods to interact
with simple git operations for future features.
Also, a test file is provided to make sure the implementation of
GitHandler is working correctly.

### DIFF
--- a/garage/config.py
+++ b/garage/config.py
@@ -56,8 +56,6 @@ FAST_CODE_SYNC = True
 
 FAST_CODE_SYNC_IGNORES = [".git", "data", ".pods"]
 
-GIT_REPO_URL = "https://github.com/rlworkgroup/garage.git"
-
 KUBE_DEFAULT_RESOURCES = {
     "requests": {
         "cpu": 0.8,
--- a/garage/config.py
+++ b/garage/config.py
@@ -56,6 +56,8 @@ FAST_CODE_SYNC = True
 
 FAST_CODE_SYNC_IGNORES = [".git", "data", ".pods"]
 
+GIT_REPO_URL = "https://github.com/rlworkgroup/garage.git"
+
 KUBE_DEFAULT_RESOURCES = {
     "requests": {
         "cpu": 0.8,
--- a/garage/misc/git_handler.py
+++ b/garage/misc/git_handler.py
@@ -7,16 +7,15 @@ from git import Repo
 from git.exc import BadName, GitCommandError
 from termcolor import colored
 
-from garage.config import GIT_REPO_URL
 from garage.config import PROJECT_PATH
+from garage.config_personal import GIT_REPO_URL
 
 
 class GitHandler:
     """Execute diverse git commands.
 
-    This class makes use of the sh package to execute git commands just like
-    in the terminal shell, with the convenience of grouping commands to perform
-    more complex tasks.
+    This class makes use of the GitPython library to execute git commands with
+    the convenience of grouping commands to perform more complex tasks.
 
     Preconditions of this class:
         - The working directory is not in the middle of a conflict
@@ -39,7 +38,7 @@ class GitHandler:
 
         """
         self.work_dir = work_directory
-        self.valid_targets = {
+        self.valid_references = {
             "branch": self.get_branch_sha,
             "sha": self.get_sha,
             "tag": self.get_tag_sha
@@ -60,20 +59,20 @@ class GitHandler:
                     break
         return valid_url
 
-    def create_branch(self, branch_name, target_ref="HEAD"):
-        """Create a branch on the target reference."""
-        self.repo.git.branch(branch_name, target_ref)
+    def create_branch(self, branch_name, ref="HEAD"):
+        """Create a branch on the reference."""
+        self.repo.git.branch(branch_name, ref)
 
-    def checkout_new_branch(self, branch_name, target_ref="HEAD"):
-        """Create a branch on the target reference."""
-        self.repo.git.checkout(branch_name, target_ref, b=True)
+    def checkout_new_branch(self, branch_name, ref="HEAD"):
+        """Create a branch on the reference."""
+        self.repo.git.checkout(branch_name, ref, b=True)
 
     def checkout(self, branch_name):
         """Checkout a branch."""
         self.repo.git.checkout(branch_name)
 
     def delete_branch(self, branch_name):
-        """Create a branch on the target reference."""
+        """Delete branch."""
         self.repo.git.branch(branch_name, D=True)
 
     def get_branch_sha(self, branch_name):
@@ -97,12 +96,12 @@ class GitHandler:
 
         return self.get_sha(full_branch_name)
 
-    def create_tag(self, tag_name, target_ref="HEAD"):
-        """Create a tag on the target reference."""
-        self.repo.git.tag(tag_name, target_ref)
+    def create_tag(self, tag_name, ref="HEAD"):
+        """Create a tag on the reference."""
+        self.repo.git.tag(tag_name, ref)
 
     def delete_tag(self, tag_name):
-        """Create a tag on the target reference."""
+        """Delete tag."""
         self.repo.git.tag(tag_name, d=True)
 
     def get_tag_sha(self, tag_name):
@@ -124,7 +123,7 @@ class GitHandler:
         HEAD is just a reference to the commit where the working directory is
         currently checked out.
 
-        If the reference is not valid, the exception sh.ErrorReturnCode_128 is
+        If the reference is not valid, the exception BadName or ValueError is
         thrown.
         """
         try:
@@ -172,18 +171,18 @@ class GitHandler:
             curr_branch_name = ""
         return curr_branch_name
 
-    def run_experiment_on_target(self, target, exp_args):
-        """Run a experiment on the indicated commit defined by target.
+    def run_experiment_on_ref(self, reference, exp_args):
+        """Run a experiment on the indicated commit defined by the reference.
 
-        After the target is verified, if there are any non-staged changes in
-        the working directory, they're stashed, the target commit is checked
+        After the reference is verified, if there are any non-staged changes in
+        the working directory, they're stashed, the reference commit is checked
         out, the experiment is executed and the working directory is restored
         at the end, including the stashed changes.
 
         Parameters
         ----------
-        - target: a key-value string, where the key indicates the type of
-            target (tag, branch, sha), and the value the reference for the
+        - reference: a key-value string, where the key indicates the type of
+            reference (tag, branch, sha), and the value the reference for the
             corresponding type. For example:
                 sha: 8d54s23a
                 branch: my_branch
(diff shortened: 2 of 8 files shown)


garage/misc/git_handler.py
```diff
@@ -0,0 +1,383 @@
...
+    Parameter
+    ---------
+    - command: it's an executed command from the sh package.
+
+    """
+    cmd_str = "%s" % command
+    srt_arr = cmd_str.split("\n")
+    # Remove next line character
+    srt_arr = srt_arr[:-1]
+    return srt_arr
```
CatherineSue: Why not `return str_arr[:-1]` here like function cmd_to_string in line 23? Also, `str_arr` instead of `srt_arr`?

gonzaiva: Nice catch, done!

garage/misc/git_handler.py
```diff
@@ -0,0 +1,382 @@
...
+        """
+        assert isinstance(target, str), ("The git target has to be a string")
+        self.target = target
+        self.target_sha = self._verify_target(self.target)
+        self.restore_sha = self.get_sha()
+        self.stash_sha = self.create_stash()
+        self.branch_name = self.get_current_branch_name()
+        self.target_branch_name = ("run_exp_on_" + str(int(time.time())))
+        try:
+            self._switch_to_target(self.target_branch_name, self.target_sha)
```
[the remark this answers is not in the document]
gonzaiva: It's true that we can use the same local branch or detached HEAD to switch between commits, but in the case of local branches, if an unexpected error happens and not even this [code](https://github.com/rlworkgroup/garage/blob/1e6d931f4fc163040b859450db769f7049365527/garage/misc/git_handler.py#L285) restores the branch correctly, I think it's less risky to not lose the reference of local commits under the branch by creating a new one just for the test.
It's also true that developers can use `git reflog` to look for these lost references, but maybe for someone not so experienced in git it could be easier just to have the reference of the original branch and delete the branch used for testing.

garage/misc/git_handler.py
```diff
@@ -0,0 +1,382 @@
...
+        # Imports are here to avoid circular dependencies within garage.misc
+        import garage.misc.instrument
+        from garage.misc import instrument
+        if "git_target" in exp_args:
+            exp_args.pop("git_target")
+        if "kwargs" in exp_args:
+            kwargs = exp_args["kwargs"]
+            exp_args.pop("kwargs")
+            exp_args.update(kwargs)
+        importlib.reload(garage.misc.instrument)
```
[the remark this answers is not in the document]
gonzaiva: For the cases where the target commit to test also includes changes in the function run_experiment (which is in garage.misc.instrument). Since the module is loaded in cache before performing the switching, the function run_experiment won't be that of the target commit unless it's reloaded.

garage/misc/git_handler.py
```diff
@@ -0,0 +1,382 @@
...
+        """Clean working directory and check out the target branch.
+
+        The target branch is created during the checkout operation.
+        SIGINT is temporarily blocked while performing the git operations to
+        avoid corrupting the working directory.
+        """
+        # Block SIGINT to avoid ending in an unknown state
+        signal.pthread_sigmask(signal.SIG_BLOCK, {signal.SIGINT})
+        # Clean working directory
+        self.reset()
```
zhanpenghe: Should we hard reset here before switching to target branch? If a launcher file is written and not commit, we will lose it with this reset right? I think a more proper way is throwing an error to guide the user to either stash the changes or commit it.

garage/misc/git_handler.py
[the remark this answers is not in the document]
zhanpenghe: Sounds good!

garage/misc/git_handler.py
gonzaiva: Sure Zhanpeng, I create a stash before reseting the working directory [here](https://github.com/rlworkgroup/garage/blob/a384e0a469d953353fac4cb210e9db30d6d0849a/garage/misc/git_handler.py#L205), and once the environment is restored, the stash is applied [here](https://github.com/rlworkgroup/garage/blob/a384e0a469d953353fac4cb210e9db30d6d0849a/garage/misc/git_handler.py#L327-L328).
I also print the sha of the stash that was created (if any unstaged changes were found) [here](https://github.com/rlworkgroup/garage/blob/a384e0a469d953353fac4cb210e9db30d6d0849a/garage/misc/git_handler.py#L227-L230), so the users are able to apply the stash if anything went wrong.

garage/misc/git_handler.py
```diff
@@ -0,0 +1,330 @@
...
+        valid_url = False
+        for remote in self.repo.remotes:
+            for remote_url in remote.urls:
+                if remote_url == repo_url:
+                    valid_url = True
+                    break
+        return valid_url
+
+    def create_branch(self, branch_name, target_ref="HEAD"):
+        """Create a branch on the targer reference."""
```
jonashen: target

garage/misc/git_handler.py
```diff
@@ -0,0 +1,330 @@
...
+        self.repo.git.branch(branch_name, target_ref)
+
+    def checkout_new_branch(self, branch_name, target_ref="HEAD"):
+        """Create a branch on the targer reference."""
```
jonashen: target

garage/misc/git_handler.py
```diff
@@ -0,0 +1,330 @@
...
+        self.repo.git.checkout(branch_name, target_ref, b=True)
+
+    def checkout(self, branch_name):
+        """Checkout a branch."""
+        self.repo.git.checkout(branch_name)
+
+    def delete_branch(self, branch_name):
+        """Create a branch on the targer reference."""
```
jonashen: target

garage/misc/git_handler.py
```diff
@@ -0,0 +1,330 @@
...
+
+        if remote_branch:
+            full_branch_name = "refs/remotes/" + branch_name
+        else:
+            full_branch_name = "refs/heads/" + branch_name
+
+        return self.get_sha(full_branch_name)
+
+    def create_tag(self, tag_name, target_ref="HEAD"):
+        """Create a tag on the targer reference."""
```
jonashen: target

garage/misc/git_handler.py
```diff
@@ -0,0 +1,330 @@
...
+        self.repo.git.tag(tag_name, target_ref)
+
+    def delete_tag(self, tag_name):
+        """Create a tag on the targer reference."""
```
jonashen: target

garage/misc/git_handler.py
```diff
@@ -0,0 +1,330 @@
...
+        finally:
+            self._restore_working_dir(self.target_branch_name, self.stash_sha,
+                                      self.branch_name, self.restore_sha)
+
+    def _verify_target(self, target):
+        """Verify the key-value target is valid.
+
+        Parameters
+        ----------
+        - target: same description as in method run_experiment_on_target.
```
jonashen: I think it's better to just copy-paste the actual description, rather than have people scroll back up to see what it was in the first place.

garage/misc/git_handler.py
gonzaiva: Done, as well  as the others.

garage/misc/git_handler.py
gonzaiva: Done.

garage/config.py
```diff
@@ -56,6 +56,8 @@
 
 FAST_CODE_SYNC_IGNORES = [".git", "data", ".pods"]
 
+GIT_REPO_URL = "https://github.com/rlworkgroup/garage.git"
```
ryanjulian: this belongs in config_personal_template.py

garage/misc/git_handler.py
```diff
@@ -0,0 +1,336 @@
...
+
+from garage.config import GIT_REPO_URL
+from garage.config import PROJECT_PATH
+from git import Repo
+
+
+class GitHandler:
+    """Execute diverse git commands.
+
+    This class makes use of the sh package to execute git commands just like
```
ryanjulian: it doesn't use the sh package anymore

garage/misc/git_handler.py
```diff
@@ -0,0 +1,336 @@
...
+    in the terminal shell, with the convenience of grouping commands to perform
+    more complex tasks.
+
+    Preconditions of this class:
+        - The working directory is not in the middle of a conflict
+        resolution.
+        - No branch is named HEAD.
```
ryanjulian: HEAD is not a valid branch name

garage/misc/git_handler.py
```diff
@@ -0,0 +1,336 @@
...
+
+class GitHandler:
+    """Execute diverse git commands.
+
+    This class makes use of the sh package to execute git commands just like
+    in the terminal shell, with the convenience of grouping commands to perform
+    more complex tasks.
+
+    Preconditions of this class:
+        - The working directory is not in the middle of a conflict
```
ryanjulian: You should detect this and exit if you can

garage/misc/git_handler.py
```diff
@@ -0,0 +1,336 @@
...
+        self.target_sha = self._verify_target(self.target)
+        self.restore_sha = self.get_sha()
+        self.stash_sha = self.create_stash()
+        self.branch_name = self.get_current_branch_name()
+        self.target_branch_name = ("run_exp_on_" + str(int(time.time())))
+        try:
+            self._switch_to_target(self.target_branch_name, self.target_sha)
+            current_sha = self.get_sha()
+            self._run_experiment(exp_args)
+        except BaseException:
```
ryanjulian: is there no more specific exception you can catch?

garage/misc/instrument.py
```diff
@@ -348,6 +349,7 @@ def run_experiment(stub_method_call=None,
                    script="scripts/run_experiment.py",
                    python_command="python",
                    mode="local",
+                   git_target=None,
```
ryanjulian: these are called refs or references in git terminology

https://git-scm.com/book/en/v2/Git-Internals-Git-References

garage/misc/git_handler.py
```diff
@@ -0,0 +1,336 @@
...
+        out, the experiment is executed and the working directory is restored
+        at the end, including the stashed changes.
+
+        Parameters
+        ----------
+        - target: a key-value string, where the key indicates the type of
+            target (tag, branch, sha), and the value the reference for the
+            corresponding type. For example:
+                sha: 8d54s23a
+                branch: my_branch
```
ryanjulian: git should be able to disambiguate these without knowing what kind of ref it is

garage/misc/git_handler.py
```diff
@@ -0,0 +1,336 @@
+"""Perform different git tasks while running experiments."""
```
ryanjulian: This is an immense amount of code for a small feature.

Can you explain what use cases this implements?

gonzaiva: Currently, run an experiment in a target commit, but all the code in this class is used for that purpose, except for the creation and removal of branches and tags, but those are required for my tests.
Also, the methods in this class will be used to implement two more features:
- Add the SHA to the experiment log.
- Create a new commit and push it to the remote

garage/misc/git_handler.py
gonzaiva: Ok, but I haven't not found a nice way to do this like a return value from a git command or an API from GitPython, so it may take some time.

garage/misc/git_handler.py
gonzaiva: I'm able to create a branch called HEAD with:
`git checkout -b HEAD`

garage/misc/git_handler.py
gonzaiva: That's true if all of them have different name, but what if they don't. For example:
```
$ git branch my_feature
$ git tag my_feature
$ git rev-parse my_feature
warning: refname 'my_feature' is ambiguous.
4421c2a81debf46079793eee05154f8a73ab3faa
```

garage/misc/git_handler.py
gonzaiva: Since this method calls run_experiment at garage.misc.instrument, I'm not sure about all the possible exceptions that may occur there, and I also want to catch exceptions such as KeyboardInterrupt.
Besides, the purpose of catching this exception is to notify the user of how to restore the working directory in case the restoration performed in the finally blocked is not executed correctly for unexpected reasons, and the exception is raised again after that.

garage/config.py
gonzaiva: Moved to config_personal_template.py

garage/misc/git_handler.py
gonzaiva: Removed all references of sh.

garage/misc/instrument.py
gonzaiva: The variables, methods and docstrings that contained target were changed to use ref or reference instead.

garage/misc/logger.py
```diff
@@ -331,7 +338,7 @@ def stub_to_json(stub_sth):
         return {stub_to_json(k): stub_to_json(v) for k, v in stub_sth.items()}
     elif isinstance(stub_sth, (list, tuple)):
         return list(map(stub_to_json, stub_sth))
-    elif type(stub_sth) == type(lambda: None):
+    elif isinstance(stub_sth, (types.FunctionType, types.BuiltinFunctionType)):
```
ryanjulian: what does this do?

gonzaiva: This changed is required to solve [E721](https://lintlyci.github.io/Flake8Rules/rules/E721.html). However, on the left side there's an anonymous function, so we need to compare if stub_sth is an instance of function or built-in function.
I've read this does not cover all cases, and using the function callable(stub_sh) could be a better approach. What do you think Ryan?

ryanjulian:

ryanjulian:

PR #208, opened 2018-07-23
Fix name scope in TensorBoardOutput

CatherineSue:
There is a bug in tensorboard_output.record_histogram, which creates `record_histogram/gass` space in TensorBaord instead of `gass`. This PR fixes it.


CatherineSue:
Comparison:
Before:
image

After:
screenshot from 2018-07-23 10-27-53

Issue #201, opened 2018-07-18
Move as many deps as possible to the 'pip' stanza of environment.yml


ryanjulian:
This will help ease the transition to a pip package in #32

PR #215, opened 2018-07-25
Move some easy deps from conda to pip

ryanjulian:
This PR moves the least-dangerous deps from conda to pip installation.

numpy/scipy have been omitted because they are fundamental deps with
binary platform-specific distributions, and thus most likely to break.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -4,55 +4,53 @@ channels:
     - https://conda.anaconda.org/conda-forge
 dependencies:
     - python==3.5.2
+    # Please keep alphabetized
+    - mkl-service=1.1.2
+    - mkl==2017.0.4
     - numpy==1.14.5
     - scipy
-    - mkl==2017.0.4
-    - path.py
-    - python-dateutil
-    - joblib==0.10.3
-    - mako
-    - ipywidgets
-    - flask
-    - matplotlib
-    - pandas
-    - mkl-service=1.1.2
-    # absl-py must be installed via conda rather than pip, because for some
-    # reason depedencies from the pip stanza are not available during the build
-    # phase of other packages in the pip stanza
-    - absl-py
     # dev dependencies below this line
     - pre_commit
     - pip:
-        - pyprind
-        - ipdb
+        # Please keep alphabetized
+        - awscli
         - boto3
-        - PyOpenGL
-        - pyzmq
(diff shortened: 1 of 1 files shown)


ryanjulian:
Partial fix for #201

environment.yml
```diff
@@ -4,55 +4,54 @@ channels:
...
         - hyperopt
-        - polling
-        - tensorboard
+        - ipdb
+        - ipywidgets
         - jsonmerge
-        - protobuf
+        - joblib==0.10.3
+        - jupyter
+        - git+https://github.com/jonashen/Lasagne.git@83065b9e1ff1758eaee9ce5bc494ceb5c804fdfd#egg=lasagne
```
jonashen: The most updated hash is now `b4fd25809c446f8c7b1c0d80283eb25c66e06960`

environment.yml
```diff
@@ -4,55 +4,54 @@ channels:
...
-    - mako
-    - ipywidgets
-    - flask
-    - matplotlib
-    - pandas
-    - mkl-service=1.1.2
-    # absl-py must be installed via conda rather than pip, because for some
-    # reason depedencies from the pip stanza are not available during the build
-    # phase of other packages in the pip stanza
-    - absl-py
```
jonashen: We don't need absl-py anymore?

ryanjulian: A "working" dm_control pip install actually fails the first time, then installs absl-py, then tries again and succeeds.

environment.yml
```diff
@@ -4,55 +4,53 @@ channels:
...
-        - pyzmq
-        - mujoco-py<1.50.2,>=1.50.1
-        - pygame
         - Box2D>=2.3.2
         - cached_property
         - cloudpickle
-        - git+https://github.com/Theano/Theano.git@3b51141a46affe9505f0e3f283020820b2c0251e#egg=theano
-        - git+https://github.com/jonashen/Lasagne.git@b4fd25809c446f8c7b1c0d80283eb25c66e06960#egg=lasagne
-        - git+https://github.com/plotly/plotly.py.git@2594076e29584ede2d09f2aa40a8a195b3f3fc66#egg=plotly
         - git+https://github.com/deepmind/dm_control.git#egg=dm_control
```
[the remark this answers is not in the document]
ryanjulian: oops. i will do that on the next commit

Issue #32, opened 2018-06-11
Replace conda with a pip package


ryanjulian:
conda makes it difficult to use rllab as a library.

We would like to transition to using the standard Python package interface. This will require getting all the dependencies to install using pip, plus probably a some custom setup scripts for setup.py.

Adding wheel compilation to the CI (e.g. appveyor) is also in-scope for this project.

Imported from https://github.com/ryanjulian/rllab/issues/81

ryanjulian:
I think pyenv+pipenv is the way to go for the dev environment, but I could be swayed otherwise.

ryanjulian:
#35

ryanjulian:
Fixed in #501 

environment.yml is still used in the CI

PR #223, opened 2018-07-26
Update to Python 3.6.6

ryanjulian:


jonashen:
Do we need to change our README? `garage only officially supports Python 3.5+.`

ryanjulian:
3.6.6 >= 3.5

ryanjulian:
I think the right thing to do would probably be to add a testing matrix to the .travis.yml which covers py35 and py36. We might also different versions of deps for py35 vs py36.

Unfortunately this would also double our CI time if run with every PR and maybe require us to maintain different dep lists (at least an environment.yml) for each version. I don't think this is warranted right now when de facto the only way to run garage is in conda, and the conda env has a set Python version.

A testing matrix is probably warranted when we do #32, which I'm slowing working towards by moving deps into the pip stanza.

PR #226, opened 2018-07-26
Utility functions to save files in Google Storage

gonzaiva:
One function checks if the authentication certificate is set and
retrieves the bucket from Google Storage. The other uploads the file(s)
specified in a path passed as parameter, giving the option of setting a
path inside the bucket.
The logs produced in run_experiment are saved if the parameter
save_in_gcs_bucket is True, and they're saved in the bucket set in the
file config_personal.py.

### DIFF
--- a/garage/config_personal_template.py
+++ b/garage/config_personal_template.py
@@ -47,3 +47,7 @@ AWS_ACCESS_KEY = os.environ.get("AWS_ACCESS_KEY", "<insert aws key>")
 AWS_ACCESS_SECRET = os.environ.get("AWS_ACCESS_SECRET", "<insert aws secret>")
 
 GCS_CREDENTIAL_PATH = "<insert file path to key>"
+
+GCS_BUCKET = "<insert bucket to drop experiment logs and checkpoints>"
+
+GCS_PATH_IN_BUCKET = ""
--- a/garage/config_personal_template.py
+++ b/garage/config_personal_template.py
@@ -45,3 +45,5 @@ DOCKER_CODE_DIR = "/root/code/garage"
 AWS_ACCESS_KEY = os.environ.get("AWS_ACCESS_KEY", "<insert aws key>")
 
 AWS_ACCESS_SECRET = os.environ.get("AWS_ACCESS_SECRET", "<insert aws secret>")
+
+GCS_CREDENTIAL_PATH = "<insert file path to key>"
--- a/garage/misc/gcs_utils.py
+++ b/garage/misc/gcs_utils.py
@@ -1,0 +1,60 @@
+import glob
+import os
+
+from google.cloud import storage
+from termcolor import colored
+
+from garage.config import GCS_CREDENTIAL_PATH
+
+
+def get_gcs_bucket(bucket: str):
+    """Obtain the GCS bucket.
+
+    It's verified that the JSON credentials to connect to GCS is set, and then
+    the bucket is retrieved.
+
+    Parameters
+    ----------
+    bucket: str
+        The bucket name as defined in GCS.
+
+    """
+    # Name of the environment variable used by the GCS library
+    GCS_ENV_VAR = "GOOGLE_APPLICATION_CREDENTIALS"
+    if GCS_ENV_VAR not in os.environ:
+        assert os.path.isfile(GCS_CREDENTIAL_PATH), colored(
+            "No credentials for GCS were provided. For instructions on " \
+            "how to setup credentials, see \"Set up a service acccount\":\n" \
+            "https://cloud.google.com/video-intelligence/docs/common"\
+            "/auth\n" \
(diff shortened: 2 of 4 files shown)


garage/misc/gcs_utils.py
```diff
@@ -0,0 +1,97 @@
...
+        assert os.path.isfile(GCS_CREDENTIAL_PATH), colored(
+            "No credentials for GCS were provided. For instructions on " \
+            "how to setup credentials, see \"Set up a service acccount\":\n" \
+            "https://cloud.google.com/video-intelligence/docs/common"\
+            "/auth\n" \
+            "Besides an environment variable, the certificate path can " \
+            "be set in field GCS_CREDENTIAL_PATH at config_personal.py.",
+            "yellow")
+        os.environ[gcs_env_var] = GCS_CREDENTIAL_PATH
+    storage_client = storage.Client()
```
zhanpenghe: Don't we need to specify a project?

gonzaiva: The project is already associated to the the GCS credential.

scripts/run_experiment.py
```diff
@@ -110,10 +111,20 @@ def run_experiment(argv):
         help='Pickled data for variant configuration')
     parser.add_argument(
         '--use_cloudpickle', type=ast.literal_eval, default=False)
+    parser.add_argument(
+        '--save_in_gcs_bucket',
```
ryanjulian: we want to also be able to save it to a prefix path in the bucket. 

this is achievable by having the user provide a full gcs path rather than just the bucket name. e.g.
`gs://my-cool-bucket/the/path/just/keeps/going`

gonzaiva: That's done by setting this [field](https://github.com/rlworkgroup/garage/blob/004c05f18b849bd6740952be69ced24a0f08ca11/garage/config_personal_template.py#L53) in the config_personal file. Should I pass it as a parameter of the function run_experiment instead?

garage/misc/gcs_utils.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+# Function get_gcs_bucket also works to verify that the certificate and bucket
+# name are correct, which is useful before running an experiment to avoid a
+# late error message once the training has finished.
+check_gcs_config = get_gcs_bucket
+
+
+def upload_to_gcs(files_to_upload_path: str,
+                  bucket_name: str = GCS_BUCKET,
```
ryanjulian: While I aspire to add type annotations to this codebase soon, for now they are out of place.

garage/config_personal_template.py
```diff
@@ -45,3 +45,9 @@
 AWS_ACCESS_KEY = os.environ.get("AWS_ACCESS_KEY", "<insert aws key>")
 
 AWS_ACCESS_SECRET = os.environ.get("AWS_ACCESS_SECRET", "<insert aws secret>")
+
+GCS_CREDENTIAL_PATH = "<insert file path to key>"
```
ryanjulian: load the key directly into memory from the environment (see AWS keys above) rather than using a file key--this makes it easier to use in Docker or on cloud hosts.

garage/misc/gcs_utils.py
```diff
@@ -0,0 +1,97 @@
...
+        Specify the path of the file or files to upload to GCS.
+    bucket_name: str
+        Name of the bucket in your GCS account.
+    path_in_bucket:
+        path prepended to the root folder or file uploaded to GCS.
+
+    """
+    bucket = get_gcs_bucket(bucket_name)
+    path_in_bucket = path_in_bucket.strip("/")
+    files_to_upload_path = files_to_upload_path.rstrip("/")
```
ryanjulian: don't mangle paths and URLs directly. use the os.path  and urllib modules, respectively.

garage/misc/gcs_utils.py
```diff
@@ -0,0 +1,97 @@
...
+    if os.path.isfile(files_to_upload_path):
```
ryanjulian: this all seems unnecessarily complicated. there is already a directly naming scheme used for local logs (in data/local). All this module needs to do is take whatever ends up in data/local as a result of an experiment and instead put it as some `gs://cool-bucket-bro/other/paths/here` URL instead.

you can use the application logic to find out what the most upper directory in use by this experiment is (usually its `data/local/exp_prefix` and individual runs go in `data/local/exp_prefix/exp_name`). By globbing `data/local/exp_prefix/exp_name` you can get every which needs uploading.

Another note: if this is only run at the very end of the experiment, it is almost certainly faster to call out to `gsutil -m rsync` with subprocess.Popen(), because gsutil implements some sophisticated multithreaded uploads which are difficult to replicate in your own script. It also implements rsync semantics which is pretty convenient.

ryanjulian:

PR #230, opened 2018-07-27
Fix bugs in garage/tf/baselines and optimizers

CatherineSue:
There are some missing lines in garage/tf/optimizers. Without them, the
penalty_lbfgs_optimizer can't work properly.

Add baselines to garage/tf so we can use tf/regressors.

### DIFF
--- a/garage/tf/optimizers/penalty_lbfgs_optimizer.py
+++ b/garage/tf/optimizers/penalty_lbfgs_optimizer.py
@@ -57,15 +57,16 @@ class PenaltyLbfgsOptimizer(Serializable):
         :param inputs: A list of symbolic variables as inputs
         :return: No return value.
         """
-        self._target = target
-        self._max_constraint_val = constraint_value
-        self._constraint_name = constraint_name
         params = target.get_params(trainable=True)
         with tf.name_scope(name, "PenaltyLbfgsOptimizer",
                            [leq_constraint, loss, params]):
             constraint_term, constraint_value = leq_constraint
             penalty_var = tf.placeholder(tf.float32, tuple(), name="penalty")
             penalized_loss = loss + penalty_var * constraint_term
+
+            self._target = target
+            self._max_constraint_val = constraint_value
+            self._constraint_name = constraint_name
 
             def get_opt_output():
                 with tf.name_scope(
@@ -132,8 +133,8 @@ class PenaltyLbfgsOptimizer(Serializable):
                     x0=cur_params,
                     maxiter=self._max_opt_itr)
 
-                _, try_loss, try_constraint_val = f_penalized_loss(
-                    *(inputs + (try_penalty, )))
+                _, try_loss, try_constraint_val = f_penalized_loss(*(
+                    inputs + (try_penalty, )))
 
                 logger.log('penalty %f => loss %f, %s %f' %
                            (try_penalty, try_loss, self._constraint_name,
--- a/garage/tf/regressors/deterministic_mlp_regressor.py
+++ b/garage/tf/regressors/deterministic_mlp_regressor.py
@@ -2,6 +2,7 @@ import numpy as np
 import tensorflow as tf
 
 from garage.core import Serializable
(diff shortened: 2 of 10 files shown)


garage/tf/regressors/gaussian_mlp_regressor.py
```diff
@@ -68,10 +69,14 @@ def __init__(self,
...
+                if optimizer_args is None:
+                    optimizer_args = dict()
                 if use_trust_region:
-                    optimizer = PenaltyLbfgsOptimizer("optimizer")
+                    optimizer = PenaltyLbfgsOptimizer(**optimizer_args)
                 else:
-                    optimizer = LbfgsOptimizer("optimizer")
+                    optimizer = LbfgsOptimizer(**optimizer_args)
+            else:
+                optimizer = optimizer(**optimizer_args)
```
gonzaiva: Is optimizer_args different from None in this else branch?

garage/tf/optimizers/penalty_lbfgs_optimizer.py
```diff
@@ -57,6 +57,10 @@ def update_opt(self,
         :param inputs: A list of symbolic variables as inputs
         :return: No return value.
         """
+        constraint_term, constraint_value = leq_constraint
+        penalty_var = tf.placeholder(tf.float32, tuple(), name="penalty")
+        penalized_loss = loss + penalty_var * constraint_term
```
gonzaiva: I see these variables are obtained again in the lines of code below, but under a scope. Why are they required outside the scope?

garage/tf/regressors/deterministic_mlp_regressor.py
```diff
@@ -42,7 +44,11 @@ def __init__(
         with tf.variable_scope(name, "DeterministicMLPRegressor"):
 
             if optimizer is None:
-                optimizer = LbfgsOptimizer(name="optimizer")
+                if optimizer_args is None:
+                    optimizer_args = dict()
+                optimizer = LbfgsOptimizer(**optimizer_args)
+            else:
+                optimizer = optimizer(**optimizer_args)
```
gonzaiva: Is optimizer_args different from None in this else branch?

garage/tf/optimizers/penalty_lbfgs_optimizer.py
CatherineSue: Oh my bad. I didn't see they are obtained under a scope.

CatherineSue: But isn't `self._max_constraint_val = constraint_value` requires the line 60?

garage/tf/regressors/gaussian_mlp_regressor.py
CatherineSue: No. Why?

gonzaiva: I saw you added this [guard](https://github.com/rlworkgroup/garage/blob/be570838f2fe4c5f291b10247209d04548565370/garage/tf/regressors/gaussian_mlp_regressor.py#L72-L73) for null args, so I was wondering if the same could happen when optimizer is already set, but if that's not the case, it's fine.

garage/tf/optimizers/penalty_lbfgs_optimizer.py
gonzaiva: Oh you're right, and using git blame it seems I'm responsible for that. Could you please change the code to:
```
        params = target.get_params(trainable=True)
        with tf.name_scope(name, "PenaltyLbfgsOptimizer",
                           [leq_constraint, loss, params]):
            constraint_term, constraint_value = leq_constraint
            penalty_var = tf.placeholder(tf.float32, tuple(), name="penalty")
            penalized_loss = loss + penalty_var * constraint_term

            self._target = target
            self._max_constraint_val = constraint_value
            self._constraint_name = constraint_name

            def get_opt_output():
```
Sorry for the inconvenience.

Issue #231, opened 2018-07-27
Add pickled mode support to GarageEnv/TfEnv/TheanoEnv


ryanjulian:
In order to support pickled mode, envs need to implement Serializable/Parameterized.

Obviously a gym.Env is not going to do that. Since these features are important and problems keep recurring, I think we should centralize the interaction for garage <--> gym into these wrapper.

* There should probably be a root class GarageEnv which TheanoEnv/TfEnv inherit from
* Eliminate special treatment of spaces -- all envs passed around garage inherit from GarageEnv (and therefore provide garage.spaces instead of gym.spaces
* All garage.envs still only implement gym.Env
* GarageEnv implements Parameterized so it will work with existing samplers and checkpointing (pickled mode)
* GarageEnv silently converts gym.Spaces to garage.spaces

ryanjulian:
if GarageEnv can implement gym.Wrapper, that's even better

ryanjulian:
I want to add type checking in #224 to avoid these recurring type compat problems, but for now we are stuck with what we have.

Feel free to add `assert issubclass(env, GarageEnv)` anywhere an env is passed and we can commit it for now to stem the bleeding.

Issue #239, opened 2018-07-30
Delete stub()


jonashen:
```
Traceback (most recent call last):
  File "/Users/jonathon/Documents/garage/garage/scripts/run_experiment.py", line 191, in <module>
    run_experiment(sys.argv)
  File "/Users/jonathon/Documents/garage/garage/scripts/run_experiment.py", line 146, in run_experiment
    logger.log_parameters_lite(params_log_file, args)
  File "/Users/jonathon/Documents/garage/garage/garage/misc/logger.py", line 372, in log_parameters_lite
    json.dump(log_params, f, indent=2, sort_keys=True, cls=MyEncoder)
  File "/anaconda2/envs/garage/lib/python3.6/json/__init__.py", line 179, in dump
    for chunk in iterable:
  File "/anaconda2/envs/garage/lib/python3.6/json/encoder.py", line 430, in _iterencode
    yield from _iterencode_dict(o, _current_indent_level)
  File "/anaconda2/envs/garage/lib/python3.6/json/encoder.py", line 404, in _iterencode_dict
    yield from chunks
  File "/anaconda2/envs/garage/lib/python3.6/json/encoder.py", line 404, in _iterencode_dict
    yield from chunks
  File "/anaconda2/envs/garage/lib/python3.6/json/encoder.py", line 404, in _iterencode_dict
    yield from chunks
  [Previous line repeated 1 more times]
  File "/anaconda2/envs/garage/lib/python3.6/json/encoder.py", line 437, in _iterencode
    o = _default(o)
  File "/Users/jonathon/Documents/garage/garage/garage/misc/logger.py", line 352, in default
    return json.JSONEncoder.default(self, o)
  File "/anaconda2/envs/garage/lib/python3.6/json/encoder.py", line 180, in default
    o.__class__.__name__)
TypeError: Object of type 'TimeLimit' is not JSON serializable
```

ryanjulian:
Seems to be attempting to serializing a gym.Env? I think this should be covered by #240

jonashen:
This error persisted even after the changes made in my PR. I'm not sure if it's because I implemented serialization wrong?

ryanjulian:
Let's hold off on merging #240 until you root cause this bug.

jonashen:
After I normalize the environment, i.e. `env = TfEnv(normalize(gym.make("CartPole-v0")))` instead of `env = TfEnv(gym.make("CartPole-v0"))`, the example works.

[...]

Issue #236, opened 2018-07-27
Replace ProxyEnv with gym.Wrapper


jonashen:
Once GarageEnv is implemented in #231, ProxyEnv will be redundant and all implementations of it should be removed and replaced with GarageEnv.

ryanjulian:
Uh I think the most logical replacement is gym.Wrapper, actually.

jonashen:
GarageEnv will eventually inherit from gym.Wrapper though

ryanjulian:
A gym.Wrapper can wrap a gym.Wrapper. We still have classes other than GarageEnv which need to wrap a gym.Env

PR #240, opened 2018-07-30
Add pickled mode support to GarageEnv/TfEnv/TheanoEnv

jonashen:
This PR refactors TfEnv and TheanoEnv to inherit from a base GarageEnv (that inherits from gym.Wrapper) which supports Parameterization and Serialization. Assert statements are also added whenever envs are passed into functions as a safety check.

In short, gym environments are now expected to be wrapped within TfEnv() or TheanoEnv() to convert to a local garage environment to ensure compatibility with garage functions and policies.

Ref: #231

### DIFF
--- a/examples/theano/trpo_cartpole.py
+++ b/examples/theano/trpo_cartpole.py
@@ -10,7 +10,6 @@ env = TheanoEnv(normalize(CartpoleEnv()))
 policy = GaussianMLPPolicy(env_spec=env.spec, hidden_sizes=(32, 32))
 
 baseline = LinearFeatureBaseline(env_spec=env.spec)
-
 algo = TRPO(
     env=env,
     policy=policy,
--- a/garage/algos/cem.py
+++ b/garage/algos/cem.py
@@ -171,7 +171,6 @@ class CEM(RLAlgorithm, Serializable):
                 np.mean([len(path['returns']) for path in paths]))
 
             self.policy.set_param_values(best_x)
-            self.env.log_diagnostics(paths)
             self.policy.log_diagnostics(paths)
             logger.save_itr_params(
                 itr,
--- a/garage/algos/cma_es.py
+++ b/garage/algos/cma_es.py
@@ -128,9 +128,7 @@ class CMAES(RLAlgorithm, Serializable):
             logger.record_tabular(
                 'AvgTrajLen',
                 np.mean([len(info['returns']) for info in infos]))
-            self.env.log_diagnostics(infos)
             self.policy.log_diagnostics(infos)
-
             logger.save_itr_params(
                 itr, dict(
                     itr=itr,
--- a/garage/contrib/bichengcao/examples/trpo_gym_Acrobot-v1.py
+++ b/garage/contrib/bichengcao/examples/trpo_gym_Acrobot-v1.py
@@ -2,25 +2,25 @@ import gym
 
 from garage.baselines import LinearFeatureBaseline
 from garage.envs import normalize
-from garage.envs.util import horizon, spec
 from garage.misc.instrument import run_experiment
 from garage.theano.algos import TRPO
+from garage.theano.envs import TheanoEnv
 from garage.theano.policies import CategoricalMLPPolicy
 
 
 def run_task(*_):
-    env = normalize(gym.make("Acrobot-v1"))
+    env = TheanoEnv(normalize(gym.make("Acrobot-v1")))
 
-    policy = CategoricalMLPPolicy(env_spec=spec(env), hidden_sizes=(32, 32))
+    policy = CategoricalMLPPolicy(env_spec=env.spec, hidden_sizes=(32, 32))
 
-    baseline = LinearFeatureBaseline(env_spec=spec(env))
+    baseline = LinearFeatureBaseline(env_spec=env.spec)
 
     algo = TRPO(
         env=env,
         policy=policy,
         baseline=baseline,
         batch_size=4000,
-        max_path_length=horizon(env),
+        max_path_length=env.max_episode_steps,
         n_itr=50,
         discount=0.99,
         step_size=0.01,
--- a/garage/contrib/bichengcao/examples/trpo_gym_CartPole-v0.py
+++ b/garage/contrib/bichengcao/examples/trpo_gym_CartPole-v0.py
@@ -20,7 +20,7 @@ def run_task(*_):
         policy=policy,
         baseline=baseline,
         batch_size=4000,
-        max_path_length=env.horizon,
+        max_path_length=env.max_episode_steps,
         n_itr=50,
         discount=0.99,
         step_size=0.01,
--- a/garage/contrib/bichengcao/examples/trpo_gym_CartPole-v0.py
+++ b/garage/contrib/bichengcao/examples/trpo_gym_CartPole-v0.py
@@ -2,25 +2,25 @@ import gym
 
 from garage.baselines import LinearFeatureBaseline
 from garage.envs import normalize
-from garage.envs.util import horizon, spec
 from garage.misc.instrument import run_experiment
 from garage.theano.algos import TRPO
+from garage.theano.envs import TheanoEnv
 from garage.theano.policies import CategoricalMLPPolicy
 
 
 def run_task(*_):
-    env = normalize(gym.make("CartPole-v0"))
+    env = TheanoEnv(normalize(gym.make("CartPole-v0")))
 
-    policy = CategoricalMLPPolicy(env_spec=spec(env), hidden_sizes=(32, 32))
+    policy = CategoricalMLPPolicy(env_spec=env.spec, hidden_sizes=(32, 32))
 
-    baseline = LinearFeatureBaseline(env_spec=spec(env))
+    baseline = LinearFeatureBaseline(env_spec=env.spec)
 
     algo = TRPO(
         env=env,
         policy=policy,
         baseline=baseline,
         batch_size=4000,
-        max_path_length=horizon(env),
+        max_path_length=env.horizon,
         n_itr=50,
         discount=0.99,
         step_size=0.01,
--- a/garage/contrib/bichengcao/examples/trpo_gym_CartPole-v1.py
+++ b/garage/contrib/bichengcao/examples/trpo_gym_CartPole-v1.py
@@ -20,7 +20,7 @@ def run_task(*_):
         policy=policy,
         baseline=baseline,
         batch_size=4000,
-        max_path_length=env.horizon,
+        max_path_length=env.max_episode_steps,
         n_itr=50,
         discount=0.99,
         step_size=0.01,
--- a/garage/contrib/bichengcao/examples/trpo_gym_CartPole-v1.py
+++ b/garage/contrib/bichengcao/examples/trpo_gym_CartPole-v1.py
@@ -2,25 +2,25 @@ import gym
 
 from garage.baselines import LinearFeatureBaseline
 from garage.envs import normalize
-from garage.envs.util import horizon, spec
 from garage.misc.instrument import run_experiment
 from garage.theano.algos import TRPO
+from garage.theano.envs import TheanoEnv
 from garage.theano.policies import CategoricalMLPPolicy
 
 
 def run_task(*_):
-    env = normalize(gym.make("CartPole-v1"))
+    env = TheanoEnv(normalize(gym.make("CartPole-v1")))
 
-    policy = CategoricalMLPPolicy(env_spec=spec(env), hidden_sizes=(32, 32))
+    policy = CategoricalMLPPolicy(env_spec=env.spec, hidden_sizes=(32, 32))
 
-    baseline = LinearFeatureBaseline(env_spec=spec(env))
+    baseline = LinearFeatureBaseline(env_spec=env.spec)
 
     algo = TRPO(
         env=env,
         policy=policy,
         baseline=baseline,
         batch_size=4000,
-        max_path_length=horizon(env),
+        max_path_length=env.horizon,
         n_itr=50,
         discount=0.99,
         step_size=0.01,
--- a/garage/contrib/bichengcao/examples/trpo_gym_MountainCar-v0.py
+++ b/garage/contrib/bichengcao/examples/trpo_gym_MountainCar-v0.py
@@ -21,7 +21,7 @@ def run_task(*_):
         policy=policy,
         baseline=baseline,
         batch_size=4000,
-        max_path_length=env.horizon,
+        max_path_length=env.max_episode_steps,
         n_itr=150,
         discount=0.99,
         step_size=0.1,
--- a/garage/contrib/bichengcao/examples/trpo_gym_MountainCar-v0.py
+++ b/garage/contrib/bichengcao/examples/trpo_gym_MountainCar-v0.py
@@ -3,25 +3,25 @@ import gym
 
 from garage.baselines import LinearFeatureBaseline
 from garage.envs import normalize
-from garage.envs.util import horizon, spec
 from garage.misc.instrument import run_experiment
 from garage.theano.algos import TRPO
+from garage.theano.envs import TheanoEnv
 from garage.theano.policies import CategoricalMLPPolicy
 
 
 def run_task(*_):
-    env = normalize(gym.make("MountainCar-v0"))
+    env = TheanoEnv(normalize(gym.make("MountainCar-v0")))
 
-    policy = CategoricalMLPPolicy(env_spec=spec(env), hidden_sizes=(32, 32))
+    policy = CategoricalMLPPolicy(env_spec=env.spec, hidden_sizes=(32, 32))
 
-    baseline = LinearFeatureBaseline(env_spec=spec(env))
+    baseline = LinearFeatureBaseline(env_spec=env.spec)
 
     algo = TRPO(
         env=env,
         policy=policy,
         baseline=baseline,
         batch_size=4000,
-        max_path_length=horizon(env),
+        max_path_length=env.horizon,
         n_itr=150,
         discount=0.99,
         step_size=0.1,
--- a/garage/contrib/bichengcao/examples/trpo_gym_Pendulum-v0.py
+++ b/garage/contrib/bichengcao/examples/trpo_gym_Pendulum-v0.py
@@ -20,7 +20,7 @@ def run_task(*_):
         policy=policy,
         baseline=baseline,
         batch_size=4000,
-        max_path_length=env.horizon,
+        max_path_length=env.max_episode_steps,
         n_itr=50,
         discount=0.99,
         step_size=0.01,
--- a/garage/contrib/bichengcao/examples/trpo_gym_Pendulum-v0.py
+++ b/garage/contrib/bichengcao/examples/trpo_gym_Pendulum-v0.py
@@ -2,25 +2,25 @@ import gym
 
 from garage.baselines import LinearFeatureBaseline
 from garage.envs import normalize
-from garage.envs.util import horizon, spec
 from garage.misc.instrument import run_experiment
 from garage.theano.algos import TRPO
+from garage.theano.envs import TheanoEnv
 from garage.theano.policies import GaussianMLPPolicy
 
 
 def run_task(*_):
-    env = normalize(gym.make("Pendulum-v0"))
+    env = TheanoEnv(normalize(gym.make("Pendulum-v0")))
 
-    policy = GaussianMLPPolicy(env_spec=spec(env), hidden_sizes=(32, 32))
+    policy = GaussianMLPPolicy(env_spec=env.spec, hidden_sizes=(32, 32))
 
-    baseline = LinearFeatureBaseline(env_spec=spec(env))
+    baseline = LinearFeatureBaseline(env_spec=env.spec)
 
     algo = TRPO(
         env=env,
         policy=policy,
         baseline=baseline,
         batch_size=4000,
-        max_path_length=horizon(env),
+        max_path_length=env.horizon,
         n_itr=50,
         discount=0.99,
         step_size=0.01,
--- a/garage/envs/base.py
+++ b/garage/envs/base.py
@@ -1,17 +1,162 @@
-import collections
-
+"""Wrapper class that converts gym.Env into GarageEnv."""
 from cached_property import cached_property
+import collections
+import warnings
 
-from garage.envs import EnvSpec
+import gym
+from gym.spaces import Box as GymBox
+from gym.spaces import Discrete as GymDiscrete
+from gym.spaces import Tuple as GymTuple
+
+from garage.core import Parameterized
+from garage.core import Serializable
+from garage.envs.env_spec import EnvSpec
+from garage.misc.overrides import overrides
+from garage.spaces import Box
+from garage.spaces import Discrete
+from garage.spaces import Product
+
+
+class GarageEnv(gym.Wrapper, Parameterized, Serializable):
+    """
+    Returns a Garage wrapper class for gym.Env.
+
+    In order to provide pickling (serialization) and parameterization
+    for gym.Envs, they must be wrapped with GarageEnv. This ensures
+    compatibility with existing samplers and checkpointing when the
+    envs are passed internally around garage.
+
+    Furthermore, GarageEnv silently converts action_space and
+    observation_space from gym.Spaces to garage.spaces.
+
+    Args:
+        env (gym.Env): the env that will be wrapped
+    """
+
+    def __init__(self, env):
+        Serializable.quick_init(self, locals())
+        Parameterized.__init__(self)
+        super().__init__(env)
+
+    @cached_property
+    @overrides
+    def action_space(self):
+        """
+        Classes inheriting from GarageEnv need to convert
+        action_space from gym.space to garage.space.
+
+        Returns:
+            NotImplementedError
+        """
+        return self._to_garage_space(self.env.action_space)
+
+    @cached_property
+    @overrides
+    def observation_space(self):
+        """
+        Classes inheriting from GarageEnv need to convert
+        observation_space from gym.space to garage.space.
+
+        Returns:
+            NotImplementedError
+        """
+        return self._to_garage_space(self.env.observation_space)
+
+    def close(self):
+        """
+        Close the wrapped env.
+
+        Returns:
+            None
+        """
+        self.env.close()
+
+    def get_params_internal(self, **tags):
+        """
+        Returns an empty list if env.get_params() is called.
+
+        Returns:
+            An empty list
+        """
+        warnings.warn("get_params_internal is deprecated", DeprecationWarning)
+        return []
+
+    @property
+    def horizon(self):
+        """
+        Get the maximum episode steps for the wrapped env.
+
+        Returns:
+            max_episode_steps (int)
+        """
+        if self.env.spec is not None:
+            return self.env.spec.max_episode_steps
+        else:
+            return NotImplementedError
+
+    def log_diagnostics(self, paths, *args, **kwargs):
+        """No env supports this function call."""
+        warnings.warn("log_diagnostics is deprecated", DeprecationWarning)
+        pass
+
+    @property
+    def spec(self):
+        """
+        Returns an EnvSpec.
+
+        Returns:
+            spec (garage.envs.EnvSpec)
+        """
+        return EnvSpec(
+            observation_space=self.observation_space,
+            action_space=self.action_space,
+        )
+
+    def reset(self, **kwargs):
+        """
+        Inheriting gym.Wrapper requires implementing this method.
+
+        Calls reset on wrapped env.
+        """
+        return self.env.reset(**kwargs)
+
+    def step(self, action):
+        """
+        Inheriting gym.Wrapper requires implementing this method.
+
+        Calls step on wrapped env.
+        """
+        return self.env.step(action)
+
+    def _to_garage_space(self, space):
+        """
+        Converts a gym space into a garage space.
+
+        Args:
+            space (gym.spaces)
+
+        Returns:
+            space (garage.spaces)
+        """
+        if isinstance(space, GymBox):
+            return Box(low=space.low, high=space.high)
+        elif isinstance(space, GymDiscrete):
+            return Discrete(space.n)
+        elif isinstance(space, GymTuple):
+            return Product(list(map(self._to_tf_space, space.spaces)))
+        else:
+            raise NotImplementedError
+
+
+def Step(observation, reward, done, **kwargs):  # noqa: N802
+    """
+    Convenience method for creating a namedtuple from the results of
+    environment.step(action). Provides the option to put extra
+    diagnostic info in the kwargs (if it exists) without demanding
+    an explicit positional argument.
+    """
+    return _Step(observation, reward, done, kwargs)
+
 
 _Step = collections.namedtuple("Step",
                                ["observation", "reward", "done", "info"])
-
-
-def Step(observation, reward, done, **kwargs):
-    """
-    Convenience method creating a namedtuple with the results of the
-    environment.step method.
-    Put extra diagnostic info in the kwargs
-    """
-    return _Step(observation, reward, done, kwargs)
--- a/garage/envs/base.py
+++ b/garage/envs/base.py
@@ -1,66 +1,34 @@
 """Wrapper class that converts gym.Env into GarageEnv."""
-from cached_property import cached_property
+from abc import ABCMeta, abstractmethod
 import collections
 import warnings
 
 import gym
-from gym.spaces import Box as GymBox
-from gym.spaces import Discrete as GymDiscrete
-from gym.spaces import Tuple as GymTuple
 
 from garage.core import Parameterized
 from garage.core import Serializable
-from garage.envs.env_spec import EnvSpec
-from garage.misc.overrides import overrides
-from garage.spaces import Box
-from garage.spaces import Discrete
-from garage.spaces import Product
 
 
-class GarageEnv(gym.Wrapper, Parameterized, Serializable):
+class GarageEnv(gym.Wrapper, Parameterized, Serializable, metaclass=ABCMeta):
     """
-    Returns a Garage wrapper class for gym.Env.
+    Returns an abstract Garage wrapper class for gym.Env.
 
     In order to provide pickling (serialization) and parameterization
-    for gym.Envs, they must be wrapped with GarageEnv. This ensures
+    for gym.Envs, they must be wrapped with a GarageEnv. This ensures
     compatibility with existing samplers and checkpointing when the
     envs are passed internally around garage.
 
-    Furthermore, GarageEnv silently converts action_space and
-    observation_space from gym.Spaces to garage.spaces.
+    Furthermore, classes inheriting from GarageEnv should silently
+    convert action_space and observation_space from gym.Spaces to
+    garage.spaces.
 
-    Args:
-        env (gym.Env): the env that will be wrapped
+    Args: env (gym.Env): the env that will be wrapped
     """
 
     def __init__(self, env):
         Serializable.quick_init(self, locals())
         Parameterized.__init__(self)
         super().__init__(env)
-
-    @cached_property
-    @overrides
-    def action_space(self):
-        """
-        Classes inheriting from GarageEnv need to convert
-        action_space from gym.space to garage.space.
-
-        Returns:
-            NotImplementedError
-        """
-        return self._to_garage_space(self.env.action_space)
-
-    @cached_property
-    @overrides
-    def observation_space(self):
-        """
-        Classes inheriting from GarageEnv need to convert
-        observation_space from gym.space to garage.space.
-
-        Returns:
-            NotImplementedError
-        """
-        return self._to_garage_space(self.env.observation_space)
 
     def close(self):
         """
@@ -99,22 +67,20 @@ class GarageEnv(gym.Wrapper, Parameterized, Serializable):
         warnings.warn("log_diagnostics is deprecated", DeprecationWarning)
         pass
 
-    @property
+    @abstractmethod
     def spec(self):
         """
-        Returns an EnvSpec.
+        Returns an EnvSpec with garage.spaces.
 
         Returns:
             spec (garage.envs.EnvSpec)
         """
-        return EnvSpec(
-            observation_space=self.observation_space,
-            action_space=self.action_space,
-        )
+        raise NotImplementedError
 
     def reset(self, **kwargs):
         """
-        Inheriting gym.Wrapper requires implementing this method.
+        This method is necessary to suppress a deprecated warning
+        thrown by gym.Wrapper.
 
         Calls reset on wrapped env.
         """
@@ -122,15 +88,17 @@ class GarageEnv(gym.Wrapper, Parameterized, Serializable):
 
     def step(self, action):
         """
-        Inheriting gym.Wrapper requires implementing this method.
+        This method is necessary to suppress a deprecated warning
+        thrown by gym.Wrapper.
 
         Calls step on wrapped env.
         """
         return self.env.step(action)
 
+    @abstractmethod
     def _to_garage_space(self, space):
         """
-        Converts a gym space into a garage space.
+        Converts a gym.space into a garage.space.
 
         Args:
             space (gym.spaces)
@@ -138,14 +106,7 @@ class GarageEnv(gym.Wrapper, Parameterized, Serializable):
         Returns:
             space (garage.spaces)
         """
-        if isinstance(space, GymBox):
-            return Box(low=space.low, high=space.high)
-        elif isinstance(space, GymDiscrete):
-            return Discrete(space.n)
-        elif isinstance(space, GymTuple):
-            return Product(list(map(self._to_tf_space, space.spaces)))
-        else:
-            raise NotImplementedError
+        raise NotImplementedError
 
 
 def Step(observation, reward, done, **kwargs):  # noqa: N802
--- a/garage/envs/box2d/box2d_env.py
+++ b/garage/envs/box2d/box2d_env.py
@@ -9,7 +9,6 @@ from garage.envs import Step
 from garage.envs.box2d.box2d_viewer import Box2DViewer
 from garage.envs.box2d.parser.xml_box2d import find_body, find_joint
 from garage.envs.box2d.parser.xml_box2d import world_from_xml
-from garage.envs.util import bounds, flat_dim
 from garage.misc import autoargs
 from garage.misc.overrides import overrides
 
@@ -124,13 +123,13 @@ class Box2DEnv(gym.Env):
 
     @property
     def action_bounds(self):
-        return bounds(self.action_space)
+        return self.action_space.low, self.action_space.high
 
     def forward_dynamics(self, action):
-        if len(action) != flat_dim(self.action_space):
+        if action.shape != self.action_space.low.shape:
             raise ValueError(
                 'incorrect action dimension: expected %d but got '
-                '%d' % (flat_dim(self.action_space)), len(action))
+                '%d' % (len(action.shape), len(self.action_space.low.shape)))
         lb, ub = self.action_bounds
         action = np.clip(action, lb, ub)
         for ctrl, act in zip(self.extra_data.controls, action):
--- a/garage/envs/box2d/car_parking_env.py
+++ b/garage/envs/box2d/car_parking_env.py
@@ -4,8 +4,8 @@ import pygame
 from garage.core import Serializable
 from garage.envs.box2d.box2d_env import Box2DEnv
 from garage.envs.box2d.parser import find_body
+from garage.envs.box2d.parser import find_joint
 from garage.envs.box2d.parser.xml_box2d import _get_name
-from garage.envs.util import flat_dim
 from garage.misc import autoargs
 from garage.misc.overrides import overrides
 
@@ -15,9 +15,8 @@ class CarParkingEnv(Box2DEnv, Serializable):
     @autoargs.arg(
         "random_start",
         type=bool,
-        help=
-        "Randomized starting position by uniforming sampling starting car angle"
-        "and position from a circle of radius 5")
+        help="Randomized starting position by uniforming sampling starting"
+        " car angle and position from a circle of radius 5")
     @autoargs.arg(
         "random_start_range",
         type=float,
@@ -120,9 +119,10 @@ class CarParkingEnv(Box2DEnv, Serializable):
 
     @overrides
     def forward_dynamics(self, action):
-        if len(action) != flat_dim(self.action_space) + 1:
-            raise ValueError('incorrect action dimension: expected %d but got '
-                             '%d' % (flat_dim(self.action_space), len(action)))
+        if len(action) != np.prod(self.action_space.low.shape) + 1:
+            raise ValueError(
+                'incorrect action dimension: expected %d but got '
+                '%d' % (np.prod(self.action_space.low.shape), len(action)))
         lb, ub = self.action_bounds
         action = np.clip(action, lb, ub)
         for ctrl, act in zip(self.extra_data.controls, action):
--- a/garage/envs/box2d/car_parking_env.py
+++ b/garage/envs/box2d/car_parking_env.py
@@ -6,6 +6,7 @@ from garage.envs.box2d.box2d_env import Box2DEnv
 from garage.envs.box2d.parser import find_body
 from garage.envs.box2d.parser import find_joint
 from garage.envs.box2d.parser.xml_box2d import _get_name
+from garage.envs.util import flat_dim
 from garage.misc import autoargs
 from garage.misc.overrides import overrides
 
@@ -119,10 +120,9 @@ class CarParkingEnv(Box2DEnv, Serializable):
 
     @overrides
     def forward_dynamics(self, action):
-        if len(action) != np.prod(self.action_space.low.shape) + 1:
-            raise ValueError(
-                'incorrect action dimension: expected %d but got '
-                '%d' % (np.prod(self.action_space.low.shape), len(action)))
+        if len(action) != flat_dim(self.action_space) + 1:
+            raise ValueError('incorrect action dimension: expected %d but got '
+                             '%d' % (flat_dim(self.action_space), len(action)))
         lb, ub = self.action_bounds
         action = np.clip(action, lb, ub)
         for ctrl, act in zip(self.extra_data.controls, action):
--- a/garage/envs/mujoco/gather/gather_env.py
+++ b/garage/envs/mujoco/gather/gather_env.py
@@ -21,7 +21,6 @@ from garage.envs.mujoco.gather.embedded_viewer import EmbeddedViewer
 from garage.envs.mujoco.mujoco_env import BIG
 from garage.envs.mujoco.mujoco_env import MODEL_DIR
 from garage.envs.proxy_env import ProxyEnv
-from garage.envs.util import flat_dim
 from garage.misc import autoargs
 from garage.misc import logger
 from garage.misc.overrides import overrides
@@ -467,7 +466,7 @@ class GatherEnv(ProxyEnv, Serializable):
                 stripped_path[k] = v
             stripped_path['observations'] = \
                 stripped_path['observations'][
-                    :, :flat_dim(self.wrapped_env.observation_space)]
+                    :, :np.prod(self.wrapped_env.observation_space.low.shape)]
             #  this breaks if the obs of the robot are d>1 dimensional (not a
             #  vector)
             stripped_paths.append(stripped_path)
--- a/garage/envs/mujoco/gather/gather_env.py
+++ b/garage/envs/mujoco/gather/gather_env.py
@@ -21,6 +21,7 @@ from garage.envs.mujoco.gather.embedded_viewer import EmbeddedViewer
 from garage.envs.mujoco.mujoco_env import BIG
 from garage.envs.mujoco.mujoco_env import MODEL_DIR
 from garage.envs.proxy_env import ProxyEnv
+from garage.envs.util import flat_dim
 from garage.misc import autoargs
 from garage.misc import logger
 from garage.misc.overrides import overrides
@@ -466,7 +467,7 @@ class GatherEnv(ProxyEnv, Serializable):
                 stripped_path[k] = v
             stripped_path['observations'] = \
                 stripped_path['observations'][
-                    :, :np.prod(self.wrapped_env.observation_space.low.shape)]
+                    :, :flat_dim(self.wrapped_env.observation_space)]
             #  this breaks if the obs of the robot are d>1 dimensional (not a
             #  vector)
             stripped_paths.append(stripped_path)
--- a/garage/envs/mujoco/hill/terrain.py
+++ b/garage/envs/mujoco/hill/terrain.py
@@ -11,8 +11,6 @@ import numpy as np
 from scipy.signal import convolve2d
 from scipy.stats import multivariate_normal
 
-from garage.envs.util import flat_dim
-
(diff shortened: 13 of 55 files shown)


garage/envs/base.py
```diff
@@ -1,17 +1,132 @@
...
+
+    Args:
+        env (gym.Env): the env that will be wrapped
+    """
+
+    def __init__(self, env):
+        super(GarageEnv, self).__init__(env)
+        Parameterized.__init__(self)
+        Serializable.quick_init(self, locals())
+        self._wrapped_env = env
```
ryanjulian: All of these are already implemented by gym.Wrapper

https://github.com/openai/gym/blob/master/gym/core.py#L238

ryanjulian: Why store this if gym.Wrapper already populates self.env?

garage/envs/base.py
```diff
@@ -1,17 +1,132 @@
...
+        self.action_space = self.env.action_space
+        self.observation_space = self.env.observation_space
```
ryanjulian: These are returning the gym.Spaces. A GarageEnv should return garage.spaces.

garage/envs/base.py
```diff
@@ -1,17 +1,132 @@
...
+        """
+        Get the maximum episode steps for the wrapped env.
+
+        Returns:
+            max_episode_steps (int)
+        """
+        return self._wrapped_env.spec.tags[
+            'wrapper_config.TimeLimit.max_episode_steps']
+
+    def log_diagnostics(self, paths, *args, **kwargs):
```
ryanjulian: log_diagnostics() is not part of the gym interface so you can't call it on the underlying env. 

If this is necessary because of calls elsewhere, implement it as pass and then make an issue to remove all said calls.

garage/envs/base.py
```diff
@@ -1,17 +1,132 @@
...
+    def wrapped_env(self):
+        """
+        Returns the original env wrapped by this class.
+
+        Returns:
+            _wrapped_env (gym.Env)
+        """
+        return self._wrapped_env
+
+    def _to_garage_space(self, space):
```
ryanjulian: This is a private method which is never called. Who uses it?

garage/theano/envs/base.py
```diff
@@ -5,47 +5,47 @@
...
+    def __init__(self, env):
+        super(TheanoEnv, self).__init__(env)
+        self.action_space = self._to_theano_space(self.action_space)
+        self.observation_space = self._to_theano_space(self.observation_space)
 
+    @overrides
     @cached_property
     def spec(self):
-        """Return an EnvSpec."""
+        """Returns an EnvSpec."""
```
ryanjulian: Why can't tis be in the parent class?

garage/envs/base.py
```diff
@@ -1,17 +1,132 @@
...
+        """
+        Returns an empty list if env.get_params() is called.
+
+        Returns:
+            An empty list
+        """
+        return []
+
+    @property
+    def horizon(self):
```
ryanjulian: Isn't this part of the gym interface?

jonashen: `horizon` is not

garage/envs/base.py
jonashen: Since both TheanoEnv and TfEnv inherit from GarageEnv, and they both have different specifications about what the Space looks like, I decided to convert them to garage.spaces inside TheanoEnv and TfEnv rather than GarageEnv, and just leave GarageEnv abstract.

garage/envs/base.py
jonashen: https://github.com/openai/gym/search?q=horizon&unscoped_q=horizon

garage/envs/base.py
ryanjulian: Then calling GarageEnv.action_space should raise NotImplementedError, instead of silently returning a mostly-but-not-fully-compatible type which will propagate an obscure error 30 stack frames deep across a process boundary.

jonashen: good idea

garage/envs/base.py
ryanjulian: Are you guaranteed that `self._wrapped_env.spec.tags['wrapper_config.TimeLimit.max_episode_steps']` exists for every env we will ever wrap?

Also, that tag is deprecated: https://github.com/openai/gym/blob/master/gym/envs/registration.py#L48

Why not just replace all calls to horizon with env.spec.max_episode_steps?
```python
>>> e = gym.make('Pendulum-v0')
WARN: gym.spaces.Box autodetected dtype as <class 'numpy.float32'>. Please provide explicit dtype.
WARN: gym.spaces.Box autodetected dtype as <class 'numpy.float32'>. Please provide explicit dtype.
>>> dir(e)
['__class__', '__delattr__', '__dict__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__gt__', '__hash__', '__init__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', '__weakref__', '_elapsed_seconds', '_elapsed_steps', '_episode_started_at', '_max_episode_seconds', '_max_episode_steps', '_past_limit', '_warn_double_wrap', 'action_space', 'class_name', 'close', 'compute_reward', 'env', 'metadata', 'observation_space', 'render', 'reset', 'reward_range', 'seed', 'spec', 'step', 'unwrapped']
>>> e.spec
EnvSpec(Pendulum-v0)
>>> dir(e.spec)
['__class__', '__delattr__', '__dict__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__gt__', '__hash__', '__init__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', '__weakref__', '_entry_point', '_env_name', '_kwargs', '_local_only', 'id', 'make', 'max_episode_seconds', 'max_episode_steps', 'nondeterministic', 'reward_threshold', 'tags', 'timestep_limit', 'trials']
>>>
```

ryanjulian: Ahh, not all envs have a spec.

jonashen: But that just means we should either first check whether env has a spec inside `horizon`, or just replace all uses of horizon with `env.spec.max_episode_steps`

ryanjulian: in case 2, what do we do if the env doesn't have a spec?

jonashen: https://github.com/openai/gym/issues/463#issuecomment-389873434

Looks like I can just return `env._max_episode_steps`

ryanjulian: uhh no you can't because it's _-private

ryanjulian: also:
```python
(garage36) rjulian@nyquist:~/code/garage$ python
Python 3.6.6 | packaged by conda-forge | (default, Jul 26 2018, 09:53:17)
[GCC 4.8.2 20140120 (Red Hat 4.8.2-15)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from garage.envs.box2d import MountainCarEnv
pygame 1.9.4
Hello from the pygame community. https://www.pygame.org/contribute.html
>>> m = MountainCarEnv()
>>> m._max_episode_steps
  File "<stdin>", line 1, in <module>
AttributeError: 'MountainCarEnv' object has no attribute '_max_episode_steps'
>>>
```

jonashen: Then I'll raise a NotImplementedError within horizon, so that if someone decides to use MountainCarEnv (for instance), `env.horizon` is not a valid call

ryanjulian: according to my ag, no environments even define horizon

```sh
(garage36) rjulian@nyquist:~/.../garage/garage/envs$ ag horizon
mujoco/randomization/randomized_env.py
77:    def horizon(self):
78:        return self._wrapped_env.horizon

util.py
13:    'bounds', 'default_value', 'flat_dim', 'flatten', 'flatten_n', 'horizon',
74:def horizon(env):

proxy_env.py
36:    def horizon(self):
37:        return self._wrapped_env.horizon

box2d/mountain_car_env.py
17:    @autoargs.arg("goal_cart_pos", type=float, help="Goal horizontal position")
```

garage/envs/base.py
```diff
@@ -1,17 +1,156 @@
...
+
+        Returns:
+            spec (garage.envs.EnvSpec)
+        """
+        return EnvSpec(
+            observation_space=self.observation_space,
+            action_space=self.action_space,
+        )
+
+    def step(self, action):
```
ryanjulian: covered by gym.Wrapper

garage/envs/base.py
```diff
@@ -1,17 +1,156 @@
...
+        return self.env.step(action)
+
+    def wrapped_env(self):
```
ryanjulian: implemented by gym.Wrapper.unwrapped

garage/envs/base.py
```diff
@@ -1,17 +1,156 @@
...
+            raise NotImplementedError
 
-from garage.envs import EnvSpec
 
 _Step = collections.namedtuple("Step",
                                ["observation", "reward", "done", "info"])
 
 
-def Step(observation, reward, done, **kwargs):
+def Step(observation, reward, done, **kwargs):  # noqa: N802
```
ryanjulian: Hmm why not just `Step = namedtuple(...)` above? Seems like directly out of the documentation.

https://docs.python.org/3.6/library/collections.html#collections.namedtuple

garage/envs/util.py
```diff
@@ -1,7 +1,7 @@
 import gym
```
ryanjulian: we can't delete this file entirely?

garage/envs/base.py
```diff
@@ -1,17 +1,156 @@
...
+    observation_space from gym.Spaces to garage.spaces.
+
+    Args:
+        env (gym.Env): the env that will be wrapped
+    """
+
+    def __init__(self, env):
+        Serializable.quick_init(self, locals())
+        Parameterized.__init__(self)
+        super(GarageEnv, self).__init__(env)
```
ryanjulian: should not be necessary in Python 3 (I think it should inherit the first __init__(), i.e. gym.Wrapper)

garage/envs/base.py
```diff
@@ -1,17 +1,156 @@
...
+        """
+        if self.env.spec is not None:
+            return self.env.spec.max_episode_steps
+        else:
+            return NotImplementedError
+
+    def log_diagnostics(self, paths, *args, **kwargs):
+        pass
+
+    def render(self, *args, **kwargs):
```
ryanjulian: implemented by gym.Wrapper

garage/envs/base.py
```diff
@@ -1,17 +1,156 @@
...
+        return self.env.render(*args, **kwargs)
+
+    def reset(self, **kwargs):
```
ryanjulian: implemented by gym.Wrapper

garage/envs/base.py
```diff
@@ -1,17 +1,156 @@
...
+    def wrapped_env(self):
+        """
+        Returns the original env wrapped by this class.
+
+        Returns:
+            env (gym.Env)
+        """
+        return self.env
+
+    def _to_garage_space(self, space):
```
ryanjulian: is this ever called anywhere?

garage/envs/base.py
```diff
@@ -1,17 +1,156 @@
...
+        """
+        Classes inheriting from GarageEnv need to convert
+        observation_space from gym.space to garage.space.
+
+        Returns:
+            NotImplementedError
+        """
+        raise NotImplementedError
+
+    def close(self):
```
ryanjulian: implemented by gym.Wrapper

garage/envs/base.py
```diff
@@ -1,17 +1,156 @@
...
+
+        Returns:
+            max_episode_steps (int)
+        """
+        if self.env.spec is not None:
+            return self.env.spec.max_episode_steps
+        else:
+            return NotImplementedError
+
+    def log_diagnostics(self, paths, *args, **kwargs):
```
ryanjulian: Perhaps raise a DeprecationWarning?

https://docs.python.org/3.6/library/warnings.html

garage/envs/base.py
```diff
@@ -1,17 +1,156 @@
...
+    def close(self):
+        """
+        Close the wrapped env.
+
+        Returns:
+            None
+        """
+        self.env.close()
+
+    def get_params_internal(self, **tags):
```
ryanjulian: Perhaps raise a DeprecationWarning?

https://docs.python.org/3.6/library/warnings.html

garage/envs/base.py
ryanjulian: You could make this also abstract (e.g. NotImplementedError) and then implement it concretely in TfEnv and TheanoEnv

garage/tf/envs/base.py
```diff
@@ -32,29 +19,26 @@ def __call__(self, *args, **kwargs):
...
+    """
 
-    @cached_property
-    def spec(self):
-        return EnvSpec(
-            observation_space=self.observation_space,
-            action_space=self.action_space,
-        )
+    def __init__(self, env):
+        super(TfEnv, self).__init__(env)
```
ryanjulian: Should not be necessary in Python 3

garage/tf/envs/base.py
```diff
@@ -32,29 +19,26 @@ def __call__(self, *args, **kwargs):
...
 
     @property
     def vectorized(self):
-        return getattr(self.wrapped_env, "vectorized", False)
+        return getattr(self._wrapped_env, "vectorized", False)
 
     def vec_env_executor(self, n_envs, max_path_length):
         return VecTfEnv(
-            self.wrapped_env.vec_env_executor(
+            self._wrapped_env.vec_env_executor(
```
ryanjulian: self.env should be populated by gym.Wrapper

garage/tf/envs/base.py
```diff
@@ -32,29 +19,26 @@ def __call__(self, *args, **kwargs):
...
 
-    @cached_property
-    def spec(self):
-        return EnvSpec(
-            observation_space=self.observation_space,
-            action_space=self.action_space,
-        )
+    def __init__(self, env):
+        super(TfEnv, self).__init__(env)
+        self.action_space = self._to_tf_space(self.action_space)
```
ryanjulian: `self.action_space` and `self.observation_space` should only ever take on one type. why not `self._to_garage_space(env.action_space)`

garage/tf/envs/base.py
```diff
@@ -32,29 +19,26 @@ def __call__(self, *args, **kwargs):
...
+        self.observation_space = self._to_tf_space(self.observation_space)
 
     @property
     def vectorized(self):
-        return getattr(self.wrapped_env, "vectorized", False)
+        return getattr(self._wrapped_env, "vectorized", False)
```
ryanjulian: Does and gym.Env have a vectorized attr?

garage/tf/envs/base.py
```diff
@@ -32,29 +19,26 @@ def __call__(self, *args, **kwargs):
...
 
     def vec_env_executor(self, n_envs, max_path_length):
```
ryanjulian: This might need some redesign

garage/envs/util.py
jonashen: Not yet no

garage/envs/base.py
jonashen: This will break existing functionality in the code

garage/envs/base.py
jonashen: I'll address the horizon issue in another PR. I'm not sure how much refactoring will be necessary.

ryanjulian: No environment seems to even define horizon.


```python
scripts/run_experiment.py:103:        help='Print only the tabular log information (in a horizontal format)')
garage/theano/envs/base.py:48:    def horizon(self):
garage/sampler/parallel_sampler.py:116:    :param max_path_length: horizon / maximum length of a single trajectory
garage/sampler/utils.py:164:        # self.horizon_terminals[self.top] = horizon_terminal
garage/sampler/utils.py:263:            # do not pick samples which terminated because of horizon
garage/sampler/utils.py:264:            # if np.any(self.horizon_terminals.take(initial_indices[0:-1],
garage/sampler/utils.py:265:            #    mode='wrap')) or self.horizon_terminals[end_index]:
garage/misc/tabulate.py:60:    """Return a segment of a horizontal line with optional colons which
garage/misc/tabulate.py:74:    """Return a horizontal line with optional colons to indicate column's
garage/misc/tabulate.py:855:    "Return a string which represents a horizontal line."
garage/tf/core/layers.py:1233:                 horizon=None,
garage/tf/core/layers.py:1241:        self.horizon = horizon
garage/tf/core/layers.py:1284:            if self.horizon is not None:
garage/tf/core/layers.py:1286:                for idx in range(self.horizon):
garage/tf/core/layers.py:1776:                 horizon=None,
garage/tf/core/layers.py:1790:        self.horizon = horizon
garage/tf/core/layers.py:1851:            if self.horizon is not None:
garage/tf/core/layers.py:1853:                for idx in range(self.horizon):
garage/tf/algos/batch_polopt.py:36:                 fixed_horizon=False,
garage/tf/algos/batch_polopt.py:81:        self.fixed_horizon = fixed_horizon
garage/envs/mujoco/randomization/randomized_env.py:77:    def horizon(self):
garage/envs/mujoco/randomization/randomized_env.py:78:        return self._wrapped_env.horizon
garage/envs/util.py:13:    'bounds', 'default_value', 'flat_dim', 'flatten', 'flatten_n', 'horizon',
garage/envs/util.py:74:def horizon(env):
garage/envs/proxy_env.py:36:    def horizon(self):
garage/envs/proxy_env.py:37:        return self._wrapped_env.horizon
garage/envs/box2d/mountain_car_env.py:17:    @autoargs.arg("goal_cart_pos", type=float, help="Goal horizontal position")
garage/algos/ddpg.py:56:                 include_horizon_terminal_transitions=False,
garage/algos/ddpg.py:90:        :param include_horizon_terminal_transitions: whether to include
garage/algos/ddpg.py:91:         transitions with terminal=True because the horizon was reached. This
garage/algos/ddpg.py:126:        self.include_horizon_terminal_transitions = \
garage/algos/ddpg.py:127:            include_horizon_terminal_transitions
garage/algos/ddpg.py:194:                    if self.include_horizon_terminal_transitions:
examples/theano/trpo_gym_pendulum.py:27:        max_path_length=env.horizon,
contrib/bichengcao/examples/trpo_gym_MountainCar-v0.py:7:from garage.envs.util import horizon, spec
contrib/bichengcao/examples/trpo_gym_MountainCar-v0.py:24:        max_path_length=horizon(env),
contrib/bichengcao/examples/trpo_gym_Pendulum-v0.py:6:from garage.envs.util import horizon, spec
contrib/bichengcao/examples/trpo_gym_Pendulum-v0.py:23:        max_path_length=horizon(env),
examples/theano/trpo_gym_cartpole.py:27:        max_path_length=env.horizon,
contrib/bichengcao/examples/trpo_gym_CartPole-v1.py:6:from garage.envs.util import horizon, spec
contrib/bichengcao/examples/trpo_gym_CartPole-v1.py:23:        max_path_length=horizon(env),
contrib/bichengcao/examples/trpo_gym_Acrobot-v1.py:6:from garage.envs.util import horizon, spec
contrib/bichengcao/examples/trpo_gym_Acrobot-v1.py:23:        max_path_length=horizon(env),
contrib/bichengcao/examples/trpo_gym_CartPole-v0.py:6:from garage.envs.util import horizon, spec
contrib/bichengcao/examples/trpo_gym_CartPole-v0.py:23:        max_path_length=horizon(env),
garage/algos/cma_es_lib.py:463:        # learning rates and back-ward time horizons
```

garage/envs/util.py
ryanjulian: What's still in use?

jonashen:
```
(garage) garage (replace_distributions) $ ag "from garage.envs.util import "
contrib/bichengcao/examples/trpo_gym_Acrobot-v1.py
5:from garage.envs.util import horizon, spec

contrib/bichengcao/examples/trpo_gym_CartPole-v0.py
5:from garage.envs.util import horizon, spec

contrib/bichengcao/examples/trpo_gym_Pendulum-v0.py
5:from garage.envs.util import horizon, spec

contrib/bichengcao/examples/trpo_gym_CartPole-v1.py
5:from garage.envs.util import horizon, spec

contrib/bichengcao/examples/trpo_gym_MountainCar-v0.py
6:from garage.envs.util import horizon, spec

contrib/ros/envs/sawyer/launchers/trpo_real_sawyer_push.py
6:from garage.envs.util import spec

contrib/ros/envs/sawyer/launchers/trpo_real_sawyer_pnp.py
6:from garage.envs.util import spec

contrib/ros/envs/sawyer/launchers/trpo_sim_sawyer_pnp.py
6:from garage.envs.util import spec

contrib/ros/envs/sawyer/launchers/trpo_real_sawyer_reacher.py
8:from garage.envs.util import spec

contrib/ros/envs/sawyer/launchers/trpo_sim_sawyer_push.py
6:from garage.envs.util import spec

contrib/ros/envs/sawyer/launchers/trpo_sim_sawyer_reacher.py
11:from garage.envs.util import spec

examples/tf/cluster_gym_mujoco_demo.py
7:from garage.envs.util import spec

examples/tf/trpo_gym_tf_cartpole.py
4:from garage.envs.util import spec

garage/envs/box2d/box2d_env.py
12:from garage.envs.util import bounds, flat_dim

garage/envs/box2d/car_parking_env.py
8:from garage.envs.util import flat_dim

garage/envs/mujoco/gather/gather_env.py
24:from garage.envs.util import flat_dim

garage/envs/mujoco/hill/terrain.py
14:from garage.envs.util import flat_dim

garage/envs/mujoco/maze/maze_env.py
17:from garage.envs.util import flat_dim

garage/envs/mujoco/mujoco_env.py
16:from garage.envs.util import bounds

garage/envs/noisy_env.py
6:from garage.envs.util import flat_dim

garage/envs/normalized_env.py
5:from garage.envs.util import flat_dim, flatten, unflatten

garage/envs/occlusion_env.py
9:from garage.envs.util import flat_dim

garage/tf/algos/ddpg.py
19:from garage.envs.util import bounds
20:from garage.envs.util import flat_dim

garage/tf/exploration_strategies/ou_strategy.py
11:from garage.envs.util import flat_dim

garage/tf/policies/continuous_mlp_policy.py
11:from garage.envs.util import flat_dim

garage/tf/q_functions/continuous_mlp_q_function.py
4:from garage.envs.util import flat_dim

garage/theano/policies/categorical_conv_policy.py
6:from garage.envs.util import flatten, flatten_n, weighted_sample

tests/envs/test_sawyer_envs.py
12:from garage.envs.util import spec

tests/test_normalized_gym.py
7:from garage.envs.util import flat_dim, flatten
```

ryanjulian: I thought this PR guarantees that everything in garage receives a garage.spaces instance.

If that's true, can't those all be replaced by some_space.flatten(foo), some_space.spec, some_space.flat_dim,  some_space.bounds, etc. thanks to this PR?

jonashen: Furthermore, we will need to refactor a lot of Envs to inherit from GarageEnv (e.g. PickAndPlaceEnv, ReacherEnv, PushEnv)

ryanjulian: We can touch whatever we want, we just don't make development a priority in contrib. There's no point in making more technical debt when it's easy to avoid (e.g. I could implement the commit using `sed`).

Why not have ReacherEnv, etc. inherit from gym.Env as now? I can't see a use case for inheriting from GarageEnv/TfEnv/TheanoEnv. All environments are still supposed to be gym.Envs.

garage/envs/base.py
jonashen: After removing `def Step(...)`, calling `Step()` in garage requires passing `info` as an explicit positional argument:
```
  File "/Users/jonathon/Documents/garage/garage/garage/envs/box2d/box2d_env.py", line 191, in step
    return Step(observation=next_obs, reward=reward, done=done)
TypeError: __new__() missing 1 required positional argument: 'info'
```
I believe it's better to leave it as-is and write a comment above `def Step` about its necessity.

ryanjulian: Ah I see. SGTM.

ryanjulian:
If this is getting too big you can submit it in (stable) pieces. Just stack some PRs branched one after the other.

garage/envs/base.py
ryanjulian: Warnings print annoying error messages--they don't crash the code.

garage/envs/box2d/box2d_env.py
```diff
@@ -124,13 +123,13 @@ def observation_space(self):
 
     @property
     def action_bounds(self):
-        return bounds(self.action_space)
+        return self.action_space.low, self.action_space.high
 
     def forward_dynamics(self, action):
-        if len(action) != flat_dim(self.action_space):
+        if len(action) != np.prod(self.action_space.low.shape):
```
ryanjulian: Does it break if we impose the stronger (and simpler) constraint: `action.shape == self.action_space.low.shape`?. There's not much point in having action spaces if they are not followed.

garage/envs/base.py
jonashen: It makes Travis fail, however:
```
======================================================================
ERROR: test_ddpg (algos.test_ddpg.TestDDPG)
  File "/root/code/garage/tests/algos/test_ddpg.py", line 31, in test_ddpg
    algo.train()
  File "/root/code/garage/garage/theano/algos/ddpg.py", line 218, in train
    self.evaluate(epoch, pool)
  File "/root/code/garage/garage/theano/algos/ddpg.py", line 379, in evaluate
    self.env.log_diagnostics(paths)
  File "/root/code/garage/garage/envs/base.py", line 88, in log_diagnostics
    raise DeprecationWarning
DeprecationWarning
Ran 54 tests in 783.391s
FAILED (errors=9)
```

garage/envs/noisy_env.py
```diff
@@ -22,7 +21,7 @@ def __init__(
         Serializable.quick_init(self, locals())
         super(NoisyObservationEnv, self).__init__(env)
         self.obs_noise = obs_noise
-        self._action_flat_dim = flat_dim(self.action_space)
+        self._action_flat_dim = self.action_space.low.shape
```
ryanjulian: this is a shape, not a flat_dim. `np.prod(self.action_space.low.shape)`

I might support keeping flat_dim just to stop people from getting this wrong.

garage/envs/normalized_env.py
```diff
@@ -25,7 +25,13 @@ def __init__(
         self._flatten_obs = flatten_obs
 
         self._obs_alpha = obs_alpha
-        flat_obs_dim = flat_dim(env.observation_space)
+
+        flat_obs_dim = None
+        if isinstance(env.observation_space, gym.spaces.Box):
```
ryanjulian: you will need flat_dim or np.prod here

garage/envs/normalized_env.py
```diff
@@ -34,7 +40,14 @@ def __init__(
         self._reward_var = 1.
 
     def _update_obs_estimate(self, obs):
-        flat_obs = flatten(self.env.observation_space, obs)
+        flat_obs = None
+        if isinstance(self.env.observation_space, gym.spaces.Box):
+            flat_obs = np.asarray(obs).flatten()
```
ryanjulian: ok fine you've convinced me that flat_dim, flatten{_n}, and unflatten{_n} are warranted.

garage/envs/occlusion_env.py
```diff
@@ -30,7 +29,7 @@ def __init__(self, env, sensor_idx):
             self._dt = env.sim.opt.timestep * env.frame_skip
 
     def _set_sensor_mask(self, env, sensor_idx):
-        obsdim = flat_dim(env.observation_space)
+        obsdim = env.observation_space.low.shape
```
ryanjulian: flat_dim != .shape

garage/tf/envs/base.py
```diff
@@ -1,68 +1,54 @@
...
-    def spec(self):
-        return EnvSpec(
-            observation_space=self.observation_space,
-            action_space=self.action_space,
-        )
-
-    @property
-    def vectorized(self):
-        return getattr(self.wrapped_env, "vectorized", False)
+    def reset(self, **kwargs):
```
ryanjulian: implemented by the superclass

garage/tf/envs/base.py
```diff
@@ -1,68 +1,54 @@
...
+        return self.env.reset(**kwargs)
 
-    def vec_env_executor(self, n_envs, max_path_length):
-        return VecTfEnv(
-            self.wrapped_env.vec_env_executor(
-                n_envs=n_envs, max_path_length=max_path_length))
+    def step(self, action):
```
ryanjulian: Implemented by the superclass

garage/tf/envs/base.py
```diff
@@ -1,68 +1,54 @@
...
+        return self.env.step(action)
 
     @classmethod
     def wrap(cls, env_cls, **extra_kwargs):
         # Use a class wrapper rather than a lambda method for smoother
         # serialization
         return WrappedCls(cls, env_cls, extra_kwargs)
 
+    def _to_garage_space(self, space):
```
ryanjulian: @overrides

garage/tf/envs/base.py
```diff
@@ -80,3 +66,14 @@ def step(self, action_n):
 
     def close(self):
         self.vec_env.close()
+
+
+class WrappedCls(object):
```
ryanjulian: Use new-style classes (no (object) needed)

garage/tf/envs/base.py
```diff
@@ -1,68 +1,54 @@
...
+            return Box(low=space.low, high=space.high)
+        elif isinstance(space, GymDiscrete):
+            return Discrete(space.n)
+        elif isinstance(space, GymTuple):
+            return Product(list(map(self._to_tf_space, space.spaces)))
+        else:
+            raise NotImplementedError
+
 
 class VecTfEnv(object):
```
ryanjulian: Use new-style classes (no (object) needed)

garage/algos/cem.py
```diff
@@ -171,8 +171,6 @@ def train(self):
                 np.mean([len(path['returns']) for path in paths]))
 
             self.policy.set_param_values(best_x)
-            self.env.log_diagnostics(paths)
-            self.policy.log_diagnostics(paths)
```
ryanjulian: uh policies still have log_diagnostics

garage/algos/cma_es.py
```diff
@@ -128,8 +128,6 @@ def train(self):
             logger.record_tabular(
                 'AvgTrajLen',
                 np.mean([len(info['returns']) for info in infos]))
-            self.env.log_diagnostics(infos)
-            self.policy.log_diagnostics(infos)
```
ryanjulian: don't remove log_diagnostics for policies

garage/contrib/bichengcao/examples/trpo_gym_Acrobot-v1.py
```diff
@@ -2,25 +2,25 @@
...
-    baseline = LinearFeatureBaseline(env_spec=spec(env))
+    baseline = LinearFeatureBaseline(env_spec=env.spec)
 
     algo = TRPO(
         env=env,
         policy=policy,
         baseline=baseline,
         batch_size=4000,
-        max_path_length=horizon(env),
+        max_path_length=env.horizon,
```
ryanjulian: since, in this example, we know that this env has a spec (from gym), you should instead call env.max_episode_steps

https://github.com/openai/gym/blob/master/gym/envs/registration.py#L64

garage/contrib/bichengcao/examples/trpo_gym_CartPole-v0.py
```diff
@@ -2,25 +2,25 @@
...
-    baseline = LinearFeatureBaseline(env_spec=spec(env))
+    baseline = LinearFeatureBaseline(env_spec=env.spec)
 
     algo = TRPO(
         env=env,
         policy=policy,
         baseline=baseline,
         batch_size=4000,
-        max_path_length=horizon(env),
+        max_path_length=env.horizon,
```
ryanjulian: https://github.com/openai/gym/blob/master/gym/envs/registration.py#L64

garage/contrib/bichengcao/examples/trpo_gym_CartPole-v1.py
```diff
@@ -2,25 +2,25 @@
...
-    baseline = LinearFeatureBaseline(env_spec=spec(env))
+    baseline = LinearFeatureBaseline(env_spec=env.spec)
 
     algo = TRPO(
         env=env,
         policy=policy,
         baseline=baseline,
         batch_size=4000,
-        max_path_length=horizon(env),
+        max_path_length=env.horizon,
```
ryanjulian: https://github.com/openai/gym/blob/master/gym/envs/registration.py#L64

garage/contrib/bichengcao/examples/trpo_gym_MountainCar-v0.py
```diff
@@ -3,25 +3,25 @@
...
-    baseline = LinearFeatureBaseline(env_spec=spec(env))
+    baseline = LinearFeatureBaseline(env_spec=env.spec)
 
     algo = TRPO(
         env=env,
         policy=policy,
         baseline=baseline,
         batch_size=4000,
-        max_path_length=horizon(env),
+        max_path_length=env.horizon,
```
ryanjulian: https://github.com/openai/gym/blob/master/gym/envs/registration.py#L64

garage/contrib/bichengcao/examples/trpo_gym_Pendulum-v0.py
```diff
@@ -2,25 +2,25 @@
...
-    baseline = LinearFeatureBaseline(env_spec=spec(env))
+    baseline = LinearFeatureBaseline(env_spec=env.spec)
 
     algo = TRPO(
         env=env,
         policy=policy,
         baseline=baseline,
         batch_size=4000,
-        max_path_length=horizon(env),
+        max_path_length=env.horizon,
```
ryanjulian: https://github.com/openai/gym/blob/master/gym/envs/registration.py#L64

garage/envs/base.py
ryanjulian: That's because you raised an exception, not a warning. In python you can raise anything as an exception, including the DeprecationWarning class.

https://pymotw.com/3/warnings/

garage/envs/mujoco/gather/gather_env.py
```diff
@@ -467,7 +466,7 @@ def log_diagnostics(self, paths, log_prefix='Gather', *args, **kwargs):
                 stripped_path[k] = v
             stripped_path['observations'] = \
                 stripped_path['observations'][
-                    :, :flat_dim(self.wrapped_env.observation_space)]
+                    :, :self.wrapped_env.observation_space.low.shape]
```
ryanjulian: .shape != flat_dim

garage/envs/mujoco/hill/terrain.py
```diff
@@ -51,10 +50,10 @@ def generate_hills(width, height, nhills):
 
 
 def clear_patch(hfield, box):
-    ''' Clears a patch shaped like box, assuming robot is placed in center of hfield
+    '''Clears a patch shaped like box, assuming robot is placed in center of hfield
     @param box: garage.spaces.Box-like
     '''
-    if flat_dim(box) > 2:
+    if box.low.shape > 2:
```
ryanjulian: .shape != flat_dim

garage/envs/mujoco/maze/maze_env.py
```diff
@@ -266,6 +265,7 @@ def _find_robot(self):
                 if structure[i][j] == 'r':
                     return j * size_scaling, i * size_scaling
         assert False
```
ryanjulian: this will always crash.

garage/envs/mujoco/maze/maze_env.py
```diff
@@ -349,8 +350,9 @@ def log_diagnostics(self, paths, *args, **kwargs):
             stripped_path = {}
             for k, v in path.items():
                 stripped_path[k] = v
-            stripped_path['observations'] = stripped_path['observations'][:,
-             :flat_dim(self.wrapped_env.observation_space)] # yapf: disable
+            stripped_path['observations'] = stripped_path[
+                'observations'][:, :
```
ryanjulian: flat_dim != .shape

garage/theano/algos/ddpg.py
```diff
@@ -376,9 +376,6 @@ def evaluate(self, epoch, pool):
         logger.record_tabular('PolicyRegParamNorm', policy_reg_param_norm)
         logger.record_tabular('QFunRegParamNorm', qfun_reg_param_norm)
 
-        self.env.log_diagnostics(paths)
-        self.policy.log_diagnostics(paths)
```
ryanjulian: don't remove calls to policy.log_diagnostics()

garage/theano/envs/base.py
```diff
@@ -1,51 +1,45 @@
 """Wrapper class that converts gym.Env into TheanoEnv."""
-from cached_property import cached_property
```
ryanjulian: uhhh the properties were cached for a reason. please keep them cached.

garage/envs/box2d/car_parking_env.py
```diff
@@ -120,9 +119,9 @@ def action_from_keys(self, keys):
 
     @overrides
     def forward_dynamics(self, action):
-        if len(action) != flat_dim(self.action_space) + 1:
+        if len(action) != self.action_space.low.shape + 1:
```
ryanjulian: action_space.low.shape != flat_dim(action_space)

garage/envs/noisy_env.py
```diff
@@ -63,7 +62,7 @@ def __init__(
         Serializable.quick_init(self, locals())
         super(DelayedActionEnv, self).__init__(env)
         self.action_delay = action_delay
-        self._action_flat_dim = flat_dim(self.action_space)
+        self._action_flat_dim = self.action_space.low.shape
```
ryanjulian: flat_dim != .shape

tests/test_normalized_gym.py
```diff
@@ -39,9 +35,8 @@ def test_unflatten(self):
             for e in range(5):
                 action = env.action_space.sample()
                 next_obs, reward, done, info = env.step(action)
-                assert flatten(env.observation_space,
-                               next_obs).shape == flat_dim(
-                                   env.observation_space)
+                assert np.asarray(next_obs).flatten(
```
ryanjulian: This is a logical error.

a.flatten().shape == b.shape iff a is already flat.

garage/contrib/bichengcao/examples/trpo_gym_Acrobot-v1.py
jonashen: `horizon` will already call `env.spec.max_episode_steps` for the underlying env. This is because we wrapped env with TheanoEnv:
```
  File "garage/contrib/bichengcao/examples/trpo_gym_Acrobot-v1.py", line 23, in run_task
    max_path_length=env.max_episode_steps,
AttributeError: 'TheanoEnv' object has no attribute 'max_episode_steps'
```

ryanjulian:
```python
if hasattr(env, "max_episode_steps") and not hasattr(TheanoEnv(env), "max_episode_steps"):
    raise SomethingIsWrongWithTheanoEnvException
```

garage/envs/base.py
jonashen: i see. thanks for the heads up

garage/envs/box2d/car_parking_env.py
jonashen: I think needs `len()` because we're checking if the dimensions align but not the values themselves

jonashen: i.e. `if len(action.shape) != len(self.action_space.low.shape):`

garage/envs/mujoco/maze/maze_env.py
jonashen: Remove `assert False` and add `return None, None` instead?

garage/envs/noisy_env.py
jonashen: I'm not sure how I missed np.prod(). My b

garage/envs/normalized_env.py
jonashen: I can make them util functions for normalized_env only, as I've removed all other usages of `util.py` in garage.

garage/tf/envs/base.py
jonashen:
```
  File "/Users/jonathon/Documents/garage/garage/garage/tf/envs/base.py", line 31, in TfEnv
    @overrides
TypeError: 'module' object is not callable
```

garage/theano/envs/base.py
jonashen: How do I cache variables like `self.observation_space`, `self.action_space`, rather than functions (as it's used here)

tests/test_normalized_gym.py
jonashen: would would be the correct logic here? a.shape == b.shape?

zhanpenghe: This should be something like this:
`observation_space.flatten(next_obs).shape[0] == observation.flat_dim`
Of course the this next_obs should not be a batch of observation.

garage/tf/envs/base.py
zhanpenghe: Is this env necessary now? I think we can just use `TfEnv(VecEnv)` now.

garage/envs/mujoco/mujoco_env.py
```diff
@@ -102,7 +101,7 @@ def observation_space(self):
 
     @property
     def action_bounds(self):
-        return bounds(self.action_space)
+        return self.action_space.low, self.action_space.high
```
zhanpenghe: Some spaces does not have a high and a low. e.g. gym.Discrete

garage/tf/envs/base.py
zhanpenghe: Can you show me how did you import it? It might just be a small error of importing.

tests/test_dm_control.py
```diff
@@ -22,10 +21,7 @@ def run_task(domain_name, task_name):
     time_step = dm_control_env.reset()
     action_spec = dm_control_env.action_space
     for _ in range(5):
-        if 'CI' in os.environ:
-            print("Skipping rendering test")
-        else:
-            dm_control_env.render()
+        dm_control_env.render()
```
zhanpenghe: Why deleting the CI condition?

jonashen: We can now render on the CI using a fake X display inside Docker

garage/envs/box2d/box2d_env.py
```diff
@@ -124,13 +123,13 @@ def observation_space(self):
...
-        return bounds(self.action_space)
+        return self.action_space.low, self.action_space.high
 
     def forward_dynamics(self, action):
-        if len(action) != flat_dim(self.action_space):
+        if action.shape != self.action_space.low.shape:
             raise ValueError(
                 'incorrect action dimension: expected %d but got '
-                '%d' % (flat_dim(self.action_space)), len(action))
+                '%d' % (np.prod(self.action_space.low.shape), len(action)))
```
ryanjulian: I think you forgot to update the error message here.

garage/envs/box2d/box2d_env.py
```diff
@@ -124,13 +123,13 @@ def observation_space(self):
 
     @property
     def action_bounds(self):
-        return bounds(self.action_space)
+        return self.action_space.low, self.action_space.high
 
     def forward_dynamics(self, action):
-        if len(action) != flat_dim(self.action_space):
+        if action.shape != self.action_space.low.shape:
```
ryanjulian: What test did you use to make sure it still works with the stronger check?

garage/envs/mujoco/mujoco_env.py
ryanjulian: Yeah not sure what to do about this. MuJoCo is continuous control, so discrete makes little sense. But conceivably someone could want to input a tuple space.

I think it's okay to leave the Box assumption as long as we assert it somewhere and nothing is using non-boxes yet. If someone hits the assert they can extend to support non-Box spaces.

garage/envs/box2d/box2d_env.py
jonashen: I tested with all Box2D envs, i.e. CartpoleEnv, CarParkingEnv, DoublePendulumEnv, CartpoleSwingupEnv, and MountainCarEnv

tests/test_normalized_gym.py
```diff
@@ -39,9 +34,8 @@ def test_unflatten(self):
             for e in range(5):
                 action = env.action_space.sample()
                 next_obs, reward, done, info = env.step(action)
-                assert flatten(env.observation_space,
-                               next_obs).shape == flat_dim(
-                                   env.observation_space)
+                assert env.observation_space.flatten(next_obs).shape[
+                    0] == env.observation_space.flat_dim
```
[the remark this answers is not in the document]
jonashen: done

garage/envs/mujoco/mujoco_env.py
jonashen: Or since we're not removing the `util.py` file, I can continue using `bounds`.

jonashen: I'll leave the code as-is for now and add some assert statements instead.

garage/theano/envs/base.py
ryanjulian: You make them properties

garage/envs/mujoco/mujoco_env.py
ryanjulian: Why do that if the definition of bounds() is just this code + NotImplementedError for non-Box spaces? You can already check that the action space is a Box.

examples/theano/trpo_cartpole.py
```diff
@@ -6,11 +6,10 @@
 from garage.theano.policies import GaussianMLPPolicy
 
 env = TheanoEnv(normalize(CartpoleEnv()))
-
+print(env.action_space)
```
ryanjulian: oops

jonashen: oops

garage/contrib/bichengcao/examples/trpo_gym_CartPole-v0.py
ryanjulian: max_episode_steps?

garage/contrib/bichengcao/examples/trpo_gym_CartPole-v1.py
ryanjulian: max_episode_steps?

garage/contrib/bichengcao/examples/trpo_gym_MountainCar-v0.py
ryanjulian: max_episode_steps?

garage/contrib/bichengcao/examples/trpo_gym_Pendulum-v0.py
ryanjulian: max_episode_steps?

garage/envs/base.py
ryanjulian: 👍

garage/envs/base.py
ryanjulian: 👍

garage/envs/mujoco/gather/gather_env.py
```diff
@@ -467,7 +466,7 @@ def log_diagnostics(self, paths, log_prefix='Gather', *args, **kwargs):
                 stripped_path[k] = v
             stripped_path['observations'] = \
                 stripped_path['observations'][
-                    :, :flat_dim(self.wrapped_env.observation_space)]
+                    :, :np.prod(self.wrapped_env.observation_space.low.shape)]
```
ryanjulian: ok you got me--i'm tired of reading np.prod(...). Can you put flat_dim back in?

garage/envs/mujoco/maze/maze_env.py
```diff
@@ -282,6 +282,7 @@ def _find_goal_range(self):  # this only finds one goal!
                     maxy = i * size_scaling + size_scaling \
                            * 0.5 - self._init_torso_y
                     return minx, maxx, miny, maxy
+        return None
```
ryanjulian: Hmm an exception is probably in order here before you actually return (but I'm assuming pylint wants you to return something, which is fine).

garage/envs/base.py
```diff
@@ -1,17 +1,138 @@
...
     """
-    Convenience method creating a namedtuple with the results of the
-    environment.step method.
-    Put extra diagnostic info in the kwargs
+
+    def __init__(self, env):
+        Serializable.quick_init(self, locals())
+        Parameterized.__init__(self)
+        super().__init__(env)
+        self.action_space = self._to_garage_space(env.action_space)
```
ryanjulian: Aren't these properties in TfEnv/TheanoEnv? Why set a member var here?

It seems better not to set them there--then if the property doesn't exists (i.e. someone constructed a GarageEnv outside of a TfEnv/TheanoEnv) we will get an error.

garage/envs/base.py
```diff
@@ -1,17 +1,138 @@
...
+from garage.core import Serializable
+from garage.envs.env_spec import EnvSpec
+from garage.spaces import Box
+from garage.spaces import Discrete
+from garage.spaces import Product
 
-_Step = collections.namedtuple("Step",
-                               ["observation", "reward", "done", "info"])
 
+class GarageEnv(gym.Wrapper, Parameterized, Serializable):
```
ryanjulian: Using an abstract base class might help you debug this

https://pymotw.com/3/abc/

garage/envs/base.py
ryanjulian: http://sayspy.blogspot.com/2009/12/how-to-handle-multiple-inheritance-of.html

garage/envs/mujoco/maze/maze_env.py
ryanjulian: oh i see. it is a way of throwing an error if nothing was found.

replace it with an exception and we're good.

garage/envs/mujoco/gather/gather_env.py
jonashen: 👍

garage/envs/mujoco/maze/maze_env.py
jonashen: 👍

garage/envs/mujoco/maze/maze_env.py
jonashen: 👍

garage/envs/base.py
jonashen: The issue here is wherever I call `self.action_space` (e.g. in `spec()`), it defaults to gym.spaces.Box, even if I have @overrides above `def action_space(...)` in TheanoEnv/TfEnv.

jonashen: As in even if the property `def action_space` does exist, calling `self.action_space` will just default to the underlying env.action_space (which is a gym space), instead of returning the property method.

garage/envs/base.py
```diff
@@ -1,17 +1,121 @@
...
+        Returns an empty list if env.get_params() is called.
+
+        Returns:
+            An empty list
+        """
+        warnings.warn("get_params_internal is deprecated", DeprecationWarning)
+        return []
+
+    @property
+    def horizon(self):
```
ryanjulian: Please remove this in #236

jonashen:
This issue is currently open here: https://github.com/rlworkgroup/garage/issues/239

jonashen:
How did you fix the (executed but not training) issue?

ryanjulian:
```python
import gym

from garage.baselines import LinearFeatureBaseline
from garage.misc.instrument import run_experiment
from garage.misc.instrument import stub
from garage.tf.policies import GaussianMLPPolicy

# Need to wrap in a tf environment and force_reset to true
# see https://github.com/openai/rllab/issues/87#issuecomment-282519288
env = TfEnv(gym.make("CartPole-v0"))

policy = GaussianMLPPolicy(
    name="policy", env_spec=env.spec, hidden_sizes=(32, 32))

algo = TRPO(
    env=env,
    policy=policy,
    baseline=baseline,
    batch_size=4000,
    max_path_length=200,
    n_itr=120,
    discount=0.99,
    step_size=0.01,
)

algo.train()
```
```sh
(garage) rjulian@tars:~/code/garage$ python examples/tf/trpo_gym_tf_cartpole.py 
WARN: gym.spaces.Box autodetected dtype as <class 'numpy.float32'>. Please provide explicit dtype.
  File "examples/tf/trpo_gym_tf_cartpole.py", line 15, in <module>
    name="policy", env_spec=env.spec, hidden_sizes=(32, 32))
  File "/home/rjulian/code/garage/garage/tf/policies/gaussian_mlp_policy.py", line 57, in __init__
    assert isinstance(env_spec.action_space, Box)
AssertionError
```

garage/envs/base.py
```diff
@@ -1,17 +1,121 @@
...
+        """
+        raise NotImplementedError
+
+    def reset(self, **kwargs):
+        """
+        Inheriting gym.Wrapper requires implementing this method.
+
+        Calls reset on wrapped env.
+        """
+        return self.env.reset(**kwargs)
```
ryanjulian: most of these should be provided by gym.Wrapper. Are you sure the gym.Wrapper constructor is being called?

garage/envs/base.py
```diff
@@ -1,17 +1,121 @@
...
+import gym
 
-from garage.envs import EnvSpec
+from garage.core import Parameterized
+from garage.core import Serializable
 
-_Step = collections.namedtuple("Step",
-                               ["observation", "reward", "done", "info"])
 
+class GarageEnv(ABC, gym.Wrapper, Parameterized, Serializable):
```
ryanjulian: don't inherit from ABC first--that means super() will call the ABC constructor.

most people following this pattern use the metaclass param instead to avoid this problem

```python
from abc import ABCMeta

class MyABC(metaclass=ABCMeta):
    pass
```

ryanjulian:
```python
class GarageEnv(gym.Wrapper, Parameterized, Serializable, metaclass=ABCMeta):
 ...

**Comment by ryanjulian** [comment]
The title of this PR is "Add pickled mode support to GarageEnv/TfEnv/TheanoEnv" yet I cannot pickle an environment until #239 is solved. It's not a problem with the launcher, it's the pickling.

File: garage/envs/base.py (hunk 60 above)
Comment (jonashen) [reply garage/envs/base.py]: Yes, gym.Wrapper requires all wrappers derived from gym.Wrapper to implement step() and reset(), or else it throws a deprecated warning:
```
    def step(self, action):
        if hasattr(self, "_step"):
            deprecated_warn_once("%s doesn't implement 'step' method, but it implements deprecated '_step' method." % type(self))
            self.step = self._step
            return self.step(action)
        else:
            deprecated_warn_once("%s doesn't implement 'step' method, " % type(self) +
                "which is required for wrappers derived directly from Wrapper. Deprecated default implementation is used.")
            return self.env.step(action)

    def reset(self, **kwargs):
        if hasattr(self, "_reset"):
            deprecated_warn_once("%s doesn't implement 'reset' method, but it implements deprecated '_reset' method." % type(self))
            self.reset = self._reset
            return self._reset(**kwargs)
        else:
            deprecated_warn_once("%s doesn't implement 'reset' method, " % type(self) +
                "which is required for wrappers derived directly from Wrapper. Deprecated default implementation is used.")
            return self.env.reset(**kwargs)
```

ryanjulian:
Can you test all the example scripts before submitting this? If you hit a known bug which already has an issue, you can of course ignore it.

Issue #214, opened 2018-07-24
Stop using a private fork of lasagne in environment.yml


ryanjulian:
Deps should either be from the source or first-party forks.

PR #241, opened 2018-07-30
Use Lasagne directly instead of a private fork

jonashen:
This PR fixes a tuple() to int() comparison bug that resulted in `tests.test_networks` and `tests.test_algos` failing. Furthermore, this bug fix enables us to install Lasagne directly rather than a private fork. 

Note: if we want to use the latest version of Lasagne we will need to clone Lasagne into `rlworkgroup/Lasagne` and address [this](https://github.com/Lasagne/Lasagne/issues/867) issue using a new commit. However, this is beyond the scope of this PR.

Ref: #182, #184, #214

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -29,7 +29,7 @@ dependencies:
         - jsonmerge
         - joblib==0.10.3
         - jupyter
-        - git+https://github.com/jonashen/Lasagne.git@b4fd25809c446f8c7b1c0d80283eb25c66e06960#egg=lasagne
+        - git+https://github.com/Lasagne/Lasagne.git@a61b76fd991f84c50acdb7bea02118899b5fefe1#egg=lasagne
         - mako
         - matplotlib
         - memory_profiler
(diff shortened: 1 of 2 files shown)


ryanjulian:
If we merge this shouldn't we be able to just clone directly from https://github.com/Lasagne/Lasagne

jonashen:
Again, like I mentioned there is a compatibility issue with the most updated version of Lasagne. It requires Theano's version to be [0.8.2](https://github.com/Lasagne/Lasagne/blob/master/requirements.txt), but we are currently using `1.0.2`. That is why we will have to fix this [issue](https://github.com/Lasagne/Lasagne/issues/867) in a private Lasagne fork.

ryanjulian:
How many usages are there of the deprecated API?

jonashen:
The deprecated API is used within Lasagne itself inside `lasagne/layers/pool.py` when it tries to `from theano.tensor.signal import downsample`. So basically, wherever Lasagne is used this error will occur.

ryanjulian:
Does it work if we use bleeding-edge lasagne, as the author suggests?

ryanjulian:
It's fixed in lasagne/master. Let's just pin to `a61b76fd991f84c50acdb7bea02118899b5fefe1`

jonashen:
got it

ryanjulian:
Does this close any issues?

jonashen:
#182, #184 were already closed, just closed #214

PR #253, opened 2018-08-01
Replace ProxyEnv with gym.Wrapper

jonashen:
Ref: #236


ryanjulian:
Looks like this depends on #240?

jonashen:
Yes, waiting on the refactor of GarageEnv.

PR #254, opened 2018-08-01
Set TRPO and PPO benchmark tests big to True

CatherineSue:


ryanjulian:
Let me know if you want

PR #258, opened 2018-08-08
Update setup_linux.sh based on Dockerfile.ci

gonzaiva:
The dependencies used in docker to build garage in Travis CI have been
added to the script used to setup garage in Linux.
Also, with the help of Argbash the following arguments have been added
to the script:
- --tf-gpu: installs the GPU flavor of TensorFlow instead of the vanilla
flavor.
- --set-envvar: it writes the file ~/.bashrc to set the environment
variables required by garage. Otherwise, the user is notified of the
required environment variables to modify.
- mjkey-path: the path to the MuJoCo license key.

The current script has been tested in a fresh installation of Ubuntu,
running the files under the examples folder (both Theano and TensorFlow)
right after to make sure this setup file is working.


scripts/setup_linux.sh
```diff
@@ -1,61 +1,254 @@
...
+  ffmpeg \
+  libjpeg-dev \
+  swig \
+  libsdl2-dev
+# OpenAI baselines
+sudo apt-get install -y \
+  libopenmpi-dev \
+  openmpi-bin
+sudo apt-get clean
+sudo rm -rf /var/lib/apt/lists/*
```
ryanjulian: do not do this on a non-Docker system. it will delete the user's apt cache.

scripts/setup_linux.sh
```diff
@@ -1,61 +1,254 @@
...
+
+# Build GLFW because the Ubuntu 16.04 version is too old
+# See https://github.com/glfw/glfw/issues/1004
+apt-get purge -y -v libglfw*
```
ryanjulian: don't make commands verbose unless they need to be

scripts/setup_linux.sh
```diff
@@ -1,61 +1,254 @@
...
+git clone https://github.com/glfw/glfw.git
```
ryanjulian: always work in a temp directory if you're on a user's system (see `man mktemp`)

scripts/setup_linux.sh
```diff
@@ -1,61 +1,254 @@
...
+rm -rf glfw
+
+# Set up MuJoCo
+if [[ ! -f "${HOME}"/.mujoco/ ]]; then
+  mkdir "${HOME}"/.mujoco
+  wget https://www.roboti.us/download/mjpro150_linux.zip -O mujoco.zip
+  unzip mujoco.zip -d "${HOME}"/.mujoco
+  rm mujoco.zip
+else
+  printf '\e[0;31m%s\e[0m\n' "Mujoco is already installed!"
```
ryanjulian: If you want to print in color, define a helper function to do that.

scripts/setup_linux.sh
```diff
@@ -1,61 +1,254 @@
...
+# Set up conda
+hash conda 2>/dev/null || {
+  wget https://repo.continuum.io/miniconda/Miniconda2-latest-Linux-x86_64.sh \
+    -O miniconda.sh
+  sudo bash miniconda.sh -b -p /opt/conda
+  rm miniconda.sh
+  sudo ln -s /opt/conda/etc/profile.d/conda.sh /etc/profile.d/conda.sh
+}
+# Add conda to executable programs
+export PATH="$PATH:/opt/conda/bin"
```
ryanjulian: It's not appropriate to put conda into /opt/conda for non-docker users. conda should go in the normal install location.

scripts/setup_linux.sh
```diff
@@ -1,61 +1,254 @@
...
+
+# File where environment variables are stored
+BASH_RC="${HOME}/.bashrc"
+
+# Install dependencies
+printf "Installing garage dependencies\n"
+printf "You will probably be asked for your sudo password\n"
+sudo apt-get -y -q update
+# To install garage
+sudo apt-get install -y \
```
ryanjulian: * Install all the apt dependencies in a single command, which is faster and risks fewer conflicts
* User `apt` instead of `apt-get` (which is the new default and has more sophisticated conflict resolution)

scripts/setup_linux.sh
```diff
@@ -1,61 +1,254 @@
...
+  -DGLFW_BUILD_TESTS=OFF -DGLFW_BUILD_DOCS=OFF ..
+make -j"$(nproc)"
+make install
+cd ../../
+rm -rf glfw
+
+# Set up MuJoCo
+if [[ ! -f "${HOME}"/.mujoco/ ]]; then
+  mkdir "${HOME}"/.mujoco
+  wget https://www.roboti.us/download/mjpro150_linux.zip -O mujoco.zip
```
ryanjulian: on a user system, always work in a temp directory

scripts/setup_linux.sh
```diff
@@ -1,61 +1,254 @@
...
+pre-commit install -t pre-push
+pre-commit install -t commit-msg
+source deactivate
+
+# Add garage to python modules
+export PYTHONPATH="$PYTHONPATH:${GARAGE_DIR}"
+PYTHON_ENV_VAR="PYTHONPATH=\"\$PYTHONPATH:${GARAGE_DIR}\""
+if [[ "${_arg_set_envvar}" = on ]]; then
+  printf "\\n# Added by the garage installer\\n" >> "${BASH_RC}"
+  printf "export ${PYTHON_ENV_VAR}\\n" >> "${BASH_RC}"
```
ryanjulian: It looks like you are bending over backwards with a lot of escapes in these. You can usually use `echo -e` to make things simpler

scripts/setup_linux.sh
```diff
@@ -1,61 +1,254 @@
...
+pre-commit install -t pre-commit
+pre-commit install -t pre-push
+pre-commit install -t commit-msg
+source deactivate
+
+# Add garage to python modules
+export PYTHONPATH="$PYTHONPATH:${GARAGE_DIR}"
+PYTHON_ENV_VAR="PYTHONPATH=\"\$PYTHONPATH:${GARAGE_DIR}\""
+if [[ "${_arg_set_envvar}" = on ]]; then
+  printf "\\n# Added by the garage installer\\n" >> "${BASH_RC}"
```
ryanjulian: Please create *only one* section which says "Added by the garage installer"

scripts/setup_linux.sh
jonashen: MuJoCo

scripts/setup_linux.sh
```diff
@@ -1,61 +1,254 @@
...
+  printf "export ${LD_LIB_ENV_VAR}\\n" >> "${BASH_RC}"
 fi
 
-echo "Creating conda environment..."
+# Set up conda
+hash conda 2>/dev/null || {
+  wget https://repo.continuum.io/miniconda/Miniconda2-latest-Linux-x86_64.sh \
+    -O miniconda.sh
+  sudo bash miniconda.sh -b -p /opt/conda
```
ryanjulian: Allow conda to put this in the default location for a user install, rather than /opt.

scripts/setup_linux.sh
```diff
@@ -1,61 +1,254 @@
-#!/bin/bash
-# Make sure that conda is available
+#!/usr/bin/env bash
+# This script installs garage on Linux distributions
```
ryanjulian: Please add a note to maintainers to update the dockerfiles in docker/ if they change this file (and vice-versa)

scripts/setup_linux.sh
```diff
@@ -114,98 +128,91 @@ cd "${GARAGE_DIR}"
...
 conda env create -f environment.yml
 if [[ "${?}" -ne 0 ]]; then
-  printf '\e[0;31m%s\e[0m\n' "Error: conda environment could not be created"
-  exit 1
+  print_error "Error: conda environment could not be created"
 fi
-sudo rm -rf /opt/conda/pkgs/*
 
 # Extras
 source activate garage
```
ryanjulian: Specify a tensorflow version. Right now 1.9 works but 1.10 has some warnings we need to address

gonzaiva:
All the requested changes have been applied. Also, if the script runs twice, the garage installation does not break:
- Apt detects when the packages are already installed, so this section won't break.
- When GLFW is already installed, make install will fail but the script will continue.
- If MuJoCo is already unzipped in ~/.mujoco, this part is skipped.
- If conda is already installed, the installer is not ran again.
- If the conda environment for garage is already installed, an error is printed, but the script continues.
- The section of extras runs all the time to fix Box2D and install the desired flavor of TF.

PR #260, opened 2018-08-08
Add ros docker

hjzh4:
- Use this to get a out-of-box ros/gazebo environment.


ryanjulian:
* Please make sure all the scripts conform to the Google Shell Style Guide
* Please remove references/paths/switches which make the user specify sawyer. We only support one robot right now so it does nothing, plus it's not clear if we supported more than one robot that we would make them the user specify which robot rather than maintaining a single config for all robots.
* The images should be tagged under the rlworkgroup org. CPU images are normal and GPU images are "special" so the tags would be rlworkgroup/garage-contrib-ros:latest and rlworkgroup/garage-contrib-ros-gpu:latest
* Please file an issue to refactor these to be based on shared garage and garage-gpu Dockerfiles later.

garage/contrib/ros/docker/build.sh
```diff
@@ -0,0 +1,19 @@
+#! /bin/bash
```
ryanjulian: Please remove this script, since it doesn't do anything more than a regular Docker command.

garage/contrib/ros/docker/run_gpu.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/sh
```
ryanjulian: Please place scripts intended to be run by the user in scripts/

garage/contrib/ros/docker/run.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/sh
```
ryanjulian: Please place scripts intended to be run by the user in scripts/

garage/contrib/ros/docker/sawyer/Dockerfile.cpu
```diff
@@ -0,0 +1,177 @@
+FROM ubuntu:16.04
```
ryanjulian: Please rename this Dockerfile.contrib-ros and put it in docker/

ryanjulian: Please base Dockerfiles intended to be run interactively on [tini](https://hub.docker.com/r/krallin/ubuntu-tini/)

garage/contrib/ros/docker/sawyer/Dockerfile.gpu
```diff
@@ -0,0 +1,235 @@
+ARG DOCKER_FROM=ubuntu:16.04
+
```
ryanjulian: This file is a mess of copy/pastes. Please consolidate it.

Please base Dockerfiles meant to be run interactively on [tini](https://hub.docker.com/r/krallin/ubuntu-tini/)

Please move this file into docker/ and make it Dockerfile.contrib-ros-gpu

garage/contrib/ros/docker/sawyer/docker-entrypoint.sh
```diff
@@ -0,0 +1,253 @@
+#!/bin/bash
+
+ORANGE='\033[0;33m'
```
ryanjulian: Please do not install anything in an entrypoint. That's what Dockerfiles are for. You're not even guaranteed internet access when running the Dockerfile.

garage/contrib/ros/docker/sawyer/docker-entrypoint.sh
```diff
@@ -0,0 +1,253 @@
...
+echo "export USER=sawyer_docker" > "/home/$USER/.bashrc"
+
+echo "source /opt/ros/kinetic/setup.bash" >> "/home/$USER/.bashrc"
+source /opt/ros/kinetic/setup.bash
+
+echo "export USER=$USER" >> "/home/$USER/.bashrc"
+echo "export HOME=/home/$USER" >> "/home/$USER/.bashrc"
+
+export PATH=$PATH:/opt/conda/bin
+echo 'export PATH=$PATH:/opt/conda/bin' >> "/home/$USER/.bashrc"
```
ryanjulian: Please don't produce unnecessary prints to the console. Unless the user absolutely needs to know something for the behavior of the command to be meaningful, remain silent.

ryanjulian:

Issue #263, opened 2018-08-09
CategorialMLPPolicy is broken


ryanjulian:
```python
import gym

from garage.baselines import LinearFeatureBaseline
from garage.misc.instrument import run_experiment
from garage.misc.instrument import stub
from garage.tf.algos import TRPO
from garage.tf.envs import TfEnv
from garage.tf.policies import CategoricalMLPPolicy

# Need to wrap in a tf environment and force_reset to true
# see https://github.com/openai/rllab/issues/87#issuecomment-282519288
env = TfEnv(gym.make("CartPole-v0"))

policy = CategoricalMLPPolicy(
    name="policy", env_spec=env.spec, hidden_sizes=(32, 32))

baseline = LinearFeatureBaseline(env_spec=env.spec)

algo = TRPO(
    env=env,
    policy=policy,
    baseline=baseline,
    batch_size=4000,
    max_path_length=200,
    n_itr=120,
    discount=0.99,
    step_size=0.01,
)

algo.train()
```

```sh
(garage) rjulian@tars:~/code/garage$ python examples/tf/trpo_gym_tf_cartpole.py 
WARN: gym.spaces.Box autodetected dtype as <class 'numpy.float32'>. Please provide explicit dtype.
Traceback (most recent call last):
  File "examples/tf/trpo_gym_tf_cartpole.py", line 27, in <module>
    step_size=0.01,
  File "/home/rjulian/code/garage/garage/tf/algos/trpo.py", line 40, in __init__
    **kwargs)
  File "/home/rjulian/code/garage/garage/tf/algos/npo.py", line 52, in __init__
    super(NPO, self).__init__(policy=policy, **kwargs)
  File "/home/rjulian/code/garage/garage/tf/algos/batch_polopt.py", line 90, in __init__
    self.init_opt()
  File "/home/rjulian/code/garage/garage/tf/algos/npo.py", line 62, in init_opt
    pol_loss, pol_kl = self._build_policy_loss(pol_loss_inputs)
  File "/home/rjulian/code/garage/garage/tf/algos/npo.py", line 218, in _build_policy_loss
    policy_entropy = self._build_entropy_term(i)
  File "/home/rjulian/code/garage/garage/tf/algos/npo.py", line 315, in _build_entropy_term
    name="policy_dist_info_flat")
  File "/home/rjulian/code/garage/garage/tf/policies/categorical_mlp_policy.py", line 71, in dist_info_sym
    return dict(prob)
[...]

PR #264, opened 2018-08-09
Fix CategoricalMLPPolicy

CatherineSue:
The problem is that CategoricalMLPPolicy._f_prob and
tf.Distribution.Categorical. It returns wrong shape array.

Fixes: #263

### DIFF
--- a/garage/tf/policies/categorical_mlp_policy.py
+++ b/garage/tf/policies/categorical_mlp_policy.py
@@ -51,7 +51,7 @@ class CategoricalMLPPolicy(StochasticPolicy, LayersPowered, Serializable):
             with tf.name_scope(self._prob_network_name):
                 prob_network_outputs = L.get_output(prob_network.output_layer)
             self._f_prob = tensor_utils.compile_function(
-                [prob_network.input_layer.input_var], [prob_network_outputs])
+                [prob_network.input_layer.input_var], prob_network_outputs)
 
             self._dist = Categorical(env_spec.action_space.n)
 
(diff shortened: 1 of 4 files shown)


garage/tf/policies/categorical_mlp_policy.py
```diff
@@ -51,7 +51,7 @@ def __init__(
             with tf.name_scope(self._prob_network_name):
                 prob_network_outputs = L.get_output(prob_network.output_layer)
             self._f_prob = tensor_utils.compile_function(
-                [prob_network.input_layer.input_var], [prob_network_outputs])
+                [prob_network.input_layer.input_var], prob_network_outputs)
```
CatherineSue: This problem also appears in garage/tf/policies/deterministic_mlp_policy.py. Should I change it?

zhanpenghe: Yes, please do that.

ryanjulian:
Can you add a test which is broken before this change and fixed after?

CatherineSue:
I added a test file into examples/tf.

ryanjulian:
If you are fixing DeterministicMLPPolicy, can you add a unit test for that as well?

CatherineSue:
I think the DeterministicMLPPolicy is supposed to be used in DDPG, right? But currently, because we need to use gym.spaces.Dict to use DDPG+HER and garage spaces don't have the wrapper for it, we can't use TfEnv() wrapper when using DDPG. But DeterministicMLPPolicy is designed to use TfEnv() and the garage.spaces. I can change it to make DeterminisitcMLPPolicy work but I don't think it is a good idea.

ryanjulian:
Okay, please file a bug and fix it with Jonathon in the next CL

Issue #243, opened 2018-07-31
Support for gym Dict spaces


ryanjulian:
See https://github.com/vitchyr/multiworld/blob/master/multiworld/core/flat_goal_env.py for a convenient wrapper to also include for algos which don't expect a dictspace.

PR #270, opened 2018-08-10
Add dict support for TfEnv

jonashen:
Relevant: #243


garage/tf/envs/base.py
```diff
@@ -41,6 +43,8 @@ def _to_garage_space(self, space):
         """
         if isinstance(space, GymBox):
             return Box(low=space.low, high=space.high)
+        elif isinstance(space, GymDict):
+            return Dict(space)
```
CatherineSue: I think the interface should be like `Dict(space.spaces)`

zhanpenghe:
You can also write a small test like this:
```
dict_1 = GymDict(some_other_spaces)
dict_2 = GarageDict(dict_1)
sampled = dict_1.sample()
assert dict_2.flat_dim == dict_2.flatten(sampled).shape[-1]
assert dict_2.unflatten(dict_2.flatten(sampled)) == sampled
# Do the sample with other methods
```

garage/spaces/dict.py
```diff
@@ -0,0 +1,68 @@
...
+class Dict(Space):
+    """
+    A dictionary of simpler spaces.
+    Example usage:
+    self.observation_space = spaces.Dict({"position": spaces.Discrete(2), "velocity": spaces.Discrete(3)})
+    """
+
+    def __init__(self, spaces):
+        if isinstance(spaces, dict):
+            spaces = OrderedDict(sorted(list(spaces.items())))
```
zhanpenghe: You might want to wrap the spaces to garage spaces.
E.g Dict(Box(3,), Box(2,)) should become GarageDict(GarageBox(3,), GarageBox(3,))

ryanjulian: Yeah, it should be recursive.

jonashen: Since GarageEnv is abstract, I will implement this in garage.tf.spaces.dict (and garage.theano.spaces.dict) instead and use TfDict(TfBox(3,), TfBox(3,))

garage/tf/spaces/dict.py
```diff
@@ -0,0 +1,95 @@
...
+            [(key, space.sample()) for key, space in self.spaces.items()])
+
+    @property
+    def flat_dim(self):
+        return sum([space.flat_dim for _, space in self.spaces.items()])
+
+    def flat_dim_with_keys(self, keys: Iterable):
+        return sum([self.spaces[key].flat_dim for key in keys])
+
+    def flatten(self, x):
```
CatherineSue: I think the x is a dict here? Then it should probably be
```
def flatten(self, x):
      return np.concatenate(space.flatten(xs) for space, xs in zip(self.spaces.values(), self.x.values())
```
We need to make sure the space and x has same order.

CatherineSue: Yes.

scripts/travisci/check_imports.py
```diff
@@ -37,13 +37,7 @@ def get_import_statement(node, lines):
...
 
 # Add file path prefixes here to ignore them.
-IGNORE_PATHS = [
-    "build/",
-    "dist/",
-    "docs/",
-    "garage/contrib/",
-    "sandbox/",
-]
+IGNORE_PATHS = ["build/", "dist/", "docs/", "garage/contrib/", "tests/"]
```
ryanjulian: Is there a reason this file was touched? It seems unrelated to the PR

jonashen: It's because for some reason `check_imports` claims that importing from the `tests` folder is invalid, e.g. `from tests.envs.dummy import DummyDictEnv` will throw an error.

jonashen: This affects the test I wrote to test this code located at `tests/tf/spaces/test_dict_space.py`

tests/tf/spaces/test_dict_space.py
```diff
@@ -0,0 +1,45 @@
...
+from garage.tf.envs import TfEnv
+
+
+class TestDictSpace(unittest.TestCase):
+    def test_dict_space(self):
+        ext.set_seed(0)
+
+        # A dummy dict env
+        dummy_env = DummyDictEnv()
+        dummy_act = dummy_env.action_space
```
CatherineSue: Shouldn't this be dummy_env.observation_space?

CatherineSue: Ignore this. I found you set action_space to dict as well. Sorry.

jonashen: I was under the impression we can't sample from observation_space b/c some observation spaces will be from -inf to inf, which is not able to be sampled.

Issue #199, opened 2018-07-17
Add recurrent policy implementation to NPO


CatherineSue:

jonashen:
`examples/tf/trpo_cartpole_recurrent.py` depends on this issue being resolved in order to work.

Issue #162, opened 2018-07-09
Running examples/tf/trpo_cartpole_recurrent.py throws attribute error


jonashen:
```Traceback (most recent call last):
  File "/anaconda2/envs/garage/lib/python3.5/site-packages/tensorflow/python/framework/ops.py", line 2327, in get_attr
    c_api.TF_OperationGetAttrValueProto(self._c_op, name, buf)
tensorflow.python.framework.errors_impl.InvalidArgumentError: Operation 'NPO/init_opt/policy/dist_info_sym/scan/TensorArrayStack/TensorArrayGatherV3' has no attr named '_XlaCompile'.
```

ryanjulian:
Related to #199

CatherineSue:
Fixed in #272

PR #272, opened 2018-08-11
Add reccurrent policy to NPO

CatherineSue:
Add recurrent policy support to NPO so that recurrent policies can be
used in NPO/TRPO/PPO. This includes:

 - Fix a bug in tf/NPO. Not all the policies have `_dist` attribute,
   change it to use distribution property.
 - Add recurrent policy implementation to NPO. Recurrent policies use
   unflatten inputs.

Fixes: #199 , #162

### DIFF
--- a/garage/tf/algos/npo.py
+++ b/garage/tf/algos/npo.py
@@ -56,9 +56,6 @@ class NPO(BatchPolopt):
 
     @overrides
     def init_opt(self):
-        if self.policy.recurrent:
-            raise NotImplementedError
-
         pol_loss_inputs, pol_opt_inputs = self._build_inputs()
         self._policy_opt_inputs = pol_opt_inputs
 
@@ -121,7 +118,7 @@ class NPO(BatchPolopt):
         observation_space = self.policy.observation_space
         action_space = self.policy.action_space
 
-        policy_dist = self.policy._dist
+        policy_dist = self.policy.distribution
 
         with tf.name_scope("inputs"):
             obs_var = observation_space.new_tensor_variable(
@@ -224,7 +221,7 @@ class NPO(BatchPolopt):
         return policy_loss_inputs, policy_opt_inputs
 
     def _build_policy_loss(self, i):
-        pol_dist = self.policy._dist
+        pol_dist = self.policy.distribution
 
         policy_entropy = self._build_entropy_term(i)
 
@@ -244,6 +241,9 @@ class NPO(BatchPolopt):
             adv_valid = filter_valids(
                 adv_flat, i.flat.valid_var, name="adv_valid")
 
+            if self.policy.recurrent:
+                advantages = tf.reshape(advantages, [-1, self.max_path_length])
+
             # Optionally normalize advantages
             eps = tf.constant(1e-8, dtype=tf.float32)
             if self.center_adv:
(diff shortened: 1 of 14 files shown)


garage/tf/algos/trpo.py
```diff
@@ -23,12 +23,13 @@ def __init__(self,
                  optimizer=None,
                  optimizer_args=None,
                  **kwargs):
-        if kl_constraint == KLConstraint.HARD:
-            optimizer = ConjugateGradientOptimizer
-        elif kl_constraint == KLConstraint.SOFT:
-            optimizer = PenaltyLbfgsOptimizer
-        else:
-            raise NotImplementedError("Unknown KLConstraint")
+        if optimizer is None:
```
ryanjulian: `if not optimizer`

CatherineSue:
Will do.

Yes. But it leads me to find other bugs in the policy file. I've fixed them. Also I will add unittests for all the garage/tf/policies.

Best,
Chang

ryanjulian:
Let me know when it's ready for another look!

garage/tf/algos/trpo.py
CatherineSue: Change all the None check to `if not`?

garage/tf/algos/npo.py
```diff
@@ -244,6 +241,9 @@ def _build_policy_loss(self, i):
             adv_valid = filter_valids(
                 adv_flat, i.flat.valid_var, name="adv_valid")
 
+            if self.policy.recurrent:
+                advantages = tf.reshape(advantages, [-1, self.max_path_length])
```
ryanjulian: hmm what shape do the recurrent policies need?

`compute_advantages` should output shape `(trajectory, step, advantage)` and then `adv_flat` should have shape `(trajectory x time, advantage)`.

Reading this, it looks like your recurrent policy is getting shape `(trajectory x advantage, step)`, which I don't really understand.

garage/tf/algos/npo.py
```diff
@@ -256,32 +256,56 @@ def _build_policy_loss(self, i):
...
-                policy_dist_info_flat,
-                i.flat.valid_var,
-                name="policy_dist_info_valid")
+            if self.policy.recurrent:
+                policy_dist_info = self.policy.dist_info_sym(
+                    i.obs_var,
+                    i.policy_state_info_vars,
+                    name="policy_dist_info")
+            else:
+                policy_dist_info_flat = self.policy.dist_info_sym(
```
ryanjulian: I wasn't aware we could compute the dense versions of `dist_info`, `kl`, and `lr`.

Please verify that this actually works.

If it works, please remove the flat/valid views from NPO and just flatten/filter valids everything right before the final loss function, which is less code.

garage/tf/algos/npo.py
CatherineSue: `compute_advantages` is actually output shape `(trajectory x step x advantage)` because here. https://github.com/rlworkgroup/garage/blob/074c2aa100617c27f73d34a31ab257ccedaf8feb/garage/tf/misc/tensor_utils.py#L218

ryanjulian: I think that's what I said above?

What is the shape of advantages after you reshape it here?

CatherineSue: Sorry, I mean `(trajectory x step x advantage, )`.

After reshaping, the shape of advantages is `(trajectory x advantage, step). Let me check if the shape is correct for recurrent policy.

garage/tf/algos/npo.py
CatherineSue: The recurrent policy's `dist_info` requires a shape of `(trajectory, step, data_flat_dim)`, but the nonrecurrent policy's `dist_info` requires an input shape of `(trajectory x step, data_flat_dim)`. I think we still need flat/valid views for nonrecurrent policies.

ryanjulian: Alright thanks for checking! We can fix this when we remove garage.distributions

ryanjulian:
@CatherineSue is this ready for submit?

CatherineSue:
Yes.

Issue #235, opened 2018-07-27
Update setup_linux.sh and setup_osx.sh


ryanjulian:
The new canonical setup steps are in `docker/Dockerfile.ci`. Most of them can be ported into a setup script for a regular machine (omitting the CI-specific parts like the fake X server). Then we can use setup_linux.sh to mostly setup the docker as well.

* There should be a --gpu switch to setup a gpu-based garage. We can't support GPU on Mac.
* setup_osx.sh will be very similar to the CPU version of setup_linux.sh, but there are always complications. I think you can add a job in travis to run on OSX. It can at least run setup_osx.sh on a cron job to make sure it stays fairly stable.

ryanjulian:
I'm not sure it's possible, but if we can get OpenGL support on Docker for OS X (host: OS X, guest: ubuntu) then we can eliminate specific OS X support entirely. GPU support is irrelevant because nobody uses CUDA on OS X

gonzaiva:
Could we pass the path of mjkey.txt as an argument of the scripts?

ryanjulian:
Yes please.

No need to ask for the zip file. We can also assume that users have internet access and use wget to retrieve mjpro150.zip.

gonzaiva:
The setup for linux is fixed in #258 
For OS X is being worked at #279, however it does not consider using Docker.

gonzaiva:
Closed with #279

Issue #222, opened 2018-07-25
Setup Docker container using setup_linux.sh


ryanjulian:
This will ensure consistency between container and non-container installs.

ryanjulian:
In retrospect this might be a bad idea because it will mean the dockerfile has much fewer layers, so the cache won't work as well. In practice that might not matter--i'm not sure.

gonzaiva:
Based on the latest comments, for the 2018.09 Release, it's sufficient to add a header comment in the setup scripts and the docker file that notifies developers to keep changes consistent across these files.

PR #279, opened 2018-08-13
Update setup script for OS X

gonzaiva:
The script has been updated based on setup_linux.sh, using homebrew to
install Linux packages and updating the requirements for mujoco-py, gym
and baselines according to their documentation for OS X.
Since mujoco-py is installed within the setup scripts for Linux and OS
X, the script to setup mujoco has been removed.
Also, the tensorflow package was added to environment.yml so it can be
installed out of the box without the scripts. The feature to install the
GPU flavor of TensorFlow may be removed from the Linux script once
everything could be installed using environment.yml only, or another
single list of dependencies.
Finally, the default for set-envvar and the correct replacement of
string in print error and warning functions have been added to the
script to setup linux.


gonzaiva:
Reviewers, I don't have access to a laptop with OS X to test setup_osx.sh, however I checked that the functions and dependencies in the script are valid for OS X.
Namely, I checked that all the Linux functions in the script are available in Mac based on this [list](https://ss64.com/osx/). Also I considered the OS X requirements for [mujoco-py](https://github.com/openai/mujoco-py#requirements), [gym](https://github.com/openai/gym#installing-everything) (gym [issue](https://github.com/openai/gym/issues/191)) and [baselines](https://github.com/openai/baselines#mac-os-x).
One of the requirements to build GLFW is Xcode as mentioned [here](http://www.glfw.org/docs/latest/compile.html#compile_deps_xcode), however I'm not sure about how to install this dependency. Are the [xcode command line tools](https://github.com/rlworkgroup/garage/blob/c580437b33f14fd905efecf00bb577637f85a16d/scripts/setup_osx.sh#L131) sufficient?
Finally, besides your useful comments, I would appreciate if those of you using OS X could test the script to make sure everything works fine.
Thanks in advance.

scripts/setup_osx.sh
```diff
@@ -1,38 +1,248 @@
...
+# THE DEFAULTS INITIALIZATION - POSITIONALS
+_positionals=()
+# THE DEFAULTS INITIALIZATION - OPTIONALS
+_arg_set_envvar="off"
 
-# Make sure that we're under the directory of the project
-cd "$(dirname "$0")/.."
+print_help ()
+{
+  printf '%s\n' "Installer of garage for Linux."
```
CatherineSue: Should this be `Installer of garage for OS X`?

gonzaiva: Yes, I will change it. Thanks Chang.

gonzaiva:
I've noticed that issue #235 considers the use of Travis to test this script, so I can try that to test my script.

gonzaiva:
On the other hand, using setup_osx in Travis may have the same problem as with setup_linux described in issue #222.

ryanjulian:
* Please test on 1-2 real OS X machines.

Check for the OS X versions you tested with in the script. If the user's version is different, warn the user that it might not work. Ask them to file a PR updating the list of known-good OSX versions (or a PR with fixes for their version) and include a link to our Github Issues page.

* XCode command line tools is a standard requirement for OSX dev machines. Brew requires it, anyway.

* We won't be making an OSX-based docker container so #222 is not a problem. Please try testing it with Travis if they have a reasonably-current array of OSX versions. Set it up as a separate build job and don't test it in the main CI build (it can be a cron job instead).

scripts/setup_osx.sh
```diff
@@ -1,38 +1,248 @@
...
+print_warning() {
+  printf "\e[0;33m%s\e[0m" "${1}"
+}
+
+# Verify there's a file in the mjkey path
+test -f "${_arg_mjkey_path}" || _PRINT_HELP=yes die \
+  "The path ${_arg_mjkey_path} of the MuJoCo key is not valid." 1
+
+# Make sure that we're under the garage directory
+GARAGE_DIR="$(readlink -f $(dirname $0)/..)"
```
jonashen: I got this error while running the script:

```
readlink: illegal option -- f
usage: readlink [-n] [file ...]
```

scripts/setup_osx.sh
```diff
@@ -1,38 +1,248 @@
...
+pre-commit install -t commit-msg
+source deactivate
+
+# Add garage to python modules
+export PYTHONPATH="$PYTHONPATH:${GARAGE_DIR}"
+PYTHON_ENV_VAR="PYTHONPATH=\"\$PYTHONPATH:${GARAGE_DIR}\""
+if [[ "${_arg_set_envvar}" = on ]]; then
+  echo "export ${PYTHON_ENV_VAR}" >> "${BASH_RC}"
+else
+  echo "Remember to export the following environment variables before" \
```
jonashen: Can you add an extra newline before this line? Or else it looks like it's part of the pre-commit install:

```
pre-commit installed at /private/var/folders/kj/dgb2_y1n1wj515xphc6clvzm0000gn/T/tmp.DWwDXgPH/glfw/.git/hooks/commit-msg
Remember to export the following environment variables before running garage:
```

scripts/setup_osx.sh
```diff
@@ -1,38 +1,248 @@
...
+    "running garage:"
+  echo "${LD_LIB_ENV_VAR}"
+  echo "${PATH_ENV_VAR}"
+  echo "${PYTHON_ENV_VAR}"
+  echo "You may wish to edit your .bashrc to prepend the exports of these"
+    "environment variables."
```
jonashen: You need an extra `\` here I think: 
```
You may wish to edit your .bashrc to prepend the exports of these
./scripts/setup_osx.sh: line 243: environment variables.: command not found
```

scripts/setup_osx.sh
```diff
@@ -1,38 +1,248 @@
...
+# Leave a note in ~/.bashrc for the added environment variables
+if [[ "${_arg_set_envvar}" = on ]]; then
+  echo -e "\n# Added by the garage installer" >> "${BASH_PROF}"
 fi
 
-echo "Creating conda environment..."
+# Set up MuJoCo
+if [[ ! -d "${HOME}"/.mujoco/mjpro150 ]]; then
+  mkdir "${HOME}"/.mujoco
+  MUJOCO_ZIP="$(mktemp --suffix=_mujoco.zip)"
```
CatherineSue: I don't think OS X has --suffix option for mktemp.
```
$ mktemp --suffix=_mujoco.zip
mktemp: illegal option -- -
usage: mktemp [-d] [-q] [-t prefix] [-u] template ...
       mktemp [-d] [-q] [-u] -t prefix
```

scripts/setup_osx.sh
```diff
@@ -1,38 +1,248 @@
...
+  echo "${PATH_ENV_VAR}"
+  echo "${PYTHON_ENV_VAR}"
+  echo "You may wish to edit your .bashrc to prepend the exports of these"
+    "environment variables."
+fi
 
-echo "Conda environment created! Make sure to run \`source activate garage\`" \
+echo -e "\nGarage is installed! To make the changes take effect, work under" \
+  "a new terminal. Also, make sure to run \`source activate garage\`" \
   "whenever you open a new terminal and want to run programs under garage."
```
jonashen: Can you divide this into two or more lines? It's too long when it's being displayed in the console.

scripts/setup_osx.sh
ryanjulian: It looks like you are installing pre-commit in the temporary git checkout for building glfw? THat seems not useful.

Also, you should (hopefully) not need to compile glfw for Mac. Frankly I'm surprised that worked.

scripts/setup_osx.sh
```diff
@@ -1,38 +1,248 @@
...
+  done
+}
+
+
+handle_passed_args_count ()
+{
+  _required_args_string="'mjkey-path'"
+  error_msg="$(echo "FATAL ERROR: Not enough positional arguments - we" \
+    "require exactly 1 (namely: $_required_args_string), but got only" \
+    "${#_positionals[@]}.")"
```
jonashen: IIRC MuJoCo recommends moving the mjkey.txt into the `.mujoco` folder, so maybe we can remind users that their `mjkey.txt` is located for example in `./scripts/setup_osx.sh ~/.mujoco/mjkey.txt`. At least for me, I had to think for a while about where I last stored my mjkey.txt after seeing this error.

scripts/setup_osx.sh
ryanjulian: OS X uses `.bash_profile` not `.bashrc`

scripts/setup_osx.sh
ryanjulian: Your console doesn't wrap it?

scripts/setup_osx.sh
```diff
@@ -1,38 +1,248 @@
...
+
+# Extras
+source activate garage
+# Prevent pip from complaining about available upgrades
+pip install --upgrade pip
+# Fix Box2D install
+# See https://github.com/openai/gym/issues/100
+# See https://github.com/pybox2d/pybox2d/issues/82
+pip uninstall -y Box2D Box2D-kengz box2d-py
+pip install Box2D
```
jonashen: This should be `pip install box2d-py`. This is the error I get from `Box2D`:

```
(garage) garage (fix_setup_osx) $ python examples/theano/trpo_cartpole.py
  File "examples/theano/trpo_cartpole.py", line 3, in <module>
    from garage.envs.box2d import CartpoleEnv
  File "/Users/jonathon/Documents/garage/garage/garage/envs/box2d/__init__.py", line 1, in <module>
    from garage.envs.box2d.box2d_env import Box2DEnv
  File "/Users/jonathon/Documents/garage/garage/garage/envs/box2d/box2d_env.py", line 9, in <module>
    from garage.envs.box2d.box2d_viewer import Box2DViewer
  File "/Users/jonathon/Documents/garage/garage/garage/envs/box2d/box2d_viewer.py", line 1, in <module>
    from Box2D import b2ContactListener
  File "/anaconda2/envs/garage/lib/python3.6/site-packages/Box2D/__init__.py", line 20, in <module>
    from .Box2D import *
  File "/anaconda2/envs/garage/lib/python3.6/site-packages/Box2D/Box2D.py", line 435, in <module>
    _Box2D.RAND_LIMIT_swigconstant(_Box2D)
AttributeError: module '_Box2D' has no attribute 'RAND_LIMIT_swigconstant'
```

scripts/setup_osx.sh
gonzaiva: It seems this option is not available in OS X. Even though greadlink is available, I wouldn't like to download a new tool just for this little thing. I will find a way to work this out with more primitive tools so there's no compatibility issue.

scripts/setup_osx.sh
gonzaiva: I will.

scripts/setup_osx.sh
gonzaiva: Ok, I will.

scripts/setup_osx.sh
jonashen: It didn't for me, but I find it a good habit in general. It's just a personal preference.

scripts/setup_osx.sh
jonashen: Some OS X users use .bashrc, and .bash_profile will default to .bashrc first. I found this inside my .bash_profile:
```
if [ -f ~/.bashrc ]; then
   . ~/.bashrc;
fi
```

scripts/setup_osx.sh
gonzaiva: A solution could be to use the dry-run option to only output the name of the temporary file (not create it), and append the suffix.
This is a little bit unsafe if the file happens to exist, but since these file are the tmp folder and with random characters in their name, I don't think there's a chance this will happen.

scripts/setup_osx.sh
gonzaiva: Sure, I can split them to a constant length.

scripts/setup_osx.sh
jonashen: You can just say "edit your .bashrc/.bash_profile" instead

scripts/setup_osx.sh
gonzaiva: I don't understand this request very well. This script asks for the path where mjkey.txt is so it can be moved into ~/.mujco/mjkey.txt as shown [here](https://github.com/rlworkgroup/garage/blob/c580437b33f14fd905efecf00bb577637f85a16d/scripts/setup_osx.sh#L223).
For example, if the key is downloaded in ~/Downloads/mjkey.txt, then it's copied into ~/.mujoco/mjkey.txt.
Do you mean printing a help message or something like that?

scripts/setup_osx.sh
jonashen: sgtm

scripts/setup_osx.sh
jonashen: Ignore what I said previously, I didn't see that line.

ryanjulian: I agree with Angel that it's fine.

scripts/setup_osx.sh
ryanjulian: OS X is BSD but not Linux, so you can only really rely on `bash` primitives and POSIX commands.

ryanjulian: This trick has worked for me before https://stackoverflow.com/questions/3572030/bash-script-absolute-path-with-osx

scripts/setup_osx.sh
```diff
@@ -1,38 +1,248 @@
...
+  cmake \
+  boost \
+  boost-python \
+  ffmpeg \
+  sdl2 \
+  swig \
+  openmpi
+
+# Build GLFW because the brew version is too old
+GLFW_DIR="$(mktemp -d)/glfw"
```
ryanjulian: Is it true that the brew version is too old? Does this even work?

scripts/setup_osx.sh
ryanjulian: Just don't add a suffix or put it in a directory instead.

scripts/setup_osx.sh
gonzaiva: The pre-commit happened there because readlink failed as pointed in a comment above, so cd does not change to the garage dir.
I thought we were compiling glfw to include the newest changes since their last release is too old. Is glfw already included in Mac or installing from brew is good enough?

scripts/setup_osx.sh
ryanjulian: There is a dumb bug in a recent version of swig which causes this problem. You may need to download a specific version of swig from brew or build it yourself to avoid this. You need to install the new version of swig *before* installing Box2D the final time. You also need to force pip to re-build the package (--no-binary or --no-wheel) on install.

scripts/setup_osx.sh
gonzaiva: I think adding the env vars to only .bashrc does not work, unless they include the snippet of code you added above in .bash_profile. I will tell them to add them into .bash_profile to avoid the confusion.

ryanjulian: tl;dr: use bash_profile

Your code snippet is what Apple placed in OS X to avoid compatibility problems with linux tools assuming .bashrc

Many Linux distros use .bashrc for user profiles even though they shouldn't, because many people use their machines in headless (e.g. SSH) mode.

https://apple.stackexchange.com/questions/51036/what-is-the-difference-between-bash-profile-and-bashrc

scripts/setup_osx.sh
jonashen: I'm actually using the brew version of glfw right now and it works fine for me.

ryanjulian:
The environment libraries are most likely to break across platform. Please thoroughly test lots of gym environments, our custom mujoco envs, dm_control, and our custom box2d envs before declaring this is working.

It is also work it to run nose2 and make sure that the unit tests succeed on OS X.

scripts/setup_osx.sh
jonashen: sgtm

scripts/setup_osx.sh
gonzaiva: Ok, then I will remove the compilation of glfw and add it to the brew.

scripts/setup_osx.sh
ryanjulian: If box2d-py solves the problem universally then we can just use that. IIRC I had trouble getting it to work reliably with Docker and started using the latest version from the author.

jonashen:
I'm still having problems with the Box2D issue.

ryanjulian:
@jonashen does everything else work with the exception of Box2D?

jonashen:
Yes, but one thing to note is I have many dependencies already installed (e.g. xcode-select), and so can't uncover them easily.

ryanjulian:
Ok I will try it in a clean VM

ryanjulian:
Here's the log of my attempt on a clean VM.

I think the problem happens at `chmod: /var/folders/cg/b3cdgh9d3nz7pr_j1r8n2_540000gn/T/tmp.eFUxB2xP/miniconda.sh: No such file or directory`

You should `set -e` at the top of your bash script so that it stops on an error, rather than just running all the way through.
[Uploading osx.log…]()

scripts/setup_osx.sh
```diff
@@ -1,38 +1,265 @@
...
+  echo "export ${LD_LIB_ENV_VAR}" >> "${BASH_PROF}"
+fi
+
+# Set up conda
+hash conda 2>/dev/null || {
+  CONDA_INSTALLER="$(mktemp -d)/miniconda.sh"
+  sudo chmod u+x "${CONDA_INSTALLER}"
+  wget https://repo.continuum.io/miniconda/Miniconda2-latest-MacOSX-x86_64.sh \
+    -O "${CONDA_INSTALLER}"
+  sudo bash "${CONDA_INSTALLER}" -b
```
ryanjulian: why use sudo here? This should not be necessary and is explicitly discouraged by conda.

scripts/setup_osx.sh
```diff
@@ -1,38 +1,265 @@
-#!/bin/bash
-# Make sure that pip is available
-hash brew 2>/dev/null || {
-  echo "Please install homebrew before continuing. You can use the" \
-    "following command to install:"
-  echo "/usr/bin/ruby -e \"\$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)\""
-  exit 0
+#!/usr/bin/env bash
+# This script installs garage on OS X distributions.
```
ryanjulian: `set -e` here

scripts/setup_osx.sh
```diff
@@ -1,38 +1,265 @@
...
+LD_LIB_ENV_VAR="LD_LIBRARY_PATH=\"\$LD_LIBRARY_PATH:${HOME}/.mujoco/mjpro150"
+LD_LIB_ENV_VAR="${LD_LIB_ENV_VAR}/bin\""
+if [[ "${_arg_set_envvar}" = on ]]; then
+  echo "export ${LD_LIB_ENV_VAR}" >> "${BASH_PROF}"
+fi
+
+# Set up conda
+hash conda 2>/dev/null || {
+  CONDA_INSTALLER="$(mktemp -d)/miniconda.sh"
+  sudo chmod u+x "${CONDA_INSTALLER}"
```
ryanjulian: don't use `sudo` for this. also, this goes *after* `wget` or your file doesn't exist.

environment.yml
```diff
@@ -48,6 +48,7 @@ dependencies:
         - pyzmq
         - scipy
         - tensorboard
+        - tensorflow
```
ryanjulian: pin tensorflow to `tensorflow<1.10,>=1.9.0`. 1.10 has breaking changes

ryanjulian:
Once I have fixed the conda permission problem I get 
image

scripts/setup_osx.sh
```diff
@@ -1,38 +1,265 @@
...
+# See https://github.com/pybox2d/pybox2d/issues/82
+pip uninstall -y Box2D Box2D-kengz box2d-py
+pip install Box2D
+# We need a MuJoCo key to import mujoco_py
+cp ${_arg_mjkey_path} "${HOME}"/.mujoco/mjkey.txt
+python -c 'import mujoco_py'
+# Set up pre-commit in local repo
+pre-commit install -t pre-commit
+pre-commit install -t pre-push
+pre-commit install -t commit-msg
```
ryanjulian: You should run `pre-commit` once after installing the hooks--it runs a lengthy process which installs a virtualenv for the hooks.

scripts/setup_osx.sh
```diff
@@ -1,38 +1,265 @@
...
+# Prevent pip from complaining about available upgrades
+pip install --upgrade pip
+# Fix Box2D install
+# See https://github.com/openai/gym/issues/100
+# See https://github.com/pybox2d/pybox2d/issues/82
+pip uninstall -y Box2D Box2D-kengz box2d-py
+pip install Box2D
+# We need a MuJoCo key to import mujoco_py
+cp ${_arg_mjkey_path} "${HOME}"/.mujoco/mjkey.txt
+python -c 'import mujoco_py'
```
ryanjulian: Please leave the comment explaining this line

scripts/setup_osx.sh
```diff
@@ -1,38 +1,265 @@
...
+PYTHON_ENV_VAR="PYTHONPATH=\"\$PYTHONPATH:${GARAGE_DIR}\""
+if [[ "${_arg_set_envvar}" = on ]]; then
+  echo "export ${PYTHON_ENV_VAR}" >> "${BASH_PROF}"
+else
+  echo -e "\nRemember to export the following environment variables before" \
+    "running garage:"
+  echo "${LD_LIB_ENV_VAR}"
+  echo "${PATH_ENV_VAR}"
+  echo "${PYTHON_ENV_VAR}"
+  echo "You may wish to edit your .bashrc to prepend the exports of these" \
```
ryanjulian: `.bash_profile`

scripts/setup_osx.sh
```diff
@@ -1,38 +1,265 @@
...
-if [[ "${?}" -ne 0 ]]; then
-  echo -e "\e[0;31mError: MuJoCo couldn't be set up\e[0m"
-  exit 1
+parse_commandline ()
+{
+  while test $# -gt 0
+  do
+    _key="$1"
+    case "$_key" in
+      --no-set-envvar|--set-envvar)
```
ryanjulian: please call this `--modify_bashrc` to make it clear we're going to touch the bashrc, not just the environment.

scripts/setup_osx.sh
```diff
@@ -1,38 +1,265 @@
...
+      -h|--help)
+        print_help
+        exit 0
+        ;;
+      -h*)
+        print_help
+        exit 0
+        ;;
+      *)
+        _positionals+=("$1")
```
ryanjulian: Please make all arguments named arguments. You can use --mjkey

scripts/setup_osx.sh
```diff
@@ -1,38 +1,265 @@
...
+}
+
+# List of verified versions of OS X where garage has been installed
+VALID_VER=()
+VER="$(sw_vers -productVersion)"
+
+if [[ ! " ${VALID_VER[@]} " =~ " ${VER} " ]]; then
+  print_warning "It has not been verified whether garage will install" \
+    "correctly under your current OS X version (${VER}). If the installation" \
+    "is successful, please update the list of valid versions in this script." \
```
ryanjulian: 1. Provide a link to this file on Github
1. Rename `VALID_VER` to `VERIFIED_OSX_VERSIONS` to make it more clear to contributors what to edit
1. Put `VERIFIED_OSX_VERSIONS` at the top of the file, and leave a block comment explaining that this is the list you should edit if you were directed here by the installer
> You are attempting to install garage on a version of OS X which we have not verified is working.
>
> garage relies on community contributions to support OS X
>
> If this installation is successful, please add your OS X version to `VERIFIED_OSX_VERSIONS` to https://github.com/rlworkgroup/garage/blob/master/scripts/setup_osx.sh on GitHub and submit a pull request to rlworkgroup/garage to help out future users. If the installation is not initially successful, but you find changes which fix it, please help us out by submitting a PR with your updates to the setup script.

scripts/setup_osx.sh
```diff
@@ -1,38 +1,265 @@
...
+  /usr/bin/ruby -e "$(curl -fsSL \
+    https://raw.githubusercontent.com/Homebrew/install/master/install)"
+}
+
+# For installing garage: bzip2, git, glfw, unzip, wget
+# For building glfw: cmake
+# Required for OpenAI gym: cmake boost boost-python ffmpeg sdl2 swig wget
+# Required for OpenAI baselines: cmake openmpi
+brew update
+brew install \
```
ryanjulian: before `brew install \` add `brew install gcc@6 --no-multilib`

scripts/setup_osx.sh
```diff
@@ -1,38 +1,265 @@
...
+pip uninstall -y Box2D Box2D-kengz box2d-py
+pip install Box2D
+# We need a MuJoCo key to import mujoco_py
+cp ${_arg_mjkey_path} "${HOME}"/.mujoco/mjkey.txt
+python -c 'import mujoco_py'
+# Set up pre-commit in local repo
+pre-commit install -t pre-commit
+pre-commit install -t pre-push
+pre-commit install -t commit-msg
+source deactivate
```
ryanjulian: Copy the `config_personal_template.py` to `config_personal.py` so that users don't see the nag message on their first attempt to run something.

ryanjulian:
image

image

scripts/setup_osx.sh
```diff
@@ -1,38 +1,265 @@
...
+  local first_option all_short_options
+  all_short_options='h'
+  first_option="${1:0:1}"
+  test "$all_short_options" = "${all_short_options/$first_option/}" && \
+    return 1 || return 0
 }
 
 
-echo "Installing system dependencies"
-echo "You will probably be asked for your sudo password."
```
ryanjulian: I don't think sudo is necessary

scripts/setup_osx.sh
```diff
@@ -1,38 +1,265 @@
...
+# Configure MuJoCo as a shared library
+export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:${HOME}/.mujoco/mjpro150/bin"
+LD_LIB_ENV_VAR="LD_LIBRARY_PATH=\"\$LD_LIBRARY_PATH:${HOME}/.mujoco/mjpro150"
+LD_LIB_ENV_VAR="${LD_LIB_ENV_VAR}/bin\""
+if [[ "${_arg_set_envvar}" = on ]]; then
+  echo "export ${LD_LIB_ENV_VAR}" >> "${BASH_PROF}"
+fi
+
+# Set up conda
+hash conda 2>/dev/null || {
```
ryanjulian: don't install conda if it already exists

scripts/setup_osx.sh
```diff
@@ -1,38 +1,265 @@
...
+  CONDA_INSTALLER="$(mktemp -d)/miniconda.sh"
+  sudo chmod u+x "${CONDA_INSTALLER}"
+  wget https://repo.continuum.io/miniconda/Miniconda2-latest-MacOSX-x86_64.sh \
+    -O "${CONDA_INSTALLER}"
+  sudo bash "${CONDA_INSTALLER}" -b
+  # Add conda to executable programs
+  CONDA_PATH="${HOME}/miniconda2/bin/"
+  export PATH="$PATH:${CONDA_PATH}"
```
ryanjulian: conda doesn't use PATH anymore. instead you execute a script within your bashrc/bash_profile

scripts/setup_osx.sh
ryanjulian: I managed to fix this with
```sh
pip uninstall Box2D Box2D-kengz box2d-py
pip install Box2D-kengz --no-binary Box2D-kengz
```

We should update the Dockerfile an setup_linux.sh so everything is consistent.

The --no-binary forces pip to compile from source using the system swig libraries, rather than using the pre-compiled distribution from PyPI (which is apparently broken).

ryanjulian:
After adding all the fixes I provided, everything seems to work on my machine.

I've hacked on rllab with this Mac before, so I might have something which a clean machine doesn't.

It would be nice if we could save the users from seeing the GLFW warning, though it appears harmless.

scripts/setup_osx.sh
ryanjulian: (see the screenshot above)

scripts/setup_osx.sh
gonzaiva: Removed

scripts/setup_osx.sh
gonzaiva: Done

scripts/setup_osx.sh
gonzaiva: If the exit status of `hash conda` is not zero, that is, conda is not installed, then the commands between the brackets are executed. Otherwise, if the exit status is zero (conda is installed), they're not executed.

gonzaiva:
I applied the requested changes and I'm testing again. For OS X, using `set -e ` makes the script exit after performing the installation with brew. Here the bottom of the the script output: 
```
######################################################################## 100.0%
==> Pouring gmp-6.1.2_2.sierra.bottle.tar.gz
🍺  /usr/local/Cellar/gmp/6.1.2_2: 18 files, 3.1MB
==> Installing open-mpi dependency: isl
==> Downloading https://homebrew.bintray.com/bottles/isl-0.20.sierra.bottle.tar.gz
######################################################################## 100.0%
==> Pouring isl-0.20.sierra.bottle.tar.gz
🍺  /usr/local/Cellar/isl/0.20: 71 files, 3.9MB
==> Installing open-mpi dependency: mpfr
==> Downloading https://homebrew.bintray.com/bottles/mpfr-4.0.1.sierra.bottle.tar.gz
######################################################################## 100.0%
==> Pouring mpfr-4.0.1.sierra.bottle.tar.gz
🍺  /usr/local/Cellar/mpfr/4.0.1: 28 files, 4.6MB
==> Installing open-mpi dependency: libmpc
==> Downloading https://homebrew.bintray.com/bottles/libmpc-1.1.0.sierra.bottle.tar.gz
######################################################################## 100.0%
==> Pouring libmpc-1.1.0.sierra.bottle.tar.gz
🍺  /usr/local/Cellar/libmpc/1.1.0: 12 files, 353.9KB
==> Installing open-mpi dependency: gcc
==> Downloading https://homebrew.bintray.com/bottles/gcc-8.2.0.sierra.bottle.tar.gz
######################################################################## 100.0%
==> Pouring gcc-8.2.0.sierra.bottle.tar.gz
🍺  /usr/local/Cellar/gcc/8.2.0: 1,494 files, 343.6MB
==> Installing open-mpi dependency: libevent
==> Downloading https://homebrew.bintray.com/bottles/libevent-2.1.8.sierra.bottle.tar.gz
######################################################################## 100.0%
==> Pouring libevent-2.1.8.sierra.bottle.tar.gz
🍺  /usr/local/Cellar/libevent/2.1.8: 847 files, 2.2MB
==> Installing open-mpi
==> Downloading https://homebrew.bintray.com/bottles/open-mpi-3.1.1.sierra.bottle.tar.gz
######################################################################## 100.0%
==> Pouring open-mpi-3.1.1.sierra.bottle.tar.gz
🍺  /usr/local/Cellar/open-mpi/3.1.1: 736 files, 10.5MB
==> Caveats
==> bzip2
This formula is keg-only, which means it was not symlinked into /usr/local,
because macOS already provides this software and installing another version in
parallel can cause all kinds of trouble.

If you need to have this software first in your PATH run:
  echo 'export PATH="/usr/local/opt/bzip2/bin:$PATH"' >> ~/.bash_profile

For compilers to find this software you may need to set:
    LDFLAGS:  -L/usr/local/opt/bzip2/lib
    CPPFLAGS: -I/usr/local/opt/bzip2/include

==> git
Bash completion has been installed to:
  /usr/local/etc/bash_completion.d

zsh completions and functions have been installed to:
  /usr/local/share/zsh/site-functions

Emacs Lisp files have been installed to:
  /usr/local/share/emacs/site-lisp/git
==> unzip
This formula is keg-only, which means it was not symlinked into /usr/local,
because macOS already provides this software and installing another version in
parallel can cause all kinds of trouble.

If you need to have this software first in your PATH run:
  echo 'export PATH="/usr/local/opt/unzip/bin:$PATH"' >> ~/.bash_profile

==> gettext
This formula is keg-only, which means it was not symlinked into /usr/local,
because macOS provides the BSD gettext library & some software gets confused if both are in the library path.

If you need to have this software first in your PATH run:
  echo 'export PATH="/usr/local/opt/gettext/bin:$PATH"' >> ~/.bash_profile

For compilers to find this software you may need to set:
    LDFLAGS:  -L/usr/local/opt/gettext/lib
    CPPFLAGS: -I/usr/local/opt/gettext/include

==> openssl
A CA file has been bootstrapped using certificates from the SystemRoots
keychain. To add additional certificates (e.g. the certificates added in
the System keychain), place .pem files in
  /usr/local/etc/openssl/certs

and run
  /usr/local/opt/openssl/bin/c_rehash

This formula is keg-only, which means it was not symlinked into /usr/local,
because Apple has deprecated use of OpenSSL in favor of its own TLS and crypto libraries.

If you need to have this software first in your PATH run:
  echo 'export PATH="/usr/local/opt/openssl/bin:$PATH"' >> ~/.bash_profile

For compilers to find this software you may need to set:
    LDFLAGS:  -L/usr/local/opt/openssl/lib
    CPPFLAGS: -I/usr/local/opt/openssl/include

==> cmake
Emacs Lisp files have been installed to:
  /usr/local/share/emacs/site-lisp/cmake
Angels-Mac:garage aigonzal$
```
Don't try this new changes yet please.

ryanjulian:
If it is only `brew` and a couple other things misbehaving, you can use [this](https://stackoverflow.com/questions/48079029/exception-to-set-e-in-bash) pattern to temporarily disable exits.

scripts/setup_osx.sh
```diff
@@ -1,38 +1,267 @@
...
 
-# Make sure that we're under the directory of the project
-cd "$(dirname "$0")/.."
+print_help ()
+{
+  printf '%s\n' "Installer of garage for OS X."
+  printf 'Usage: %s [--mjkey <arg>] [--(no-)modify-bashprof] ' "$0"
+  printf '[-h|--help]\n'
+  printf '\t%s\n' "--mjkey: Path of the MuJoCo key (no default)"
+  printf '\t%s' "--modify-bashprof,--no-modify-bashprof: Set environment "
```
ryanjulian: Please keep the arg naming consistent across scripts. --modify-bashrc is fine on mac, even though it's actually in bash_profile.

scripts/setup_osx.sh
```diff
@@ -1,38 +1,267 @@
...
+    if [[ "${continue_var}" = "n" ]]; then
+      exit
+    fi
+  done
 fi
 
-echo "Creating conda environment..."
+# Verify there's a file in the mjkey path
+test -f "${_arg_mjkey}" || _PRINT_HELP=yes die \
+  "The path ${_arg_mjkey} of the MuJoCo key is not valid." 1
```
ryanjulian: How about "File ${_arg_mjkey} of the MuJoCo not found."

scripts/setup_osx.sh
ryanjulian: (this fixes the error with building mujoco_py)

scripts/setup_osx.sh
gonzaiva: Actually not even conda works inside the subshell. I don't know if this may be related to not using PATH but the script suggested by conda.

gonzaiva: If I add the binaries of conda in the PATH, calling conda inside the script works again

ryanjulian: I took a look at this problem, and it's actually really difficult to detect a conda installation from a shell script by inspecting just the shell environment. Considering #32 I don't think we need to address this in general.

I think it's sufficient to detect `[[ -d "${HOME}/miniconda2" ]]`. If that exists, then you can `. ${HOME}/miniconda2/etc/profile.d/conda.sh` to use the conda commands. If it doesn't just install it and use it.

Expert users with more than one conda install will just have to be careful to not use `--modify-bashrc`. 

If you want to be really kind to experts, you could add the arg `--use-conda` which takes the path to an existing conda, but I think that's overkill. If someone it juggling multiple conda installs, they should be enough of an expert to modify the script.

scripts/setup_osx.sh
```diff
@@ -1,38 +1,267 @@
...
+PYBOX2D_DIR="$(mktemp -d)/pybox2d"
+git clone https://github.com/pybox2d/pybox2d "${PYBOX2D_DIR}"
+cd "${PYBOX2D_DIR}"
+python setup.py build
+python setup.py install
+python setup.py develop
+cd "${GARAGE_DIR}"
+# We need a MuJoCo key to import mujoco_py
+cp ${_arg_mjkey} "${HOME}"/.mujoco/mjkey.txt
+brew reinstall gcc@7 --without-multilib
```
ryanjulian: Please place this with the other brew commands.

scripts/setup_osx.sh
```diff
@@ -1,38 +1,267 @@
...
+
+# Add garage to python modules
+export PYTHONPATH="$PYTHONPATH:${GARAGE_DIR}"
+PYTHON_ENV_VAR="PYTHONPATH=\"\$PYTHONPATH:${GARAGE_DIR}\""
+if [[ "${_arg_modify_bashprof}" = on ]]; then
+  echo "export ${PYTHON_ENV_VAR}" >> "${BASH_PROF}"
+else
+  echo -e "\nRemember to export the following environment variables before" \
+    "running garage:"
+  echo "${LD_LIB_ENV_VAR}"
```
ryanjulian: You may want to include the full commands, e.g. "export $MY_FOO_VAR". If a user needs your help with this message, they probably also don't know what `export` does.

scripts/setup_osx.sh
```diff
@@ -1,38 +1,267 @@
...
+  echo "${PATH_ENV_VAR}"
+  echo "${PYTHON_ENV_VAR}"
+  echo "You may wish to edit your .bash_profile to prepend the exports of" \
+    "these environment variables."
+fi
 
-echo "Conda environment created! Make sure to run \`source activate garage\`" \
-  "whenever you open a new terminal and want to run programs under garage."
+echo -e "\nGarage is installed! To make the changes take effect, work under" \
```
ryanjulian: garage is lower case

scripts/setup_osx.sh
ryanjulian: The problem with using PATH is that conda explicitly discourages it, to the point of displaying a nag message on every call. There is probably a reason for that. We also can't detect whether conda is installed anyway, we can only guess at the path (and it could be installed in many places).

gonzaiva: Got it, I will implement this approach  then.

scripts/setup_osx.sh
```diff
@@ -1,38 +1,267 @@
...
+  wget \
+  cmake \
+  boost \
+  boost-python \
+  ffmpeg \
+  sdl2 \
+  swig \
+  openmpi
+# Required by MuJoCo
+brew reinstall gcc@7 --without-multilib
```
ryanjulian: why reinstall? and why not just install it before all the brew dependencies (which might also need to be compiled using the older GCC).

Also, unless something needs a higher version, gcc6 is more broadly supported than gcc7, which is why I picked it.

mine worked with `brew install gcc@6 --without-multilib` before all the other brew deps

scripts/setup_osx.sh
```diff
@@ -1,38 +1,267 @@
...
+# Extras
+conda activate garage
+# Prevent pip from complaining about available upgrades
+pip install --upgrade pip
+# Fix Box2D install
+# See https://github.com/openai/gym/issues/100
+# See https://github.com/pybox2d/pybox2d/issues/82
+pip uninstall -y Box2D Box2D-kengz box2d-py
+PYBOX2D_DIR="$(mktemp -d)/pybox2d"
+git clone https://github.com/pybox2d/pybox2d "${PYBOX2D_DIR}"
```
ryanjulian: Did `pip install Box2D-kengz --no-binary Box2D-kengz` not work on your system?

Checking out and installing from source should be considered a last resort. Your checkout is using `master`, not a release tag or hash, so it could break without warning for no reason.

scripts/setup_osx.sh
```diff
@@ -1,38 +1,267 @@
...
+# Make sure that we're under the garage directory
+GARAGE_DIR="$(dirname $(script_dir_path))"
+cd "${GARAGE_DIR}"
+
+# File where environment variables are stored
+BASH_PROF="${HOME}/.bash_profile"
+
+# Install dependencies
+echo "Installing garage dependencies"
+echo "You will probably be asked for your sudo password"
```
ryanjulian: `sudo` does not appear in this file

scripts/setup_osx.sh
gonzaiva: I didn't try installing Box2D kengz, sorry for not noticing this one before. I will try it.

gonzaiva: It worked!

gonzaiva:
Scripts are working in fresh installations of Ubuntu (16.04) and OS X (10.12).

scripts/setup_osx.sh
```diff
@@ -1,38 +1,261 @@
...
+# Build mujoco-py and check is correctly installed
+python -c 'import mujoco_py'
+# Install git hooks
+pre-commit install -t pre-commit
+pre-commit install -t pre-push
+pre-commit install -t commit-msg
+# Install a virtualenv for the hooks
+pre-commit
+conda deactivate
+# Set personal configuration parameters
```
ryanjulian: This actually copies the template so that the user can set their own parameters

scripts/setup_linux.sh
```diff
@@ -215,46 +202,50 @@ if [[ "${?}" -ne 0 ]]; then
...
 
 # Extras
-source activate garage
+conda activate garage
 # Prevent pip from complaining about available upgrades
 pip install --upgrade pip
 # TensorFlow is not in environment.yml because of CPU vs GPU
 # Remove any TensorFlow installations before installing the desired flavor
 pip uninstall -y tensorflow tensorflow-gpu
 if [[ "${_arg_tf_gpu}" = on ]]; then
```
ryanjulian: You only need to remove the existing TFs if you are installing tensorflow-gpu

scripts/setup_linux.sh
```diff
@@ -215,46 +202,50 @@ if [[ "${?}" -ne 0 ]]; then
...
-  pip install tensorflow-gpu
+  pip install "tensorflow-gpu<1.10,>=1.9.0"
 else
-  pip install tensorflow
+  pip install "tensorflow<1.10,>=1.9.0"
 fi
 # Fix Box2D install
```
ryanjulian: Can you split each of these items into blocks with a newline? It's hard to tell which steps are related.

gonzaiva:
@ryanjulian I tested the installation on my Ubuntu 16.04 VM by running the scripts found under the script section in .travis.yml, particularly the tests under nose2. Also, I separately ran examples/theano/trpo_swimmer.py and examples/tf/trpo_cartpole.py, both with run_experiment, plot=True and parallel_sampler=10, and both worked well.

gonzaiva:
Okay,  I will create an issue to address the quiet option.

PR #295, opened 2018-08-17
Fix sleeping processes

gonzaiva:
The joblib package responsible of the MemmappingPool has been updated to
consider any bugs that could produce the sleeping processes in the
parallel sampler. Also the environment variable JOBLIB_START_METHOD has
been removed since it's not implemented by joblib anymore.
However, if run_experiment is interrupted during the optimization steps,
the sleeping processes are still produced. To fix the problem, the child
processes of the parallel sampler ignore SIGINT so they're not killed
while holding a lock that is also acquired by the parent process,
avoiding a dead lock.
To make sure the child processes are terminated, the SIGINT handler in
the parent process is overridden to call the terminate and join
functions in the processes pool.
The process (thread in TF) used in Plotter is terminated thanks to
registering the method shutdown with function atexit, but one important
step missing was to clean the Queue that interacts with worker process.


ryanjulian:
@jonashen did you test this on your machine?

jonashen:
Oops, good point. Sorry.

environment.yml
```diff
@@ -27,7 +27,7 @@ dependencies:
         - ipdb
         - ipywidgets
         - jsonmerge
-        - joblib==0.10.3
+        - joblib<0.13,>=0.12
```
jonashen:
```(garage) garage (fix_sleeping_proc *) $ pip install joblib<0.13,>=0.12
bash: 0.13,: No such file or directory```

File: garage/plotter/plotter.py [hunk 2]
```diff
@@ -84,6 +84,9 @@ def shutdown(self):
         if not Plotter.enable:
             return
         if self._process and self._process.is_alive():
+            while not self._queue.empty():
+                self._queue.get()
+                self._queue.task_done()
```
jonashen:
```
  File "/Users/jonathon/Documents/garage/garage/garage/plotter/plotter.py", line 89, in shutdown
    self._queue.task_done()
AttributeError: 'Queue' object has no attribute 'task_done'
```

scripts/run_experiment.py
```diff
@@ -2,9 +2,9 @@
 import ast
```
jonashen:
```
(garage) garage (fix_sleeping_proc *) $ python scripts/run_experiment.py
2018-08-17 09:45:25.883375 PDT | tensorboard data will be logged into:/Users/jonathon/Documents/garage/garage/data/experiment_2018_08_17_09_45_25_880884_PDT_95e7c
  File "scripts/run_experiment.py", line 201, in <module>
    run_experiment(sys.argv)
  File "scripts/run_experiment.py", line 187, in run_experiment
    data = pickle.loads(base64.b64decode(args.args_data))
  File "/anaconda2/envs/garage/lib/python3.6/base64.py", line 80, in b64decode
    s = _bytes_from_decode_data(s)
  File "/anaconda2/envs/garage/lib/python3.6/base64.py", line 46, in _bytes_from_decode_data
    "string, not %r" % s.__class__.__name__) from None
TypeError: argument should be a bytes-like object or ASCII string, not 'NoneType'
```

garage/tf/plotter/plotter.py
```diff
@@ -98,6 +98,9 @@ def _start_worker(self):
 
     def shutdown(self):
         if self.worker_thread.is_alive():
+            while not self.queue.empty():
+                self.queue.get()
+                self.queue.task_done()
```
jonashen:
```
  File "examples/tf/trpo_cartpole.py", line 25, in <module>
    algo.train()
  File "/Users/jonathon/Documents/garage/garage/garage/tf/algos/batch_polopt.py", line 146, in train
    self.shutdown_worker()
  File "/Users/jonathon/Documents/garage/garage/garage/tf/algos/batch_polopt.py", line 101, in shutdown_worker
    self.plotter.shutdown()
  File "/Users/jonathon/Documents/garage/garage/garage/tf/plotter/plotter.py", line 103, in shutdown
    self.queue.task_done()
  File "/anaconda2/envs/garage/lib/python3.6/queue.py", line 68, in task_done
    raise ValueError('task_done() called too many times')
ValueError: task_done() called too many times
Error in atexit._run_exitfuncs:
  File "/anaconda2/envs/garage/lib/python3.6/queue.py", line 68, in task_done
    raise ValueError('task_done() called too many times')
ValueError: task_done() called too many times
```

ryanjulian:
Maybe a good exercise is to thing about how to test this bug, since it's so severe and recurring.

You can use subprocess in Python to spawn a new process and send it signals (e.g. SIGINT). I think the standard library should also have the tools to crawl the spawned process tree and make sure everything terminates.

You can test the sampler by

* starting up "run_experiment" with subprocess and having it sample some environment.
* Once it is started, you crawl the process tree to find the PIDs of all it's child processes.
* Then you send SIGINT to the "run_experiment" parent process and wait some time
* Finally, verify all child PIDs are gone

Seeing as this is also a script to reliably *reproduce* this bug on all platforms, I propose that we write it and then just wrap it in a unit test so this doesn't happen again.

environment.yml
ryanjulian: try `pip install "joblib<0.13,>=0.12"`

scripts/run_experiment.py
ryanjulian: You forgot an argument. `run_experiment.py` is usually called by `garage.misc.instrument.run_experiment`

environment.yml
jonashen: it works

jonashen: Is there a reason we're specifying a version here? The installed version (0.12.2) is equivalent to if I simply called `pip install joblib`.

garage/plotter/plotter.py
zhanpenghe: Use [JoinableQueue](https://docs.python.org/3.6/library/multiprocessing.html#multiprocessing.JoinableQueue) instead of Queue. I think python 3.6 clean up the multiprocessing a little bit.

ryanjulian:
There is a nice tutorial here about using process groups to jail fork trees created by subprocess. It might be of use.
https://pymotw.com/3/subprocess/

`psutil` will also let you crawl the process tree.

garage/plotter/plotter.py
gonzaiva: Done

garage/tf/plotter/plotter.py
gonzaiva: I fixed this by moving some calls of task_done previously inserted to the right places. Right after the tasks are completed in the worker thread.

environment.yml
gonzaiva: We are being more careful with the major changes introduced in joblib, so we are restricting the updates of this library to only minor changes.

ryanjulian:
Please create a test fixture that can reproduce (and then prove fixed) this bug before we submit.

environment.yml
ryanjulian: The semantic versioning scheme used by many open source libraries:
XX.YY.ZZ

XX changes -- no backwards compatibility. literally anything can happen. whole packages can disappear.
YY changes -- backwards compatible for the same value of YY. features may be added but not removed. most increments to YY will be small changes but they may sometimes be backwards-incompatible, especially for libraries <1.0.
ZZ changes -- bug fixes/maintenance within a release only. generally no new features.

scripts/run_experiment.py
```diff
@@ -113,15 +113,25 @@ def run_experiment(argv):
 
     args = parser.parse_args(argv[1:])
 
-    assert (os.environ.get("JOBLIB_START_METHOD", None) == "forkserver")
     if args.seed is not None:
         set_seed(args.seed)
 
+    sigint_hdlr = signal.getsignal(signal.SIGINT)
+
+    def terminte_sampler(signum, frame):
```
ryanjulian: terminte --> terminate

scripts/run_experiment.py
```diff
@@ -113,15 +113,25 @@ def run_experiment(argv):
...
+        parallel_sampler.terminate()
+        parallel_sampler.join()
```
ryanjulian: what if parallel_sampler hands after terminate? can we join() with a timeout and take more drastic termination steps if it hangs?

scripts/run_experiment.py
```diff
@@ -113,15 +113,25 @@ def run_experiment(argv):
...
+        sigint_hdlr(signum, frame)
+
+    signal.signal(signal.SIGINT, terminte_sampler)
+
+    signal.pthread_sigmask(signal.SIG_BLOCK, [signal.SIGINT])
```
ryanjulian: can you make this a context manager?

```python
# SIGINT unblocked here
with mask_signal(signal.SIGINT):
    # SIGINT blocked in here
    # do uninterruptible stuff

# SIGINT unblocked again out here

File: scripts/run_experiment.py [hunk 8]
```diff
@@ -113,15 +113,25 @@ def run_experiment(argv):
 
     args = parser.parse_args(argv[1:])
 
-    assert (os.environ.get("JOBLIB_START_METHOD", None) == "forkserver")
     if args.seed is not None:
         set_seed(args.seed)
 
+    sigint_hdlr = signal.getsignal(signal.SIGINT)
```
ryanjulian: Please include a large block comment explaining why this is necessary and how it works.

It seems to me that it might be more appropriate to implement this inside stateful_pool if possible.

ryanjulian:
Can you verify that this bug https://bugs.python.org/issue8296 doesn't affect Python 3.5+? Otherwise we may need to change stateful_pool to only use _async mappers.

scripts/run_experiment.py
gonzaiva: Unfortunately the Pool class that MemmappingPool inherits from does not have a [join](https://docs.python.org/3.6/library/multiprocessing.html#multiprocessing.pool.Pool.join) method with timeout parameter.
From the Pool API, [close](https://docs.python.org/3.6/library/multiprocessing.html#multiprocessing.pool.Pool.close) is the nice way to ask to finish all tasks, and terminate just interrupts and finishes right away.

ryanjulian: Ok SGTM. We can test this and see how robust it is before going to greater lengths.

gonzaiva:
Regarding the bug above mentioned, I did an experiment with python 3.6.6 based on this [example](http://jessenoller.com/2009/01/08/multiprocessingpool-and-keyboardinterrupt/):
```
import multiprocessing    
import time    
import os    
     
def create():    
    try:    
        time.sleep(3)     
    except KeyboardInterrupt:    
        print("Exiting child gracefully", os.getpid())        
        return                                                
    return "Finishing child creation %s"%(os.getpid())    
               
def main():          
    def cb(what):                   
        print("Callback:", what)    
                                    
    print("Parent", os.getpid())      
    pool = multiprocessing.Pool(2)    
    try:                      
        for i in range(2):                                    
            pool.apply_async(create, args=(), callback=cb)      
        print("Initialization of child processes requested")    
        pool.close()                                                       
        print("Child processes were requested to finish")        
        pool.join()                        
        print("Child processes joined")    
    except KeyboardInterrupt:                 
        print("Keyboard Interrupt caught")    
        pool.terminate()            
        print("Pool terminated")    
                              
if __name__ == "__main__":    
    main()                                                                     
```

Without interrupting the execution, the outcome is:
```
Parent 2578
Initialization of child processes requested
Child processes were requested to finish
Callback: Finishing child creation 2584
Callback: Finishing child creation 2585
Child processes joined
```

Interrupting the execution:
```
Parent 3070
Initialization of child processes requested
Child processes were requested to finish
^CKeyboard Interrupt caught
Exiting child gracefully 3072
Exiting child gracefully 3071
Pool terminated
```

Even though the child processes catch the KeyboardInterruption, they become joinable and the parent process exits without leaving any zombie processes hanging around.

ryanjulian:
Alright. To merge this still need a test which reproduces the bug (before the change) and verifies it's gone after.

gonzaiva:
I've been thinking in a way to perform the tests for this change. My algorithm would be the following:
1. Create a subprocess to run an example file using parallel sampler and plotter.
2. Once the parallel sampler and plotter are up an running, assert the number of child processes based on the configuration of the example file.
3. After a random time, send SIGINT to the subprocess in step one.
4. Wait for all the child process to finish, and assert that the number of child processes are zero.

To make sure we're trying this test in different execution points, we can run the above sequence for a certain number of iterations, but please let me know what could be improved.

One of the things I'm not so sure about is how to detect that the child processes under the parallel sampler are running. How I'm currently doing this in my sandbox is reading stdout and catching the string "Populated", which is written once all the child processes finished their initialization. Please let me know if you have a better idea to do this.

ryanjulian:
Some thoughts on this plan:

* Tests should be deterministic so that they can reliably detect bugs when run once. Choosing a random time point to stop the process will automatically make your test flaky.
* You should always test your desired outcome directly if at all possible. Our desired outcome is that there are 0 child processes for run_experiment after it receives a SIGINT. This is easiest to verify IMO if we enumerate all of the child processes when it launches and then verify they are all gone when it terminates. It also means if you detect an orphan you can tell the user which it is. The number and sources of child processes generated by run_experiment might change as the code evolves, but we will always want all of them to terminate on SIGINT.
* There are actually several test cases here. Don't try to blur them with a giant integration test. At a minimum, we need to verify successful termination at several points in the run_experiment lifecycle (setup, sampling, optimizing, shutdown).
* Don't use pipe hacks and other things which read stdout to figure out the process tree. We can change the fact that new processes print "populated" at any time and your test would failing silently. The plotter process doesn't even print this currently. Unix and Python provide you with ample tools to enumerate the process tree created by your subprocess, and I have provided pointers to them. Use the right tool for the job and you will only  ever have to write this once.

scripts/run_experiment.py
gonzaiva: Actually calling join here is wrong since the signal handler is called by the Python main thread and join has an assert call to make sure the parent process is trying to join.
Also, I found some handling that I did for this in a previous fix [here](https://github.com/rlworkgroup/garage/blob/a3f6fe4e2cc030b0b58d9dd403ba9a04c04dce7f/scripts/run_experiment.py#L174). I will add the join call there instead of the overriding the signal handler.

gonzaiva:
I added the test, but now I'm wondering about three things:
1) Add a message to the user when a forced finalization is produced. Currently only a forced finalization is executed since the join call for Pool does not have a timeout parameter that would wake up the process if the nice finalization never returns. Maybe we could use a polling approach instead by checking on each process in the pool instead of calling join, and sleeping in relative small intervals to avoid hogging the CPU.
2) Add more malicious cases. One case I have in mind is sending SIGINT not to the launcher but to any of its children and check all of them die.
3) Right now I added the test considering Theano, but I will also add the same for TF to make sure there's no errors on that side.

gonzaiva:
It seems my test is not behaving the same way as in my working area. As shown at the bottom of these [logs](https://travis-ci.com/rlworkgroup/garage/builds/82714356), the tests hangs.
It seems that SIGINT is propagated to all children even if the signal is masked when they're created, since there's a KeyboardInterrupt traceback for all processes, but I am not really sure in which call of the test it's getting stuck, since the call that waits for the children processes to die has a [timeout](https://github.com/rlworkgroup/garage/blob/b17dde53466b3a56c29c77060c0feef9a013946f/tests/theano/test_exp_interrupt/test_exp_interruption.py#L49).

gonzaiva:
The test is working now. The root cause of why it wasn't working in docker is that docker runs the test script with the init process (PID 1), that is, it doesn't create a subprocess as it happens when running the script in the shell.
Since the init process is the parent of all processes, the algorithm in the test was interrupting all other processes in the container, which leads to unexpected results.
The algorithm has been changed to ask for the children of the launcher, not of the test script, and adding the launcher to the list of children when waiting for all the generated processes in the test to die.
Amongst other things, the pre-execution call to run the launcher was os.setsid to create a session, but this may not work as well to create a group inside the docker container, so the pre-execution call was changed to os.setpgrp instead.

scripts/run_experiment.py
```diff
@@ -113,15 +115,24 @@ def run_experiment(argv):
...
         set_seed(args.seed)
 
-    if args.n_parallel > 0:
-        from garage.sampler import parallel_sampler
-        parallel_sampler.initialize(n_parallel=args.n_parallel)
-        if args.seed is not None:
-            parallel_sampler.set_seed(args.seed)
+    # SIGINT is blocked for all processes created in parallel_sampler to avoid
+    # the creation of sleeping and zombie processes.
+    # If the user interrupts run_experiment, there's a chance some processes
```
ryanjulian: Please add a newline between paragraphs

ryanjulian:
Great work! I suspect we might still find ways of leaving zombies after this change.

Let's merge this and have people start using it. If we find new zombies, we can amend the tests until we catch all corner cases.

gonzaiva:
Okay, I just added some comments in the test, but I will merge after that. Also, I want to make sure that Travis won't break for this test, so I may rebuild three times to make sure.

tests/theano/test_exp_interrupt/trpo_cartpole.py
```diff
@@ -0,0 +1,39 @@
+from garage.baselines import LinearFeatureBaseline
```
ryanjulian: please put this in `tests/fixtures/tf/trpo_cartpole_instrumented.py`

ryanjulian:
Does `parallel_sampler.terminate()` block? If not, I think the best option here is to busy wait for all the children to disappear, the either `join()` once they have disappeared, or exit with an error if they don't disappear within some timeout.

gonzaiva:
Okay, I have added the corresponding calls to shutdown the parallel sampler and plotters in run_experiment. However, when run_experiment runs under test_sigint_theano, a bug happens in some tests, where some sleeping processes remain and the corresponding user warning is produced. The output comes with a traceback that points to the fork of the process that is remaining:
```
2018-08-28 23:09:36.891154 PDT | Perplexity                 4.13273
2018-08-28 23:09:36.891207 PDT | StdReturn                 65.6678
2018-08-28 23:09:36.891261 PDT | dLoss                      0.0341225
2018-08-28 23:09:36.891313 PDT | -----------------------  -------------
  File "tests/fixtures/theano/trpo_cartpole_instrumented.py", line 38, in <module>
    plot=True,
  File "/home/aigonzal/ivanWorkspace/garage/garage/misc/instrument.py", line 524, in run_experiment
    command, shell=True, env=dict(os.environ, **env))
  File "/home/aigonzal/miniconda2/envs/garage/lib/python3.6/subprocess.py", line 269, in call
    return p.wait(timeout=timeout)
  File "/home/aigonzal/miniconda2/envs/garage/lib/python3.6/subprocess.py", line 1457, in wait
    (pid, sts) = self._try_wait(0)
  File "/home/aigonzal/miniconda2/envs/garage/lib/python3.6/subprocess.py", line 1404, in _try_wait
    (pid, sts) = os.waitpid(self.pid, wait_flags)
KeyboardInterrupt
The following processes didn't die after the shutdown of run_experiment:
{'status': 'sleeping', 'name': 'python', 'pid': 12288}
This is a sign of an unclean shutdown. Please reopen the following issue 
 with a detailed description of how the error was produced:
https://github.com/rlworkgroup/garage/issues/120
  File "/home/aigonzal/ivanWorkspace/garage/scripts/run_experiment.py", line 235, in <module>
    run_experiment(sys.argv)
  File "/home/aigonzal/ivanWorkspace/garage/scripts/run_experiment.py", line 185, in run_experiment
    method_call(variant_data)
  File "tests/fixtures/theano/trpo_cartpole_instrumented.py", line 26, in run_task
    algo.train()
  File "/home/aigonzal/ivanWorkspace/garage/tests/fixtures/theano/batch_polopt_instrumented.py", line 35, in train
    paths = self.sampler.obtain_samples(itr)
  File "/home/aigonzal/ivanWorkspace/garage/garage/algos/batch_polopt.py", line 28, in obtain_samples
    scope=self.algo.scope,
  File "/home/aigonzal/ivanWorkspace/garage/garage/sampler/parallel_sampler.py", line 130, in sample_paths
    show_prog_bar=True)
  File "/home/aigonzal/ivanWorkspace/garage/garage/sampler/stateful_pool.py", line 162, in run_collect
    manager = mp.Manager()
  File "/home/aigonzal/miniconda2/envs/garage/lib/python3.6/multiprocessing/context.py", line 56, in Manager
    m.start()
  File "/home/aigonzal/miniconda2/envs/garage/lib/python3.6/multiprocessing/managers.py", line 513, in start
    self._process.start()
  File "/home/aigonzal/miniconda2/envs/garage/lib/python3.6/multiprocessing/process.py", line 105, in start
    self._popen = self._Popen(self)
  File "/home/aigonzal/miniconda2/envs/garage/lib/python3.6/multiprocessing/context.py", line 277, in _Popen
    return Popen(process_obj)
  File "/home/aigonzal/miniconda2/envs/garage/lib/python3.6/multiprocessing/popen_fork.py", line 19, in __init__
    self._launch(process_obj)
  File "/home/aigonzal/miniconda2/envs/garage/lib/python3.6/multiprocessing/popen_fork.py", line 66, in _launch
    self.pid = os.fork()
KeyboardInterrupt
```
It seems the problem is related to the use of psutil.wait_procs in both run_experiment and test_sigint_theano at the same time, but I need to look further into this.
This bug does not happen when a launcher is used.

garage/plotter/plotter.py
```diff
@@ -21,13 +21,16 @@ class Op(Enum):
 
 Message = namedtuple("Message", ["op", "args", "kwargs"])
 
+__plotters__ = []
```
ryanjulian: This can be a class member of `Plotter` rather than package-global.

See https://stackoverflow.com/a/12102666 for a useful pattern.

scripts/run_experiment.py
```diff
@@ -186,5 +202,34 @@ def run_experiment(argv):
     logger.pop_prefix()
 
 
+def child_proc_shutdown(shutdown_sampler=False):
```
ryanjulian: If you change the Plotter API to use `close()` (or the sampler API to use `shutdown()`) then this could be called as

```python
def child_proc_shutdown(children):
    run_exp_proc = psutil.Process()
    alive = run_exp_proc.children(recursive=True)
    for c in children:
        c.shutdown()
    # etc...

# example at the callsite
child_proc_shutdown(__plotters__ + [parallel_sampler])
```

scripts/travisci/check_flake8.sh
```diff
@@ -2,7 +2,7 @@
 
 # Python packages considered local to the project, for the purposes of import
 # order checking. Comma-delimited.
-garage_packages="garage,sandbox,examples,tests"
+garage_packages="tests,garage,sandbox,examples,contrib"
```
ryanjulian: contrib was removed for a reason--it not longer exists

setup.cfg
```diff
@@ -4,7 +4,7 @@
 # style, this rule is ignored.
 ignore = W503
 import-order-style = google
-application-import-names = sandbox,garage,examples,tests
+application-import-names = tests,sandbox,garage,examples,contrib
```
ryanjulian: Remove contrib

tests/fixtures/theano/__init__.py
```diff
@@ -0,0 +1,4 @@
+from tests.fixtures.theano.batch_polopt_instrumented import \
```
ryanjulian: PEP8 says to use () to split lines instead of \

tests/fixtures/theano/__init__.py
```diff
@@ -0,0 +1,4 @@
...
+        InstrumentedBatchPolopt
+from tests.fixtures.theano.npo_instrumented import InstrumentedNPO
```
ryanjulian: Package names should be `instrumented_batchpolopt`, `instrumented_npo` and `instrumented_trpo` to match class names.

tests/sampler/test_sigint_theano.py
```diff
@@ -0,0 +1,81 @@
+from enum import Enum
```
ryanjulian: Please put this file in `tests/integration_tests`

tests/sampler/test_sigint_theano.py
```diff
@@ -0,0 +1,81 @@
...
+from multiprocessing.connection import Listener
+import os
```
ryanjulian: Can you add an equivalent for TensorFlow?

tests/sampler/test_sigint_theano.py
```diff
@@ -0,0 +1,81 @@
...
+        os.kill(child.pid, signal.SIGINT)
+
+    if not clean_exit:
+        raise AssertionError(colorize(error_msg, "red"))
+
+
+class TestSigInt(unittest.TestCase):
+    def test_sigint(self):
+        """Interrupt the experiment in different stages of its lifecyle."""
+        for stage in list(ExpLifecycle):
```
ryanjulian: use a parameterized test instead of for loop https://nose2.readthedocs.io/en/latest/params.html

tests/sampler/test_sigint_theano.py
```diff
@@ -0,0 +1,81 @@
...
+        clean_exit = False
+
+    error_msg = ("These processes didn't die during %s:\n" % (lifecycle_stage)
+                 + error_msg)
+
+    for child in alive:
+        os.kill(child.pid, signal.SIGINT)
+
+    if not clean_exit:
+        raise AssertionError(colorize(error_msg, "red"))
```
ryanjulian: Don't raise exceptions in tests. Just use `assert`.

ryanjulian:
NICE

naeioi:
It there an corresponding issue to demonstrate what this PR is trying to fix?

PR #299, opened 2018-08-17
Fixed DmControlEnv

jonashen:
Ref: #294

One thing to note is after running `nose2 tests.test_dm_control_env`, `dm_control` throws a GLFW error.
However, it looks to be a minor issue that can be [ignored](https://github.com/deepmind/dm_control/issues/32#issuecomment-381205808).
```
Ran 1 test in 5.551s

OK
Exception ignored in: <bound method ContextBase.__del__ of <dm_control.render.glfw_renderer.GLFWContext object at 0x125792198>>
  File "/anaconda2/envs/garage/lib/python3.6/site-packages/dm_control/render/base.py", line 234, in __del__
  File "/anaconda2/envs/garage/lib/python3.6/site-packages/dm_control/render/base.py", line 230, in free
  File "/anaconda2/envs/garage/lib/python3.6/site-packages/dm_control/render/base.py", line 95, in release_context
  File "/anaconda2/envs/garage/lib/python3.6/site-packages/dm_control/render/base.py", line 153, in release_context
  File "/anaconda2/envs/garage/lib/python3.6/site-packages/dm_control/render/base.py", line 158, in _active_context
TypeError: getattr(): attribute name must be string
```


ryanjulian:
This always happens when `dm_control` is destructed. Not sure why, but I think it can be ignored. We might be able to mask is by overloading __del__ in `DmControlEnv`

garage/envs/dm_control/dm_control_env.py
```diff
@@ -39,13 +39,14 @@ def step(self, action):
         if time_step.reward:
             self._total_reward += time_step.reward
         return Step(
-            flatten_observation(time_step.observation), time_step.reward,
-            time_step.step_type == StepType.LAST, **time_step.observation)
+            flatten_observation(time_step.observation)['observations'],
```
gonzaiva: Although it's not enforced by the lint tools since the legacy code has a lot of mixed cases, better use double quotes with strings.

gonzaiva: Although it's not enforced by the lint tools since the legacy code has a lot of mixed cases, better use double quotes with strings.

tests/test_dm_control_env.py
```diff
@@ -0,0 +1,31 @@
...
+from garage.baselines import LinearFeatureBaseline
+from garage.envs.dm_control import DmControlEnv
+from garage.tf.algos import TRPO
+from garage.tf.envs import TfEnv
+from garage.tf.policies import GaussianMLPPolicy
+
+
+class TestDmControlEnv(unittest.TestCase):
+    def test_dm_control_env(self):
+        env = TfEnv(DmControlEnv(domain_name='cartpole', task_name='balance'))
```
ryanjulian: This can be one test case (algo integration test)

Can you also add parameterized tests (which cover every dm_control env) like the ones I wrote for gym?
* Test pickling/unpickling
* Test each env can be wrapped, constructed, stepped, rendered, and closed.
* Test the types observation_space, action_space, and observations produced by step/reset for the wrapped version.

ryanjulian:
This needs a test before it can be submitted.

The test can replace the DmControlEnv tests in test_dm_control.py. The test should verify that every dm_control environment can be constructed and stepped. It should also have a separate case which verifies the same, but for an environment which is constructed, pickled, and stepped.

You can use a parameterized test and the symbol ALL_TASKS found here https://github.com/deepmind/dm_control/blob/master/dm_control/suite/__init__.py

tests/dm_control/test_dm_control_envs.py
```diff
@@ -0,0 +1,64 @@
...
+                batch_size=100,
+                max_path_length=10,
+                n_itr=1,
+                discount=0.99,
+                step_size=0.01,
+            )
+            algo.train()
+
+
+# class TestEnvs(unittest.TestCase):
```
zhanpenghe: Please remove commented out codes.

jonashen:
Here is the situation: the reason why these tests (specifically `test_dm_control_policies`) are OOM on the CI is due to the need to create a completely new TensorFlow graph each time we generate a new dm_control environment. 

*Why don’t we use the same graph?*
We cannot reuse the same graph (e.g. `tf.get_default_graph()`) because each environment has different node specifications (e.g. hidden_sizes), and thus the graph generated from one environment is likely incompatible for the next environment. 

Even if we could reuse the graph, however, we would need extensive code refactoring to enable reusing nodes in all variable scopes to prevent this error: `ValueError: Variable <weight_node> already exists, disallowed. Did you mean to set reuse=True in VarScope?`. See [here](https://stackoverflow.com/a/46894945) for some reasons why this error arises.

Furthermore, we cannot call `tf.reset_default_graph()` due to a nesting error: “Do not use tf.reset_default_graph() to clear nested graphs. If you need a cleared graph, exit the nesting and create a new graph”. 

However, even if we could call `tf.reset_default_graph()`, memory isn’t being [freed](https://github.com/tensorflow/tensorflow/blob/r1.11/tensorflow/python/framework/ops.py#L5017) by TensorFlow, and it actually generates a new [graph](https://stackoverflow.com/a/39353329) and keeps the same Tensors. This is also the reason why we can’t test Box environments alongside Discrete environments (`ValueError: Tensor("CategoricalMLPPolicy/Actions:0", shape=(), dtype=string) must be from the same graph as Tensor("GaussianMLPPolicy/Actions:0", shape=(), dtype=string)`). I also was unable to find a way to manually delete the individual nodes in the default graph.

All of this culminates into the reason why we need to create completely new TensorFlow graphs each time. That being said, due to TensorFlow’s nonexistent memory cleanup, we need to explicitly call `del graph` and python’s internal garbage collection `gc.collect()` after each graph creation. As you can see from the images, there is a 33% decrease in memory usage from these calls. However, this is still inadequate because the graphs are still increasing, i.e. some TensorFlow objects are still being kept in memory.  Here is proof of TensorFlow eating up all memory usage using [tracemalloc](https://docs.python.org/3/library/tracemalloc.html):
```
Top 10 lines
#1: framework/ops.py:3008: 62440.1 KiB
   frame_info))
#2: framework/ops.py:3000: 7543.9 KiB
   lineno = f.f_lineno
#3: framework/ops.py:3005: 7468.6 KiB
   func_start_lineno = co.co_firstlineno
#4: framework/ops.py:1876: 3163.3 KiB
   return c_api.TF_OperationName(self._c_op)
#5: framework/ops.py:4179: 2980.3 KiB
   name_key = name.lower()
#6: wrapper/util.py:195: 2186.8 KiB
   buf = np.frombuffer(ptr.contents, dtype=ctype)
#7: framework/ops.py:1691: 2070.2 KiB
   self._graph = g
#8: collections/__init__.py:423: 1990.7 KiB
   for index, name in enumerate(field_names))
#9: collections/__init__.py:429: 1987.9 KiB
   exec(class_definition, namespace)
#10: python/pywrap_tensorflow_internal.py:44: 1970.6 KiB
   self.__dict__[name] = value
1633 other: 42258.7 KiB
Total allocated size: 136061.1 KiB
```

**Conclusion**
Please let me know if I have missed anything or if anything is unclear. In short, I do not see any other solution but to limit the number of dm_control environments we test to a half or a third. TensorFlow does not free up its graph memory, nor does it provide an interface to reuse the default graph by supporting an API to adjust the properties of the nodes.

ryanjulian:
@krzentner thoughts?

ryanjulian:
This behavior from TF is concerning when thinking about future test suites, but I think we can probably get around it for the purposes of these tests.

I don't think it's necessary to run an integration test which steps *every* dm_control env with an algorithm. It should be sufficient to validate that our DmControlEnv wrapper makes every dm_control env satisfy the gym.Env interface. That should require no TF session at all.

tests/dm_control/test_dm_control.py
```diff
@@ -0,0 +1,38 @@
...
+            task_name=task_name,
+            plot=True,
+            width=600,
+            height=400),
+        normalize_obs=False,
+        normalize_reward=False)
+
+    time_step = dm_control_env.reset()
+    action_spec = dm_control_env.action_space
+    for _ in range(5):
```
ryanjulian: Isn't there a helper which does exactly this stepping test in Angel's PR?

tests/dm_control/test_dm_control.py
```diff
@@ -0,0 +1,38 @@
...
+
+    dm_control_env.close()
+
+
+class TestDmControl(unittest.TestCase):
+    def test_dm_control(self):
+        for domain, task in suite.ALL_TASKS:
+            run_task(domain, task)
+
+        print("Congratulation! All tasks are done!")
```
ryanjulian: This is unnecessary logspam. Please remove.

tests/dm_control/test_dm_control.py
```diff
@@ -0,0 +1,38 @@
...
+            height=400),
+        normalize_obs=False,
+        normalize_reward=False)
+
+    time_step = dm_control_env.reset()
+    action_spec = dm_control_env.action_space
+    for _ in range(5):
+        dm_control_env.render()
+        action = action_spec.sample()
+        next_obs, reward, done, info = dm_control_env.step(action)
```
ryanjulian: This test doesn't assert anything and is therefore not really a test.

tests/dm_control/test_dm_control_policies.py
```diff
@@ -0,0 +1,51 @@
...
+
+from garage.baselines import LinearFeatureBaseline
+from garage.envs.dm_control import DmControlEnv
+from garage.tf.algos import TRPO
+from garage.tf.envs import TfEnv
+from garage.tf.policies import GaussianMLPPolicy
+from garage.tf.spaces import Box
+
+
+class TestDmControlPolicies(unittest.TestCase):
```
ryanjulian: This test is unnecessary. We don't need to run an integration test with every dm_control environment. We should perhaps run one of these.

tests/dm_control/test_dm_control_policy.py
```diff
@@ -0,0 +1,57 @@
...
+from garage.tf.spaces import Box
+from tests.fixtures import GarageTestCase
+
+all_tasks_list = []
+for task in ALL_TASKS[0:1]:
+    all_tasks_list.append([task[0], task[1]])
+
+
+class TestDmControlPolicy(GarageTestCase):
+    @tools.params(*all_tasks_list)
```
ryanjulian: This many integration tests is unnecessary.

tests/fixtures/fixtures.py
```diff
@@ -8,9 +8,13 @@
 
 class GarageTestCase(unittest.TestCase):
```
ryanjulian: I think we should make clear that this is a TfTestCase. There is more than one backend in garage.

jonashen: sgtm

tests/dm_control/test_dm_control_policy.py
jonashen: This only runs one dm_control environment.

tests/dm_control/test_dm_control_policies.py
jonashen: So I should just keep the implementation of `test_dm_control_policy`? They are functionally equivalent.

tests/dm_control/test_dm_control.py
jonashen: I'll remove this test then

tests/dm_control/test_dm_control.py
jonashen: sgtm

tests/dm_control/test_dm_control.py
jonashen: Yes, I will use his implementation or remove this file altogether.

tests/dm_control/test_dm_control_policy.py
ryanjulian: `@tools.params(*all_task_lists)` implies that this runs a test case for every element of `all_tasks_list`. Are you saying `all_tasks_list` has only a single element?

tests/dm_control/test_dm_control.py
ryanjulian: I meant that you should assert some things here about the expected behavior

tests/dm_control/test_dm_control_policy.py
jonashen: No, I was going to replace this code with all environments. I simply limited it to a single element for the purposes of the CI.

tests/dm_control/test_dm_control_policies.py
ryanjulian: I think it's sufficient to test one dm_control environment against one algorithm from each backend.

tests/dm_control/test_dm_control.py
jonashen: i see

tests/dm_control/test_dm_control_policy.py
ryanjulian: The CI memory problems occur when testing a single environment against a single algorithm? (!)

jonashen: No, it's due to the other file that's testing all dm_control environments.

jonashen: Basically I'm making sure that both implementations are running in the CI, but since we don't need to test all dm_control environments, some renaming / refactoring is in order.

tests/dm_control/test_dm_control_policy.py
```diff
@@ -0,0 +1,57 @@
...
+from garage.baselines import LinearFeatureBaseline
+from garage.envs.dm_control import DmControlEnv
+from garage.tf.algos import TRPO
+from garage.tf.envs import TfEnv
+from garage.tf.policies import CategoricalMLPPolicy, GaussianMLPPolicy
+from garage.tf.spaces import Box
+from tests.fixtures import GarageTestCase
+
+all_tasks_list = []
+for task in ALL_TASKS[0:1]:
```
ryanjulian: What does this do exactly? Why can't we pass ALL_TASKS directly to `@tools.params`?

tests/dm_control/test_dm_control_policy.py
ryanjulian: ...but this test is virtually identical to the other one. Does that mean that the memory leak happens with GarageTestCase but not with your custom test here?

tests/dm_control/test_dm_control_policy.py
jonashen: Because it's in tuple form, I wasn't able to pass it directly as ALL_TASKS

jonashen: don't remember the exact error though

tests/dm_control/test_dm_control_policy.py
jonashen: Both are experiencing memory leaks, and it's due to the same core problem of needing to create new TensorFlow graphs.

tests/dm_control/test_dm_control_envs.py
```diff
@@ -0,0 +1,32 @@
...
+
+from dm_control.suite import ALL_TASKS
+import numpy as np
+
+from garage.envs.dm_control import DmControlEnv
+
+
+class TestDmControlEnvs(unittest.TestCase):
+    def test_dm_control_envs(self):
+        for task in ALL_TASKS:
```
ryanjulian: Please make this a parameterized test

tests/dm_control/test_dm_control_envs.py
```diff
@@ -0,0 +1,32 @@
...
+            env = DmControlEnv(domain_name=task[0], task_name=task[1])
+
+            # Test pickling
+            pickle.loads(pickle.dumps(env))
+
+            print("Testing", env.__class__)
```
ryanjulian: This is logspam. Please remove.

tests/dm_control/test_dm_control_envs.py
```diff
@@ -0,0 +1,32 @@
...
+            ob_space = env.observation_space
+            act_space = env.action_space
+            ob = env.reset()
```
ryanjulian: Can you use the helper function we already have for this? Feel free to add your own asserts.

tests/dm_control/test_dm_control_tf_policy.py
```diff
@@ -0,0 +1,36 @@
...
+                env=env,
+                policy=policy,
+                baseline=baseline,
+                batch_size=10,
+                max_path_length=5,
+                n_itr=1,
+                discount=0.99,
+                step_size=0.01,
+            )
+            print("Testing", env.__class__)
```
ryanjulian: This is logspam. Please remove.

tests/fixtures/__init__.py
```diff
@@ -1 +1 @@
-from tests.fixtures.fixtures import GarageTestCase  # noqa: F401
+from tests.fixtures.fixtures import TfTestCase  # noqa: F401
```
ryanjulian: Please remove the `# noqa: F401` and add an `__all__` to avoid this error.

tests/dm_control/test_dm_control_envs.py
```diff
@@ -0,0 +1,32 @@
+import pickle
+import unittest
+
+from dm_control.suite import ALL_TASKS
+import numpy as np
+
+from garage.envs.dm_control import DmControlEnv
+
+
+class TestDmControlEnvs(unittest.TestCase):
```
ryanjulian: A better way to construct this test is as follows:

Make the TestCase (not the method) parameterized. In setup, create the environment. Then add test methods for vanilla stepping `test_can_step_and_render()` and pickling `test_pickling()`.

This way, if the pickling test fails, the error will point right to the pickling test, not into the middle of a large test.

jonashen:
This PR is waiting on #318

Issue #298, opened 2018-08-17
CI only running 19 tests


ryanjulian:
My local suite runs 54 tests. Is something still guarded on the CI?

https://travis-ci.com/rlworkgroup/garage/builds/82191539
https://travis-ci.com/rlworkgroup/garage/builds/82168745

jonashen:
Some CIs are running 19, others 54. Not sure yet what the source of the deviation is.

jonashen:
I'm fairly certain it's because of this: `ImportError: cannot import name 'MemmapingPool'`.

This is probably why some tests aren't running, e.g. `AttributeError: module 'tests.theano.algos' has no attribute 'test_algos'`.

ryanjulian:
That only occurs when you upgrade joblib to >0.10, but the environment.yml on master is using joblib 0.10 and we have not committed Angel's change to upgrade.

Where did you see these errors?

ryanjulian:
`master` is still using joblib 0.10.3, but its builds are still only running 19 tests (see https://travis-ci.com/rlworkgroup/garage/builds/82260970)

It appears that the regression was introduced by #270 
`master` build before #270 https://travis-ci.com/rlworkgroup/garage/builds/81870318
`master` build after #270 https://travis-ci.com/rlworkgroup/garage/builds/81941351

Since you introduced the regression which wiped out >50% of our test suite, can you please fix it ASAP?

ryanjulian:
I suspect this might be because nose2 doesn't recognize test packages withoout `__init__.py` files, despite the fact that Python 3 was released a decade ago.

ryanjulian:
Reopened until comments on #311 are addressed in a new commit

PR #311, opened 2018-08-22
Fix Travis tests

jonashen:
This PR brings the total number of tests running in Travis from 19 to 93. Also fixes some tests that were previously broken but left undiscovered due to them being disabled.

Ref: #298

### DIFF
--- a/garage/envs/mujoco/gather/gather_env.py
+++ b/garage/envs/mujoco/gather/gather_env.py
@@ -33,7 +33,7 @@ PyMjrRect = cymj.PyMjrRect
 class GatherViewer(MjViewer):
     def __init__(self, env):
         self.env = env
-        super(GatherViewer, self).__init__(self.env.wrapped_env.sim)
+        super(GatherViewer, self).__init__(self.env.env.sim)
         green_ball_model = load_model_from_path(
             osp.abspath(osp.join(MODEL_DIR, 'green_ball.xml')))
         self.green_ball_renderer = EmbeddedViewer()
@@ -66,7 +66,7 @@ class GatherViewer(MjViewer):
 
     def render(self):
         super(GatherViewer, self).render()
-        ctx = MjRenderContext(self.env.wrapped_env.sim)
+        ctx = MjRenderContext(self.env.env.sim)
         scn = ctx.scn
         con = ctx.con
         functions.mjv_makeScene(scn, 1000)
@@ -97,9 +97,8 @@ class GatherViewer(MjViewer):
                                        self.red_ball_ctx.vopt,
                                        self.red_ball_ctx.pert, CAT_ALL, scn)
 
-        functions.mjv_addGeoms(self.env.wrapped_env.sim.model,
-                               self.env.wrapped_env.sim.data, ctx.vopt,
-                               ctx.pert, CAT_ALL, scn)
+        functions.mjv_addGeoms(self.env.env.sim.model, self.env.env.sim.data,
+                               ctx.vopt, ctx.pert, CAT_ALL, scn)
         functions.mjr_render(self.green_ball_renderer.get_rect(), scn, con)
 
         try:
@@ -252,6 +251,11 @@ class GatherEnv(gym.Wrapper, Serializable):
         # pylint: enable=not-callable
         super().__init__(inner_env)
 
+        # Redefine observation space
+        shp = self.get_current_obs().shape
+        ub = BIG * np.ones(shp)
+        self.observation_space = gym.spaces.Box(ub * -1, ub, dtype=np.float32)
+
     def reset(self, also_wrapped=True):
         self.objects = []
         existing = set()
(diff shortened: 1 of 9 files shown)


tests/tf/policies/test_policies.py
```diff
@@ -19,6 +22,11 @@
 
 
 class TestTfPolicies(unittest.TestCase):
+    def setUp(self):
+        self.sess = tf.Session(graph=tf.Graph())
+        self.sess.__enter__()
```
zhanpenghe: Please avoid using sess.__enter__()

CatherineSue: Why?

zhanpenghe: There are nicer ways to handle tensorflow session lifecycle. `__enter__` and `__exit__` are mostly used for the `with` statement. Using `__enter__`, especially without further releasing the resources, is not a very clean way to use tensorflow sessions(even though we used it before). 

You can look up [with statement](https://docs.python.org/3/reference/compound_stmts.html#the-with-statement) and [tf.Session](https://github.com/tensorflow/tensorflow/blob/r1.10/tensorflow/python/client/session.py)

ryanjulian: ZP is correct in general but I think this is exactly actually what you want in a unit test. 

Rules are often bent in unit tests to make the tests easier to write and read.

A good way to deal with this is to add a custom class (inherit from TestCase) which overloads setUp/tearDown to always make a clean session and enter. I think TF already has one of these in the tf.tests, but unsure if it fits our use case.

zhanpenghe: I think we should at least add an `__exit__` in tearDown.

CatherineSue: I've already tried tf.tests.TestCase.test_session() before but not work.
I agree we should add sess.close in tearDown.

jonashen: Done.

garage/envs/mujoco/gather/gather_env.py
```diff
@@ -438,14 +430,12 @@ def get_ori(self):
         successful, falls back to the default based on the ORI_IND specified in
         Maze (not accurate for quaternions)
         """
-        obj = self.wrapped_env
-        while not hasattr(obj, 'get_ori') and hasattr(obj, 'wrapped_env'):
```
ryanjulian: uhh does this still implement the behavior in the comment? i know it's a gross implementation, but we shouldn't break functionality to aesthetics.

tests/envs/test_envs.py
```diff
@@ -80,13 +78,7 @@ def test_env(self, env):
         a = act_space.sample()
         assert act_space.contains(a)
         res = env.step(a)
-        _, _, _, _ = env.step(a)
         assert ob_space.contains(res[0])  # res[0] --> observation
         assert np.isscalar(res[1])  # res[1] --> reward
         env.render()
-        if all(not isinstance(env, T) for T in [GridWorldEnv, Box2DEnv]):
```
ryanjulian: please leave this here. it verifies that off-screen rendering is working.

tests/tf/algos/test_trpo.py
```diff
@@ -22,6 +22,9 @@ def setUp(self):
         self.sess.__enter__()
         logger._tensorboard = TensorBoardOutput()
 
+    def tearDown(self):
```
ryanjulian: You can make a custom TestCase class (inheriting from `unittest.TestCase`) which overloads setUp/tearDown, and inherit from that (put it in tests/fixtures.py).

tests/envs/test_envs.py
jonashen: Actually this breaks the test. There are multiple assertion errors, i.e. `img is not None`

garage/envs/mujoco/gather/gather_env.py
jonashen: I'm not sure. I can put it back in but since we're inheriting from gym.Wrapper I'm not sure if `hasattr(obj, 'env')` will still work.

ryanjulian:
Please address these comments in a new commit. My apologize for not getting them in before merge.
* https://github.com/rlworkgroup/garage/pull/311#discussion_r212050536
* https://github.com/rlworkgroup/garage/pull/311#discussion_r212063832
* https://github.com/rlworkgroup/garage/pull/311#discussion_r212049587

garage/envs/mujoco/gather/gather_env.py
```diff
@@ -386,38 +383,33 @@ def maze_observation_space(self):
...
-
     @property
     def action_bounds(self):
-        return self.wrapped_env.action_bounds
+        return self.env.action_bounds
 
     # @property
     # def viewer(self):
-    #     return self.wrapped_env.viewer
+    #     return self.env.viewer
```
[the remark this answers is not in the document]
jonashen: I did not comment it out, so I'm not sure. I can test out `self.get_viewer()`

tests/tf/algos/test_trpo.py
jonashen: sounds good

garage/envs/mujoco/gather/gather_env.py
ryanjulian: i mean, if it doesn't, then we need to figure out how to implement the same logic with the new wrapper. i don't think the original author put this here for just for fun.

tests/envs/test_envs.py
ryanjulian: then the code is broken, not the test.

PR #316, opened 2018-08-26
Add code coverage tracking to TravisCI

ryanjulian:


ryanjulian:
This actually adds coverage uploads to 4 providers (Codecov, Coveralls, CodeClimate, and Codacy) so we can evaluate. We'll remove the ones we don't want later.

PR #318, opened 2018-09-05
Fix serialization of Gym Environments

gonzaiva:
By adding a new parameter for the gym environment name instead of the
environment object, we're able to pickle the gym environment using the
Serializable class.
A parameterized test to check on all Gym environments (except for Atari
since tests can take much more time) was added to back up this change
and future changes in the gym environments.


garage/envs/base.py
```diff
@@ -25,10 +25,14 @@ class GarageEnv(gym.Wrapper, Parameterized, Serializable, metaclass=ABCMeta):
     Args: env (gym.Env): the env that will be wrapped
     """
 
-    def __init__(self, env):
-        Serializable.quick_init(self, locals())
+    def __init__(self, env=None, env_name=""):
+        constructor_args = locals()
```
ryanjulian: There is no need for this. Just call `Serializable.quick_init(self, locals())`

tests/integration_tests/test_gym_envs.py
```diff
@@ -0,0 +1,130 @@
...
+
+# White list of environments that cannot be rendered
+not_rendering_envs = [
+    "Blackjack-v0", "NChain-v0", "Roulette-v0", "GuessingGame-v0",
+    "HotterColder-v0", "Blackjack-v0", "NChain-v0", "Roulette-v0",
+    "GuessingGame-v0", "HotterColder-v0"
+]
+
+# White list of environments that cannot be closed
+not_closing_envs = [
```
ryanjulian: Please use the pattern I implemented (but didn't commit) here https://github.com/rlworkgroup/garage/blob/env_pickle_tests/tests/quirks.py
https://github.com/rlworkgroup/garage/blob/env_pickle_tests/tests/helpers.py

In fact, I think you will find that entire `test/` tree useful.

tests/integration_tests/test_gym_envs.py
```diff
@@ -0,0 +1,130 @@
...
+            assert spec.id in not_closing_envs, "It seems environment %s " \
+                    "cannot be closed. Add it to the corresponding white" \
+                    "list in test_gym_envs.py." % (spec.id)
+
+    @params(*list(product(garage_envs, list(gym.envs.registry.all()))))
+    def test_step_and_render(self, garage_env, spec):
+        # Skip Atari environments
+        if spec._entry_point == "gym.envs.atari:AtariEnv":
+            return
+        process = Process(
```
ryanjulian: Please don't launch new processes in tests.

garage/envs/base.py
```diff
@@ -25,10 +25,14 @@ class GarageEnv(gym.Wrapper, Parameterized, Serializable, metaclass=ABCMeta):
     Args: env (gym.Env): the env that will be wrapped
     """
 
-    def __init__(self, env):
-        Serializable.quick_init(self, locals())
+    def __init__(self, env=None, env_name=""):
```
ryanjulian: Good idea using a single wrapper. I wonder if the rest of the wrappers (e.g. Normalized, Noisy) could inherit from this.

tests/integration_tests/test_gym_envs.py
```diff
@@ -0,0 +1,130 @@
...
+import pickle
+import unittest
+
+import gym
+from nose2.tools import params
+
+from garage.tf.envs.base import TfEnv
+from garage.theano.envs.base import TheanoEnv
+
+# White list of environments that cannot be rendered
```
ryanjulian: Please see the test suite I implemented in branch `env_pickle_tests`. I think you should be able to commit all of it along with this PR.

tests/integration_tests/test_gym_envs.py
```diff
@@ -0,0 +1,130 @@
...
+            assert spec.id in not_closing_envs, "It seems environment %s " \
+                    "cannot be closed. Add it to the corresponding white" \
+                    "list in test_gym_envs.py." % (spec.id)
+
+    @params(*list(product(garage_envs, list(gym.envs.registry.all()))))
+    def test_pickling_wrapper(self, garage_env, spec):
+        # Skip Atari environments
+        if spec._entry_point == "gym.envs.atari:AtariEnv":
+            return
+        process = Process(
```
ryanjulian: Please do not launch processes in unit tests.

tests/integration_tests/test_gym_envs.py
```diff
@@ -0,0 +1,130 @@
...
+
+        # Step and render work
+        env.reset()
+        done = False
+        for _ in range(10):
+            _, _, done, _ = env.step(env.action_space.sample())
+            try:
+                env.render()
+            except NotImplementedError:
+                assert spec.id in not_rendering_envs, "It seems environment " \
```
ryanjulian: I don't think this is necessary and it might tell someone in the future to do the wrong thing.

There are many reasons why an environment might fail to render during tests, and most of the time it's because we broke something (in which case it should not be whitelisted).

tests/integration_tests/test_gym_envs.py
```diff
@@ -0,0 +1,130 @@
...
+                        "%s cannot be rendered. Add it to the corresponding " \
+                        "white list in test_gym_envs.py." % (spec.id)
+            if done:
+                break
+
+        try:
+            env.close()
+        except AttributeError:
+            assert spec.id in not_closing_envs, "It seems environment %s " \
+                    "cannot be closed. Add it to the corresponding white" \
```
ryanjulian: Remove this. The future programmer should root cause it (and read the test) before reflexively whitelisting it.

tests/integration_tests/test_gym_envs.py
```diff
@@ -0,0 +1,130 @@
...
+                    "list in test_gym_envs.py." % (spec.id)
+
+    @params(*list(product(garage_envs, list(gym.envs.registry.all()))))
+    def test_step_and_render(self, garage_env, spec):
+        # Skip Atari environments
+        if spec._entry_point == "gym.envs.atari:AtariEnv":
```
ryanjulian: Why skip atari environments? They should work.

tests/integration_tests/test_gym_envs.py
```diff
@@ -0,0 +1,130 @@
...
+    "HandManipulatePenRotateDense-v0", "HandManipulatePenFullDense-v0",
+    "HandManipulatePenDense-v0"
+]
+
+
+class TestGymEnvs(unittest.TestCase):
+
+    garage_envs = [TfEnv, TheanoEnv]
+
+    def _routine_test_pickling_wrapper(self, garage_env, spec):
```
ryanjulian: If you need a fixture, please put it in `tests.fixtures`. See the env test suite I made in `env_pickle_tests`

tests/integration_tests/test_gym_envs.py
```diff
@@ -0,0 +1,130 @@
...
+
+        try:
+            env.render()
+        except NotImplementedError:
+            assert spec.id in not_rendering_envs, "It seems environment %s " \
+                    "cannot be rendered. Add it to the corresponding white" \
+                    "list in test_gym_envs.py." % (spec.id)
+
+        # Test serialization once the viewer object has been set in env, but
+        # don't reassign env with the one produced with pickle to avoid
```
ryanjulian: I'm not sure what this means. `env.close()` and `pickled_env.close()` should *always* execute without error.

tests/integration_tests/test_gym_envs.py
```diff
@@ -0,0 +1,130 @@
...
+            try:
+                env.render()
+            except NotImplementedError:
+                assert spec.id in not_rendering_envs, "It seems environment " \
+                        "%s cannot be rendered. Add it to the corresponding " \
+                        "white list in test_gym_envs.py." % (spec.id)
+            if done:
+                break
+
+        try:
```
ryanjulian: try/catch in tests is an antipattern. The test should always work without error. If you expect an exception, you can use `with self.assertRaises():`

tests/integration_tests/test_gym_envs.py
ryanjulian: Why did you launch a process?

gonzaiva: I'm launching processes here due to MjViewers. By running this test in my computer, the windows for the MjViewers aren't closed with env.close(), but until the parent process finishes, and since there's a lot of MuJoCo related environments, there are a lot of windows rendering MuJoCo simulations by the middle of the test which leads to shortage of RAM and sometimes termination of the test process to keep the computer alive.
By adding a process for each MjViewer I'm able to close them right after the env is closed, but I agree that this may not the best way to solve this problem, and it also seems to be generating problems in Travis, specifically this error:
```
XIO:  fatal IO error 11 (Resource temporarily unavailable) on X server ":0"
      after 81 requests (81 known processed) with 0 events remaining.
```
Maybe there's a way to detect and close this windows with python.

ryanjulian: Jonathon and I have this problem too, both on Linux and Mac. It's a bug in mujoco_py. However it's only a problem when testing--because most people don't launch 100+ viewer windows.

It's bad to launch a new process when you test, because it messes with the test runner, and also is not faithful to how the code is executed.

I deal with it when testing locally by running tests with xvfb-run, which makes the windows go to a virtual display.

`xvfb-run --server-args="-screen 0 1024x768x16 +iglx" nose2 ...`

tests/integration_tests/test_gym_envs.py
gonzaiva: Some environments produced by gym.make have an attribute called viewer, which is initially equal to None, but after `env.render()` is called, it's assigned to the object that is responsible of the rendering, and more specifically, of closing the window displaying the environment.
If viewer has been set and then we do the call:
`env = pickle.loads(pickle.dumps(env))`
A new env is built again but with the viewer object assigned to None, losing the reference to the first viewer, so when `env.close()` is called, the viewer window is not closed, which of course doesn't produce an error, it only doesn't close the window until the testing process is closed, but this may not be a problem if execute the test with the virtual displays you mentioned above.

ryanjulian: I don't think this needs a block comment. We should keep both copies around, anyway.

tests/integration_tests/test_gym_envs.py
gonzaiva: The reason I skipped them is because there's a lot of them, but I will re-enable them again then.

garage/envs/mujoco/maze/maze_env.py
```diff
@@ -236,9 +237,9 @@ def reset(self):
         self.env.reset()
         return self.get_current_obs()
 
-    @property
-    def viewer(self):
-        return self.env.viewer
+    # @property
```
zhanpenghe: I think we can preserve the viewer property.

tests/helpers.py
```diff
@@ -0,0 +1,96 @@
+import pickle
```
jonashen: Considering combining this file with `tests/fixtures.py`

tests/mock.py
```diff
@@ -0,0 +1,6 @@
+from unittest import mock
```
jonashen: Can I ask what this file is used for?

tests/garage/envs/test_point_env.py
```diff
@@ -0,0 +1,13 @@
...
+
+from garage.envs.point_env import PointEnv
+from tests.helpers import step_env
+
+
+class TestPointEnv(unittest.TestCase):
+    def test_pickleable(self):
+        env = PointEnv()
+        round_trip = pickle.loads(pickle.dumps(env))
+        assert round_trip
```
jonashen: Are you just asserting round_trip exists because there are no parameters in env that you can compare equality to?

garage/theano/envs/base.py
```diff
@@ -22,8 +22,8 @@ class TheanoEnv(GarageEnv):
         env (gym.Env): the env that will be wrapped
     """
 
-    def __init__(self, env):
-        super(TheanoEnv, self).__init__(env)
+    def __init__(self, env=None, env_name=""):
+        super(TheanoEnv, self).__init__(env, env_name)
```
jonashen: Should it be `super().__init__(env, env_name)`?

tests/garage/envs/test_point_env.py
jonashen: This question also extends to all of your other tests that only have `assert round_trip` and no other assertions.

tests/mock.py
ryanjulian: https://medium.com/@piraveenaparalogarajah/what-is-mocking-in-testing-d4b0f2dbe20a

tests/helpers.py
ryanjulian: I didn't read the file, but basically helpers are small helper functions and fixtures are larger constructs (generally they subclass unittest.TestCase, but not always).

tests/garage/envs/test_point_env.py
ryanjulian: Trivially we want to make sure any object was unpickled at all.

garage/theano/envs/base.py
ryanjulian: super() will call the constructor of the *first* parent class only. If your class uses mixins/multiple inheritance, you have to explicitly call the test of the parent constructors if you need them.

scripts/travisci/check_tests.sh
```diff
@@ -1,6 +1,76 @@
 #!/usr/bin/env bash
+
```
ryanjulian: I don't think we should have scripts this complicated in the CI stack. Just have different scripts for the cron job versus the regular builds.

You don't want to report coverage for the cron_job--it will make the coverage system think that the base coverage on master is changing. The coverage system just uses the latest report from master. The cron job doesn't cover much of the code, so I think the coverage rating will start oscillating based on whether the cron or master job ran last.

tests/garage/envs/mujoco/test_inverted_double_pendulum_env.py
```diff
@@ -0,0 +1,15 @@
+import pickle
+import unittest
+
+from garage.envs.mujoco.inverted_double_pendulum_env import \
```
ryanjulian: PEP8 says to use parentheses to split lines, not \

tests/garage/envs/test_envs.py
```diff
@@ -47,6 +47,12 @@
     SwimmerMazeEnv,
     PointMazeEnv,
```
ryanjulian: Please remove this file now that we have a proper environment test suite.

tests/garage/envs/test_occlusion_env.py
```diff
@@ -0,0 +1,20 @@
...
+        inner_env = CartpoleEnv(obs_noise=5.)
+        env = OcclusionEnv(inner_env, [1])
+        round_trip = pickle.loads(pickle.dumps(env))
+        assert round_trip
+        obs = inner_env.reset()
+        assert round_trip.occlude(obs) == env.occlude(obs)
+        assert round_trip.env.obs_noise == env.env.obs_noise
+        step_env(round_trip)
+
+    test_pickleable.broken = True
```
ryanjulian: Is there an issue filed to fix this?

tests/garage/envs/test_sliding_mem_env.py
```diff
@@ -0,0 +1,19 @@
...
+    def test_pickleable(self):
+        inner_env = CartpoleEnv(obs_noise=5.)
+        env = SlidingMemEnv(inner_env, n_steps=10)
+        round_trip = pickle.loads(pickle.dumps(env))
+        assert round_trip
+        assert round_trip.n_steps == env.n_steps
+        assert round_trip.env.obs_noise == env.env.obs_noise
+        step_env(round_trip)
+
+    test_pickleable.broken = True
```
ryanjulian: Is there an issue filed to fix this?

tests/garage/spaces/test_box.py
```diff
@@ -0,0 +1,11 @@
...
+import unittest
+
+from garage.spaces.box import Box
+
+
+class TestBox(unittest.TestCase):
+    def test_pickleable(self):
+        obj = Box(-1.0, 1.0, (3, 4))
+        round_trip = pickle.loads(pickle.dumps(obj))
+        assert round_trip
```
ryanjulian: Make sure you got the same box back

tests/garage/spaces/test_dict.py
```diff
@@ -0,0 +1,12 @@
...
+import unittest
+
+from garage.spaces.dict import Dict
+from garage.spaces.discrete import Discrete
+
+
+class TestDict(unittest.TestCase):
+    def test_pickleable(self):
+        obj = Dict({"position": Discrete(2), "velocity": Discrete(3)})
+        round_trip = pickle.loads(pickle.dumps(obj))
```
ryanjulian: Make sure you got the same dict back

tests/garage/spaces/test_discrete.py
```diff
@@ -0,0 +1,11 @@
...
+import unittest
+
+from garage.spaces.discrete import Discrete
+
+
+class TestDiscrete(unittest.TestCase):
+    def test_pickleable(self):
+        obj = Discrete(10)
+        round_trip = pickle.loads(pickle.dumps(obj))
+        assert round_trip
```
ryanjulian: Make sure you got the same space back

tests/garage/spaces/test_product.py
```diff
@@ -0,0 +1,12 @@
+import pickle
+import unittest
+
+from garage.spaces.product import Product
+from garage.theano.spaces import Discrete
+
+
+class TestProduct(unittest.TestCase):
+    def test_pickleable(self):
+        obj = Product([Discrete(3), Discrete(2)])
```
ryanjulian: Make sure you got the same space back

tests/garage/tf/spaces/test_box.py
```diff
@@ -0,0 +1,11 @@
...
+import unittest
+
+from garage.tf.spaces.box import Box
+
+
+class TestBox(unittest.TestCase):
+    def test_pickleable(self):
+        obj = Box(-1.0, 1.0, (3, 4))
+        round_trip = pickle.loads(pickle.dumps(obj))
+        assert round_trip
```
ryanjulian: Make sure you got the same box back

tests/garage/tf/spaces/test_dict.py
```diff
@@ -0,0 +1,25 @@
...
+            gym.spaces.Box(
+                low=-200., high=200., shape=(3, ), dtype=np.float32),
+            "desired_goal":
+            gym.spaces.Box(
+                low=-200., high=200., shape=(3, ), dtype=np.float32),
+            "observation":
+            gym.spaces.Box(
+                low=-200., high=200., shape=(25, ), dtype=np.float32)
+        }
+        obj = Dict(spaces)
```
ryanjulian: Make sure you get the same space back.

tests/garage/tf/spaces/test_discrete.py
```diff
@@ -0,0 +1,11 @@
...
+import unittest
+
+from garage.tf.spaces.discrete import Discrete
+
+
+class TestDiscrete(unittest.TestCase):
+    def test_pickleable(self):
+        obj = Discrete(10)
+        round_trip = pickle.loads(pickle.dumps(obj))
+        assert round_trip
```
ryanjulian: Make sure you get the same space back

tests/garage/tf/spaces/test_product.py
```diff
@@ -0,0 +1,12 @@
...
+
+from garage.tf.spaces import Discrete
+from garage.tf.spaces.product import Product
+
+
+class TestProduct(unittest.TestCase):
+    def test_pickleable(self):
+        obj = Product([Discrete(3), Discrete(2)])
+        round_trip = pickle.loads(pickle.dumps(obj))
+        assert round_trip
```
ryanjulian: Make sure you get the same space back

tests/garage/theano/spaces/test_box.py
```diff
@@ -0,0 +1,11 @@
...
+import unittest
+
+from garage.theano.spaces.box import Box
+
+
+class TestBox(unittest.TestCase):
+    def test_pickleable(self):
+        obj = Box(-1.0, 1.0, (3, 4))
+        round_trip = pickle.loads(pickle.dumps(obj))
+        assert round_trip
```
ryanjulian: Make sure you get the same box back

tests/garage/theano/spaces/test_dict.py
```diff
@@ -0,0 +1,25 @@
...
+            "desired_goal":
+            gym.spaces.Box(
+                low=-200., high=200., shape=(3, ), dtype=np.float32),
+            "observation":
+            gym.spaces.Box(
+                low=-200., high=200., shape=(25, ), dtype=np.float32)
+        }
+        obj = Dict(spaces)
+        round_trip = pickle.loads(pickle.dumps(obj))
+        assert round_trip
```
ryanjulian: Make sure you get the same space back

tests/garage/theano/spaces/test_discrete.py
```diff
@@ -0,0 +1,11 @@
...
+import unittest
+
+from garage.theano.spaces.discrete import Discrete
+
+
+class TestDiscrete(unittest.TestCase):
+    def test_pickleable(self):
+        obj = Discrete(10)
+        round_trip = pickle.loads(pickle.dumps(obj))
+        assert round_trip
```
ryanjulian: Make sure you get the same space back

tests/garage/theano/spaces/test_product.py
```diff
@@ -0,0 +1,12 @@
...
+
+from garage.theano.spaces import Discrete
+from garage.theano.spaces.product import Product
+
+
+class TestProduct(unittest.TestCase):
+    def test_pickleable(self):
+        obj = Product([Discrete(3), Discrete(2)])
+        round_trip = pickle.loads(pickle.dumps(obj))
+        assert round_trip
```
ryanjulian: Make sure you get the same space back

tests/helpers.py
```diff
@@ -0,0 +1,96 @@
...
+        env.close()
+    else:
+        with test_case.assertRaisesRegex(
+                AttributeError, "'MjViewer' object has no attribute 'finish'"):
+            env.close()
+    if any([s in spec._entry_point for s in KNOWN_GYM_NOT_CLOSE_VIEWER]):
+        close_viewer_window(env)
+
+
+def pickle_env_with_gym_quirks(test_case, env, spec, render=True):
```
ryanjulian: I don't think this routine is necessary at all. It just pickles the environment followed by copy/paste of the helper above. Why not just write one after the other?

tests/helpers.py
```diff
@@ -0,0 +1,96 @@
...
+import glfw
+from mujoco_py.mjviewer import MjViewer
+import tensorflow as tf
+
+from tests.quirks import KNOWN_GYM_CLOSE_BROKEN
+from tests.quirks import KNOWN_GYM_NOT_CLOSE_VIEWER
+from tests.quirks import KNOWN_GYM_RENDER_NOT_IMPLEMENTED
+
+
+class TfTest(unittest.TestCase):
```
jonashen: I have this equivalent implementation in my PR, but also addresses memory leakage and is located in `tests/fixtures`: https://github.com/rlworkgroup/garage/pull/299/files#diff-ef6796cafe4c24fd8bb646d79094d9f6R14. Let's discuss about how to combine our implementations

tests/garage/envs/test_sliding_mem_env.py
gonzaiva: All of these have the broken flag:
```
tests/garage/envs/mujoco/gather/test_swimmer_gather_env.py:16:    test_pickleable.broken = True
tests/garage/envs/mujoco/gather/test_point_gather_env.py:16:    test_pickleable.broken = True
tests/garage/envs/mujoco/gather/test_ant_gather_env.py:16:    test_pickleable.broken = True
tests/garage/envs/mujoco/maze/test_swimmer_maze_env.py:16:    test_pickleable.broken = True
tests/garage/envs/mujoco/maze/test_point_maze_env.py:16:    test_pickleable.broken = True
tests/garage/envs/mujoco/maze/test_ant_maze_env.py:16:    test_pickleable.broken = True
tests/garage/envs/test_sliding_mem_env.py:19:    test_pickleable.broken = True
tests/garage/envs/test_occlusion_env.py:20:    test_pickleable.broken = True
```
But the flag is not used with any nose2 configuration (I checked with grep), and if I run:
```
nose2 -c setup.cfg -A broken
```
All of the tests are passing.

I will remove this flag since it's not used anywhere else and the tests are passing. There's an issue for the rendering of the GatherEnv but it's already documented at #323.

tests/helpers.py
gonzaiva: I agree, however, I think adding a parameter in method step_env_with_gym_quirks to perform the pickle testing is a better option, since in these [lines](https://github.com/rlworkgroup/garage/blob/9f06a1e1754d73fac7ea8c077a77e60e147a8ef0/tests/helpers.py#L73-L74) I'm checking that the environment can be serialized once the viewer object is present (this is something important because it tests a bug we have already solved), but it has to be done before closing the environment, which is in the same block of code.

tests/helpers.py
gonzaiva: Me and Jonathon concluded that this TestCase and the one on his PR only differ in setting and removing the TF graph, so there's no big conflict for him to rebase his changes on mine.

PR #320, opened 2018-09-07
Fixed remaining CI tests

jonashen:
Enabled and fixed all remaining CI tests. Also added GarageTestCase that inherits from unittest.TestCase with overriding setUp and tearDown function calls.

Ref: #298

### DIFF
--- a/docker/docker-entrypoint.sh
+++ b/docker/docker-entrypoint.sh
@@ -28,4 +28,8 @@ fi
 # Activate conda environment
 source activate garage
 
+# Fixes Segmentation Fault
+# See: https://github.com/openai/mujoco-py/pull/145#issuecomment-356938564
+export LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libGLEW.so:/usr/lib/x86_64-linux-gnu/mesa/libGL.so.1
+
 exec "$@"
--- a/tests/fixtures/fixtures.py
+++ b/tests/fixtures/fixtures.py
@@ -1,0 +1,16 @@
+import unittest
+
+import tensorflow as tf
+
+import garage.misc.logger as logger
+from garage.misc.tensorboard_output import TensorBoardOutput
+
+
+class GarageTestCase(unittest.TestCase):
+    def setUp(self):
+        self.sess = tf.Session(graph=tf.Graph())
+        self.sess.__enter__()
+        logger._tensorboard = TensorBoardOutput()
+
+    def tearDown(self):
+        self.sess.close()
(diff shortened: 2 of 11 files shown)


tests/fixtures/fixtures.py
```diff
@@ -0,0 +1,16 @@
...
+from garage.misc.tensorboard_output import TensorBoardOutput
+
+
+class GarageTestCase(unittest.TestCase):
+    def setUp(self):
+        self.sess = tf.Session(graph=tf.Graph())
+        self.sess.__enter__()
+        logger._tensorboard = TensorBoardOutput()
+
+    def tearDown(self):
```
ryanjulian: I had trouble with this crashing when I did this, so glad that this worked for you lol.

tests/garage/envs/test_envs.py
```diff
@@ -43,28 +43,43 @@
     MountainCarEnv,
 ]
 
+mujoco_classes = [
+    SwimmerMazeEnv,
+    PointMazeEnv,
+    AntMazeEnv,
+]
```
ryanjulian: I think it's time to split out test_envs into individual tests. Can you check out @gonzaiva's PR and coordinate with him on how to merge yours and his?

ryanjulian: Alternatively you can just merge and let him rebase, since this is relatively small. But please make sure he integrates your new tests in his change (it's pretty big so it would be easy for him to miss).

docker/docker-entrypoint.sh
```diff
@@ -28,4 +28,7 @@ fi
 # Activate conda environment
 source activate garage
 
+# Fixes GLEW initialization / segmentation errors
```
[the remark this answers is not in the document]
jonashen: Sure thing, I'll add it

tests/fixtures/fixtures.py
```diff
@@ -0,0 +1,16 @@
+import unittest
+
+import tensorflow as tf
+
+import garage.misc.logger as logger
+from garage.misc.tensorboard_output import TensorBoardOutput
+
+
+class GarageTestCase(unittest.TestCase):
```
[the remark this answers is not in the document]
jonashen: Not sure, I can ask her

tests/garage/envs/test_envs.py
jonashen: I think rebasing this will be easier as it's smaller.

tests/fixtures/fixtures.py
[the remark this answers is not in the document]
jonashen: We decided I will merge first, and she will build on top of this.

PR #321, opened 2018-09-07
Install fixes: CONDA_SH, garage packaging

gautams3:
* Defining CONDA_SH for all installations (unconditionally)
* Updating garage/setup.py with package name
* Updated setup scripts to install garage as an editable package


setup.py
```diff
@@ -2,7 +2,7 @@
 from setuptools import setup
 
 setup(
-    name='rlgarage',
+    name='garage',
```
ryanjulian: unfortunately garage is taken on PyPI so I have named it rlgarage for now.

ryanjulian:
@eric-heiden `pip install -e .` merely adds the garage dir to the PYTHONPATH of the conda environment permanently, but without the hassle of mangled PYTHONPATH.

The only people this affects are those who maintain multiple parallel garage installs, which AFAIK is only me. I should have a different conda env for each garage repo, anyway.

docker/Dockerfile.ci
```diff
@@ -85,6 +85,7 @@ RUN conda env create -f /root/code/garage/environment.yml && \
 # Extras
 # prevent pip from complaining about available upgrades
 RUN ["/bin/bash", "-c", "source activate garage && pip install --upgrade pip"]
+RUN ["/bin/bash", "-c", "pip install -e ."]
```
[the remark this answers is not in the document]
gautams3: It is an editable install, not a vanilla pip install. So your local dev directory is the source of the package
https://pip.pypa.io/en/stable/reference/pip_install/#editable-installs

@ryanjulian, anything you want to add?

ryanjulian: Please leave a comment about that this does for future victims.

scripts/setup_linux.sh
```diff
@@ -207,6 +207,9 @@ conda activate garage
   # Prevent pip from complaining about available upgrades
   pip install --upgrade pip
 
+  # 'Install' garage as an editable package
+  pip install -e .
```
gonzaiva: Please remove all things related to PYTHONPATH and PYTHON_ENV_VAR here and in setup_linux since we're using pip install now.

scripts/setup_linux.sh
```diff
@@ -178,14 +178,14 @@ if [[ "${_arg_modify_bashrc}" = on ]]; then
 fi
 
 # Set up conda
+# Add conda to executable programs
```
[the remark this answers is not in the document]
gautams3: Done

setup.py
gautams3: but does that mean our package will be named rlgarage? That will change all our example files

scripts/setup_linux.sh
gautams3: Done

ryanjulian:
Ignore codecov.

docker/Dockerfile.ci
```diff
@@ -110,7 +110,7 @@ RUN curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest
 
 # Add code stub last
 COPY . /root/code/garage
-ENV PYTHONPATH=$PYTHONPATH:/root/code/garage
+RUN ["/bin/bash", "-c", "pip install -e /root/code/garage"]
```
ryanjulian: You need to `source activate garage` before you do this, or else you're using the system pip and not the conda pip.

e.g.
```Dockerfile
RUN ["/bin/bash", "-c", "source activate garage && pip install -e /root/code/garage"]
```

I'm surprised the CI passed, but it's probably because `WORKDIR` happens to be right in the repo root.

docs/user/installation.rst
```diff
@@ -5,15 +5,6 @@
 Installation
 ============
 
-Preparation
-===========
-
-You need to edit your :code:`PYTHONPATH` to include the garage directory:
-
-.. code-block:: bash
```
ryanjulian: You should tell folks to `pip install -e .` after `source activate garage` instead, or else they won't have `garage` in their path.

gautams3: Is there no way to do it from within the docker files?

gautams3: I made changes to Dockerfile.ci, in an attempt to automate this instead of asking users to `pip install` 🤞

gautams3:
Note: I also added macOS 10.13 (High Sierra) in setup_osx.sh as a verified OSX version

docs/user/installation.rst
ryanjulian: Users don't use Dockerfile.ci--it's currently only used by the CI.

docker/Dockerfile.ci
```diff
@@ -110,7 +110,9 @@ RUN curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest
 
 # Add code stub last
 COPY . /root/code/garage
-ENV PYTHONPATH=$PYTHONPATH:/root/code/garage
+RUN ["/bin/bash", "-c", "source activate garage"]
```
ryanjulian: These all have to be in the same RUN layer. The conda environment only lasts for a single bash session (these lines create a new session every time).

docs/user/installation.rst
gautams3: The setup scripts were already updated to run `pip install`

[setup_osx](https://github.com/rlworkgroup/garage/pull/321/files#diff-9b3afda2da20b8b001323b10f7fa3baeR228)

[setup_linux](https://github.com/rlworkgroup/garage/pull/321/files#diff-078766c0ef4038e6f79e94cec003a016R210)

Verified manually on macOS 10.13.6

Issue #257, opened 2018-08-02
Remove env.horizon from garage


jonashen:
This is a deprecated call that can be replaced with `env.spec.max_episode_steps` only where appropriate, e.g. if env.spec does not exist, further logic is required.

PR #328, opened 2018-09-25
Removed all remaining env.horizon calls

jonashen:
Ref: #257

### DIFF
--- a/garage/envs/mujoco/randomization/randomized_env.py
+++ b/garage/envs/mujoco/randomization/randomized_env.py
@@ -1,11 +1,13 @@
 import os.path as osp
 
+from cached_property import cached_property
 import gym
 from mujoco_py import load_model_from_xml
 from mujoco_py import MjSim
 
 from garage.core import Serializable
 from garage.envs.mujoco.mujoco_env import MODEL_DIR
+from garage.misc.overrides import overrides
 
 
 class RandomizedEnv(gym.Env, Serializable):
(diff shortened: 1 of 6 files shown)


garage/theano/envs/base.py
```diff
@@ -58,3 +58,14 @@ def spec(self):
         return EnvSpec(
             observation_space=self.observation_space,
             action_space=self.action_space)
+
+    @cached_property
```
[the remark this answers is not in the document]
jonashen: So we don't have to keep retrieving it two levels down (i.e. normalized -> gym.Env)

garage/envs/mujoco/randomization/randomized_env.py
```diff
@@ -73,9 +73,5 @@ def action_space(self):
     def observation_space(self):
         return self._wrapped_env.observation_space
 
-    @property
-    def horizon(self):
```
gonzaiva: Why is horizon completely removed instead of renamed/replaced?

gonzaiva: I see, it can be replaced with env.spec.max_episode_steps. Although if we're adding a lot of wrappers to the env, for example, two wrappers, we would need to call it like env.env.env.spec.max_espisode_steps, right?

jonashen: Yes, but that's why I also added `max_episode_steps` functions to all of our env wrappers.

jonashen: So we wouldn't need several levels of env.env.env

gonzaiva: Okay, but randomize is also used as an env wrapper and you didn't add max_episodes_steps to it.

jonashen: Good catch, I'll add it now.

Issue #267, opened 2018-08-10
Merge HERReplayBuffer and ReplayBuffer into a single interface


CatherineSue:

CatherineSue:
See also: #265

CatherineSue:
Fixed in #329

PR #329, opened 2018-09-26
Fix VPG and DDPG to use VectorizedSampler

CatherineSue:
- Rename the current VectorizedSampler to OnPolicyVectorizedSampler
   since it's more customized for on policy algorithms.
 - Write an OffPolicyVectorizedSampler for off polocy optimization
   algorithms.
 - Merge HerReplayBuffer and ReplayBuffer into same interface.
 - Fix VPG to use OnPolicyPolopt.

Closes: #252 , #267


CatherineSue:
The benchmark of DDPG and HER will be provided soon. You can start the code review after it's finished.

garage/tf/algos/batch_polopt.py
```diff
@@ -168,3 +112,7 @@ def get_itr_snapshot(self, itr, samples_data):
 
     def optimize_policy(self, itr, samples_data):
         raise NotImplementedError
+
+    @property
+    def off_policy(self):
```
zhanpenghe: Where is this property used?

garage/tf/samplers/off_policy_vectorized_sampler.py
```diff
@@ -0,0 +1,155 @@
...
+import numpy as np
+
+from garage.misc import tensor_utils
+from garage.misc.overrides import overrides
+from garage.replay_buffer.base import Buffer
+from garage.tf.envs import VecEnvExecutor
+from garage.tf.samplers import BatchSampler
+
+
+class OffPolicyVectorizedSampler(BatchSampler):
```
zhanpenghe: Why do we need two samplers for on policy and off policy? I believe that these two could be easily unified.

garage/tf/algos/ddpg.py
```diff
@@ -36,359 +27,87 @@ class DDPG(RLAlgorithm):
...
-                 critic_optimizer=None,
-                 use_her=False,
-                 clip_obs=np.inf,
-                 clip_pos_returns=True,
+                 policy_lr=1e-4,
+                 qf_lr=1e-3,
+                 replay_buffer_type=Buffer.REGULAR,
+                 policy_weight_decay=0,
+                 qf_weight_decay=0,
+                 policy_optimizer=None,
```
zhanpenghe: I think we can set adam as default.

garage/tf/algos/off_policy_batch_polopt.py
```diff
@@ -0,0 +1,179 @@
...
+from garage.envs.util import dims_to_shapes
+import garage.misc.logger as logger
+from garage.misc.overrides import overrides
+from garage.replay_buffer import HerReplayBuffer, RegularReplayBuffer
+from garage.replay_buffer.base import Buffer
+from garage.replay_buffer.her_replay_buffer import make_her_sample
+from garage.tf.algos.batch_polopt import BatchPolopt
+
+
+class OffPolicyBatchPolopt(BatchPolopt):
```
zhanpenghe: I don't think BatchPolopt is an appropriate name for these algorithms

garage/tf/policies/continuous_mlp_policy.py
```diff
@@ -137,23 +139,29 @@ def get_actions(self, observations):
         """Return multiple actions."""
         return self._f_prob_online(observations), dict()
 
-    @property
-    def trainable_vars(self):
-        """Return trainable vars in the network."""
-        return tf.get_collection(
-            tf.GraphKeys.TRAINABLE_VARIABLES, scope=self.name)
+    def get_trainable_vars(self, scope=None):
```
zhanpenghe: This can be moved to a super class.

garage/tf/policies/continuous_mlp_policy.py
```diff
@@ -137,23 +139,29 @@ def get_actions(self, observations):
...
+        scope = scope if scope else self.name
+        return tf.get_collection(tf.GraphKeys.TRAINABLE_VARIABLES, scope=scope)
 
-    @property
-    def global_vars(self):
-        """Return the global vars in the network."""
-        return tf.get_collection(
-            tf.GraphKeys.GLOBAL_VARIABLES, scope=self.name)
+    def get_global_vars(self, scope=None):
+        scope = scope if scope else self.name
```
zhanpenghe: This can also be moved to a super class.

garage/tf/algos/ddpg.py
```diff
@@ -36,359 +27,87 @@ class DDPG(RLAlgorithm):
...
-                    sample_transitions=make_her_sample(
-                        self.replay_k, self.env.compute_reward))
-            else:
-                replay_buffer = ReplayBuffer(
-                    buffer_shapes=input_shapes,
-                    max_buffer_size=self.replay_buffer_size)
+            # Create target policy and qf network
+            self.target_policy_f_prob_online, _, _ = self.policy._build_net(
+                trainable=False, name="target_policy")
+            self.target_qf_f_prob_online, _, _, _ = self.baseline._build_net(
```
zhanpenghe: There is no baseline in ddpg.

garage/sampler/stateful_pool.py
```diff
@@ -60,7 +60,7 @@ def initialize(self, n_parallel):
             self.manager = mp.Manager()
             self.queue = mp.Queue()
             self.worker_queue = mp.Queue()
-            self.pool = MemmappingPool(
+            self.pool = MemmapingPool(
```
ryanjulian: Please use `conda env update -n garage` to get the latest joblib, which corrected this typo.

garage/tf/algos/batch_polopt.py
```diff
@@ -66,22 +55,16 @@ def __init__(self,
...
-        self.positive_adv = positive_adv
         self.store_paths = store_paths
-        self.whole_paths = whole_paths
-        self.fixed_horizon = fixed_horizon
+        self.es = exploration_strategy
+        self.input_include_goal = input_include_goal
         if sampler_cls is None:
             if self.policy.vectorized and not force_batch_sampler:
-                sampler_cls = VectorizedSampler
+                sampler_cls = OffPolicyVectorizedSampler if self.off_policy else OnPolicyVectorizedSampler
```
ryanjulian: @krzentner again to look at the dispatch strategy used here, before it gets out-of-hand

garage/tf/algos/ddpg.py
ryanjulian: Yes--sane defaults please.

garage/tf/algos/ddpg.py
```diff
@@ -36,359 +27,87 @@ class DDPG(RLAlgorithm):
...
-        self.evaluate = False
-        self.replay_k = replay_k
-        self.clip_return = (
-            1. / (1. - self.discount)) if clip_return is None else clip_return
-        self.clip_obs = clip_obs
         self.clip_pos_returns = clip_pos_returns
-        self.success_history = deque(maxlen=100)
-        self._initialize()
+        if not clip_return:
+            if replay_buffer_type == Buffer.HER:
```
ryanjulian: How would you design this if there are 10 different choices of replay buffer?

garage/tf/algos/ddpg.py
ryanjulian: To be less smug, a baseline is a value function (estimates vf: state --> \E{sum of discounted rewards}), while a Q-Function estimates  (qf: (state, action) --> \E{sum of discounted rewards})

garage/tf/algos/off_policy_batch_polopt.py
```diff
@@ -0,0 +1,179 @@
...
+                    logger.record_tabular('QFunction/AverageY',
+                                          np.mean(epoch_ys))
+                    logger.record_tabular('QFunction/MaxY', np.max(epoch_ys))
+                    logger.record_tabular('QFunction/AverageAbsY',
+                                          np.mean(np.abs(epoch_ys)))
+                    if self.replay_buffer_type == Buffer.HER:
+                        logger.record_tabular('AverageSuccessRate',
+                                              np.mean(self.success_history))
+
+                # Uncomment the following if you want to calculate the average
```
ryanjulian: Please find a different way of doing this or remove it--commenting/uncommenting code is not a valid way of changing a module's behavior.

garage/tf/samplers/off_policy_vectorized_sampler.py
ryanjulian: A better question is "which parts are the same and which parts are different?"

garage/envs/util.py
```diff
@@ -116,21 +117,12 @@ def configure_dims(env):
...
-                "action": flat_dim(action),
-                "terminal": 1,
-                "reward": 1,
-                "next_observation": flat_dim(obs),
-            }
+        dims = {
+            "observation": obs.flat_dim,
+            "action": action.flat_dim,
+            "terminal": 0,
+            "reward": 0,
```
jonashen: What's the reason you converted 1 to 0 here for `terminal` and `reward`?

garage/envs/util.py
```diff
@@ -2,7 +2,7 @@
 import numpy as np
 
 from garage.misc import special
-from garage.spaces import Space
+from garage.tf.spaces import Dict
```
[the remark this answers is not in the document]
jonashen: The code is exactly the same across both theano and tf implementations of Dict. If anything, we can use `from garage.spaces.dict import Dict` here, and I can make a new PR adding the missing functions that are not currently in the base Dict class but are used here (e.g. `flat_dim_with_keys `, `flatten_with_keys`)

CatherineSue:
# Benchmarks for Mujoco1M

Here is the plot of benchmarks Mujoco1M. Each plot represents the average reward comparison between ddpg-garage(using primitives from garage) and ddpg-baselines. There are seven tasks in Mujoco1M.

## InvertedDoublePendulum-v2

## InvertedPendulum-v2
invertedpendulum-v2_benchmark

## HalfCheetah-v2
halfcheetah-v2_benchmark

## Hopper-v2
hopper-v2_benchmark

## Walker2d-v2
walker2d-v2_benchmark

## Reacher-v2
reacher-v2_benchmark

## Swimmer-v2
swimmer-v2_benchmark

garage/tf/algos/ddpg.py
CatherineSue: I didn't get a better idea on how to address this.
For now, OffPolicyBatchPolopt and OnPolicyBatchPolopt both inherit from BatchPolopt, `BatchPolopt.log_diagnostics` needs to use `self.baseline`, so I used policy and baseline in BatchPolopt's constructor. 
I can override `log_diagnostics` in OffPolicyBatchPolopt

garage/tf/policies/continuous_mlp_policy.py
CatherineSue: Yes I agree. But I didn't know where to put this. Because I also used this function in continuous_q_fuction.py, so should I put this in garage.tf.Policy or somewhere else?

garage/envs/util.py
```diff
@@ -102,12 +102,13 @@ def configure_dims(env):
     obs = env.observation_space
     action = env.action_space
 
-    if isinstance(obs, gym.spaces.Dict):
```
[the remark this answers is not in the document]
CatherineSue: `configure_dims` is used to build a `input_dims` dict for replay_buffer. `env.reset()` is removeable here.

garage/replay_buffer/base.py
```diff
@@ -0,0 +1,92 @@
...
+import numpy as np
+
+
+@unique
+class Buffer(Enum):
+    REGULAR = "regular"
+    HER = "her"
+
+
+class ReplayBuffer:
```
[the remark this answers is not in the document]
CatherineSue: I can change to use only ReplayBuffer and HERReplayBuffer here

garage/replay_buffer/base.py
```diff
@@ -0,0 +1,92 @@
...
+                                 self._current_size + size_increment)
+
+        if size_increment == 1:
+            idx = idx[0]
+        return idx
+
+    def _convert_episode_to_batch_major(self):
+        transitions = {}
+        for key in self._episode_buffer.keys():
+            val = np.array(self._episode_buffer[key]).copy()
```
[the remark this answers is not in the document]
CatherineSue: I swapped the axes of `val` in the next line. That's why we don't store them into batch major. 
It's okay to not use `copy()` here. The `episode_buffer` is going to be cleared after calling `_conver_episode_to_batch_major`.

garage/sampler/stateful_pool.py
CatherineSue: Done. Will updated on next push.

CatherineSue:
# HER+DDPG Benchmarks for Mujoco1M

## FetchReach-v1
fetchreach-v1_benchmark

garage/tf/samplers/off_policy_vectorized_sampler.py
```diff
@@ -0,0 +1,155 @@
...
+                    observation=obs,
+                    action=actions,
+                    goal=d_g,
+                    achieved_goal=a_g,
+                    terminal=dones,
+                    **info_dict,
+                )
+
+            if self.algo.replay_buffer_type == Buffer.HER and (
+                    rollout == self.algo.max_path_length - 1):
```
krzentner: If I understand correctly, this only constructs Hindsight samples from the very end of the rollout.
However, didn't the HER paper show that it's better to generate hindsight samples throughout the rollout? Has this already been discussed somewhere?

garage/tf/samplers/off_policy_vectorized_sampler.py
```diff
@@ -0,0 +1,155 @@
...
+                if running_paths[idx] is None:
+                    running_paths[idx] = dict(
+                        rewards=[],
+                        env_infos=[],
+                    )
+                running_paths[idx]["rewards"].append(reward)
+                running_paths[idx]["env_infos"].append(env_info)
+
+                if self.algo.replay_buffer_type == Buffer.REGULAR:
+                    add_to_paths = done or (
```
krzentner: Maybe I don't understand the method, but why don't we add to the paths when done if using HER?
I might just be confused by the bigger picture somehow, but it looks like this logic can be simplified?

garage/tf/q_functions/continuous_mlp_q_function.py
```diff
@@ -59,19 +57,22 @@ def __init__(self,
...
+        defined in garage.tf.
 
         Args:
             reuse: A bool indicates whether reuse variables in the same scope.
-            custom_getter: A customized getter object used to get variables.
             trainable: A bool indicates whether variables are trainable.
         """
-        with tf.variable_scope(
-                self.name, reuse=reuse, custom_getter=custom_getter):
+        with tf.variable_scope(name, reuse=reuse):
```
ryanjulian: I think you should create this scope once in the constructor and then reuse it whenever you need it.

garage/tf/q_functions/continuous_mlp_q_function.py
```diff
@@ -23,7 +22,9 @@ def __init__(self,
                  action_merge_layer=-2,
                  output_nonlinearity=None,
                  input_include_goal=False,
-                 bn=False):
+                 bn=False,
+                 reuse=False,
```
ryanjulian: You need to make this decision for the user. Is the scope reused or not? Your object controls the graph.

garage/tf/q_functions/continuous_mlp_q_function.py
ryanjulian: See the style used in my version of GaussianMLPPolicy.

garage/tf/samplers/off_policy_vectorized_sampler.py
ryanjulian: I don't think it's been discussed.

garage/theano/algos/ddpg.py
```diff
@@ -153,10 +153,15 @@ def start_worker(self):
...
         input_shapes = dims_to_shapes(self.input_dims)
-        pool = ReplayBuffer(
-            buffer_shapes=input_shapes,
-            max_buffer_size=self.replay_pool_size,
-        )
+        buffer_shapes = {
+            key: (self.max_path_length, *input_shapes[key])
+            for key, val in input_shapes.items()
+        }
+        pool = RegularReplayBuffer(
```
ryanjulian: Is there a reason the user can't construct one of these and hand it to you?

garage/replay_buffer/base.py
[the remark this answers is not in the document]
CatherineSue: I'm thinking to get rid of ABC here because there are not many particular methods for HerReplayBuffer.

garage/tf/algos/batch_polopt.py
CatherineSue: We could have a function that builds the sampler that subclasses, i.e.OffPolicyBatchPolopt can override. 
I have doubts about the OffPolicyBatchPolopt class. The `OffPolicyBatchPolopt.train()` is more specific for DDPG and not generalized. It might be better to get rid of OffPolicyBatchPolopt.

garage/tf/algos/trpo.py
```diff
@@ -38,4 +38,5 @@ def __init__(self,
             pg_loss=PGLoss.VANILLA,
             optimizer=optimizer,
             optimizer_args=optimizer_args,
+            name="TRPO",
```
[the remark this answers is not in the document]
CatherineSue: I don't get the question.

garage/tf/samplers/off_policy_vectorized_sampler.py
CatherineSue: > If I understand correctly, this only constructs Hindsight samples from the very end of the rollout.
Yes. But I don't think it makes a lot difference.
Here is the process of DDPG.
```
for episode in range(1, M):
    for rollout in range(1, T):
        replay_buffer.add_transition(obs, action, reward, goal)
        for train in range(1, N):
            transitions = replay_buffer.sample(buffer_batch_size)
```
If we are using HerReplayBuffer, then in `replay_buffer.sample()`, it constructs Hindsight samples by calling `HerReplayBuffer._her_sample_transitions(    def _her_sample_transitions(episode_batch, sample_batch_size):
)`. It randomly selects `sample_batch_size` samples, then constructs Hindsight samples based on `future` strategy.

CatherineSue: >However, didn't the HER paper show that it's better to generate hindsight samples throughout the rollout? Has this already been discussed somewhere?

Do you mean the algorithm pseudocode in the original paper?
```
for episode in range(1, M):
    for rollout in range(1, T):
        replay_buffer.add_transition(obs, action, reward, goal)
        sample a set of additional goals G:=S
        for g in G:
            repaly_buffer.add_transitions(obs, actions, reward, g)
```
S here is the sample function of the goal in current episode. In the original paper, there are three strategies to sample the goal: `future`, `episode`, `random`, `final`. Experiments show `future` is the best. 
The `future` strategy replays with k random states which come from the same episode as the transition being replayed and were observed *after* it. I don't think it's possible to do this process during s<sub>t</sub> since it needs the goals in s<sub>t+future_offset</sub>.

garage/tf/q_functions/continuous_mlp_q_function.py
CatherineSue: I removed `reuse`. I created the scope in the constructor. I need to move this in `_build_net()` because it is used to create the target network. I call `self.policy._build_net(trainable=False, name="target_policy")` to build the target network graph.

garage/replay_buffer/base.py
```diff
@@ -0,0 +1,140 @@
...
+"""
+
+from abc import ABC, abstractmethod
+
+import numpy as np
+
+from garage.spaces import Dict
+
+
+class ReplayBuffer(ABC):
```
ryanjulian: i think you want 
```suggestion
class ReplayBuffer(metaclass=abc.ABCMeta):
```

garage/tf/algos/__init__.py
```diff
@@ -1,8 +1,11 @@
...
 from garage.tf.algos.ddpg import DDPG
 from garage.tf.algos.npo import NPO
+from garage.tf.algos.off_policy_rl_algorithm import OffPolicyRLAlgorithm
 from garage.tf.algos.ppo import PPO
 from garage.tf.algos.trpo import TRPO
 from garage.tf.algos.vpg import VPG
 
-__all__ = ["BatchPolopt", "DDPG", "NPO", "PPO", "TRPO", "VPG"]
+__all__ = [
+    "OffPolicyRLAlgorithm", "BatchPolopt", "DDPG", "NPO", "PPO", "TRPO", "VPG"
```
ryanjulian: Does this mean that `BatchPolopt` is equivalent to `OnPolicyRLAlgorithm`? Should we then rename it?

garage/tf/algos/ddpg.py
```diff
@@ -36,427 +29,241 @@ class DDPG(RLAlgorithm):
...
+            self.f_train_qf = f_train_qf
             self.f_init_target = f_init_target
             self.f_update_target = f_update_target
-            self.replay_buffer = replay_buffer
-            self.target_critic = target_critic
-            self.target_actor = target_actor
 
-    def _learn(self):
+    @overrides
+    def train(self, sess=None):
```
ryanjulian: Are you sure that this can't go into `OffPolicyRLAlgorithm`?

garage/tf/policies/continuous_mlp_policy.py
ryanjulian: Let's leave dealing with this for later when we get rid of layers.

garage/tf/policies/continuous_mlp_policy.py
ryanjulian: Leave for later.

garage/replay_buffer/her_replay_buffer.py
```diff
@@ -63,7 +60,7 @@ def _her_sample_transitions(episode_batch, sample_batch_size):
             for k, rk in zip(["next_achieved_goal", "goal"],
                              list(reward_params_keys)[:-1])
         }
-        reward_params["info"] = info
+        reward_params["info"] = {}
```
zhanpenghe: I noticed that info is not included in HER replay buffer. I think this is useful for reward computation for some cases.

krzentner: What's the reason for this change? It doesn't look correct to me.

garage/theano/algos/ddpg.py
```diff
@@ -37,11 +35,11 @@ def __init__(self,
                  policy,
                  qf,
                  es,
+                 pool,
```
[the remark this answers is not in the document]
krzentner: It definitely makes sense to move it out, since we don't want the algorithm to care (much) about the replay buffer type.
A default of None (which produces a SimpleReplayBuffer), sounds reasonable to me. However, I don't think that needs to go in this pull request.

garage/replay_buffer/base.py
```diff
@@ -0,0 +1,140 @@
...
+
+    @abstractmethod
+    def sample(self, batch_size):
+        """Sample a transition of batch_size."""
+        raise NotImplementedError
+
+    def add_transition(self, **kwargs):
+        """Add one transition into the replay buffer."""
+        for key, value in kwargs.items():
+            self._episode_buffer[key].append(kwargs[key])
```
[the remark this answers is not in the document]
krzentner: Yeah, please simplify as Zhanpeng is suggesting.

garage/tf/algos/__init__.py
CatherineSue: Yes. `BatchPolopt` refers to all the batch sampling based policy gradient algorithms. But I think it can be used for most on-policy algorithms. (I don't have much knowledge on rl algorithms that are not policy gradient based.)

garage/tf/algos/ddpg.py
CatherineSue: I don't think this can for now. The training process here is more customized to actor-critic algorithms, i.e. it calculates TD error. This doesn't fit DQN at least. We can generalize this later.

garage/replay_buffer/base.py
CatherineSue: Done.

garage/replay_buffer/base.py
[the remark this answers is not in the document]
CatherineSue: Done.

garage/replay_buffer/her_replay_buffer.py
CatherineSue: For now all the GoalEnv.compute_reward in gym doesn't use information in `info`, though it needs this parameter. 

The reason I don't store `info` in HER replay buffer is, when construct replay buffer, we need to get all the keys and value shapes. To get all the keys in info, I didn't find other way but call
```
env.reset()
_, _, _, info = env.step(env.action_space.sample())
```
But is it proper to call `env.reset()` in the constructor of replay buffer?

Issue #319, opened 2018-09-06
Make box2d fix consistent across scripts/setup_foo.sh


ryanjulian:
https://github.com/rlworkgroup/garage/blob/0d373273116e97fb502ee5ecbff571fe0ed6f80b/scripts/setup_linux.sh#L220

`pip install Box2D-kengz --no-binary Box2D-kengz` is in setup_osx.sh

ryanjulian:
https://github.com/rlworkgroup/garage/blob/master/docker/Dockerfile.ci#L91

ryanjulian:
ref #319

naeioi:
This issue is fixed by #335 . Closed.

Issue #293, opened 2018-08-17
Calling render() on Box2D envs results in a segfault


ryanjulian:
Tested with gym.make("BipedalWalker-v2")

naeioi:
Tested on box2d-py=2.3.6 on Ubuntu, Box2D-kengz=2.3.3 on OSX (both module are latest, but Box2D-kengz is not maintained since Apr 2017). Both rendering work as expected.

Issue #250, opened 2018-08-01
Missing attribute 'RAND_LIMIT' in modules Box2D


gonzaiva:
The following traceback is produced when running `examples/tf/trpo_cartpole.py`:
```
  File "examples/tf/trpo_cartpole.py", line 3, in <module>
    from garage.envs.box2d import CartpoleEnv
  File "/home/aigonzal/ivanWorkspace/garage/garage/envs/box2d/__init__.py", line 1, in <module>
    from garage.envs.box2d.box2d_env import Box2DEnv
  File "/home/aigonzal/ivanWorkspace/garage/garage/envs/box2d/box2d_env.py", line 9, in <module>
    from garage.envs.box2d.box2d_viewer import Box2DViewer
  File "/home/aigonzal/ivanWorkspace/garage/garage/envs/box2d/box2d_viewer.py", line 1, in <module>
    from Box2D import b2ContactListener
  File "/home/aigonzal/anaconda2/envs/garage/lib/python3.6/site-packages/Box2D/__init__.py", line 20, in <module>
    from .Box2D import *
  File "/home/aigonzal/anaconda2/envs/garage/lib/python3.6/site-packages/Box2D/Box2D.py", line 434, in <module>
    RAND_LIMIT = _Box2D.RAND_LIMIT
```
It seems the problem is solved by running:
`pip install box2d-py`

ryanjulian:
This is documented in dockerfile/Dockerfile.ci. See the do

The solution is:
```sh
conda activate garage
pip uninstall -y Box2D Box2D-kengz box2d-py
pip install Box2D
```

See https://github.com/openai/gym/issues/100 and https://github.com/pybox2d/pybox2d/issues/82

gonzaiva:
Yes, I just noticed with `grep -n -r "box2d-py"` in garage.

ryanjulian:
We should probably keep this open until gym/box2d fix the upstream bug.

gonzaiva:
Okay, in the meant time, the patch has been implemented in #258

ryanjulian:
ref #319

naeioi:
This issue is fixed in #335 . Closed.

PR #335, opened 2018-10-08
Fix broken Box2D dependency in fresh install

### LINKED ISSUE #334: Box2D breaks in fresh installation on Ubuntu
Box2D breaks in freash installation on Ubuntu. Running `examples/tf/trpo_cartpole.py` triggers following traceback. 

```
python3 examples/tf/trpo_cartpole.py 
-----
  File "examples/tf/trpo_cartpole.py", line 3, in <module>
    from garage.envs.box2d import CartpoleEnv
  File "/media/kzhu/OS/Users/naeio/Documents/prog/18USC/resl/garage/garage/envs/box2d/__init__.py", line 1, in <module>
    from garage.envs.box2d.box2d_env import Box2DEnv
  File "/media/kzhu/OS/Users/naeio/Documents/prog/18USC/resl/garage/garage/envs/box2d/box2d_env.py", line 9, in <module>
    from garage.envs.box2d.box2d_viewer import Box2DViewer
  File "/media/kzhu/OS/Users/naeio/Documents/prog/18USC/resl/garage/garage/envs/box2d/box2d_viewer.py", line 1, in <module>
    from Box2D import b2ContactListener
  File "/home/kzhu/miniconda2/envs/garage/lib/python3.6/site-packages/Box2D/__init__.py", line 20, in <module>
    from .Box2D import *
  File "/home/kzhu/miniconda2/envs/garage/lib/python3.6/site-packages/Box2D/Box2D.py", line 435, in <module>
    _Box2D.RAND_LIMIT_swigconstant(_Box2D)
AttributeError: module '_Box2D' has no attribute 'RAND_LIMIT_swigconstant'
```

This issue is mentioned in https://github.com/openai/gym/issues/100 and resolved using https://github.com/openai/gym/issues/100#issuecomment-220940994.

naeioi:
Hotfix of Box2D python module in linux install script. Fix #334 .


scripts/setup_linux.sh
```diff
@@ -219,7 +219,7 @@ conda activate garage
   # See https://github.com/openai/gym/issues/100
   # See https://github.com/pybox2d/pybox2d/issues/82
   pip uninstall -y Box2D Box2D-kengz box2d-py
-  pip install Box2D
+  pip install Box2D box2d-py
```
ryanjulian: i think you need to pick one, because these have the same python namespace.

See the issue links above for info on the dumpster going on with gym+Box2D right now.

We've implemented a couple fixes for this across the install scripts/Dockerfile. Can you verify which fix actually works and make sure they all use the same fix?

ryanjulian:
ref https://github.com/rlworkgroup/garage/issues/293

scripts/setup_linux.sh
naeioi: OK I will check

naeioi:
Now use box2d-py>=2.3.4 on both Linux and OSX. Tested on both platforms.

environment.yml
```diff
@@ -15,7 +15,8 @@ dependencies:
         # Please keep alphabetized
         - awscli
         - boto3
-        - Box2D>=2.3.2
+        # box2d-py>=2.3.4 works both on Linux and OSX.
```
ryanjulian: you can remove this

scripts/setup_linux.sh
```diff
@@ -216,10 +216,15 @@ conda activate garage
   fi
 
   # Fix Box2D install
+  # box2d-py 2.3.4 works on both Linux & OSX and is
+  # actively mantained by OpenAI.
+  # Before that, box2d-py works only on Linux,
+  # and Box2D-kengz works only on OSX.
   # See https://github.com/openai/gym/issues/100
   # See https://github.com/pybox2d/pybox2d/issues/82
+  # See https://github.com/openai/gym/pull/1120 (fixed)
```
ryanjulian: do we still need the step where we remove the other Box2D/box2d packages?

We previously did this because pip found various Box2D packages in the dependency tree and installed them all.

scripts/setup_osx.sh
```diff
@@ -228,10 +228,15 @@ conda activate garage
   pip install -e .
 
   # Fix Box2D install
+  # box2d-py 2.3.4 works on both Linux & OSX and is
+  # actively mantained by OpenAI.
+  # Before that, box2d-py works only on Linux,
+  # and Box2D-kengz works only on OSX.
```
ryanjulian: Same comment. Do we still need the uninstall/reinstall step?

scripts/setup_linux.sh
naeioi: Yeah I agree, they are not necessary. It's a fresh installation in a new conda environment.

jonashen:
Relevant: #319

ryanjulian:
What about Dockerfile.ci?

ryanjulian:
ref #250 #319

naeioi:
Extra pip of box2d-py in setup_foo.sh and Dockerfile.ci have been removed.

ryanjulian:
Looks like your fix doesn't work in the CI https://travis-ci.com/rlworkgroup/garage/builds/87381137

naeioi:
It turns out that gym 0.10.5 installs Box2D-kengz. In the latest release gym 0.10.8, OpenAI replaces Box2D-kengz with box2d-py. 

We have two candidate solutions. In environment.yml, whether we upgrade to gym==0.10.8 and remove box2d-py (it's included in gym), or we manually uninstall Box2D-kengz and install box2d-py like before. 
The first solution is preferable, but need to test if upgrade breaks other things. 

Ref: https://github.com/openai/gym/commit/5c116fb3c91e872505300031d2bd60672b3a6e03#diff-2eeaed663bd0d25b7e608891384b7298L11

ryanjulian:
I think it's preferably to just upgrade gym. We still need to specify a version of box2d in environment.yml, because we have a direct dependency on Box2D in garage.envs.

clarification: box2d-py doesn't come with gym. it's installed by pip as a gym dependency.

naeioi:
SIGINT test failed. I am looking into the problem.

ryanjulian:
SIGINT test is probably unrelated.

gonzaiva:
Keren, while doing some testing in my PR #340, I found out that updating to gym 0.10.8 will impact the testing of gym environments.
There's a bug a in gym 0.10.5 that doesn't close the viewer windows of the MuJoCo environments, and this leads to "out of memory" issues when a lot of environments are ran one after the other in the same process. To fix that, I pushed the following change to forcefully close the windows when the AttributeError exception mentioned [here](https://github.com/openai/gym/issues/1000#issuecomment-382106495) occurs. My fix is [here](https://github.com/rlworkgroup/garage/pull/318/files#diff-c1bb02df5151a785915187090d65c631R47).
For gym 0.10.8, the developers only commented the finish method that caused the exception, so they just hid the bug. The problem is that if no exception is produced, my fix won't work anymore and also [this test assertion](https://github.com/rlworkgroup/garage/blob/f3b5ec1d97485a2f164698574c205bccebae17ec/tests/helpers.py#L50) will fail.
In order to update to gym 0.10.8 I'm [changing](https://github.com/rlworkgroup/garage/pull/340/files#diff-c1bb02df5151a785915187090d65c631R45) my detection schema from checking on the exception to checking on the package (gym.envs.mujoco or gym.envs.robotics) where the environment is from, but the exception in the test still has to be removed to avoid assertion errors.
You were not aware of this because these tests are run under a cron job in Travis on the master branch. Those tests are marked as Cron in the [build history](https://travis-ci.com/rlworkgroup/garage/builds), for example, [this one](https://travis-ci.com/rlworkgroup/garage/builds/87724641). They were moved into a cron job because they take a long time and were consuming the time for other more relevant tests in our regular PR builds.
**TL;DR:**
Before pushing your PR, please remove [this assertion](https://github.com/rlworkgroup/garage/blob/f3b5ec1d97485a2f164698574c205bccebae17ec/tests/helpers.py#L50) (and the surrounding if statement, since it's now useless) to make the tests pass.

@ryanjulian It would be nice to now your opinion on this as well.

ryanjulian:
@gonzaiva sgtm. all PRs need fix any tests they break.

please coordinate to make sure the build does not break

naeioi:
New issue: CI failed with GLEW segfault after syncing with master branch.
Running `nose2 tests.test_dm_control_env` throws segfault. I believe there are some conflicts with #299.

naeioi:
OK so the problem is that we retrieve latest version of dm_control from their github repo in `environment.yml`, and today (10/15/2018) they updated dependency of Mujoco from 1.5 to 2.0. 

Ref: [dm_control commit](https://github.com/deepmind/dm_control/commit/ae2d628c1437e5cf47211b964fb4039ecb456d46#diff-2eeaed663bd0d25b7e608891384b7298R35)

tests/helpers.py
```diff
@@ -43,10 +42,3 @@ def step_env_with_gym_quirks(test_case,
         # Roundtrip serialization
         round_trip = pickle.loads(pickle.dumps(env))
         assert round_trip.env.spec == env.env.spec
-
-    if not spec.id in KNOWN_GYM_CLOSE_BROKEN:
-        env.close()
```
gonzaiva: You still need to call env.close() here.

gonzaiva:
I pushed my PR #340, so I'll be checking on this one once it gets merged to see if the cron job is still working.

ryanjulian:
@gonzaiva please unblock this if it's ready

tests/helpers.py
```diff
@@ -46,7 +46,3 @@ def step_env_with_gym_quirks(test_case,
 
     if not spec.id in KNOWN_GYM_CLOSE_BROKEN:
```
gonzaiva: Remove this line but keep line 48: we want to call close for all environments.

gonzaiva:
Sorry for my late response. Please address my small comment and it's all good.

tests/helpers.py
```diff
@@ -1,6 +1,5 @@
 import pickle
 
-from tests.quirks import KNOWN_GYM_CLOSE_BROKEN
```
ryanjulian: I think this will leave dead code? Best to remove `KNOWN_GYM_CLOSE_BROKEN` if it's now unused.

naeioi: I think I already removed?

naeioi:
pre-commit has a breaking updates today that prevents our installation. I have opened an issue in https://github.com/pre-commit/pre-commit/issues/853.

tests/helpers.py
naeioi: @gonzaiva Can you confirm that we are not going to use `KNOWN_GYM_CLOSE_BROKEN` anywhere now and in the future.

ryanjulian: No need to preemptively retain dead code. If it has no references, it is dead and should be removed. If it's needed in the future, it can be revived from the commit history.

naeioi: OK i got it.

ryanjulian:
Not sure if it's hit deployment, but the fix was merged last night https://github.com/conda-forge/pre_commit-feedstock/pull/17/files

PR #336, opened 2018-10-08
Updated Docs so they mention that MuJoCo is required

pelillian:
Quick docs update for clarification and easy install for external collaborators.

### DIFF
--- a/docs/user/installation.rst
+++ b/docs/user/installation.rst
@@ -10,21 +10,23 @@ Express Install
 
 The fastest way to set up dependencies for garage is via running the setup script.
 
-Clone this repo and navigate to its directory.
+Clone our repo (https://github.com/rlworkgroup/garage) and navigate to its directory.
 
-A MuJoCo key is required for installation.
+A MuJoCo key is required for installation. You can get one here: https://www.roboti.us/license.html
+
+Make sure you run these scripts from the root directory of the repo, not from the scripts directory.
 
 - On Linux, run the following:
 
 .. code-block:: bash
 
-    ./scripts/setup_linux.sh --mjkey path-to-your-mjkey.txt
+    ./scripts/setup_linux.sh --mjkey path-to-your-mjkey.txt --modify-bashrc
 
 - On Mac OS X, run the following:
 
 .. code-block:: bash
 
-    ./scripts/setup_osx.sh --mjkey path-to-your-mjkey.txt
+    ./scripts/setup_osx.sh --mjkey path-to-your-mjkey.txt --modify-bashrc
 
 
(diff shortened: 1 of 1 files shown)


ryanjulian:
Please pick some appropriate reviewers when you submit a PR

docs/user/installation.rst
```diff
@@ -10,23 +10,28 @@ Express Install
...
-    ./scripts/setup_osx.sh
+    ./scripts/setup_osx.sh --mjkey path-to-your-mjkey.txt
+
 
 The script sets up a conda environment, which is similar to :code:`virtualenv`. To start using it, run the following:
 
 .. code-block:: bash
 
-    source activate garage
+    conda activate garage
```
jonashen: Let's mention that `conda activate garage` only works for conda versions 4.4+, while those older than 4.4 should use `source activate <envname>`.

pelillian: I changed it back to source activate garage for simplicity. You think I should include both?

jonashen: Since we are eventually phasing out conda anyway, just keep `source` for simplicity.

docs/user/installation.rst
```diff
@@ -10,17 +10,22 @@ Express Install
 
 The fastest way to set up dependencies for garage is via running the setup script.
 
+Clone this repo and navigate to its directory.
+
+A MuJoCo key is required for installation.
```
ryanjulian: Can you provide a link for where people can get a MuJoCo key?

docs/user/installation.rst
```diff
@@ -10,17 +10,22 @@ Express Install
 
 The fastest way to set up dependencies for garage is via running the setup script.
 
+Clone this repo and navigate to its directory.
```
ryanjulian: Which repo?

These docs appear at rlgarage.readthedocs.io. The user doesn't necessarily have any context to the GitHub repo.

docs/user/installation.rst
```diff
@@ -10,17 +10,22 @@ Express Install
...
+
+A MuJoCo key is required for installation.
+
 - On Linux, run the following:
 
 .. code-block:: bash
 
-    ./scripts/setup_linux.sh
+    ./scripts/setup_linux.sh --mjkey path-to-your-mjkey.txt
```
ryanjulian: You should probably mention the other options, such as the one which amends bashrc for you. I think you would want that for a truly "express" install.

gonzaiva: Peter, you can get the argument name by running `./scripts/setup_linux.sh -h`. Also, could you please emphasize users to run the script at the root at garage?
I see you wrote the command to be executed at the root of the garage folder, but I have noticed sometimes people run the script inside the `scripts` folder and it won't work that way.

PR #340, opened 2018-10-09
Import mujoco_py in GarageEnv only if necessary

gonzaiva:
Currently the viewer window for MuJoCo environments is not closed
correctly, leading to out of memory issues. This bug is on the side of
the gym environments, but until gym provides a fix, we're closing the
viewer window directly.
However, by importing the mujoco_py package into the environment base of
garage, this enforces the use of MuJoCo for all environments, which is
not necessary for non-MuJoCo environments.
That's why the import for the mjviewer was moved into the exception
where the closing of the mjviewer is performed.

### DIFF
--- a/garage/envs/base.py
+++ b/garage/envs/base.py
@@ -5,10 +5,10 @@ import warnings
 
 import glfw
 import gym
-from mujoco_py.mjviewer import MjViewer
 
 from garage.core import Parameterized
 from garage.core import Serializable
+from tests.quirks import KNOWN_GYM_NOT_CLOSE_VIEWER
 
 
 class GarageEnv(gym.Wrapper, Parameterized, Serializable, metaclass=ABCMeta):
@@ -42,14 +42,10 @@ class GarageEnv(gym.Wrapper, Parameterized, Serializable, metaclass=ABCMeta):
         Returns:
             None
         """
-        try:
-            self.env.close()
-        except AttributeError as e:
-            if str(e) == "'MjViewer' object has no attribute 'finish'":
-                self._close_viewer_window()
-            raise
+        self._close_mjviewer_window()
+        self.env.close()
 
-    def _close_viewer_window(self):
+    def _close_mjviewer_window(self):
         """
         Close the MjViewer window.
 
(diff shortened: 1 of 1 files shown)


garage/envs/base.py
```diff
@@ -61,6 +60,7 @@ def _close_viewer_window(self):
         search in those as well.
         This method can be removed once OpenAI solves the issue.
```
ryanjulian: can you provide a link to an openai github issue?

garage/envs/base.py
```diff
@@ -61,6 +60,7 @@ def _close_viewer_window(self):
...
         """
+        from mujoco_py.mjviewer import MjViewer
```
jonashen: Is this the only solution? The CI / Ryan will probably complain about import formatting.

jonashen: nvm looks like Ryan approves

garage/envs/base.py
jonashen: And maybe a comment about why we're importing it here (so when we read the code in the future we won't forget)

garage/envs/base.py
gonzaiva: Well, unless we had a MuJoCo-Env base class, I don't know where to move the method _close_viewer_window, but I'm open to suggestions since I also don't like to have imports at the middle of the code.
[This](https://travis-ci.com/rlworkgroup/garage/builds/87391938) build didn't complain.

garage/envs/base.py
gonzaiva: @ryanjulian The issue has been reported to OpenAI [here](https://github.com/openai/gym/issues/1000#issuecomment-382106495), but it seems that the only change they have done from our current version 0.10.5 to 0.10.8 is [commenting](https://github.com/openai/gym/blob/70ca46a9685de333493722ee74d0cd51782b2b44/gym/envs/mujoco/mujoco_env.py#L122) (diff is [here](https://github.com/openai/gym/compare/c8f8149...70ca46a#diff-3f48026043ba85dc635adababe52094c)) the finish method in the viewer so the exception is not thrown. The guys at gym closed the issue, so if you agree I can create a new one to specifically address this concern, since the issue I referenced above comments on other things as well.
Regarding the exception I mentioned above, we're using it to detect when we have to force the closing ourselves, but for the new changes, the exception is not thrown any more and my current implementation won't close the viewer windows.
I actually tested this change (upgrading gym to v0.10.8), and the windows remain open.
If we update gym as in Keren's PR #335, we'll see the "out of memory" issue again in the cron jobs.
I will use the list of packages defined [here](https://github.com/rlworkgroup/garage/blob/f3b5ec1d97485a2f164698574c205bccebae17ec/tests/quirks.py#L90-L91) for not closing environments to detect the environments with not closing windows to avoid this dependency.
@jonashen Okay, I will add a comment to explain why the import is in not in the header.

PR #341, opened 2018-10-12
Implemented MLP class without layers + unit test

ahtsan:
Implemented MLP class without garage.tf.core.layers. Also with an unit test.
However it does not handle batch normalization and weight normalization yet. (It does now)

### DIFF
--- a/garage/tf/core/networks.py
+++ b/garage/tf/core/networks.py
@@ -28,6 +28,11 @@ def mlp(input_var,
                 bias_initializer=hidden_b_init,
                 name="hidden_%d" % idx,
             )
+            if batch_normalization:
+                l_hid = tf.layers.batch_normalization(
+                    l_hid, training=is_training)
+            if weight_normalization:
+                l_hid = tf.contrib.layers.layer_norm(l_hid)
         l_out = tf.layers.dense(
             inputs=l_hid,
             units=output_dim,
@@ -37,5 +42,7 @@ def mlp(input_var,
             name="output")
         if batch_normalization:
             l_out = tf.layers.batch_normalization(l_out, training=is_training)
+        if weight_normalization:
+            l_out = tf.contrib.layers.layer_norm(l_out)
 
     return l_out
--- a/garage/tf/core/networks.py
+++ b/garage/tf/core/networks.py
@@ -1,71 +1,52 @@
 import tensorflow as tf
-from garage.core import Serializable
 
 
-class MLPs(Serializable):
-    def __init__(
-            self,
-            output_dim,
-            hidden_sizes,
-            hidden_nonlinearity,
-            output_nonlinearity,
-            name=None,
-            hidden_w_init=tf.contrib.layers.xavier_initializer(),
-            hidden_b_init=tf.zeros_initializer(),
-            output_w_init=tf.contrib.layers.xavier_initializer(),
-            output_b_init=tf.zeros_initializer(),
-            input_var=None,
-            input_shape=None,
-            batch_normalization=False,
-            weight_normalization=False,
-    ):
+def mlp(input_var,
+        output_dim,
+        hidden_sizes,
+        hidden_nonlinearity=tf.nn.relu,
+        output_nonlinearity=None,
+        name="mlp",
+        hidden_w_init=tf.contrib.layers.xavier_initializer(),
+        hidden_b_init=tf.zeros_initializer(),
+        output_w_init=tf.contrib.layers.xavier_initializer(),
+        output_b_init=tf.zeros_initializer(),
+        batch_normalization=False,
+        weight_normalization=False,
+        is_training=True,
+        debug=False):
 
-        Serializable.quick_init(self, locals())
+    _layers = []
 
-        with tf.variable_scope(name):
-            if input_var is None:
-                l_in = tf.placeholder(
-                    tf.float32,
-                    shape=(None, ) + input_shape,  # dynamic batch dimension
-                    name="input")
-            else:
-                l_in = input_var
-            self._layers = [l_in]
-            l_hid = l_in
-            for idx, hidden_size in enumerate(hidden_sizes):
-                l_hid = tf.layers.dense(
-                    inputs=l_hid,
-                    units=hidden_size,
-                    activation=hidden_nonlinearity,
-                    kernel_initializer=hidden_w_init,
-                    bias_initializer=hidden_b_init,
-                    name="hidden_%d" % idx)
-
-                self._layers.append(l_hid)
-            l_out = tf.layers.dense(
+    with tf.variable_scope(name):
+        l_hid = input_var
+        _layers = [l_hid]
+        if batch_normalization:
+            l_hid = tf.layers.batch_normalization(l_hid, training=is_training)
+        for idx, hidden_size in enumerate(hidden_sizes):
+            l_hid = tf.layers.dense(
                 inputs=l_hid,
-                units=output_dim,
-                activation=output_nonlinearity,
-                kernel_initializer=output_w_init,
-                bias_initializer=output_b_init,
-                name="output")
+                units=hidden_size,
+                activation=hidden_nonlinearity,
+                kernel_initializer=hidden_w_init,
+                bias_initializer=hidden_b_init,
+                name="hidden_%d" % idx,
+            )
 
-            self._layers.append(l_out)
-            self._l_in = l_in
-            self._l_out = l_out
+            _layers.append(l_hid)
+        l_out = tf.layers.dense(
+            inputs=l_hid,
+            units=output_dim,
+            activation=output_nonlinearity,
+            kernel_initializer=output_w_init,
+            bias_initializer=output_b_init,
+            name="output")
+        if batch_normalization:
+            l_out = tf.layers.batch_normalization(l_out, training=is_training)
+        _layers.append(l_out)
+        _l_out = l_out
 
-    @property
-    def input_layer(self):
-        return self._l_in
-
-    @property
-    def output_layer(self):
-        return self._l_out
-
-    @property
-    def input_var(self):
-        return self._l_in
-
-    @property
-    def layers(self):
-        return self._layers
+    if debug:
+        return _l_out, _layers
+    else:
+        return _l_out
--- a/garage/tf/core/networks.py
+++ b/garage/tf/core/networks.py
@@ -1,0 +1,71 @@
+import tensorflow as tf
+from garage.core import Serializable
+
+
+class MLPs(Serializable):
+    def __init__(
+            self,
+            output_dim,
+            hidden_sizes,
+            hidden_nonlinearity,
+            output_nonlinearity,
+            name=None,
+            hidden_w_init=tf.contrib.layers.xavier_initializer(),
+            hidden_b_init=tf.zeros_initializer(),
+            output_w_init=tf.contrib.layers.xavier_initializer(),
+            output_b_init=tf.zeros_initializer(),
+            input_var=None,
+            input_shape=None,
+            batch_normalization=False,
+            weight_normalization=False,
+    ):
+
+        Serializable.quick_init(self, locals())
+
+        with tf.variable_scope(name):
+            if input_var is None:
+                l_in = tf.placeholder(
+                    tf.float32,
+                    shape=(None, ) + input_shape,  # dynamic batch dimension
+                    name="input")
+            else:
+                l_in = input_var
+            self._layers = [l_in]
+            l_hid = l_in
+            for idx, hidden_size in enumerate(hidden_sizes):
+                l_hid = tf.layers.dense(
+                    inputs=l_hid,
+                    units=hidden_size,
+                    activation=hidden_nonlinearity,
+                    kernel_initializer=hidden_w_init,
+                    bias_initializer=hidden_b_init,
+                    name="hidden_%d" % idx)
+
+                self._layers.append(l_hid)
+            l_out = tf.layers.dense(
+                inputs=l_hid,
+                units=output_dim,
+                activation=output_nonlinearity,
+                kernel_initializer=output_w_init,
+                bias_initializer=output_b_init,
+                name="output")
+
+            self._layers.append(l_out)
+            self._l_in = l_in
+            self._l_out = l_out
+
+    @property
+    def input_layer(self):
+        return self._l_in
+
+    @property
+    def output_layer(self):
+        return self._l_out
+
+    @property
+    def input_var(self):
+        return self._l_in
+
+    @property
+    def layers(self):
+        return self._layers
--- a/garage/tf/core/networks.py
+++ b/garage/tf/core/networks.py
@@ -1,24 +1,51 @@
+"""
+Multilayer Perceptrons (MLP) with tensorflow as the only dependency.
+
+The module contains MLP which serves as the base of all networks.
+It aims to replace existing implementation of MLP class
+(garage.tf.core.network), which is under development.
+"""
 import tensorflow as tf
 
 
 def mlp(input_var,
         output_dim,
         hidden_sizes,
+        name,
         hidden_nonlinearity=tf.nn.relu,
-        output_nonlinearity=None,
-        name="mlp",
         hidden_w_init=tf.contrib.layers.xavier_initializer(),
         hidden_b_init=tf.zeros_initializer(),
+        output_nonlinearity=None,
         output_w_init=tf.contrib.layers.xavier_initializer(),
         output_b_init=tf.zeros_initializer(),
-        batch_normalization=False,
-        weight_normalization=False,
-        is_training=True):
+        layer_normalization=False):
+    """
+    MLP function.
 
+    Args:
+        input_var: Input tf.Tensor to the MLP.
+        output_dim: Dimension of the network output.
+        hidden_sizes: Output dimension of dense layer(s).
+        name: variable scope of the mlp.
+        hidden_nonlinearity: Activation function for
+                    intermediate dense layer(s).
+        hidden_w_init: Initializer function for the weight
+                    of intermediate dense layer(s).
+        hidden_b_init: Initializer function for the bias
+                    of intermediate dense layer(s).
+        output_nonlinearity: Activation function for
+                    output dense layer.
+        output_w_init: Initializer function for the weight
+                    of output dense layer(s).
+        output_b_init: Initializer function for the bias
+                    of output dense layer(s).
+        layer_normalization: Bool for using layer normalization or not.
+
+    Return:
+        The output tf.Tensor of the MLP
+    """
     with tf.variable_scope(name):
         l_hid = input_var
-        if batch_normalization:
-            l_hid = tf.layers.batch_normalization(l_hid, training=is_training)
         for idx, hidden_size in enumerate(hidden_sizes):
             l_hid = tf.layers.dense(
                 inputs=l_hid,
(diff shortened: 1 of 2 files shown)


tests/garage/tf/core/test_networks.py
```diff
@@ -0,0 +1,68 @@
+import unittest
+
+import tensorflow as tf
+from garage.tf.core.networks import MLPs
+import numpy as np
+
+
+class TestNetworks(unittest.TestCase):
```
ryanjulian: There is a test fixture which sets up a TF session for you.

tests/garage/tf/core/test_networks.py
```diff
@@ -0,0 +1,68 @@
...
+    def test_networks(self):
```
ryanjulian: You may want to take a look at the test suite for lasagne to get some ideas for tests you can run on NN layers.

https://github.com/Lasagne/Lasagne

ryanjulian:
I would like to hear what people think about implementing layers (e.g. MLP) as classes rather than functions. I am partial to using functions, and only having full models (e.g. GaussanMLP) as classes. Anson does it here with a class.

ahtsan:
I think the pros of having it as a class is
-we can set more properties with intuitive name, e.g. input_shape, input_shape_flat and handle the transformation internally (if it were a function, we have to trace back to the input placeholder and see what's the input shape, and also use util functions to format it). By this, we can abstract the MLP class better. The same may also be applied to other networks.
-for full model class, some of the property can be reused (GaussianMLP.input_shape can actually be GaussianMLP.mean_network.input_shape which is a MLP)
My idea is modelling networks as a class with standardized properties is easier to read and more intuitive, especially when it is likely that there will be more people using garage.

garage/tf/core/networks.py
```diff
@@ -0,0 +1,71 @@
...
+            hidden_nonlinearity,
+            output_nonlinearity,
+            name=None,
+            hidden_w_init=tf.contrib.layers.xavier_initializer(),
+            hidden_b_init=tf.zeros_initializer(),
+            output_w_init=tf.contrib.layers.xavier_initializer(),
+            output_b_init=tf.zeros_initializer(),
+            input_var=None,
+            input_shape=None,
+            batch_normalization=False,
```
zhanpenghe: Batch norm is not actually implemented yet. Could you please also add it?

zhanpenghe: Batch norm is not actually implemented yet. Could you please also add it?

garage/tf/core/networks.py
```diff
@@ -0,0 +1,71 @@
+import tensorflow as tf
+from garage.core import Serializable
+
+
+class MLPs(Serializable):
```
zhanpenghe: I think it would be better if we can abstract out a model with appropriate API's for different models like GaussianMLP, LSTM and etc.

ryanjulian:
I prefer common layers/networks to be simple utility functions, and whole models to be classes. I think otherwise we risk reinventing layers/keras.

The point of getting rid of the layers library is to allow everyone to use raw tensorflow, because nobody really agrees on an interface for neural network components. If we make an MLP a class, then we have to define a type/interface for MLPs--which is just the layers library all over again.

If we make an MLP a function, then anyone can use it and they only need to be concerned with inputs and outputs. IMO we lose nothing.

I think there's even an argument to be made for making GaussianMLP a function, rather than a model. These utilities define subgraphs which we want to reuse over-and-over. They don't necessarily need to have state.

ryanjulian:
I really want @krzentner 's opinion on this PR, since it involves questions of how we're going to organize the PyTorch and TF trees in the future.

Sorry for the hold-up @ahtsan

garage/tf/core/networks.py
```diff
@@ -0,0 +1,52 @@
...
+        for idx, hidden_size in enumerate(hidden_sizes):
+            l_hid = tf.layers.dense(
+                inputs=l_hid,
+                units=hidden_size,
+                activation=hidden_nonlinearity,
+                kernel_initializer=hidden_w_init,
+                bias_initializer=hidden_b_init,
+                name="hidden_%d" % idx,
+            )
+
```
CatherineSue: There should have batch_normalization.

tests/garage/tf/core/test_networks.py
```diff
@@ -0,0 +1,75 @@
+import tensorflow as tf
+from garage.tf.core.networks import mlp
+import numpy as np
```
CatherineSue: Please group imports and reorder them. There should be error msgs in CI.
https://www.python.org/dev/peps/pep-0008/#imports

garage/tf/core/networks.py
```diff
@@ -0,0 +1,52 @@
...
+        hidden_sizes,
+        hidden_nonlinearity=tf.nn.relu,
+        output_nonlinearity=None,
+        name="mlp",
+        hidden_w_init=tf.contrib.layers.xavier_initializer(),
+        hidden_b_init=tf.zeros_initializer(),
+        output_w_init=tf.contrib.layers.xavier_initializer(),
+        output_b_init=tf.zeros_initializer(),
+        batch_normalization=False,
+        weight_normalization=False,
```
CatherineSue: `weight_normalization` is unused.

garage/tf/core/networks.py
```diff
@@ -0,0 +1,52 @@
...
+            inputs=l_hid,
+            units=output_dim,
+            activation=output_nonlinearity,
+            kernel_initializer=output_w_init,
+            bias_initializer=output_b_init,
+            name="output")
+        if batch_normalization:
+            l_out = tf.layers.batch_normalization(l_out, training=is_training)
+        _layers.append(l_out)
+        _l_out = l_out
```
CatherineSue: Why not just return `l_out`?

ahtsan: I was trying to get all the layers for unit test, since it is not a class anymore, I can only put it in the return value. If that is not needed for testing, I can remove it.

ahtsan: for some reason the unit test was not included in the commit... I will add that back today

tests/garage/tf/core/test_networks.py
krzentner: Please fix this.

garage/tf/core/networks.py
```diff
@@ -0,0 +1,48 @@
+import tensorflow as tf
+
+
+def mlp(input_var,
```
krzentner: There's been some discussion towards using the `tf.keras` namespace instead of `tf.contrib`, but I don't think it matters in this case.

garage/tf/core/networks.py
```diff
@@ -0,0 +1,48 @@
...
+                l_hid = tf.contrib.layers.layer_norm(l_hid)
+        l_out = tf.layers.dense(
+            inputs=l_hid,
+            units=output_dim,
+            activation=output_nonlinearity,
+            kernel_initializer=output_w_init,
+            bias_initializer=output_b_init,
+            name="output")
+        if batch_normalization:
+            l_out = tf.layers.batch_normalization(l_out, training=is_training)
```
ryanjulian: I'm not sure that we have a clean way of setting this variable within algorithms, unfortunately. This needs to be thought through before it's included or required.

garage/tf/core/networks.py
```diff
@@ -0,0 +1,48 @@
...
+        if batch_normalization:
+            l_hid = tf.layers.batch_normalization(l_hid, training=is_training)
+        for idx, hidden_size in enumerate(hidden_sizes):
+            l_hid = tf.layers.dense(
+                inputs=l_hid,
+                units=hidden_size,
+                activation=hidden_nonlinearity,
+                kernel_initializer=hidden_w_init,
+                bias_initializer=hidden_b_init,
+                name="hidden_%d" % idx,
```
ryanjulian: please use Python 3 `.format()` syntax. `%` syntax is considered deprecated in Python 3.

garage/tf/core/networks.py
```diff
@@ -0,0 +1,48 @@
...
+        name="mlp",
+        hidden_w_init=tf.contrib.layers.xavier_initializer(),
+        hidden_b_init=tf.zeros_initializer(),
+        output_w_init=tf.contrib.layers.xavier_initializer(),
+        output_b_init=tf.zeros_initializer(),
+        batch_normalization=False,
+        weight_normalization=False,
+        is_training=True):
+
+    with tf.variable_scope(name):
```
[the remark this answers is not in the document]
ryanjulian: don't forget to include the input tensors in the name_scope call, as well as using their style of defining a default name (i.e. your function signature will have `name=None`)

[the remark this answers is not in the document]
ahtsan: Why do we need name_scope here? Since we would always have a variable scope wrapping the mlp and both the variables and tensors would be already under the variable scope. Is there a case that we need to separate the operations and variables under different scope (when all of them belong to mlp)?

Also is it not a good practice if we put the input placeholder outside the scopes of mlp? In this way the input placeholder can be fed to different mlp to reduce redundancy (e.g. q_network and target q_network).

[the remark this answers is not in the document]
ryanjulian: * Using name scope will create a block in the TensorFlow graph which contains all ops created by your function call. This makes the graph much easier to read when viewed with TensorBoard
* Correct. A utility function for making networks (like this one) should never create a placeholder. It should only produce outputs from inputs. If you're suggesting that my comment implied you need to create a placeholder, I did not intend to say that.

garage/tf/core/networks.py
```diff
@@ -0,0 +1,48 @@
+import tensorflow as tf
+
+
+def mlp(input_var,
+        output_dim,
+        hidden_sizes,
+        hidden_nonlinearity=tf.nn.relu,
+        output_nonlinearity=None,
+        name="mlp",
```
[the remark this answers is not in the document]
ahtsan: Do you mean last argument? Since it needs to be optional, it cannot follow default arguments.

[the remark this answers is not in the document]
ryanjulian: Oops. Perhaps let's make it always the first optional argument.

ahtsan:
screenshot from 2018-10-23 17-06-51
The graph is like this after I provide mlp with a name "mlp" and wrap it with variable scope "mlp_var"

ahtsan:
screenshot from 2018-10-23 17-28-21

> with tf.variable_scope("mlp_var", reuse=True):
    mlp1 = mlp(..., name="mlp")
    mlp2 = mlp(..., name="mlp")

screenshot from 2018-10-23 17-29-36

> with tf.variable_scope("mlp_var", reuse=True):
    mlp1 = mlp(..., name="mlp1")
    mlp2 = mlp(..., name="mlp2")

screenshot from 2018-10-23 17-30-23

> with tf.variable_scope("mlp_var1"):
    mlp1 = mlp(..., name="mlp1")
with tf.variable_scope("mlp_var2"):
    mlp2 = mlp(..., name="mlp2")

ryanjulian:
I think I may have mislead you and you might actually need to use variable_scope. Let's chat tomorrow.

ryanjulian:
Can you plot the TensorBoard examples above, but using `variable_scope()` rather than `name_scope()`?

tests/garage/tf/core/test_networks.py
```diff
@@ -0,0 +1,62 @@
...
+import tensorflow as tf
+
+from garage.tf.core.networks import mlp
+from tests.fixtures import TfGraphTestCase
+
+
+class TestNetworks(TfGraphTestCase):
```
ryanjulian: Please add tests of batch normalization and weight normalization.

See missed lines here https://coveralls.io/builds/19684171/source?filename=garage%2Ftf%2Fcore%2Fnetworks.py#L21

ahtsan: I discussed with Chang and we both agreed that batch normalization should be implemented later, since it is actually not being used at all, and it makes thing a little bit complicated. What do you think?

ahtsan:
Case 1: For 2 same mlps. Name scope outside is better.
screenshot from 2018-10-24 12-30-23
```python
with tf.variable_scope("mlp_var"):
    mlp1 = mlp(..., name="mlp")
    mlp2 = mlp(..., name="mlp")
```
screenshot from 2018-10-24 12-33-24
```python
with tf.name_scope("mlp_var"):
    mlp1 = mlp(..., name="mlp")
    mlp2 = mlp(..., name="mlp")
```
Case 2: 2 different mlps under same scope. Both is fine.
screenshot from 2018-10-24 12-36-45
```python
with tf.variable_scope("mlp_var"):
    mlp1 = mlp(..., name="mlp1")
    mlp2 = mlp(..., name="mlp2")
```
screenshot from 2018-10-24 12-37-43
```python
with tf.name_scope("mlp_var"):
    mlp1 = mlp(..., name="mlp1")
    mlp2 = mlp(..., name="mlp2")
```
Case 3: 2 different mlps under different scope. Both is fine.
screenshot from 2018-10-24 13-15-40
```python
with tf.variable_scope("mlp_var1"):
    mlp1 = mlp(..., name="mlp1")
with tf.variable_scope("mlp_var2"):
    mlp2 = mlp(..., name="mlp2")
```
screenshot from 2018-10-24 13-01-16
```python
with tf.name_scope("mlp_var1"):
    mlp1 = mlp(..., name="mlp1")
with tf.name_scope("mlp_var2"):
    mlp2 = mlp(..., name="mlp2")
```
screenshot from 2018-10-24 13-02-54
Case 4: for same mlp under different name scope:
```python
with tf.name_scope("mlp_var1"):
    mlp1 = mlp(..., name="mlp1")
with tf.name_scope("mlp_var2"):
    mlp2 = mlp(..., name="mlp1")
```
P.S. We can't have same mlp under different variable scope.

garage/tf/core/networks.py
ryanjulian: Please add docstrings for this function using the PEP287/PEP257 style discussed [here](https://github.com/rlworkgroup/garage/blob/master/CONTRIBUTING.md#garage-specific-python-style)

ahtsan:
Is it reasonable to assume user must give a name for variable scope of mlp? Since variable scope can't be None.

ahtsan:
> Is it reasonable to assume user must give a name for variable scope of mlp? Since variable scope can't be None.

Or we set default name to be "mlp"

ryanjulian:
In this case I think the best practice is to make it a required argument (e.g. not a kwarg), rather than having a silent default.

tests/garage/tf/core/test_networks.py
```diff
@@ -0,0 +1,122 @@
...
+            output_dim=self._output_shape,
+            hidden_sizes=(32, 32),
+            hidden_nonlinearity=self.hidden_nonlinearity,
+            name="mlp2",
+            weight_normalization=True)
+
+        self.sess.run(tf.global_variables_initializer())
+
+        mlp_output = self.sess.run(
+            self.mlp_f_w_n, feed_dict={self._input: self.obs_input})
```
ryanjulian: Can  you please assert something about the outputs here?

garage/tf/core/networks.py
```diff
@@ -0,0 +1,55 @@
...
+        hidden_w_init: Initializer function for the weight of intermediate dense layer(s).
+        hidden_b_init: Initializer function for the bias of intermediate dense layer(s).
+        output_w_init: Initializer function for the weight of output dense layer(s).
+        output_b_init: Initializer function for the bias of output dense layer(s).
+        weight_normalization: Bool for using weight normalization or not.
+
+    Return:
+        A MLP tf.Tensor object.
+    """
+    with tf.variable_scope(name, reuse=tf.AUTO_REUSE):
```
ryanjulian: Do not set the reuse policy here. It should be inherited from the parent scope.

garage/tf/core/networks.py
```diff
@@ -0,0 +1,55 @@
...
+        hidden_nonlinearity: Activation function for intermediate dense layer(s).
+        output_nonlinearity: Activation function for output dense layer.
+        hidden_w_init: Initializer function for the weight of intermediate dense layer(s).
+        hidden_b_init: Initializer function for the bias of intermediate dense layer(s).
+        output_w_init: Initializer function for the weight of output dense layer(s).
+        output_b_init: Initializer function for the bias of output dense layer(s).
+        weight_normalization: Bool for using weight normalization or not.
+
+    Return:
+        A MLP tf.Tensor object.
```
ryanjulian: The output tf.Tensor of the MLP.

garage/tf/core/networks.py
```diff
@@ -0,0 +1,55 @@
...
+        output_nonlinearity=None,
+        hidden_w_init=tf.contrib.layers.xavier_initializer(),
+        hidden_b_init=tf.zeros_initializer(),
+        output_w_init=tf.contrib.layers.xavier_initializer(),
+        output_b_init=tf.zeros_initializer(),
+        weight_normalization=False):
+    """Multilayer Perceptrons (MLP) implemention with tensorflow
+
+    Args:
+        input_var: User predefined tf.Tensor for feeding input.
```
ryanjulian: Input tf.Tensor to the MLP

garage/tf/core/networks.py
```diff
@@ -0,0 +1,48 @@
+import tensorflow as tf
+
+
+def mlp(input_var,
+        output_dim,
+        hidden_sizes,
+        hidden_nonlinearity=tf.nn.relu,
```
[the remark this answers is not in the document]
ryanjulian: Please fix this before submitting.

ryanjulian:
Your build is failing because of style checks. Did you have pre-commit setup? https://github.com/rlworkgroup/garage/blob/master/CONTRIBUTING.md#preparing-your-repo-to-make-contributions

ryanjulian:
Please get this rebased/reformatted/submitted ASAP

garage/tf/core/networks.py
```diff
@@ -0,0 +1,55 @@
+import tensorflow as tf
```
CatherineSue: Please add module docstring here.

CatherineSue: Is networks a good name for this module

garage/tf/core/networks.py
```diff
@@ -0,0 +1,55 @@
...
+        hidden_sizes,
+        name,
+        hidden_nonlinearity=tf.nn.relu,
+        output_nonlinearity=None,
+        hidden_w_init=tf.contrib.layers.xavier_initializer(),
+        hidden_b_init=tf.zeros_initializer(),
+        output_w_init=tf.contrib.layers.xavier_initializer(),
+        output_b_init=tf.zeros_initializer(),
+        weight_normalization=False):
+    """Multilayer Perceptrons (MLP) implemention with tensorflow
```
CatherineSue: New line. And end with a period.

tests/garage/tf/core/test_networks.py
```diff
@@ -0,0 +1,122 @@
...
+        self.sess.run(tf.global_variables_initializer())
+
+    def test_multiple_same_mlp(self):
+        # We create another mlp with the same name
+        self.mlp_same_copy = mlp(
+            input_var=self._input,
+            output_dim=self._output_shape,
+            hidden_sizes=(32, 32),
+            hidden_nonlinearity=self.hidden_nonlinearity,
+            name="mlp1")
```
CatherineSue: What's the purpose for this test? Is this the reason you use `with tf.variable_scope(name, reuse=tf.AUTO_REUSE)` in mlp function?

garage/tf/core/networks.py
ahtsan: Why shouldn't we set mlp to be reused if possible? If the user wants another mlp, he can give it another name (another variable scope).

garage/tf/core/networks.py
ahtsan: I want to use network but...

tests/garage/tf/core/test_networks.py
ahtsan: I want to show how to reuse the same mlp by referring to the same network name.

garage/tf/core/networks.py
ryanjulian: networks.py is fine

PR #348, opened 2018-10-17
Setup scripts now check directory

### LINKED ISSUE #324: Fix setup_xxx.sh scripts
### There are some major errors that need to be fixed in order for the scripts to work correctly.
- `Directory ‘.’ is not installable. File ‘setup.py’ not found.`
This script assumes that the individual will run the script in the `garage/` home directory. This should not be assumed.
 >**Potential solution**: recursively determine where the garage directory is located (relative or absolute path), then use that path to locate `setup.py` and replace `.` in the script.

- `SpecNotFound: Runtime error: Can't process without a name.`
This is a conda error, specifically on line 216 in `setup_osx.sh`: `conda env create -f environment.yml`. For some users, running this line with the relative path `environment.yml` will not work.
>**Potential solution**: recursively determine where the garage directory is located (relative or absolute path), then use that path to locate environment.yml and pass the full path in the script.
E.g. `conda env create -f /path/to/environment.yml`.

- Miniconda is not adding a name (i.e. "garage") to the created environment.
```$ conda env list 
# conda environments:
#
                  *  /miniconda/path/to/environment
^ missing garage here
```
>**Potential solution**: Not sure, don't have experience with miniconda. Maybe miniconda doesn't have naming functionality? Though I highly doubt it.

pelillian:
Fixes #324

### DIFF
--- a/scripts/setup_linux.sh
+++ b/scripts/setup_linux.sh
@@ -103,7 +103,8 @@ print_warning() {
 }
 
 # Verify this script is running from the correct folder (root directory)
-(more setup.py | grep -q "name='rlgarage'") || _PRINT_HELP=yes die \
+([ -e setup.py ] && (more setup.py | grep -q "name='rlgarage'")) \
+  || _PRINT_HELP=yes die \
   "This setup script must be run from the root garage directory." 1
 
 # Verify there's a file in the mjkey path
--- a/scripts/setup_linux.sh
+++ b/scripts/setup_linux.sh
@@ -106,9 +106,10 @@ print_warning() {
 dir_err_txt="Please run this script only from the root of the garage \
 repository, i.e. you should run it using the command \
 \"bash scripts/setup_linux.sh\""
-([ -e setup.py ] && (grep -Fq "name='rlgarage'," setup.py)) \
-  || _PRINT_HELP=yes die \
+if ! test -f setup.py && ! grep -Fq "name='rlgarage'," setup.py; then
+  _PRINT_HELP=yes die \
   "${dir_err_txt}" 1
+fi
 
 # Verify there's a file in the mjkey path
 test "$(file -b --mime-type ${_arg_mjkey})" == "text/plain" \
--- a/scripts/setup_linux.sh
+++ b/scripts/setup_linux.sh
@@ -103,9 +103,11 @@ print_warning() {
 }
 
 # Verify this script is running from the correct folder (root directory)
-([ -e setup.py ] && (more setup.py | grep -q "name='rlgarage'")) \
+dir_err_txt="This script must be run from the root garage git directory. Use:
+./scripts/setup_linux.sh"
+([ -e setup.py ] && (grep -Fq "name='rlgarage'," setup.py)) \
   || _PRINT_HELP=yes die \
-  "This setup script must be run from the root garage directory." 1
+  "${dir_err_txt}" 1
 
 # Verify there's a file in the mjkey path
 test "$(file -b --mime-type ${_arg_mjkey})" == "text/plain" \
--- a/scripts/setup_linux.sh
+++ b/scripts/setup_linux.sh
@@ -107,8 +107,7 @@ dir_err_txt="Please run this script only from the root of the garage \
 repository, i.e. you should run it using the command \
 \"bash scripts/setup_linux.sh\""
 if [[ ! -f setup.py ]] && ! grep -Fq "name='rlgarage'," setup.py; then
-  _PRINT_HELP=yes die \
-  "${dir_err_txt}" 1
+  _PRINT_HELP=yes die "${dir_err_txt}" 1
 fi
 
 # Verify there's a file in the mjkey path
--- a/scripts/setup_linux.sh
+++ b/scripts/setup_linux.sh
@@ -106,7 +106,7 @@ print_warning() {
 dir_err_txt="Please run this script only from the root of the garage \
 repository, i.e. you should run it using the command \
 \"bash scripts/setup_linux.sh\""
-if ! test -f setup.py && ! grep -Fq "name='rlgarage'," setup.py; then
+if [[ ! -f setup.py ]] && ! grep -Fq "name='rlgarage'," setup.py; then
   _PRINT_HELP=yes die \
   "${dir_err_txt}" 1
 fi
--- a/scripts/setup_linux.sh
+++ b/scripts/setup_linux.sh
@@ -103,8 +103,9 @@ print_warning() {
 }
 
 # Verify this script is running from the correct folder (root directory)
-dir_err_txt="This script must be run from the root garage git directory. Use:
-./scripts/setup_linux.sh"
+dir_err_txt="Please run this script only from the root of the garage \
+repository, i.e. you should run it using the command \
+\"bash scripts/setup_linux.sh\""
 ([ -e setup.py ] && (grep -Fq "name='rlgarage'," setup.py)) \
   || _PRINT_HELP=yes die \
   "${dir_err_txt}" 1
--- a/scripts/setup_linux.sh
+++ b/scripts/setup_linux.sh
@@ -102,6 +102,10 @@ print_warning() {
   printf "\e[0;33m%s\e[0m" "${1}"
 }
 
+# Verify this script is running from the correct folder (root directory)
+(more setup.py | grep -q "name='rlgarage'") || _PRINT_HELP=yes die \
+  "This setup script must be run from the root garage directory." 1
+
 # Verify there's a file in the mjkey path
 test "$(file -b --mime-type ${_arg_mjkey})" == "text/plain" \
   || _PRINT_HELP=yes die \
--- a/scripts/setup_linux.sh
+++ b/scripts/setup_linux.sh
@@ -106,7 +106,7 @@ print_warning() {
 dir_err_txt="Please run this script only from the root of the garage \
 repository, i.e. you should run it using the command \
 \"bash scripts/setup_linux.sh\""
-if [[ ! -f setup.py ]] && ! grep -Fq "name='rlgarage'," setup.py; then
+if [[ ! -f setup.py ]] && [[ ! $(grep -Fq "name='rlgarage'," setup.py) ]]; then
   _PRINT_HELP=yes die "${dir_err_txt}" 1
 fi
 
--- a/scripts/setup_osx.sh
+++ b/scripts/setup_osx.sh
@@ -45,7 +45,7 @@ _arg_modify_bashrc="off"
 print_help ()
 {
   printf '%s\n' "Installer of garage for OS X."
-  printf 'Usage: %s [--mjkey <arg>] [--(no-)modify-bashrc] ' "$0"
+  printf 'Usage: ./scripts/setup_osx.sh [--mjkey <arg>] [--(no-)modify-bashrc] '
   printf '[-h|--help]\n'
   printf '\t%s\n' "--mjkey: Path of the MuJoCo key (no default)"
   printf '\t%s' "--modify-bashrc,--no-modify-bashrc: Set environment "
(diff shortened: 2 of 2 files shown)


ryanjulian:
Can you provide some shell logs showing examples of what happens when I try to install from various (wrong) directories? Please include Linux and Mac (you can recruit someone else to help with linux).

pelillian:
MacOS:

Scripts Directory: (Script does not run)
```
(garage) Peters-MacBook-Pro:scripts pelillian$ ./setup_osx.sh 
Installer of garage for OS X.
Usage: ./setup_osx.sh [--mjkey <arg>] [--(no-)modify-bashrc] [-h|--help]
	--mjkey: Path of the MuJoCo key (no default)
	--modify-bashrc,--no-modify-bashrc: Set environment variables in .bash_profile (off by default)
	-h,--help: Prints help
This setup script must be run from the root garage directory.
```

Root Directory: (This is expected behavior)
```
(garage) Peters-MacBook-Pro:garage pelillian$ ./scripts/setup_osx.sh 
Installer of garage for OS X.
Usage: ./scripts/setup_osx.sh [--mjkey <arg>] [--(no-)modify-bashrc] [-h|--help]
	--mjkey: Path of the MuJoCo key (no default)
	--modify-bashrc,--no-modify-bashrc: Set environment variables in .bash_profile (off by default)
	-h,--help: Prints help
The path  of the MuJoCo key is not valid.
```

pelillian:
I'll test it on Linux when I get home

scripts/setup_linux.sh
```diff
@@ -102,6 +102,11 @@ print_warning() {
   printf "\e[0;33m%s\e[0m" "${1}"
 }
 
+# Verify this script is running from the correct folder (root directory)
+([ -e setup.py ] && (more setup.py | grep -q "name='rlgarage'")) \
```
ryanjulian: * Please follow the [Google Shell Style Guide](https://google.github.io/styleguide/shell.xml), as mentioned in CONTRIBUTING.md.
* There's no need to open `setup.py` with `more` and pipe to `grep`. `grep` can access the file directly (see `man grep`)

scripts/setup_linux.sh
```diff
@@ -102,6 +102,11 @@ print_warning() {
...
+  || _PRINT_HELP=yes die \
+  "This setup script must be run from the root garage directory." 1
```
ryanjulian: Can you make this message more helpful? If a user hits this error, they might be confused if they only got this message.

For instance, there's a path in the repo which is `garage/garage`--which do you mean? Tell them to go to the root directory of the git repository.

I think it's even best to provide a snippet they can run, e.g. `./scripts/setup_linux.sh`, which will only work if they're in the right directory.

scripts/setup_osx.sh
```diff
@@ -129,6 +129,11 @@ if [[ ! " ${VERIFIED_OSX_VERSIONS[@]} " =~ " ${VER} " ]]; then
   done
 fi
 
+# Verify this script is running from the correct folder (root directory)
+([ -e setup.py ] && (more setup.py | grep -q "name='rlgarage'")) \
```
ryanjulian: Same comments as above

pelillian:
Linux

scripts Directory: (Script does not run because we're not in the correct directory)
```
(base) peter@peter-desktop:~/git/garage/scripts$ ./setup_linux.sh 
Installer of garage for Linux.
Usage: ./scripts/setup_linux.sh [--mjkey <arg>] [--(no-)modify-bashrc] [--(no-)tf-gpu] [-h|--help]
	--mjkey: Path of the MuJoCo key (no default)
	--tf-gpu,--no-tf-gpu: Install TensorFlow GPU instead of the regular version (off by default)
	--modify-bashrc,--no-modify-bashrc: Set environment variables required by garage in .bashrc (off by default)
	-h,--help: Prints help
This script must be run from the root garage git directory. Use:
./scripts/setup_linux.sh
```
garage Root Directory: (This is expected behavior)
```
(base) peter@peter-desktop:~/git/garage$ ./scripts/setup_linux.sh 
Installer of garage for Linux.
Usage: ./scripts/setup_linux.sh [--mjkey <arg>] [--(no-)modify-bashrc] [--(no-)tf-gpu] [-h|--help]
	--mjkey: Path of the MuJoCo key (no default)
	--tf-gpu,--no-tf-gpu: Install TensorFlow GPU instead of the regular version (off by default)
	--modify-bashrc,--no-modify-bashrc: Set environment variables required by garage in .bashrc (off by default)
	-h,--help: Prints help
The path  of the MuJoCo key is not valid.
```

pelillian:
@ryanjulian Made the changes you mentioned! :+1:

ryanjulian:
```
This script must be run from the root garage git directory. Use:
./scripts/setup_linux.sh
```
Ah but the root of the garage git directory would be `garage/.git` :)

How about 

```
Please run this script only from the root of the garage repository, i.e. you should run it using the command "bash scripts/setup_linux.sh".
```

pelillian:
MacOS
scripts Directory: (Script does not run because we're not in the correct directory)
```
(garage) Peters-MacBook-Pro:scripts pelillian$ bash setup_osx.sh 
Installer of garage for OS X.
Usage: ./scripts/setup_osx.sh [--mjkey <arg>] [--(no-)modify-bashrc] [-h|--help]
	--mjkey: Path of the MuJoCo key (no default)
	--modify-bashrc,--no-modify-bashrc: Set environment variables in .bash_profile (off by default)
	-h,--help: Prints help
Please run this script only from the root of the garage repository, i.e. you should run it using the command "bash scripts/setup_osx.sh"
```
garage Root Directory: (This is expected behavior)
```
(garage) Peters-MacBook-Pro:garage pelillian$ bash scripts/setup_osx.sh 
Installer of garage for OS X.
Usage: ./scripts/setup_osx.sh [--mjkey <arg>] [--(no-)modify-bashrc] [-h|--help]
	--mjkey: Path of the MuJoCo key (no default)
	--modify-bashrc,--no-modify-bashrc: Set environment variables in .bash_profile (off by default)
	-h,--help: Prints help
The path  of the MuJoCo key is not valid.
```

scripts/setup_linux.sh
```diff
@@ -102,6 +102,14 @@ print_warning() {
   printf "\e[0;33m%s\e[0m" "${1}"
 }
 
+# Verify this script is running from the correct folder (root directory)
+dir_err_txt="Please run this script only from the root of the garage \
+repository, i.e. you should run it using the command \
+\"bash scripts/setup_linux.sh\""
+([ -e setup.py ] && (grep -Fq "name='rlgarage'," setup.py)) \
```
ryanjulian: I don't think this conforms to the style guide for conditionals
https://google.github.io/styleguide/shell.xml#Test,_%5B_and_%5B%5B

@gonzaiva opinion?

pelillian: @ryanjulian I wanted to keep code consistency with what existed already. I can change all the conditionals to if statements if that's better.

scripts/setup_linux.sh
```diff
@@ -40,7 +40,7 @@ _arg_tf_gpu="off"
 print_help ()
 {
   printf '%s\n' "Installer of garage for Linux."
-  printf 'Usage: %s [--mjkey <arg>] [--(no-)modify-bashrc] ' "$0"
+  printf 'Usage: ./scripts/setup_linux.sh [--mjkey <arg>] [--(no-)modify-bashrc] '
```
ryanjulian: Do not edit this part of the file. It is autogenerated (see block comments above and below).

scripts/setup_linux.sh
ryanjulian: I don't see anywhere else in this file which uses the subshell `()` style you used here for a conditional statement. everything below either uses `test -f` or `if [[  ]]``

scripts/setup_osx.sh
```diff
@@ -42,7 +42,7 @@ _arg_modify_bashrc="off"
 print_help ()
 {
   printf '%s\n' "Installer of garage for OS X."
-  printf 'Usage: %s [--mjkey <arg>] [--(no-)modify-bashrc] ' "$0"
+  printf 'Usage: ./scripts/setup_osx.sh [--mjkey <arg>] [--(no-)modify-bashrc] '
```
ryanjulian: All comments above also apply to this file

scripts/setup_linux.sh
gonzaiva: Yes, please use test or if. Makes things more verbose but also clearer. Also, I think a better way to verify this is using the current working directory (pwd): if we're executing this script one level down the path we're good. The problem in relying in setup.py is that if in the future we move/rename the file (not likely but it could happen), nobody is going to remember to update this verification as well.
Please also remove/update my failed code for this :(
https://github.com/rlworkgroup/garage/blob/dfba4206c8d938d41a592e9fd8ce1f16f4aab1c6/scripts/setup_linux.sh#L109

gonzaiva: Okay, but I think for our case it would be sufficient to check if we're running the script one directory below. However, it's true that if we move the installation script, my suggestion will fail.

ryanjulian: What if I execute the script from `/`? What if I execute it from `/home/rjulian/code/notgarage` but there is a directly `notgarage/scripts`?

gonzaiva: If we use `$(dirname ${0})` in the script, we get the relative path from our current working directory to the script, which should be `./scripts`, and we don't rely on the garage folder or the root, but I accept the setup.py solution as well.

ryanjulian: Angel's solution is fine with me, too.

pelillian: Code has been changed to use test in an if statement for clarity. 👍

scripts/setup_linux.sh
```diff
@@ -102,6 +102,15 @@ print_warning() {
   printf "\e[0;33m%s\e[0m" "${1}"
 }
 
+# Verify this script is running from the correct folder (root directory)
+dir_err_txt="Please run this script only from the root of the garage \
+repository, i.e. you should run it using the command \
+\"bash scripts/setup_linux.sh\""
+if ! test -f setup.py && ! grep -Fq "name='rlgarage'," setup.py; then
```
ryanjulian: @gonzaiva don't you use *either* `test -f` *or* `if`

gonzaiva: Ok, I will only use one from now on

ryanjulian: haha. i was asking you what's appropriate according to convention/style guide.

pelillian: Just replaced the test statement with brackets

scripts/setup_osx.sh
```diff
@@ -129,6 +129,15 @@ if [[ ! " ${VERIFIED_OSX_VERSIONS[@]} " =~ " ${VER} " ]]; then
   done
 fi
 
+# Verify this script is running from the correct folder (root directory)
+dir_err_txt="Please run this script only from the root of the garage \
+repository, i.e. you should run it using the command \
+\"bash scripts/setup_osx.sh\""
+if [[ ! -f setup.py ]] && ! grep -Fq "name='rlgarage'," setup.py; then
```
ryanjulian: all conditions in an `if` statement must be enclosed in `[[ ]]`. I know the style guide doesn't show this explicitly, but their motivation behind `[[ ]]` is to force all statements in the conjunction into an unambiguous boolean value.

scripts/setup_linux.sh
```diff
@@ -102,6 +102,15 @@ print_warning() {
   printf "\e[0;33m%s\e[0m" "${1}"
 }
 
+# Verify this script is running from the correct folder (root directory)
+dir_err_txt="Please run this script only from the root of the garage \
+repository, i.e. you should run it using the command \
+\"bash scripts/setup_linux.sh\""
+if [[ ! -f setup.py ]] && ! grep -Fq "name='rlgarage'," setup.py; then
+  _PRINT_HELP=yes die \
+  "${dir_err_txt}" 1
```
ryanjulian: why split the line here?

pelillian: Meant to fix that!

scripts/setup_osx.sh
pelillian: I'm getting
`conditional binary operator expected`
when I add the braces around grep.

I can put curly braces around it though.

ryanjulian:
You probably need to use `$()` to evaluate grep in a subshell.

scripts/setup_osx.sh
ryanjulian: You probably need to use `$()` to evaluate grep in a subshell.

ryanjulian:
```shell
if [[ ! -f setup.py ]] && [[ ! $(grep -Fq "name='rlgarage'," setup.py) ]]; then
  # stuff
fi
```

ryanjulian:
See CONTRIBUTING.md about how and why not to create merge commits. Please amend the head commit of your branch to have a meaningful commit message.

ryanjulian:
@pelillian please get this closed quickly, since it's blocking the release.

Issue #273, opened 2018-08-11
Always copy actions if they are modified by the environment


ryanjulian:
When using VectorizedSampler, the actions ndarrary (or a slice thereof) of passed directly to the environment. If the environment modifies the action (e.g. by re-scaling or clipping it), it will modify the action vector used by the optimizer. This is *not* the correct behavior--the actions used by the optimizer should be the ones produced by the policy, regardless of how the environment mangles the actions.

To close this issue, we need to audit each environment (including wrappers) and make sure that any environment which modifies actions in `step()` copies them first.

ryanjulian:
@jonashen status update?

jonashen:
I can work on this issue for the release sprint.

PR #350, opened 2018-10-19
Remove side effects from step() in policies

jonashen:
Fixes: #273.


jonashen:
Since all environment tests (including test_normalized_env) are/should use step_env() in helpers.py, I decided to add the unittest within the function directly.

tests/garage/envs/test_normalized_gym.py
```diff
@@ -35,8 +43,16 @@ def test_unflatten(self):
...
             env.reset()
             for e in range(5):
-                action = env.action_space.sample()
-                next_obs, reward, done, info = env.step(action)
+                a = env.action_space.sample()
+                a_copy = a if isinstance(a, int) else a.copy()
+                next_obs, reward, done, info = env.step(a)
+
+                # Check for side effects
+                if isinstance(a, int):
```
gonzaiva: I will get a little picky here, but would this work for a scalar continuous action (e.g. float)?

tests/helpers.py
```diff
@@ -29,7 +38,16 @@ def step_env_with_gym_quirks(test_case,
...
     env.reset()
     for _ in range(n):
-        _, _, done, _ = env.step(env.action_space.sample())
+        a = env.action_space.sample()
+        a_copy = a if isinstance(a, int) else a.copy()
+        _, _, done, _ = env.step(a)
+
+        # Check for side effects
+        if isinstance(a, int):
+            assert a == a_copy
```
gonzaiva: Some error messages would be useful here, so if another guy finds this issue in Travis, it takes less time to figure out what exactly went wrong.

tests/garage/envs/test_normalized_gym.py
jonashen: Good point.

tests/helpers.py
jonashen: done

garage/envs/box2d/box2d_env.py
```diff
@@ -176,11 +176,15 @@ def step(self, action):
         Note: override this method with great care, as it post-processes the
         observations, etc.
         """
-        reward_computer = self.compute_reward(action)
+        # Copy action first to remove side effects
+        action_copy = action if isinstance(action,
```
ryanjulian: i think you can remove this switch by using the copy stdlib

https://docs.python.org/3.5/library/copy.html

garage/envs/mujoco/half_cheetah_env.py
```diff
@@ -35,10 +35,15 @@ def get_body_com(self, body_name):
         return self.data.get_body_xpos(body_name)
 
     def step(self, action):
-        self.forward_dynamics(action)
+        # Copy action first to remove side effects
+        action_copy = action if isinstance(action,
```
ryanjulian: `copy.copy()`

tests/garage/envs/test_normalized_gym.py
```diff
@@ -18,8 +18,18 @@ def test_flatten(self):
...
                 env.render()
-                action = env.action_space.sample()
-                next_obs, reward, done, info = env.step(action)
+                a = env.action_space.sample()
+                a_copy = a if isinstance(a, (int, float)) else a.copy()
+                next_obs, _, done, _ = env.step(a)
+
+                # Check for side effects
+                if isinstance(a, (int, float)):
+                    assert a == a_copy,\
```
ryanjulian: PEP8 says to split lines with `()`, not `\`

tests/garage/envs/test_normalized_gym.py
```diff
@@ -18,8 +18,18 @@ def test_flatten(self):
...
+                    "Action was modified by environment!"
+                else:
+                    assert a.all() == a_copy.all(),\
```
ryanjulian: Split with ()

tests/garage/envs/test_normalized_gym.py
```diff
@@ -35,8 +45,18 @@ def test_unflatten(self):
...
             for e in range(5):
-                action = env.action_space.sample()
-                next_obs, reward, done, info = env.step(action)
+                a = env.action_space.sample()
+                a_copy = a if isinstance(a, (int, float)) else a.copy()
+                next_obs, reward, done, info = env.step(a)
+
+                # Check for side effects
+                if isinstance(a, (int, float)):
+                    assert a == a_copy,\
```
ryanjulian: Split with ()

tests/helpers.py
```diff
@@ -7,7 +7,18 @@
...
     env.reset()
     for _ in range(n):
-        _, _, done, _ = env.step(env.action_space.sample())
+        a = env.action_space.sample()
+        a_copy = a if isinstance(a, int) else a.copy()
+        _, _, done, _ = env.step(a)
+
+        # Check for side effects
+        if isinstance(a, (int, float)):
+            assert a == a_copy,\
```
ryanjulian: Split with ()

tests/helpers.py
```diff
@@ -7,7 +7,18 @@
...
+            "Action was modified by environment!"
+        else:
+            assert a.all() == a_copy.all(),\
+             "Action was modified by environment!"
```
ryanjulian: Split with ()

tests/helpers.py
```diff
@@ -29,7 +40,18 @@ def step_env_with_gym_quirks(test_case,
 
     env.reset()
     for _ in range(n):
-        _, _, done, _ = env.step(env.action_space.sample())
+        a = env.action_space.sample()
+        a_copy = a if isinstance(a, int) else a.copy()
+        _, _, done, _ = env.step(a)
+
+        # Check for side effects
+        if isinstance(a, (int, float)):
```
ryanjulian: Split with ()

tests/helpers.py
```diff
@@ -7,7 +7,18 @@
 def step_env(env, n=10, render=True):
     env.reset()
     for _ in range(n):
-        _, _, done, _ = env.step(env.action_space.sample())
+        a = env.action_space.sample()
+        a_copy = a if isinstance(a, int) else a.copy()
+        _, _, done, _ = env.step(a)
+
+        # Check for side effects
+        if isinstance(a, (int, float)):
```
ryanjulian: TestCase (`self`) has `assertEqual()` to avoid exactly this code.

tests/helpers.py
ryanjulian: assertEqual()

tests/garage/envs/test_normalized_gym.py
```diff
@@ -35,8 +45,18 @@ def test_unflatten(self):
...
+                    "Action was modified by environment!"
+                else:
+                    assert a.all() == a_copy.all(),\
```
ryanjulian: assertEqual()

tests/helpers.py
```diff
@@ -7,7 +7,18 @@
...
+            assert a == a_copy,\
+            "Action was modified by environment!"
+        else:
+            assert a.all() == a_copy.all(),\
```
ryanjulian: Don't assert inside test helper functions unless the helper function name makes it crystal clear that it's a *test* and not a *function*. This helper is a function.

It makes it difficult for the test author to predict why the test might fail. It also bundles together a bunch of test cases into one function, rather than making them individual test cases.

tests/helpers.py
```diff
@@ -29,7 +40,18 @@ def step_env_with_gym_quirks(test_case,
...
     for _ in range(n):
-        _, _, done, _ = env.step(env.action_space.sample())
+        a = env.action_space.sample()
+        a_copy = a if isinstance(a, int) else a.copy()
+        _, _, done, _ = env.step(a)
+
+        # Check for side effects
+        if isinstance(a, (int, float)):
+            assert a == a_copy,\
+            "Action was modified by environment!"
```
ryanjulian: Don't assert inside helper functions.

tests/garage/envs/test_normalized_gym.py
```diff
@@ -35,8 +45,18 @@ def test_unflatten(self):
         for i in range(10):
```
ryanjulian: It's not clear to me why this is repeated 10 times.

tests/garage/envs/test_normalized_gym.py
ryanjulian: This code does not test flattening and therefore does not belong in this test case. It is a separate test case (function). If the test cases share common code, use `setUp()`.

tests/garage/envs/test_normalized_gym.py
ryanjulian: This doesn't test flattening at all and is therefore a different test function (e.g. `test_does_not_modify_action()`. If they share code, use `setUp()` to separate them.

Test code should be as granular as possible. When there's a failure, you want to see the exact few lines of code which make up the failing test, not a 100 line test function with 10 possible broken tests inside of it.

tests/garage/envs/dm_control/test_dm_control_env.py
```diff
@@ -10,13 +12,19 @@
 
 class TestDmControlEnv(unittest.TestCase):
     @params(*dm_control.suite.ALL_TASKS)
-    def test_all_pickleable(self, domain, task):
+    def test_all_pickleable_and_step(self, domain, task):
```
ryanjulian: We can save some time by parameterizing this class (rather than the individual functions) and then making a single envi n `setUp()`

jonashen: I will combine it with `DmTestCase` in fixtures.py and delete this file.

tests/garage/envs/dm_control/test_dm_control_env.py
```diff
@@ -1,22 +0,0 @@
...
-
-class TestDmControlEnv(unittest.TestCase):
-    @params(*dm_control.suite.ALL_TASKS)
-    def test_all_pickleable(self, domain, task):
-        env = DmControlEnv(domain, task)
-        round_trip = pickle.loads(pickle.dumps(env))
-        assert round_trip
-        step_env(round_trip)
-
-    @params(*dm_control.suite.ALL_TASKS)
```
ryanjulian: Actually please move DmTestCase here and delete it from fixtures. Fixtures are fixtures--they shouldn't have tests themselves.

PR #353, opened 2018-10-21
Update Dockerfile.ci for new mujoco_py install

ryanjulian:
mujoco_py changed their setup.py to build the native bindings during
install, rather than during first import. This affects when we need to
make mjkey.txt available to the Dockferfile, and also means we don't
need to trigger the build manually by importing mujoco_py within the
Dockerfile.


ryanjulian:
I rebased @naeioi 's GPU PR and it appears to be working. https://travis-ci.com/rlworkgroup/garage/builds/88669853

naeioi:
> I rebased @naeioi 's GPU PR and it appears to be working. https://travis-ci.com/rlworkgroup/garage/builds/88669853

Great news!

PR #361, opened 2018-10-26
Copy MuJoCo key before installing the conda env

gonzaiva:
Currently the dm-control installation calls the MuJoCo package during
its installation, and since no MuJoCo key is present, there's an error
during the installation.

### DIFF
--- a/scripts/setup_linux.sh
+++ b/scripts/setup_linux.sh
@@ -103,7 +103,8 @@ print_warning() {
 }
 
 # Verify there's a file in the mjkey path
-test -f "${_arg_mjkey}" || _PRINT_HELP=yes die \
+test "$(file -b --mime-type ${_arg_mjkey})" == "text/plain" \
+  || _PRINT_HELP=yes die \
   "The path ${_arg_mjkey} of the MuJoCo key is not valid." 1
 
 # Make sure that we're under the garage directory
@@ -194,6 +195,9 @@ fi
 . "${CONDA_SH}"
 conda update -q -y conda
 
+# We need a MuJoCo key to import mujoco_py
+cp "${_arg_mjkey}" "${HOME}/.mujoco/mjkey.txt"
+
 # Create conda environment
 conda env create -f environment.yml
 if [[ "${?}" -ne 0 ]]; then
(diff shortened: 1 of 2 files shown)


gonzaiva:
Tested on fresh installations of:
- Ubuntu 16.04
- OSX 10.12

scripts/setup_linux.sh
```diff
@@ -194,6 +194,9 @@ fi
 . "${CONDA_SH}"
 conda update -q -y conda
 
+# We need a MuJoCo key to import mujoco_py
+cp ${_arg_mjkey} "${HOME}"/.mujoco/mjkey.txt
```
gautams3: I think we need failsafes in case:
1. the user doesn't present a mujoco key
2. the mujoco key is not a text file
3. (if possible) the mujoco key is not valid, but we may not be able to check that before installing mujoco

scripts/setup_osx.sh
```diff
@@ -215,6 +215,9 @@ fi
 . "${CONDA_SH}"
 conda update -q -y conda
 
+# We need a MuJoCo key to import mujoco_py
+cp ${_arg_mjkey} "${HOME}"/.mujoco/mjkey.txt
```
gautams3: Same comment for the OSX setup

gonzaiva: Same reply but  in [here](https://github.com/rlworkgroup/garage/blob/master/scripts/setup_osx.sh#L135-L137)

scripts/setup_linux.sh
ryanjulian: 1. That's already checked above
2. `$(file --mime-type "${_arg_mjkey}") == "text/plain")` or similar
3. Not possible.

ryanjulian: Remember to quote all variable expansions, per the style guide.

gautams3: Great!! And I don't think 3 is possible outside the mujoco install process

ryanjulian: `test -f` checks that it's a file, but does not check that it's text.

gonzaiva:
I tested again to make sure the mime type verification works in OSX and Linux.

scripts/setup_linux.sh
```diff
@@ -103,7 +103,8 @@ print_warning() {
 }
 
 # Verify there's a file in the mjkey path
-test -f "${_arg_mjkey}" || _PRINT_HELP=yes die \
+test "$(file -b --mime-type ${_arg_mjkey})" == "text/plain" \
```
ryanjulian: in the future i really prefer `if` statements but this is fine.

Issue #281, opened 2018-08-14
policy_ent_coeff does not seem to work in tf/NPO


ryanjulian:
I can choose any value (or GaussianMLPPolicy params) I want, and it doesn't change the policy standard deviation. e.g. -1.0 won't cause the policy stddev to fall, and 1.0 won't cause it to rise. This may be because a gradient path is broken between the reward function and the policy std network.

ryanjulian:
Things to check:

* Assuming your augmented rewards equation looks like `r + (policy_ent_coeff * policy_entropy)`, ensure that `policy_entropy` is differentiable (symbolic) all the way back to the `std_network` of the policy. This means that your loss function retrieves policy_entropy by purely by feeding actions and observations, not by using the recorded mean/std (e.g. agent_infos, dist_infos).
* Make sure that dist_infos are only used for forward pass calculations (e.g. KL divergence)

ryanjulian:
@CatherineSue were you able to reproduce this?

CatherineSue:
Y

ryanjulian:
For this issue, let's use
* MaxEnt (e.g. augment the reward with an entropy term and a coefficient): \hat r = r + \alpha_1 * H
* log-likelihood estimator: H ~ -logli(action| policy(state))
* Turn off the gradient: H = tf.stop_gradient(H)

PR #368, opened 2018-10-29
Fix policy entropy

CatherineSue:
This commit fixes policy entropy. Currently it uses the following three
options:
• maximum entropy (augmented reward)
• with a logli estimator (-logli(action|policy)
• with the gradient turned off (tf.stop_gradient(entropy))

See: #281

### DIFF
--- a/garage/tf/algos/npo.py
+++ b/garage/tf/algos/npo.py
@@ -34,12 +34,16 @@ class NPO(BatchPolopt):
                  optimizer_args=None,
                  name="NPO",
                  policy=None,
-                 policy_ent_coeff=1e-2,
+                 policy_ent_coeff=0.0,
                  use_softplus_entropy=True,
+                 use_neg_logli_entropy=True,
+                 stop_entropy_gradient=True,
                  **kwargs):
         self.name = name
         self._name_scope = tf.name_scope(self.name)
         self._use_softplus_entropy = use_softplus_entropy
+        self._use_neg_logli_entropy = use_neg_logli_entropy
+        self._stop_entropy_gradient = stop_entropy_gradient
 
         self._pg_loss = pg_loss
         if optimizer is None:
@@ -95,7 +99,8 @@ class NPO(BatchPolopt):
         logger.record_tabular("{}/KL".format(self.policy.name), policy_kl)
 
         pol_ent = self.f_policy_entropy(*policy_opt_input_values)
-        logger.record_tabular("{}/Entropy".format(self.policy.name), pol_ent)
+        logger.record_tabular("{}/Entropy".format(self.policy.name),
+                              np.mean(pol_ent))
 
         num_traj = self.batch_size // self.max_path_length
         actions = samples_data["actions"][:num_traj, ...]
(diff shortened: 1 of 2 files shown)


garage/tf/algos/npo.py
```diff
@@ -358,22 +363,37 @@ def _build_entropy_term(self, i):
...
                                         [-1, self.max_path_length])
 
             # This prevents entropy from becoming negative for small policy std
             if self._use_softplus_entropy:
                 policy_entropy = tf.nn.softplus(policy_entropy)
 
-            policy_entropy = tf.reduce_mean(policy_entropy * i.valid_var)
+            policy_entropy = policy_entropy * i.valid_var
+
+            if self._stop_entropy_gradients:
```
ryanjulian: stop_entropy_gradient

garage/tf/algos/npo.py
```diff
@@ -36,10 +36,14 @@ def __init__(self,
                  policy=None,
                  policy_ent_coeff=1e-2,
                  use_softplus_entropy=True,
+                 log_estimate_entropy=True,
```
ryanjulian: use_logli_entropy

garage/tf/algos/npo.py
```diff
@@ -36,10 +36,14 @@ def __init__(self,
                  policy=None,
                  policy_ent_coeff=1e-2,
                  use_softplus_entropy=True,
```
ryanjulian: is this still needed?

garage/tf/algos/npo.py
```diff
@@ -358,22 +363,37 @@ def _build_entropy_term(self, i):
...
                     i.flat.obs_var,
                     i.flat.policy_state_info_vars,
                     name="policy_dist_info_flat")
+                policy_log_likeli_flat = self.policy.distribution.log_likelihood_sym(  # noqa: E501
+                    i.flat.action_var,
+                    policy_dist_info_flat,
+                    name="policy_log_likeli")
+
+            if self._use_logli_entropy:
+                policy_entropy_flat = policy_log_likeli_flat
```
zhanpenghe: @ryanjulian Shouldn't this be negative log likelihood if using it for entropy?

garage/tf/algos/npo.py
CatherineSue: Yes

garage/tf/algos/npo.py
CatherineSue: Are you suggesting to change the name? 
log_likelihood_sym returns a negative log likelihood.

zhanpenghe: Oh ok. Then just change the name. I did not notice that it returns negative log likelihood already.

ryanjulian:
Do any of the examples even use entropy?

zhanpenghe:
`policy_ent_coeff` was set as `1e-2` before so any of the examples that does not explicit set it to 0 would use entropy.

CatherineSue:
The default value of `use_logli_entropy` is True, so every related example and benchmark will be using negative log likelihood as entropy term

garage/tf/algos/npo.py
CatherineSue: Done.

CatherineSue:
I changed the default policy_ent_coeff to 0.0.

PR #370, opened 2018-10-29
Update GCC to avoid build errors

gonzaiva:
In OSX 10.14 there are build errors for dm-control due to gcc 7 which
invalidates the installation of the conda environment for garage. This
is solved by upgrading to the latest version of gcc, and it was tested
in both OSX 10.12 and 10.14.


CatherineSue:
Yes the default is now gcc@8 I believe. I tried gcc@7, it doesn't work on 10.14, but gcc@8 works. gcc@8 also works on 10.12. I haven't tested on 10.13.

ryanjulian:
I'm willing to bet gcc@8 works in 10.13 if it works on 10.12.

Don't forget to update the known-good versions

gautams3:
scripts/setup_osx.sh has been renamed to setup_macos.sh. Rebase off of master and push

ryanjulian:
Please get this submitted ASAP since we want it in the release

PR #375, opened 2018-10-29
Fix TensorFlow warning in NPO about IndexedSlices

### LINKED ISSUE #289: Investigate TensorFlow warning in NPO about IndexedSlices
```sh
/home/rjulian/miniconda2/envs/garage-joblib/lib/python3.6/site-packages/tensorflow/python/ops/gradients_impl.py:108: UserWarning: Converting sparse IndexedSlices to a dense Tensor of unknown shape. This may consume a large amount of memory.
  "Converting sparse IndexedSlices to a dense Tensor of unknown shape. "
```

I think this has something to do with the design of our loss function in NPO. 

* Determine if we can address the warning
* If we can't address the warning, determine if it's actually a problem
* If it's not a (big) problem, I think we can find a way to suppress it

naeioi:
NPO uses a boolean valid_var tensor to select a slice from a variety
of tensors related to action, policy and advantages. Previously we
used `tf.boolean_mask()` to do the selection, which internally calls
`tf.gather()` that emits `tf.IndexedSlices`. When calculating BP,
`tf.IndexedSlices` will be casted into a dense tf.Tensor with a
`"Converting sparse IndexedSlices to a dense Tensor of unknown shape"`
warning. This fix replaces `tf.boolean_mask()` with
`tf.dynamic_partition()`, which don't have such problem.

This difference of behavior is not documented in Tensorflow. It is
not clear if `tf.dynamic_partition()` will cause performance
regression.

This fixes #289.

### DIFF
--- a/garage/tf/algos/npo.py
+++ b/garage/tf/algos/npo.py
@@ -126,7 +126,7 @@ class NPO(BatchPolopt):
             reward_var = tensor_utils.new_tensor(
                 name="reward", ndim=2, dtype=tf.float32)
             valid_var = tf.placeholder(
-                tf.bool, shape=[None, None], name="valid")
+                tf.float32, shape=[None, None], name="valid")
             baseline_var = tensor_utils.new_tensor(
                 name="baseline", ndim=2, dtype=tf.float32)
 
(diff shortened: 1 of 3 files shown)


ryanjulian:
Looks like you need to rebase

Issue #42, opened 2018-06-11
Get rid of layers library in the TensorFlow tree


ryanjulian:
This is largely superseded by `tf.nn`, and it adds another unnecessary layer of complexity--it would be easier to understand if primitives were just written in pure TF.

Imported from https://github.com/ryanjulian/rllab/issues/42

ahtsan:
@ryanjulian Is it done and closed by #939 ?

PR #378, opened 2018-11-01
Add parameter without layers library

zhanpenghe:
Part of rewriting GaussianMLPPolicy without layers.


garage/tf/core/networks.py
```diff
@@ -64,3 +65,37 @@ def mlp(input_var,
             bias_initializer=output_b_init,
             name="output")
     return l_out
+
+
+def parameter(input_var,
```
ryanjulian: how could a parameter have an `input_var`?

garage/tf/core/networks.py
```diff
@@ -64,3 +65,37 @@ def mlp(input_var,
...
+              initializer=tf.zeros_initializer(),
+              dtype=tf.float32,
+              trainable=True,
+              name="parameter"):
+    """
+    Paramter function that creates  variables that could be broadcast to a certain shape to match with input var.
+        
+    Args:
+        input_var: Input tf.Tensor.
+        lenth: Integer dimension of the variables.
```
ryanjulian: length

garage/tf/core/networks.py
```diff
@@ -64,3 +65,37 @@ def mlp(input_var,
             bias_initializer=output_b_init,
             name="output")
     return l_out
+
+
+def parameter(input_var,
+              length,
```
ryanjulian: why can't my parameter have an arbitrary shape?

tests/garage/tf/core/test_networks.py
```diff
@@ -156,3 +156,26 @@ def test_layer_normalization(self):
...
+
+        init = tf.constant_initializer(initial_params)
+
+        params = parameter(
+            input_var=input_vars,
+            length=3,
+            initializer=init,
+        )
+
+        data = np.random.uniform(size=[5, 2, 3, 4])
```
ryanjulian: generally avoid using random values in tests unless you're actually testing randomness. it can lead to flaky tests.

zhanpenghe:
I think this is not necessary but the main purpose of this function is flexibly broadcast the parameters to a certain shape to match the input_vars (e.g. the std of gaussian_mlp primitives).

ryanjulian:
Hmm can you provide an example of what it looks like in the wild with/without this utility? I'm not against it I just want to understand what it looks like.

garage/tf/core/networks.py
zhanpenghe: This is used for broadcast the parameter shape.

garage/tf/core/networks.py
```diff
@@ -64,3 +65,38 @@ def mlp(input_var,
...
+              dtype=tf.float32,
+              trainable=True,
+              name="parameter"):
+    """
+    Paramter function that creates  variables that could be
+    broadcast to a certain shape to match with input var.
+
+    Args:
+        input_var: Input tf.Tensor.
+        lenth: Integer length of the variables.
```
CatherineSue: length: Integer length of the variables.

tests/garage/tf/core/test_networks.py
zhanpenghe: done

ryanjulian:
ref #42

ahtsan:
Does this essentially "extending" each batch to length `length`? i.e.
```python
input_var = tf.placeholder(shape=[None, 2], dtype=tf.float32)
p = parameter(
    input_var=input_var,
    length=5,
    ...)
original_inputs = np.zeros(shape=[3, 2])
'''
original value = [
    [0, 0],
    [0, 0],
    [0, 0]
]
'''
result = sess.run(p, feed_dict={input_var: inputs})
'''
result = [
    [0, 0, 0, 0, 0],
    [0, 0, 0, 0, 0],
    [0, 0, 0, 0, 0]
]
'''
```
In this case, broadcasting from 2 to 5

PR #384, opened 2018-11-02
Default-enable all rules in flake8 checks

ryanjulian:
This PR changes the flake8 CI script to enable all rules by default.
All disabled rules are explicitly specified, making it easier to
inspect which rules we are not yet fully-compliant with.


ryanjulian:
Note that I have verified with intentionally-failed builds that changed and added files are still checked properly. They are in the Travis build history for this branch.

e.g. https://travis-ci.com/rlworkgroup/garage/builds/90140084

PR #385, opened 2018-11-03
Add GaussianMLPModel and abstract class Model

zhanpenghe:
Implemented GaussianMLPModel. A import feature/contract of this abstraction is assuming that all tensors should be created in `build_model`. This allows users to make a copy of the graph with the same parameters. This is used in some cases e.g. MAML(#374).

Also, the model is not written with garage layers library. ref: #42 
Currently the model is using parameterized and serializable for pickling.

### DIFF
--- a/garage/tf/core/parameter.py
+++ b/garage/tf/core/parameter.py
@@ -1,7 +1,6 @@
 """Parameter layer in TensorFlow."""
 
 import tensorflow as tf
-from tensorflow.python.ops.gen_array_ops import broadcast_to
 
 
 def parameter(input_var,
@@ -37,7 +36,9 @@ def parameter(input_var,
             initializer=initializer,
             trainable=trainable)
 
-        broadcast_shape = tf.concat(
-            axis=0, values=[tf.shape(input_var)[:-1], [length]])
-        p_broadcast = broadcast_to(p, shape=broadcast_shape)
-        return p_broadcast
+        ndim = input_var.get_shape().ndims
+        reshaped_p = tf.reshape(p, (1, ) * (ndim - 1) + (length, ))
+        tile_arg = tf.concat(
+            axis=0, values=[tf.shape(input_var)[:ndim - 1], [1]])
+        tiled = tf.tile(reshaped_p, tile_arg)
+        return tiled
--- a/garage/tf/models/__init__.py
+++ b/garage/tf/models/__init__.py
@@ -1,0 +1,4 @@
+from garage.tf.models.base import Model
+from garage.tf.models.gaussian_mlp_model import GaussianMLPModel
+
+__all__ = ["Model", "GaussianMLPModel"]
--- a/garage/tf/models/__init__.py
+++ b/garage/tf/models/__init__.py
@@ -1,3 +1,5 @@
+"""Tensorflow models."""
+
 from garage.tf.models.base import Model
 from garage.tf.models.gaussian_mlp_model import GaussianMLPModel
 
--- a/garage/tf/models/__init__.py
+++ b/garage/tf/models/__init__.py
@@ -1,4 +1,4 @@
-"""Tensorflow models."""
+"""TensorFlow models."""
 
 from garage.tf.models.base import Model
 from garage.tf.models.gaussian_mlp_model import GaussianMLPModel
--- a/garage/tf/models/base.py
+++ b/garage/tf/models/base.py
@@ -1,50 +1,43 @@
-"""
-This file contains the abstraction class for models
-"""
+"""This file contains the abstraction class for models."""
 import tensorflow as tf
 
-from garage.tf.core import layers as L
-from garage.tf.core import Parameterized
 from garage.misc.overrides import overrides
+from garage.tf.core import Parameterized
 
 
 class Model(Parameterized):
-    """
-    Abstraction class for models.
-    """
+    """Abstraction class for models."""
 
     def __init__(self):
+        """Initialize a model."""
         Parameterized.__init__(self)
 
     def build_model(self):
         """
-        This function will build the whole graph for
-        the model. All tensors should be created here.
-        By calling this function, a copy of the same graph
-        should be created with the same parameters.
-        Currently three things are returned: inputs, outputs and
-        model_info. All of them are returned as dictionaries.
+        Build the whole graph for the model.
+
+        All tensors should be created here. By calling this function,
+        a copy of the same graph should be created with the same
+        parameters. Currently three things are returned: inputs,
+        outputs and model_info. All of them are returned as dictionaries.
         """
         raise NotImplementedError
 
     @property
     def inputs(self):
-        """
-        Tensors of the inputs
-        """
+        """Tensors of the inputs."""
         return self._inputs
 
     @property
     def outputs(self):
-        """
-        Tensors of the outputs
-        """
+        """Tensors of the outputs."""
         return self._outputs
 
     @overrides
     def get_params_internal(self, **tags):
         """
-        Function for pickling the model.
+        Retrieve the parameters to pickle the model.
+
         Args
             tags: a dictionary of tags, only support trainable for now.
         Return
--- a/garage/tf/models/base.py
+++ b/garage/tf/models/base.py
@@ -12,7 +12,7 @@ class Model(Parameterized):
         """Initialize a model."""
         Parameterized.__init__(self)
 
-    def build_model(self):
+    def build_model(self, inputs=None, reuse=tf.AUTO_REUSE):
         """
         Build the whole graph for the model.
 
@@ -24,7 +24,7 @@ class Model(Parameterized):
         raise NotImplementedError
 
     @property
-    def inputs(self, inputs=None, reuse=tf.AUTO_REUSE):
+    def inputs(self):
         """Tensors of the inputs."""
         return self._inputs
 
--- a/garage/tf/models/base.py
+++ b/garage/tf/models/base.py
@@ -8,6 +8,8 @@ from garage.tf.core import Parameterized
 
 class Model(KerasModel, Parameterized):
     """
+    This is a tautology for all the garage models.
+
     A model define a neural network model to build a graph for
     other primitivs (e.g. GaussianMLPPolicy and GaussianMLPBaseline).
     A model can be built by assembling different models. For example,
--- a/garage/tf/models/base.py
+++ b/garage/tf/models/base.py
@@ -24,7 +24,7 @@ class Model(Parameterized):
         raise NotImplementedError
 
     @property
-    def inputs(self):
+    def inputs(self, inputs=None, reuse=tf.AUTO_REUSE):
         """Tensors of the inputs."""
         return self._inputs
 
--- a/garage/tf/models/base.py
+++ b/garage/tf/models/base.py
@@ -1,0 +1,57 @@
+"""
+This file contains the abstraction class for models
+"""
+import tensorflow as tf
+
+from garage.tf.core import layers as L
+from garage.tf.core import Parameterized
+from garage.misc.overrides import overrides
+
+
+class Model(Parameterized):
+    """
+    Abstraction class for models.
+    """
+
+    def __init__(self):
+        Parameterized.__init__(self)
+
+    def build_model(self):
+        """
+        This function will build the whole graph for
+        the model. All tensors should be created here.
+        By calling this function, a copy of the same graph
+        should be created with the same parameters.
+        Currently three things are returned: inputs, outputs and
+        model_info. All of them are returned as dictionaries.
+        """
+        raise NotImplementedError
+
+    @property
+    def inputs(self):
+        """
+        Tensors of the inputs
+        """
+        return self._inputs
+
+    @property
+    def outputs(self):
+        """
+        Tensors of the outputs
+        """
+        return self._outputs
+
+    @overrides
+    def get_params_internal(self, **tags):
+        """
+        Function for pickling the model.
+        Args
+            tags: a dictionary of tags, only support trainable for now.
+        Return
+            parameters of the model.
+        """
+        if tags.get("trainable"):
+            params = [v for v in tf.trainable_variables(scope=self.name)]
+        else:
+            params = [v for v in tf.global_variables(scope=self.name)]
+        return params
--- a/garage/tf/models/base.py
+++ b/garage/tf/models/base.py
@@ -1,15 +1,24 @@
-"""This file contains the abstraction class for models."""
+"""This file contains the tautology for garage models."""
 import tensorflow as tf
+from tensorflow.keras.models import Model as KerasModel
 
 from garage.misc.overrides import overrides
 from garage.tf.core import Parameterized
 
 
-class Model(Parameterized):
-    """Abstraction class for models."""
+class Model(KerasModel, Parameterized):
+    """
+    A model define a neural network model to build a graph for
+    other primitivs (e.g. GaussianMLPPolicy and GaussianMLPBaseline).
+    A model can be built by assembling different models. For example,
+    to define a VAEGaussianMLP Model, we need to build a VAE model
+    and a GaussianMLP Model. Then, we can assemble them in the
+    build_model function.
+    """
 
-    def __init__(self):
+    def __init__(self, *args, **kwargs):
         """Initialize a model."""
+        super(Model, self).__init__(*args, **kwargs)
         Parameterized.__init__(self)
 
     def build_model(self, inputs=None, reuse=tf.AUTO_REUSE):
@@ -23,16 +32,6 @@ class Model(Parameterized):
         """
         raise NotImplementedError
 
-    @property
-    def inputs(self):
-        """Tensors of the inputs."""
-        return self._inputs
-
-    @property
-    def outputs(self):
-        """Tensors of the outputs."""
-        return self._outputs
-
     @overrides
     def get_params_internal(self, **tags):
         """
@@ -41,7 +40,7 @@ class Model(Parameterized):
         Args
             tags: a dictionary of tags, only support trainable for now.
         Return
-            parameters of the model.
+            parameters of the model in tf.Tensor.
         """
         if tags.get("trainable"):
             params = [v for v in tf.trainable_variables(scope=self.name)]
--- a/garage/tf/models/gaussian_mlp_model.py
+++ b/garage/tf/models/gaussian_mlp_model.py
@@ -1,0 +1,211 @@
+import numpy as np
+import tensorflow as tf
+
+from garage.core import Serializable
+from garage.misc import ext
+from garage.tf.core.networks import mlp, parameter
+from garage.tf.models import Model
+
+
+class GaussianMLPModel(Model, Serializable):
+    """Gaussian MLP Model"""
+    def __init__(self,
+                 input_dim,
+                 output_dim,
+                 name="GaussianMLPModel",
+                 hidden_sizes=(32, 32),
+                 learn_std=True,
+                 init_std=1.0,
+                 adaptive_std=False,
+                 std_share_network=False,
+                 std_hidden_sizes=(32, 32),
+                 min_std=1e-6,
+                 max_std=None,
+                 std_hidden_nonlinearity=tf.nn.tanh,
+                 hidden_nonlinearity=tf.nn.tanh,
+                 output_nonlinearity=None,
+                 std_parameterization='exp',
+                ):
+        """
+        :param input_dim: input dimension
+        :param output_dim: output dimension
+        :param name: name of the model
+        :param hidden_sizes: list of sizes for the fully-connected hidden
+          layers
+        :param learn_std: Is std trainable
+        :param init_std: Initial std
+        :param adaptive_std:
+        :param std_share_network:
+        :param std_hidden_sizes: list of sizes for the fully-connected layers
+          for std
+        :param min_std: whether to make sure that the std is at least some
+          threshold value, to avoid numerical issues
+        :param std_hidden_nonlinearity:
+        :param hidden_nonlinearity: nonlinearity used for each hidden layer
+        :param output_nonlinearity: nonlinearity for the output layer
+        :param mean_network: custom network for the output mean
+        :param std_network: custom network for the output log std
+        :param std_parametrization: how the std should be parametrized. There
+          are a few options:
+            - exp: the logarithm of the std will be stored, and applied a
+                exponential transformation
+            - softplus: the std will be computed as log(1+exp(x))
+        :return:
+        """
+
+        Serializable.quick_init(self, locals())
+        super(GaussianMLPModel, self).__init__()
+
+        self.name = name
+        self._variable_scope = tf.variable_scope(
+            self.name, reuse=tf.AUTO_REUSE)
+        self._name_scope = tf.name_scope(self.name)
+
+        # Network parameters
+        self._input_dim = input_dim
+        self._output_dim = output_dim
+        self._hidden_sizes = hidden_sizes
+        self._learn_std = learn_std
+        self._init_std = init_std
+        self._adaptive_std = adaptive_std
+        self._std_share_network = std_share_network
+        self._std_hidden_sizes = std_hidden_sizes
+        self._min_std = min_std
+        self._max_std = max_std
+        self._std_hidden_nonlinearity = std_hidden_nonlinearity
+        self._hidden_nonlinearity = hidden_nonlinearity
+        self._output_nonlinearity = output_nonlinearity
+        self._std_parameterization = std_parameterization
+
+        # Tranform std arguments to parameterized space
+        self._init_std_param = None
+        self._min_std_param = None
+        self._max_std_param = None
+        if self._std_parameterization == 'exp':
+            self._init_std_param = np.log(init_std)
+            if min_std:
+                self._min_std_param = np.log(min_std)
+            if max_std:
+                self._max_std_param = np.log(max_std)
+        elif self._std_parameterization == 'softplus':
+            self._init_std_param = np.log(np.exp(init_std) - 1)
+            if min_std:
+                self._min_std_param = np.log(np.exp(min_std) - 1)
+            if max_std:
+                self._max_std_param = np.log(np.exp(max_std) - 1)
+        else:
+            raise NotImplementedError
+
+        inputs, outputs, model_info = self.build_model()
+        self._inputs = inputs["input_var"]
+        
+        self._mean = outputs["mean"]
+        self._std = outputs["std"]
+        self._std_param = outputs["std_param"]
+        self._sample = outputs["sample"]
+        self._dist = model_info["dist"]
+
+        self._outputs = outputs
+
+    def build_model(self):
+        """
+        Build the graph
+        return:
+            inputs: a dict that contains input tensor
+            outputs: a dict that contains mean, 
+                std, parameterized std and sample tensors
+            model_info: a dict that contains the distribution tensor
+        """
+
+        input_var = tf.placeholder(
+            shape=[None, self._input_dim],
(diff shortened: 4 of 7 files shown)


garage/tf/models/base.py
```diff
@@ -0,0 +1,49 @@
...
+
+class Model(Parameterized):
+    """Abstraction class for models."""
+
+    def __init__(self):
+        """Initialize a model."""
+        Parameterized.__init__(self)
+
+    def build_model(self):
+        """
```
CatherineSue: What if we want to create a target model with the same graph except for variable scope?

garage/tf/models/gaussian_mlp_model.py
```diff
@@ -0,0 +1,220 @@
...
+                exponential transformation
+            - softplus: the std will be computed as log(1+exp(x))
+        :return:
+        """
+        Serializable.quick_init(self, locals())
+        super(GaussianMLPModel, self).__init__()
+
+        self.name = name
+        self._variable_scope = tf.variable_scope(
+            self.name, reuse=tf.AUTO_REUSE)
```
CatherineSue: Why use tf.AUTO_REUSE here?

zhanpenghe: I assume parameters have unique fullnames (I.e. “variable_scope/name”)

zhanpenghe:
Sorry.. I accidentally pressed the update button that causes the merge... I am about to board now. Will resolve this after landing.

garage/tf/models/gaussian_mlp_model.py
ahtsan: I have the same thought for MLP too (each different mlp should have unique fullname)

garage/tf/models/gaussian_mlp_model.py
```diff
@@ -0,0 +1,220 @@
...
+                std, parameterized std and sample tensors
+            model_info: a dict that contains the distribution tensor
+        """
+        input_var = tf.placeholder(
+            shape=[None, self._input_dim],
+            dtype=tf.float32,
+        )
+
+        with self._variable_scope:
+            if self._std_share_network:
```
CatherineSue: Should we add the gaussian mlp model structure which has two outputs, one refers to mean, the other refers to std,  instead of one output that needs to be sliced to mean and std?

garage/tf/models/gaussian_mlp_model.py
```diff
@@ -0,0 +1,220 @@
...
+    ):
+        """
+        Initialize a gaussian mlp model and build the graph.
+
+        :param input_dim: input dimension
+        :param output_dim: output dimension
+        :param name: name of the model
+        :param hidden_sizes: list of sizes for the fully-connected hidden
+          layers
+        :param learn_std: Is std trainable
```
CatherineSue: Lowercase 'I'

ryanjulian:
@krzentner can you check this out before we merge?

garage/tf/models/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""Tensorflow models."""
```
ryanjulian: TensorFlow

garage/tf/models/base.py
ryanjulian: How will I compose models if `build_model()` doesn't take an input?

garage/tf/models/gaussian_mlp_model.py
ryanjulian: You should not set the reuse scope. Force the outer scope (e.g. the author) to decide which scope behavior to use.

garage/tf/models/gaussian_mlp_model.py
ryanjulian: They are not the same. A two headed-MLP is not the same as slicing the output of a single-headed MLP.

I think this should be split into several models. **Invariant: constructor arguments cannot change the architecture of the model, just parameters (e.g. hidden layers and sizes).**

ryanjulian:
We also need to ask ourselves if this is the correct path, or should be use keras?

Is the endgame for this abstraction just a worse version of keras?
Do we expect keras to be supported for the forseeable future?
Is keras extensible-enough for us to adopt their abstractions while implementing our own models?

etc.

garage/tf/models/gaussian_mlp_model.py
zhanpenghe: Oh, but I think two headed MLP is actually just a MLP connected with two MLPs right? I think chang was just saying that splitting the output layer?

garage/tf/models/base.py
zhanpenghe: Good point. I forgot to add this. Thanks!

garage/tf/models/gaussian_mlp_model.py
ryanjulian: In my mind, two-headed means shared hidden layers, two output layers. You are variously proposing 
* Shared MLP, split the output
* One MLP (with hidden and output layers) outputs to two other MLPs (minimum 3 layers).

garage/tf/models/gaussian_mlp_model.py
ryanjulian: To add detail: all parameters internal to this scope still need unique fullnames, but if there is a name collision, tf.AUTO_REUSE can cause TF to silently reuse parameters in a way that the author doesn't expect and can be difficult to debug.

That's why it's not the default in TF and I don't think it should be default here.

If an author wants tf.AUTO_REUSE they can declare that in the enclosing scope.

garage/tf/models/base.py
```diff
@@ -0,0 +1,50 @@
+"""This file contains the abstraction class for models."""
```
ryanjulian: This is a tautology. You should describe what models are.

garage/tf/models/base.py
```diff
@@ -0,0 +1,50 @@
...
+
+    @overrides
+    def get_params_internal(self, **tags):
+        """
+        Retrieve the parameters to pickle the model.
+
+        Args
+            tags: a dictionary of tags, only support trainable for now.
+        Return
+            parameters of the model.
```
ryanjulian: What are their types?

garage/tf/models/gaussian_mlp_model.py
```diff
@@ -0,0 +1,220 @@
...
+
+from garage.core import Serializable
+from garage.misc import ext
+from garage.misc.overrides import overrides
+from garage.tf.core.networks import mlp, parameter
+from garage.tf.models import Model
+
+
+class GaussianMLPModel(Model, Serializable):
+    """Gaussian MLP Model."""
```
ryanjulian: Describe in more detail what this is

garage/tf/models/gaussian_mlp_model.py
```diff
@@ -0,0 +1,220 @@
...
+            self.name, reuse=tf.AUTO_REUSE)
+        self._name_scope = tf.name_scope(self.name)
+
+        # Network parameters
+        self._input_dim = input_dim
+        self._output_dim = output_dim
+        self._hidden_sizes = hidden_sizes
+        self._learn_std = learn_std
+        self._init_std = init_std
+        self._adaptive_std = adaptive_std
```
ryanjulian: adaptive_std and std_share_network are different architectures so IMO they should be different models

garage/tf/models/gaussian_mlp_model.py
```diff
@@ -0,0 +1,220 @@
...
+            "input_var": input_var,
+        }
+        outputs = {
+            "mean": mean_var,
+            "std": std_var,
+            "std_param": std_param_var,
+            "sample": sample_var,
+        }
+        model_info = {
+            "dist": dist,
```
ryanjulian: I think we need to think harder about what the outputs are here. Is it just sample()? Or what about just the dist? Mean and std/std_param seem more like infos than outputs.

Should we even have infos? Isn't that a violation of encapsulation? What is the use case for infos? If your model provides a symbolic output as part of its interface, that is by definition an output.

ryanjulian:
I think it's important that we bring this PR to a resolution to decide on how we're going to define models/modules in TF. It's blocking a lot of redesigns which could really improve the TF tree.

It's very relevant to the discussion here https://github.com/rlworkgroup/proposals/pull/3

ryanjulian:
@CatherineSue @krzentner @zuoxingdong please comment

ryanjulian:
Note that this RFC outlining TF 2.0 semantics is quite relevant

https://github.com/tensorflow/community/pull/20/files

Here's a relevant question to keep the discussion going:

Let's assume the framework provides a native model encapsulation type (i.e. like `torch.nn.Module`) and we will use that type to build the computation graph associated with primitives.
### Should primitives (e.g. GaussianMLPPolicy, LinearBaseline) have a "has a" or "is a" relationship with their underlying models?

#### LinearBaseline "has a" model
class LinearBaseline(Baseline):

    def __init__(self, ...):
        self._model = self._build_model(...)

    def predict(self, inputs):
        return self._model.forward(inputs)

#### LinearBaseline "is a" model
class LinearBaseline(Baseline, tf.Model):

    def __init__(self, ...):
        super(self, tf.Model).__init__(...)

    def predict(self, inputs):
        return self.forward(inputs)

ahtsan:
Following my previous comment, I think `has-a` model can encapsulate the underlying model as well as preserving the class name semantics?

garage/tf/policies/gaussian_mlp_policy.py
```diff
@@ -47,214 +47,89 @@ def __init__(self,
...
+            std_hidden_sizes=std_hidden_sizes,
+            min_std=min_std,
+            max_std=max_std,
+            std_hidden_nonlinearity=std_hidden_nonlinearity,
+            hidden_nonlinearity=hidden_nonlinearity,
+            output_nonlinearity=output_nonlinearity,
+            std_parameterization=std_parameterization,
+        )
+
+        self._dist = DiagonalGaussian(self.env_spec.action_space.flat_dim)
```
CatherineSue: Why not using self.model["dist"]?

zhanpenghe: Because of the current implementation of NPO (which I don't want to fix for now).

zhanpenghe:
Sorry I pushed some of my maml stuff to this branch yesterday. I will remove those changes later.

For me, I would prefer "has a" relationship (or "having a" lol) since this can reduce the difficulties of testing `GaussianMLP*`.

garage/tf/models/base.py
CatherineSue: Can we extend the functionality of this to get parameters under a specific scope? 
```
model = GaussianMLPModel(**kwargs)
with tf.name_scope("target") as scope:
    target_model = model.build_model()
```
How do I get the params of the `target_model`?

garage/tf/policies/gaussian_mlp_policy.py
CatherineSue: Are you going to do benchmarking for the new GaussianMLPPolicy? If you are going to benchmark test the policy on algorithms, you need to use NPO/PPO/TRPO as they are the algorithms that use GaussianMLPPolicy for now.

garage/tf/models/gaussian_mlp_model.py
ryanjulian: i think dist is an output. This model is a Gaussian with an MLP input. How is the dist not the output?

On the contrary, I think mean, std, and std_param are infos, not outputs. I'm not sure sample() needs to be in either output or infos...that's something you derive from the model.

garage/tf/policies/gaussian_mlp_policy.py
```diff
@@ -47,214 +47,89 @@ def __init__(self,
...
-                log_std_var = std_param_var
-            elif self.std_parametrization == 'softplus':
-                log_std_var = tf.log(tf.log(1. + tf.exp(std_param_var)))
-            else:
-                raise NotImplementedError
-            return dict(mean=mean_var, log_std=log_std_var)
-
-    @overrides
     def get_action(self, observation):
+        sess = tf.get_default_session()
```
ryanjulian: This really makes me nervous.

Shouldn't we have a global session singleton if we're going to do this?

garage/tf/policies/gaussian_mlp_policy.py
```diff
@@ -4,16 +4,15 @@
 from garage.core import Serializable
```
ryanjulian: Can you rewrite the rest of the GaussianMLP* primitives to use this model?

garage/tf/policies/gaussian_mlp_policy.py
zhanpenghe: This imply the default session of the thread is used but we can always have the session as argument.

garage/tf/policies/gaussian_mlp_policy.py
zhanpenghe: Sure.

garage/tf/policies/gaussian_mlp_policy.py
ryanjulian: I know. I am positing we should eliminate this assumption in the TF tree, since the session can change at any time anywhere in the code.

That's probably out-of-scope for this PR.

However, you should at lease provide `sess=get_default_session()` as an optional kwarg, so that callers can control which session you're using.

ryanjulian:
We need to get the coverage to ~100% and benchmark this a lot before committing.

@ahtsan can you focus on extending this to the rest of the GaussianMLP primitives
@CatherineSue @cheng-chang can you help benchmark?
@zhanpenghe can coordinate

@krzentner I'd appreciate if you took a look at this for software engineering advice.

ryanjulian:
https://medium.com/tensorflow/standardizing-on-keras-guidance-on-high-level-apis-in-tensorflow-2-0-bad2b04c819a

@zhanpenghe does keras cover this use case? Or a subclass of keras.Model?

garage/tf/models/gaussian_mlp_model.py
zhanpenghe: The reason why dist is not an output is that we cannot do `session.run(dist, feed_dict=fd)`.

zhanpenghe:
@ryanjulian this is ready.

garage/tf/models/gaussian_mlp_model.py
ryanjulian: is `session.run(dist, feed_dict=foo)` a requirement of the API for keras.Model?

garage/tf/core/parameter.py
```diff
@@ -37,7 +36,9 @@ def parameter(input_var,
             initializer=initializer,
             trainable=trainable)
 
-        broadcast_shape = tf.concat(
-            axis=0, values=[tf.shape(input_var)[:-1], [length]])
-        p_broadcast = broadcast_to(p, shape=broadcast_shape)
-        return p_broadcast
+        ndim = input_var.get_shape().ndims
+        reshaped_p = tf.reshape(p, (1, ) * (ndim - 1) + (length, ))
+        tile_arg = tf.concat(
```
ryanjulian: why tile instead of broadcast?

I think broadcast is the canonical TensorFlow API to use. It allows the graph compiler to resolve the shape and optimize. Tiling forces a particular memory layout which might be inefficient.

In any case, this is not in scope for the PR unless it's necessary for models to work.

garage/tf/models/base.py
```diff
@@ -0,0 +1,51 @@
...
+import tensorflow as tf
+from tensorflow.keras.models import Model as KerasModel
+
+from garage.misc.overrides import overrides
+from garage.tf.core import Parameterized
+
+
+class Model(KerasModel, Parameterized):
+    """
+    This is a tautology for all the garage models.
```
ryanjulian: This is not a tautology. Not sure what you meant here.

garage/tf/models/base.py
```diff
@@ -0,0 +1,51 @@
...
+
+    A model define a neural network model to build a graph for
```
ryanjulian: Grammar: a model defines
Spelling: primitives

garage/tf/models/base.py
```diff
@@ -0,0 +1,51 @@
...
+    other primitivs (e.g. GaussianMLPPolicy and GaussianMLPBaseline).
+    A model can be built by assembling different models. For example,
```
ryanjulian: New paragraph

garage/tf/core/parameter.py
ryanjulian: It seems like these should become keras Layers at some point

garage/tf/models/base.py
```diff
@@ -0,0 +1,51 @@
...
+    and a GaussianMLP Model. Then, we can assemble them in the
+    build_model function.
+    """
+
+    def __init__(self, *args, **kwargs):
+        """Initialize a model."""
+        super(Model, self).__init__(*args, **kwargs)
+        Parameterized.__init__(self)
+
+    def build_model(self, inputs=None, reuse=tf.AUTO_REUSE):
```
ryanjulian: This is called `call()` in the keras API

https://www.tensorflow.org/api_docs/python/tf/keras/models/Model (see bullet point 2)

garage/tf/models/base.py
```diff
@@ -0,0 +1,51 @@
...
+    other primitivs (e.g. GaussianMLPPolicy and GaussianMLPBaseline).
+    A model can be built by assembling different models. For example,
+    to define a VAEGaussianMLP Model, we need to build a VAE model
+    and a GaussianMLP Model. Then, we can assemble them in the
+    build_model function.
+    """
+
+    def __init__(self, *args, **kwargs):
+        """Initialize a model."""
+        super(Model, self).__init__(*args, **kwargs)
```
ryanjulian: `super().__init__()` should be sufficient here

garage/tf/models/gaussian_mlp_model.py
```diff
@@ -0,0 +1,227 @@
...
+            std_share_network=False,
+            std_hidden_sizes=(32, 32),
+            min_std=1e-6,
+            max_std=None,
+            std_hidden_nonlinearity=tf.nn.tanh,
+            hidden_nonlinearity=tf.nn.tanh,
+            output_nonlinearity=None,
+            std_parameterization='exp',
+    ):
+        """
```
ryanjulian: We are now putting constructor params in the class docstring rather than` __init__` docstring

garage/tf/models/gaussian_mlp_model.py
```diff
@@ -0,0 +1,227 @@
...
+        :param std_network: custom network for the output log std
+        :param std_parametrization: how the std should be parametrized. There
+          are a few options:
+            - exp: the logarithm of the std will be stored, and applied a
+                exponential transformation
+            - softplus: the std will be computed as log(1+exp(x))
+        :return:
+        """
+        Serializable.quick_init(self, locals())
+        super(GaussianMLPModel, self).__init__()
```
ryanjulian: `super().__init__()`

garage/tf/policies/gaussian_mlp_policy.py
```diff
@@ -47,214 +46,89 @@ def __init__(self,
...
          are a few options:
             - exp: the logarithm of the std will be stored, and applied a
              exponential transformation
             - softplus: the std will be computed as log(1+exp(x))
         :return:
         """
         assert isinstance(env_spec.action_space, Box)
-
         Serializable.quick_init(self, locals())
+        super(GaussianMLPPolicy, self).__init__(env_spec)
```
ryanjulian: `super().__init__()`

garage/tf/policies/gaussian_mlp_policy.py
```diff
@@ -4,16 +4,14 @@
...
+from garage.tf.core import Parameterized
 from garage.tf.distributions import DiagonalGaussian
-from garage.tf.misc import tensor_utils
+from garage.tf.models import GaussianMLPModel
 from garage.tf.policies import StochasticPolicy
 from garage.tf.spaces import Box
 
 
-class GaussianMLPPolicy(StochasticPolicy, LayersPowered, Serializable):
+class GaussianMLPPolicy(StochasticPolicy, Parameterized, Serializable):
```
ryanjulian: * hmm i wonder if this has-a model or is-a model?

if it is-a model, then it gets `Parameterized` for free.

garage/tf/models/gaussian_mlp_model.py
```diff
@@ -0,0 +1,227 @@
...
+
+from garage.core import Serializable
+from garage.misc import ext
+from garage.misc.overrides import overrides
+from garage.tf.core.mlp import mlp
+from garage.tf.core.parameter import parameter
+from garage.tf.models import Model
+
+
+class GaussianMLPModel(Model, Serializable):
```
ryanjulian: does this behave properly if I use `tf.keras.Model` APIs like `to_json()` `weights` ,etc?

garage/tf/policies/gaussian_mlp_policy.py
zhanpenghe: This is has-a model

ryanjulian: haha i was asking what it *should be* not what it *is*

ryanjulian: @krzentner PTAL

garage/tf/models/gaussian_mlp_model.py
ryanjulian: still waiting on an answer to this question

ahtsan: No. From Keras documentation, 
```
In subclassed models, the model's topology is defined as Python code (rather than as a static graph of layers). That means the model's topology cannot be inspected or serialized. As a result, the following methods and attributes are not available for subclassed models:

model.inputs and model.outputs.
model.to_yaml() and model.to_json()
model.get_config() and model.save().
```
`model.get_weights()` and `model.set_weights(weights)` are still available, however they return a empty list, so I guess subclassed models will have to implement them as well.

ahtsan: or are we actually using the primitives, e.g. mlp, from keras?

ahtsan: If we want to use `keras.models.Model` and comply with keras API, we may also want to use their primitives (e.g. mlp) so that we reuse their properties (e.g. get_config(), to_json(), get_weights()). For custom model (e.g. GaussianMLPModel), we can also write it in `keras.layers.core` (e.g. Dense, Activation). 

Otherwise, we would have to write those functions ourselves and it seems we don't get significant benefit from following `keras.models.Model`. In this case, why don't we just use our own model class? (garage.Model)

garage/tf/models/base.py
```diff
@@ -0,0 +1,51 @@
+"""This file contains the tautology for garage models."""
+import tensorflow as tf
+from tensorflow.keras.models import Model as KerasModel
+
+from garage.misc.overrides import overrides
+from garage.tf.core import Parameterized
+
+
+class Model(KerasModel, Parameterized):
```
krzentner: I have a few high level questions about this class, which are hinted at in the discussion below, but don't look like they are being discussed directly:
 - Do we still need parameterized if we're using KerasModel, or does the Keras API provide sufficient functionality?
 - Do we really want to put Parameterized (a subclass of Serializable) this far up the inheritance heirarchy? Note that the conclusion from the discussion [here](https://github.com/rlworkgroup/proposals/pull/3) was specifically not to do that.
 - If we're using the Keras API explicitly, how much do we want to follow their naming conventions? Should we ensure that we use all of their names, or do we prefer to use our own?

ryanjulian: I agree with KR's questions. Part of the hope here is to removed Parameterized.

I think we should follow the "when in rome" principle and adopt their naming conventions.

@krzentner can you give an example of their naming convention on which we might differ?

CatherineSue: - Keras Model API now doesn't provide serializable mothods for subclassing models. 

- Keras layers support `layer.get_config` and `layer.get_weights` for serialization. In order to use layers methods, we need access to layers in the network. i.e.
``` python
class Model(KerasModel):
    def __getstate__(self,):
        weights = [layer.get_weights() for layer in model.layers]
```
In garage, Model calls mlp or other networks in `garage.tf.core` to build its network. According to Keras documentation, I don't think we can keep track of layers here unless we make network to a class.

CatherineSue: I also have some questions here:
- Is the purpose of Model only to use serialization in Keras? 
- After reading the Keras Model doc, it looks like they are useful to compose layers for complex networks. Is this the same idea as Model in garage? i.e. According to the docstring below, we can assemble multiple models in garage, but Keras Model doesn't support the same idea.

CatherineSue: Another concern on removing `Parameterized` here:
Some optimizers in `garage.tf.optimizers` uses `policy.get_param_values`. 
Optimizers in TF:
- ConjugateGradientOptimizer.
- LBFGSOptimizer. There is one [BFGSOptimizer](https://github.com/tensorflow/probability/blob/master/tensorflow_probability/python/optimizer/bfgs.py) in tf-prob.
- FirstOrderOptimizer, which is basically Adam.
- PenaltyLbfgsOptimizer. 

Can we find alternative optimizers in TF to replace them?

ryanjulian: @CatherineSue re: Graph Networks vs Subclass Networks.

This is certainly disappointing, and I think there are a couple ways to deal with it.

Firstly, I think their explanation is a bit unclear. I think that they are saying is: if you *simply* subclass `Model` *without* using the Keras composition APIs to build your model, then you won't get serialization *automatically*.

You are of course free to implement those methods yourself--it's still Python after all.

This leaves us with 2 choices:
1. Implement all primitives using Keras-style graph composition (Sequential and Functional APIs) -- this is basically a "Layers" library
1. Implement our own subclass(es) of Model which provide `clone()`, `get_config()`/`from_config()`, `to_json()`/`to_yaml()`, and `save()` -- this is basically what we have now with `Serializable` and `Parameterized`, but they are separate classes rather than being part of one Model API

If we implement the Keras API (using either 1 or 2), then future users will have both options (they can mix/match models built using the Keras builder APIs with our own).

Now for what to do. This is not 100% clear.

* I think that if the Keras Sequential/Functional APIs suit our needs, and implementing custom Layers is possible and easy, and that keras.Model will properly save Models with custom layers, we should implement our own Layers and then do 1.
* If the above does not hold, we should probably do (2).

Justification:
* We satisfy the original goal or removing Parameterized and Serializable
* In the worst case, it seems to me you can always make one giant layer for your whole primitive and then have a Model with one layer (not ideal of course, but possible)
* We are already implementing "Layers" with our network functions. This would just make the functions we already have into classes.
* Since TF has committed to the Keras API, it's likely that support for custom models will get better in the future
* We might be able to reduce our testing surface by using Layers already provided by Keras
* We get to keep the original intent of the Layers library without having to maintain any of it.

zhanpenghe:
I think deepmind sonnet’s implementation of module also provides an insightful way to do this (even though I think this is a bit different from our purpose of the model classes). They basically define part of the graph as module and provides API to connects them.

https://github.com/deepmind/sonnet/blob/master/sonnet/python/modules/base.py

ryanjulian:
@zhanpenghe thanks for the pointer.

My meta-comment is that, if you look at both Sonnet and Keras (which are both trying to achieve the same thing in different styles), you will find that they both have to implement a similar amount of code to achieve it.

That's why I suggest we just adopt Keras, because our own version would end up being similar in size.

I would be willing to entertain using Sonnet instead of Keras, if it fits our needs.

garage/tf/models/base.py
ahtsan: I tested it and `keras.Model` does not properly save Models with primitives using Keras-style graph composition. 
The only condition that `keras.Model` save the model is that the model is created by Sequential/Functional API, i.e. either `keras.models.Sequential[...]` or `keras.model.Model(inputs=..., outputs=...)`

PR #387, opened 2018-11-08
[WIP] Integrate rlkit

### LINKED ISSUE #19: Merge rlkit code to start garage/torch
It will be the base for `garage.torch`

naeioi:
This PR integrates rlkit under `garage.torch`. Examples are under `examples/torch`.

Supported algorithms:
- TDM
- DDPG
- DQN
- (Twin) SAC
- TD3

At the current stage all examples are ready to run but the overall integration is not finished. Most importantly, the original rlkit comes without any test. I have this PR open even before completion so that everyone can comment, especially on the reuse of common infrastructure (e.g. logger). I believe we were facing similar challenges when moving tensorflow out of sandbox.  

**Progress**
- [x] Overall availability (runnable examples)
- [x] Integration test on algorithms
- [x] Reuse code from garage base 
- [x] Check and unify experiment, algorithm, policy and sampler interface
  - [x]  Unify logger interface
   - Don't ignore `garage.config` that uses `LOG_DIR` to globally specify where to log.
   - The log naming convection should also be unified. 
  - [x] Unify experiment runner **(Only tested in local mode)**
- [ ] Documentation
  - [ ] Docstring. There are a daunting number of public classes and methods to be commented. 
  - [ ] User Manual
- [ ] Finalize
  - [ ] Pass flake8
  - [ ] Pass codecov

We also need to have supported algorithms synced up with Tensorflow but let's leave it for future PRs. 

This will close #19 .


garage/torch/policies/simple.py
```diff
@@ -0,0 +1,13 @@
+from garage.torch.policies.base import SerializablePolicy
+
+
+class RandomPolicy(SerializablePolicy):
+    """
+    Policy that always outputs zero.
```
zhanpenghe: Random policy always outputs a random action.

garage/torch/envs/mujoco_env.py
```diff
@@ -0,0 +1,64 @@
...
+import mujoco_py
+import numpy as np
+from gym.envs.mujoco import mujoco_env
+
+from garage.torch.core.serializable import Serializable
+
+ENV_ASSET_DIR = os.path.join(os.path.dirname(__file__), 'assets')
+
+
+class MujocoEnv(mujoco_env.MujocoEnv, Serializable):
```
zhanpenghe: Is this necessary? How is this different from the MujocoEnv we have?

garage/torch/data_management/simple_replay_buffer.py
```diff
@@ -0,0 +1,59 @@
+import numpy as np
+
+from garage.torch.data_management.replay_buffer import ReplayBuffer
+
+
+class SimpleReplayBuffer(ReplayBuffer):
```
zhanpenghe: I think we can just use the ReplayBuffer we have.

garage/torch/algos/sac/policies.py
```diff
@@ -0,0 +1,136 @@
...
+
+from garage.torch.policies.base import ExplorationPolicy, Policy
+from garage.torch.algos.distributions import TanhNormal
+from garage.torch.algos.networks import Mlp
+
+LOG_SIG_MAX = 2
+LOG_SIG_MIN = -20
+
+
+class TanhGaussianPolicy(Mlp, ExplorationPolicy):
```
CatherineSue: Should this be moved to policies?

environment.yml
```diff
@@ -2,15 +2,17 @@ name: garage
 channels:
     - https://conda.anaconda.org/menpo
     - https://conda.anaconda.org/conda-forge
+    - pytorch
```
ryanjulian: hmm is it possible to do pytorch without conda? we are trying to keep all dependencies in pip so we can abandon conda easily.

examples/torch/ddpg_halfcheetah.py
```diff
@@ -0,0 +1,64 @@
...
+            num_steps_per_eval=1000,
+            use_soft_update=True,
+            tau=1e-2,
+            batch_size=128,
+            max_path_length=1000,
+            discount=0.99,
+            qf_learning_rate=1e-3,
+            policy_learning_rate=1e-4,
+        ), )
+    setup_logger('name-of-experiment', variant=variant)
```
ryanjulian: can we modify run_experiment to have torch support instead?

environment.yml
naeioi: Yes it can. Fixed it.

garage/torch/envs/mujoco_env.py
naeioi: They both inherit gym.Env and share the most important APIs. But there are also several differences on initialization and behavior control that requires some effort to sync up.
- Different parameters.
Some parameters in base but not in torch, some are vice versa. For the latter ones we need to modify MujocoEnv in base.  e.g torch exposures manual control of `frame_skip`.
- Torch's MujocoEnv inherits Serializable but base MujocoEnv does not. I need some experiments to see if there will be problem. 

Luckily there are now only two mujoco envs supported in torch. I will try working on this issue later.

garage/torch/data_management/simple_replay_buffer.py
naeioi: They have very different APIs. Seems like syncing them up requires more than trivial work. For now, let's focus on divergent APIs exposed to end users. 

That being said we definitely need to have them unified in the long term. Could you give shot here and estimate how much effort we might need? @CatherineSue

garage/torch/algos/sac/policies.py
naeioi: Nice catch

garage/torch/policies/simple.py
naeioi: nice catch.

garage/torch/data_management/simple_replay_buffer.py
CatherineSue: I've taken a look at the replay_buffers here. Though APIs are different, the utility of each function looks like the same. So I think a couple of hours is sufficient to do this. Definitely less than the time that would take on unifying samplers. 

I could help on this task since I'm familiar with ReplayBuffer in garage. Looks like this would belong to the 3rd step and 4th step in your head comments.

ryanjulian: Let's just get a demo of torch running before we spend time unifying interfaces. It's not clear if we're going to merge this into garage directly, or just keep it as a branch to make sure we can support torch once we split the repo.

garage/torch/envs/mujoco_env.py
naeioi: MujocoEnv is removed from torch now.

naeioi:
Close because we are not merging rlkit, at lease before splitting of garage is finished.

PR #413, opened 2018-11-15
Added CNN to networks and QFunction class

ahtsan:


garage/tf/q_functions/continuous_mlp_q_function.py
```diff
@@ -5,10 +5,10 @@
 import garage.tf.core.layers as L
 from garage.tf.core.layers import batch_norm
 from garage.tf.misc import tensor_utils
-from garage.tf.q_functions import QFunction
+from garage.tf.q_functions import NetworkBuilder
```
CatherineSue: Where is NetworkBuilder? Am I missing something here?

garage/tf/q_functions/mlp_q_function.py
```diff
@@ -0,0 +1,47 @@
+import tensorflow as tf
+
+from garage.misc.overrides import overrides
+from garage.tf.core.networks import mlp
+from garage.tf.q_functions import QFunction
+
+
+class MLPQFunction(QFunction):
```
CatherineSue: Can we replace ContinuousMLPQFunction with MLPQFunction?

garage/tf/core/networks.py
```diff
@@ -1,10 +1,11 @@
 """
-Multilayer Perceptrons (MLP) with tensorflow as the only dependency.
+Different networks with tensorflow as the only dependency.
 
-The module contains MLP which serves as the base of all networks.
+The module contains MLP and CNN implementation.
 It aims to replace existing implementation of MLP class
```
CatherineSue: Perhaps delete MLP?

garage/tf/q_functions/continuous_mlp_q_function.py
ahtsan: oh I renamed QFunction to NetworkBuilder before
I will change that back

garage/tf/core/networks.py
ahtsan: why?

garage/tf/q_functions/mlp_q_function.py
ahtsan: What does the merge layer do in ContinuousMLPQFunction?

garage/tf/core/networks.py
CatherineSue: I meant delete 'MLP' in line 5.

ahtsan: oh yes

garage/tf/q_functions/continuous_mlp_q_function.py
ahtsan: Done

garage/tf/core/networks.py
ahtsan: Done

garage/tf/q_functions/__init__.py
```diff
@@ -1,5 +1,6 @@
 from garage.tf.q_functions.base import QFunction
 from garage.tf.q_functions.continuous_mlp_q_function import (
     ContinuousMLPQFunction)
+from garage.tf.q_functions.mlp_q_function import MLPQFunction
```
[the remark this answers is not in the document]
ahtsan: yes

garage/tf/core/networks.py
```diff
@@ -100,3 +102,98 @@ def parameter(input_var,
             axis=0, values=[tf.shape(input_var)[:ndim - 1], [length]])
         p_broadcast = broadcast_to(p, shape=broadcast_shape)
         return p_broadcast
+
+
+def cnn(input_var,
```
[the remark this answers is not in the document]
ahtsan: should I create garage.tf.networks? or just put the mlp.py/cnn.py in garage.tf.core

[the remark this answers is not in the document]
ahtsan: if I do that, how about the parameter function? Should I also do parameter.py?

garage/tf/core/cnn.py
```diff
@@ -0,0 +1,96 @@
...
+
+def cnn(input_var,
+        output_dim,
+        filter_dims,
+        num_filters,
+        stride,
+        name,
+        padding="SAME",
+        max_pooling=False,
+        pool_shape=(2, 2),
```
ryanjulian: Is this really a sane default (i.e. that's what people usually use) or should we force the user to specify?

garage/tf/q_functions/continuous_mlp_q_function.py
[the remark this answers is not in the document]
ahtsan: That class doesn't exist. I updated.

garage/tf/core/mlp.py
```diff
@@ -1,12 +1,11 @@
 """
-Multilayer Perceptrons (MLP) with tensorflow as the only dependency.
+MLP with tensorflow as the only dependency.
 
-The module contains MLP which serves as the base of all networks.
-It aims to replace existing implementation of MLP class
+The module contains MLP implementation.
+It aims to replace existing implementation of network class
```
ryanjulian: Editorial comments about the development process aren't usually necessary in documentation.

garage/tf/core/parameter.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def parameter(input_var,
+              length,
+              initializer=tf.zeros_initializer(),
+              dtype=tf.float32,
+              trainable=True,
+              name="parameter"):
+    """
+    Paramter function.
```
ryanjulian: It's not a function. It's either a network or a layer.

garage/tf/core/parameter.py
```diff
@@ -0,0 +1,42 @@
...
+
+    It creates variables that could be
```
ryanjulian: Weird wrapping here.

garage/tf/core/parameter.py
ryanjulian: Perhaps explain why we would want such a thing

garage/tf/q_functions/base.py
```diff
@@ -1,5 +1,25 @@
+import tensorflow as tf
+
 from garage.tf.core import Parameterized
 
 
 class QFunction(Parameterized):
-    pass
+    def _build_net(self, name, input_var):
```
ryanjulian: Making a private function part of a public interface is a strange choice. We will have to revisit this as we define TensorFlow models.

garage/tf/q_functions/discrete_mlp_q_function.py
```diff
@@ -0,0 +1,52 @@
...
+import tensorflow as tf
+
+from garage.misc.overrides import overrides
+from garage.tf.core.mlp import mlp
+from garage.tf.q_functions import QFunction
+
+
+class DiscreteMLPQFunction(QFunction):
+    """
+    Discrelte MLP Functoin class.
```
ryanjulian: Function

garage/tf/q_functions/discrete_mlp_q_function.py
```diff
@@ -0,0 +1,52 @@
...
+
+    This class implements a q value network to predict q based on the input
```
ryanjulian: Please be consistent with literature in your comments

Q-value network. Predicts the Q-value based on....

tests/garage/tf/core/test_cnn.py
```diff
@@ -0,0 +1,36 @@
...
+            self.cnn = cnn(
+                input_var=self._input_ph,
+                output_dim=self._output_shape,
+                filter_dims=(3, 3, 3),
+                num_filters=(32, 64, 128),
+                stride=1,
+                name="cnn1",
+                hidden_nonlinearity=self.hidden_nonlinearity)
+
+        self.sess.run(tf.global_variables_initializer())
```
ryanjulian: Is this not done already in your fixture? Perhaps you called super too early?

tests/garage/tf/core/test_cnn.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def test_shape(self):
```
ryanjulian: Please assert something about actual input/output pairs rather than just shape.

garage/tf/core/parameter.py
ahtsan: It is from @zhanpenghe

ryanjulian: You are responsible for every line of code you commit to this repository, regardless of its origin.

garage/tf/core/cnn.py
ahtsan: In many cases people do not use max pooling layers and in that case, pool shape is not needed. What should be a good way for that?

tests/garage/tf/core/test_cnn.py
ahtsan: I didn't see this in `TfGraphTestCase`. Should I add this there?

tests/garage/tf/core/test_cnn.py
ahtsan: Do you mean to obtain the same output given an input without using tensorflow conv2d?

tests/garage/tf/core/test_cnn.py
ahtsan: `TfGraphTestCase` is for testing graphs, where `tf.global_variables_initializer()` cannot be called in `setUp()`. I can create a new fixture for testing trivial things and initialize in `setUp()`.

CatherineSue: Is it possible to delete TfTestCase and only retain TfGraphTestCase? @jonashen 
Looks like creating a new graph and pass it to a session is relatively important when we have multiple tests in a test class, as in Anson's tests. And I can't find other tests that use TfTestCase.

We may as well call `tf.global_variables_initializer()` in setUp. Some algorithms test cases also use TfGraphTestCase. And algo.train() also calls `tf.global_variables_initializer()`. But it looks ok to call it twice in those cases.

ryanjulian: I defer to @CatherineSue 's judgement

garage/tf/core/cnn.py
```diff
@@ -0,0 +1,96 @@
...
+
+    with tf.variable_scope(name):
+        h = input_var
+        for index, (filter_dim, num_filter) in enumerate(
+                zip(filter_dims, num_filters)):
+            h = hidden_nonlinearity(
+                _conv(h, 'h{}'.format(index), filter_dim, num_filter, strides,
+                      hidden_w_init, hidden_b_init, padding))
+            if max_pooling:
+                h = tf.nn.max_pool(
```
ryanjulian: you don't test the max_pooling case. please add a test to cover this

garage/tf/core/parameter.py
eric-heiden: Typo ("Paramter")

garage/tf/core/parameter.py
eric-heiden: Typo ("Paramter")

garage/tf/core/parameter.py
```diff
@@ -0,0 +1,42 @@
+"""Parameter function with tensorflow as the only dependency."""
```
eric-heiden: "TensorFlow"

garage/tf/q_functions/discrete_mlp_q_function.py
eric-heiden: Discrelte

tests/garage/tf/core/test_cnn.py
```diff
@@ -0,0 +1,226 @@
...
+        # flatten
+        h_out = np.full(
+            (self.batch_size, 8 * 8 * self.out_channels[-1]),
+            filter_sum,
+            dtype=np.float32)
+        # pass to a dense layer
+        dense_in = tf.matmul(h_out, out_w) + out_b
+        np.testing.assert_array_equal(dense_in.eval(), result)
+
+    def convolve(self, input, filter_weights, filter_biass, stride):
```
CatherineSue: Please use a different name for `input`.

cpnota: Any specific reason? In the generic case `input` is a bad name for a function parameter, but in this case `convolve` is a mathematical function of `input`, and it's helpful to know which parameter is the "input" to the function being represented.

ryanjulian: `input()` is a [Python builtin](https://docs.python.org/3/library/functions.html), so using the variable name `input` masks the symbol in this context.

PEP8 [gently recommends](https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles) using a trailing underscore to avoid overloading builtins. I've also used `an_input` in the past.

garage/tf/core/networks.py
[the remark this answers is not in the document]
ryanjulian: yeah that sounds fine.

ahtsan:
@ryanjulian can you give a pass on it?

tests/garage/tf/core/test_cnn.py
```diff
@@ -0,0 +1,221 @@
...
+        # flatten
+        h_out = np.full(
+            (self.batch_size, 8 * 8 * self.out_channels[-1]),
+            filter_sum,
+            dtype=np.float32)
+        # pass to a dense layer
+        dense_in = tf.matmul(h_out, out_w) + out_b
+        np.testing.assert_array_equal(dense_in.eval(), result)
+
+    def convolve(self, _input, filter_weights, filter_biass, stride):
```
ryanjulian: i believe numpy and scipy provide a convolution function. If you can use that instead, you should

PR #414, opened 2018-11-15
Added gym env wrappers and tests

ahtsan:


garage/misc/wrappers.py
```diff
@@ -0,0 +1,88 @@
+"""Gym env wrappers"""
+import cv2
```
ryanjulian: OpenCV is an enormous dependency. Is there no other Python library which can do this (e.g. PIL or just plain numpy)?

tests/garage/misc/test_wrappers.py
```diff
@@ -0,0 +1,20 @@
+import gym
+
+from garage.misc.wrappers import ResizeAndGrayscaleWrapper, StackFramesWrapper
+from tests.fixtures import TfGraphTestCase
+
+
+class TestWrappers(TfGraphTestCase):
```
ryanjulian: One test class per target class, please.

garage/misc/wrappers.py
```diff
@@ -0,0 +1,88 @@
+"""Gym env wrappers"""
```
ryanjulian: Please do not put these in `misc`, which we are trying to clean out. I think it should go under `garage.envs`.

garage/misc/wrappers.py
```diff
@@ -0,0 +1,88 @@
...
+        return obs
+
+    def reset(self):
+        return self._observation(self.env.reset())
+
+    def step(self, action):
+        obs, reward, done, info = self.env.step(action)
+        return self._observation(obs), reward, done, info
+
+    def _display(self, obs):
```
ryanjulian: What is the purpose of this method?

garage/misc/wrappers.py
```diff
@@ -0,0 +1,88 @@
+"""Gym env wrappers"""
+import cv2
+import gym
+import numpy as np
+
+from collections import deque
+from gym.spaces import Box
+
+
+class RepeatActionWrapper(gym.core.Wrapper):
```
ryanjulian: I don't think any of these need the suffix `Wrapper`

ryanjulian: I think you should be inheriting from `gym.Wrapper` rather than `gym.core.Wrapper`

garage/misc/wrappers.py
```diff
@@ -0,0 +1,88 @@
...
+class StackFramesWrapper(gym.core.Wrapper):
+    """
+    Gym env wrapper to stack multiple frames.
+    Useful for training feed-forward agents on dynamic games.
+    """
+
+    def __init__(self, env, n_frames_stacked):
+        super(StackFramesWrapper, self).__init__(env)
+        if len(env.observation_space.shape) != 2:
+            raise Exception('Stack frames works with 2D single channel images')
```
ryanjulian: Exceptions should be as specific as possible. This is a `ValueError` because someone has provided an invalid input.

garage/misc/wrappers.py
```diff
@@ -0,0 +1,88 @@
...
+    def __init__(self, env, frame_to_repeat):
+        super(RepeatActionWrapper, self).__init__(env)
+        self.frame_to_repeat = frame_to_repeat
+
+    def step(self, action):
+        for i in range(self.frame_to_repeat):
+            obs, reward, done, info = self.env.step(action)
+        return obs, reward, done, info
+
+    def reset(self):
```
ryanjulian: Isn't this covered by the superclass?

garage/misc/wrappers.py
```diff
@@ -0,0 +1,88 @@
...
+        self.observation_space = Box(0.0, 1.0, shape=[w, h], dtype=np.float32)
+        self.w = w
+        self.h = h
+        self.plot = plot
+
+    def _observation(self, obs):
+        obs = cv2.cvtColor(obs, cv2.COLOR_RGB2GRAY)
+        obs = cv2.resize(obs, (self.w, self.h), interpolation=cv2.INTER_AREA)
+        obs = obs.astype(np.float32) / 255.0
+        if self.plot:
```
ryanjulian: You should not be plotting inside some random helper function. The only valid place to plot is when someone calls `render()`. This also means your `plot` argument is irrelevant.

garage/misc/wrappers.py
```diff
@@ -0,0 +1,88 @@
...
+        self._n_stack_past = n_frames_stacked
+        self._frames = None
+
+        new_obs_space_shape = env.observation_space.shape + (
+            n_frames_stacked, )
+        self.observation_space = Box(
+            0.0, 1.0, shape=new_obs_space_shape, dtype=np.float32)
+
+    def _frames_as_numpy(self):
+        np_frames = np.asarray(self._frames)
```
ryanjulian: This allocates new memory, but we already have memory allocated for the frame (in `observation`). Can't we concatenate the frames we already have and return them?

garage/misc/wrappers.py
ryanjulian: I think it would be best if each of these classes were in its own file.

Generally you should only put more than one class into a file if the classes are all part of the same API, or they are very obviously related.

tests/garage/misc/test_wrappers.py
```diff
@@ -0,0 +1,20 @@
...
+
+    def test_stack_frames_wrapper(self):
+        # StackFrameWrapper only supports input with channel=1
+        env = StackFramesWrapper(
+            ResizeAndGrayscaleWrapper(gym.make("Breakout-v0"), w=84, h=84),
+            n_frames_stacked=4)
+
+        env.reset()
+        obs, _, _, _ = env.step(0)
+        assert obs.shape == (84, 84, 4)
```
ryanjulian: You can verify a lot more about this than its size, so you should. You know exactly what the contents should contain.

tests/garage/misc/test_wrappers.py
```diff
@@ -0,0 +1,20 @@
+import gym
+
+from garage.misc.wrappers import ResizeAndGrayscaleWrapper, StackFramesWrapper
+from tests.fixtures import TfGraphTestCase
+
+
+class TestWrappers(TfGraphTestCase):
+    def test_resize_and_gray_scale_wrapper(self):
+        env = ResizeAndGrayscaleWrapper(gym.make("Breakout-v0"), w=84, h=84)
+        assert env.observation_space.shape == (84, 84)
```
ryanjulian: You verified it was resized but did not verify that it's greyscale.

tests/garage/misc/test_wrappers.py
ryanjulian: Why no test for repeated wrappers?

tests/garage/misc/test_wrappers.py
```diff
@@ -0,0 +1,20 @@
+import gym
+
+from garage.misc.wrappers import ResizeAndGrayscaleWrapper, StackFramesWrapper
+from tests.fixtures import TfGraphTestCase
```
ryanjulian: These classes have nothing to do with TensorFlow so they shouldn't be using this test fixture

garage/misc/wrappers.py
ahtsan: I will remove this, as user can call `render` to see.

garage/misc/wrappers.py
ahtsan: When reset is not added, it complains 
`WARN: <class 'garage.envs.wrappers.repeat_action.RepeatAction'> doesn't implement 'reset' method, which is required for wrappers derived directly from Wrapper. Deprecated default implementation is used.`

tests/garage/misc/test_wrappers.py
ahtsan: can I say that the shape is now (84, 84, (1)) so the channel now becomes 1, which means grayscale?

garage/misc/wrappers.py
ahtsan: How about scipy?

garage/envs/wrappers/resize_and_gray_scale.py
```diff
@@ -0,0 +1,36 @@
...
+    def __init__(self, env, w, h):
+        """Construct the class."""
+        super(ResizeAndGrayscale, self).__init__(env)
+        self.observation_space = Box(0.0, 1.0, shape=[w, h], dtype=np.float32)
+        self.w = w
+        self.h = h
+
+    def _observation(self, obs):
+        obs = imresize(obs, (self.w, self.h))
+        obs = np.dot(obs[:, :, :3], [0.299, 0.587, 0.114]) / 255.0
```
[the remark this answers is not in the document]
ryanjulian: Please find a library function for this. If you really need to, define a utility function and write a test for it.

I think scikit-image should have what you need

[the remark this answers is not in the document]
ahtsan: Skimage is not installed in garage. Should it?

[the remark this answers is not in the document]
CatherineSue: If you find the function in scikit-image and you decide to use it, you can add scikit-image to environments.yml.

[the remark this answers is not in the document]
ahtsan: scikit uses a different weighting as opencv for RGB = `0.2125 R + 0.7154 G + 0.0721 B`
http://scikit-image.org/docs/dev/api/skimage.color.html
I used scikit since it is a smaller dependency comparing with opencv.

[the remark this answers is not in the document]
ahtsan: result
There is a slight difference but I guess it is fine.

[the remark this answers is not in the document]
ryanjulian: interestingly i was able to see the difference on my phone display but not my laptop. macbooks are calibrated, so perhaps there is no chroma difference for a calibrated display.

anyway, i don't think it should make any difference for learning.

garage/tf/core/parameter.py
```diff
@@ -0,0 +1,44 @@
...
+        p = tf.get_variable(
+            "parameter",
+            shape=(length, ),
+            dtype=dtype,
+            initializer=initializer,
+            trainable=trainable)
+
+        ndim = input_var.get_shape().ndims
+        broadcast_shape = tf.concat(
+            axis=0, values=[tf.shape(input_var)[:ndim - 1], [length]])
```
jonashen: Can you just call `tf.shape(input_var)[:-1]`?

tests/garage/envs/test_stack_frames_env.py
```diff
@@ -0,0 +1,47 @@
...
+        self.obs_stack = self.env_stack.reset()
+        self.frame_width = self.env.observation_space.shape[0]
+        self.frame_height = self.env.observation_space.shape[1]
+
+    def test_stack_frames_multiple_channel(self):
+        with self.assertRaises(ValueError):
+            StackFrames(gym.make("Breakout-v0"), n_frames=4)
+
+    def test_stack_frames_for_reset(self):
+
```
jonashen: Extra newline

tests/garage/envs/test_stack_frames_env.py
```diff
@@ -0,0 +1,47 @@
...
+        frame_stack = self.obs
+        for i in range(3):
+            frame_stack = np.dstack((frame_stack, self.obs))
+
+        assert self.obs_stack.shape == (self.frame_width, self.frame_height, 4)
+
+        np.testing.assert_array_equal(self.obs_stack, frame_stack)
+
+    def test_stack_frames_for_step(self):
+
```
jonashen: Extra newline

garage/tf/core/parameter.py
ahtsan: Yes

garage/envs/wrappers/gray_scale.py
```diff
@@ -0,0 +1,35 @@
+"""Grayscale wrapper for gym.Env."""
+import gym
+from gym.spaces import Box
+import numpy as np
+from skimage import color
+
+
+class GrayScale(gym.Wrapper):
+    """Converting frames to grayscale."""
```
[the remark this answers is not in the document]
ahtsan: Calling env.render() on the wrapped environment still render colour image. Probably it's their way of doing it. 

Env.render(mode='rgb_array') still gives the colour image, but it returns the raw-pixel as array instead of displaying it.

But yes, if the user wants to see gray image, we have to override render() and display the processed image.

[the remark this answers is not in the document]
ryanjulian: what is the gym API contract? isn't it

* `mode == "human"` --> show a window, return none
* `mode == "rgb_array"` --> show nothing, return pixels

If I want to get pixels and show a window, I should be able to call:
```python
env.render(mode="human")  # shows a window
pixels = env.render(mode="rgb_array")  # gives me the pixels
```

garage/envs/wrappers/gray_scale.py
```diff
@@ -0,0 +1,35 @@
+"""Grayscale wrapper for gym.Env."""
+import gym
+from gym.spaces import Box
+import numpy as np
+from skimage import color
+
+
+class GrayScale(gym.Wrapper):
```
ryanjulian: `Grayscale`

garage/envs/wrappers/gray_scale.py
```diff
@@ -0,0 +1,35 @@
...
+    """Converting frames to grayscale."""
+
+    def __init__(self, env):
+        """
+        Grayscale wrapper.
```
ryanjulian: Please make this description more explicit.

garage/envs/wrappers/resize.py
```diff
@@ -0,0 +1,37 @@
...
+            height: resized frame height.
+        """
+        super().__init__(env)
+        self.observation_space = Box(
+            0.0, 1.0, shape=[width, height], dtype=np.float32)
+        self.width = width
+        self.height = height
+
+    def _observation(self, obs):
+        obs = imresize(obs, (self.width, self.height))
```
ryanjulian: Why keep obs at all?

garage/envs/wrappers/stack_frames.py
```diff
@@ -0,0 +1,58 @@
...
+        Raises:
+            ValueError: If observation space shape is not 2.
+
+        """
+        super().__init__(env)
+        if len(env.observation_space.shape) != 2:
+            raise ValueError(
+                "Stack frames only works with 2D single channel images")
+
+        self.n_frames = n_frames
```
ryanjulian: Please keep members private in non-data classes

tests/garage/envs/test_gray_scale_env.py
```diff
@@ -0,0 +1,32 @@
...
+
+import gym
+import numpy as np
+
+from garage.envs.wrappers import GrayScale
+from garage.misc.overrides import overrides
+
+
+class TestGrayScale(unittest.TestCase):
+    @overrides
```
ryanjulian: Missing newline. Does this pass pre-commit?

tests/garage/envs/test_gray_scale_env.py
```diff
@@ -0,0 +1,32 @@
...
+    def setUp(self):
+        self.env = gym.make("Breakout-v0")
+        self.env_r = GrayScale(gym.make("Breakout-v0"))
```
ryanjulian: By including gym.make you accidentally made this an integration test.

use a mock to provide a known-value input instead.

https://semaphoreci.com/community/tutorials/getting-started-with-mocking-in-python

tests/garage/envs/test_repeat_action_env.py
```diff
@@ -0,0 +1,27 @@
...
+from garage.envs.wrappers import RepeatAction
+from garage.misc.overrides import overrides
+
+
+class TestRepeatAction(unittest.TestCase):
+    @overrides
+    def setUp(self):
+        self.env_repeat = RepeatAction(
+            gym.make("Breakout-v0"), n_frame_to_repeat=4)
+        self.env = gym.make("Breakout-v0")
```
ryanjulian: Use mocks to avoid interacting with gym

https://semaphoreci.com/community/tutorials/getting-started-with-mocking-in-python

tests/garage/envs/test_resize_env.py
```diff
@@ -0,0 +1,11 @@
+import unittest
+
+import gym
+
+from garage.envs.wrappers import Resize
+
+
+class TestResize(unittest.TestCase):
+    def test_resize_output_shape(self):
+        env_r = Resize(gym.make("Breakout-v0"), width=84, height=84)
```
ryanjulian: Use mocks to avoid interacting with gym

https://semaphoreci.com/community/tutorials/getting-started-with-mocking-in-python

tests/garage/envs/test_stack_frames_env.py
```diff
@@ -0,0 +1,45 @@
...
+from garage.envs.wrappers import StackFrames
+from garage.misc.overrides import overrides
+
+
+class TestStackFrames(unittest.TestCase):
+    @overrides
+    def setUp(self):
+        self.env_stack = StackFrames(
+            GrayScale(gym.make("Breakout-v0")), n_frames=4)
+        self.env = GrayScale(gym.make("Breakout-v0"))
```
ryanjulian: Use mocks to avoid interacting with gym

garage/envs/wrappers/stack_frames.py
```diff
@@ -0,0 +1,58 @@
...
+    def __init__(self, env, n_frames):
+        """
+        Stack frames wrapper.
+
+        Args:
+            env: gym.Env to wrap.
+            n_frames: number of frames to stack.
+
+        Raises:
+            ValueError: If observation space shape is not 2.
```
ryanjulian: Would a shape of `(120, 240)` be valid? Isn't that a single-channel 2D image?

garage/tf/core/parameter.py
```diff
@@ -0,0 +1,43 @@
+"""Parameter layer in TensorFlow."""
```
ryanjulian: Did this sneak in from a different PR?

garage/envs/wrappers/gray_scale.py
ahtsan: I feel that class docstring will always contain the same content for __init__ docstring. From [google styleguide](http://google.github.io/styleguide/pyguide.html), there is no docstring in __init__. Should we skip the flake8 checking for __init__ docstring?

garage/envs/wrappers/gray_scale.py
```diff
@@ -0,0 +1,35 @@
...
+
+        Args:
+            env: gym.Env to wrap.
+        """
+        super().__init__(env)
+        width = env.observation_space.shape[0]
```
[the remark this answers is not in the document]
ahtsan: I think we should handle mujoco environment separately as it requires other dependency. Like we can have gray_scale_mujoco.

garage/envs/wrappers/stack_frames.py
ahtsan: It would be valid.

garage/tf/core/parameter.py
ahtsan: Yes, sorry for that.

ahtsan: Yes, sorry for that.

garage/envs/wrappers/gray_scale.py
```diff
@@ -0,0 +1,35 @@
...
+        height = env.observation_space.shape[1]
+        self.observation_space = Box(
```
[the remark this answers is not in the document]
ahtsan: using `@property`?

garage/envs/wrappers/gray_scale.py
ryanjulian: The Sphinx example for Google docstring style says you can document `__init__` args in either the class docstring or the `__init__` docstring, but to only do one of those and be consistent.


Let's go with only doing it in the class. This means you can edit `setup.cfg` and `scripts/travisci/check_flake8.sh` to tell flake8 to ignore [D107](http://www.pydocstyle.org/en/2.1.1/error_codes.html).

I'd appreciate it if you made a PR on https://github.com/rlworkgroup/akro/blob/master/setup.cfg to update it there as well.

garage/envs/wrappers/gray_scale.py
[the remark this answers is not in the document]
ryanjulian: yes

garage/envs/wrappers/stack_frames.py
ryanjulian: alright

garage/envs/wrappers/gray_scale.py
```diff
@@ -0,0 +1,35 @@
+"""Grayscale wrapper for gym.Env."""
```
ryanjulian: `grayscale.py`

garage/envs/wrappers/gray_scale.py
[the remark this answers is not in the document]
ahtsan: It raises `AttributeError: can't set attribute` when the gym.Env runs `self.observation_space = self.env.observation_space` in `__init__`. By adding `@property` we made `self.observation_space` immutable.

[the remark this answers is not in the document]
ryanjulian: i think you can resolve this by adding a setter.

garage/envs/wrappers/resize.py
```diff
@@ -0,0 +1,37 @@
...
+        return obs
+
+    def reset(self):
```
ryanjulian: missing from test coverage

garage/envs/wrappers/resize.py
```diff
@@ -0,0 +1,37 @@
...
+        """gym.Env reset function."""
+        return self._observation(self.env.reset())
+
+    def step(self, action):
```
ryanjulian: missing from test coverage

ahtsan:
@ryanjulian @CatherineSue @jonashen

ryanjulian:
please address/resolve all my comments and I'll take another look

garage/envs/wrappers/stack_frames.py
```diff
@@ -0,0 +1,70 @@
...
+        n_frames: number of frames to stack.
+
+    Raises:
+        ValueError: If observation space shape is not 2 or environment is not gym.spaces.Box.  # noqa: E501
+
+    """
+
+    def __init__(self, env, n_frames):
+        if not isinstance(env.observation_space, Box):
+            raise ValueError("Stack frames only works with Box environment.")
```
ryanjulian: missing from test coverage

garage/envs/wrappers/stack_frames.py
```diff
@@ -0,0 +1,70 @@
...
+        new_obs_space_shape = env.observation_space.shape + (n_frames, )
+        _low = env.observation_space.low.flatten()[0]
+        _high = env.observation_space.high.flatten()[0]
+        self._observation_space = Box(
+            _low, _high, shape=new_obs_space_shape, dtype=np.float32)
+
+    @property
+    def observation_space(self):
+        """gym.Env observation space."""
+        return self._observation_space
```
ryanjulian: missing from test coverage

PR #416, opened 2018-11-21
Add MultitaskEnv

zhanpenghe:
Part of #374. Next step will be a multi-task sampler.

### DIFF
--- a/garage/envs/multitask_env.py
+++ b/garage/envs/multitask_env.py
@@ -1,7 +1,4 @@
-"""
-This file contains a Multi-tasks Environment wrapper
-that is used for meta learning or multitask RL.
-"""
+"""Multitask environment."""
 
 import gym
 import numpy as np
@@ -9,7 +6,7 @@ import numpy as np
 from garage.core import Serializable
 
 
-class MultitaskEnv(gym.Wrapper):
+class MultitaskEnv(gym.Wrapper, Serializable):
     """
     Multitask Environment wrapper.
 
@@ -20,13 +17,14 @@ class MultitaskEnv(gym.Wrapper):
 
     def __init__(self, wrapped_env, tasks, task_sample_method="round_robin"):
         """
-        Initialize a MultitaskEnv
+        Initialize a MultitaskEnv.
 
         Args
             wrapped_env: the environment that is wrapped. Note that
                          this environment need a task space.
             tasks: a list of tasks to be sampled from.
-            task_sample_method: how to sample/reset a task. Options are: "round_robin" and "random".
+            task_sample_method: how to sample/reset a task.
+                          Options are: "round_robin" and "random".
         """
         assert wrapped_env.task_space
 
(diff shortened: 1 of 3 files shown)


garage/envs/multitask_env.py
```diff
@@ -0,0 +1,67 @@
...
+class MultitaskEnv(gym.Wrapper, Serializable):
+    """
+    Multitask Environment wrapper.
+
+    This class take an task-based environment and wrap it.
+    It provides functions like resetting/sampling tasks
+    that is used for multitask sampling.
+    """
+
+    def __init__(self, wrapped_env, tasks, task_sample_method="round_robin"):
```
ryanjulian: What is the data type of `tasks`? If `wrapped_env` needs a task space, then what is the purpose of tasks?

garage/envs/point_env.py
```diff
@@ -6,7 +6,12 @@
 
 
 class PointEnv(gym.Env, Serializable):
```
ryanjulian: Maybe start a PR to move the embed2learn PointEnv into here?

garage/envs/point_env.py
```diff
@@ -28,13 +33,21 @@ def reset(self):
...
         return Step(observation=next_observation, reward=reward, done=done)
 
-    def render(self):
+    def render(self, mode=None):
         print('current state:', self._state)
 
     def log_diagnostics(self, paths):
         pass
+
+    def reset_task(self, task):
```
ryanjulian: ## Don't create interfaces without specifying them
Violating the `gym.Env` interface for convenience is not the way to go here. You have added a new interface (`reset_task`, `task_space`) but did not endorsed anywhere what that interface is.

## Meet existing code where it is -- changes are expensive
It's also very inflexible to force the base `gym.Env` to implement a multitask env interface. The entire purpose of a `MultitaskEnv` *wrapper* should be to *avoid* forcing a new interface onto an existing environment, since most environments are not multitask. You will of course need to assert that all of those envs have the same action space and observation space.

## Separate the API contract from the implementation
Of course, there's nothing wrong with implementing an environment which implements a multitask API directly (it's strictly better), but in the beginning most environments simply won't do that.

If you want an API contract for Multitask environments, you need to say so. You've spread the API across two layers here, in a pretty confusing way. Why is there even a multitask wrapper if you forced PointEnv to implement the multitask interface???

Here's a blueprint for this
* An ABC specifying the set of functions a MultitaskEnv needs to provide (it does not provide any of them itself). 
* A wrapper which takes a list of `gym.Env`'s and turns them into a `MultitaskEnv` (it implements your MultitaskEnv interface)
* (optional) An example of an environment implementing MultitaskEnv directly (MultitaskPointEnv would be a decent option)

My existing implementation in

ryanjulian:
Meta-RL is learning over potentially many POMDPs, but those POMDPs might all have the same task.

Perhaps a better term for this is MultiEnv or MetaEnv -- an environment which is actually many environments.

* For now it can take a constructor with a list of environments. In the future, it might make more sense for some algorithms to instead provide a thunk/future/generator which generates a new environment when asked.

For instance, you can phrase dynamics randomization as a MultiEnv whose thunk produces an environment with new dynamics every time it is executed.

* We can also provide a method on this class which returns a gym.Env backed by the MultiEnv, but switches the underlying env every `reset()`. That makes for easy compatibility with single task algorithms or meta-RL algorithms which don't need task labels.

zhanpenghe:
This will be reopen later.

PR #418, opened 2018-11-26
Fix bugs in tf/VPG

CatherineSue:
1. Fix typo in pg_loss.
2. The default batch_size of the optimizer in VPG should be None.
3. The default max_epochs of the optimizer in VPG should be 1.

Refer: #417

PR #427, opened 2018-12-05
Fix bugs in tf/VPG

CatherineSue:
See: #423, #418


ryanjulian:
Please use the same commit title as on the commit on master.

PR #430, opened 2018-12-07
v2018.10: Fix bugs in tf/VPG (#418)

ryanjulian:
1. Fix typo in pg_loss.
2. The default batch_size of the optimizer in VPG should be None.
3. The default max_epochs of the optimizer in VPG should be 1.


ryanjulian:
Also includes commits for #428

Issue #433, opened 2018-12-18
garage/envs/base.py calls tests.quirks


amolchanov86:
As the title says one of the garage modules calls tests.quirks which semantically should be part of the garage module, for example, garage.tests.quirks instead (or any other if you would like).
The problem arrives when I am trying to run something inside of my projects folder, obviously, since tests are somewhere up the folder hierarchy it can not find it. 
At the same time, the name "tests" is too generic to just put it into PYTHONPATH. 
Hence, I believe, some restructuring is appropriate.

ryanjulian:
good catch!

@jonashen @gonzaiva can one of you fix this? I think it would be sufficient to just copy `KNOWN_GYM_NOT_CLOSER_VIEWER` to `base.py`.

This one will require a backport into `release-2018.10`

gonzaiva:
Okay, I'll work on this today.

gonzaiva:
`KNOWN_GYM_NOT_CLOSER_VIEWER` was moved to `base.py` in #434, however, should we close this issue until we have decided how to repackage `tests`?

gonzaiva:
I understood by Artem's issue that `tests` is too generic for the PYTHONPATH, so we should repackage it as `garage.tests`.

amolchanov86:
I guess after `KNOWN_GYM_NOT_CLOSER_VIEWER` moving to the `base.py` and excluding import of tests it should be fine.

ryanjulian:
Ok. We will close once Angel backports it to `release-2018.10`

PR #434, opened 2018-12-18
Move KNOWN_GYM_NOT_CLOSE_VIEWER to envs base

gonzaiva:
Since tests is one step above garage.envs in the directory hierarchy,
the class GarageEnv cannot be correctly imported when running code from
inside the garage folder.
This dependency is removed by moving the list KNOWN_GYM_NOT_CLOSE_VIEWER
which is the only dependency from garage/envs/base in tests/quirks.

### DIFF
--- a/garage/envs/base.py
+++ b/garage/envs/base.py
@@ -8,7 +8,14 @@ import gym
 
 from garage.core import Parameterized
 from garage.core import Serializable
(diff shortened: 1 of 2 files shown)


gonzaiva:
See issue #433 
@ryanjulian, I will do cherry-pick on v2018.10 once this PR gets approved

PR #440, opened 2018-12-18
Add git-ignored files in flake8 excluded files

gonzaiva:
Since now flake8 runs locally before pushing a commit remotely, it may
throw errors for ignored files. This commit fixes that by adding all the
files in .gitignore in all the lists of excluded files inside the script
check_flake8.


ryanjulian:
Please backport this to 2018.10 if it's a clean merge

PR #441, opened 2018-12-19
Add tf/ERWR

CatherineSue:
1. Add tf/ERWR.
2. Fix warning in garage.baselines.LinearFeatureBaseline.
3. Set seed in TfGraphTestCase to avoid large flakiness.

### DIFF
--- a/garage/tf/algos/erwr.py
+++ b/garage/tf/algos/erwr.py
@@ -1,0 +1,39 @@
+"""Episodic Reward Weighted Regression."""
+from garage.tf.algos.vpg import VPG
+from garage.tf.optimizers import LbfgsOptimizer
+
+
+class ERWR(VPG):
+    """
+    Episodic Reward Weighted Regression [1].
+
+    Notes
+    -----
+    This does not implement the original RwR [2]_ that deals with "immediate
+    reward problems" since it doesn't find solutions that optimize for
+    temporally delayed rewards.
+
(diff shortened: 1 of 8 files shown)


garage/tf/algos/erwr.py
```diff
@@ -0,0 +1,39 @@
...
+                 positive_adv=None,
+                 **kwargs):
+        if optimizer is None:
+            optimizer = LbfgsOptimizer
+            if optimizer_args is None:
+                optimizer_args = dict()
+        super(ERWR, self).__init__(
+            optimizer=optimizer,
+            optimizer_args=optimizer_args,
+            positive_adv=True if positive_adv is None else positive_adv,
```
ryanjulian: Isn't this equivalent to `positive_adv=True` above?

CatherineSue: Yes. Addressed.

garage/tf/algos/erwr.py
```diff
@@ -0,0 +1,39 @@
...
+    def __init__(self,
+                 optimizer=None,
+                 optimizer_args=None,
+                 positive_adv=True,
+                 **kwargs):
+        if optimizer is None:
+            optimizer = LbfgsOptimizer
+            if optimizer_args is None:
+                optimizer_args = dict()
+        super(ERWR, self).__init__(
```
ahtsan: just a comment: super().__init__ also works

ryanjulian: super().__init__() is the preferred style in Python 3

PR #445, opened 2018-12-20
Fix missing flatten method

bohan-zhang:
Fix the error when calling rollout:
```
  File "./scripts/sim_policy.py", line 43, in <module>
    main()
  File "./scripts/sim_policy.py", line 39, in main
    simulate_policy(args)
  File "./scripts/sim_policy.py", line 34, in simulate_policy
    animated=True, speedup=args.speedup)
  File "./garage/sampler/utils.py", line 28, in rollout
    observations.append(env.observation_space.flatten(o))
AttributeError: 'Box' object has no attribute 'flatten'
```

### DIFF
--- a/garage/sampler/utils.py
+++ b/garage/sampler/utils.py
@@ -3,6 +3,7 @@ import time
 
 import numpy as np
 
+from garage.envs.util import flatten
 from garage.misc import tensor_utils
 
 
@@ -25,9 +26,9 @@ def rollout(env,
     while path_length < max_path_length:
         a, agent_info = agent.get_action(o)
         next_o, r, d, env_info = env.step(a)
-        observations.append(env.observation_space.flatten(o))
+        observations.append(flatten(env.observation_space, o))
         rewards.append(r)
-        actions.append(env.action_space.flatten(a))
+        actions.append(flatten(env.action_space, a))
         agent_infos.append(agent_info)
         env_infos.append(env_info)
         path_length += 1


bohan-zhang:
It seems to be the same reason that my PR failed last time @ryanjulian and I couldn't write a clean unit test as it requires to write a custom agent / policy.

ryanjulian:
Thanks for the PR! I think there is some confusion.

This should not be necessary if your environment is properly wrapped in [TfEnv](https://github.com/rlworkgroup/garage/blob/master/garage/tf/envs/base.py#L16) or [TheanoEnv](https://github.com/rlworkgroup/garage/blob/master/garage/theano/envs/base.py).

Right now, garage code assumes you are using garage.spaces, not gym.spaces. TfEnv and TheanoEnv silently convert the types for you, and are required wrappers for using gym.Env.

I realize this is not optimal and we are working hard to make this easier-to-use (see https://github.com/rlworkgroup/akro).

For now I think the appropriate pull request would be to update documentation to make this more clear. If you find an example or script which does not use TfEnv/TheanoEnv, please let us know so we can fix it.

bohan-zhang:
Hey @ryanjulian I'm fixing a library, `softqlearning`, that's creating an environment inheriting from `MujocoEnv` from `garage` and it's causing this bug!

ryanjulian:
Can you provide a link to the launcher file which triggers the bug?

It's likely that some code in the library needs to be updated, since garage changed some of the rllab APIs.

ryanjulian:
This is reversed somewhere in your algorithms (or in our code).

The sampler should only encounter `garage.Spaces` objects, which have the `flatten()` method.

ryanjulian:
Did you wrap your environment in TfEnv?

bohan-zhang:
Hey @ryanjulian I did not and I will consider doing that! Thank you so much for your help and I'll close this PR now :)

Issue #436, opened 2018-12-18
TF still depends on Theano


amolchanov86:
Somewhere TF scripts still depend on theano.
For example try running:
```sh
export THEANO_FLAGS="floatX=float32,device=**gpu**"
python examples/tf/ppo_pendulum.py
```
It will complain:
ValueError: You are trying to use the old GPU back-end. It was removed from Theano. Use device=cuda* now

PR #446, opened 2018-12-21
Conditionally import lasagne or tf to set the seed

gonzaiva:
When launching experiments with run_experiment the function set_seed is
in charge of setting the random seed across the different packages used
in garage. However, we should keep from import theano when only using
tensorflow and vice versa.
To detect if theano and/or tensorflow are being used, we check if they
appear in the list of packages already imported.


gonzaiva:
See issue #436

PR #447, opened 2018-12-21
Add vanilla loss to NPO

CatherineSue:
1. VPG uses a different loss function than NPO and its variants. VPG
uses the vanilla loss, as VPG uses direct gradient estimator from policy
score function, NPO calculates the derivation using importance sampling.
Thus, we need to add the 'real' vanilla loss to NPO to inherit from NPO.
The Enum is kept in this commit instead of string.
2. Update examples.

### DIFF
--- a/examples/tf/ppo_pendulum.py
+++ b/examples/tf/ppo_pendulum.py
@@ -39,7 +39,7 @@ def run_task(*_):
         max_path_length=100,
         n_itr=488,
         discount=0.99,
-        step_size=0.01,
+        lr_clip_range=0.01,
         optimizer_args=dict(batch_size=32, max_epochs=10),
         plot=False)
     algo.train()
--- a/garage/tf/algos/npo.py
+++ b/garage/tf/algos/npo.py
@@ -1,3 +1,4 @@
+"""Natural Policy Gradient Optimization."""
 from enum import Enum
 from enum import unique
 
@@ -23,13 +24,28 @@ from garage.tf.optimizers import LbfgsOptimizer
 @unique
 class PGLoss(Enum):
     VANILLA = "vanilla"
-    CLIP = "clip"
+    SURROGATE = "surrogate"
+    SURROGATE_CLIP = "surrogate_clip"
 
 
 class NPO(BatchPolopt):
+    """
+    Natural Policy Gradient Optimization.
+
+    Attributes:
+        name(str): The name of the algorithm.
+        lr_clip_range(float): The limit on the likelihood ratio between
+            policies, as in PPO.
+        max_kl_step(float): The maximum KL divergence between old and new
+            policies, as in TRPO.
+        policy_ent_coeff(float): The coefficient of the policy entropy.
+        optimizer(float): The optimizer of the algorithm.
+    """
+
     def __init__(self,
-                 pg_loss=PGLoss.VANILLA,
-                 clip_range=0.01,
+                 pg_loss=PGLoss.SURROGATE,
+                 lr_clip_range=0.01,
+                 max_kl_step=0.01,
                  optimizer=None,
                  optimizer_args=None,
                  name="NPO",
@@ -53,10 +69,11 @@ class NPO(BatchPolopt):
 
         with self._name_scope:
             self.optimizer = optimizer(**optimizer_args)
-            self.clip_range = float(clip_range)
+            self.lr_clip_range = float(lr_clip_range)
+            self.max_kl_step = float(max_kl_step)
             self.policy_ent_coeff = float(policy_ent_coeff)
 
-        super(NPO, self).__init__(policy=policy, **kwargs)
+        super().__init__(policy=policy, **kwargs)
 
     @overrides
     def init_opt(self):
@@ -67,7 +84,7 @@ class NPO(BatchPolopt):
         self.optimizer.update_opt(
             loss=pol_loss,
             target=self.policy,
-            leq_constraint=(pol_kl, self.clip_range),
+            leq_constraint=(pol_kl, self.max_kl_step),
             inputs=flatten_inputs(self._policy_opt_inputs),
             constraint_name="mean_kl")
 
@@ -291,8 +308,23 @@ class NPO(BatchPolopt):
                     )
                     pol_mean_kl = tf.reduce_mean(kl)
 
+            # Calculate vanilla loss
+            with tf.name_scope("vanilla_loss"):
+                if self.policy.recurrent:
+                    ll = pol_dist.log_likelihood_sym(
+                        i.action_var, policy_dist_info, name="log_likelihood")
+
+                    vanilla = ll * advantages * i.valid_var
+                else:
+                    ll = pol_dist.log_likelihood_sym(
+                        i.valid.action_var,
+                        policy_dist_info_valid,
+                        name="log_likelihood")
+
+                    vanilla = ll * adv_valid
+
             # Calculate surrogate loss
-            with tf.name_scope("surr_loss"):
+            with tf.name_scope("surrogate_loss"):
                 if self.policy.recurrent:
                     lr = pol_dist.likelihood_ratio_sym(
                         i.action_var,
@@ -300,7 +332,7 @@ class NPO(BatchPolopt):
                         policy_dist_info,
                         name="lr")
 
-                    surr_vanilla = lr * advantages * i.valid_var
+                    surrogate = lr * advantages * i.valid_var
                 else:
                     lr = pol_dist.likelihood_ratio_sym(
                         i.valid.action_var,
@@ -308,33 +340,36 @@ class NPO(BatchPolopt):
                         policy_dist_info_valid,
                         name="lr")
 
-                    surr_vanilla = lr * adv_valid
+                    surrogate = lr * adv_valid
 
+            # Finalize objective function
+            with tf.name_scope("loss"):
                 if self._pg_loss == PGLoss.VANILLA:
-                    # VPG, TRPO use the standard surrogate objective
-                    surr_obj = tf.identity(surr_vanilla, name="surr_obj")
-                elif self._pg_loss == PGLoss.CLIP:
+                    # VPG uses the vanilla objective
+                    obj = tf.identity(vanilla, name="vanilla_obj")
+                elif self._pg_loss == PGLoss.SURROGATE:
+                    # TRPO uses the standard surrogate objective
+                    obj = tf.identity(surrogate, name="surr_obj")
+                elif self._pg_loss == PGLoss.SURROGATE_CLIP:
                     lr_clip = tf.clip_by_value(
                         lr,
-                        1 - self.clip_range,
-                        1 + self.clip_range,
+                        1 - self.lr_clip_range,
+                        1 + self.lr_clip_range,
                         name="lr_clip")
                     if self.policy.recurrent:
                         surr_clip = lr_clip * advantages * i.valid_var
                     else:
                         surr_clip = lr_clip * adv_valid
-                    surr_obj = tf.minimum(
-                        surr_vanilla, surr_clip, name="surr_obj")
+                    obj = tf.minimum(surrogate, surr_clip, name="surr_obj")
                 else:
                     raise NotImplementedError("Unknown PGLoss")
 
                 # Maximize E[surrogate objective] by minimizing
                 # -E_t[surrogate objective]
                 if self.policy.recurrent:
-                    surr_loss = (-tf.reduce_sum(surr_obj)) / tf.reduce_sum(
-                        i.valid_var)
+                    loss = -tf.reduce_sum(obj) / tf.reduce_sum(i.valid_var)
                 else:
-                    surr_loss = -tf.reduce_mean(surr_obj)
+                    loss = -tf.reduce_mean(obj)
 
             # Diagnostic functions
             self.f_policy_kl = tensor_utils.compile_function(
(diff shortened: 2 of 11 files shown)


CatherineSue:
Regression of tf/VPG vs. theano/VPG.
screen shot 2018-12-20 at 5 14 18 pm

ryanjulian:
isn't PenaltyLbfgsOptimizer TRPO? Because the penalty term is for KL-divergence?

garage/tf/algos/npo.py
```diff
@@ -17,18 +17,19 @@
...
 from garage.tf.misc.tensor_utils import graph_inputs
-from garage.tf.optimizers import LbfgsOptimizer
+from garage.tf.optimizers import PenaltyLbfgsOptimizer
 
 
 @unique
 class PGLoss(Enum):
     VANILLA = "vanilla"
-    CLIP = "clip"
+    SURROGATE_VANILLA = "surrogate_vanilla"
```
ryanjulian: I think just `SURROGATE`/`"surrogate"` is fine here

garage/tf/algos/npo.py
```diff
@@ -300,6 +301,10 @@ def _build_policy_loss(self, i):
                         policy_dist_info,
                         name="lr")
 
+                    ll = pol_dist.log_likelihood_sym(
```
ryanjulian: seems like the scope above should now just be "loss" since it's not necessarily surrogate

garage/tf/algos/npo.py
```diff
@@ -308,12 +313,21 @@ def _build_policy_loss(self, i):
...
+                        policy_dist_info_valid,
+                        name="log_likelihood")
+
+                    vanilla = ll * adv_valid
                     surr_vanilla = lr * adv_valid
 
                 if self._pg_loss == PGLoss.VANILLA:
-                    # VPG, TRPO use the standard surrogate objective
+                    # VPG uses the vanilla objective
+                    surr_obj = tf.identity(vanilla, name="surr_obj")
```
ryanjulian: i think the variable `surr_obj` now needs to be renamed since it's not necessarily surrogate

garage/tf/algos/npo.py
```diff
@@ -308,12 +313,21 @@ def _build_policy_loss(self, i):
                         policy_dist_info_valid,
                         name="lr")
 
+                    ll = pol_dist.log_likelihood_sym(
+                        i.valid.action_var,
+                        policy_dist_info_valid,
+                        name="log_likelihood")
+
+                    vanilla = ll * adv_valid
```
ryanjulian: I think it might be more clear if you moved `vanilla` and `surr_vanilla` inside the `if` statement, even if it means copying a line or two.

garage/tf/algos/npo.py
```diff
@@ -300,6 +301,10 @@ def _build_policy_loss(self, i):
                         policy_dist_info,
                         name="lr")
 
+                    ll = pol_dist.log_likelihood_sym(
+                        i.action_var, policy_dist_info, name="log_likelihood")
+
+                    vanilla = ll * advantages * i.valid_var
```
ryanjulian: i think it might be more clear if you moved these inside the `if` statement, even if it means copying a couple lines

garage/tf/algos/npo.py
```diff
@@ -17,18 +17,19 @@
...
-    CLIP = "clip"
+    SURROGATE_VANILLA = "surrogate_vanilla"
+    SURROGATE_CLIP = "surrogate_clip"
 
 
 class NPO(BatchPolopt):
     def __init__(self,
-                 pg_loss=PGLoss.VANILLA,
+                 pg_loss=PGLoss.SURROGATE_VANILLA,
                  clip_range=0.01,
```
ryanjulian: while we are here, i think we should separate the `clip_range` and `step_size` parameters, since they are different numbers in the literature.

`step_size` refers to the maximum KL divergence between old and new policies, as in TRPO
`clip_range` refers to the limit on the likelihood ratio between policies, as in PPO.

garage/tf/algos/npo.py
```diff
@@ -49,7 +50,7 @@ def __init__(self,
         if optimizer is None:
             if optimizer_args is None:
                 optimizer_args = dict()
-            optimizer = LbfgsOptimizer
+            optimizer = PenaltyLbfgsOptimizer
```
ryanjulian: I think this is an error. True VPG is unconstrainted in KL divergence. Using `PenaltyLbfgsOptimizer` turns it into the soft-penalty version of TRPO.

@amolchanov86 do you agree?

amolchanov86: you are right

garage/tf/algos/npo.py
CatherineSue: Is the `step_size` same as the `epsilon` in f(kl) < epsilon?

garage/tf/algos/npo.py
amolchanov86: The shorter the better

garage/tf/algos/npo.py
amolchanov86: I would then explicitly rename them for what they are `kl_max_step` , `loss_clip_range`

CatherineSue:
This is accidentally closed because of a wrong git reset cmd, I've pushed the new changes.

garage/tf/algos/npo.py
```diff
@@ -310,10 +326,15 @@ def _build_policy_loss(self, i):
 
                     surr_vanilla = lr * adv_valid
 
+            # Finalize objective function
+            with tf.name_scope("final_loss"):
```
ryanjulian: just "loss"

garage/tf/algos/npo.py
```diff
@@ -310,10 +326,15 @@ def _build_policy_loss(self, i):
 
                     surr_vanilla = lr * adv_valid
```
ryanjulian: "surrogate", to be consistent with the enum

garage/tf/algos/npo.py
```diff
@@ -291,6 +292,21 @@ def _build_policy_loss(self, i):
...
+                else:
+                    ll = pol_dist.log_likelihood_sym(
+                        i.valid.action_var,
+                        policy_dist_info_valid,
+                        name="log_likelihood")
+
+                    vanilla = ll * adv_valid
+
             # Calculate surrogate loss
             with tf.name_scope("surr_loss"):
```
ryanjulian: "surrogate_loss"

garage/tf/algos/npo.py
```diff
@@ -323,18 +344,16 @@ def _build_policy_loss(self, i):
...
+                    obj = tf.minimum(surr_vanilla, surr_clip, name="surr_obj")
                 else:
                     raise NotImplementedError("Unknown PGLoss")
 
                 # Maximize E[surrogate objective] by minimizing
                 # -E_t[surrogate objective]
                 if self.policy.recurrent:
-                    surr_loss = (-tf.reduce_sum(surr_obj)) / tf.reduce_sum(
-                        i.valid_var)
+                    loss = (-tf.reduce_sum(obj)) / tf.reduce_sum(i.valid_var)
```
ryanjulian: unnecessary ()

garage/tf/algos/npo.py
```diff
@@ -323,18 +344,16 @@ def _build_policy_loss(self, i):
                         surr_clip = lr_clip * advantages * i.valid_var
                     else:
                         surr_clip = lr_clip * adv_valid
-                    surr_obj = tf.minimum(
-                        surr_vanilla, surr_clip, name="surr_obj")
+                    obj = tf.minimum(surr_vanilla, surr_clip, name="surr_obj")
                 else:
                     raise NotImplementedError("Unknown PGLoss")
```
ryanjulian: can you test this case?

garage/tf/algos/npo.py
```diff
@@ -323,18 +344,16 @@ def _build_policy_loss(self, i):
                         surr_clip = lr_clip * advantages * i.valid_var
```
ryanjulian: it seems that the recurrent case here is never tested. can you add a test which uses recurrent too, so that this line is covered?

garage/tf/algos/npo.py
CatherineSue: Tested. Also tested "Unknown KLConstraint"

garage/tf/algos/npo.py
CatherineSue: Added in tests/garage/tf/algos/test_ppo.py.

garage/tf/algos/npo.py
CatherineSue: Deleted

setup.cfg
```diff
@@ -10,6 +10,8 @@ import-order-style = google
 application-import-names = tests,sandbox,garage,examples
 
 [pylint]
+#[MASTER]
+init-hook='import sys; sys.path.append("./")'
```
ryanjulian: did this slip in on accident?

CatherineSue: Oh Yes. I'll delete it now.

garage/tf/algos/npo.py
```diff
@@ -23,13 +24,28 @@
...
+    SURROGATE_CLIP = "surrogate_clip"
 
 
 class NPO(BatchPolopt):
+    """
+    Natural Policy Gradient Optimization.
+
+    Attributes:
+        name(str): The name of the algorithm.
+        loss_clip_range(float): The limit on the likelihood ratio between
```
ryanjulian: lr_clip_range

examples/tf/ppo_pendulum.py
```diff
@@ -39,7 +39,7 @@ def run_task(*_):
         max_path_length=100,
         n_itr=488,
         discount=0.99,
-        step_size=0.01,
+        loss_clip_range=0.01,
```
CatherineSue: I've separated `loss_clip_range` and `max_kl_step`. Theano/NPO still uses `step_size`, but I think we can leave it there. 
There is a problem when I change the corresponding params in tests/benchmarks/test_benchmark_*.py. The pylint raises the following errors:
```
hookid: pylint

************* Module tests.benchmarks.test_benchmark_ppo
tests/benchmarks/test_benchmark_ppo.py:E0401: 14,0: Unable to import 'baselines.bench'
tests/benchmarks/test_benchmark_ppo.py:E0401: 16,0: Unable to import 'baselines.bench'
tests/benchmarks/test_benchmark_ppo.py:E0401: 18,0: Unable to import 'baselines.common'
tests/benchmarks/test_benchmark_ppo.py:E0401: 19,0: Unable to import 'baselines.common.vec_env.dummy_vec_env'
tests/benchmarks/test_benchmark_ppo.py:E0401: 20,0: Unable to import 'baselines.common.vec_env.vec_normalize'
tests/benchmarks/test_benchmark_ppo.py:E0401: 21,0: Unable to import 'baselines.logger'
tests/benchmarks/test_benchmark_ppo.py:E0401: 22,0: Unable to import 'baselines.ppo2'
tests/benchmarks/test_benchmark_ppo.py:E0401: 23,0: Unable to import 'baselines.ppo2.policies'
tests/benchmarks/test_benchmark_ppo.py:W0611: 17,0: Unused common imported from baselines
************* Module tests.benchmarks.test_benchmark_trpo
tests/benchmarks/test_benchmark_trpo.py:E0401: 15,0: Unable to import 'baselines.bench'
tests/benchmarks/test_benchmark_trpo.py:E0401: 16,0: Unable to import 'baselines.common.cmd_util'
tests/benchmarks/test_benchmark_trpo.py:E0401: 17,0: Unable to import 'baselines.common.tf_util'
tests/benchmarks/test_benchmark_trpo.py:E0401: 18,0: Unable to import 'baselines.ppo1.mlp_policy'
tests/benchmarks/test_benchmark_trpo.py:E0401: 19,0: Unable to import 'baselines.trpo_mpi'
```
Anybody have ideas why?

ryanjulian: this is because pylint is now checking whether all of your imports are actually importable from the default garage install.

are you sure that those baselines modules exist in the default environment.yml? the packages changed and it needs updating?

garage/tf/algos/ppo.py
```diff
@@ -24,7 +24,7 @@ def __init__(self, optimizer=None, optimizer_args=None, **kwargs):
             if optimizer_args is None:
                 optimizer_args = dict()
```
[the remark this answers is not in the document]
ryanjulian: no i think this is okay

garage/tf/algos/npo.py
naeioi: This is probably not an error. This is not relevant to VPG because VPG overrides optimizer to FirstOrderOptimizer. I think Chang was to sync up TF with Theano. https://github.com/rlworkgroup/garage/blob/241ae5c02f17e6fb4c6ec59b35454fe4b3dd1c23/garage/theano/algos/npo.py#L26

NPO also uses interface unique of PenaltyLbfgsOptimizer. https://github.com/rlworkgroup/garage/blob/241ae5c02f17e6fb4c6ec59b35454fe4b3dd1c23/garage/theano/algos/npo.py#L93-L99

ryanjulian: ah i see. so it doesn't matter.

i guess Artem and I just assumed that VPG would be the "default" version of NPO.

ryanjulian: I guess in the end it doesn't matter because nobody runs NPO directly, and if an algorithm uses a specific optimizer it will set it in the constructor.

I'm ambivalent about whether to revert it or not. Whatever Chang wants.

ryanjulian:
Anything blocking merging this?

PR #448, opened 2018-12-21
Backport PRs 434 and 446

gonzaiva:


naeioi:
Seems like Travis failed because it tested flake8 against diff between master and backport-434-446 branch. backport-434-446 contains some old version of files that fails flake8.

ryanjulian:
Verified there is no simple fix, so we will just force-push.

PR #451, opened 2018-12-21
Add dockerfiles for garage

gonzaiva:
Two versions were added:
- Release: image based on the latest stable release from garage
- Local: image built on top of the release image with local changes of
the garage repository

A makefile was added as well to build and run the corresponding
containers, as well as a README.md with the documentation for the docker
images.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,9 @@ addons:
     packages:
       - docker-ce
 
+env:
+  - DOCKER_COMPOSE_VERSION=1.23.2
+
 python:
   - "3.6.6"
 
@@ -24,49 +27,29 @@ before_install:
       "max-concurrent-uploads": 50
     }' | sudo tee /etc/docker/daemon.json
   - sudo service docker restart
-  # prevent garage from exiting on config
-  - cp garage/config_personal_template.py garage/config_personal.py
+  # Reinstall docker compose with the specified version. For more information, see:
+  # https://docs.travis-ci.com/user/docker/#using-docker-compose
+  - sudo rm /usr/local/bin/docker-compose
+  - curl -L https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-`uname -s`-`uname -m` > docker-compose
+  - chmod +x docker-compose
+  - sudo mv docker-compose /usr/local/bin
   # Pull cached docker image
   - docker pull rlworkgroup/garage-ci:latest
+  # Pull CodeClimate reporter
+  - curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > cc-test-reporter
+  - chmod +x cc-test-reporter
 
 install:
   - tag="rlworkgroup/garage-ci:${TRAVIS_BUILD_NUMBER}"
-  - |
-    docker build \
-      -f docker/Dockerfile.ci \
-      --cache-from rlworkgroup/garage-ci:latest \
-      --build-arg MJKEY="${MJKEY}" \
-      -t "${tag}" \
-      .
-
-before_script:
-  # Codecov utility to capture environment variables it needs
-  # This should be passed as an environment variable after `docker run`.
-  # e.g. `docker run ${ci_env} ...`
-  - ci_env="$(bash <(curl -s https://codecov.io/env))"
-  - |
-    docker-run() {
-      docker run \
-      ${ci_env} \
-      -e COVERALLS_REPO_TOKEN \
-      -e COVERALLS_SERVICE_NAME \
-      -e CODACY_PROJECT_TOKEN \
-      -e CC_TEST_REPORTER_ID \
-      -e TRAVIS_BRANCH \
-      -e TRAVIS_PULL_REQUEST \
-      -e TRAVIS_COMMIT_RANGE \
-      -e TRAVIS \
-      -e MJKEY \
-      "${tag}" "$@"
-    }
+  - TAG="${tag}" make build-ci
 
 script:
-  - docker-run scripts/travisci/check_precommit.sh
+  - TAG="${tag}" make run-ci RUN_CMD="scripts/travisci/check_precommit.sh"
   - |
     if [[ "${TRAVIS_EVENT_TYPE}" == "cron" ]]; then
-      docker-run scripts/travisci/check_cronjob_tests.sh
+      TAG="${tag}" make run-ci RUN_CMD="scripts/travisci/check_cronjob_tests.sh"
     else
-      docker-run scripts/travisci/check_tests.sh
+      TAG="${tag}" make run-ci RUN_CMD="scripts/travisci/check_tests.sh"
     fi
 
 after_success:
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,89 @@
+.PHONY: build-ci build-headless build-nvidia \
+	run-ci run-headless run-nvidia
+
+# Path in host where the experiment data obtained in the container is stored
+DATA_PATH ?= $(shell pwd)/data
+# Set the environment variable MJKEY with the contents of the file specified by
+# MJKEY_PATH.
+MJKEY_PATH ?= ~/.mujoco/mjkey.txt
+# Prevent garage from exiting on config by making sure the personal config file
+# is already created
+CONFIG_PERSONAL := garage/config_personal.py
+
+build-ci: TAG ?= rlworkgroup/garage-ci
+build-ci: docker/docker-compose-headless.yml copy_config_personal
+	TAG=${TAG} \
+	docker-compose \
+		-f docker/docker-compose-headless.yml \
+		build \
+		--build-arg MJKEY="$${MJKEY}"
+
+build-headless: TAG ?= rlworkgroup/garage-headless
+build-headless: docker/docker-compose-headless.yml check-mjkey copy_config_personal
+	TAG=${TAG} \
+	docker-compose \
+		-f docker/docker-compose-headless.yml \
+		build \
+		--build-arg MJKEY="$$(cat ${MJKEY_PATH})"
+
+build-nvidia: docker/docker-compose-nvidia.yml check-mjkey copy_config_personal
+	docker-compose \
+		-f docker/docker-compose-nvidia.yml \
+		build \
+		--build-arg MJKEY="$$(cat ${MJKEY_PATH})"
+
+run-ci: SHELL := /usr/bin/env bash
+run-ci: TAG ?= rlworkgroup/garage-ci
+# The CI container requires environment variables to run CodeCov
+run-ci: CI_ENV := $$(bash <(curl -s https://codecov.io/env))
+run-ci:
+	docker run \
+		"${CI_ENV}" \
+		-e COVERALLS_REPO_TOKEN \
+		-e COVERALLS_SERVICE_NAME \
+		-e CODACY_PROJECT_TOKEN \
+		-e CC_TEST_REPORTER_ID \
+		-e TRAVIS_BRANCH \
+		-e TRAVIS_PULL_REQUEST \
+		-e TRAVIS_COMMIT_RANGE \
+		-e TRAVIS \
+		-e MJKEY \
+		${TAG} ${RUN_CMD}
+
+run-headless: CONTAINER_NAME ?= garage-headless
+run-headless: build-headless
+	docker run \
+		-it \
+		--rm \
+		-v $(DATA_PATH)/data/$(CONTAINER_NAME):/root/code/garage/data \
+		-e MJKEY="$$(cat $(MJKEY_PATH))" \
+		--name $(CONTAINER_NAME) \
+		rlworkgroup/garage-headless $(RUN_CMD)
+
+run-nvidia: CONTAINER_NAME ?= garage-nvidia
+run-nvidia: build-nvidia
+	xhost +local:docker
+	docker run \
+		-it \
+		--rm \
+		--runtime=nvidia \
+		-v /tmp/.X11-unix:/tmp/.X11-unix \
+		-v $(DATA_PATH)/data/$(CONTAINER_NAME):/root/code/garage/data \
+		-e DISPLAY=$(DISPLAY) \
+		-e QT_X11_NO_MITSHM=1 \
+		-e MJKEY="$$(cat $(MJKEY_PATH))" \
+		--name $(CONTAINER_NAME) \
+		rlworkgroup/garage-nvidia $(RUN_CMD)
+
+check-mjkey:
+ifeq (0, $(shell [ ! -f $(MJKEY_PATH) ]; echo $$? ))
+	$(error The MJKEY was not set: make sure to pass a valid MJKEY_PATH \
+		or put your key at ~/.mujoco/mjkey.txt)
+endif
+
+copy_config_personal:
+ifeq (0, $(shell [ ! -f $(CONFIG_PERSONAL) ]; echo $$? ))
+	$(warning No file for personal configuration was found. Copying from \
+		garage/config_personal_template.py)
+	cp garage/config_personal_template.py garage/config_personal.py
+endif
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -2,7 +2,8 @@
 # and scripts/setup_macos.sh, if there's any changes applied to this file,
 # specially regarding the installation of dependencies, apply those same
 # changes to the mentioned files.
-FROM ubuntu:16.04
+ARG PARENT_IMAGE=ubuntu:16.04
+FROM $PARENT_IMAGE
 
 # http://bugs.python.org/issue19846
 # > At the moment, setting "LANG=C" on a Linux system *fundamentally breaks Python 3*, and that's not OK.
@@ -99,13 +100,6 @@ WORKDIR /root/code/garage
 COPY .pre-commit-config.yaml /root/code/garage
 RUN ["/bin/bash", "-c", "source activate garage && git init && pre-commit"]
 
-# Pull CodeClimate reporter
-RUN curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > /root/code/garage/cc-test-reporter && \
-  chmod +x /root/code/garage/cc-test-reporter
-
 # Add code stub last
 COPY . /root/code/garage
 RUN ["/bin/bash", "-c", "source activate garage && pip install -e /root/code/garage && source deactivate garage"]
-
-# Ready, set, go.
-ENTRYPOINT ["docker/docker-entrypoint.sh"]
--- a/docker/Dockerfile.headless
+++ b/docker/Dockerfile.headless
@@ -1,0 +1,5 @@
+ARG PARENT_IMAGE=rlworkgroup/garage-base
+FROM $PARENT_IMAGE
+
+# Ready, set, go.
+ENTRYPOINT ["docker/entrypoint-headless.sh"]
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,0 +1,111 @@
+# Using Docker to run garage
+
+Currently there are two types of garage images:
+  - headless: garage without environment visualization.
+  - nvidia: garage with environment visualization using an NVIDIA graphics
+    card.
+
+## Headless image
+
+If you already have a copy of garage, proceed to subsection ["Build and run the
+image"], otherwise, keep reading.
+
+To run an example launcher in the container, execute:
+```
+docker run -it --rm rlworkgroup/garage-headless python examples/tf/trpo_cartpole.py
+```
+
+To run environments using MuJoCo, pass the contents of the MuJoCo key in a
+variable named MJKEY in the same docker-run command using `cat`. For example,
+if your key is at `~/.mujoco/mjkey.txt`, execute:
+```
+docker run \
+  -it \
+  --rm \
+  -e MJKEY="$(cat ~/.mujoco/mjkey.txt)" \
+  rlworkgroup/garage-headless python examples/theano/trpo_swimmer.py
+```
+
+To save the experiment data generated in the container, you need to specify a
+path where the files will be saved inside your host computer with the argument
+`-v` in the docker-run command. For example, if the path you want to use is
+at `/home/tmp/data`, execute:
+```
+docker run \
+  -it \
+  --rm \
+  -v /home/tmp/data:/root/code/garage/data \
+  rlworkgroup/garage-headless python examples/tf/trpo_cartpole.py
+```
+This binds a volume between your host path and the path in garage at
+`/root/code/garage/data`.
+
+### Build and run the image
+
+To build the headless image, first clone this repository and move to the root
+folder of your local repository. If you have your MuJoCo key at
+`~/.mujoco/mjkey.txt`, simply execute:
+```
+make build-headless
+```
+
+Alternatively, you can specify the path to your MuJoCo key with the variable
+`MJKEY_PATH`:
+```
+make build-headless MJKEY_PATH="..."
+```
+
+To build and run the container, execute;
+```
+make run-headless RUN_CMD="python examples/tf/trpo_cartpole.py"
+```
+Where RUN_CMD specifies the executable to run in the container
+
+The previous command adds a volume from the data folder inside your cloned
+garage repository to the data folder in the garage container, so any experiment
+results ran in the container will be saved in the data folder inside your
+cloned repository. The data is saved in a folder with the name of the container
+that generated the data, which by default is the name of the image type the
+container is based on with the date and time the container was launched.
+
+If you want to specify another name for the container, do so with the variable
+`CONTAINER_NAME`:
+```
+make run-headless RUN_CMD="..." CONTAINER_NAME="my_container"
+```
+
+#### Prerequisites
+
+Be aware of the following prerequisites to build the image.
+
+- Install [Docker CE](https://docs.docker.com/install/linux/docker-ce/ubuntu/#install-docker-ce)
+- Install [Docker Compose](https://docs.docker.com/compose/install/#install-compose)
+
+Tested on Ubuntu 16.04.
+
+## nvidia image
+
+The same commands for the headless image mentioned above apply for the nvidia
+image, except that the image name is defined by `rlworkgroup/garage-nvidia`.
+For example, to execute a launcher file:
+```
+docker run -it --rm rlworkgroup/garage-nvidia python examples/tf/trpo_cartpole.py
+```
+
+### Build and run the image
+
+The same rules for the headless image apply here, except that the target names
+are the following:
+```
+make build-nvidia
+make run-nvidia
+```
+
+#### Prerequisites
+
+Additional to the prerequisites for the headless image, make sure to have:
+- Install the latest NVIDIA driver, tested
+  on [nvidia-390](https://tecadmin.net/install-latest-nvidia-drivers-ubuntu/)
+- [Install nvidia-docker2](https://github.com/NVIDIA/nvidia-docker#ubuntu-140416041804-debian-jessiestretch)
+
+Tested on Ubuntu 16.04.
--- a/docker/docker-compose-headless.yml
+++ b/docker/docker-compose-headless.yml
@@ -1,0 +1,16 @@
+version: '2.2'
+services:
+  garage-base:
+    build:
+      cache_from:
+        - ${TAG}:latest
+      context: ../
+      dockerfile: docker/Dockerfile
+    image: rlworkgroup/garage-base
+  garage-headless:
+    build:
+      context: ../
+      dockerfile: docker/Dockerfile.headless
+      args:
+        - PARENT_IMAGE=rlworkgroup/garage-base
+    image: ${TAG}
--- a/docker/Dockerfile.nvidia
+++ b/docker/Dockerfile.nvidia
@@ -1,0 +1,5 @@
+ARG PARENT_IMAGE=rlworkgroup/garage-base
+FROM $PARENT_IMAGE
+
+# Ready, set, go.
+ENTRYPOINT ["docker/entrypoint-nvidia.sh"]
--- a/docker/docker-compose-nvidia.yml
+++ b/docker/docker-compose-nvidia.yml
@@ -1,0 +1,18 @@
+version: '2.2'
+services:
+  garage-base-nvidia:
+    build:
+      cache_from:
+        - rlworkgroup/garage-nvidia:latest
+      context: ../
+      dockerfile: docker/Dockerfile
+      args:
+        - PARENT_IMAGE=nvidia/opengl:1.0-glvnd-runtime-ubuntu16.04
+    image: rlworkgroup/garage-nvidia-base
+  garage-nvidia:
+    build:
+      context: ../
+      dockerfile: docker/Dockerfile.nvidia
+      args:
+        - PARENT_IMAGE=rlworkgroup/garage-nvidia-base
+    image: rlworkgroup/garage-nvidia
--- a/docker/entrypoint-headless.sh
+++ b/docker/entrypoint-headless.sh
@@ -6,7 +6,10 @@ set -e
 echo "${MJKEY}" > /root/.mujoco/mjkey.txt
 
 # Setup dummy X server display
-display_num=0
+# Socket for display :0 may already be in use if the container is connected
+# to the network of the host, and other low-numbered socket could also be in
+# use, that's why we use 100.
+display_num=100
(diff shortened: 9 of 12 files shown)


docker/README.md
```diff
@@ -0,0 +1,81 @@
+# Garage Docker
```
gonzaiva: I forgot to add a pre-requisites section, mainly for docker-ce and docker-compose.

.gitignore
```diff
@@ -32,6 +32,7 @@ iterate.dat
 src/
 .settings
 .pods
+docker/mjkey.txt
```
ryanjulian: don't copy this into the repo. retrieve it from wherever it lives on the filesystem and then pass it as an environment variable. see Dockerfile.ci for an example.

Makefile
```diff
@@ -0,0 +1,34 @@
+.PHONY: build-garage-release run-garage-release
+
+GARAGE_RELEASE="v2018.10.0"
+
+build-garage-release: docker/Dockerfile.release docker/mjkey.txt
```
ryanjulian: nah this makes no sense. there is only one set of canonical Dockerfiles with a single set of names.

the ones in the release branch are compliant with the release. the ones on the master branch are compliant with master.

Dockerfiles are code, just like any other Python file. we don't have -release and -local versions of logger.py, do we?

Makefile
```diff
@@ -0,0 +1,34 @@
...
+	docker build \
+		-f docker/Dockerfile.release \
+		--build-arg RELEASE="${GARAGE_RELEASE}" \
+		-t rlworkgroup/garage:release .
```
ryanjulian: tags (garage:release) are for specifying versions, not variants. if you want an image variant, you should name it with a dash.

for instance, the ci version of this Dockerfile is garage-ci

Makefile
```diff
@@ -0,0 +1,34 @@
...
+
+build-garage-local: docker/Dockerfile.local
+	docker build \
+		-f docker/Dockerfile.local \
+		-t rlworkgroup/garage:local .
```
ryanjulian: tags are for versions, not variants.

docker/README.md
ryanjulian: I think this is more appropriately called "Using Docker to run garage"

docker/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+Currently there's three versions of garage containers:
+  - ci: used for the continuous integration of garage in TravisCI and not
+    intended for final users.
+  - release: it's a published image based on the latest official release of
```
ryanjulian: this should just live in the release branch

docker/README.md
```diff
@@ -0,0 +1,81 @@
...
+    garage.
+  - local: it's an image built on top of the release with a copy of the files
```
ryanjulian: this is not called .local, it's just called the Dockerfile. 

whereas we put special Dockerfiles in docker/, the canonical Dockerfile usually goes in the repository root. We might move all Dockerfiles to the repo root--not sure yet. That's out of scope for this PR.

docker/README.md
```diff
@@ -0,0 +1,81 @@
...
+```
+docker run --rm rlworkgroup/garage:release RUN_CMD="python examples/tf/trpo_cartpole.py"
+```
+This image does not come with a MuJoCo key, so it's not possible to run MuJuCo
+environments with it. To enable MuJoCo, you need to build a local image and
+set a valid key in the image.
+
+If you want to build the release image, follow this steps:
+  1. Clone this repository
+  2. Copy a valid MuJoCo key in `docker/mjkey.txt` inside the cloned repository
```
ryanjulian: No. See how we do it in Dockerfile.ci. We do that for a reason.

docker/README.md
```diff
@@ -0,0 +1,81 @@
...
+```
+
+This previous command adds a volume from the data folder inside your cloned
+garage repository to the garage-release container, so any experiment results
+ran in the container will be saved in the data folder inside your cloned
+repository.
+
+## Local image
+
+The local image is built on top of the release image and it includes the
```
ryanjulian: This doesn't make any sense to me. What if I want to use the latest garage, not the release?

docker/README.md
```diff
@@ -0,0 +1,81 @@
...
+changes in your cloned repository as well as configuring the MuJoCo key
+(optionally).
+
+If you want to build the local image, follow this steps:
+  1. Clone this repository
+  2. Copy a valid MuJoCo key in `docker/mjkey.txt` inside the cloned repository
```
ryanjulian: No. Use environment variables.

docker/docker-entrypoint-release.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+# Dockerfile entrypoint
+set -e
+
+source ~/.bashrc
+
+# Activate conda environment
+source activate garage
```
ryanjulian: How does this differ from the CI entrypoint?

docker/Dockerfile.release
```diff
@@ -0,0 +1,108 @@
+# NOTICE: To keep consistency across this docker file, scripts/setup_linux.sh
+# and scripts/setup_macos.sh, if there's any changes applied to this file,
+# specially regarding the installation of dependencies, apply those same
+# changes to the mentioned files.
+FROM ubuntu:16.04
```
ryanjulian: This image uses a fake display, meaning I can't ever see the training visualization. Why not have fake display, docker-opengl, and nvidia-docker versions of this?

The canonical one would be docker-opengl, since that will run on any machine interactively.

docker/README.md
```diff
@@ -0,0 +1,81 @@
...
+    in your local repository.
+
+## Release image
+
+The release image is published to be downloaded without the need of having
+a local repository. The command to run in the corresponding container can be
+specified through the argument RUN_CMD. To run an example launcher in the
+container, execute:
+```
+docker run --rm rlworkgroup/garage:release RUN_CMD="python examples/tf/trpo_cartpole.py"
```
ryanjulian: why do i need RUN_CMD? I can just put the command right on the terminal and it will be `$@` in the entrypoint

gonzaiva:
Hi guys, I'm back and I'll be reviewing your comments today.

.gitignore
gonzaiva: Done, now the mjkey is read from an environment variable instead.

Makefile
gonzaiva: Got it! Once this PR is approved, I will back port this commit to the 2018.10 release to build the corresponding version images.

Makefile
gonzaiva: I followed this nomenclature in the new changes

docker/README.md
gonzaiva: Done

docker/README.md
gonzaiva: Once approved, I will back port this PR to the release branch. In the meantime, I have removed the image named rlworkgroup/garage:release.

docker/README.md
gonzaiva: Done

docker/README.md
gonzaiva: Got it! This PR will create the images for latest changes in master, and each release branch will have their corresponding images based on the release number.

docker/README.md
gonzaiva: Done

docker/docker-entrypoint-release.sh
gonzaiva: I removed this file and with some light modifications I'm using the CI entry point.

docker/Dockerfile.release
gonzaiva: I could generate images for the fake display (I named it headless) and nvidia versions. However, I couldn't find/create a proper opengl image that works with ubuntu 16.04. I'm taking a look at which packages the nvidia-opengl uses to copy the opengl parts only, but I can submit this version in a next PR.

docker/README.md
```diff
@@ -0,0 +1,81 @@
+# Garage Docker
+
+Currently there's three versions of garage containers:
```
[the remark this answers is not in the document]
gonzaiva: Done

docker/README.md
gonzaiva: Fixed!

Makefile
ryanjulian: I honestly don't think that's necessary. 2018.10 release only guarantees bug fixes, not new features.

.travis.yml
```diff
@@ -75,6 +76,10 @@ after_success:
       echo "${DOCKER_API_KEY}" | docker login -u "${DOCKER_USERNAME}" --password-stdin
       docker tag "${tag}" rlworkgroup/garage-ci:latest
       docker push rlworkgroup/garage-ci
+      make build-garage-headless
+      docker push rlworkgroup/garage-headless
+      make build-garage-nvidia
```
ryanjulian: No no no. Do not build the nvidia dockerfile in the CI.

This won't work because the CI has no graphics card.

Also the purpose of this part of the CI script is not to maintain the Docker hub. It's to keep a cache of the CI docker image. Please remove this.

docker/Dockerfile
```diff
@@ -105,7 +106,14 @@ RUN curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest
 
 # Add code stub last
 COPY . /root/code/garage
+# Prevent garage from exiting on config
+RUN cp /root/code/garage/garage/config_personal_template.py \
+  /root/code/garage/garage/config_personal.py
 RUN ["/bin/bash", "-c", "source activate garage && pip install -e /root/code/garage && source deactivate garage"]
 
+# Specifies which docker image flavor we're dealing with
+ARG IMAGE_TYPE
```
ryanjulian: this is strange. why does the Dockerfile need to know this?

docker/Dockerfile
```diff
@@ -105,7 +106,14 @@ RUN curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest
 
 # Add code stub last
 COPY . /root/code/garage
+# Prevent garage from exiting on config
+RUN cp /root/code/garage/garage/config_personal_template.py \
```
ryanjulian: Why is this necessary here but not in the headless version?

Makefile
```diff
@@ -0,0 +1,43 @@
...
+	xhost +local:docker
+	@ docker run \
+		-it \
+		--rm \
+		--runtime=nvidia \
+		-v /tmp/.X11-unix:/tmp/.X11-unix \
+		-v $(shell pwd)/data:/root/code/garage/data \
+		-e DISPLAY=$(DISPLAY) \
+		-e QT_X11_NO_MITSHM=1 \
+		-e MJKEY="$(MJKEY)" \
```
ryanjulian: suggestion:
if the environment variable MJKEY is empty, attempt to read it from the disk (~/.mujoco/mjkey.txt) into the environment variable and produce a message saying you automatically detected the key at that location.

docker/Dockerfile
```diff
@@ -84,7 +85,7 @@ COPY environment.yml /root/code/garage/environment.yml
 # MAKE SURE TO DELETE MJKEY.TXT SO THAT WE DON'T PUBLISH THE KEY
 # NOTE: You MUST create and delete the mjkey.txt within the same RUN command
 ARG MJKEY
-RUN echo "${MJKEY}" > /root/.mujoco/mjkey.txt && \
+RUN echo -e "${MJKEY}" > /root/.mujoco/mjkey.txt && \
```
ryanjulian: why is -e necessary?

docker/README.md
```diff
@@ -0,0 +1,79 @@
...
+docker run -it --rm rlworkgroup/garage-headless python examples/tf/trpo_cartpole.py
+```
+
+This image does not come with a MuJoCo key, so if you want to run environments
+using MuJoCo, the environment variable `MJKEY` has to be set with the
+corresponding content of the key file in your file system:
+```
+# New line characters in MuJoCo key files (they're in DOS format) are
+# replaced by '\n' to pass the file as a command argument.
+export MJKEY="$(sed ':a;N;$!ba;s/\r\r//g;s/\n/\\n/g' <your_key_path>)"
```
ryanjulian: Do not instruct users to export their MJKEY. This is a security risk.

It's extremely easy to include your MJKEY as an argument

MJKEY="$(cat ~/.mujoco/mjkey.txt)"

docker/README.md
```diff
@@ -0,0 +1,79 @@
+# Using Docker to run garage
+
+Currently there are three types of garage images:
+  - ci: used for the continuous integration of garage in TravisCI and not
```
ryanjulian: you appear to only have CI and headless images? also it's not clear how CI differs from headless

docker/README.md
```diff
@@ -0,0 +1,79 @@
...
+    intended for final users.
```
ryanjulian: what is a final user?

docker/README.md
```diff
@@ -0,0 +1,79 @@
...
+  - headless: garage without environment visualization.
+  - NVIDIA: garage with environment visualization using an NVIDIA graphics
```
ryanjulian: lower-case nvidia

docker/README.md
```diff
@@ -0,0 +1,79 @@
...
+    card.
+
```
ryanjulian: what about users who want to run interactively but without an nvidia card? see docker-opengl

docker/README.md
```diff
@@ -0,0 +1,79 @@
+# Using Docker to run garage
+
+Currently there are three types of garage images:
```
ryanjulian: it's not clear how headless and CI differ

docker/README.md
```diff
@@ -0,0 +1,79 @@
...
+  -it \
+  --rm \
+  -e MJKEY="${MJKEY}" \
+  rlworkgroup/garage-headless python examples/theano/trpo_swimmer.py
+```
+
+For every release of garage there's a version of this container, which you can
+specify through the release number. For example:
+```
+docker run --rm rlworkgroup/garage-headless:2018.10 python examples/tf/trpo_cartpole.py
```
ryanjulian: nah

docker/docker-entrypoint.sh
```diff
@@ -3,37 +3,41 @@
 set -e
 
 # Get MuJoCo key from the environment
-echo "${MJKEY}" > /root/.mujoco/mjkey.txt
+echo -e "${MJKEY}" > /root/.mujoco/mjkey.txt
```
ryanjulian: why -e??

docker/docker-entrypoint.sh
```diff
@@ -3,37 +3,41 @@
...
 # See: https://github.com/openai/mujoco-py/pull/145#issuecomment-356938564
-export LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libGLEW.so:/usr/lib/x86_64-linux-gnu/mesa/libGL.so.1
+export LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libGLEW.so
 
 export TF_CPP_MIN_LOG_LEVEL=3      # shut TensorFlow up
-export DISABLE_MUJOCO_RENDERING=1  # silence glfw
 export LD_LIBRARY_PATH="${LD_LIBRARY_PATH}:${HOME}/.mujoco/mjpro150/bin"
 
-exec "$@"
+source /root/.bashrc && exec "$@"
```
ryanjulian: do not source a bashrc in a docker environment.

docker/docker-entrypoint.sh
```diff
@@ -3,37 +3,41 @@
...
+  export DISABLE_MUJOCO_RENDERING=1  # silence glfw
 fi
 
 # Activate conda environment
 source activate garage
 
 # Fixes Segmentation Fault
 # See: https://github.com/openai/mujoco-py/pull/145#issuecomment-356938564
-export LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libGLEW.so:/usr/lib/x86_64-linux-gnu/mesa/libGL.so.1
+export LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libGLEW.so
```
ryanjulian: does this work on nvidia? you would need to preload the nvidia version.

Makefile
```diff
@@ -0,0 +1,43 @@
+.PHONY: build-garage-headless build-garage-nvidia \
```
ryanjulian: i'm not sure this setup exactly works for interactive use cases. what will happen to all my data files if i'm using these containers to run an experiment?

by default, they will be destroyed when the container stops.

.travis.yml
gonzaiva: Removed

docker/Dockerfile
gonzaiva: I realized that the docker-entrypoint for headless and nvidia shares the same commands except for creating the socket of the fake display for the headless type.
In order to distinguish between both I added this variable and I use it [here](https://github.com/rlworkgroup/garage/blob/b97856de130d0b5eebdfeb400796dc1eee458a78/docker/docker-entrypoint.sh#L8). My effort was to keep one canonical entry point for all image types, but now I see that the idea is that each type should have its own Dockerfile and entrypoint.

docker/Dockerfile
gonzaiva: It is necessary in the headless (CI) version as well, but the config files is copied in [Travis](https://github.com/rlworkgroup/garage/blob/77714c38d5b575a5cfd6d1e42f0a045eebbe3484/.travis.yml#L28), and this won't work for local builds.
I forgot to remove the line from the .travis.yml though, so I will do that.

Makefile
gonzaiva: Done

docker/Dockerfile
gonzaiva: Since I'm passing the MJKEY as a build argument in docker, I have to pass it as a single line string, that is, replace the new lines and returns as \n and \r, otherwise the make recipes are ruined by expanding these characters.
By using "echo -e", the backslash characters are expanded to reconstruct mjkey.txt.

docker/README.md
gonzaiva: I agree that is safer to pass the MJKEY file as a variable to the makefile, but the cat command is not enough because once MJKEY is expanded in the recipe, its first return/newline will indicate the end of the build command, and the rest of the build command will remain unprocessed, generating an error.
For example:
```
(garage) aigonzal@gonzaiva:~/ivanWorkspace/garage$ make build-garage-headless MJKEY="$(cat ~/.mujoco/mjkey.txt)"
docker build \
	-f docker/Dockerfile \
	--build-arg MJKEY="MuJoCo Pro Individual license activation key, number ????, type ?.
/bin/sh: 3: Syntax error: Unterminated quoted string
Makefile:14: recipe for target 'build-garage-headless' failed
make: *** [build-garage-headless] Error 2
```
That's why I'm using sed.

docker/README.md
gonzaiva: A difference between the headless and CI images could be CodeClimate, since only the latter uses it, but I don't see any problem in having CodeClimate for headless. I will remove the references to the CI image.

docker/README.md
gonzaiva: The user that only wants to test/deploy experiments with garage, sorry if the term is misleading. I will remove this section of text anyway.

docker/README.md
gonzaiva: I've checked docker-opengl, but it doesn't work when installing some apt dependencies. I'm still working on the opengl flavor.

docker/README.md
gonzaiva: I will replace the references from CI to headless.

docker/README.md
gonzaiva: ?

docker/docker-entrypoint.sh
gonzaiva: Since I'm passing the MJKEY as a build argument in docker, I have to pass it as a single line string, that is, replace the new lines and returns as \n and \r, otherwise the make recipes are ruined by expanding these characters.
By using "echo -e", the backslash characters are expanded to reconstruct mjkey.txt.

docker/docker-entrypoint.sh
gonzaiva: It does work on nvidia.

docker/docker-entrypoint.sh
gonzaiva: I'm sourcing bashrc because python packages used in the ROS workspace and the intera_interface are loaded in that way. Otherwise, if we try to run a python script that imports `intera_interface` and/or `rospy`, it will not find them.
A way to avoid this could be an environment variable with a list of all the setup scripts, and then source each of the scripts in the entrypoint, but that list has to be carried on from one docker build to another in dockercompose.
Any ideas to solve this problem are really appreciated :smiley_cat:

Makefile
gonzaiva: That's not completely true, I bind the data folder in the container with the data folder in the garage workspace using the argument:
```
-v $(shell pwd)/data:/root/code/garage/data
```
However, I accept this won't work when the container runs anywhere else in the file system. I will add a variable to pass the path to bind to the data folder in the container, having the garage/data folder as default, and if there's no folder in the path, then only in that case the data files will be lost.

gonzaiva:
Variable expansion is not working when calling make in Travis as in my computer, and that's why the build fails. It may be related to using make 3.9.2 instead of 4.1 since the former comes installed with Ubuntu Trust, but I'm investigating other reasons.

docker/Dockerfile
ryanjulian: what if i am an interactive user and i actually customized my config_personal.py though?

docker/Dockerfile.headless
```diff
@@ -0,0 +1,9 @@
+ARG PARENT_IMAGE=rlworkgroup/garage-base
+FROM $PARENT_IMAGE
+
+# Pull CodeClimate reporter
```
ryanjulian: perhaps you can do this in .travis.yml to avoid splitting the entrypoints? if the file is in code/garage it should get pulled into the image.

docker/README.md
ryanjulian: we aren't adding containers for the previous release, so this tag won't exist yet.

docker/docker-compose-headless.yml
```diff
@@ -0,0 +1,14 @@
+version: '2.2'
```
ryanjulian: is docker-compose really necessary here? we are only using one container at a time. docker-compose is for launching groups of containers which work together.

ryanjulian: it's okay if it makes this more maintainable, but makefile+docker compose is starting to be a lot of layers.

docker/docker-entrypoint-headless.sh
```diff
@@ -6,34 +6,36 @@ set -e
 echo "${MJKEY}" > /root/.mujoco/mjkey.txt
 
 # Setup dummy X server display
-display_num=0
+# Socket for display :0 may already be in use if the container is connected
+# to the network of the host, that's why we use 1.
```
ryanjulian: in that case you should start with a very high display number (e.g. 100) to avoid collisions with similar reasons.

docker/docker-entrypoint.sh
ryanjulian: that problem should be dealt with in garage-ros, not here. it's bad form to be sourcing bashrc inside a Docker container.

I realize ROS is full of bad design decisions, but we shouldn't let them percolate into these unrelated containers.

for instance, your ROS container can just be FROM this image, but with a different entrypoint.

docker/docker-entrypoint-nvidia.sh
```diff
@@ -0,0 +1,18 @@
...
+source activate garage
+
+# Fixes Segmentation Fault
+# See: https://github.com/openai/mujoco-py/pull/145#issuecomment-356938564
+export LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libGLEW.so
+
+export TF_CPP_MIN_LOG_LEVEL=3      # shut TensorFlow up
+export LD_LIBRARY_PATH="${LD_LIBRARY_PATH}:${HOME}/.mujoco/mjpro150/bin"
+
+source /root/.bashrc && exec "$@"
```
ryanjulian: please don't source bashrc inside docker.

Makefile
```diff
@@ -0,0 +1,69 @@
...
+	run-garage-headless run-garage-nvidia
+
+# Path in host where the experiment data obtained in the container is stored
```
ryanjulian: isn't this set in config.py/config_personal.py?

there needs to be a single source of truth here. you might need to ask Python where to put it.

docker/README.md
ryanjulian: okay. can you explain the difference between headless and CI then? i am still confused

Makefile
gonzaiva: Done

ryanjulian: that solution (attempting to glob the conda install) is really hacky and makes me nervous. you have no guarantee that conda is even inside $HOME. i have machines where it lives in /opt, for instance. furthermore, you have no guarantee the host is running bash.

this is a good indication that at least one of the following things are true:
* Makefile is not a good way for us to manage the Dockerfiles (current config assumes Docker is launched from Python, so doesn't have this problem)
* We should not configure the user's environment in config.py

I think the least worst solution here is to pick a default place to map data/ from the docker container into (perhaps `$(pwd)/data/dockerfile_name` and stick to it. Allow users to override the default with an argument  to the Makefile

It also means we should remove the Docker-related entries from config/config_personal.py

It's clear here that how we do configuration in garage needs to be rewritten, but that can be left for another day.

Makefile
ryanjulian: in retrospect, something like this is actually the right thing to do

docker/Dockerfile
ryanjulian: i think this logic needs to be moved to the Makefile

Makefile
```diff
@@ -0,0 +1,80 @@
...
+		--build-arg MJKEY="$$(cat ${MJKEY_PATH})"
+
+build-garage-nvidia: docker/Dockerfile check-mjkey
+	docker-compose \
+		-f docker/docker-compose-nvidia.yml \
+		build \
+		--build-arg MJKEY="$$(cat ${MJKEY_PATH})"
+
+run-garage-ci: TAG ?= rlworkgroup/garage-ci
+run-garage-ci: docker/Dockerfile check-mjkey
```
ryanjulian: docker/Dockerfile-headless?

Makefile
gonzaiva: For headless and nvidia I'm doing the following binding:
```
-v $(DATA_PATH)/$(CONTAINER_NAME):/root/code/garage/data
```
Where `DATA_PATH` is equal to `$(shell pwd)/data`
By default, `CONTAINER_NAME` for headless is equal to `garage-headless-$(DATE_TIME)`, and `garage-nvidia-$(DATE_TIME)` for nvidia. The DATE_TIME is defined by `$(shell date "+%Y_%m_%d_%H_%M_%S")`.
`CONTAINER_NAME` can be set by the user by passing it as an argument when calling make.

I created the target copy_config_personal to copy the config_personal.py file in case it doesn't exist.

.travis.yml
```diff
@@ -24,49 +27,28 @@ before_install:
       "max-concurrent-uploads": 50
     }' | sudo tee /etc/docker/daemon.json
   - sudo service docker restart
-  # prevent garage from exiting on config
-  - cp garage/config_personal_template.py garage/config_personal.py
+  # Reinstall docker compose with the specified version
```
ryanjulian: whoah this is strange. why can't we just pull the right version of docker from apt?

Makefile
ryanjulian: i think the path should be `$(shell pwd)/data/$(CONTAINER_NAME)`. the default experiment runner config already adds timecodes.

docker/Dockerfile
```diff
@@ -99,13 +100,6 @@ WORKDIR /root/code/garage
 COPY .pre-commit-config.yaml /root/code/garage
 RUN ["/bin/bash", "-c", "source activate garage && git init && pre-commit"]
 
-# Pull CodeClimate reporter
-RUN curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > /root/code/garage/cc-test-reporter && \
-  chmod +x /root/code/garage/cc-test-reporter
-
 # Add code stub last
 COPY . /root/code/garage
 RUN ["/bin/bash", "-c", "source activate garage && pip install -e /root/code/garage && source deactivate garage"]
```
ryanjulian: the `pip install -e` step should be done in the Dockerfile, not the entrypoint

Makefile
```diff
@@ -0,0 +1,87 @@
...
+# Date/time is appended to the container names
+DATE_TIME ?= $(shell date "+%Y_%m_%d_%H_%M_%S")
+# Prevent garage from exiting on config by making sure the personal config file
+# is already created
+CONFIG_PERSONAL := garage/config_personal.py
+
+build-garage-ci: TAG ?= rlworkgroup/garage-ci
+build-garage-ci: docker/docker-compose-headless.yml copy_config_personal
+	TAG=${TAG} \
+	docker-compose \
```
ryanjulian: does this properly use `--cache-from`? otherwise, builds in the CI will take a very long time.

Makefile
```diff
@@ -0,0 +1,87 @@
+.PHONY: build-garage-ci build-garage-headless build-garage-nvidia \
```
ryanjulian: perhaps you can remove `-garage-` from all these directives, since it's obvious they are garage

`run-nvidia`, etc.

docker/Dockerfile
gonzaiva: But this is the Dockerfile :confused:

Makefile
```diff
@@ -0,0 +1,85 @@
...
+	run-garage-ci run-garage-headless run-garage-nvidia
+
+SHELL = bash
```
ryanjulian: i don't think you have a guarantee bash is installed. is there a reason you need bash? i've basically never seen this in a Makefile before.

Makefile
```diff
@@ -0,0 +1,85 @@
...
+
+check-mjkey:
+ifeq (0, $(shell [ ! -f $(MJKEY_PATH) ]; echo $$? ))
+	$(error The MJKEY was not set: make sure to pass a valid MJKEY_PATH \
+		or put your key at ~/.mujoco/mjkey.txt)
+endif
+
+copy_config_personal:
+ifeq (0, $(shell [ ! -f $(CONFIG_PERSONAL) ]; echo $$? ))
+	cp garage/config_personal_template.py garage/config_personal.py
```
ryanjulian: you should warn a user when you do this

.travis.yml
ryanjulian: okay i see this in the Travis documentation.

can you include this link for explanation.

https://docs.travis-ci.com/user/docker/#using-docker-compose

Makefile
gonzaiva: bash is required to obtain the CodeCov arguments and pass them to the CI container:
```
run-ci:                                                                         
         docker run \                                                            
                 "$$(bash <(curl -s https://codecov.io/env))" \ 
                 ...
```
If the SHELL is not set as bash, even if the bash command is explicitly called in the target, the bourne shell is used. I can set the shell to be bash only in the run-ci target, for example:
```
run-ci: SHELL := /usr/bin/env bash
run-ci:                                                                         
         docker run \                                                            
                 "$$(bash <(curl -s https://codecov.io/env))" \ 
                 ...
```

.travis.yml
gonzaiva: Done

Makefile
gonzaiva: Done

docker/Dockerfile.headless
```diff
@@ -0,0 +1,5 @@
+ARG PARENT_IMAGE=rlworkgroup/garage-base
+FROM $PARENT_IMAGE
+
+# Ready, set, go.
+ENTRYPOINT ["docker/docker-entrypoint-headless.sh"]
```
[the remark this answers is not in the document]
gonzaiva: Done

Makefile
gonzaiva: DOne

Makefile
gonzaiva: I did that and worked, so let me know if this is okay with you, otherwise I can explore other options without using bash such as using a temporary file to save the script.

Makefile
gonzaiva: I have added the cache-from argument to the docker-compose files:
```

Makefile
ryanjulian: i like this solution. i think it might be more clear if you can put the results of the curl into a descriptive variable first (e.g. CI_ENV).

otherwise it would be hard for someone to figure out why we have a random curl call in our Docker run command.

gonzaiva: Done

PR #453, opened 2018-12-22
Fix AttributeError in updating 'mujoco_py.cymj.PyMjData'

bohan-zhang:
Fix error: `AttributeError: attribute 'qpos' of 'mujoco_py.cymj.PyMjData' objects is not writable` when running `reset_mujoco()` on `MujocoEnv`. See https://github.com/rlworkgroup/garage/issues/452.

More specifically:
```
  File "garage/scripts/run_experiment.py", line 242, in <module>
    run_experiment(sys.argv)
  File "garage/scripts/run_experiment.py", line 185, in run_experiment
    method_call(variant_data)
  File "./examples/pusher_pretrain.py", line 81, in run_experiment
    env = normalize(PusherEnv(goal=variant.get('goal')))
  File "softqlearning/softqlearning/environments/pusher.py", line 37, in __init__
    super(PusherEnv, self).__init__(file_path=self.FILE_PATH)
  File "garage/garage/envs/mujoco/mujoco_env.py", line 85, in __init__
    self.reset()
  File "softqlearning/softqlearning/environments/pusher.py", line 130, in reset
    super(PusherEnv, self).reset(full_state)
  File "garage/garage/envs/mujoco/mujoco_env.py", line 134, in reset
    self.reset_mujoco(init_state)
  File "garage/garage/envs/mujoco/mujoco_env.py", line 122, in reset_mujoco
    self.sim.data.qpos = datum
AttributeError: attribute 'qpos' of 'mujoco_py.cymj.PyMjData' objects is not writable
```

### DIFF
--- a/garage/envs/mujoco/mujoco_env.py
+++ b/garage/envs/mujoco/mujoco_env.py
@@ -115,35 +115,17 @@ class MujocoEnv(gym.Env):
             self.sim.data.qacc[:] = self.init_qacc
             self.sim.data.ctrl[:] = self.init_ctrl
         else:
-            start = 0
+            datums = [
+                self.sim.data.qpos,
+                self.sim.data.qvel,
+                self.sim.data.qacc,
+                self.sim.data.ctrl,
+            ]
+            dims = [d.shape[0] for d in datums]
+            offsets = [0] + list(np.cumsum(dims)[:-1])
 
-            # update qpos
-            datum = self.sim.data.qpos
-            datum_dim = datum.shape[0]
-            datum = init_state[start:start + datum_dim]
-            self.sim.data.qpos[:] = datum
-            start += datum_dim
-
-            # update qvel
-            datum = self.sim.data.qvel
-            datum_dim = datum.shape[0]
-            datum = init_state[start:start + datum_dim]
-            self.sim.data.qvel[:] = datum
-            start += datum_dim
-
-            # update qacc
-            datum = self.sim.data.qacc
-            datum_dim = datum.shape[0]
-            datum = init_state[start:start + datum_dim]
-            self.sim.data.qacc[:] = datum
-            start += datum_dim
-
-            # update ctrl
-            datum = self.sim.data.ctrl
-            datum_dim = datum.shape[0]
-            datum = init_state[start:start + datum_dim]
-            self.sim.data.ctrl[:] = datum
-            start += datum_dim
+            for datum, dim, offset in zip(datums, dims, offsets):
+                datum[:] = init_state[offset:offset + dim]
 
     @overrides
     def reset(self, init_state=None):
--- a/garage/envs/mujoco/mujoco_env.py
+++ b/garage/envs/mujoco/mujoco_env.py
@@ -116,12 +116,34 @@ class MujocoEnv(gym.Env):
             self.sim.data.ctrl[:] = self.init_ctrl
         else:
             start = 0
-            for datum_name in ["qpos", "qvel", "qacc", "ctrl"]:
-                datum = getattr(self.sim.data, datum_name)
-                datum_dim = datum.shape[0]
-                datum = init_state[start:start + datum_dim]
-                exec('self.sim.data.%s[:] = datum' % datum_name)
-                start += datum_dim
+
+            # update qpos
+            datum = self.sim.data.qpos
+            datum_dim = datum.shape[0]
+            datum = init_state[start:start + datum_dim]
+            self.sim.data.qpos[:] = datum
+            start += datum_dim
+
+            # update qvel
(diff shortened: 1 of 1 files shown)


garage/envs/mujoco/mujoco_env.py
```diff
@@ -120,7 +120,7 @@ def reset_mujoco(self, init_state=None):
                 datum = getattr(self.sim.data, datum_name)
                 datum_dim = datum.shape[0]
                 datum = init_state[start:start + datum_dim]
-                setattr(self.sim.data, datum_name, datum)
+                exec('self.sim.data.%s[:] = datum' % datum_name)
```
ryanjulian: ### The problem
This bug is caused by mujoco_py making the references to the underlying numpy arrays in `self.sim.data` read-only. You are free to write into the underlying array (`self.sim.data.ctrl[:] = self.init_ctrl`), but you may not change the array to which the attr points (`self.sim.data.ctrl = self.init_ctrl`)

### How to fix it
I believe your line is equivalent to
```python
getattr(self.sim.data, datum_name).__setitem__(slice(None), datum)
# same as
self.sim.data.datum_name[:] = datum
# also the same as
exec('self.sim.data.%s[:] = datum' % datum_name)
```
which is ugly, but at least it doesn't use `exec()`

### I would prefer if you just unrolled the `for` loop and wrote everything out. 
The original code is some obscure Python meta-programming, and it's difficult to read. There are exactly 4 well-known elements to cover, and nobody would ever be confused about what's going on if it was written out (which is not the case at all for the original version). 

### Comments
* please don't use `exec()` in almost all cases. meta-programming like that is very hard to test and this case is not exceptional-enough to warrant it.

bohan-zhang: @ryanjulian Thank you so much for the comments and I learnt a lot from you! I'll patch the fix right now!

garage/envs/mujoco/mujoco_env.py
```diff
@@ -116,12 +116,34 @@ def reset_mujoco(self, init_state=None):
...
         else:
             start = 0
-            for datum_name in ["qpos", "qvel", "qacc", "ctrl"]:
-                datum = getattr(self.sim.data, datum_name)
-                datum_dim = datum.shape[0]
-                datum = init_state[start:start + datum_dim]
-                setattr(self.sim.data, datum_name, datum)
-                start += datum_dim
+
+            # update qpos
```
ryanjulian: * Please use different variables for each case instead of reusing `datum` and `datum_dim`

### Justification
Code should be tolerant to refactoring. If I rearrange some of this code and the variables share names, I might end up with a different result

garage/envs/mujoco/mujoco_env.py
```diff
@@ -116,12 +116,34 @@ def reset_mujoco(self, init_state=None):
...
+            datum = self.sim.data.qpos
+            datum_dim = datum.shape[0]
+            datum = init_state[start:start + datum_dim]
+            self.sim.data.qpos[:] = datum
+            start += datum_dim
```
ryanjulian: I think this code would be most impervious to refactoring and shorter if you just pre-calculated the offsets (`start`). Right now, if these lines get rearranged, the final array will end up with the wrong ordering (very difficult to debug).

Something like:
```python
datums = [
    self.sim.data.qpos,
    self.sim.data.qvel,
    self.sim.data.qacc,
    self.sim.data.ctrl,
]
dims = [d.shape[0] for d in datums]
offsets = [0] + list(np.cumsum(dims)[:-1])

for datum, dim, offset in zip(datums, dims, offsets):
    datum[:] = init_state[offset:offset + dim] 
```

that's it! no metaprogramming!

garage/envs/mujoco/mujoco_env.py
bohan-zhang: That's an awesome point Ryan! I'll patch that fix right now!

garage/envs/mujoco/mujoco_env.py
bohan-zhang: Thank you so much for your pointers, @ryanjulian and I tested your solution out! it's working 👍

ryanjulian:
Hey Bohan,

Can you please rebase this on `master`? Once you do that, I'll send it to the CI to be checked and submitted.

bohan-zhang:
@ryanjulian I just rebased this on master!

ryanjulian:
Okay thanks. I'm traveling right now but should be able to merge this in about a week.

Sorry for the long delay.

ryanjulian:
Hey Bohan,

These files have been moved to https://github.com/rlworkgroup/metaworlds, so I'm going to close this PR.

Feel free to reopen it on the new repo!

PR #457, opened 2019-01-03
Add necessary classes for DQN

ahtsan:


garage/tf/policies/epsilon_greedy_policy.py
```diff
@@ -0,0 +1,118 @@
...
+    max_epsilon to min_epsilon within decay_ratio * total_step.
+
+    At state s, with probability
+    1 − ϵ: select action = argmax Q(s, a)
+    ϵ    : select a random action
+
+    Args:
+        env_spec: environment specification
+        total_step: Total steps in the training.
+        max_epsilon: the maximum(starting) value of epsilon.
```
CatherineSue: The

garage/tf/policies/epsilon_greedy_policy.py
```diff
@@ -0,0 +1,118 @@
...
+        Serializable.quick_init(self, locals())
+        super(EpsilonGreedyPolicy, self).__init__(env_spec)
+
+        self._env_spec = env_spec
+        self._max_epsilon = max_epsilon
+        self._epsilon = max_epsilon
+        self._min_epsilon = min_epsilon
+        self._decay_period = int(total_step * decay_ratio)
+        self._action_space = env_spec.action_space
+        self._act = None
```
CatherineSue: Why not set a default value for `_act`?

garage/tf/policies/epsilon_greedy_policy.py
```diff
@@ -0,0 +1,118 @@
...
+    @overrides
+    def get_action(self, observation):
+        """
+        Get action from this policy for the input observation.
+
+        Args:
+            observation: Observation from environment.
+
+        Returns:
+            opt_action: optimal action from this policy.
```
CatherineSue: Optimal

garage/tf/policies/epsilon_greedy_policy.py
```diff
@@ -0,0 +1,118 @@
...
+    @overrides
+    def get_actions(self, observations):
+        """
+        Get actions from this policy for the input observations.
+
+        Args:
+            observations: Observations from environment.
+
+        Returns:
+            opt_action: optimal actions from this policy.
```
CatherineSue: Optimal

garage/tf/policies/epsilon_greedy_policy.py
ahtsan: Like random action?

garage/tf/policies/epsilon_greedy_policy.py
```diff
@@ -0,0 +1,118 @@
...
+
+        self._env_spec = env_spec
+        self._max_epsilon = max_epsilon
+        self._epsilon = max_epsilon
+        self._min_epsilon = min_epsilon
+        self._decay_period = int(total_step * decay_ratio)
+        self._action_space = env_spec.action_space
+        self._act = None
+
+    def set_act(self, act):
```
ryanjulian: This is not how things are usually designed in garage.

The Q-Function should be set during the constructor.

I believe what you actually want here is an ExplorationStrategy which wraps a Q-Function-derived policy of your choice.

I am not opposed to changing the inheritance tree to make ExplorationStrategy inherit from/wrap policies (which is basically what you did here). If you want to make that change, you should make it consistent across the codebase.

garage/tf/policies/epsilon_greedy_policy.py
```diff
@@ -0,0 +1,118 @@
...
+        """
+        Set the act function.
+
+        The act function is used to choose action
+        given observations.
+        """
+        self._act = act
```
ryanjulian: Though completely valid Python, using function assignment to create dynamically-defined methods like this is definitely an anti-pattern. It's not always wrong, but it requires a really strong justification (which this case does not warrant).

garage/tf/policies/epsilon_greedy_policy.py
```diff
@@ -0,0 +1,118 @@
...
+class EpsilonGreedyPolicy(Policy, Serializable):
+    """
+    ϵ-greedy policy.
+
+    Select action based on the value of ϵ. ϵ will decrease from
+    max_epsilon to min_epsilon within decay_ratio * total_step.
+
+    At state s, with probability
+    1 − ϵ: select action = argmax Q(s, a)
+    ϵ    : select a random action
```
ryanjulian: Random how? What is the distribution?

garage/tf/policies/epsilon_greedy_policy.py
ahtsan: Yes, I want a ExplorationStrategy which wraps a Q-Function-derived policy. The reason I pass the Q-Function here is that I won't have a Q-Function built until I constructed the algorithm class, i.e. `DQN` in this case. However, ExplorationStrategy is one of the argument needed. I can only either set the Q-Function in the ExplorationStrategy or set the ExplorationStrategy in the DQN class afterwards.

garage/tf/policies/epsilon_greedy_policy.py
```diff
@@ -0,0 +1,118 @@
...
+        if self._epsilon > self._min_epsilon:
+            self._epsilon -= (
+                self._max_epsilon - self._min_epsilon) / self._decay_period
+
+        opt_action = np.argmax(self._act(observations), axis=1)
+
+        for itr in range(len(opt_action)):
+            if np.random.random() < self._epsilon:
+                if isinstance(self._action_space, Box):
+                    opt_action[itr] = np.random.randint(
```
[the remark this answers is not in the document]
ahtsan: That's right, I will change it to 
`np.random.uniform(self._action_space.low, self._action_space.high)`

garage/tf/policies/epsilon_greedy_policy.py
ahtsan: I will change it to `np.random.uniform(...)` in choosing random actions so it will be from a uniform distribution.

garage/tf/policies/epsilon_greedy_policy.py
ahtsan: I will try to use ExplorationStrategy class and pass q-values to `get_actions`.

ahtsan:
Current design:
qf_derived_policy: greedy policy based on q-values of actions.
epsilon_greedy_strategy: handle the epsilon part. Wrapping policy to get optimal actions.

qf is passed to policy constructor, and tf.session is given to the policy and exploration strategy to run qf for getting q-values.

garage/tf/exploration_strategies/__init__.py
```diff
@@ -1,6 +1,8 @@
 """
 Exploration strategies used for RL algorithms.
 """
+from garage.tf.exploration_strategies.epsilon_greedy_strategy import (
+    EpsilonGreedyStrategy)
 from garage.tf.exploration_strategies.ou_strategy import OUStrategy
 
-__all__ = ["OUStrategy"]
+__all__ = ["OUStrategy", "EpsilonGreedyStrategy"]
```
jonashen: Does this need to be in alphabetical order

CatherineSue: Yes please do it in alphabetical order

garage/tf/exploration_strategies/ou_strategy.py
```diff
@@ -64,7 +64,7 @@ def reset(self):
             self.action_dim)
 
     @overrides
-    def get_action(self, t, observation, policy, **kwargs):
+    def get_action(self, t, observation, policy):
```
CatherineSue: Why delete `**kwargs` here?

garage/tf/exploration_strategies/epsilon_greedy_strategy.py
```diff
@@ -0,0 +1,109 @@
...
+        """
+        if self._epsilon > self._min_epsilon:
+            self._epsilon -= (
+                self._max_epsilon - self._min_epsilon) / self._decay_period
+
+        opt_action = policy.get_action(observation, sess)
+
+        if np.random.random() < self._epsilon:
+            if isinstance(self._action_space, Box):
+                opt_action = np.random.uniform(self._action_space.low,
```
CatherineSue: Could you use `self._action_space.sample()` here? What if the dtype is uinit8?

tests/garage/tf/exploration_strategies/test_epsilon_greedy_strategy.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def test_epsilon_greedy_strategy(self):
+        obs_ph = tf.placeholder(tf.float32, shape=(None, 4))
+        qf = mlp(
+            input_var=obs_ph,
+            output_dim=2,
+            hidden_sizes=(32, 32),
+            hidden_nonlinearity=tf.nn.relu,
+            name="mlp")
+        policy = QfDerivedPolicy(
```
CatherineSue: What should the qf param of `QfDerivedPolicy` be if `qf=DiscreteMLPQFunction(...)`?

garage/tf/exploration_strategies/epsilon_greedy_strategy.py
```diff
@@ -0,0 +1,109 @@
...
+        env_spec: Environment specification
+        total_step: Total steps in the training.
+        max_epsilon: The maximum(starting) value of epsilon.
+        min_epsilon: The minimum(terminal) value of epsilon.
+        decay_ratio: Fraction of total steps for epsilon decay.
+    """
+
+    def __init__(self,
+                 env_spec,
+                 total_step,
```
CatherineSue: Is `total_step` the same as `max_path_length`? If it is, please use `max_path_length` then.

garage/tf/exploration_strategies/epsilon_greedy_strategy.py
CatherineSue: And it can skip the check of the space type.

tests/garage/tf/policies/test_qf_derived_policy.py
```diff
@@ -0,0 +1,53 @@
...
+from garage.tf.policies import QfDerivedPolicy
+from tests.fixtures import TfGraphTestCase
+
+
+class TestQfDerivedPolicy(TfGraphTestCase):
+    @overrides
+    def setUp(self):
+        self.observation_shape = (4, )
+        self.action_shape = 2
+        self.env = mock.Mock()
```
CatherineSue: What's the difference with mock.Mock() here and tests.fixtures.DummyEnv? I think DummyEnv is more convenient to use.

garage/tf/exploration_strategies/epsilon_greedy_strategy.py
CatherineSue: Got it. It'd be better to add this, i.e. max_path_length * n_epochs, in docstring I think.

garage/tf/exploration_strategies/epsilon_greedy_strategy.py
ahtsan: Yes

tests/garage/tf/exploration_strategies/test_epsilon_greedy_strategy.py
CatherineSue: LGTM.

garage/tf/exploration_strategies/ou_strategy.py
```diff
@@ -81,7 +81,8 @@ def get_action(self, t, observation, policy, **kwargs):
         return np.clip(action + ou_state, self.action_space.low,
                        self.action_space.high), agent_infos
 
-    def get_actions(self, observations, policy):
+    @overrides
+    def get_actions(self, t, observations, policy):
```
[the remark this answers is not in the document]
ahtsan: ok

garage/exploration_strategies/base.py
```diff
@@ -1,5 +1,8 @@
 class ExplorationStrategy:
-    def get_action(self, t, observation, policy, **kwargs):
+    def get_action(self, t, observation, policy, sess=None, **kwargs):
```
ryanjulian: why would i ever need to specify the session? This pattern isn't used anywhere else in primitives. 

Imagine how dangerous it could be if policy were in one session and you passed a different session to ExplorationStrategy.

I think you want something like

```python
with my_session:
    foo.get_action(t, obs, policy)
```

ahtsan: since we are removing layers, sometimes policy needs to session to get the actions, e.g. in `discrete_qf_derived_policy`.
 
Do you mean we should use `with my_session:` in ExplorationStrategy and do `tf.get_default_session()` inside policy?

ryanjulian: You should not need a to initiate a session anywhere in your primitives because the algorithm should have done that for you.

i.e. `tf.get_default_session()` should always work inside your primitives.

ahtsan: got it

CatherineSue: It looks like the change serves the purpose of removing `tensor_utils.compile_function()`. But if we are to replace it with `make_callable` instead of `sess.run()`, we may not need to feed sess here. The issue is beyond the scope of this PR. For this PR I suggest we do minimum or none API changes if possible. 
- Remove the API changes in exploration_strategy and sampler.
- Add get_qval in DiscreteMLPQFunction. This should avoid the need for passing sess to policy.

ryanjulian: As always, I defer to Chang's judgement on TF issues.

garage/tf/q_functions/discrete_mlp_q_function.py
```diff
@@ -8,52 +8,79 @@
...
         self._hidden_sizes = hidden_sizes
         self._hidden_nonlinearity = hidden_nonlinearity
         self._output_nonlinearity = output_nonlinearity
         self._layer_norm = layer_norm
 
+        with tf.name_scope(name):
+            self.q_val, self.obs_ph = self.build_net(name)
+
+    def _build_ph(self, scope):
+
```
CatherineSue: Please add a docstring here.

ahtsan:
Epsilon greedy exploration strategy and discrete qf-derived policy will be used in DQN.

ahtsan:
@ryanjulian @CatherineSue @krzentner please review.

tests/garage/tf/exploration_strategies/test_epsilon_greedy_strategy.py
```diff
@@ -0,0 +1,37 @@
...
+            min_epsilon=0.02,
+            decay_ratio=0.1)
+
+        self.sess.run(tf.global_variables_initializer())
+
+        env.reset()
+        obs, _, _, _ = env.step(1)
+
+        action = epilson_greedy_strategy.get_action(0, obs, policy)
+        assert action in np.arange(env.action_space.n)
```
ryanjulian: is this equivalent to `env.action_space.contains(action)`?

tests/garage/tf/exploration_strategies/test_epsilon_greedy_strategy.py
```diff
@@ -0,0 +1,37 @@
...
+            max_epsilon=1.0,
+            min_epsilon=0.02,
+            decay_ratio=0.1)
+
+        self.sess.run(tf.global_variables_initializer())
+
+        env.reset()
+        obs, _, _, _ = env.step(1)
+
+        action = epilson_greedy_strategy.get_action(0, obs, policy)
```
ryanjulian: can you verify that the probability of random is actually near \epsilon?

tests/garage/tf/exploration_strategies/test_epsilon_greedy_strategy.py
```diff
@@ -0,0 +1,37 @@
...
+
+from garage.tf.envs import TfEnv
+from garage.tf.exploration_strategies import EpsilonGreedyStrategy
+from garage.tf.policies import DiscreteQfDerivedPolicy
+from garage.tf.q_functions import DiscreteMLPQFunction
+from tests.fixtures import TfGraphTestCase
+from tests.fixtures.envs.dummy import DummyDiscreteEnv
+
+
+class TestEpsilonGreedyStrategy(TfGraphTestCase):
```
ryanjulian: try to write this as a unit test rather than an integration test. i don't think you need any TensorFlow dependency to test the epislon greedy strategy.

garage/tf/exploration_strategies/epsilon_greedy_strategy.py
```diff
@@ -0,0 +1,97 @@
+"""
```
ryanjulian: this doesn't seem to have any tf dependencies, so why put it in the TF tree?

garage/tf/exploration_strategies/ou_strategy.py
```diff
@@ -81,7 +81,8 @@ def get_action(self, t, observation, policy, **kwargs):
         return np.clip(action + ou_state, self.action_space.low,
                        self.action_space.high), agent_infos
 
-    def get_actions(self, observations, policy):
+    @overrides
```
ryanjulian: this file also doesn't seem to have any tf dependencies. move out of the TF tree?

ahtsan: yes

garage/tf/q_functions/discrete_mlp_q_function.py
```diff
@@ -8,52 +8,78 @@
...
+        network = mlp(
+            input_var=obs_ph,
             output_dim=self._action_dim,
             hidden_sizes=self._hidden_sizes,
             name=name,
             hidden_nonlinearity=self._hidden_nonlinearity,
             output_nonlinearity=self._output_nonlinearity,
             layer_normalization=self._layer_norm)
+
+        return network, obs_ph
```
ryanjulian: in the rest of the TF tree, the algorithm provides a placeholder to build. i think it would be good to follow that convention (since it aligns with Models)

garage/tf/q_functions/discrete_mlp_q_function.py
```diff
@@ -8,52 +8,78 @@
...
+
+    @overrides
+    def get_qval_sym(self, input_phs):
+        assert len(input_phs) == 1
```
ryanjulian: please test and raise a ValueError, rather than just doing a naked assert

garage/tf/samplers/batch_sampler.py
```diff
@@ -61,8 +61,8 @@ def process_samples(self, itr, paths):
         for idx, path in enumerate(paths):
             path_baselines = np.append(all_path_baselines[idx], 0)
             deltas = path["rewards"] + \
-                     self.algo.discount * path_baselines[1:] - \
-                     path_baselines[:-1]
+                self.algo.discount * path_baselines[1:] - \
```
ryanjulian: not sure why this file got touched?

tests/garage/tf/policies/test_qf_derived_policy.py
```diff
@@ -0,0 +1,28 @@
...
+from garage.tf.policies import DiscreteQfDerivedPolicy
+from garage.tf.q_functions import DiscreteMLPQFunction
+from tests.fixtures import TfGraphTestCase
+from tests.fixtures.envs.dummy import DummyDiscreteEnv
+
+
+class TestQfDerivedPolicy(TfGraphTestCase):
+    def test_discrete_qf_derived_policy(self):
+        env = TfEnv(DummyDiscreteEnv())
+        qf = DiscreteMLPQFunction(env.spec)
```
ryanjulian: can you use a simpler QFunction to test this as more of a unit test? why not just `tf.identity`?

tests/garage/tf/exploration_strategies/test_epsilon_greedy_strategy.py
ahtsan: yes

garage/tf/exploration_strategies/ou_strategy.py
ahtsan: I checked the original `garage.exploration_strategies.ou_strategy` is not used anywhere. I will remove that and move the tf one there.

garage/tf/policies/discrete_qf_derived_policy.py
```diff
@@ -0,0 +1,75 @@
+"""
+QfDerived policy.
```
CatherineSue: DiscreteQfDerived

garage/tf/policies/discrete_qf_derived_policy.py
```diff
@@ -0,0 +1,75 @@
...
+
+        Args:
+            observation: Observation from environment.
+            sess: tf.Session provided.
+
+        Returns:
+            opt_action: Optimal action from this policy.
+
+        """
+        sess = tf.get_default_session()
```
CatherineSue: Is this gonna be a problem? Have we settled down how do we design this?

garage/tf/q_functions/discrete_mlp_q_function.py
```diff
@@ -8,52 +8,78 @@
 
 class DiscreteMLPQFunction(QFunction):
     """
-    Discrete MLP Function class.
+    Discrete MLP Function.
```
CatherineSue: Discrete MLP Q Function

garage/tf/exploration_strategies/epsilon_greedy_strategy.py
```diff
@@ -0,0 +1,97 @@
...
+        Returns:
+            opt_action: optimal actions from this policy.
+
+        """
+        if self._epsilon > self._min_epsilon:
+            self._epsilon -= (
+                self._max_epsilon - self._min_epsilon) / self._decay_period
+
+        opt_actions = policy.get_actions(observations)
+        for itr in range(len(opt_actions)):
```
CatherineSue: I think replace with a random mask and np.where will be fatster than a for loop.

garage/tf/exploration_strategies/epsilon_greedy_strategy.py
```diff
@@ -0,0 +1,97 @@
...
+        Args:
+            t: Iteration.
+            observation: Observation from the environment.
+            policy: Policy network to predict action based on the observation.
+
+        Returns:
+            opt_action: optimal actions from this policy.
+
+        """
+        if self._epsilon > self._min_epsilon:
```
CatherineSue: You can wrap this update in a self.stimulate since it’s used twice

garage/tf/exploration_strategies/epsilon_greedy_strategy.py
CatherineSue: I can't seem to understand this equation, it looks related to your training architecture of dqn. Which is
```
for epoch in range(n_epochs):
    for t in range(n_training_steps):
```
but the regular way is:
```
for epoch in range(n_epochs):
    for epoch_cycle in range(n_epoch_cycles):
        for t in range(n_training_steps):
```
Please discuss the details about this because I think we need to have a way to circumvent this.

garage/tf/exploration_strategies/ou_strategy.py
ryanjulian: sgtm

garage/tf/policies/discrete_qf_derived_policy.py
ryanjulian: AFAICT this should work fine now. I don't think we settled that. Can you propose some designs?

garage/tf/exploration_strategies/epsilon_greedy_strategy.py
CatherineSue: After discussing with you, I think we can rename it to total_timesteps or include it is the same as total time steps in docstring.

garage/tf/policies/discrete_qf_derived_policy.py
CatherineSue: Yeah, I agree it works properly now. Since we introduced Model class and made the assumption that a session is created when we instantiate a primitive, we can later replace this with make_callable.

garage/tf/exploration_strategies/epsilon_greedy_strategy.py
```diff
@@ -0,0 +1,97 @@
...
+
+Random exploration according to the value of epsilon.
+"""
+import numpy as np
+
+from garage.exploration_strategies import ExplorationStrategy
+from garage.misc.overrides import overrides
+
+
+class EpsilonGreedyStrategy(ExplorationStrategy):
```
CatherineSue: Is this pickable?

ahtsan: Yes, I will add a test.

garage/tf/exploration_strategies/epsilon_greedy_strategy.py
ahtsan: since the delay is done per sample (observation in this case), I think we still need the for loop.

CatherineSue: Ok

garage/tf/q_functions/discrete_mlp_q_function.py
```diff
@@ -8,52 +8,70 @@
...
             name=name,
             hidden_nonlinearity=self._hidden_nonlinearity,
             output_nonlinearity=self._output_nonlinearity,
             layer_normalization=self._layer_norm)
+
+    @overrides
+    def get_qval_sym(self, input_phs):
+        assert len(input_phs) == 1, "Only one input should be given."
+
+        return mlp(
```
ryanjulian: how do i know this is in the same variable_scope as all the other calls to `build_net()`? this is exactly the same problem that Models tries to solve...

ahtsan: you are right, we should have a `tf.variable_scope(name, reuse=False)` in `build_net()` and `tf.variable_scope(name, reuse=True)` in `get_qval_sym()`.

ryanjulian: if you can do it in <30 mins, why not replace this with a DiscreteMLPModel?

ahtsan: Yes that's a good idea.

ahtsan:
- Used `DiscreteMLPModel`
- Added pickle test for `DiscreteMLPQFunction`

We might want to make `model.network_output_spec()` from an interface into a field, e.g. I want to have `model.network.q_val` when I use `DiscreteMLPModel` for `DiscreteMLPQFunction`, and `model.network.action` when I use `DiscreteMLPModel` for `DiscreteMLPPolicy`. If that makes sense, I will do that in another PR.

ryanjulian:
@ahtsan i think what you describe is probably best handled by inheriting primitives from their models?

ahtsan:
That's right.

garage/tf/q_functions/discrete_mlp_q_function.py
```diff
@@ -2,58 +2,72 @@
...
 
     @overrides
-    def build_net(self, name, input):
+    def get_qval_sym(self, input_var, name):
         """
-        Set up q network based on class attributes.
+        Symbolic graph for q-network.
 
         Args:
+            input_var: Input tf.placeholder to the network.
```
ryanjulian: what is the input_var? is it state?

ahtsan: Yes

ryanjulian: can you update the variable names/docstrings accordingly?

ryanjulian:
Your tests are actually failing but they are masked by #542 . Take a look at the build log.

ahtsan:
CI passed.

ryanjulian:
feel free to click the big green button

ryanjulian:
please remember to rewrite the commit message (into one which is not just a bunch of random strings from your previous commits).

remember the commit message guidelines from CONTRIBUTING.md

PR #458, opened 2019-01-05
Add tf/TD3

jonashen:
1. Add TD3 to TF. TD3 is DDPG with two critics.
2. Add example.

### DIFF
--- a/examples/tf/td3_pendulum.py
+++ b/examples/tf/td3_pendulum.py
@@ -1,0 +1,83 @@
+"""
+This is an example to train a task with TD3 algorithm.
+
+Here, we create a gym environment InvertedDoublePendulum
+and use a TD3 with 1M steps.
+
+Results:
+    AverageReturn: 250
+    RiseTime: epoch 499
+"""
+import gym
+import tensorflow as tf
+
+from garage.experiment import run_experiment
+from garage.np.exploration_strategies.gaussian_strategy import GaussianStrategy
+from garage.replay_buffer import SimpleReplayBuffer
+from garage.tf.algos import TD3
+from garage.tf.envs import TfEnv
+from garage.tf.experiment import LocalTFRunner
+from garage.tf.policies import ContinuousMLPPolicy
+from garage.tf.q_functions import ContinuousMLPQFunction
+
+
+def run_task(snapshot_config, *_):
+    """
+    Wrap TD3 training task in the run_task function.
+
+    :param _:
+    :return:
+    """
+    with LocalTFRunner(snapshot_config) as runner:
+        env = TfEnv(gym.make('InvertedDoublePendulum-v2'))
+
+        action_noise = GaussianStrategy(env.spec, max_sigma=0.1, min_sigma=0.1)
+
+        policy = ContinuousMLPPolicy(env_spec=env.spec,
+                                     hidden_sizes=[400, 300],
+                                     hidden_nonlinearity=tf.nn.relu,
+                                     output_nonlinearity=tf.nn.tanh)
+
+        qf = ContinuousMLPQFunction(name='ContinuousMLPQFunction',
+                                    env_spec=env.spec,
+                                    hidden_sizes=[400, 300],
+                                    action_merge_layer=0,
+                                    hidden_nonlinearity=tf.nn.relu)
+
+        qf2 = ContinuousMLPQFunction(name='ContinuousMLPQFunction2',
+                                     env_spec=env.spec,
+                                     hidden_sizes=[400, 300],
+                                     action_merge_layer=0,
+                                     hidden_nonlinearity=tf.nn.relu)
+
+        replay_buffer = SimpleReplayBuffer(env_spec=env.spec,
+                                           size_in_transitions=int(1e6),
+                                           time_horizon=250)
+
+        td3 = TD3(env_spec=env.spec,
+                  policy=policy,
+                  policy_lr=1e-4,
+                  qf_lr=1e-3,
+                  qf=qf,
+                  qf2=qf2,
+                  replay_buffer=replay_buffer,
+                  target_update_tau=1e-2,
+                  n_epoch_cycles=20,
+                  n_train_steps=1,
+                  smooth_return=False,
+                  discount=0.99,
+                  buffer_batch_size=100,
+                  min_buffer_size=1e4,
+                  exploration_strategy=action_noise,
+                  policy_optimizer=tf.train.AdamOptimizer,
+                  qf_optimizer=tf.train.AdamOptimizer)
+
+        runner.setup(td3, env)
+        runner.train(n_epochs=500, n_epoch_cycles=20, batch_size=250)
+
+
+run_experiment(
+    run_task,
+    snapshot_mode='last',
+    seed=1,
+)
--- a/setup.py
+++ b/setup.py
@@ -62,6 +62,7 @@ extras['dev'] = [
     'flake8',
     'flake8-docstrings==1.3.0',
     'flake8-import-order',
+    'gtimer',
     'pandas',
     'pep8-naming==0.7.0',
     'pre-commit',
@@ -73,6 +74,7 @@ extras['dev'] = [
     'pytest-cov',
     'pytest-xdist',
     'recommonmark',
+    'rlkit @ git+https://github.com/vitchyr/rlkit/@1d469a509b797ca04a39b8734c1816ca7d108fc8',  # noqa: E501
     'sphinx',
     'sphinx_rtd_theme',
     'yapf==0.28.0',
--- a/src/garage/np/exploration_strategies/gaussian_strategy.py
+++ b/src/garage/np/exploration_strategies/gaussian_strategy.py
@@ -1,3 +1,4 @@
+"""Gaussian exploration strategy."""
 import gym
 import numpy as np
 
@@ -6,10 +7,7 @@ from garage.np.exploration_strategies.base import ExplorationStrategy
 
 
 class GaussianStrategy(ExplorationStrategy, Serializable):
-    """
-    This strategy adds Gaussian noise to the action taken by the deterministic
-    policy.
-    """
+    """Add Gaussian noise to the action taken by the deterministic policy."""
 
     def __init__(self,
                  env_spec,
@@ -24,9 +22,46 @@ class GaussianStrategy(ExplorationStrategy, Serializable):
         self._decay_period = decay_period
         self._action_space = env_spec.action_space
 
-    def get_action(self, t, observation, policy, **kwargs):
+    def get_action(self, iteration, observation, policy, **kwargs):
+        """
+        Get action from this policy for the input observation.
+
+        Args:
+            iteration(int): Iteration.
+            observation(numpy.ndarray): Observation from the environment.
+            policy(garage.tf.policies.base.Policy):
+                Policy network to predict action based on the observation.
+
+        Returns:
+            opt_action(numpy.ndarray): optimal action from this policy.
+            agent_info(dict): Agent information.
+
+        """
         action, agent_info = policy.get_action(observation)
         sigma = self._max_sigma - (self._max_sigma - self._min_sigma) * min(
-            1.0, t * 1.0 / self._decay_period)
+            1.0, iteration * 1.0 / self._decay_period)
         return np.clip(action + np.random.normal(size=len(action)) * sigma,
-                       self._action_space.low, self._action_space.high)
+                       self._action_space.low,
+                       self._action_space.high), agent_info
+
+    def get_actions(self, iteration, observations, policy, **kwargs):
+        """
+        Get actions from this policy for the input observation.
+
+        Args:
+            iteration(int): Iteration.
+            observatioan(list): Observationa from the environment.
+            policy(garage.tf.policies.base.Policy):
+                Policy network to predict action based on the observation.
+
+        Returns:
+            opt_actions(numpy.ndarray): optimal actions from this policy.
+            agent_infos(dict): Agent information.
+
+        """
+        actions, agent_infos = policy.get_actions(observations)
+        sigma = self._max_sigma - (self._max_sigma - self._min_sigma) * min(
+            1.0, iteration * 1.0 / self._decay_period)
+        return np.clip(actions + np.random.normal(size=len(actions)) * sigma,
+                       self._action_space.low,
+                       self._action_space.high), agent_infos
--- a/src/garage/tf/algos/ddpg.py
+++ b/src/garage/tf/algos/ddpg.py
@@ -1,3 +1,5 @@
+"""This module implements a DDPG model."""
+
 from collections import deque
 
 from dowel import logger, tabular
@@ -106,26 +108,26 @@ class DDPG(OffPolicyRLAlgorithm):
 
         self.target_policy = policy.clone('target_policy')
 
-        super(DDPG, self).__init__(
-            env_spec=env_spec,
-            policy=policy,
-            qf=qf,
-            n_train_steps=n_train_steps,
-            n_epoch_cycles=n_epoch_cycles,
-            max_path_length=max_path_length,
-            buffer_batch_size=buffer_batch_size,
-            min_buffer_size=min_buffer_size,
-            rollout_batch_size=rollout_batch_size,
-            exploration_strategy=exploration_strategy,
-            replay_buffer=replay_buffer,
-            use_target=True,
-            discount=discount,
-            reward_scale=reward_scale,
-            input_include_goal=input_include_goal,
-            smooth_return=smooth_return)
+        super(DDPG, self).__init__(env_spec=env_spec,
+                                   policy=policy,
+                                   qf=qf,
+                                   n_train_steps=n_train_steps,
+                                   n_epoch_cycles=n_epoch_cycles,
+                                   max_path_length=max_path_length,
+                                   buffer_batch_size=buffer_batch_size,
+                                   min_buffer_size=min_buffer_size,
+                                   rollout_batch_size=rollout_batch_size,
+                                   exploration_strategy=exploration_strategy,
+                                   replay_buffer=replay_buffer,
+                                   use_target=True,
+                                   discount=discount,
+                                   reward_scale=reward_scale,
+                                   input_include_goal=input_include_goal,
+                                   smooth_return=smooth_return)
 
     @overrides
     def init_opt(self):
+        """Build the loss function and init the optimizer."""
         with tf.name_scope(self.name, 'DDPG'):
             # Create target policy and qf network
             self.target_policy_f_prob_online = tensor_utils.compile_function(
@@ -157,20 +159,21 @@ class DDPG(OffPolicyRLAlgorithm):
                         flat_dim_with_keys(['observation', 'desired_goal'])
                 else:
                     obs_dim = self.env_spec.observation_space.flat_dim
-                y = tf.compat.v1.placeholder(
-                    tf.float32, shape=(None, 1), name='input_y')
-                obs = tf.compat.v1.placeholder(
-                    tf.float32,
-                    shape=(None, obs_dim),
-                    name='input_observation')
+                y = tf.compat.v1.placeholder(tf.float32,
+                                             shape=(None, 1),
+                                             name='input_y')
+                obs = tf.compat.v1.placeholder(tf.float32,
+                                               shape=(None, obs_dim),
+                                               name='input_observation')
                 actions = tf.compat.v1.placeholder(
                     tf.float32,
                     shape=(None, self.env_spec.action_space.flat_dim),
                     name='input_action')
             # Set up policy training function
             next_action = self.policy.get_action_sym(obs, name='policy_action')
-            next_qval = self.qf.get_qval_sym(
-                obs, next_action, name='policy_action_qval')
+            next_qval = self.qf.get_qval_sym(obs,
+                                             next_action,
+                                             name='policy_action_qval')
             with tf.name_scope('action_loss'):
                 action_loss = -tf.reduce_mean(next_qval)
                 if self.policy_weight_decay > 0.:
@@ -213,6 +216,7 @@ class DDPG(OffPolicyRLAlgorithm):
             self.f_update_target = f_update_target
 
     def __getstate__(self):
+        """Object.__getstate__."""
         data = self.__dict__.copy()
         del data['target_policy_f_prob_online']
         del data['target_qf_f_prob_online']
@@ -223,10 +227,18 @@ class DDPG(OffPolicyRLAlgorithm):
         return data
 
     def __setstate__(self, state):
+        """Object.__setstate__."""
         self.__dict__ = state
         self.init_opt()
 
     def train_once(self, itr, paths):
+        """Perform one step of policy optimization given one batch of samples.
+
+        Args:
+            itr (int): Iteration number.
+            paths (list[dict]): A list of collected paths.
+
+        """
         paths = self.process_samples(itr, paths)
 
         epoch = itr / self.n_epoch_cycles
@@ -288,14 +300,17 @@ class DDPG(OffPolicyRLAlgorithm):
 
     @overrides
     def optimize_policy(self, itr, samples_data):
-        """
-        Perform algorithm optimizing.
+        """Perform algorithm optimizing.
+
+        Args:
+            itr (int): Iterations.
+            samples_data (list): Processed batch data.
 
         Returns:
-            action_loss: Loss of action predicted by the policy network.
-            qval_loss: Loss of q value predicted by the q network.
-            ys: y_s.
-            qval: Q value predicted by the q network.
+            action_loss (float): Loss of action predicted by the policy network
+            qval_loss (float): Loss of q value predicted by the q network.
+            ys (float): y_s.
+            qval (float): Q value predicted by the q network.
 
         """
         transitions = self.replay_buffer.sample(self.buffer_batch_size)
@@ -335,4 +350,5 @@ class DDPG(OffPolicyRLAlgorithm):
 
     @overrides
     def get_itr_snapshot(self, itr):
+        """Return data saved in the snapshot for this iteration."""
         return dict(itr=itr, policy=self.policy)
--- a/tests/benchmarks/test_benchmark_td3.py
+++ b/tests/benchmarks/test_benchmark_td3.py
@@ -1,0 +1,320 @@
+"""
+This script creates a regression test over garage-TD3.
+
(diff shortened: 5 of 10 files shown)


garage/tf/algos/td3.py
```diff
@@ -0,0 +1,365 @@
...
+                 clip_pos_returns=False,
+                 clip_return=np.inf,
+                 discount=0.99,
+                 max_action=None,
+                 name=None,
+                 **kwargs):
+        """
+        Construct class.
+
+        Args:
```
CatherineSue: I believe we are now moving constructor params to class docstring.

jonashen: Thanks for the heads up!

ryanjulian:
Great to see this PR up.

Can you work with Chang and Kevin to setup the benchmarks? We need to prove we can reproduce the performance from the paper and rlkit before we can merge this.

It also looks like you're missing tests, but I assume you're working on those already.

tests/garage/tf/algos/test_td3.py
```diff
@@ -0,0 +1,64 @@
...
+            env,
+            policy=policy,
+            policy_lr=1e-4,
+            qf_lr=1e-3,
+            qf=qf,
+            qf2=qf2,
+            replay_buffer=replay_buffer,
+            plot=False,
+            target_update_tau=1e-2,
+            n_epochs=500,
```
ryanjulian: how long does this test take to run?

ryanjulian: we need a small version which can be run by the CI in ~<1 minute. long tests will be run by the benchmark system.

ryanjulian:
Have you started writing the benchmark yet? @CatherineSue can help you set it up.

We need to show equivalent performance to the original paper and the version in rlkit before we can merge this.

jonashen:
Average Return over 1M steps
image

CatherineSue:
Is this running using the same hyper params as rlkit/td3?

jonashen:
Results using rlkit's hyperparameters.
hopper-v2_benchmark
invertedpendulum-v2_benchmark
reacher-v2_benchmark
swimmer-v2_benchmark
walker2d-v2_benchmark
halfcheetah-v2_benchmark

ryanjulian:
Can you make the same plots but for rlkit (or link to pre-baked ones?)

examples/tf/td3_pendulum.py
```diff
@@ -0,0 +1,86 @@
...
+            qf2=qf2,
+            replay_buffer=replay_buffer,
+            target_update_tau=1e-2,
+            n_epoch_cycles=20,
+            max_path_length=1000,
+            n_train_steps=50,
+            smooth_return=False,
+            discount=0.99,
+            buffer_batch_size=100,
+            min_buffer_size=int(1e4),
```
ryanjulian: you should probably apply int() to this argument in TD3 instead

garage/tf/algos/off_policy_rl_algorithm.py
```diff
@@ -87,3 +82,61 @@ def get_itr_snapshot(self, itr, samples_data):
     def optimize_policy(self, itr, samples_data):
         """Optimize policy network."""
         raise NotImplementedError
+
```
ryanjulian: is this really identical for all off-policy algorithms, or just DDPG? is DQN the same? what about SAC?

maybe you just want to inherit from DDPG instead.

garage/tf/algos/td3.py
```diff
@@ -0,0 +1,270 @@
+"""
+This module implements a TD3 model.
```
ryanjulian: TD3 algorithm

Provide a link to the paper

tests/garage/tf/algos/test_td3.py
```diff
@@ -0,0 +1,71 @@
...
+from garage.exploration_strategies import OUStrategy
+from garage.replay_buffer import SimpleReplayBuffer
+from garage.tf.algos import TD3
+from garage.tf.envs import TfEnv
+from garage.tf.policies import ContinuousMLPPolicy
+from garage.tf.q_functions import ContinuousMLPQFunction
+from tests.fixtures import TfGraphTestCase
+
+
+class TestTD3(TfGraphTestCase):
```
ryanjulian: how long does this take to run?

garage/tf/algos/off_policy_rl_algorithm.py
naeioi: That makes sense.

ryanjulian:
why closed???

naeioi:
Sorry. I made a mistake and github automatically closed this PR. I am reverting it.

tests/garage/tf/algos/test_td3.py
naeioi: 47s on my machine (i7-8700K)

naeioi:
This PR is ready for review.

garage/tf/algos/td3.py
ryanjulian: please fix @naeioi

garage/tf/algos/td3.py
```diff
@@ -0,0 +1,252 @@
...
+            target_update_tau(float): Interpolation parameter for doing the
+        soft target update.
+            discount(float): Discount factor for the cumulative return.
+            policy_lr(float): Learning rate for training policy network.
+            qf_lr(float): Learning rate for training q value network.
+            policy_weight_decay(float): L2 weight decay factor for parameters
+        of the policy network.
+            qf_weight_decay(float): L2 weight decay factor for parameters
+        of the q value network.
+            policy_optimizer(): Optimizer for training policy network.
```
ryanjulian: please complete the types here @naeioi

garage/tf/algos/td3.py
```diff
@@ -0,0 +1,252 @@
...
+            clip_pos_returns=clip_pos_returns,
+            clip_return=clip_return,
+            discount=discount,
+            max_action=max_action,
+            name=name,
+            **kwargs)
+
+    @overrides
+    def init_opt(self):
+        """Init the optimizer."""
```
ryanjulian: i think this actually builds the loss function (and then later initializes the optimizer)

garage/tf/algos/td3.py
```diff
@@ -0,0 +1,252 @@
...
+This module implements a TD3 model.
+
+TD3, or Twin Delayed Deep Deterministic Policy Gradient, uses actor-critic
+method to optimize the policy and reward prediction. Notably, it uses the
+minimum value of two critics instead of one to limit overestimation.
+"""
+
+import numpy as np
+import tensorflow as tf
+import tensorflow.contrib as tc
```
ryanjulian: please ensure that these are not deprecated on TF2.0. I think `tf.contrib` is getting removed

garage/tf/algos/td3.py
```diff
@@ -0,0 +1,252 @@
...
+            self.f_train_policy = f_train_policy
+            self.f_train_qf = f_train_qf
+            self.f_init_target = f_init_target
+            self.f_update_target = f_update_target
+            self.f_train_qf2 = f_train_qf2
+            self.f_init_target2 = f_init_target2
+            self.f_update_target2 = f_update_target2
+
+    @overrides
+    def optimize_policy(self, itr, samples_data):
```
ryanjulian: does this use the new runner interface?

tests/benchmarks/test_benchmark_td3.py
```diff
@@ -0,0 +1,205 @@
+"""
```
ryanjulian: please make sure this conforms to the format used by rlworkgroup/benchmarks

garage/tf/algos/td3.py
```diff
@@ -0,0 +1,252 @@
...
+                        qval_loss, var_list=self.qf2.get_trainable_vars())
+
+            f_train_qf = tensor_utils.compile_function(
+                inputs=[y, obs, actions],
+                outputs=[qf_train_op, qval_loss, qval])
+            f_train_qf2 = tensor_utils.compile_function(
+                inputs=[y, obs, actions],
+                outputs=[qf2_train_op, qval_loss, q2val])
+
+            self.f_train_policy = f_train_policy
```
ryanjulian: 1. can you use Session.make_callable() for this?
1. Can you put these all in this section (and only assign them once), so that people don't have to go hunting for them.

garage/tf/algos/td3.py
```diff
@@ -0,0 +1,252 @@
...
+            policy_weight_decay=policy_weight_decay,
+            qf_weight_decay=qf_weight_decay,
+            policy_optimizer=policy_optimizer,
+            qf_optimizer=qf_optimizer,
+            clip_pos_returns=clip_pos_returns,
+            clip_return=clip_return,
+            discount=discount,
+            max_action=max_action,
+            name=name,
+            **kwargs)
```
ryanjulian: please pass args explicitly rather than using `**kwargs`

garage/tf/algos/td3.py
naeioi: TD3 inherits the `train_once()` from DDPG.

garage/tf/algos/td3.py
naeioi: Are you suggesting that the name `init_opt` is not appropriate? I think DDPG also builds the loss function in `init_opt()`

ryanjulian: it was a remark about the comment, which doesn't seem to say much.

garage/tf/algos/td3.py
naeioi: Fixed.

garage/tf/algos/td3.py
naeioi: Removed.

tests/benchmarks/test_benchmark_td3.py
naeioi: Fixed.

garage/tf/algos/td3.py
naeioi: Fixed.

garage/tf/algos/td3.py
naeioi: I am working on it.

garage/tf/algos/td3.py
ryanjulian: ok

garage/tf/algos/ddpg.py
```diff
@@ -42,28 +44,58 @@ def __init__(self,
...
+                 n_train_steps=50,
+                 buffer_batch_size=64,
+                 min_buffer_size=1e4,
+                 rollout_batch_size=1,
+                 reward_scale=1.,
+                 input_include_goal=False,
+                 smooth_return=True,
+                 exploration_strategy=None):
         """
         Construct class.
```
ryanjulian: docstrings for the constructor go in the class docstring, per CONTRIBUTING.md

garage/tf/algos/td3.py
```diff
@@ -0,0 +1,290 @@
...
+            self.f_update_target2 = tf.get_default_session().make_callable(
+                target_update_op2)
+
+    @overrides
+    def optimize_policy(self, itr, samples_data):
+        """
+        Perform algorithm optimizing.
+
+        Returns:
+            action_loss: Loss of action predicted by the policy network.
```
ryanjulian: don't forget to provide types!

naeioi: Oops. My fault.

garage/tf/algos/ddpg.py
naeioi: I'll also move constructor docstring to class docstring for DDPG.

naeioi:
I am running benchmarks. Results will be posted here later.

naeioi:
Hopper-v2_benchmark
InvertedPendulum-v2_benchmark
Reacher-v2_benchmark
Swimmer-v2_benchmark
Walker2d-v2_benchmark
HalfCheetah-v2_benchmark

ryanjulian:
@naeioi can you compare these results to rlkit performance with the same tasks/hyperparameters? how about with the original paper?

utkarshjp7:
@ryanjulian @naeioi What is blocking us from merging this PR besides rebasing?

ryanjulian:
We need to verify the performance again other implementations. To my knowledge, other implementations are in rlkit (the one from the paper) and tf_agents.

ryanjulian:
Yes -- please refrain from using the layers library and do all tests with Models. We can prioritize the Models which are necessary for this change, for instance.

utkarshjp7:
This PR is ready for another review.

ryanjulian:
wow, these plots are a bit hard to interpret. do you mind making some smoothed versions which are easier to read? https://stackoverflow.com/a/26337730

garage seems a lot less stable than rlkit, even if it's getting higher performance on some benchmarks. one of the main advantages of TD3 is supposed to be better stability.

i think another thing you can do is to ensure that TF and Torch are only using CPUs only -- we recently found out that using GPU makes the benchmark non-deterministic.

utkarshjp7:
Smoothed Plots. `garage` and `rlkit` are only using CPUs.

HalfCheetah-v2
Hopper-v2
InvertedDoublePendulum-v2
InvertedPendulum-v2
Swimmer-v2
Walker2d-v2

ryanjulian:
So I'm not convinced this algorithm is working, in particular because garage fails to learn in several seeds on a few benchmarks where rlkit succeeds.

I'm also concerned about the performance of garage collapsing on several seeds after learning, since it's my understanding that TD3 is designed to make DDPG more stable (and collapsing is the opposite of stability). 

Please do a very close reading of the garage code + rlkit code + TD3 paper and see if you can find a discrepancy. If you get stuck, we can reach out to Vitchyr for help.

For future plots, it would be helpful to have a variance overlay + smoothed curve. You can do this very easily using `seaborn` (which is mostly interchangeable with matplotlib). https://seaborn.pydata.org/tutorial/relational.html#relational-tutorial

ryanjulian:
Looking forward to the new benchmarks!

utkarshjp7:
The benchmarks still doesn't look good. I will check our implementation against the official implementation (https://github.com/sfujim/TD3) next.

Hopper-v2_benchmark
InvertedPendulum-v2_benchmark
Swimmer-v2_benchmark

utkarshjp7:
Benchmarks are looking much better now! I changed `n_train_steps` from 50 to 1, and `n_epochs` from 200 to 1000, and `action_merge_layer` from 1 to 0 (as per the paper).

HalfCheetah-v2
Hopper-v2
InvertedDoublePendulum-v2
InvertedPendulum-v2
Swimmer-v2
Walker2d-v2

ryanjulian:
These look way better! Great sleuthing. Looks like this implementation is meeting/exceeding the original performance. 

I think this is ready-to-go.

There are a couple failure-to-learn cases on the pendulum tasks, which is hard to explain since it's doing so well on much harder tasks.

Can you look into these questions before committing?
* Perhaps the task is just high-variance for off-policy algorithms? You can verify by running both versions on a bunch of seeds (e.g. 30) and plotting the mean/variance curves.
* Is our exploration strategy different from the original?

tests/helpers.py
```diff
@@ -320,14 +318,12 @@ def plot(b_csvs, g_csvs, g_x, g_y, b_x, b_y, trials, seeds, plt_file, env_id,
...
-        plt.plot(
-            df_b[b_x],
-            df_b[b_y],
-            label='baselines_trial%d_seed%d' % (trial + 1, seed))
+        plt.plot(df_g[g_x],
+                 df_g[g_y],
+                 label='garage_trial%d_seed%d' % (trial + 1, seed))
+        plt.plot(df_b[b_x],
+                 df_b[b_y],
+                 label='baselines_trial%d_seed%d' % (trial + 1, seed))
```
ryanjulian: please replace 'baselines' with rlkit, or the name of the repo you're actually running in this benchmark.

src/garage/np/exploration_strategies/gaussian_strategy.py
```diff
@@ -24,9 +24,20 @@ def __init__(self,
         self._decay_period = decay_period
         self._action_space = env_spec.action_space
 
+    @overrides
```
ryanjulian: you can feel free not to use this. i think we will be eliminating it soon.

ryanjulian:
I think you are clear to merge, assuming the new plots look good. Please send final plot along before you merge, in any case.

tests/benchmarks/test_benchmark_td3.py
```diff
@@ -0,0 +1,321 @@
...
+import random
+
+from baselines.bench import benchmarks
+import dowel
+from dowel import logger as dowel_logger
+import gtimer as gt
+import gym
+import pytest
+try:
+    from rlkit.data_management.env_replay_buffer import EnvReplayBuffer
```
ryanjulian: please instead just add rlkit to the `dev` set of dependencies.

tests/garage/tf/algos/test_td3.py
```diff
@@ -0,0 +1,74 @@
...
+
+            algo = TD3(env_spec=env.spec,
+                       policy=policy,
+                       policy_lr=1e-3,
+                       qf_lr=1e-3,
+                       qf=qf,
+                       qf2=qf2,
+                       replay_buffer=replay_buffer,
+                       target_update_tau=0.005,
+                       n_epoch_cycles=20,
```
ryanjulian: aren't these two params handled by the runner?

@naeioi @zequnyu can you take a look to make sure this is using the runner API properly?

tests/benchmarks/test_benchmark_td3.py
utkarshjp7: I have added it to `dev` set but because it doesn't get installed properly, pytest throws error when it tries to collect all the tests and the build fails. Doing this will make pytest happy, and the build will pass. When we run benchmarks, we just install rlkit manually.

tests/garage/tf/algos/test_td3.py
```diff
@@ -0,0 +1,74 @@
...
+                       qf_weight_decay=0.001,
+                       exploration_strategy=action_noise,
+                       policy_optimizer=tf.train.AdamOptimizer,
+                       qf_optimizer=tf.train.AdamOptimizer)
+
+            runner.setup(algo, env)
+            last_avg_ret = runner.train(n_epochs=10,
+                                        n_epoch_cycles=20,
+                                        batch_size=250,
+                                        store_algo=False)
```
zequnyu: Do you mean `store_paths`?

tests/garage/tf/algos/test_td3.py
zequnyu: For now I think it is. I actually had some problems about #736 and want to clarify with @naeioi. Maybe I should do a separate PR to remove `n_epoch_cycles` from algos first.

tests/garage/tf/algos/test_td3.py
utkarshjp7: I added `store_algo` to train function because storing TD3 algo was giving some picking error.

zequnyu: Hmm, would you explain to me why this picking error happens?

utkarshjp7: This is the exact error I was getting `AttributeError: Can't pickle local object 'compile_function.<locals>.run'`. You can recreate this with following code
```
import pickle

def test():
    def run():
        return "Hello"
    return run

class A:
  def __init__(self):
    self._test = test()

a = A()
pickle.dump(a, open('test.pickle', 'wb'))
```

ahtsan: I think you can remove the member function in __getstate__() and restore it in __setstate()__ to make it pickleable, but that may requires more code refactoring.

utkarshjp7: @ahtsan Similar to what is done in `DDPG` ?

utkarshjp7:
@zequnyu @ahtsan I removed `store_algo` from local runner. Using `__getstate__` and `__setstate__` seems to solve the pickle error.

@ryanjulian I will merge once I get the benchmark results and everything looks okay.

src/garage/np/exploration_strategies/gaussian_strategy.py
```diff
@@ -25,8 +23,43 @@ def __init__(self,
         self._action_space = env_spec.action_space
 
     def get_action(self, t, observation, policy, **kwargs):
+        """
+        Get action from this policy for the input observation.
+
+        Args:
+            t: Iteration.
```
ryanjulian: please make sure all args and returns have types.

src/garage/experiment/local_runner.py
```diff
@@ -146,7 +146,6 @@ def save(self, epoch, paths=None):
 
         # Save states
         params['env'] = self.env
-        params['algo'] = self.algo
```
ryanjulian: uh is this here intentionally? @naeioi @zequnyu

examples/tf/td3_pendulum.py
```diff
@@ -0,0 +1,83 @@
...
+        td3 = TD3(env_spec=env.spec,
+                  policy=policy,
+                  policy_lr=1e-4,
+                  qf_lr=1e-3,
+                  qf=qf,
+                  qf2=qf2,
+                  replay_buffer=replay_buffer,
+                  target_update_tau=1e-2,
+                  n_epoch_cycles=20,
+                  n_train_steps=1,
```
ahtsan: @ryanjulian I think we have discussed it before - Garage is batch-based, sampler generates a bunch of samples in each iteration so algorithm should utilize that and optimize more than once, where another way is to generate **1** sample at a time and train once. We encountered this when I was implementing DQN and we decided to implement the algorithm in the garage-way.

If we set `n_train_steps` to 1, we don't actually need sampler anymore because that only introduces overhead and we don't want that?

src/garage/experiment/local_runner.py
utkarshjp7: oh.. I removed it by mistake. I will add it back. 

Whether it should stay or not is a separate issue I think.

tests/benchmarks/test_benchmark_td3.py
naeioi: Does it involve manually effort to install rlkit other than adding to setup.py? People might want to verify the benchmark later.

tests/garage/tf/algos/test_td3.py
naeioi: Zequn and I just discussed on this. `n_epoch_cycles` is duplicate on algorithm and runner's interface but it is required for training. We will have a PR to fix this later, but let's not put it in this PR.

examples/tf/td3_pendulum.py
naeioi: I don't get you on why sampler is not needed.

examples/tf/td3_pendulum.py
```diff
@@ -0,0 +1,83 @@
+"""
+This is an example to train a task with TD3 algorithm.
+
+Here, we create a gym environment InvertedDoublePendulum
+and use a TD3 with 1M steps.
+
+Results:
+    AverageReturn: 250
+    RiseTime: epoch 499
```
naeioi: Have you updated these numbers?

setup.py
```diff
@@ -62,6 +62,7 @@
     'flake8',
     'flake8-docstrings==1.3.0',
     'flake8-import-order',
+    'gtimer',
```
naeioi: What is this used for?

examples/tf/td3_pendulum.py
ahtsan: If we train once, why not just do env.step one time in each iteration?

setup.py
utkarshjp7: Its needed to run rlkit algorithm. Check `test_benchmark_td3.py`

utkarshjp7:
New plots after changing the exploration strategy. 

HalfCheetah-v2
Hopper-v2
InvertedDoublePendulum-v2
InvertedPendulum-v2
Swimmer-v2
Walker2d-v2

examples/tf/td3_pendulum.py
utkarshjp7: I haven't updated it. How do we get these numbers?

tests/garage/tf/algos/test_td3.py
krzentner: It would be good to remove this, so that we test that TD3 can be pickled.

src/garage/tf/algos/ddpg.py
```diff
@@ -213,6 +216,7 @@ def init_opt(self):
             self.f_update_target = f_update_target
 
     def __getstate__(self):
+        """Object.__getstate__."""
```
krzentner: These are private functions, so they shouldn't need docstrings?

src/garage/tf/algos/ddpg.py
```diff
@@ -223,10 +227,18 @@ def __getstate__(self):
         return data
 
     def __setstate__(self, state):
+        """Object.__setstate__."""
```
krzentner: Same as `__getstate__`.

src/garage/np/exploration_strategies/gaussian_strategy.py
krzentner: Yeah, at a glance it's not obvious what type t is (it could also benefit from a longer name, like `iteration`).

src/garage/np/exploration_strategies/gaussian_strategy.py
```diff
@@ -25,8 +23,43 @@ def __init__(self,
...
-                       self._action_space.low, self._action_space.high)
+                       self._action_space.low,
+                       self._action_space.high), agent_info
+
+    def get_actions(self, t, observations, policy, **kwargs):
+        """
+        Get actions from this policy for the input observation.
+
+        Args:
+            t: Iteration.
```
krzentner: Same as above.

src/garage/experiment/local_runner.py
krzentner: I think it's fine to keep this. There are a few scenarios where one might not want to save the algorithm.

src/garage/tf/algos/ddpg.py
utkarshjp7: As per `D105`, doc strings are required for magic methods.

src/garage/np/exploration_strategies/gaussian_strategy.py
utkarshjp7: Done.

src/garage/tf/algos/ddpg.py
krzentner: Huh. That feels pretty dumb. We can discuss changing that at a later date though.

utkarshjp7:
> There's still a few cleanup steps that would be good to have for this change. After that, I'll force push this past the CI if codecov still isn't responding.

Done.

Issue #33, opened 2018-06-11
Rewrite the logger


ryanjulian:
The current logger interface is okay, but the implementation is a bit messy and the TensorBoard integration is kind of a bolt-on afterthought. It's also package-global which can induce bad implementation decisions.

I'd like to rewrite the logger to be properly encapsulated as a class(es). There can still be a global singleton instance for easy access.

Ideas:
* Eliminate global scope
* First-class TensorBoard support
* Multiprocess-aware logging
* Make the logger API appropriate for all aspects of rllab (e.g. bring-up, training algo, random debugging, etc.) to avoid code peppered with print statements
* Decouple logged datapoints from output formats
* Decouple checkpointing and logging
* Sophisticated checkpoint/log destinations (e.g. remote buckets?)
* Take advantage of a minimalist logging framework rather than hand-crafting logger formats?

Imported from https://github.com/ryanjulian/rllab/issues/80

pelillian:
#464

PR #464, opened 2019-01-18
New Logger

pelillian:
Created new logger API & updated all references to use the new system.

Many commits were required to be squashed into 96afc22 for rebasing


pelillian:
#33

garage/envs/mujoco/randomization/variation.py
```diff
@@ -1,7 +1,8 @@
+# flake8: noqa
```
CatherineSue: Why skip flake8 here as well as other files?

pelillian: I had to edit almost 100 files in order to changed the logger API. I couldn't add docstrings and fix formatting errors in every single one.

pelillian: Actually I think that specific file doesn't need it! I must have edited it and then undone my edit

pelillian:
@ryanjulian 
For some reason Travis is giving me the following errors:
`E0401:  8,0: Unable to import 'theano.tensor'`
`E0401: 16,0: Unable to import 'baselines.bench'`

naeioi:
Could you give a brief on the design of new logger and also an example script to illustrate the basic usage?

CatherineSue:
> @ryanjulian
> For some reason Travis is giving me the following errors:
> `E0401: 8,0: Unable to import 'theano.tensor'`
> `E0401: 16,0: Unable to import 'baselines.bench'`
> 
> Any ideas what's going on?

I have met the similar error [here](https://github.com/rlworkgroup/garage/pull/447#discussion_r243751424). But I met this error in local pre-commit. I updated the garage env then solved the problem.

naeioi:
> > @ryanjulian
> > For some reason Travis is giving me the following errors:
> > `E0401: 8,0: Unable to import 'theano.tensor'`
> > `E0401: 16,0: Unable to import 'baselines.bench'`
> > Any ideas what's going on?
> 
> I have met the similar error [here](https://github.com/rlworkgroup/garage/pull/447#discussion_r243751424). But I met this error in local pre-commit. I updated the garage env then solved the problem.

Can you elaborate on how to update garage env? Do you mean reinstall deps in environment.yml?

CatherineSue:
> > > @ryanjulian
> > > For some reason Travis is giving me the following errors:
> > > `E0401: 8,0: Unable to import 'theano.tensor'`
> > > `E0401: 16,0: Unable to import 'baselines.bench'`
> > > Any ideas what's going on?
> > 
> > 
> > I have met the similar error [here](https://github.com/rlworkgroup/garage/pull/447#discussion_r243751424). But I met this error in local pre-commit. I updated the garage env then solved the problem.
> 
> Can you elaborate on how to update garage env? Do you mean reinstall deps in environment.yml?

``` bash
conda env update -n garage
```

examples/tf/example_tensorboard_logger.py
```diff
@@ -1,19 +1,20 @@
...
-        'uniform', key='uniform', maxval=k_val * 10)
-    logger.record_tabular("app", k_val)
-    logger.record_histogram("gass", k_val)
-    logger.dump_tensorboard(step=i)
+    # logger.record_histogram_by_type('gamma', key='gamma', alpha=k_val)
+    # logger.record_histogram_by_type(
+    #     'normal', key='normal', mean=5 * k_val, stddev=1.0)
+    # logger.record_histogram_by_type('poisson', key='poisson', lam=k_val)
+    # logger.record_histogram_by_type(
+    #     'uniform', key='uniform', maxval=k_val * 10)
```
jonashen: Don't push commented-out code

garage/envs/mujoco/maze/maze_env.py
```diff
@@ -229,7 +227,7 @@ def get_ori(self):
             obj = obj.env
         try:
             return obj.get_ori()
-        except (NotImplementedError, AttributeError) as e:
+        except (NotImplementedError, AttributeError):
             pass
```
jonashen: I don't think it's ideal to pass an exception, maybe add a description

tests/benchmarks/test_benchmark_her.py
```diff
@@ -5,16 +5,16 @@
 import unittest
 
 from baselines.bench import benchmarks
-from baselines.her.experiment.config import CACHED_ENVS
-from baselines.her.experiment.config import DEFAULT_PARAMS as BASELINES_PARAMS
+from baselines.her.experiment.config import CACHED_ENVS, \
+    DEFAULT_PARAMS as BASELINES_PARAMS
```
jonashen: use () instead of \

tests/fixtures/tf/instrumented_npo.py
```diff
@@ -3,24 +3,18 @@
...
 from garage.tf.misc import tensor_utils
-from garage.tf.misc.tensor_utils import compute_advantages
-from garage.tf.misc.tensor_utils import discounted_returns
-from garage.tf.misc.tensor_utils import filter_valids
-from garage.tf.misc.tensor_utils import filter_valids_dict
-from garage.tf.misc.tensor_utils import flatten_batch
-from garage.tf.misc.tensor_utils import flatten_batch_dict
-from garage.tf.misc.tensor_utils import flatten_inputs
-from garage.tf.misc.tensor_utils import graph_inputs
+from garage.tf.misc.tensor_utils import compute_advantages, \
```
jonashen: use () instead of \

tests/garage/tf/policies/test_gaussian_policies.py
```diff
@@ -4,17 +4,18 @@
...
 from garage.tf.envs import TfEnv
-from garage.tf.optimizers import ConjugateGradientOptimizer
-from garage.tf.optimizers import FiniteDifferenceHvp
-from garage.tf.policies import GaussianGRUPolicy
-from garage.tf.policies import GaussianLSTMPolicy
-from garage.tf.policies import GaussianMLPPolicy
+from garage.tf.optimizers import ConjugateGradientOptimizer, \
+    FiniteDifferenceHvp
+from garage.tf.policies import GaussianGRUPolicy, GaussianLSTMPolicy, \
+    GaussianMLPPolicy
```
jonashen: use () instead of \

garage/misc/autoargs.py
```diff
@@ -132,7 +132,7 @@ def wrap(func):
 
 def get_all_parameters(cls, parsed_args):
     prefix = _get_prefix(cls)
-    if prefix is None or len(prefix) == 0:
+    if prefix is None or not prefix:
```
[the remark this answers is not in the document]
pelillian: Thanks a million! What tooling do you use, if you don't mind my asking?

garage/misc/snapshotter.py
```diff
@@ -0,0 +1,72 @@
...
+import pickle
+
+
+class Snapshotter(object):
+    def __init__(self):
+        self._snapshot_dir = None
+        self._snapshot_mode = 'all'
+        self._snapshot_gap = 1
+
+    def set_snapshot_dir(self, dir_name):
```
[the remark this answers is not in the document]
pelillian: Cool, wasn't sure how much you wanted me to change the existing API

garage/misc/logger/logger_utils.py
```diff
@@ -0,0 +1,58 @@
...
+        json.dump(log_params, f, indent=2, sort_keys=True)
+
+
+def dump_variant(log_file, variant_data):
+    mkdir_p(os.path.dirname(log_file))
+    with open(log_file, "w") as f:
+        json.dump(variant_data, f, indent=2, sort_keys=True, cls=MyEncoder)
+
+
+class MyEncoder(json.JSONEncoder):
```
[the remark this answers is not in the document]
pelillian: I didn't write this code, just refactored it from what existed already. Do you want me to still change it?

garage/misc/logger/logger_utils.py
```diff
@@ -0,0 +1,58 @@
+# flake8: noqa
+from enum import Enum
+import json
+import os
+
+from garage.misc.autoargs import get_all_parameters
+from garage.misc.console import mkdir_p
+
+
+def log_parameters_lite(log_file, args):
```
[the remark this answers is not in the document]
pelillian: Not sure what this even does. Do you want me to get rid of it?

garage/misc/logger/logger_utils.py
```diff
@@ -0,0 +1,58 @@
...
+    for param_name, param_value in args.__dict__.items():
+        log_params[param_name] = param_value
+    if args.args_data is not None:
+        log_params["json_args"] = dict()
+    mkdir_p(os.path.dirname(log_file))
+    with open(log_file, "w") as f:
+        json.dump(log_params, f, indent=2, sort_keys=True, cls=MyEncoder)
+
+
+def log_parameters(log_file, args, classes):
```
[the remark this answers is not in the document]
pelillian: ditto as earlier

garage/envs/mujoco/maze/maze_env.py
pelillian: I didn't write this code. Do you want me to still edit it?

garage/misc/logger/logger_utils.py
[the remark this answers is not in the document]
pelillian: OK. I didn't want to mess with this code because it was copied from the old logger, but I'll delete `log_parameters` and renames `log_parameters_lite` to `log_parameters`

garage/misc/logger/logger_utils.py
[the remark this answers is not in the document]
pelillian: see above comment ^^

pelillian:
> Just a general question, is there a reason why we have logger under misc instead of it being standalone? i.e. garage.misc.logger instead of garage.logger?

@krzentner @ryanjulian What do you think about this?

ryanjulian:
> > Just a general question, is there a reason why we have logger under misc instead of it being standalone? i.e. garage.misc.logger instead of garage.logger?
> 
> @krzentner @ryanjulian What do you think about this?

Put it in garage.logger

ryanjulian:
> > This is a pretty large change, but mostly not scary.
> > You added a bunch of noqa, which kinda defeats the point of using `flake8`. I was going to tell you to fix them, but then I found that they mostly don't appear to be necessary.
> > I have good tooling to fix this, so I just added a commit which fixes them. Please pull it before making more changes. I didn't fix the test/ directory, since that didn't really seem as important.
> > The rest of the comments are hopefully pretty minor? Let me know if you have questions, as always.
> 
> @krzentner
> Thanks a million! What tooling do you use, if you don't mind my asking? I had tried autopep8 and a couple other tools, yet they seem to differ somewhat from what garage wants.

You have to make sure to use our configurations with automated tools. They live in setup.cfg

garage/envs/mujoco/maze/maze_env.py
ryanjulian: you can ignore it

garage/misc/logger/__init__.py
```diff
@@ -1,15 +1,15 @@
...
+from garage.misc.logger.singleton_tabular import TabularInput
+from garage.misc.logger.tensorboard_output import TensorBoardOutput
 
-if getattr(config, "LOG_TENSORBOARD", False):
-    from garage.misc.logger.tensorboard_logger import TensorboardLogger
-    logger = TensorboardLogger()
-else:
-    from garage.misc.logger.base_logger import Logger
-    logger = Logger()
+logger = Logger()
```
ryanjulian: i think you can put logger and snapshotter under the same namespace `garage.logger`

@naeioi what do you think?

garage/misc/logger/logger_outputs.py
```diff
@@ -0,0 +1,107 @@
+"""Contains the output classes for the logger.
```
ryanjulian: this file can be `outputs.py`. it's already into the logger namespace

krzentner:
> @krzentner
> Thanks a million! What tooling do you use, if you don't mind my asking? I had tried autopep8 and a couple other tools, yet they seem to differ somewhat from what garage wants.

I'm not sure if there's any fully automated tooling which works with our current linters. My workflow (which I've used across lots of environments / teams / companies) is basically:
 - Run the linter to find failures (e.g. `flake8`).
 - Filter them down to a single failure type (e.g. `flake8 --select=F401`).
 - Pipe them all into (n)vim (e.g. `flake8 --format="+%(row)s %(path)s" --select=F401 | xargs -L 1 nvim`).
 - Write a vim macro which fixes the error.
 - Spam the macro and `:x` or `:bNext` until all the files are fixed.

Obviously, a fully automated tooling solution is preferred, but this workflow has served me well for making formatting fixes and refactorings.

garage/misc/logger/logger_outputs.py
```diff
@@ -0,0 +1,107 @@
...
+import os
+import sys
+
+import dateutil.tz
+
+from garage.misc.console import colorize, mkdir_p
+from garage.misc.logger.singleton_tabular import TabularInput
+
+
+class LoggerOutput(ABC):
```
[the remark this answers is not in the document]
pelillian: I thought it might be confusing with the log_output method, but I'll change it!

garage/misc/tabulate.py
```diff
@@ -57,8 +57,8 @@
 
 
 def _pipe_segment_with_colons(align, colwidth):
-    # Return a segment of a horizontal line with optional colons which
-    # indicate column's alignment (as in `pipe` output format).
+    """Return a segment of a horizontal line with optional colons which
```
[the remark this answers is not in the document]
pelillian: @ryanjulian Thanks for pointing this out! I didn't want to mess with this too much, as it seemed to work.

scripts/run_experiment.py
```diff
@@ -197,10 +197,10 @@ def run_experiment(argv):
                 for _ in maybe_iter:
                     pass
 
-    logger.set_snapshot_mode(prev_mode)
-    logger.set_snapshot_dir(prev_snapshot_dir)
-    logger.remove_tabular_output(tabular_log_file)
-    logger.remove_text_output(text_log_file)
+    snapshotter.snapshot_mode = prev_mode
+    snapshotter.snapshot_dir = prev_snapshot_dir
+    logger.remove_output(CsvOutput)
```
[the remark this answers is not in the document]
pelillian: @ryanjulian I'm using a singleton because you told me to. Would you rather me do it that way?

garage/misc/snapshotter.py
```diff
@@ -0,0 +1,88 @@
...
+    @property
+    def snapshot_gap(self, ):
+        """The number of iterations between snapshots."""
+        return self._snapshot_gap
+
+    @snapshot_gap.setter
+    def snapshot_gap(self, gap):
+        self._snapshot_gap = gap
+
+    def save_extra_data(self, data, file_name='extra_data.pkl', mode='joblib'):
```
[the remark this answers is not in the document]
pelillian: In run_experiment

garage/misc/snapshotter.py
```diff
@@ -0,0 +1,88 @@
+"""Allows the taking of snapshots."""
+from os import path as osp
+import pickle
+
+
+class Snapshotter(object):
+    """This singleton handles the creation of snapshots."""
```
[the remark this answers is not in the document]
pelillian: It is a singleton, we talked about this earlier

garage/misc/logger/singleton_logger.py
```diff
@@ -0,0 +1,99 @@
...
+        if not self._outputs:
+            warn('No outputs have been added to the logger.')
+
+        kwargs['prefix'] = self._prefix_str
+        if 'with_prefix' in kwargs and not kwargs['with_prefix']:
+            kwargs['prefix'] = ''
+
+        for output in self._outputs:
+            if isinstance(data, output.accept_types):
+                output.log_output(data, **kwargs)
```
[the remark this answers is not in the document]
pelillian: I did this a bit above

garage/logger/snapshotter.py
```diff
@@ -0,0 +1,88 @@
+"""Allows the taking of snapshots."""
+from os import path as osp
+import pickle
+
+
+class Snapshotter(object):
+    """This singleton handles the creation of snapshots."""
```
ryanjulian: https://en.wikipedia.org/wiki/Singleton_pattern

> ...a software design pattern that __restricts the instantiation of a class to one__.

This object can be instantiated as many times as I want.

garage/logger/snapshotter.py
```diff
@@ -0,0 +1,88 @@
+"""Allows the taking of snapshots."""
+from os import path as osp
+import pickle
+
+
+class Snapshotter(object):
```
ryanjulian: No need to inherit from object in Python 3

garage/logger/singleton_tabular.py
```diff
@@ -0,0 +1,91 @@
+"""Contains the singleton tabular class.
```
ryanjulian: It's not a singleton, it's just a class. I can make as many of them as I want.

garage/logger/singleton_logger.py
```diff
@@ -0,0 +1,99 @@
+"""This is Garage's logger singleton.
+
+It takes in many different types of input and directs them to the correct
+output.
+"""
+from contextlib import contextmanager
+from warnings import warn
+
+
+class Logger():
```
ryanjulian:
```python
class Logger:
    ...
```

garage/logger/singleton_logger.py
```diff
@@ -0,0 +1,99 @@
+"""This is Garage's logger singleton.
```
ryanjulian: Can you provide a summary here of the principles of operation.

garage/logger/outputs.py
```diff
@@ -0,0 +1,121 @@
...
+class LogOutput(abc.ABC):
+    """Abstract class for Logger Outputs."""
+
+    @property
+    def types_accepted(self):
+        """The types that the logger may pass to this output."""
+        return ()
+
+    @abc.abstractmethod
+    def log_output(self, data, **kwargs):
```
ryanjulian: Perhaps there's a better name for this API? record? sink? 

@naeioi @krzentner

garage/logger/outputs.py
```diff
@@ -0,0 +1,121 @@
...
+
+class StdOutput(LogOutput):
+    """Standard console output for the logger."""
+
+    @property
+    def types_accepted(self):
+        """The types that the logger may pass to this output."""
+        return str, TabularInput
+
+    def log_output(self,
```
ryanjulian: I'm concerned by this API.

It seems that you are silently accepting args from the upstream `Logger.log()` calls?

@naeioi @krzentner  PTAL

garage/logger/singleton_logger.py
```diff
@@ -0,0 +1,99 @@
...
+        self.add_output(output)
+
+    def has_output(self, output_type):
+        """Checks to see if a given logger output is attached to the logger."""
+        for output in self._outputs:
+            if isinstance(output, output_type):
+                return True
+        return False
+
+    def get_output(self, output_type):
```
ryanjulian: Please don't clobber me previous comments about how you address the log outputs here. It seems strange to address them on type which I could completely validly want 2 CSV outputs.

ryanjulian: @krzentner @naeioi

pelillian:
@ryanjulian I'm in the process of fixing everything right now. I didn't just clobber your comments.

ryanjulian:
Okay thanks. Generally you wait for the commenter to resolve comments on your PR, not resolve them yourself.

garage/logger/singleton_logger.py
pelillian: I have removed this method in an upcoming commit

garage/logger/singleton_logger.py
pelillian: Fixed

garage/logger/singleton_tabular.py
pelillian: Fixed

garage/logger/snapshotter.py
pelillian: Fixed

garage/logger/outputs.py
```diff
@@ -0,0 +1,121 @@
...
+        """The types that the logger may pass to this output."""
+        return ()
+
+    @abc.abstractmethod
+    def log_output(self, data, **kwargs):
+        """This method is called by the logger when it needs to pass data."""
+        pass
+
+    def dump(self, step=None):
+        """This method is called by the logger to dump an output to file."""
```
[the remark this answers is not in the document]
pelillian: Fixed

garage/logger/snapshotter.py
pelillian: Fixed

garage/logger/__init__.py
```diff
@@ -0,0 +1,18 @@
+"""Global logger module.
```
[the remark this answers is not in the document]
pelillian: Fixed

garage/misc/logger/singleton_logger.py
```diff
@@ -0,0 +1,99 @@
...
+class Logger(object):
+    """This is the singleton class that handles logging."""
+
+    def __init__(self):
+        self._outputs = []
+        self._prefixes = []
+        self._prefix_str = ''
+
+    def log(self, data, **kwargs):
+        """Magic method that takes in all different types of input."""
```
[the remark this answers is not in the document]
pelillian: Fixed

garage/__init__.py
```diff
@@ -0,0 +1 @@
+# flake8: noqa
```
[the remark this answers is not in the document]
naeioi: Yes. I submitted a PR to add `__init__.py` in all directories. Please rebase after it's merged onto master.

garage/logger/outputs.py
naeioi: Why not just log()?

garage/logger/outputs.py
naeioi: Better remove `**kwargs` so that `log_output` rejects unidentified args.

examples/tf/example_tensorboard_logger.py
```diff
@@ -1,19 +1,13 @@
...
-    logger.record_histogram_by_type('gamma', key='gamma', alpha=k_val)
-    logger.record_histogram_by_type(
-        'normal', key='normal', mean=5 * k_val, stddev=1.0)
-    logger.record_histogram_by_type('poisson', key='poisson', lam=k_val)
-    logger.record_histogram_by_type(
-        'uniform', key='uniform', maxval=k_val * 10)
-    logger.record_tabular("app", k_val)
-    logger.record_histogram("gass", k_val)
-    logger.dump_tensorboard(step=i)
+    tabular.record("app", k_val)
```
naeioi: This tabular goes nowhere, neither logged to screen or dumped to file. Maybe log it using `logger.log()` so its usage is better illustrated.

examples/tf/example_tensorboard_logger.py
```diff
@@ -1,19 +1,13 @@
...
+    logger.log(("gass", k_val), record='histogram')
```
naeioi: The `record` parameter is specific to `TensorBoardOutput` but not in base logger interface. I think we should not pass ad hoc configuration to `Output` when calling `logger.log()` - all output options should be configured through only `output_instance` but not logger interface.

In this particular case, this `record` parameter is instructing logger to log as histogram. What about using a `Histogram` class to wrap up data, just like using `Tabular` to log tabular.

Also, instantiating helper wrapper every time can be pretty annoying and cumbersome. I also suggest add sugar methods like `logger.scalar()`, `.histogram()`, `.text()` and `.tabular()` to logger.

naeioi: `tf.summary` as a ref for my last point: https://www.tensorflow.org/api_docs/python/tf/summary/histogram

garage/logger/outputs.py
naeioi: I don't have a preference. `log_output` looks good to me.

tests/garage/tf/algos/test_ddpg.py
```diff
@@ -18,7 +18,7 @@
 class TestDDPG(TfGraphTestCase):
     def test_ddpg_pendulum(self):
         """Test PPO with Pendulum environment."""
-        logger.reset()
+        logger._tensorboard = TensorBoardOutput()
```
naeioi: `logger.reset_output(TensorBoardOutput())`

ryanjulian:
FTFY there are now 2 commits in the PR:
* Actual new logger files
* Refactors to use the new logger

ryanjulian:
@naeioi @krzentner this should make it much easier to review changes. you can choose to view changes from just one of the commits in the Github review interface.

garage/logger/outputs.py
```diff
@@ -0,0 +1,121 @@
...
+
+from garage.logger.tabular_input import TabularInput
+from garage.misc.console import colorize, mkdir_p
+
+
+class LogOutput(abc.ABC):
+    """Abstract class for Logger Outputs."""
+
+    @property
+    def types_accepted(self):
```
ryanjulian: Hmm `types_accepted()` or `input_types_accepted()`? @naeioi @krzentner

garage/logger/outputs.py
ryanjulian: @pelillian @naeioi @krzentner 

Let's look at the vocabulary for the API so that we can make it consistent/easy for novices to reason about.

* The logger has _inputs_ (sources) and _outputs_ (sinks)
* Between the _inputs_ and _outputs_ sits, the _logger_ which mediates flow between these two sides
* A user _logs_ data by calling `logger.log()` with an _input type_. Most _inputs_ are Python primitives like `string`, but there are also special cases (e.g. `tabular`, probably some stuff for `numpy.ndarrays` later)
* _outputs_ subscribe to _input types_ they _accept_ by declaring a property which returns a collection of Python _input types_ they can _accept_
* the _logger_ tell the _outputs_ when to _dump_ their data to their backing stores (e.g. flush or record)

At issue here: what happens in between _log_ and _dump_, when the _logger_ passes an _input_ to an _output_?
Suggestions:
* `record()` 
* `accept()`
* `put()`
* `post()`

I am slightly partial to `put()` or `record()`

Another question: _dump_ or _flush_? Generally `dump()` implies a one-time write-to-backing, whereas `flush()` implies an incremental update. `commit()` is another option, but might imply atomicity we can't actually live up to.

garage/logger/outputs.py
ryanjulian: should users even be able to propagate arbitrary args/kwargs from `Logger.log()` to `LogOutput.log_output()`?

garage/logger/outputs.py
```diff
@@ -0,0 +1,121 @@
...
+    """Text file output for logger."""
+
+    def __init__(self, file_name):
+        mkdir_p(os.path.dirname(file_name))
+        self._log_file = open(file_name, 'a')
+
+    @property
+    def types_accepted(self):
+        """The types that the logger may pass to this output."""
+        return str,  # we need the comma here so the return value is iterable
```
ryanjulian:
```python
return (str, )
```
?

garage/logger/outputs.py
```diff
@@ -0,0 +1,121 @@
...
+    def __init__(self, file_name):
+        mkdir_p(os.path.dirname(file_name))
+        self._log_file = open(file_name, 'w')
+
+        self._tabular_header_written = False
+
+    @property
+    def types_accepted(self):
+        """The types that the logger may pass to this output."""
+        return TabularInput,  # we need the comma here so the return value
```
ryanjulian:
```python
return (TabularInput, )
```
would probably be a lot less surprising to future maintainers

garage/logger/outputs.py
```diff
@@ -0,0 +1,121 @@
...
+            now = datetime.datetime.now(dateutil.tz.tzlocal())
+            timestamp = now.strftime('%Y-%m-%d %H:%M:%S.%f %Z')
+            out = "%s | %s" % (timestamp, out)
+
+        self._log_file.write(out + '\n')
+        self._log_file.flush()
+
+
+class CsvOutput(LogOutput):
+    """CSV file output for logger."""
```
ryanjulian: Please document constructor arguments in the class docstring

garage/logger/outputs.py
```diff
@@ -0,0 +1,121 @@
...
+            out = str(data)
+        else:
+            return
+
+        print(out)
+        sys.stdout.flush()
+
+
+class TextOutput(LogOutput):
+    """Text file output for logger."""
```
ryanjulian: Please document constructor arguments in the class docstring

garage/logger/outputs.py
```diff
@@ -0,0 +1,121 @@
...
+        elif isinstance(data, TabularInput):
+            out = str(data)
+        else:
+            return
+
+        print(out)
+        sys.stdout.flush()
+
+
+class TextOutput(LogOutput):
```
ryanjulian: I think you can merge much of `TextOutput` and `StdOutput` by inheriting them from a parent class for all `TextIO`. stdout and files are both just file descriptors in text mode.

```python
import io
import sys

class TextIOOutput:

    def __init__(self, output):
        assert isinstance(output, io.TextIOBase)
        self._output = output

    @property
    def types_accepted(self):
        return (str, )
    
    def log_output(self, data, ...):
        # ... stuff about time codes and prefixes to build out
        self._output.write("{}\n".format(out))
        self._output.flush()

    def dump(self, step=None):
        self._output.flush()


class StdOutOutput(TextIOOutput):

    def __init__(self):
        super().__init__(sys.stdout)


class TextFileOutput(TextIOOutput):

    def __init__(self):
        # ... stuff about opening a file called self._log_file
        super().__init__(self._log_file)
```

garage/logger/outputs.py
```diff
@@ -0,0 +1,121 @@
...
+        """The types that the logger may pass to this output."""
+        return ()
+
+    @abc.abstractmethod
+    def log_output(self, data, **kwargs):
+        """This method is called by the logger when it needs to pass data."""
+        pass
+
+    def dump(self, step=None):
+        """This method is called by the logger to dump an output."""
```
ryanjulian: please document all arguments in the docstring

garage/logger/snapshotter.py
```diff
@@ -0,0 +1,90 @@
...
+class Snapshotter:
+    """This class handles the creation of snapshots."""
+
+    def __init__(self):
+        self._snapshot_dir = None
+        self._snapshot_mode = 'all'
+        self._snapshot_gap = 1
+
+    @property
+    def snapshot_dir(self, ):
```
ryanjulian: hmm does this really need to be changed throughout the lifecycle of a snapshotter? why not just force users to make a new Snapshotter for a new directory? Then we can put this logic in the constructor

garage/logger/snapshotter.py
```diff
@@ -0,0 +1,90 @@
...
+    def snapshot_mode(self, mode):
+        self._snapshot_mode = mode
+
+    @property
+    def snapshot_gap(self, ):
+        """The number of iterations between snapshots."""
+        return self._snapshot_gap
+
+    @snapshot_gap.setter
+    def snapshot_gap(self, gap):
```
ryanjulian: Ditto for all these properties. What if we just make them constructor parameters?

@naeioi @krzentner

garage/logger/snapshotter.py
```diff
@@ -0,0 +1,90 @@
...
+        self._snapshot_gap = gap
+
+    def save_extra_data(self, data, file_name='extra_data.pkl', mode='joblib'):
```
ryanjulian: I think we should delete this API. There's already an interface for saving stuff to a snapshot--it's below. Just add it there.

garage/logger/snapshotter.py
```diff
@@ -0,0 +1,90 @@
...
+                joblib.dump(data, file_name, compress=3)
+            elif mode == 'pickle':
+                pickle.dump(data, open(file_name, "wb"))
+            else:
+                raise ValueError("Invalid mode: {}".format(mode))
+            return file_name
+
+        return None
+
+    def save_itr_params(self, itr, params):
```
ryanjulian: API name options:
* `save()`
* `capture()`
* `commit()`
* `put()`
* `record()`
* `write()`

@naeioi @krzentner

garage/logger/tabular_input.py
```diff
@@ -0,0 +1,91 @@
...
+This class allows the recording of tabular information for later output.
+TabularInput may be passed to the logger via its log() method.
+"""
+from contextlib import contextmanager
+
+import numpy as np
+import tabulate
+
+
+class TabularInput(object):
```
ryanjulian: No need to inherit from object in Python 3

garage/logger/tensorboard_output.py
```diff
@@ -0,0 +1,272 @@
...
+import tensorflow as tf
+
+from garage import config
+from garage.logger.outputs import LogOutput
+from garage.logger.tabular_input import TabularInput
+from garage.misc.console import mkdir_p
+
+
+class TensorBoardOutput(LogOutput):
+    def __init__(self, log_dir=config.LOG_DIR):
```
ryanjulian: Please document constructor parameters in the class docstring

garage/logger/tensorboard_output.py
```diff
@@ -0,0 +1,272 @@
...
+    def log_output(self, data, **kwargs):
+        if isinstance(data, tf.Tensor):
+            self.record_tensor(data.name, data)
+        elif isinstance(data, TabularInput):
+            for key, value in data.get_table_dict().items():
+                self.record_scalar(key, value)
+        elif isinstance(data, tuple) and kwargs['record'] == 'histogram':
+            self.record_histogram(data[0], data[1])
+
+    def set_dir(self, dir_name):
```
ryanjulian: Hmm this API is strange. Why make this mutable? We can just force the user to make a new TensorBoardOutput. That makes the log much simpler.

@naeioi @krzentner

garage/logger/tensorboard_output.py
```diff
@@ -0,0 +1,272 @@
...
+from tensorboard.plugins.custom_scalar import layout_pb2, metadata
+import tensorflow as tf
+
+from garage import config
+from garage.logger.outputs import LogOutput
+from garage.logger.tabular_input import TabularInput
+from garage.misc.console import mkdir_p
+
+
+class TensorBoardOutput(LogOutput):
```
ryanjulian: This might be easily replaced by a wrapper around tensorboardX

https://github.com/lanpa/tensorboardX

examples/tf/example_tensorboard_logger.py
ryanjulian: i agree tho we can keep that as an issue for future enhancement. let's focus on the API and design fundamentals in this PR.

garage/logger/logger.py
```diff
@@ -0,0 +1,107 @@
...
+            kwargs['prefix'] = self._prefix_str
+
+        at_least_one_logged = False
+        for output in self._outputs:
+            if isinstance(data, output.types_accepted):
+                output.log_output(data, **kwargs)
+                at_least_one_logged = True
+
+        if not at_least_one_logged:
+            warn("Log data was not accepted by any output")
```
ryanjulian: include the type?

garage/logger/tensorboard_output.py
ryanjulian: (this can be in a future PR)

garage/logger/__init__.py
```diff
@@ -0,0 +1,18 @@
+"""Logger module.
+
+This module instantiates a global logger singleton.
+"""
+from garage.logger.logger import Logger
+from garage.logger.outputs import CsvOutput, StdOutput, TextOutput
+from garage.logger.snapshotter import Snapshotter
```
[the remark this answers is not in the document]
naeioi: Let's it leave it there in this PR. The snapshotter should be replaced by checkpointer in exp runner.

examples/tf/example_tensorboard_logger.py
pelillian: Sorry, had spent all day editing everything to the new API. Changing this now

examples/tf/example_tensorboard_logger.py
pelillian: @naeioi I like this idea! Definitely something that can be added later

tests/garage/tf/algos/test_ddpg.py
pelillian:
```suggestion
        logger.reset_output(TensorBoardOutput())
```

garage/logger/singleton_logger.py
pelillian: @krzentner @ryanjulian What do you mean by "principles of operation?"

garage/logger/outputs.py
pelillian: The only use right now for this is so that we can log histograms to `TensorBoardOutput`. If we added the `Histogram` class like @naeioi mentioned we could get rid of this.

pelillian: Maybe that's the easiest solution?

garage/logger/__init__.py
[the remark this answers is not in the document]
pelillian: I was told to put this in here @ryanjulian

garage/logger/outputs.py
pelillian: I removed the parentheses for redundancy; do you think I should put it back in?

garage/logger/outputs.py
pelillian: Fixed

garage/logger/logger.py
pelillian: Fixed

garage/logger/tensorboard_output.py
pelillian: I checked in a bunch of the other code and it's been done in the __init__ docstring. Should I do it there for consistency or put it in the class docstring?

garage/logger/tabular_input.py
pelillian: Fixed

garage/logger/outputs.py
krzentner: I think it's fine as is.

garage/logger/__init__.py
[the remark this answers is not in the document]
krzentner: This seems like a change we can make outside this PR.

garage/logger/snapshotter.py
krzentner: `save()`, `record()` and `write()` all seem reasonable. `capture()` is weird, since it usually implies **active** observation, and `commit()` implies automicity. `save()` is the one there now, and the top of the list, so let's just use that?

Honestly, this seems like a name change we could land in a separate PR, if it matters.

garage/logger/outputs.py
krzentner: I don't have a strong preference. This method won't be defined or called much compared to the number of `log` calls. `log_output` is a fairly obvious name, so I don't really see a strong reason to change it. `record()` would also be fine, `put()` seems not so great, since it's often associated with networking or queuing (neither of which is happening here). `post()` and `accept()` have the same issues, but to a larger degree. Methods named `flush()` typically don't take parameters, and are generally associated with buffering, which isn't really the behavior here.

If the motivating reason is to have a different verb to describe the output's operation than the logger itself, then `record()` seems fine to me. If we don't think that's a major concern (and it doesn't seem that major to me), then `log_output` seems fine.

garage/logger/snapshotter.py
krzentner: `save()`, `record()` and `write()` all seem reasonable. `capture()` is weird, since it usually implies **active** observation, and `commit()` implies automicity. `save()` is the one there now, and the top of the list, so let's just use that?

Honestly, this seems like a name change we could land in a separate PR, if it matters.

garage/envs/mujoco/hopper_env.py
```diff
@@ -3,8 +3,8 @@
 from garage.core import Serializable
 from garage.envs import Step
 from garage.envs.mujoco import MujocoEnv
+from garage.logger import tabular
```
ryanjulian: AFAICT there is no call to `logger.log` in here

garage/logger/tensorboard_output.py
ryanjulian: class docstring

garage/envs/mujoco/hopper_env.py
pelillian: The file doesn't use the logger but the `log diagnostics` method logs to the tabular.

pelillian: The way it was implemented it was the responsibility of the model to dump the tabular, so I tried to keep it that way. Is that the right way to go?

ryanjulian: hmm i see. you're right. this is more a design flaw of the env than anything (implicitly sharing global data)

garage/logger/logger.py
```diff
@@ -0,0 +1,279 @@
...
++------+      +---------+
+|logger|
++------+      +----------+
+       +------>TextOutput|
+              +----------+
+
+# And another output.
+
+from garage.logger import CsvOutput
+logger.add_output(CsvOutput("log_folder/table.csv"
```
ryanjulian: missing some )) here?

garage/logger/logger.py
```diff
@@ -0,0 +1,279 @@
+"""This is Garage's logger.
+
+It takes in many different types of input and directs them to the correct
+output.
+
+The logger has 4 major steps:
```
ryanjulian: This is really great. Thank you.

garage/logger/logger.py
```diff
@@ -0,0 +1,279 @@
...
+    def add_output(self, output):
+        """Add a new output to the logger.
+
+        All data that is compatible with this output will be sent there.
+
+        :param output: An instantiation of a LogOutput subclass to be added.
+        """
+        self._outputs.append(output)
+
+    def remove_all(self):
```
ryanjulian: `remove_all_outputs()` (to be consistent with `add_output()`

garage/tf/algos/batch_polopt.py
```diff
@@ -132,11 +131,11 @@ def train(self, sess=None):
...
                     params["paths"] = samples_data["paths"]
-                logger.save_itr_params(itr, params)
+                snapshotter.save_itr_params(itr, params)
                 logger.log("Saved")
-                logger.record_tabular('Time', time.time() - start_time)
-                logger.record_tabular('ItrTime', time.time() - itr_start_time)
-                logger.dump_tabular(with_prefix=False)
+                tabular.record('Time', time.time() - start_time)
+                tabular.record('ItrTime', time.time() - itr_start_time)
+                logger.log(tabular)
```
ryanjulian: do we need to add a call to `dump()` here?

garage/tf/algos/npo.py
```diff
@@ -1,23 +1,17 @@
...
 from garage.tf.misc import tensor_utils
-from garage.tf.misc.tensor_utils import compute_advantages
-from garage.tf.misc.tensor_utils import discounted_returns
-from garage.tf.misc.tensor_utils import filter_valids
-from garage.tf.misc.tensor_utils import filter_valids_dict
-from garage.tf.misc.tensor_utils import flatten_batch
-from garage.tf.misc.tensor_utils import flatten_batch_dict
-from garage.tf.misc.tensor_utils import flatten_inputs
-from garage.tf.misc.tensor_utils import graph_inputs
+from garage.tf.misc.tensor_utils import (
```
ryanjulian: please don't replace the multi-line version unless the formatter is forcing you to

tests/benchmarks/test_benchmark_ddpg.py
```diff
@@ -27,8 +27,8 @@
 import pandas as pd
 import tensorflow as tf
 
+from garage.logger import CsvOutput, logger as garage_logger, TensorBoardOutput
```
ryanjulian: please don't contract complicated import lines like this
it would be more readable as
```python
from garage.logger import logger  # no need to rename
from garage.logger import CsvOutput, TensorBoardOutput
```

tests/benchmarks/test_benchmark_her.py
```diff
@@ -5,16 +5,16 @@
 import unittest
 
 from baselines.bench import benchmarks
-from baselines.her.experiment.config import CACHED_ENVS
-from baselines.her.experiment.config import DEFAULT_PARAMS as BASELINES_PARAMS
+from baselines.her.experiment.config import (CACHED_ENVS, DEFAULT_PARAMS as
```
ryanjulian: please don't contract import lines like this

tests/benchmarks/test_benchmark_ppo.py
```diff
@@ -11,8 +11,7 @@
 import random
 import unittest
 
-from baselines import bench
-from baselines import logger as baselines_logger
+from baselines import bench, logger as baselines_logger
```
ryanjulian: please don't contract your imports like this

tests/benchmarks/test_benchmark_ppo.py
```diff
@@ -26,8 +25,8 @@
 import tensorflow as tf
 
 from garage.envs import normalize
+from garage.logger import CsvOutput, logger as garage_logger, TensorBoardOutput
```
ryanjulian: import contractions

ryanjulian:
Looking really good.

IMO the contractions of many import lines reduces readability in a lot of places. Please don't do this in edited files (I assume your editor may have done it automatically?). For new files, use your best judgement.

tests/garage/logger/test_logger.py
```diff
@@ -0,0 +1,94 @@
+import unittest
+
+from garage.logger import CsvOutput, logger, tabular, TextOutput
```
ryanjulian: a simpler way to organize these tests is as unit tests rather than integration tests.

i.e. you can write a small test for each of your `*Output` classes which verifies it produces the correct output given an input from the logger, which you manually construct rather than using the `Logger` to produce. then you can write a single small test for `Logger` using a single output (or a dummy output) which verifies simple things (e.g. add/remove, passes prefixes correctly etc.) but doesn't have to examine any output.

ryanjulian: you might find the library `unittest.mock` helpful

tests/garage/logger/test_snapshotter.py
```diff
@@ -0,0 +1,69 @@
...
+import pickle
+import random
+import shutil
+import unittest
+
+from garage.logger import snapshotter
+from garage.misc.console import mkdir_p
+
+
+class TestSnapshotter(unittest.TestCase):
```
ryanjulian: Thanks for the test! Though I thought we agreed you didn't have to write tests for the snapshotter?

pelillian: Lol I refactored a bit so I just did it; didn't take too long

tests/benchmarks/test_benchmark_ddpg.py
ryanjulian: ping

garage/logger/tabular_input.py
```diff
@@ -0,0 +1,125 @@
...
+
+    @property
+    def primitive_dict(self):
+        """Return the dictionary, excluding all nonprimitive types."""
+        primitives = (int, float, str, bool)
+
+        def is_primitive(x):
+            if isinstance(x, primitives) or isinstance(x, numbers.Number):
+                return True
+            elif type(x).__module__ == np.__name__:
```
ryanjulian: not covered by tests

garage/logger/tabular_input.py
```diff
@@ -0,0 +1,125 @@
...
+
+    @property
+    def dict(self):
+        """Return a dictionary of the tabular items."""
+        return self._dict
+
+    @property
+    def key_set(self):
+        """Return a set of the table's keys."""
+        return set(self._dict.keys())
```
ryanjulian: not covered by tests

garage/logger/tensorboard_output.py
```diff
@@ -38,7 +49,26 @@ def __init__(self):
...
+        self._set_dir(log_dir)
+
+    @property
+    def types_accepted(self):
+        """The types that the logger may pass to this output."""
+        return (TabularInput, )
+
+    def record(self, data, prefix=''):
+        if isinstance(data, TabularInput):
+            for key, value in data.dict.items():
```
ryanjulian: a few of these branches are never tested

garage/logger/tensorboard_output.py
```diff
@@ -38,7 +49,26 @@ def __init__(self):
...
+                if isinstance(value, HistogramInput):
+                    self.record_histogram(key, value.data)
+                elif isinstance(value, HistogramInputDistribution):
+                    self.record_histogram_by_type(**vars(value))
+                elif isinstance(value, tf.Tensor):
+                    self.record_tensor(key, value)
+                else:
+                    self.record_scalar(key, value)
+
+    def _set_dir(self, dir_name):
```
ryanjulian: not fully tested

garage/logger/tensorboard_output.py
```diff
@@ -54,7 +84,11 @@ def set_dir(self, dir_name):
             self._default_step = 0
             assert self._writer is not None
 
-    def dump_tensorboard(self, step=None):
+    def dump(self, step=None):
+        """Dumps this output to the log dir.
+
+        :param step: The current run step.
+        """
         if not self._writer:
```
ryanjulian: not tested

garage/logger/logger.py
```diff
@@ -0,0 +1,293 @@
...
+        return None
+
+    def add_output(self, output):
+        """Add a new output to the logger.
+
+        All data that is compatible with this output will be sent there.
+
+        :param output: An instantiation of a LogOutput subclass to be added.
+        """
+        if isinstance(output, type):
```
ryanjulian: it looks like these errors are untested

garage/logger/logger.py
```diff
@@ -0,0 +1,293 @@
...
+        """Check to see if a given logger output is attached to the logger.
+
+        :param output_type: A LogOutput subclass type to be checked for.
+        """
+        for output in self._outputs:
+            if isinstance(output, output_type):
+                return True
+        return False
+
+    def dump_output_type(self, output_type, step=None):
```
ryanjulian: never visited during testing

garage/logger/logger.py
```diff
@@ -0,0 +1,293 @@
...
+        """Dump all outputs of the given type.
+
+        :param output_type: A LogOutput subclass type to be dumped.
+        :param step: The current run step.
+        """
+        for output in self._outputs:
+            if isinstance(output, output_type):
+                output.dump(step=step)
+
+    def dump_all(self, step=None):
```
ryanjulian: never visited during testing

garage/logger/outputs.py
```diff
@@ -0,0 +1,178 @@
...
+
+    @property
+    def types_accepted(self):
+        """Pass these types to this logger output.
+
+        The types in this tuple will be accepted by this output.
+
+        :return: A tuple containing all valid input types.
+        """
+        return ()
```
ryanjulian: unimplemented ABC methods usually `pass`. do you need this return type for a reason?

garage/logger/outputs.py
```diff
@@ -0,0 +1,178 @@
...
+    def types_accepted(self):
+        """Accept all output types."""
+        return (object, )
+
+    def record(self, data, prefix=''):
+        """Don't do anything."""
+        pass
+
+
+class StdOutput(LogOutput):
```
ryanjulian: it looks like this class is completely uncovered during testing?

garage/logger/outputs.py
```diff
@@ -0,0 +1,178 @@
...
+        elif isinstance(data, TabularInput):
+            out = str(data)
+        else:
+            return
+
+        print(out)
+        sys.stdout.flush()
+
+
+class FileOutput(LogOutput):
```
ryanjulian: if you never intend for anyone to instantiate this, use `class FileOutput(LogOutput, metaclass=abc.ABCMeta)`

garage/logger/outputs.py
```diff
@@ -0,0 +1,178 @@
...
+
+    def __init__(self, file_name, mode='w'):
+        mkdir_p(os.path.dirname(file_name))
+        self._log_file = open(file_name,
+                              mode)  # Open the log file in child class
+
+    @property
+    def types_accepted(self):
+        """Pass these types to this logger output."""
+        return ()
```
ryanjulian: pass?

garage/logger/outputs.py
```diff
@@ -0,0 +1,178 @@
...
+
+    def close(self):
+        """Close any files used by the output."""
+        if self._log_file and not self._log_file.closed:
```
ryanjulian: one of these branches is uncovered

garage/logger/outputs.py
```diff
@@ -0,0 +1,178 @@
...
+
+    @property
+    def types_accepted(self):
+        """Accept str objects only."""
+        return (str, )
+
+    def record(self, data, prefix=''):
+        """Log data to text file."""
+        out = data
+        if self._with_timestamp:
```
ryanjulian: apparently this doesn't happen during testing?

garage/logger/outputs.py
```diff
@@ -0,0 +1,178 @@
...
+        return (TabularInput, )
+
+    def record(self, data, prefix=''):
+        """Log tabular data to CSV."""
+        dictionary = data.primitive_dict
+
+        writer = csv.DictWriter(
+            self._log_file, fieldnames=set(dictionary.keys()))
+
+        if not self._tabular_header_written:
```
ryanjulian: this seems uncovered in testing

garage/logger/tensorboard_inputs.py
```diff
@@ -0,0 +1,76 @@
+"""Contains input classes for tensorboard.
+
+Tensorboard accepts different types of inputs, so they are defined here to keep
+tensorboard separate from the rest of the logger.
+"""
+
+
+class HistogramInput:
```
ryanjulian: these are naked data classes, so they generally don't get unit tests. however, it's expected that they get covered through usage in other unit tests.

ryanjulian:
It is going to be really hard to merge this without class-by-class unit tests. New code is the perfect opportunity to add real unit tests, because everything every class should do is all in your head right now.

Incidental coverage from other tests is not nearly as valuable as unit tests.

I think you can make them mostly by reusing the tests you already have.

Try running coverage with just your tests:
```sh
coverage run --source=garage/logger -m nose2 tests.garage.logger
coverage report
coverage html
xdg-open htmlcov/index.html
```

Output:
```sh
........
Ran 8 tests in 0.010s

OK
Name                                  Stmts   Miss Branch BrPart  Cover
-----------------------------------------------------------------------
garage/logger/__init__.py                16      0      0      0   100%
garage/logger/logger.py                  69     17     26      3    73%
garage/logger/outputs.py                 80     20     12      3    68%
garage/logger/snapshotter.py             45      0     18      1    98%
garage/logger/tabular_input.py           59      6     10      1    90%
garage/logger/tensorboard_inputs.py      27     13      0      0    52%
garage/logger/tensorboard_output.py     169    140     60      0    13%
-----------------------------------------------------------------------
TOTAL                                   465    196    126      8    53%
```

One-liner for rapid development:
```sh
coverage run --source=garage/logger -m nose2 tests.garage.logger && coverage report && coverage html && xdg-open htmlcov/index.html 
```

garage/logger/outputs.py
pelillian: The way I was thinking was that if it's unimplemented it returns an empty tuple. I'll change it to pass though

pelillian: Ok, it doesn't want me to `pass` because it's a getter for a property

garage/logger/outputs.py
pelillian: I totally forgot I could just redirect standard out to test this

tests/garage/logger/test_tensorboard_logger.py
```diff
@@ -0,0 +1,66 @@
...
+from garage.logger import HistogramInput, HistogramInputGamma
+from garage.logger import HistogramInputNormal, HistogramInputPoisson
+from garage.logger import HistogramInputUniform, TensorBoardOutput
+from garage.logger import logger, tabular
+from garage.misc.console import remove_dir_if_exists
+
+
+class TestTensorBoardLogger(unittest.TestCase):
+    def test_histogram(self):
+        log_dir = 'test'
```
ryanjulian: these 2 lines in `def setUp()`

tests/garage/logger/test_tensorboard_logger.py
```diff
@@ -0,0 +1,66 @@
...
+            logger.dump_all()
+
+            tabular.clear()
+            tabular.record("foo", foo + 1)
+            tabular.record("bar", bar + 1)
+            logger.log(tabular)
+            logger.dump_output_type(TensorBoardOutput)
+
+        finally:
+            logger.remove_all()
```
ryanjulian: these 2 lines in `def tearDown(self)`

tests/garage/logger/test_tabular.py
```diff
@@ -0,0 +1,118 @@
...
+
+            kv_dict = {'foo': foo, 'bar': bar}
+            self.check_csv_first_row(log_file, kv_dict)
+        finally:
+            logger.remove_all()
+            remove_if_exists(log_file)
+
+    def test_tabular_misc_stat(self):
+        log_file = 'test_tabular.csv'
+        try:
```
ryanjulian: setUp/tearDown

tests/garage/logger/test_tabular.py
```diff
@@ -0,0 +1,118 @@
...
+
+import numpy as np
+
+from garage.logger import CsvOutput, HistogramInput, logger, tabular
+from garage.misc.console import remove_if_exists
+
+
+class TestTabular(unittest.TestCase):
+    def test_tabular_record(self):
+        log_file = 'test_tabular.csv'
```
ryanjulian: setUp/tearDown

tests/garage/tf/algos/test_vpg.py
```diff
@@ -9,6 +10,7 @@
 class TestVPG(TfGraphTestCase):
     def test_vpg_cartpole(self):
         """Test VPG with CartPole-v1 environment."""
+        logger.reset_output(TensorBoardOutput())
```
ryanjulian: there's no point in logging to TensorBoard in the CI, so I think you can just remove all of these (or replace them with a NullOutput)

tests/garage/logger/test_snapshotter.py
```diff
@@ -0,0 +1,70 @@
...
+                else:
+                    self.fail("Snapshot mode exception not raised.")
+
+                for f, num in files.items():
+                    filename = osp.join(snap_dir, f)
+                    assert osp.exists(filename)
+                    with open(filename, 'rb') as pkl_file:
+                        data = pickle.load(pkl_file)
+                        assert data == snapshot_data[num]
+        finally:
```
ryanjulian: setUp/tearDown

tests/garage/logger/test_logger.py
```diff
@@ -0,0 +1,103 @@
...
+            assert not logger.has_output_type(TextOutput)
+
+            for file in log_files:
+                logger.add_output(TextOutput(file))
+            assert logger.has_output_type(TextOutput)
+            assert len(logger._outputs) == len(log_files)
+            logger.reset_output(TextOutput(log_file))
+            assert logger.has_output_type(TextOutput)
+            assert len(logger._outputs) == 1
+        finally:
```
ryanjulian: setUp/tearDown

tests/garage/logger/test_logger.py
```diff
@@ -0,0 +1,103 @@
...
+            assert logger.has_output_type(TextOutput)
+            logger.remove_output_type(CsvOutput)
+            assert not logger.has_output_type(CsvOutput)
+            assert logger.has_output_type(TextOutput)
+
+            tabular.record_misc_stat("stat", 1)
+            warn = "Log data of type " + type(tabular).__name__
+            warn += " was not accepted by any output"
+            assert logger.log(tabular) == warn
+        finally:
```
ryanjulian: setUp/tearDown

tests/garage/logger/test_logger.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+class TestLogger(unittest.TestCase):
+    def test_logger(self):
+        logger.disable_warnings()
+        assert logger.log(
+            "test") == "No outputs have been added to the logger."
+
+        log_file = 'text_logger.txt'
+        try:
```
ryanjulian: setUp/tearDown

tests/garage/logger/test_snapshotter.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+class TestSnapshotter(unittest.TestCase):
+    def test_snapshotter(self):
+        snap_dir = "test_snapshotter"
+        try:
+            snapshotter.snapshot_dir = snap_dir
+            assert snapshotter.snapshot_dir == snap_dir
+
+            snapfiles = {
```
ryanjulian: Use a parameterized test for this instead

tests/garage/logger/test_snapshotter.py
ryanjulian: setUp/tearDown + a parameterized test

here's an example of a parameterized test: https://github.com/rlworkgroup/garage/blob/master/tests/garage/envs/dm_control/test_dm_control_env.py

here's an example of setUp/tearDown:
https://github.com/rlworkgroup/garage/blob/40546df20cb3b42570786f9cf1514c49393f75b5/tests/garage/exploration_strategies/test_epsilon_greedy_strategy.py#L24

tests/garage/logger/test_logger.py
```diff
@@ -0,0 +1,103 @@
...
+        with self.assertRaises(ValueError):
+            logger.add_output("test_string")
+
+    def test_add_remove_outputs(self):
+        log_file = 'test.txt'
+        log_files = ['test_%u.txt' % i for i in range(5)]
+        csv_file = 'test.csv'
+        try:
+            logger.add_output(CsvOutput(csv_file))
+            logger.add_output(TextOutput(csv_file))
```
ryanjulian: log_file?

CatherineSue:
Issues to resolve:
* I rebased the branch but didn't address unresolved comments. 
* The CI failed because LoggerWarning is triggered before test_logger. We should probably reset `_warn` in test_logger.setUp.

garage/logger/simple_outputs.py
```diff
@@ -0,0 +1,105 @@
...
+    """Text file output for logger.
+
+    :param file_name: The file this output should log to.
+    :param with_timestamp: Whether to log a timestamp before the data.
+    """
+
+    def __init__(self, file_name, with_timestamp=True):
+        super().__init__(file_name, 'a')
+        self._with_timestamp = with_timestamp
+        self._delimiter = " | "
```
ryanjulian: single quote

garage/logger/simple_outputs.py
```diff
@@ -0,0 +1,105 @@
...
+class FileOutput(LogOutput, metaclass=abc.ABCMeta):
+    """File output abstract class for logger.
+
+    :param file_name: The file this output should log to.
+    :param mode: File open mode ('a', 'w', etc).
+    """
+
+    def __init__(self, file_name, mode='w'):
+        mkdir_p(os.path.dirname(file_name))
+        self._log_file = open(file_name,
```
ryanjulian: probably less awkward for the comment to go above this line

PR #467, opened 2019-01-23
Fix caching and CodeCov in garage-ci

gonzaiva:
The docker image for garage-ci was having the following issues:
- The codecov environment variables were not expanding correctly when
calling docker-run though a makefile in Travis.
- Caching was not working

First, the ci image got its own compose file docker-compose-ci to simply
things. To solve the codecov issue, the (bash<curl) operation was moved
from the makefile to the travis file, since the operation was not 
expanding properly when called from the makefile.
To pass the environment variables of codecov generated in travis, the 
variable ADD_ARGS was added in the makefile to pass these or any other
environment variables required in the future. This was documented in the 
README.md.
For caching, the wrong tag was used, since ":latest" was getting
appended to the name tag, giving an invalid value for cache image name.
Since each image type has its own compose file now, the tags for caching
are now hard-coded to their corresponding image types.
Also, only the presence of the mjkey.txt is needed to install mujoco-py, so an
empty file named mjkey.txt is created in the Dockerfile instead of passing the
real key.
Finally, the file Dockerfile was renamed to Dockerfile.base.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,35 +10,33 @@ MJKEY_PATH ?= ~/.mujoco/mjkey.txt
 # is already created
 CONFIG_PERSONAL := garage/config_personal.py
 
-build-ci: TAG ?= rlworkgroup/garage-ci
-build-ci: docker/docker-compose-headless.yml copy_config_personal
+build-ci: TAG ?= rlworkgroup/garage-ci:latest
+build-ci: docker/docker-compose-ci.yml copy_config_personal
 	TAG=${TAG} \
 	docker-compose \
-		-f docker/docker-compose-headless.yml \
+		-f docker/docker-compose-ci.yml \
 		build \
-		--build-arg MJKEY="$${MJKEY}"
+		${ADD_ARGS}
 
-build-headless: TAG ?= rlworkgroup/garage-headless
+build-headless: TAG ?= rlworkgroup/garage-headless:latest
 build-headless: docker/docker-compose-headless.yml check-mjkey copy_config_personal
 	TAG=${TAG} \
 	docker-compose \
 		-f docker/docker-compose-headless.yml \
 		build \
-		--build-arg MJKEY="$$(cat ${MJKEY_PATH})"
+		${ADD_ARGS}
 
+build-nvidia: TAG ?= rlworkgroup/garage-nvidia:latest
 build-nvidia: docker/docker-compose-nvidia.yml check-mjkey copy_config_personal
+	TAG=${TAG} \
 	docker-compose \
 		-f docker/docker-compose-nvidia.yml \
 		build \
-		--build-arg MJKEY="$$(cat ${MJKEY_PATH})"
+		${ADD_ARGS}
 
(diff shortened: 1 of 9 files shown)


gonzaiva:
@ryanjulian I have applied all changes we discussed and it caching and codecov are working again!

docker/docker-compose-nvidia.yml
```diff
@@ -5,7 +5,7 @@ services:
       cache_from:
         - rlworkgroup/garage-nvidia:latest
       context: ../
-      dockerfile: docker/Dockerfile
+      dockerfile: docker/Dockerfile.base
       args:
         - PARENT_IMAGE=nvidia/opengl:1.0-glvnd-runtime-ubuntu16.04
     image: rlworkgroup/garage-nvidia-base
```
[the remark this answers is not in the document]
gonzaiva: Fixed

Makefile
```diff
@@ -10,35 +10,33 @@ MJKEY_PATH ?= ~/.mujoco/mjkey.txt
 # is already created
 CONFIG_PERSONAL := garage/config_personal.py
 
-build-ci: TAG ?= rlworkgroup/garage-ci
-build-ci: docker/docker-compose-headless.yml copy_config_personal
+build-ci: TAG ?= rlworkgroup/garage-ci:latest
```
ryanjulian: any reason you called these latest? those are usually just tags for docker hub.

Makefile
```diff
@@ -10,35 +10,33 @@ MJKEY_PATH ?= ~/.mujoco/mjkey.txt
...
 	TAG=${TAG} \
 	docker-compose \
-		-f docker/docker-compose-headless.yml \
+		-f docker/docker-compose-ci.yml \
 		build \
-		--build-arg MJKEY="$${MJKEY}"
+		${ADD_ARGS}
 
-build-headless: TAG ?= rlworkgroup/garage-headless
+build-headless: TAG ?= rlworkgroup/garage-headless:latest
```
ryanjulian: latest?

Makefile
```diff
@@ -10,35 +10,33 @@ MJKEY_PATH ?= ~/.mujoco/mjkey.txt
...
 build-headless: docker/docker-compose-headless.yml check-mjkey copy_config_personal
 	TAG=${TAG} \
 	docker-compose \
 		-f docker/docker-compose-headless.yml \
 		build \
-		--build-arg MJKEY="$$(cat ${MJKEY_PATH})"
+		${ADD_ARGS}
 
+build-nvidia: TAG ?= rlworkgroup/garage-nvidia:latest
```
ryanjulian: latest?

Makefile
gonzaiva: It was just the default in case no TAG is provided. Docker will add latest as version anyway if the TAG is  `rlworkgroup/garage-ci`, right?

naeioi:
Seems like the commit message exceeds 72 chars. Could you do a quick fix?
https://travis-ci.com/rlworkgroup/garage/builds/98403667

Issue #363, opened 2018-10-26
tf/REPS (relative entropy policy search)


ryanjulian:
See https://github.com/rlworkgroup/garage/issues/331

PR #471, opened 2019-01-24
Add tf/REPS

CatherineSue:
Add Relative Entropy Policy Search to garage/tf.
It is the same as garage theano/REPS except for RNN policy support. The
RNN support will be added after the RNN primitives in tf have tested.

Close: #363

### DIFF
--- a/garage/tf/algos/reps.py
+++ b/garage/tf/algos/reps.py
@@ -1,0 +1,442 @@
+"""Relative Entropy Policy Search."""
+
+import numpy as np
+import scipy.optimize
+import tensorflow as tf
+
+from garage.misc import logger
+from garage.misc.overrides import overrides
+from garage.tf.algos import BatchPolopt
+from garage.tf.misc import tensor_utils
+from garage.tf.misc.tensor_utils import filter_valids
+from garage.tf.misc.tensor_utils import filter_valids_dict
+from garage.tf.misc.tensor_utils import flatten_batch
+from garage.tf.misc.tensor_utils import flatten_batch_dict
+from garage.tf.misc.tensor_utils import flatten_inputs
+from garage.tf.misc.tensor_utils import graph_inputs
+from garage.tf.optimizers import LbfgsOptimizer
+
+
+class REPS(BatchPolopt):
+    """
+    Relative Entropy Policy Search.
+
+    References
+    ----------
+    [1] J. Peters, K. Mulling, and Y. Altun, "Relative Entropy Policy Search,"
+        Artif. Intell., pp. 1607-1612, 2008.
+
+    Example:
+        $ python garage/examples/tf/reps_gym_cartpole.py
+
+    Args:
+        epsilon: dual func parameter.
+        l2_reg_dual: coefficient for dual func l2 regularization.
+        l2_reg_loss: coefficient for policy loss l2 regularization.
+        optimizer: policy optimizer.
+        optimizer_args: arguments of the policy optimizer.
+        dual_optimzier: dual func optimizer.
+        dual_optimizer_args: arguments of the dual optimizer.
+
+    """
+
+    def __init__(self,
+                 epsilon=0.5,
+                 l2_reg_dual=0.,
+                 l2_reg_loss=0.,
+                 optimizer=LbfgsOptimizer,
+                 optimizer_args=dict(max_opt_itr=50),
+                 dual_optimizer=scipy.optimize.fmin_l_bfgs_b,
+                 dual_optimizer_args=dict(maxiter=50),
+                 name="REPS",
+                 **kwargs):
+        self.name = name
+        self._name_scope = tf.name_scope(self.name)
+
+        with self._name_scope:
+            self.optimizer = optimizer(**optimizer_args)
+            self.dual_optimizer = dual_optimizer
+            self.dual_optimizer_args = dual_optimizer_args
+            self.epsilon = float(epsilon)
+            self.l2_reg_dual = float(l2_reg_dual)
+            self.l2_reg_loss = float(l2_reg_loss)
+
+        super().__init__(**kwargs)
+
+    @overrides
+    def init_opt(self):
+        """Initialize the optimization procedure."""
+        pol_loss_inputs, pol_opt_inputs, dual_opt_inputs = self._build_inputs()
+        self._policy_opt_inputs = pol_opt_inputs
+        self._dual_opt_inputs = dual_opt_inputs
+
+        pol_loss = self._build_policy_loss(pol_loss_inputs)
+        self.optimizer.update_opt(
+            loss=pol_loss,
+            target=self.policy,
+            inputs=flatten_inputs(self._policy_opt_inputs))
+
+    @overrides
+    def get_itr_snapshot(self, itr, samples_data):
+        """Return the data should saved in the snapshot."""
+        return dict(
+            itr=itr,
+            policy=self.policy,
+            env=self.env,
+        )
+
+    @overrides
+    def optimize_policy(self, itr, samples_data):
+        """Perform the policy optimization."""
+        # Initial BFGS parameter values.
+        x0 = np.hstack([self.param_eta, self.param_v])
+        # Set parameter boundaries: \eta>=1e-12, v unrestricted.
+        bounds = [(-np.inf, np.inf) for _ in x0]
+        bounds[0] = (1e-12, np.inf)
+
+        # Optimize dual
+        eta_before = self.param_eta
+        logger.log("Computing dual before")
+        self.feat_diff = self._features(samples_data)
+        dual_opt_input_values = self._dual_opt_input_values(samples_data)
+        dual_before = self.f_dual(*dual_opt_input_values)
+        logger.log("Optimizing dual")
+
+        def eval_dual(x):
+            self.param_eta = x[0]
+            self.param_v = x[1:]
+            dual_opt_input_values = self._dual_opt_input_values(samples_data)
+            return self.f_dual(*dual_opt_input_values)
+
+        def eval_dual_grad(x):
+            self.param_eta = x[0]
+            self.param_v = x[1:]
+            dual_opt_input_values = self._dual_opt_input_values(samples_data)
+            grad = self.f_dual_grad(*dual_opt_input_values)
+            eta_grad = np.float(grad[0])
+            v_grad = grad[1]
+            return np.hstack([eta_grad, v_grad])
+
+        params_ast, _, _ = self.dual_optimizer(
+            func=eval_dual,
+            x0=x0,
+            fprime=eval_dual_grad,
+            bounds=bounds,
+            **self.dual_optimizer_args,
+        )
+
+        logger.log("Computing dual after")
+        self.param_eta, self.param_v = params_ast[0], params_ast[1:]
+        dual_opt_input_values = self._dual_opt_input_values(samples_data)
+        dual_after = self.f_dual(*dual_opt_input_values)
+
+        # Optimize policy
+        policy_opt_input_values = self._policy_opt_input_values(samples_data)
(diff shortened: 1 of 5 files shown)


garage/tf/algos/reps.py
```diff
@@ -0,0 +1,419 @@
...
+
+    def _build_inputs(self):
+        """Decalre graph inputs variables."""
+        observation_space = self.policy.observation_space
+        action_space = self.policy.action_space
+        policy_dist = self.policy.distribution
+
+        with tf.name_scope("inputs"):
+            obs_var = observation_space.new_tensor_variable(
+                name="obs", extra_dims=2)
```
ryanjulian: i know YAPF won't force you to, but this block is probably more readable if each of these arguments are split into different lines

```python
valid_var = tensor_utils.new_tensor(
    name="valid",
    ndim=2,
    dtype=tf.float32)
```
...or similar

CatherineSue: em, but YAPF will force me to split into different lines
```
--- garage/tf/algos/reps.py	(original)
+++ garage/tf/algos/reps.py	(reformatted)
@@ -163,36 +163,22 @@
 
         with tf.name_scope("inputs"):
             obs_var = observation_space.new_tensor_variable(
-                name="obs",
-                extra_dims=2)
+                name="obs", extra_dims=2)

```

ryanjulian: oh i see

i think you can do 
```python
valid_var = tensor_utils.new_tensor(
    name="valid",
    ndim=2,
    dtype=tf.float32)  # yapf: disable
```

garage/tf/algos/reps.py
```diff
@@ -0,0 +1,442 @@
...
+            return self.f_dual(*dual_opt_input_values)
+
+        def eval_dual_grad(x):
+            self.param_eta = x[0]
+            self.param_v = x[1:]
+            dual_opt_input_values = self._dual_opt_input_values(samples_data)
+            grad = self.f_dual_grad(*dual_opt_input_values)
+            eta_grad = np.float(grad[0])
+            v_grad = grad[1]
+            return np.hstack([eta_grad, v_grad])
```
ahtsan: Should we avoid creating inline functions?

tests/garage/tf/algos/test_reps.py
```diff
@@ -0,0 +1,37 @@
...
+            baseline=baseline,
+            batch_size=4000,
+            max_path_length=100,
+            n_itr=10,
+            discount=0.99,
+            max_kl_step=1e6,
+            plot=False)
+
+        last_avg_ret = algo.train(sess=self.sess)
+        assert last_avg_ret > 5
```
ahtsan: Should the average return be higher to make sure it works?

garage/tf/algos/reps.py
```diff
@@ -0,0 +1,442 @@
...
+    def _build_inputs(self):
+        """Decalre graph inputs variables."""
+        observation_space = self.policy.observation_space
+        action_space = self.policy.action_space
+        policy_dist = self.policy.distribution
+
+        with tf.name_scope("inputs"):
+            obs_var = observation_space.new_tensor_variable(
+                name="obs",
+                extra_dims=2)   # yapf: disable
```
gonzaiva: I suppose these extra dims are for the batch sizes, but could we make that clearer by defining some constant instead of having these hard coded numbers.

garage/tf/algos/reps.py
```diff
@@ -0,0 +1,442 @@
...
+                    name=k)
+                for k, shape in self.policy.state_info_specs
+            }   # yapf: disable
+            policy_state_info_vars_list = [
+                policy_state_info_vars[k]
+                for k in self.policy.state_info_keys
+            ]   # yapf: disable
+
+            policy_old_dist_info_vars = {
+                k: tf.placeholder(
```
gonzaiva: We could call [new_tensor](https://github.com/rlworkgroup/garage/blob/1c4eebe59e4609be1a890c968c293f2d592ed3ad/garage/tf/misc/tensor_utils.py#L72) here.

garage/tf/regressors/gaussian_conv_regressor.py
```diff
@@ -86,7 +86,7 @@ def __init__(self,
                             np.zeros(output_dim),
                             np.full(output_dim, np.log(init_std))
                         ],
-                        axis=0)
+                        axis=0)  # yapf: disable
```
gonzaiva: Is this due to a new version of yapf?

garage/tf/algos/reps.py
```diff
@@ -0,0 +1,442 @@
...
+
+        return policy_loss_inputs, policy_opt_inputs, dual_opt_inputs
+
+    def _build_policy_loss(self, i):
+        """Initialize policy loss complie function based on inputs i."""
+        pol_dist = self.policy.distribution
+        is_recurrent = self.policy.recurrent
+
+        # Initialize dual params
+        self.param_eta = 15.
```
gonzaiva: Is there any particular reason why eta is 15?

garage/tf/algos/reps.py
```diff
@@ -0,0 +1,442 @@
...
+        )
+
+        return flatten_inputs(policy_opt_input_values)
+
+    def _features(self, samples_data):
+        """Get valid view features based on samples data."""
+        paths = samples_data["paths"]
+        feat_diff = []
+        for path in paths:
+            o = np.clip(path["observations"], -10, 10)
```
gonzaiva: Why is clipping min/max set to -10/10?

garage/tf/algos/reps.py
CatherineSue: We should. But I don't know any simple way to make `eval_dual` and `eval_dual_grad` to be member functions or helper functions outside the class.

tests/garage/tf/algos/test_reps.py
CatherineSue: This is low because REPS is highly unstable and flaky.

garage/tf/algos/reps.py
CatherineSue: em,,, we need to set `shape=[None] * 2 + list(shape)` here, but `new_tensor` returns a shape of `[None] * ndim`.

garage/tf/regressors/gaussian_conv_regressor.py
CatherineSue: I think so but I don't remember.

garage/tf/algos/reps.py
gonzaiva: oh, you're right, my bad

garage/tf/regressors/gaussian_conv_regressor.py
gonzaiva: Okay, I just thought it's weird that nothing changed in this file and yapf complained.

garage/tf/algos/reps.py
```diff
@@ -0,0 +1,442 @@
...
+        logger.record_tabular("{}/LossAfter".format(self.policy.name),
+                              loss_after)
+        logger.record_tabular("{}/dLoss".format(self.policy.name),
+                              loss_before - loss_after)
+        logger.record_tabular("{}/KLBefore".format(self.policy.name),
+                              policy_kl_before)
+        logger.record_tabular("{}/KL".format(self.policy.name), policy_kl)
+
+    def _build_inputs(self):
+        """Decalre graph inputs variables."""
```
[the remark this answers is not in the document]
CatherineSue: Good catch, thx!

garage/tf/algos/reps.py
ryanjulian: this is pretty common around the TF algos. the 2 extra dimensions are the valid bit and time.

i agree that this is annoying, but it's best dealt with as a structural change in the TF tree (probably after Anson gets models in)--not just here in this algo.

garage/tf/algos/reps.py
ryanjulian: we can only squeeze so much magic out of hyperparameters, unfortunately.

garage/tf/algos/reps.py
ryanjulian: this is worth a comment. i'm pretty sure it's to limit the dynamic range of the observations.

garage/tf/algos/reps.py
ryanjulian: in this case i think it's fine. it comes down do a design issue with the optimizers

garage/tf/algos/reps.py
```diff
@@ -0,0 +1,442 @@
...
+
+    """
+
+    def __init__(self,
+                 epsilon=0.5,
+                 l2_reg_dual=0.,
+                 l2_reg_loss=0.,
+                 optimizer=LbfgsOptimizer,
+                 optimizer_args=dict(max_opt_itr=50),
+                 dual_optimizer=scipy.optimize.fmin_l_bfgs_b,
```
naeioi: It is feasible to use LbfgsOptimizer instead? Otherwise, we have a divergent interface for optimizer against the other one above.

garage/tf/algos/reps.py
```diff
@@ -0,0 +1,442 @@
...
+        policy_state_info_list = [
+            samples_data["agent_infos"][k]
+            for k in self.policy.state_info_keys
+        ]   # yapf: disable
+        policy_old_dist_info_list = [
+            samples_data["agent_infos"][k]
+            for k in self.policy.distribution.dist_info_keys
+        ]
+
+        self.feat_diff = self._features(samples_data)
```
naeioi: Features will be recalculated every time when dual_optimizer search for eta and v. This is not necessary.

garage/tf/algos/reps.py
```diff
@@ -0,0 +1,442 @@
...
+            i.flat.policy_state_info_vars,
+            name="policy_dist_info_flat")
+
+        policy_dist_info_valid = filter_valids_dict(
+            policy_dist_info_flat,
+            i.flat.valid_var,
+            name="policy_dist_info_valid")
+
+        with tf.name_scope("bellman_error"):
+            delta_v = i.valid.reward_var + tf.tensordot(
```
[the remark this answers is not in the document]
zhanpenghe: Based on my understanding, this should be something like `r(s, a) + E(Phi * Theta) - Phi_s * Theta`.

garage/tf/algos/reps.py
CatherineSue: No. I did find it is possible

garage/tf/algos/reps.py
CatherineSue: I remember I put it here for a reason. But I will check if it is unnecessary here. Thanks

CatherineSue: `feat_diff` is used to calculate the bellman error. This is required here.

garage/tf/algos/reps.py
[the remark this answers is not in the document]
CatherineSue: I think the above formula is true for REPS. Here the algorithm is policy iteration with REPS, which uses Table 2 in the paper.

garage/tf/algos/reps.py
CatherineSue: I couldn't find how to compute features in the paper, so this method is ported from theano/algos/reps.py.

CatherineSue: I agree this is to limit the range of the observations. I don't know why they do it here instead of the sampler, so that we could normalize the observations, like Zhanpeng proposed. Another constant param is [100.0 in here](https://github.com/rlworkgroup/garage/pull/471/files#diff-1bc9d53a38cba8890b10755ab1366305R435). Is it related to max_path_length or max_reward?

garage/tf/algos/reps.py
CatherineSue: I've moved the line out of this function.

PR #473, opened 2019-01-24
Ignore line break after binary op & in-place yapf

naeioi:
This PR contains several small fixes for code formatter.
- Enable in-place formatting in yapf.
Free your hand from running yapf manually.
- Ignore `W504  # line break after binary operator` flake8 check for added files. 
This is a blocking issue in commit check.
yapf and flake8 have conflict opinions on this rule. This check is currently ignored for modified files but not for added files. Adopt yapf's style. 
- Update yapf version to v0.25.0 in pre-commit.
This is a blocking issue in commit check.
There are two yapf in our system, one in pre-commit, the other in `environment.yml`. Currently the version of the first one is v0.22.0 but the later is v0.25.0 (the latest one in pypi). 
This is problematic because there are some small differences between these two versions. When someone formats their code using `yapf -i` (the newer version), the formatted code might be rejected by yapf in pre-commit (the older version).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -87,10 +87,19 @@ repos:
     #     stages: [commit, push]
     -   id: trailing-whitespace         # Trims trailing whitespace.
         stages: [commit, push]
-
+# First pass: check format
 -   repo: https://github.com/pre-commit/mirrors-yapf
     rev: v0.25.0
     hooks:
     -   id: yapf
-        args: ['-dipr']
+        name: Check format by yapf
+        args: ['-dpr']
+        stages: [commit, push]
+# Second pass: format in-place
+-   repo: https://github.com/pre-commit/mirrors-yapf
+    rev: v0.25.0
+    hooks:
+    -   id: yapf
+        name: Format in-place by yapf
+        args: ['-ipr']
         stages: [commit, push]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -89,8 +89,8 @@ repos:
         stages: [commit, push]
 
 -   repo: https://github.com/pre-commit/mirrors-yapf
-    rev: v0.22.0
+    rev: v0.25.0
     hooks:
     -   id: yapf
-        args: ['-dpr']
+        args: ['-dipr']
         stages: [commit, push]
--- a/scripts/travisci/check_flake8.sh
+++ b/scripts/travisci/check_flake8.sh
@@ -47,11 +47,10 @@ $(cat .gitignore)
 
 # Error codes ignored for changed files
 ignored_errors_changed=(
-# BREAK BEFORE BINARY OPERATOR
-# It enforces the break after the operator, which is acceptable, but it's
-# preferred to do it before the operator. Since YAPF enforces the preferred
-# style, this rule is ignored.
 D     # docstring rules disabled
+# We prefer break after binary operator, but YAPF breaks before and the behavior
+# is not configurable, so we disable related rules in flake8.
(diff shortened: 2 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -89,8 +89,8 @@ repos:
         stages: [commit, push]
 
 -   repo: https://github.com/pre-commit/mirrors-yapf
-    rev: v0.22.0
+    rev: v0.25.0
     hooks:
     -   id: yapf
-        args: ['-dpr']
+        args: ['-dipr']
```
CatherineSue: Have you tested this?
``` bash
usage: yapf [-h] [-v] [-d | -i] [-r | -l START-END] [-e PATTERN]
            [--style STYLE] [--style-help] [--no-local-style] [-p] [-vv]
            [files [files ...]]
yapf: error: argument -i/--in-place: not allowed with argument -d/--diff

```

naeioi: uhh, I totally missed that. Seems like we must run yapf for two passes if we want in-place formatting.

ryanjulian:
Can you give an example of the binary operator format which is acceptable under this change?

naeioi:
Concerning this line https://github.com/rlworkgroup/garage/blob/095bb5d25b32df1d44b47e99a78a9b01796941d9/garage/experiment/experiment.py#L993-L995

flake8 complaints as follows
```
garage/experiment/experiment.py:994:13: W504 line break after binary operator
```
yapf complaints as follows
```
--- garage/run_experiment/run_experiment.py     (original)
+++ garage/run_experiment/run_experiment.py     (reformatted)
@@ -991,8 +991,8 @@
             current_commit = None
 
         file_name = str(timestamp) + "_" + hashlib.sha224(
-            subprocess.check_output(["pwd"]) + str(current_commit).encode()
-            + str(timestamp).encode()).hexdigest() + ".tar.gz"
+            subprocess.check_output(["pwd"]) + str(current_commit).encode() +
+            str(timestamp).encode()).hexdigest() + ".tar.gz"
 
         file_path = "/tmp/" + file_name

```

.pre-commit-config.yaml
```diff
@@ -87,10 +87,19 @@ repos:
...
+# First pass: check format
 -   repo: https://github.com/pre-commit/mirrors-yapf
-    rev: v0.22.0
+    rev: v0.25.0
     hooks:
     -   id: yapf
+        name: Check format by yapf
         args: ['-dpr']
         stages: [commit, push]
+# Second pass: format in-place
```
ryanjulian: hmm why not just do the in-place format and be done with it?

ryanjulian:
Our preferred style is W504 (which is the new flake8/PEP8 standard), but it appears that YAPF has not caught up: https://github.com/google/yapf/issues/647

.pre-commit-config.yaml
ryanjulian: Oh I see `-dpr` returns non-zero if it finds a change, which will fail the CI. `-ipr` will not return non-zero on a change. never mind.

CatherineSue:
Please change the comment [here](https://github.com/rlworkgroup/garage/pull/473/files#diff-7bb4f5052141ea4a268c1283eb0af7baR51), too. YAPF seems to support break before the binary operator before, that's why we ignored W503.

ryanjulian:
What did you two figure out with respect the YAPF/flake8 and W503/W504?

Can we reliably enforce break-before-operator?

e.g.
```python
my_sum = first_number
         + second_number
         + third_number
```

naeioi:
No, we can't. Since yapf breaks after binary op and there's no way to configure that.

PR #474, opened 2019-01-24
Experiment runner

naeioi:
This PR introduces experiment runner interface. I squashed the whole PR into 3 commits for easy review. 
- commit `Add experiment runner`
Important modules includes
    - `Experiment` class: experiment coordinator. Middleware between agent, sampler and checkpointer.
    - `Agent` interface: wrapper abstraction of algorithm related components. 
    - `Sampler` interface and its implemtation `BatchSampler`.
    - General purpose`Checkpointer` interface and its implemetation `DiskCheckpointer`.
- commit `Add Pytorch implementation of VPG`
Reference implementation of VPG in Pytorch under `Agent` interface.
**Note:** Manually install pytorch by `pip install pytorch` is required if you want to try. 
- commit `Move exp runner to garage.experiment`
This commit moves old `run_experiment()` from `garage.experiment` to `garage.run_experiment` and moves new experiment runner into `garage.experiment`.

Important files to look at for a quick overview:
- [examples/experiment/vpg-pendulum.py](https://github.com/rlworkgroup/garage/blob/exp_runner/examples/experiment/vpg-pendulum.py). Experiment setup script.
- [garage/garage/experiment/experiment.py](https://github.com/rlworkgroup/garage/blob/exp_runner/garage/experiment/experiment.py)
Experiment runner. It coordinates algo, sampler and checkpointer. These logics are currently attributed to algorithm implementation in garage.
- [ garage/garage/experiment/torch/algos/vpg.py](https://github.com/rlworkgroup/garage/blob/exp_runner/garage/experiment/torch/algos/vpg.py). VPG implementation.

**Also**: Code under `agent/garage` is my early unsuccessful attemp to make a non-intrusive interface. I cannot get rid of it from git history and you don't bother to look at it.


garage/contrib/exp/agents/agent.py
```diff
@@ -0,0 +1,43 @@
+import abc
+
+
+class Agent(abc.ABC):
+    @abc.abstractmethod
+    def get_actions(self, obs):
```
ryanjulian: i think it would be permissible to include the singular `get_action()` (gets one action only), as a convenience.

i think you can implement it in this ABC since the implementation is obvious

garage/contrib/exp/agents/agent.py
```diff
@@ -0,0 +1,43 @@
...
+        """Get actions given a batch of observations.
+
+        Args:
+            obs: Tensor(batch_size, observation_dim)
```
ryanjulian: hmm what is a `Tensor`? Is it an `np.ndarray`? Does it conform to some `akro.space`?

garage/contrib/exp/agents/agent.py
```diff
@@ -0,0 +1,43 @@
...
+        """
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def train_once(self, samples):
+        """Train policy given trajectories sampled under latest policy.
+
+        Args:
+            samples: [Path]
+                Path: {
```
ryanjulian: Does this mean samples is a _collection of dictionaries of Tensors_, or a _dictionary of collections of Tensors_?

The distinction is important.

@krzentner do you think we should type these things with `akro`, or keep the types vanilla? There are upsides and downsides to both options.

garage/contrib/exp/checkpointers/checkpointer.py
```diff
@@ -0,0 +1,45 @@
...
+    return '_'.join(args) + '.pkl'
+
+
+class Checkpointer(abc.ABC):
+    def __init__(self, prefix, resume=True):
+        self.resume = resume
+        self.prefix = prefix
+
+    @abc.abstractmethod
+    def load(self, **kwargs):
```
ryanjulian: is `**kwargs` necessary in an ABC if you want child classes to be able to add additional args?

garage/contrib/exp/checkpointers/disk_checkpointer.py
```diff
@@ -0,0 +1,161 @@
...
+
+        """
+        latest_checkpoint, _ = self._get_latest_checkpoint(kwargs.keys(), dry=True)
+        if not latest_checkpoint or not self.resume:
+            self.save(**kwargs)
+            return kwargs
+        else:
+            return self._load(**kwargs)
+
+    def save(self, **kwargs):
```
ryanjulian: i think concrete implementations should have (only) concrete arguments.

garage/contrib/exp/core/misc.py
```diff
@@ -0,0 +1,17 @@
...
+
+EnvSpec = namedtuple('EnvSpec', ['action_space', 'observation_space'])
+Space = namedtuple('Space', 'flat_dim')
+
+
+def _get_flat_dim(space):
+    return np.prod(space.low.shape)
+
+
+def get_env_spec(env):
```
ryanjulian: I would be really surprised if this isn't already somewhere in garage.spaces

garage/contrib/exp/core/policy.py
```diff
@@ -0,0 +1,33 @@
+import abc
+
+
+class Policy(abc.ABC):
```
ryanjulian: I'm not sure why this interface is necessary when `Agent` is nearly exactly the same.

garage/contrib/exp/experiment.py
```diff
@@ -0,0 +1,58 @@
+import gym
+
+from garage.contrib.exp import Agent
+from garage.contrib.exp import Checkpointer
+from garage.contrib.exp import Logger
+from garage.contrib.exp import Sampler
+
+
+class Experiment():
```
ryanjulian:
```python
class Experiment:
    # ...
```

garage/contrib/exp/experiment.py
```diff
@@ -0,0 +1,58 @@
...
+    def __init__(
+            self,
+            env: gym.Env,
+            agent: Agent,
+        checkpointer: Checkpointer,
+            logger: Logger,
+        sampler: Sampler = None,
+        # common experiment parameters:
+            n_itr=10,
+            batch_size=1000,
```
ryanjulian: hmm is this actually a parameter of the Sampler instead?

garage/contrib/exp/checkpointers/disk_checkpointer.py
ryanjulian: I see what's going on here.

You are using kwargs as a dict to allow the user to avoid the boilerplate of constructing one themselves. Perhaps.

@krzentner what do you think of this interface? what if instead all you can give is something pickleable?

garage/contrib/exp/experiment.py
```diff
@@ -0,0 +1,58 @@
...
+    ):
+        if sampler is None:
+            from garage.contrib.exp.samplers import BatchSampler
+            sampler = BatchSampler(env=env)
+
+        if checkpointer.resume:
```
ryanjulian: this makes `.resume` part of the public `Checkpointer` interface, but it's not in the ABC

garage/contrib/exp/experiment.py
```diff
@@ -0,0 +1,58 @@
...
+from garage.contrib.exp import Agent
+from garage.contrib.exp import Checkpointer
+from garage.contrib.exp import Logger
+from garage.contrib.exp import Sampler
+
+
+class Experiment():
+    def __init__(
+            self,
+            env: gym.Env,
```
ryanjulian: let's hold off on type annotations until we actually decide whether to use them in the codebase.

garage/contrib/exp/experiment.py
```diff
@@ -0,0 +1,58 @@
...
+            agent: Agent,
+        checkpointer: Checkpointer,
+            logger: Logger,
+        sampler: Sampler = None,
+        # common experiment parameters:
+            n_itr=10,
+            batch_size=1000,
+    ):
+        if sampler is None:
+            from garage.contrib.exp.samplers import BatchSampler
```
ryanjulian: how many CPUs does the BatchSampler get to use?

garage/contrib/exp/experiment.py
```diff
@@ -0,0 +1,58 @@
...
+        self.itr = 0
+
+    def train_once(self):
+        self.itr = self.itr + 1
+        print('Itration', self.itr)
+
+        obs = self.sampler.reset()
+        while self.sampler.sample_count < self.batch_size:
+            actions = self.agent.get_actions(obs)
+            obs = self.sampler.step(actions)
```
ryanjulian: hmmm the sampler is _not necessarily_ stepping. for instance it might just return some samples from a dataset.

This whole method really confuses me actually. Why does the experiment runner have a method called `train_once()`?

Maybe it should be more like

```python
def train(self):
    # these should probably be constructor parameters, but then we can only accept a sampler Type
    #  in the Experiment constructor. That makes it easier to serialize an experiment, anyway.
    self._sampler.set_agent(self._agent)  
    self._sampler.set_environment(self._environment)
    while self.itr < self.n_itr:
        samples = self.sampler.get_samples()  # sampler already has agent and environment
        self._agent.train_once(samples)
        self._sampler.update_agent(self._agent)

        # Logging
        # TODO: how does this work Peter's logger if the agent has access to the Logger? The agent only has a facade logger, right? Then what type does get_summary() produce?
        self._logger.log(self._agent.get_summary())
        self._logger.log(self._sampler.get_summary())
        self._logger.dumps()

        # Checkpointing
        self._checkpointer.maybe_save(self)  # ??? save the whole experiment
```

Issue #331, opened 2018-10-03
Remove Theano tree


ryanjulian:
There is no interest in Theano among current users, and it presents a maintenance burden.

The first step in this process is to make sure we don't lose anything.

1. Ensure there is a version of every algorithm and primitive in the Theano tree in the TensorFlow tree.
1. Ensure there are no useful utilities in the Theano tree which should be moved into common directories
1. Bring TF test coverage of primitives and algorithms up to the same level (or better) as the Theano tree
1. Delete Theano tree (and dependencies) from the repo.

ryanjulian:
We will do this after cutting the 2018.10 release (i.e. 2018.10 will be the last tree to support Theano).

ryanjulian:
More primitives available in Theano but not TF. We need to decide which of these we will port over.

* baselines.GaussianConvBaseline
* regressors.GaussianConvRegressor
* distributions.Delta
* optimizers.HessianFreeOptimizer
* optimizers.HfOptimizer

PR #475, opened 2019-01-24
Add tf/GaussianConv*

CatherineSue:
Add GaussianConvBaseline and GaussianConvRegressor to tf.
Change the default baseline name in tf/*Baselines.
Add simple initialize test of tf/baselines.

See: #331


ryanjulian:
I would prefer if this came with a test, but I am more interested in deleting Theano as fast as possible.

PR #477, opened 2019-01-25
Remove theano

CatherineSue:
See: #331.


environment.yml
```diff
@@ -53,7 +52,6 @@ dependencies:
         - scipy
         - tensorboard
         - tensorflow<1.10,>=1.9.0
-        - Theano==1.0.2
```
ryanjulian: i think that mkl and mkl-service above are also Theano-only dependencies.

ryanjulian: You can also likely remove conda-forge and menpo from channels.

ryanjulian:
Can you take a look at the coverage report? it looks like some lines of code are no longer covered, but they're not Theano lines.

Perhaps there were some tests in the theano tree which covered those lines? It would be nice to port them to TF if they are easily ported.

CatherineSue:
I checked the cov report. Most of the missing lines come from `garage.sampler`, `garage.algos`, `garage.experiment`.
Sampler and algos are not easy to port to tf because they require no tf dep. So I added some dummy distribution and policies to test algos.

I couldn't find what theano tests covered the missing lines in `garage.experiment`.

CatherineSue:
I moved the experiment test to `tests.garage.experiment`. But it doesn't help to cover the experiment lines. Any idea why?

ryanjulian:
Have you verified the test actually runs in the CI? Can you see its output in the build log?

CatherineSue:
Yes it runs on the CI. The test doesn't have outputs but the total number of running tests increased after I added it.

ryanjulian:
I looked at the new coverage report and I am happy after your changes. Don't worry too much about whatever happened with experiment--it will be replaced soon anyway.

I think you can delete this:
https://codecov.io/gh/rlworkgroup/garage/compare/3bad524a2cb80e73f4ba5f9682e519b8ab15380f...b87bc92d556e0484e8b3c5677aed2ddb25ac655d/changes#D1-153

examples/theano/cluster_demo.py
```diff
@@ -1,57 +0,0 @@
-import sys
```
ryanjulian: it would be nice to port this example to TF since I don't think there's an example of complex run_experiment usage there.

PR #481, opened 2019-01-27
Autopickable and BaseModel class

ahtsan:
This PR only serves the purpose for Autopickable and BaseModel blueprint, and examples. Names are not finalized.

`Autopickable` class aims to replace the existing `Serializable` and `Parameterzied` by exploiting serialization modules from `tf.keras.models.Model` class, and a BaseModel class is introduced.


garage/tf/core/parameterLayer.py
```diff
@@ -0,0 +1,54 @@
...
+
+class ParameterLayer(KerasLayer):
+    """
+    Parameter layer based on tf.keras.layers.Layer.
+
+    Used as layer that could be broadcast to a certain shape to
+    match with input variable during training.
+    Example: A trainable parameter variable with shape (2,), it needs to be
+    broadcasted to (32, 2) when applied to a batch with size 32.
+    """
```
CatherineSue: Please add doc string of the constructor params.

garage/tf/core/parameterLayer.py
```diff
@@ -0,0 +1,54 @@
...
+    def get_config(self):
+        """Cusomterized configuration for serialization."""
+        config = {
+            'length': self.length,
+            'name': self.name,
+            'initializer': self.initializer,
+            'trainable': self.trainable
+        }
+        base_config = super().get_config()
+        return dict(list(base_config.items()) + list(config.items()))
```
CatherineSue: I'd prefer unpack the two dicts using [PEP448](https://www.python.org/dev/peps/pep-0448/).

garage/tf/core/mlp2.py
```diff
@@ -0,0 +1,80 @@
...
+            _out = Dense(
+                units=hidden_sizes,
+                activation=hidden_nonlinearity,
+                kernel_initializer=hidden_w_init,
+                bias_initializer=hidden_b_init)(_out)
+            if batch_normalization:
+                _out = BatchNormalization(
+                    hidden_sizes, activation=hidden_nonlinearity)(_out)
+        else:
+            _out = Dense(
```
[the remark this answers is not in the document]
ahtsan: not sure why I did that

garage/tf/models/autopickable.py
```diff
@@ -0,0 +1,75 @@
+"""
```
CatherineSue: Should the file rename to `auto_pickable.py`?

tests/garage/tf/baselines/test_baselines.py
```diff
@@ -13,5 +13,5 @@ class TestTfBaselines(TfGraphTestCase):
     def test_baseline(self):
         """Test the baseline initialization."""
         box_env = TfEnv(DummyBoxEnv())
-        deterministic_mlp_baseline = DeterministicMLPBaseline(env_spec=box_env)
-        gaussian_mlp_baseline = GaussianMLPBaseline(env_spec=box_env)
+        DeterministicMLPBaseline(env_spec=box_env)
```
CatherineSue: This file is updated in master branch. Please rebase.

garage/tf/models/gaussian_mlp_model.py
```diff
@@ -0,0 +1,194 @@
...
+            - exp: The logarithm of the std will be stored, and applied a
+                exponential transformation
+            - softplus: The std will be computed as log(1+exp(x))
+        :return:
+    """
+
+    def __init__(self,
+                 input_dim,
+                 output_dim,
+                 dist=tf.contrib.distributions.MultivariateNormalDiag,
```
CatherineSue: I wonder if we should let users to pass the distribution. If a user chooses tf.distributions.Categorical, what makes this model to?

garage/tf/models/gaussian_mlp_model.py
```diff
@@ -0,0 +1,194 @@
...
+                 adaptive_std=False,
+                 std_share_network=False,
+                 std_hidden_sizes=(32, 32),
+                 min_std=1e-6,
+                 max_std=None,
+                 std_hidden_nonlinearity='relu',
+                 hidden_nonlinearity='relu',
+                 output_nonlinearity=None,
+                 std_parameterization='exp',
+                 *args,
```
CatherineSue: Same here. I can't find where the args and kwargs are used.

garage/tf/models/gaussian_mlp_model.py
```diff
@@ -0,0 +1,194 @@
...
+                        initializer=p,
+                        trainable=self._learn_std,
+                        scope="std_network")(input_var)
+
+            with tf.variable_scope("std_limits"):
+                if self._min_std_param:
+                    std_param_var = Lambda(
+                        lambda x, f, p: f(x, p),
+                        arguments={
+                            'f': tf.maximum,
```
CatherineSue: Please update all the quotation marks.

garage/tf/core/parameterLayer.py
```diff
@@ -0,0 +1,54 @@
...
+    Used as layer that could be broadcast to a certain shape to
+    match with input variable during training.
+    Example: A trainable parameter variable with shape (2,), it needs to be
+    broadcasted to (32, 2) when applied to a batch with size 32.
+    """
+
+    def __init__(self,
+                 length,
+                 scope='ParameterLayer',
+                 initializer='ones',
```
CatherineSue: Here the default value of initializer is a string, but in [here](https://github.com/rlworkgroup/garage/pull/481/files#diff-f27d474d4aed442478e7e44fdc2de671R148) you passed a function, will this be a problem?

garage/tf/models/base.py
```diff
@@ -0,0 +1,84 @@
...
+    @property
+    def output(self):
+        """Tensor output of the Model."""
+        if len(self.model.outputs) == 1:
+            return self.model.outputs[0]
+
+        return self.model.outputs
+
+    @property
+    def inputs(self):
```
CatherineSue: Any reason not merge input and inputs into a single interface?

garage/tf/core/mlp2.py
```diff
@@ -0,0 +1,80 @@
...
+        output_b_init: Initializer function for the bias
+                    of output dense layer(s).
+        batch_normalization: Bool for using batch normalization or not.
+
+    Return:
+        The MLP Model.
+    """
+    with tf.variable_scope(scope):
+        _out = input_var
+        if isinstance(hidden_sizes, int):
```
CatherineSue: Do we actually need this? Because in GaussianMLPModel, the doc saying the default of hidden_sizes is a list.

garage/tf/models/base.py
ahtsan: comply with `tf.keras.model`

garage/tf/core/mlp2.py
ahtsan: we don't

garage/tf/models/autopickable.py
ahtsan: yes that sounds better.

garage/tf/models/gaussian_mlp_model.py
ahtsan: GaussgoricalMLP? Yes we shouldn't let user pass the dist, my original intent is to show that serialization from `tf.keras.model` allows us to have `tf.distribution` as argument.

tests/garage/tf/models/test_model_autopickable.py
```diff
@@ -0,0 +1,116 @@
...
+
+        model = GaussianMLPModel(
+            input_dim=5, output_dim=2, hidden_sizes=(4, 4), init_std=2.0)
+
+        self.sess.run(tf.global_variables_initializer())
+
+        data = np.random.random((3, 5))
+
+        result_from_model = []
+        for output in model.outputs:
```
CatherineSue: `result_from_model = self.sess.run(model.outputs, feed_dict={model.input: data})`?

garage/tf/models/gaussian_mlp_model.py
ahtsan: useless, will remove that.

garage/tf/core/parameterLayer.py
ahtsan: `tf.keras` accepts predefined strings and callable, so I think should be fine.

tests/garage/tf/models/test_model_autopickable.py
```diff
@@ -0,0 +1,116 @@
...
+            result_from_model.append(
+                self.sess.run(output, feed_dict={model.input: data}))
+
+        model_pickled = pickle.loads(pickle.dumps(model))
+
+        result_from_pickled_model = []
```
CatherineSue: `result_from_pickled_model = self.sess.run(outputs, feed_dict={model_pickled.input: data}))`?

ahtsan: Yes, good catch!

garage/tf/core/mlp2.py
```diff
@@ -0,0 +1,80 @@
...
+         hidden_sizes,
+         scope='mlp',
+         hidden_nonlinearity='relu',
+         hidden_w_init='glorot_uniform',
+         hidden_b_init='zeros',
+         output_nonlinearity=None,
+         output_w_init='glorot_uniform',
+         output_b_init='zero',
+         batch_normalization=False,
+         *args,
```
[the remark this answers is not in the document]
ahtsan: useless, will remove that.

garage/tf/models/base.py
```diff
@@ -0,0 +1,84 @@
+"""This file contains the abstraction class for models."""
+from garage.tf.models import AutoPickable
+
+
+class BaseModel:
```
CatherineSue: I don't see BaseModel is used anywhere. When will we use it?

garage/tf/core/parameterLayer.py
```diff
@@ -0,0 +1,54 @@
+"""Parameter layer in TensorFlow."""
```
CatherineSue: Please change the file name.

garage/tf/models/base.py
ahtsan: When people don't want `AutoPickable` but want to use the API from `Model`.

CatherineSue: How does the user pickle model if he uses BaseModel?

garage/tf/models/gaussian_mlp_model.py
```diff
@@ -0,0 +1,194 @@
...
+                    })(std_param_var)
+            else:
+                raise NotImplementedError
+
+        sample_var = Lambda(
+            lambda x, f, s: f(x[0], x[1]).sample(seed=s),
+            arguments={
+                'f': self._dist,
+                's': ext.get_seed()
+            })([mean_var, std_param_var])
```
CatherineSue: Could you pass a `name=MultivariateNormalDiagDistribution` to Lambda here?

ahtsan: Yes.

garage/tf/models/gaussian_mlp_model.py
```diff
@@ -0,0 +1,194 @@
...
+                    scope="mean_network")
+                mean_var = mean_model.output
+
+                if self._adaptive_std:
+                    std_model = mlp2(
+                        input_var=input_var,
+                        output_dim=self._output_dim,
+                        hidden_sizes=self._std_hidden_sizes,
+                        hidden_nonlinearity=self._std_hidden_nonlinearity,
+                        output_nonlinearity=self._output_nonlinearity,
```
CatherineSue: Why isn't there a _std_output_nonlinearity?

garage/tf/models/gaussian_mlp_model.py
```diff
@@ -0,0 +1,194 @@
...
+                 scope="GaussianMLPModel",
+                 hidden_sizes=(32, 32),
+                 learn_std=True,
+                 init_std=1.0,
+                 adaptive_std=False,
+                 std_share_network=False,
+                 std_hidden_sizes=(32, 32),
+                 min_std=1e-6,
+                 max_std=None,
+                 std_hidden_nonlinearity='relu',
```
CatherineSue: Please group the parameters of std and mean network.

garage/tf/core/parameterLayer.py
```diff
@@ -0,0 +1,54 @@
...
+                 **kwargs):
+        self.length = length
+        self.initializer = initializer
+        self.trainable = trainable
+        self.scope = scope
+        super().__init__(**kwargs)
+
+    def build(self, input_shape):
+        """tf.keras.layers.Layer build."""
+        with tf.variable_scope(self.scope):
```
ryanjulian: how does Keras handle scope? I wonder, does it handle scope automatically?

garage/tf/models/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""Models."""
```
ryanjulian: this package is very exciting :)

garage/tf/models/autopickable.py
ryanjulian: generally underscores follow whatever CamelCase you use in the class name

`Autopickleable` --> `autopickleable.py`
`AutoPickleable` --> `auto_pickleable.py`

garage/tf/models/autopickable.py
```diff
@@ -0,0 +1,75 @@
...
+        else:
+            # when unpickling
+            return __pickle_target(*args, **kwargs)
+
+    def __getnewargs__(self):
+        """object.__getnewargs__."""
+        return (None, self.args, self.kwargs, self.to_call)
+
+
+def build_layers(config, weights, **kwargs):
```
ryanjulian: hmm should this be _ private?

garage/tf/models/autopickable.py
```diff
@@ -0,0 +1,75 @@
...
+    """
+
+    def __getstate__(self):
+        """object.__getstate__."""
+        state = self.__dict__.copy()
+        for k, v in self.__dict__.items():
+            if isinstance(v, tf.keras.models.Model):
+                custom_objects = {}
+                for c in state[k].layers:
+                    if "garage" in str(type(c)):  # detect subclassed layer
```
ryanjulian: ohh this is strange

@krzentner

garage/tf/models/autopickable.py
ahtsan: Yes

garage/tf/core/mlp2.py
```diff
@@ -0,0 +1,80 @@
+"""MLP model in TensorFlow using tf.keras.models.Model."""
+import tensorflow as tf
+from tensorflow.keras.layers import BatchNormalization
+from tensorflow.keras.layers import Dense
+from tensorflow.keras.models import Model
+
+
+def mlp2(input_var,
```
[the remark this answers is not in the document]
ahtsan: It would be a model but whether we want it to be stored in a class or returned by a function?

garage/tf/models/autopickable.py
ahtsan: Both custom layers and keras layers inherit tf.keras.layers.Layer so I couldn't do class check. I assume custom class must be implemented in garage codebase so has garage in their class name.

garage/tf/core/mlp2.py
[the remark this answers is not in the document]
ryanjulian: it is a stylistic preference. 

`mlp2()` is technically a factory function, since it returns an instance of a class.

I think it is probably more straightforward to make it a class. That makes it obvious to the user that they are constructing an object and not just calling a function.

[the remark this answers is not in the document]
ahtsan: Ok. I think that would become the `PickableMLP` inside `test_model_autopickable.py`

garage/tf/core/parameterLayer.py
ahtsan: TensorFlow handles variable handling by variable scope, i.e. if two variables with same name are declared under the same variable, they share the same variable/weight. 

Keras handles variable handling in a different way. Variable sharing should be done via calling a same Keras layer (or model) instance multiple times. **A TensorFlow variable scope will have no effect on a Keras layer or model**. (Reference: [Keras](https://blog.keras.io/keras-as-a-simplified-interface-to-tensorflow-tutorial.html))

I am not completely positive whether tf.Keras follows this, but for sure we can use `tf.name_scope` for better graph visualization.

garage/tf/models/base.py
ahtsan: If we just want to return the config for the underlying model, since we construct every model with the Functional API I think we can do the "Aggregation" way:
```python
class SomeModel:
    def __init__(...):
        self.model = self.build_model()

    @property
    def input:
        return self.model.input

    def get_config():
        return self.model.get_config()
```
But if we want to return the config that reflects the entire model, I think that's not easy.

CatherineSue: i c. so the pickler of `GaussianMLPModel` actually pickles the config and weights of `GaussianMLPModel.model` instead of `GaussianMLPModel`?

ahtsan: Yes. `GaussianMLPModel` is the abstract class that holds the underlying `GaussianMLPModel` as `self.model`.

ahtsan:
Should we also preserve the name_scope when unpickled? If not, it will look like this
screen shot 2019-01-27 at 2 56 02 pm
where `mlp` is the original MLP model, the others belong to the pickled one.

garage/tf/models/gaussian_mlp_model.py
CatherineSue: Could you show what a tf graph would like if you don't pass a name to the layer? It would be better to pass a name here otherwise the name in the config would be just Lambda. It is more informative to show it is a distribution layer

garage/tf/core/parameterLayer.py
CatherineSue: No, tf.keras doesn't handle scope automatically. Because a tf.keras.Layer will generate it's [unique_layer_name](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/engine/base_layer.py#L1257), so if we want it to handle the scope, we have pass it to every layers. But it seems in 2.x they will fix [this](https://github.com/tensorflow/tensorflow/issues/24079).

tests/garage/tf/models/test_model_auto_pickable.py
```diff
@@ -0,0 +1,106 @@
...
+    def test_autopickable_gaussian_mlp_pickling(self):
+        model = GaussianMLPModel(
+            input_dim=5, output_dim=2, hidden_sizes=(4, 4), init_std=2.0)
+
+        self.sess.run(tf.global_variables_initializer())
+
+        data = np.random.random((3, 5))
+
+        result_from_model = []
+        result_from_model.append(
```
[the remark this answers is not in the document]
ahtsan: Yes.

garage/tf/core/parameterLayer.py
ahtsan: so looks like we can use `tf.variable_scope` for sharing variables, and pass the name to the layers for now?

garage/tf/models/gaussian_mlp_model.py
```diff
@@ -0,0 +1,189 @@
...
+                mean_var = mean_model.output
+
+                if self._adaptive_std:
+                    std_model = MLPModel(
+                        input_dim=self._input_dim,
+                        output_dim=self._output_dim,
+                        hidden_sizes=self._std_hidden_sizes,
+                        hidden_nonlinearity=self._std_hidden_nonlinearity,
+                        output_nonlinearity=self._std_output_nonlinearity,
+                        output_b_init=b,
```
CatherineSue: Isn't `b` undefined here?

garage/tf/core/parameterLayer.py
CatherineSue: I think so but you need to test it. 
FYI, I think using name_scope here doesn't seem to work:
``` python
model = GaussianMLPModel(
            input_dim=5, output_dim=2, hidden_sizes=(4, 4), init_std=2.0, adaptive_std=True)
```
error:
```
ERROR: test_autopickable_gaussian_mlp_pickling (tests.garage.tf.models.test_model_auto_pickable.TestKerasModel)
  File "/home/chang/Code/garage/tests/garage/tf/models/test_model_auto_pickable.py", line 93, in test_autopickable_gaussian_mlp_pickling
    input_dim=5, output_dim=2, hidden_sizes=(4, 4), init_std=2.0, adaptive_std=True)
  File "/home/chang/Code/garage/garage/tf/models/gaussian_mlp_model.py", line 98, in __init__
    self.model = self.build_model()
  File "/home/chang/Code/garage/garage/tf/models/gaussian_mlp_model.py", line 189, in build_model
    outputs=[mean_var, std_var, std_param_var, sample_var])
  File "/home/chang/miniconda2/envs/garage/lib/python3.6/site-packages/tensorflow/python/keras/engine/training.py", line 111, in __init__
    super(Model, self).__init__(*args, **kwargs)
  File "/home/chang/miniconda2/envs/garage/lib/python3.6/site-packages/tensorflow/python/keras/engine/network.py", line 78, in __init__
    self._init_graph_network(*args, **kwargs)
  File "/home/chang/miniconda2/envs/garage/lib/python3.6/site-packages/tensorflow/python/keras/engine/network.py", line 244, in _init_graph_network
    self.inputs, self.outputs)
  File "/home/chang/miniconda2/envs/garage/lib/python3.6/site-packages/tensorflow/python/keras/engine/network.py", line 1714, in _map_graph_network
    str(layers_with_complete_input))
ValueError: Graph disconnected: cannot obtain value for tensor Tensor("GaussianMLPModel/input_2:0", shape=(?, 5), dtype=float32) at layer "input_2". The following previous layers were accessed without issue: []

Ran 5 tests in 2.418s

```

CatherineSue:
> Should we also preserve the name_scope when unpickled? If not, it will look like this
> screen shot 2019-01-27 at 2 56 02 pm
> where `mlp` is the original MLP model, the others belong to the pickled one.

I think this issue can be resolved by passing unique name `scope/layer_name`, i.e., `mean_network/hidden_1` to each layer. It will be automatically organized in tf graph.

garage/tf/core/parameterLayer.py
CatherineSue: tf.keras.Model and tf.keras.Layer don't handle variable_scope, because the name of the layer is assigned by passing a name value when constructing the object.
When constructing a tf.keras.Model, it will check all the layer names inside the model. If there is a hidden_1 in mean_network, another hidden_1 in std_network, the summary of the GaussianMLPModel would become:
```
Layer (type)        Output Shape         Param #     Connected to                            
__________________________________________________________________________________________________
hidden_1 (Dense)    (None, 4)            20          hidden_0[0][0]                    
__________________________________________________________________________________________________
hidden_1 (Dense)    (None, 4)            20          hidden_0[0][0]       

```
This is invalid because there are two `hidden_1` layers.
Based on the above test,we may as well conclude that they can't share the same variable by using `with tf.variable_scope("variable_scope", reuse=True)`.

garage/tf/models/base.py
ryanjulian: should this inherit from `keras.Model` (or `keras.Layer` or neither)?

@krzentner

ahtsan:
> > Should we also preserve the name_scope when unpickled? If not, it will look like this
> > screen shot 2019-01-27 at 2 56 02 pm
> > where `mlp` is the original MLP model, the others belong to the pickled one.
> 
> I think this issue can be resolved by passing unique name `scope/layer_name`, i.e., `mean_network/hidden_1` to each layer. It will be automatically organized in tf graph.

Consider a single GaussianMLPModel.

Appending `scope` to layer name will have:
screenshot from 2019-01-28 22-48-18
The custom layer `parameter_layer` will contain the `GaussianMLPModel` scope.

While using `tf.variable_scope` or `tf.name_scope` will have:
screenshot from 2019-01-28 22-50-39
where the custom layer will be inside the `GaussianMLPModel` scope.

CatherineSue:
I don't want to quote the long reply above so I'll reply here. 
That is interesting. Have you passed the name to Model too? Like Model=(name=“GaussianMLPModel”). I've tested this to work but I deleted my code.

garage/tf/models/base.py
krzentner: I think neither is slightly preferable. There's no particular reason to add the behavior of `tf.keras.Model` here.

krzentner: Yeah, I think this is good for the same reasons expressed above.

garage/tf/models/base.py
krzentner: Wait, does `tf.keras.Model` have both `input` and `inputs` properties? If so, can we make it unnecessary to implement both (probably `input`)?

garage/tf/core/parameterLayer.py
ahtsan: Right. `tf.variable_scope` is simply useless here. I tried with appending the scope to the variables name in the latest commit. It looks pretty good like this
screen shot 2019-01-29 at 7 05 37 pm
where there is one GaussianMLP with `adaptive_std` and one doesn't, and they share the input placeholder.

garage/tf/models/base.py
ahtsan: Yes, it does.

garage/tf/models/gaussian_mlp_model.py
```diff
@@ -0,0 +1,190 @@
...
+            lambda x, f, s: f(x[0], x[1]).sample(seed=s),
+            arguments={
+                "f": tf.contrib.distributions.MultivariateNormalDiag,
+                "s": ext.get_seed()
+            },
+            name=self._scope + "/sample_var")([mean_var, std_param_var])
+
+        return Model(
+            inputs=input_var,
+            outputs=[mean_var, std_var, std_param_var, sample_var])
```
CatherineSue: I think I added a comment about add `self._dist` in the last commit. What if you need to calculate entropy?

garage/tf/models/gaussian_mlp_model.py
```diff
@@ -0,0 +1,190 @@
...
+        else:
+            raise NotImplementedError
+
+        sample_var = Lambda(
+            lambda x, f, s: f(x[0], x[1]).sample(seed=s),
+            arguments={
+                "f": tf.contrib.distributions.MultivariateNormalDiag,
+                "s": ext.get_seed()
+            },
+            name=self._scope + "/sample_var")([mean_var, std_param_var])
```
CatherineSue: I still think this name style is weird. Because this looks like to pass a name_scope but slightly different. Normally if there are two same name scopes without `reuse=True`, tf graph will rename the latter one. 

If you prefer to do it this way, then please to set `scope` to a required param to prevent a messy tf graph. And please change the `scope` to `name` because it is passed to layer's name. Using `scope` is confusing to me. But making  `scope` a required param is also strange.

garage/tf/models/autopickable.py
ryanjulian: that assumption won't hold for very long. garage is used as a dependency in research repos.

we will need to find a different way to do this.

garage/tf/core/parameterLayer.py
ryanjulian: i think in this case variable sharing should be a non-issue, because models will allow us to abandon re-building the whole network over-and-over again like rllab did.

for example, GaussianMLPPolicy.entropy_sym() rebuilds the entire network (sharing weights with the original) and then returns the entropy variable. A GaussianMLPPolicy written with Models should just have a call entropy() which returns the the entropy tensor op of the underlying distribution.

garage/tf/models/base.py
ryanjulian: sorry i guess i was asking "to what extend does this interface overlap with tf.keras.Model"?

garage/tf/models/gaussian_mlp_model.py
ryanjulian: I think you will also need outputs for important distribution parameters such as entropy() or logpdf() (or some other way of accessing those, such as providing an output for the entire distribution).

garage/tf/models/mlp_model.py
```diff
@@ -0,0 +1,82 @@
...
+        self._hidden_w_init = hidden_w_init
+        self._hidden_b_init = hidden_b_init
+        self._output_nonlinearity = output_nonlinearity
+        self._output_w_init = output_w_init
+        self._output_b_init = output_b_init
+        self._batch_normalization = batch_normalization
+
+        self.model = self.build_model(input_var)
+
+    def build_model(self, input_var):
```
ryanjulian: I'm kind of confused about why this doesn't use the `tf.keras.Model` `__init__() and `call()` interfaces, but maybe I haven't been paying enough attention.

@krzentner can you explain?

garage/tf/models/autopickable.py
CatherineSue: I think using str in str is not a safe way to check this. What if user has a keras module that defines some custom layers, here the `if` check will return `False`, but the custom layers should be custom objects.

CatherineSue: Why do we check custom_objects in here instead of defining a `custom_objects` member in the custom model class then just use it here?

ahtsan: Do you mean to define the custom layers in the `GaussianMLPModel` and get by `self.custom_layers`? 
My original idea is they are detected automatically without the need of manual definition. But if we do it in that way, it would be much easier.

ryanjulian:
Thanks for working on this. Introducing distributions seems to have made some of the design fall apart.

It's not your fault, it's TensorFlows.

Let's talk tomorrow about a simpler path out of this (without Keras). I think it can be done quickly and I'm willing to help, so you don't have to be tortured by this anymore.

garage/tf/models/gaussian_mlp_model.py
```diff
@@ -0,0 +1,193 @@
...
+                            "g": tf.exp
+                        })(std_param_var)
+                else:
+                    raise NotImplementedError
+
+            with tf.variable_scope("sample_var"):
+                sample_var = DistributionLayer(
+                    tf.contrib.distributions.MultivariateNormalDiag,
+                    ext.get_seed())([mean_var, std_param_var])
+                custom_objects['DistributionLayer'] = DistributionLayer
```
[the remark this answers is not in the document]
CatherineSue: What if user defined custom objects on his own?

garage/tf/models/base.py
```diff
@@ -0,0 +1,92 @@
...
+        """Tensor inputs of the Model."""
+        return self.model.inputs
+
+    @property
+    def outputs(self):
+        """Tensor outputs of the Model."""
+        return self.model.outputs
+
+
+class PickableModel(AutoPickable):
```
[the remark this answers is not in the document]
ahtsan: Yes.

For `inputs` in model, it is defined during graph initialization.
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/engine/network.py#L224

In the latest version, they also check if inputs/outputs are singular.

garage/tf/core/distribution_layer.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def __init__(self, dist, seed=0, **kwargs):
+        kwargs['name'] = "distribution_layer"
+        super().__init__(**kwargs)
+        self._dist_callable = dist
+        self._seed = seed
+
+    def build(self, input_shape):
+        """tf.keras.layers.Layer build."""
+        super().build(input_shape)
```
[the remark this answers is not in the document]
ahtsan: https://keras.io/layers/writing-your-own-keras-layers/

It is recommended to call to make sure `self.built=True` in the layer.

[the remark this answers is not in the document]
ahtsan: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/engine/base_layer.py#L76

ryanjulian:
see #512

PR #486, opened 2019-01-30
Fix garage.spaces and enviroment wrapper bugs

ryanjulian:
* Fixes bug in garage/sampler/utils.py, where the sampler calls the
  flatten() helper (used for gym.spaces) rather than
  GarageEnv.flatten(). This is incorrect because the sampler assumes it
  is operating on a GarageEnv
* Changes GarageEnv to no longer be an ABC -- there are algorithms
  which depend only on numpy (not a NN library like TF or PyTorch), and
  those algorithms can use GarageEnv directly. No need for subclassing.
* Updates non-TF algorithm tests to use GarageEnv

### DIFF
--- a/garage/envs/base.py
+++ b/garage/envs/base.py
@@ -1,13 +1,21 @@
 """Wrapper class that converts gym.Env into GarageEnv."""
-from abc import ABCMeta, abstractmethod
 import collections
 import warnings
 
 import glfw
 import gym
+from gym.spaces import Box as GymBox
+from gym.spaces import Dict as GymDict
+from gym.spaces import Discrete as GymDiscrete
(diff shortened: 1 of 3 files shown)


garage/tf/algos/npo.py
```diff
@@ -131,7 +131,7 @@ def get_itr_snapshot(self, itr, samples_data):
             itr=itr,
             policy=self.policy,
             baseline=self.baseline,
-            env=self.env,
+            # env=self.env,
```
jonashen: No commented code

ryanjulian: thanks

ryanjulian: fixed

PR #492, opened 2019-02-01
Remove some dead files from garage/misc

ryanjulian:
This PR deletes a few files from garage/misc which are not imported
anywhere.


akumaraguru:
Looks like `garage/envs/box2d/models/{cartpole,double_pedulum}.xml.mako` depend on `mako_utils.py`. Should this be refactored, or kept as is?

ryanjulian:
I will remove it

ryanjulian:
`nb_utils.py` is part of the experiment package, actually. I will move it there :). 

it's a set of utilities for processing experiment files using Jupyter Notebook.

PR #494, opened 2019-02-02
Install MuJoCo 2.0

ryanjulian:
This PR adds MuJoCo 2.0 installation to Dockerfiles and setup scripts,
in preparation for upgrading dm_control.

We install MuJoCo 2.0 alongside MuJoCo 1.5, because gym and mujoco_py
still requre MuJoCo 1.5. We also don't edit any of the LD_PRELOAD or
LD_LIBRARY_PATH additions (yet), since dm_control uses its own lookup
logic to find MuJoCo libraries. TBD whether our linker hacks will
interfere with dm_control's operation, though.

### DIFF
--- a/docker/Dockerfile.base
+++ b/docker/Dockerfile.base
@@ -58,12 +58,18 @@ RUN git clone https://github.com/glfw/glfw.git && \
   cd ../../ && \
   rm -rf glfw
 
-# MuJoCo
-RUN mkdir /root/.mujoco && \
+# MuJoCo 1.5 (for gym)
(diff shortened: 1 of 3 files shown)


zhanpenghe:
`mujoco_py ` has not been supporting mujoco 2.0 right?

ryanjulian:
@zhanpenghe nope

naeioi:
How are we switching Mujoco version after this PR? Do gym and mujoco_py automatically find v1.5 and dm_control finds v2.0?

ryanjulian:
Correct.

Issue #359, opened 2018-10-25
Remove custom environments which duplicate openai/gym


ryanjulian:

ryanjulian:
Inventory of custom environments

Category    | Name (<>Env)           | openai/gym? | Meta- dep? | Notes |
----------- | ---------------------- | ----------- | ---------- | ----- |
Box2D       | CarParking             | No          | No         | |
Box2D       | Cartpole               | Yes         | No         | |
Box2D       | CartpoleSwingUp        | No          | No         | Pendulum-v0 is a radial joint rather than a cart-pole |
Box2D       | DoublePendulum         | Yes         | No         | Acrobot-v1 |
Box2D       | MountainCar            | Yes         | No         | |
MuJoCo      | Ant                    | Yes         | Yes        | |
MuJoCo      | HalfCheetah            | Yes         | Yes        | |
MuJoCo      | Hopper                 | Yes         | Yes        | |
MuJoCo      | Humanoid               | Yes         | No         | |
MuJoCo      | InvertedDoublePendulum | Yes         | No         | |
MuJoCo      | Point                  | No          | Yes        | |
MuJoCo      | SimpleHumanoid         | No          | No         | |
MuJoCo      | Swimmer3D              | No          | Yes        | Not clear this one is working |
MuJoCo      | Swimmer                | Yes         | Yes        | Not clear this one is working |
MuJoCo      | Walker2D               | Yes         | Yes        | |
Meta-MuJoCo | AntGather              | No          | -          | |
Meta-MuJoCo | PointGather            | No          | -          | |
Meta-MuJoCo | SwimmerGather          | No          | -          | |
Meta-MuJoCo | AntHill                | No          | -          | |
Meta-MuJoCo | HalfCheetahHill        | No          | -          | |
Meta-MuJoCo | HopperHill             | No          | -          | |
Meta-MuJoCo | Swimmer3DHill          | No          | -          | |
Meta-MuJoCo | Walker2DHill           | No          | -          | |
Meta-MuJoCo | AntMaze                | No          | -          | |
Meta-MuJoCo | PointMaze              | No          | -          | |
Meta-MuJoCo | SwimmerMaze            | No          | -          | |

ryanjulian:
Based on above, the first candidates for deletion are:
* Box2D/Cartpole
* Box2D/DoublePendulum
* Box2D/MountainCar
* MuJoCo/Humanoid
* MuJoCo/InvertedDoublePendulum

[...]

Issue #381, opened 2018-11-02
Remove examples and dead code from scripts/


ryanjulian:
Examples which live in scripts:
* resume_training.py
* sim_env.py
* sim_policy.py

Code to just delete:
* submit_gym.py

PR #495, opened 2019-02-02
Move some scripts to examples/

ryanjulian:
This PR moves a couple files from scripts/ to examples/, since they
are better classified as garage usage examples than general purpose
utility scripts.

Notably it does not touch scripts/sim_env.py. That script is only
relevant to the custom MuJoCo environments, which will be removed by
addressing #359


ryanjulian:
ref #381

PR #504, opened 2019-02-03
Upgrade dm_control to 92f9913

ryanjulian:
This PR upgrades dm_control to the latest version from master.

dm_control has been significantly upgraded since we last updated.

* Requires MuJoCo 2.0 (introduced in #494)
* Rewrites DmControlViewer for the new graphical viewer
* Updates setup scripts and Dockerfiles with new environment variables

### DIFF
--- a/garage/envs/dm_control/dm_control_env.py
+++ b/garage/envs/dm_control/dm_control_env.py
@@ -14,49 +14,51 @@ class DmControlEnv(gym.Env, Serializable):
     Binding for [dm_control](https://arxiv.org/pdf/1801.00690.pdf)
     """
 
-    def __init__(
-            self,
-            domain_name,
-            task_name,
-            plot=False,
-            width=320,
-            height=240,
-    ):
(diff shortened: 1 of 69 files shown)


garage/envs/dm_control/dm_control_env.py
```diff
@@ -14,49 +14,45 @@ class DmControlEnv(gym.Env, Serializable):
...
+            return None
+        elif mode == 'rgb_array':
+            return self._env.physics.render()
+        else:
+            raise NotImplementedError
 
     def close(self):
         if self._viewer:
-            self._viewer.finish()
+            self._viewer.close()
```
zhanpenghe: Should we also delete the reference to `self._viewer` here?

ryanjulian: Theoretically Python should do that as soon as DmControlEnv goes out of scope.

ryanjulian: i will try it

PR #507, opened 2019-02-05
WIP Bare bones models

ryanjulian:
@krzentner @ahtsan PTAL


garage/tf/models/base.py
```diff
@@ -0,0 +1,241 @@
...
+        self._name = type(self).__name__  # name default to class
+        self._inputs = None
+        self._outputs = None
+        self._default_weights = None
+        self._built = False
+
+    def build(self, *inputs):
+        """Output of model with the given input placeholder(s)."""
+
+        # Generally, you should not be rebuilding models
```
zhanpenghe: So can we rebuild a model with customized inputs?

examples/tf/test_models_ryan.py
```diff
@@ -0,0 +1,50 @@
...
+sess = tf.Session()
+with sess.as_default():
+    h = pickle.loads(h_data)
+    state2 = tf.placeholder(shape=[None, 10, 5], dtype=tf.float32)
+    outs = h.build(state2)
+
+    # sess.run(tf.global_variables_initializer())
+
+    print('weights (after pickling): {}'.format(h.weights))
+    ipdb.set_trace()
```
zhanpenghe: Hmm I think this file should be rewritten/removed to be one that conformed to the garage tests

garage/tf/models/base.py
ryanjulian: If you want to rebuild a model then you want to make a new subgraph in TF for that model. In which case you should copy the model, not rebuild the same model.

Forbidding rebuilding keeps a 1:1 Python object:TF subgraph mapping.

examples/tf/test_models_ryan.py
ryanjulian: This is just an example for Anson's benefit. He can finish the  PR

garage/tf/models/base.py
```diff
@@ -0,0 +1,241 @@
...
+        """Tensor inputs of the Model."""
+        pass
+
+    @property
+    def outputs(self):
+        """Tensor outputs of the Model."""
+        pass
+
+    @property
+    def weights(self):
```
ryanjulian: this should be called parameters

ryanjulian: and should return only a list of the params

garage/tf/models/base.py
```diff
@@ -164,30 +194,40 @@ def build(self, *inputs):
...
-        return {v.name: v for v in tf.global_variables(scope=self.name)}
+        return {
+            v.name: v
+            for v in tf.global_variables(
+                scope=self.variable_scope._name_or_scope)
+        }
+
+    def __getstate__(self):
+        if not self._is_parent:
+            raise ValueError("Child of a model should not be pickled!")
```
CatherineSue: What is the child of a model?

garage/tf/models/gaussian_mlp_model_ryan.py
```diff
@@ -163,16 +159,16 @@ def _build(self, inputs):
                         trainable=self._learn_std,
                         name="std_network")
 
-            mean_var = mean_network
-            std_param_var = std_network
+                mean_var = mean_network
```
CatherineSue: Why the indent change?

tests/garage/tf/models/test_new_model.py
```diff
@@ -0,0 +1,78 @@
...
+        data = np.random.random((3, 5))
+        self.sess.run(model.output, feed_dict={model.input: data})
+
+    def test_model_creation_with_custom_name(self):
+        input_var = tf.placeholder(tf.float32, shape=(None, 5))
+        model = GaussianMLPModel(output_dim=2)
+        model.build(input_var, name="custom_model")
+        data = np.random.random((3, 5))
+        result, result2 = self.sess.run(
+            [model.outputs[:2], model.custom_model.outputs[:2]],
```
CatherineSue: What is the purpose of having `model.custom_model`?

garage/tf/models/base.py
```diff
@@ -117,32 +104,75 @@ def outputs(self):
...
 
 
 class TfModel(Model):
-
     def __init__(self, name):
         super().__init__()
         # All TF models need a valid unique name
         # TODO(ahtsan): produce a message on collision?
         self._name = name or self._name
+        self._is_parent = True
```
CatherineSue: Could you explain the design of parent, child, siblings, and twins to me, please? I can't seem to understand them after I reviewed all the code.

garage/tf/models/base.py
```diff
@@ -117,32 +104,75 @@ def outputs(self):
...
+            if 'self' in self._siblings:
+                raise ValueError("Subgraph {} already exists!".format(
+                    self.name))
+            with self._name_scope:
+                _variable_scope = self.variable_scope if not self._is_built else tf.variable_scope(
+                    self._variable_scope._name_or_scope, reuse=True)
+                with _variable_scope:
+                    self._inputs = inputs
+                    outputs = self._build(*inputs)
+                    self._outputs = list(outputs)
```
CatherineSue: Why have to set them a list?

garage/tf/models/base.py
```diff
@@ -117,32 +104,75 @@ def outputs(self):
...
+                        if self._default_parameters:
+                            self.parameters = self._default_parameters
+                        self._is_built = True
+        elif name in self._siblings:
+            raise ValueError("Subgraph {} already exists!".format(name))
+        elif not self._is_parent:
+            raise ValueError("Child of a model should not be built!")
+        else:
+            twin = type(self)._new_child(self, name)
+            setattr(self, name, twin)
```
CatherineSue: Is it a good idea to use setattr here?

garage/tf/models/base.py
ahtsan: ### Design ###
It's a `Parent-Children` design, where parent model is the abstract model that holds the model/network structure, i.e. the variables, and the child model is spawned when given an input, to establish connectivity. This can also be interpreted as creating a symbolic model with the given input, using the same shared variables. It enforces 1-to-1 relationship in model and tf subgraph. Siblings and twins are not important.

### Usage ###
The suggested way to use the model is to first create a parent model (declare the structure of the model)
```python
model = SomeCustomModel()
```
and build the model with some given input.
The input and output of the first child model will be set as the default of the parent model.
```python
model.build(some_input_ph)
sess.run(model.outputs, feed_dict={model.inputs: some_data})
```

### Reusing the model for different input ###
This is the reason why we have child models. We can reuse the underlying variables by creating another child
```python
model.build(other_input_ph, name="another_child")
```
and access by
```python
model.another_child
```
### Scope ###
Unique name scope is required for each child model, and all of them share the same variable scope. So we cannot build a child model with the same name scope twice.
```python
model = SomeCustomModel()
model.build(some_input_ph) # it uses the default name scope
model.build(other_input_ph, name="another_child") # name scope here is "another_child"
model.build(other_input_ph) # error
```

### Pickling ###
Since the parent model contains the model/network structure, a.k.a the configuration, we only have to pickle the parent model. Therefore, child model are not supposed to be pickled.
Example:
```python
model = SomeCustomModel()
model_config = pickle.dumps(model)
...
model_restored = pickle.loads(model_config)
#establish connectivity in order to use it
model_restored.build(some_input_ph)
```

ahtsan: Sorry I should have put some docstring.

garage/tf/models/base.py
ahtsan: Maybe not necessary.

garage/tf/models/base.py
krzentner: Wait, so passing in a name to the build method basically causes a copy of the graph to be made, but calling it normally builds it in place? Or do we end up with a mis-match between the model objects and the graph?

Either way, this seems quite surprising.

garage/tf/models/base.py
```diff
@@ -0,0 +1,281 @@
...
+        # Initialize our variables so that they're guaranteed to exist after
+        # build
+        #
+        # NOTE: It's very important that we don't run
+        # `tf.global_variable_initializer()` after this step (which will call
+        # the mostly-random initializers on each variable, destroying the
+        # presets loaded from the file)
+        # @gautams has already hit this problem in his own code.
+        #
+        # TODO(ahtsan): We can deal with this by sanitizing
```
krzentner: Sanitization is not a real solution, because in practice people only run the sanitizer after they are done debugging. Since apparently people are already hitting this bug, I think we should do something else to fix it.

Is there any way we can dynamically detect that `tf.global_variable_initializer()` was called and throw an exception?

garage/tf/models/base.py
```diff
@@ -0,0 +1,281 @@
...
+
+    def _get_variables(self):
+        return {
+            v.name: v
+            for v in tf.global_variables(
+                scope=self.variable_scope._name_or_scope)
+        }
+
+    def __getstate__(self):
+        if not self._is_parent:
```
krzentner: I don't understand why this check is needed.

garage/tf/models/auto_pickable.py
```diff
@@ -0,0 +1,70 @@
...
+        return (None, self.args, self.kwargs, self.to_call)
+
+
+def _build_layers(config, weights, **kwargs):
+    model = tf.keras.models.Model.from_config(config, custom_objects=kwargs)
+    model.set_weights(weights)
+    return model
+
+
+class AutoPickable:
```
krzentner:
```suggestion
class AutoPicklable:
```

garage/tf/models/base.py
ryanjulian: We could compare the output of https://www.tensorflow.org/api_docs/python/tf/global_variables and https://www.tensorflow.org/api_docs/python/tf/report_uninitialized_variables and ensure they are 1:1.

garage/tf/models/base.py
ryanjulian: see my comments on #512 on creating separate Graph an Model classes

In short, yes, but I think we should implement it to represent the individual Graphs and the Model (family of all graphs sharing variables) with separate classes.

I'm not sure it's really possible to improve on this, since this is exactly what TensorFlow does, and it's basically the only way to share variables.

I think the improvement here is to represent the process explicitly (TF makes does this silently and it's _very_ confusing as a newbie)

Anson has revised it so that calling `build` with no arguments creates a graph called "default".

ryanjulian: One alternative (which KR might prefer) is to treat `build` like a factory which returns a `Graph`. Then people can only interconnect `Graphs` in their code, not models (i.e. models would not have inputs and outputs).

garage/tf/models/base.py
ryanjulian: not sure it's necessary

garage/tf/models/base.py
ryanjulian: @krzentner would like your thoughts

ryanjulian:
see #512 

please try to consolidate feedback there

garage/tf/models/base.py
krzentner: I don't really see why we would want to use `setattr()` here, since we could already access the graph by name using `_siblings`. If we really are going with a design like this, we should just rename `_siblings` to `instances`. Using `setattr()` looks like it's just asking for trouble down the road.

garage/tf/models/base.py
krzentner: So as I understand it, the current design lets us do this:
```
model = SomeModel()
model.build(some_input)
sess.run(model.outputs, feed_dict={model.input: some_data})

other_output = model.build(some_other_input, name='copy1')
sess.run(other_output, feed_dict={some_other_input: some_data})
```

This allows having multiple copies of the subgraph that share variables, which is important only if you're going to assemble those copies and run them together. However, as I understand it, it's also possible to achieve this by just copying the variable around? I'm afraid I may not understand Tensorflow well enough to comprehend why we're doing this.

It's also still not clear to me what happens if I deserialize multiple instances of the `Model` object, and have them all build a default subgraph. Is that just broken? If so, can we detect this dynamically and throw an error?

But generally, Ryan's most recent comment seems most appealing to me, since it makes all of the subgraph instances "the same" instead of having one "default" subgraph instance, plus any number of "siblings."

PR #510, opened 2019-02-05
Add matplotlib support to TensorBoardLogger

ryanjulian:
Note: this is pretty slow

### DIFF
--- a/examples/tf/example_tensorboard_logger.py
+++ b/examples/tf/example_tensorboard_logger.py
@@ -11,6 +11,7 @@ N = 400
 for i in range(N):
     sess = tf.Session()
     sess.__enter__()
+    logger.log("Starting epoch {}".format(i))
     k_val = i / float(N)
     logger.record_histogram_by_type('gamma', key='gamma', alpha=k_val)
     logger.record_histogram_by_type(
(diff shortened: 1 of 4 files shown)


examples/tf/example_tensorboard_logger.py
```diff
@@ -18,4 +20,14 @@
         'uniform', key='uniform', maxval=k_val * 10)
     logger.record_tabular("app", k_val)
     logger.record_histogram("gass", k_val)
+
+    # Make a random plot
+    fig = plt.figure()
```
gautams3: Since this is slow, add a line like `logger.log("Running itr {}".format(i))`

garage/misc/logger/tensorboard_output.py
```diff
@@ -128,6 +131,18 @@ def record_tensor(self, key, val):
             self._scalars.value.add(
                 tag=key + '/' + str(idx).strip('()'), simple_value=float(v))
 
+    def record_matplotlib(self, key, fig):
+        height, width = fig.get_size_inches() * fig.dpi
+        image_buffer = io.BytesIO()
+        fig.savefig(image_buffer, format='png')
```
gautams3: thanks for using compressed images (it still takes up a lot of space for 400 images in the example runner)

ryanjulian: SVG would be tiny because it's a vector format. Bitmap by far the largest because it is lossless and unencoded. PNGs very but sometimes use internal lossless compression (LZ i *think*)

PR #533, opened 2019-02-23
Move custom environments to rlworkgroup/metaworlds

ryanjulian:
This PR removes the remaining custom environments, most of which are
designed for testing meta-RL.

To use these in the future, please see https://github.com/rlworkgroup/metaworlds

We remove the Python dependencies for these environments from setup.py,
but leave the Docker and CI dependencies in-place for now, since many
projects would like to use these dependencies (e.g. MuJoCo) with
garage.

We also install the metaworlds package in environment.yml, to make it
available to garage projects, and to ensure that the Docker containers
continue to successfully install all dependencies.


naeioi:
This example uses custom mujoco environment https://github.com/rlworkgroup/garage/blob/24465c9e3f22b9af40111ab503755cda353fa2ea/examples/tf/trpo_swimmer.py#L5 Should use gym in instead.

PR #539, opened 2019-02-26
Fix GaussianMLPolicyWithModel pickling bug

ahtsan:
Previously GaussianMLPolicyWithModel cannot be pickled. To make it picklable, the operations that will create unpicklable objects (e.g. `tf.Tensor` object) need to be under `model.build()`, so that those unpicklable objects will be inside `Network`. `Model` will discard all `Network`s when pickling.

It also uses `tensor_util` so first stage refactoring would be easier.

### DIFF
--- a/garage/tf/policies/gaussian_mlp_policy_with_model.py
+++ b/garage/tf/policies/gaussian_mlp_policy_with_model.py
@@ -3,7 +3,6 @@ from akro.tf import Box
 import numpy as np
 import tensorflow as tf
 
-from garage.tf.misc import tensor_utils
 from garage.tf.models.gaussian_mlp_model import GaussianMLPModel
 from garage.tf.policies.base2 import StochasticPolicy2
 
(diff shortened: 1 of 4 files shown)


garage/tf/policies/gaussian_mlp_policy_with_model.py
```diff
@@ -77,37 +78,42 @@ def __init__(self,
...
+    def dist_info_sym(self, obs_var, state_info_vars=None, name='default'):
         """Symbolic graph of the distribution."""
         _, mean_var, log_std_var, _, _ = self.model.build(obs_var, name=name)
         return dict(mean=mean_var, log_std=log_std_var)
 
     def get_action(self, observation):
         """Get action from the policy."""
         flat_obs = self.observation_space.flatten(observation)
-        mean, log_std = [x[0] for x in self._f_dist([flat_obs])]
+        _f_dist = tensor_utils.compile_function(
```
ryanjulian: i think at this point it would be clearer to define an actual method _f_dist which just has the tf.get_default_session.run() call in it.

tests/garage/tf/policies/test_gaussian_mlp_policy_with_model.py
```diff
@@ -178,6 +178,31 @@ def test_gaussian_mlp_policy_likehihood_sym(self):
                 obs_ph2, self.policy3.model.networks['default'].dist)
 
     def test_guassian_mlp_policy_pickle(self):
```
ryanjulian: test methods are usually named as assertions, e.g.

test_is_pickleable

ryanjulian:
Please address comments and submit

Issue #516, opened 2019-02-14
Snapshot algorithms from the experiment runner


ryanjulian:
Currently, each algorithm calls the snapshotter from its `train()` method.

After this refactor, the experiment runner will snapshot the algorithm according to the chosen schedule. This requires implementing `train_once()` so that control flow returns to the experiment runner after every optimization epoch.

ryanjulian:
@naeioi is this done?

Issue #515, opened 2019-02-14
Move algorithm control flow to the experiment runner


ryanjulian:
Rather than ceding control flow by calling `train()`, the algorithms will expose `train_once()` which returns after every optimization epoch.

naeioi:
Implemented in #541 .

PR #541, opened 2019-02-26
Decouple sampler from algo

### LINKED ISSUE #514: Extract the sampler from algorithms
Refactor the algorithms so that the sampler is passed in, rather than constructed by the algorithm

naeioi:
This PR decouples sampler for algorithm by adding a `LocalRunner` that pipelines batch of samples from sampler to algorithm. 

This PR involves several interface changes,
- Add `train_once()` in algorithm. 
This method accepts one batch of preprocessed samples (i.e. processed by `process_samples()`) and use it to train policy.
- Remove `n_itr` from on-policy algorithms.
`n_epochs`, `n_epoch_cycles` from off-policy algorithms.
These parameters are not necessary for algorithm if it only needs to care about updating policy using one batch of samples. They are thus moved to `Runner.train(n_epochs, n_epoch_cycles)`.

In addition, a runner provides a default tensorflow session using python context. This is useful for components requiring a session during construction. The runner will initialize all variables that are not initialized yet before training start.

This closes #514, #515 and #516.


examples/tf/cluster_demo.py
```diff
@@ -21,30 +22,34 @@ def run_task(v):
     run_experiment can easily execute variants of the experiment on different
     machines
     """
-    env = TfEnv(env_name="CartPole-v1")
+    with LocalRunner() as runner:
```
ryanjulian: can you explain the use of a context manager here? what are you trying to express?

examples/tf/cluster_demo.py
```diff
@@ -10,6 +10,7 @@
 
 from garage.baselines import LinearFeatureBaseline
 from garage.experiment import run_experiment
+from garage.runners import LocalRunner
```
ryanjulian: can you put this in experiment? i think it is most appropriate for now. we can always split it in the future.

examples/tf/cluster_demo.py
ryanjulian: hmm now that i look at the code, maybe these should be called "trainers". i wonder, are they algorithm trainers or experiment runners

garage/runners/local_tf_runner.py
```diff
@@ -0,0 +1,228 @@
...
+    """
+
+    def __init__(self, sess=None):
+        """Create a new local runner.
+
+        Args:
+            sess: A optional tensorflow session.
+                  A new session will be created immediately if not provided.
+        """
+        self.sess = sess if sess else tf.Session()
```
ryanjulian: i recently learned you can express this as `self.sess = sess or tf.Session()`

garage/tf/algos/batch_polopt.py
```diff
@@ -46,109 +32,40 @@ def __init__(self,
...
-                logger.record_tabular('Time', time.time() - start_time)
-                logger.record_tabular('ItrTime', time.time() - itr_start_time)
-                logger.dump_tabular(with_prefix=False)
-                if self.plot:
-                    self.plotter.update_plot(self.policy, self.max_path_length)
-                    if self.pause_for_plot:
-                        input("Plotting evaluation run: Press Enter to "
-                              "continue...")
+    def initialize(self, sess=None):
+        self.sess = tf.Session() if sess is None else sess
```
ryanjulian: `self.sess = sess or tf.Session()`

garage/tf/samplers/off_policy_vectorized_sampler.py
```diff
@@ -29,16 +29,16 @@ def __init__(self, algo, n_envs=None):
         :param algo: Algorithms.
         :param n_envs: Number of parallelized sampling envs.
         """
-        super(OffPolicyVectorizedSampler, self).__init__(algo)
+        if n_envs is None:
+            n_envs = int(algo.rollout_batch_size)
+            n_envs = max(1, min(n_envs, 100))
```
ryanjulian: i get what you're doing here, but i think an `if` would be 100% clearer

also, why limit to 100? we have some big computers.

garage/tf/samplers/off_policy_vectorized_sampler.py
```diff
@@ -58,7 +58,7 @@ def shutdown_worker(self):
         self.vec_env.close()
 
     @overrides
-    def obtain_samples(self, itr):
+    def obtain_samples(self, itr, batch_size=None):
```
ryanjulian: hmm is there an algorithm which wants this option?

tests/garage/tf/algos/test_npo.py
```diff
@@ -17,34 +18,34 @@
...
+            )
+            algo = NPO(
+                env=env,
+                policy=policy,
+                baseline=baseline,
+                max_path_length=100,
+                discount=0.99,
+                gae_lambda=0.98,
+                policy_ent_coeff=0.0,
+                plot=False,
```
ryanjulian: is `plot=` really needed in the algorithm anymore? why not have the sampler plot? the plotter is basically a sampler thread with a window attached

garage/sampler/stateful_pool.py
```diff
@@ -34,10 +33,8 @@ def initialize(self, n_parallel):
             self.manager = mp.Manager()
             self.queue = mp.Queue()
             self.worker_queue = mp.Queue()
-            self.pool = MemmappingPool(
-                self.n_parallel,
-                temp_folder="/tmp",
-            )
+            self.pool = mp.Pool(self.n_parallel)
```
ryanjulian: joblib seemed like an essential part of the original design. is there a reason you are removing this? are you sure it has no performance implications?

garage/tf/algos/npo.py
```diff
@@ -119,7 +119,7 @@ def optimize_policy(self, itr, samples_data):
         logger.record_tabular("{}/Entropy".format(self.policy.name),
                               np.mean(pol_ent))
 
-        num_traj = self.batch_size // self.max_path_length
+        num_traj = min(samples_data["actions"].shape[0], 100)
```
ryanjulian: why 100??

ryanjulian:
excellent PR -- a small change but such a big difference

examples/tf/cluster_demo.py
naeioi: This is to provide a default tensorflow session for env, policy, and algorithm. We are having some mess with which session to use and where to create a session now, so I make the following assumptions. It's the most natrual solution to me.

- A session can only be created by runner.
- A default session is always readily available when policy and algo are constructed and used.
Note: The new `GaussianMLPPolicyWithModel` requires a session during construction, so creating session as we are doing now is not a good idea.
- No where in policy and algo implementation should `tf.global_variables_initializer()` be called. Instead, they only (optionally) initialize variables they own.
- The runner initializes all uninitialized variable after policy and algo are constructed and before the training starts.

naeioi: Yes, right now it is more appropriate to call it "trainer". My idea is that ultimately we remove run_experiment, and replace it with `LocalRunner`, `RemoteRunner`, etc. The Runner take care of the logic of both setting up an experiment and training it.

garage/tf/samplers/off_policy_vectorized_sampler.py
naeioi: No there's not. `obtain_samples()` is only called by runner now and the batch_size is piped from `Runner.train()` to `Sampler.obtain_sample()`. I think it's better to make it an argument of `Runner.train()` rather than to fix it in sampler or runner constrctor, so that I can call `Runner.train()` multiple times with different batch_size.

examples/tf/cluster_demo.py
ryanjulian: sgtm

tests/garage/tf/algos/test_npo.py
naeioi: No, it's not. It's a parameter of `Runner.train()` now. I'll remove it. And I think the plotter doesn't have a sampler, but it saves the current policy and obtains a single trajectory using rollout() helper function.

garage/sampler/stateful_pool.py
naeioi: I will revert it to use the original joblib pool. Tensorflow has a caveat of not being fork-safe. (See [here](https://github.com/rlworkgroup/garage/blob/7013b23a86d4b1eba159289c7e602aae9dc050cd/examples/tf/trpo_cartpole_batch_sampler.py#L22)) I was playing around to see if using `multiprocessing.Pool` would bring me any luck.

naeioi:
@ryanjulian hyperparameter search example is at https://github.com/rlworkgroup/garage/blob/decouple_sampler_from_algo/examples/tf/cluster_demo.py

garage/tf/algos/npo.py
naeioi: I was thinking that logging all actions will hurt performance. Removed it.

garage/tf/samplers/off_policy_vectorized_sampler.py
ryanjulian: Okay I see what you're saying. One thing to keep in mind is that `batch_size` is an important hyperparameter for algorithms/experiments, so eventually it will be an important parameter of the experiment configuration. You need to make sure it can be set once for an experiment and reliably followed by the rest of the code (if that's not the case right now, it can be a fix for a future PR)

i guess what i'm really asking is:

_"what is the interface for `obtain_samples`?"_

it seems right now you have settled on

_`obtain_samples` returns `n_envs * max_path_length` samples unless a `batch_size` parameter is provided, in which case it provides `batch_size` samples_

This seems mostly fine (not going to block the PR on this), but **here's why I think it might be confusing**:
Using `batch_size = n_envs * max_path_length` will not necessarily return `n_envs` trajectories, because some trajectories will end early. Your sampler will keep going until it fills `n_samples >= batch_size`, which in general will be more than `n_envs` trajectories. 

If I encountered your API in the wild and called it with no `batch_size` parameter, I think I would expect to get back `n_envs` trajectories, with those trajectories having however many samples they have (some end early)--not some unpredictable number of trajectories `n_envs * max_path_length`. After all, if I wanted that, I could have just called it myself.

I think the tension here between two definitions of how many samples you want: do you want `batch_size` time steps or do you want some number of episodes `n_episodes`? Some algorithms want a fixed number of episode trajectories, and others want a fixed batch size. There's no reason, by the way, the desired number of  trajectory episodes must be `n_episodes == n_envs`. That's just a limitation of our current code.

So perhaps what really makes sense here is two options: you can ask `obtain_samples()` for either some number of episodes `n_episodes` OR some number of samples `n_samples` -- but you must provide me with one and exactly one of those numbers.

This is perhaps fodder for another PR, but just be aware it's a real issue. To avoid mangling arguments to determine behavior, you can always split the API: those wishing to be trajectory-oriented could call `def obtain_episodes(self, itr, n_episodes)` and those wishing to be sample-oriented could call `def obtain_samples(self, itr, n_samples)`

garage/tf/samplers/on_policy_vectorized_sampler.py
```diff
@@ -12,16 +13,13 @@
 
 
 class OnPolicyVectorizedSampler(BatchSampler):
-    def __init__(self, algo, n_envs=None):
-        super(OnPolicyVectorizedSampler, self).__init__(algo)
+    def __init__(self, algo, n_envs=cpu_count()):
```
ryanjulian: i'm not sure having a default here is the best choice. defaults are good at the top-level of an API (e.g. the runner) which people actually call. If you don't expect most users to be instantiating Samplers, its better to make it a required parameter and let the code above you decide.

This is especially true for code which consumes resources (e.g. CPU) because you don't want small errors to unwittingly introduce resource-hogging bugs, which a really difficult to find.

It's much easier to debug "why am i only training on one core?" than it is "oh my god why is my machine not responding???"

That is I don't think you should have a default here. If you really need one, it should be 1.

tests/garage/tf/algos/test_npo.py
ryanjulian: Sounds good.

ryanjulian: please resolve once you do this

examples/tf/cluster_demo.py
ryanjulian: I like these assumptions, though I want to explore in the future whether a ContextManager is the best way to implement them. I am happy with this for noe

By the way, you could implement your 3rd bullet by monkey-patching `tf.global_variables_initializer()` inside the context manager:
```python
import mock

_TF_GLOBAL_VARIABLES_INITIALIZER_SYMBOLS = [
    'tf.global_variables_initializer',
    'tf.initializers.global_variables',
    'tf.initialize_all_variables',
]

def _complain(symbol):
    raise RuntimeError('garage does not supporting calling {}'.format(symbol))

class TfRunner:
    def __init__(self):

        self._patch_no_globals = [mock.patch(s, _complain(s)) for s in TF_GLOBAL_VARIABLES_INITIALIZER_SYMBOLS]

    def __enter__(self):
        for p in self._patch_no_globals:
            p.__enter__()

    def __exit__(self):
        for p in reversed(self._patch_no_globals):
            p.__exit__(self):


class LocalRunner(TfRunner):
    def __init__(self):
        super().__init__()

    def __enter__(self):
        super().__enter__()
        # stuff

    def __exit__(self):
        # stuff
        super().__exit__()
```

garage/tf/samplers/off_policy_vectorized_sampler.py
naeioi: Removed 100 limit.

ryanjulian:
One sign this is an excellent PR is that it deletes a lot of code without deleting any functionality. Kudos.

Please let me know when the tests are in and I'm happy to +1

garage/tf/samplers/off_policy_vectorized_sampler.py
naeioi: I see your points on `batch_size`. Indeed, this parameter is not well defined and our current implementation doesn't have a clear logic with respect to it. The default value of `batch_size` is in fact copied from original `OffPolicyVectorizedSampler`. It's not configurable (but should be IMO)
https://github.com/rlworkgroup/garage/blob/9ba429f3f5e2c4593170aef61f38bf94b8d1fae1/garage/tf/samplers/off_policy_vectorized_sampler.py#L73

Let's make it future PR to offer the option between `obtain_episodes()` and `obtain_samples()`.

examples/tf/cluster_demo.py
naeioi: Cool! LocalRunner depends on tensorflow now and one thing we need in the future is to support for frameworks other than tensorflow. Your snippet demonstrates a nice shape of doing it.

garage/tf/samplers/on_policy_vectorized_sampler.py
naeioi: Make a lot of sense. Changed to 1.

garage/tf/samplers/off_policy_vectorized_sampler.py
CatherineSue: This definition is like a quick fix for using VectorizedSampler, I have some concerns on sampling fixed samples. Can we discuss it on stand up?

naeioi: Sure thing. Let's talk tomorrow.

tests/benchmarks/test_benchmark_trpo.py
```diff
@@ -103,48 +104,50 @@ def run_garage(env, seed, log_dir):
...
+                ),
+            )
+
+            algo = TRPO(
+                env=env,
+                policy=policy,
+                baseline=baseline,
+                batch_size=1024,
+                max_path_length=100,
+                n_itr=976,
```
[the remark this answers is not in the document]
naeioi: You are right. They should not be there anymore. Fixed.

tests/benchmarks/test_benchmark_trpo.py
```diff
@@ -103,48 +104,50 @@ def run_garage(env, seed, log_dir):
...
+                plot=False,
+            )
+
+            # Set up logger since we are not using run_experiment
+            tabular_log_file = osp.join(log_dir, "progress.csv")
+            garage_logger.add_tabular_output(tabular_log_file)
+            garage_logger.set_tensorboard_dir(log_dir)
+
+            runner.setup(algo, env)
+            runner.train(n_epochs=876, batch_size=1024)
```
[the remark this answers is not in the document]
naeioi: Nice catch. Fixed.

ryanjulian:
@CatherineSue @krzentner please review. i think we should try to get this in ASAP before the codebase changes underneath it.

garage/tf/algos/batch_polopt.py
```diff
@@ -46,109 +32,40 @@ def __init__(self,
...
         self.fixed_horizon = fixed_horizon
-        if sampler_cls is None:
-            if self.policy.vectorized and not force_batch_sampler:
-                sampler_cls = OnPolicyVectorizedSampler
-            else:
-                sampler_cls = BatchSampler
-        if sampler_args is None:
-            sampler_args = dict()
-        self.sampler = sampler_cls(self, **sampler_args)
+        self.whole_paths = whole_paths
```
zhanpenghe: Shouldn't this be moved to sampler also?

tests/garage/experiment/test_local_tf_runner.py
```diff
@@ -0,0 +1,100 @@
...
+    #   when this test is run, tensorflow has already been initialized, and
+    #   later singleton_pool will hangs because tensorflow is not fork-safe.
+    test_batch_sampler.flaky = True
+
+    def test_train(self):
+        with LocalRunner() as runner:
+            env = TfEnv(env_name='CartPole-v1')
+
+            policy = CategoricalMLPPolicy(
+                name="policy", env_spec=env.spec, hidden_sizes=(32, 32))
```
zhanpenghe: I think the model size could be smaller if it is not a benchmark test.

ryanjulian:
i am more then happy with this PR. please get this in ASAP so the can start enjoying it (and iterating on it).

zhanpenghe:
Me too. This is neat!

zhanpenghe:
But I think we need to make things very clear on the changelog. This is an important change even if it is easy to migrate. At this point I will start working on merging the meta learning algorithms.

garage/tf/algos/batch_polopt.py
naeioi: Agreed. I move it to `Sampler.train()`.

tests/garage/experiment/test_local_tf_runner.py
naeioi: I changed the hidden size to (8,8).

naeioi:
Many thanks for your reviews @zhanpenghe @ryanjulian . I am glad to see that you are happy with this new interface. I spent some unexpected extra time with the CI last night but right now it works OK. I will have this merged soon. 

Regarding the changelog, let's put it in the next release. I think we need more time to verify and iterate so that this comes stable and and the implication is well understood :)

ryanjulian:
@zhanpenghe CHANGELOG won't get updated until the next release (v2019.06) but I think you make a good general point about us needing more documentation.

Issue #542, opened 2019-02-27
Broken tests result in green CI


ryanjulian:
See https://travis-ci.com/rlworkgroup/garage/builds/102405158

I suspect this has something to do with Docker and/or make

gonzaiva:
The root cause is that the exit status when an command fails in [check_tests.sh](https://github.com/rlworkgroup/garage/blob/master/scripts/travisci/check_tests.sh#L3) is not passed to the parent process. This can be fixed by adding `set -e` to the script.

ryanjulian:
Can you point me to before/after TravisCI links which show this fixes the problem?

gonzaiva:
How can I re-run your build to see if it fails now that is merged? Since all the tests in nose2 are passing [right now](https://travis-ci.com/rlworkgroup/garage/builds/102543187), should I introduce an error in the tests to make sure is detected?

ryanjulian:
thanks. please let me know when it finishes

gonzaiva:
Logs are here. For both, I failed the test `test_output_shape` on purpose. The first one doesn't consider my fix and the CI still passes:
https://travis-ci.com/rlworkgroup/garage/builds/102551165
Second one considers my fix and fails the CI:
https://travis-ci.com/rlworkgroup/garage/builds/102553337

PR #544, opened 2019-02-27
Exit with a non-zero status in check_tests script

gonzaiva:
Both scripts check_test and check_cronjob_test are used when running
Travis CI, and it was observed that the CI was still passing if any of
the commands inside the scripts failed.
The argument -e is now set in both scripts to exit the scripts with non-
zero status in case any of their commands fail.


ryanjulian:
ref #542

ryanjulian:
please backport to release 2018.10 (if relevant)

PR #549, opened 2019-02-28
Packaging improvements

ryanjulian:
* Set `license` in setup.py to a simple string (reading the file was an
  error because setuptools does not support multi-line strings in that
  field)
* Add automatic versioning script


gonzaiva:
Is it failing due to the cached image?

ryanjulian:
I think it's because I forgot to copy garage/__version__.py in the Dockerfile.base

PR #554, opened 2019-03-02
Fix replay buffer dtype

ahtsan:
Replay buffer should not have a default dtype, since each of the element in the replay buffer should have dtype same as the source, e.g. observation should have dtype same as env.observation_space. One example is DQN with pixel environment, we want the observation in replay buffer to be type `np.uint8`, same as the observation space.

### DIFF
--- a/tests/garage/replay_buffer/test_replay_buffer.py
+++ b/tests/garage/replay_buffer/test_replay_buffer.py
@@ -10,7 +10,7 @@ class TestReplayBuffer(unittest.TestCase):
         obs = env.reset()
         replay_buffer = SimpleReplayBuffer(
             env_spec=env, size_in_transitions=100, time_horizon=1)
-        replay_buffer.add_transition(
+        replay_buffer.add_transitions(
             observation=[obs], action=[env.action_space.sample()])
         sample = replay_buffer.sample(1)
         sample_obs = sample['observation']
--- a/tests/garage/replay_buffer/test_replay_buffer.py
+++ b/tests/garage/replay_buffer/test_replay_buffer.py
@@ -1,0 +1,20 @@
+import unittest
+
+from garage.replay_buffer import SimpleReplayBuffer
+from tests.fixtures.envs.dummy import DummyDiscreteEnv
+
+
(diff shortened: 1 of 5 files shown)


tests/garage/replay_buffer/test_replay_buffer.py
```diff
@@ -0,0 +1,20 @@
...
+from tests.fixtures.envs.dummy import DummyDiscreteEnv
+
+
+class TestReplayBuffer(unittest.TestCase):
+    def test_replay_buffer_dtype(self):
+        env = DummyDiscreteEnv()
+        obs = env.reset()
+        replay_buffer = SimpleReplayBuffer(
+            env_spec=env, size_in_transitions=100, time_horizon=1)
+        replay_buffer.add_transition(
```
ryanjulian: if this API operates on collections instead of single values, shouldn't it be `add_transitions`?

ahtsan: I think the API works specifically for `VecEnvExecutor`, where it adds a single transition for all of the n `vec_env`, resulting in a list of length n. If I have two `vec_env`, I will be calling `replay_buffer.add_transition(observation=[obs1, obs2], action=[act1, act2])`. But I think this assumption is not clear. We should add docstring that this replay buffer only works this way, rename it to something like `VecEnvReplayBuffer` and create another replay buffer which works with a single environment.
@CatherineSue Please correct me if I am wrong.

CatherineSue: No there is no such assumption.

CatherineSue: It adds a collection of transitions instead of a single transition

ahtsan: I see, we can also add more than one transition even when there is only one `vec_env`. I will rename the api and submit.

ryanjulian: if you would like to also have the `add_transition` API, it's trivial to define given `add_transitions` (just wrap the args in a list and call `add_transitions`)

PR #556, opened 2019-03-04
More environment wrappers and tests

ahtsan:
Introduced more environment wrappers for DQN. Also modified some existing dummyEnvs for better testing.


garage/envs/wrappers/fire_reset.py
```diff
@@ -0,0 +1,28 @@
+"""Fire reset wrapper for gym.Env."""
+import gym
+
+
+class FireReset(gym.Wrapper):
```
ryanjulian: again, this wrapper seems really environment-specific, which is no the intended purpose of wrappers.

It seems strange to me that you would even have a wrapper which takes a specific action on reset. Isn't that the role of the RL agent?

garage/envs/wrappers/noop.py
```diff
@@ -0,0 +1,34 @@
+"""Noop wrapper for gym.Env."""
+import gym
+import numpy as np
+
+
+class Noop(gym.Wrapper):
+    """
```
ryanjulian: Can you explain that this is used for?

tests/fixtures/envs/dummy/dummy_discrete_pixel_env.py
```diff
@@ -1,31 +1,60 @@
...
 from tests.fixtures.envs.dummy import DummyEnv
 
 
 class DummyDiscretePixelEnv(DummyEnv):
-    """A dummy discrete environment."""
+    """
+    A dummy discrete pixel environment.
+
+    It follows Atari game convention, where actions are 'NOOP', 'FIRE', ...
+    It also contains self.unwrapped.ale.lives, get_action_meanings for testing.
```
ryanjulian: are you using openai/gym for this, or introducing ALE support?

tests/garage/envs/wrappers/test_clip_reward.py
```diff
@@ -0,0 +1,25 @@
...
+
+from garage.envs.wrappers import ClipReward
+from tests.fixtures.envs.dummy import DummyRewardBoxEnv
+
+
+class TestClipReward(unittest.TestCase):
+    def test_clip_reward(self):
+        # reward = 10 when action = 0, otherwise -10
+        env = DummyRewardBoxEnv()
+        env_wrap = ClipReward(env)
```
ryanjulian: i'm not sure this is appropriate in a wrapper.

if an RL agent would like to clip the environment rewards, it can do that itself before processing the samples. this clipping is part of the algorithm itself, and it's important it be put right next to the rest of the algorithm code.

it seems like you are slowly building a sample preprocessing pipeline by stacking wrapper after wrapper, but that's not really the intended use of wrappers. they are just there to allow quick and algorithm-agnostic adaptation of parts of the environment interface.

ryanjulian: @naeioi examine this PR for an example of a preprocessing use case

ahtsan: The above wrappers are all used in Atari games. Perhaps I should make them a specific function call for any Atari env, rather than environment wrappers.

CatherineSue:
> i think we should probably chat about this before you keep working on it.

I agree. I have the same questions as most of the comments.

tests/fixtures/envs/dummy/dummy_discrete_pixel_env.py
```diff
@@ -1,31 +1,60 @@
...
     @property
     def action_space(self):
         """Return an action space."""
-        return gym.spaces.Discrete(2)
+        return gym.spaces.Discrete(5)
+
+    def _get_action_meanings(self):
+        return ['NOOP', 'FIRE', 'SLEEP', 'EAT', 'PLAY']
+
+    def get_lives(self):
```
ryanjulian: i don't see the point of this method -- it's only used once in the constructor and it's not part of the wrapper interface

tests/fixtures/envs/dummy/dummy_discrete_pixel_env.py
```diff
@@ -1,31 +1,60 @@
...
 
     def __init__(self, random=True):
         super().__init__(random)
+        self.unwrapped.get_action_meanings = self._get_action_meanings
+        self.unwrapped.ale = mock.Mock()
+        self.unwrapped.ale.lives = self.get_lives
+        self._shape = (10, 10, 3)
+        self._observation_space = gym.spaces.Box(
+            low=0, high=255, shape=self._shape, dtype=np.uint8)
+        self.step_called = 0
```
ryanjulian: this should be _private

ryanjulian: why is this variable here? it seems only for testing?

you can use mock to count how many times something is called

tests/fixtures/envs/dummy/dummy_discrete_pixel_env.py
```diff
@@ -1,31 +1,60 @@
...
 
     @property
     def observation_space(self):
         """Return an observation space."""
-        self.shape = (10, 10, 3)
-        self._observation_space = gym.spaces.Box(
-            low=0, high=255, shape=self.shape, dtype=np.uint8)
         return self._observation_space
 
+    @observation_space.setter
```
ryanjulian: observation spaces do not have setters.

garage/envs/wrappers/noop.py
```diff
@@ -0,0 +1,34 @@
...
+    Args:
+        env: The environment to be wrapped.
+        noop_max: Maximum number no-op to be performed on reset.
+    """
+
+    def __init__(self, env, noop_max=30):
+        super().__init__(env)
+        self.noop_max = noop_max
+        self.noop_action = 0
+        assert noop_max > 0, "Noop_max should be larger than 0!"
```
ryanjulian: noop_max

garage/envs/wrappers/noop.py
```diff
@@ -0,0 +1,34 @@
...
+    No-op is assumed to be action 0.
+
+    Args:
+        env: The environment to be wrapped.
+        noop_max: Maximum number no-op to be performed on reset.
+    """
+
+    def __init__(self, env, noop_max=30):
+        super().__init__(env)
+        self.noop_max = noop_max
```
ryanjulian: these should be private

tests/garage/envs/wrappers/test_clip_reward.py
```diff
@@ -0,0 +1,25 @@
+import unittest
+
+from garage.envs.wrappers import ClipReward
+from tests.fixtures.envs.dummy import DummyRewardBoxEnv
+
+
+class TestClipReward(unittest.TestCase):
```
ryanjulian: never hits the `done` condition

tests/garage/envs/wrappers/test_episodic_life.py
```diff
@@ -0,0 +1,32 @@
+import unittest
+
+import numpy as np
+
+from garage.envs.wrappers import EpisodicLife
+from tests.fixtures.envs.dummy import DummyDiscretePixelEnv
+
+
+class TestEpisodicLife(unittest.TestCase):
```
ryanjulian: never hits `done` condition

tests/garage/envs/wrappers/test_fire_reset.py
```diff
@@ -0,0 +1,17 @@
+import unittest
+
+import numpy as np
+
+from garage.envs.wrappers import FireReset
+from tests.fixtures.envs.dummy import DummyDiscretePixelEnv
+
+
+class TestFireReset(unittest.TestCase):
+    def test_fire_reset(self):
```
ryanjulian: never hits `done` branch

tests/garage/envs/wrappers/test_noop.py
```diff
@@ -0,0 +1,13 @@
+import unittest
+
+from garage.envs.wrappers import Noop
+from tests.fixtures.envs.dummy import DummyDiscretePixelEnv
+
+
+class TestNoop(unittest.TestCase):
```
ryanjulian: never hits `done` branch

garage/envs/wrappers/episodic_life.py
```diff
@@ -0,0 +1,43 @@
...
+    This wrapper makes episode end when a life is lost, but only reset
+    when all lives are lost.
+
+    Args:
+        env: The environment to be wrapped.
+    """
+
+    def __init__(self, env):
+        super().__init__(env)
+        self.lives = 0
```
ryanjulian: these should be _private

garage/envs/wrappers/fire_reset.py
```diff
@@ -0,0 +1,28 @@
...
+        super().__init__(env)
+        assert env.unwrapped.get_action_meanings()[1] == 'FIRE', (
+            "Only use fire reset wrapper for suitable environment!")
+        assert len(env.unwrapped.get_action_meanings()) >= 3, (
+            "Only use fire reset wrapper for suitable environment!")
+
+    def reset(self, **kwargs):
+        """gym.Env reset function."""
+        self.env.reset(**kwargs)
+        obs, _, done, _ = self.env.step(1)
```
CatherineSue: What do we need FIRE here?

tests/fixtures/envs/dummy/dummy_discrete_pixel_env.py
CatherineSue: why is this a method?

tests/garage/envs/wrappers/test_episodic_life.py
```diff
@@ -0,0 +1,32 @@
...
+    def test_episodic_life_reset(self):
+        env = EpisodicLife(DummyDiscretePixelEnv())
+        obs = env.reset()
+
+        # env has reset
+        assert np.array_equal(obs, np.zeros(env.observation_space.shape))
+        assert env.unwrapped.ale.lives() == 5
+
+        obs, _, d, info = env.step(0)
+        assert d
```
CatherineSue: why done is `True` here?

garage/envs/wrappers/fire_reset.py
ahtsan: In most of the Atari games, the agent has to FIRE in order to start the game (like when you press START button when you play games). This wrapper avoids the agent doing random actions that have no effect on the environment right after environment reset.

tests/garage/envs/wrappers/test_episodic_life.py
ahtsan: DummyDiscretePixelEnv is designed to lose 1 life after each step. Since a life is lost, done will be True.
Also, in the wrapper design, calling env.reset() when number of lives is not 0 will result in one single step(). That's why in this test step() is only called 4 times but 5 lives are lost because 1 reset() is called.

I will add more docstring about why DummyDiscretePixelEnv is designed in this way for the ease of testing.

PR #560, opened 2019-03-05
Use the right env variable for .bash_rc

gonzaiva:
The variable BASH_PROF is not defined by setup_linux.sh, causing
problems when attempting to append a string to the file path that should
be in BASH_PROF, which expands to nothing.
The right variable to use is BASH_RC.


gonzaiva:
Tested on a fresh installation of Ubuntu 16.04

gonzaiva:
I did, only v2019.10 applies for this one

ryanjulian:
i didn't see a backport for v2018.10?

PR #561, opened 2019-03-05
Use the right env variable for .bash_rc

gonzaiva:
The variable BASH_PROF is not defined by setup_linux.sh, causing
problems when attempting to append a string to the file path that should
be in BASH_PROF, which expands to nothing.
The right variable to use is BASH_RC.


gonzaiva:
This is a backport of #560

PR #570, opened 2019-03-06
Solve ignored flake8 rules

gonzaiva:
Many flake8 rules that were previously ignored were solved and now
removed from the list of ignored rules.


garage/misc/krylov.py
```diff
@@ -3,24 +3,24 @@
 EPS = np.finfo('float64').tiny
 
 
-def cg(f_Ax, b, cg_iters=10, callback=None, verbose=False, residual_tol=1e-10):
-    """
-    Demmel p 312
-    """
+def cg(f_ax, b, cg_iters=10, callback=None, verbose=False, residual_tol=1e-10):
```
ryanjulian: perhaps just `# noqa: error-code` this file for now, since it seems like the capitalization are important for meaning in this algorithm (they correspond to notation from equations)

ryanjulian:
never mind -- let's keep scope small. don't worry about check_flake8.sh.

docstrings will still pose a challenge.

does this clean everything except for docstrings?

gonzaiva:
> never mind -- let's keep scope small. don't worry about check_flake8.sh.
> 
> docstrings will still pose a challenge.
> 
> does this clean everything except for docstrings?

Except docstrings and line breaks:
W503  # line break before binary operator
W504  # line break after binary operator

ryanjulian:
ok let's start with this and  then we can see what we can do about docstrings. it would be nice to get rid of the custom scripts.

ryanjulian:
i think we will just end up ignoring both W503 and W504 because the right usage is context-dependent. so basically we are down to docstrings.

PR #571, opened 2019-03-07
Clean up most ignored flake8 rules

ryanjulian:


garage/misc/krylov.py
```diff
@@ -3,24 +3,24 @@
 EPS = np.finfo('float64').tiny
 
 
-def cg(f_Ax, b, cg_iters=10, callback=None, verbose=False, residual_tol=1e-10):
-    """
-    Demmel p 312
-    """
+def cg(f_ax, b, cg_iters=10, callback=None, verbose=False, residual_tol=1e-10):
```
ryanjulian: Just leave these short mathematical names as-is and `# noqa: error-code` the whole file

gonzaiva: Done, although specific errors cannot be ignored per file with `# noqa: error-code`, but with the argument [per-file-ignores](http://flake8.pycqa.org/en/latest/user/options.html?highlight=per-file-ignores#cmdoption-flake8-per-file-ignores), and that's what I did.

gonzaiva:
The decrease in test coverage is because we're expanding one-liners in [garage/misc/krylov.py](https://github.com/rlworkgroup/garage/pull/571/files#diff-8c1bd89b1dc2e40ba53e834dcd37c364).

PR #572, opened 2019-03-07
Add MamlPolicy

zhanpenghe:
This is based on current implementation of Model.

### DIFF
--- a/garage/tf/policies/maml_policy.py
+++ b/garage/tf/policies/maml_policy.py
@@ -1,0 +1,153 @@
+"""Maml Policy."""
+import tensorflow as tf
+
+from garage.core import Serializable
+from garage.tf.policies.base2 import StochasticPolicy2
+
+# This is an alternative variable store for maml.
+MAML_VARIABLE_STORE = set()
+MAML_TASK_INDEX = 0
+
+
+class MamlPolicy(StochasticPolicy2):
+    """
+    MamlPolicy used by Model-Agnostic Meta-Learning.
+
+    This policy take any garage policies that use models and augment
+    them with one step adaptation tensors.
+
+    Args:
+        wrapped_policy: A garage policy using model.
+        n_tasks: Number of task for meta learning.
+        adaptation_step_size: The step size for one step adaptation.
+        name: Name of this policy.
+    """
+
+    def __init__(self,
+                 wrapped_policy,
+                 n_tasks,
+                 adaptation_step_size=0.5,
+                 name="MamlPolicy"):
+
+        self.wrapped_policy = wrapped_policy
+        self.n_tasks = n_tasks
+        self.name = name
+        self._initialized = False
+        self._adaptation_step_size = adaptation_step_size
+        self._adapted_param_store = dict()
+        self._create_update_opts()
+
+        super().__init__(wrapped_policy._env_spec)
+        Serializable.quick_init(self, locals())
+
+    def initialize(self, gradient_var, inputs):
+        """
+        Initialize the MAML Policy.
+
+        This funtion will create all the one step adapted parameters
+        and rebuild a model using these adapted parameters for each
+        task. A MamlPolicy can only be initialized once.
+
+        Args:
+            gradient_var: Gradient variables for one step adaptation.
+            inputs: Input tensors for rebuilding the models.
+
+        Returns:
+            all_model_outputs: The outputs of rebuilt models.
+            update_opts: Adapted parameter tensors.
+            update_opts_input: Input for calculating adapted parameters.
+        """
+
+        assert not self._initialized, "The MAML policy is initialized and can be initialized once."
+
+        global MAML_VARIABLE_STORE, MAML_TASK_INDEX
+        update_opts = []
+        with tf.name_scope(self.name):
+            # One step adaptation
+            with tf.name_scope("Adaptation"):
+                for i in range(self.n_tasks):
+                    params = self.wrapped_policy.get_params()
+                    gradient_i = gradient_var[i]
+
+                    for p, g in zip(params, gradient_i):
+                        adapted_param = p - self._adaptation_step_size * g
+                        name = "maml_policy/{}/{}".format(i, p.name)
+                        self._adapted_param_store[name] = adapted_param
+                        if i == 0:
+                            update_opts.append(adapted_param)
+
+            def maml_get_variable(name, shape=None, **kwargs):
+                scope = tf.get_variable_scope()
+                idx = 0
+                if MAML_TASK_INDEX >= self.n_tasks:
+                    raise ValueError("Invalid TASK_INDEX.")
+                fullname = "maml_policy/{}/{}/{}:{}".format(
+                    MAML_TASK_INDEX, scope.name, name, idx)
+                while fullname in MAML_VARIABLE_STORE:
+                    idx += 1
+                    fullname = "maml_policy/{}/{}/{}:{}".format(
+                        MAML_TASK_INDEX, scope.name, name, idx)
+                MAML_VARIABLE_STORE.add(fullname)
+                return self._adapted_param_store[fullname]
+
+            # build the model with these parameters
+            model = self.wrapped_policy.model
+
+            # overload the whole tf.get_variable function
+            # this allows us to use an operation as a variable
+            from tensorflow.python.ops import variable_scope
+            original_get_variable = variable_scope.get_variable
+            variable_scope.get_variable = maml_get_variable
+
+            all_model_outputs = list()
+            for i in range(self.n_tasks):
+                input_for_adapted = inputs[i]
+                model_infos = model.build(
+                    input_for_adapted, name="task{}".format(i))
+                MAML_TASK_INDEX += 1
+                all_model_outputs.append(model_infos)
+
+            self._initialized = True
+
+        # Use the original get_variable
+        variable_scope.get_variable = original_get_variable
+        update_opts_input = inputs[0]
+
+        return all_model_outputs, update_opts, update_opts_input
+
+    @property
+    def recurrent(self):
+        """Indicates whether the policy is recurrent."""
+        return self.wrapped_policy.recurrent
+
+    def get_action(self, observation):
+        """Get an action from the policy."""
+        return self.wrapped_policy.get_action(observation)
+
+    def get_actions(self, observations):
+        """Get actions from the policy."""
+        return self.wrapped_policy.get_actions(observations)
+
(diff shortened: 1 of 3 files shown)


zhanpenghe:
I will upload some tensor graphs later.

garage/tf/policies/maml_policy.py
```diff
@@ -0,0 +1,153 @@
...
+        self.wrapped_policy = wrapped_policy
+        self.n_tasks = n_tasks
+        self.name = name
+        self._initialized = False
+        self._adaptation_step_size = adaptation_step_size
+        self._adapted_param_store = dict()
+        self._create_update_opts()
+
+        super().__init__(wrapped_policy._env_spec)
+        Serializable.quick_init(self, locals())
```
ahtsan: Do we still need Serializable here?

zhanpenghe: Good catch. Forgot to remove it. Thanks!

zhanpenghe:
The graph that is generated by the test file.
maml_tf_graph
Adaptation step in details.
adaptation
Rebuilt models with adapted params.
rebuilt_models

garage/tf/policies/maml_policy.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def initialize(self, gradient_var, inputs):
```
ryanjulian: isn't this really just the `build()` method of a MamlModel?

tests/garage/tf/policies/test_maml_policy.py
```diff
@@ -0,0 +1,40 @@
...
+from garage.tf.policies import MamlPolicy
+from tests.fixtures import TfGraphTestCase
+from tests.fixtures.envs.dummy import DummyBoxEnv
+
+
+class TestMamlPolicies(TfGraphTestCase):
+    def test_maml_policy(self):
+        box_env = TfEnv(DummyBoxEnv())
+        with self.sess.as_default():
+            policy = GaussianMLPPolicyWithModel(
```
ryanjulian: what are some asserts you can actually test this class with? right now all it does is construct something.

note that we can mock out the wrapped policy, and only test the relationship between wrapped_policy and MamlPolicy.

ryanjulian: using a real model class makes this an integration test.

the tests will be simpler to write if you mock out `policy` (or use something more like a dummy policy) and make some assertions about the relationship between MamlPolicy and its wrapped_policy

garage/tf/policies/maml_policy.py
zhanpenghe: Yes and that was actually my previous attempt. But if we wrap a model, which is normally built during initialization, we will create many tensors (subtractions and placeholders for adaptations) that we will never use. We cannot reuse operations. The tensorboard looks kinda messy in this case. I admit that wrapping models is a better usage of models.

ryanjulian: perhaps you, Anson, and I can talk about it this afternoon. it sounds complicated-enough that we're unlikely to understand it over github comments.

tests/garage/tf/policies/test_maml_policy.py
zhanpenghe: I used a model because I would like to test the gradient flow.

zhanpenghe: Done.

ryanjulian: i recommend you test almost everything with mocking, and then add a smaller test with the full gradient. it will probably save you a lot of time over testing everything with full gradients

see for instance https://github.com/rlworkgroup/garage/blob/newer-logger/tests/garage/logger/test_tensor_board_output.py

garage/tf/policies/maml_policy.py
ryanjulian: i think anson ended up adding an initialize API to his policies as well. it's intended to be temporary while we make it so that Policy and inherit directly from model.

this LGTM as long as you follow his _initialize convention

garage/tf/policies/maml_policy.py
```diff
@@ -0,0 +1,151 @@
...
+                self.adapated_placeholders.append(ph)
+                self.update_opts.append(tf.assign(p, ph))
+
+    def update_params(self, params):
+        """Update parameters for the wrapped_policy"""
+        feed_dict = dict(zip(self.adapated_placeholders, params))
+        sess = tf.get_default_session()
+        sess.run(self.update_opts, feed_dict=feed_dict)
+
+    def get_params(self, trainable=True):
```
ryanjulian: i think you might need `__setstate__()` and `__getstate__()` here, but unsure.

ryanjulian:
tests look mostly fine. in addition to the integration test you posted, please add a more traditional unit test which mocks out the policy and just tests the logic inside MamlPolicy (there is not a lot of it, so this should be quick/easy).

For good examples of mocking in the codebase, see:
https://github.com/rlworkgroup/garage/blob/newer-logger/tests/garage/logger/test_tensor_board_output.py
https://github.com/rlworkgroup/garage/blob/newer-logger/tests/garage/logger/test_simple_outputs.py
https://github.com/rlworkgroup/garage/blob/mlp_policy_with_model/tests/garage/tf/policies/test_gaussian_mlp_policy_with_model.py

garage/tf/policies/maml_policy.py
```diff
@@ -0,0 +1,151 @@
...
+                ph = tf.placeholder(
+                    dtype=p.dtype,
+                    shape=p.shape,
+                )
+                self.adapated_placeholders.append(ph)
+                self.update_opts.append(tf.assign(p, ph))
+
+    def update_params(self, params):
+        """Update parameters for the wrapped_policy"""
+        feed_dict = dict(zip(self.adapated_placeholders, params))
```
ryanjulian: adapted_placeholders

garage/tf/policies/maml_policy.py
```diff
@@ -0,0 +1,151 @@
...
+        params = self.get_params()
+        with tf.name_scope("{}/UpdateWrappedPolicy".format(self.name)):
+            self.update_opts = []
+            self.adapated_placeholders = []
+            for p in params:
+                ph = tf.placeholder(
+                    dtype=p.dtype,
+                    shape=p.shape,
+                )
+                self.adapated_placeholders.append(ph)
```
ryanjulian: adapted

garage/tf/policies/maml_policy.py
```diff
@@ -0,0 +1,151 @@
...
+    def get_actions(self, observations):
+        """Get actions from the policy."""
+        return self.wrapped_policy.get_actions(observations)
+
+    def _create_update_opts(self):
+        """Create assign operations for updating parameters."""
+        params = self.get_params()
+        with tf.name_scope("{}/UpdateWrappedPolicy".format(self.name)):
+            self.update_opts = []
+            self.adapated_placeholders = []
```
ryanjulian: adapted

ryanjulian:
I also suggest the name MAMLPolicy (since MAML is an acronym). I think this is slightly against PEP8 but IMO pragmatic for our research field.

garage/tf/policies/maml_policy.py
```diff
@@ -0,0 +1,151 @@
...
+        with tf.name_scope(self.name):
+            # One step adaptation
+            with tf.name_scope("Adaptation"):
+                for i in range(self.n_tasks):
+                    params = self.wrapped_policy.get_params()
+                    gradient_i = gradient_var[i]
+
+                    for p, g in zip(params, gradient_i):
+                        adapted_param = p - self._adaptation_step_size * g
+                        name = "maml_policy/{}/{}".format(i, p.name)
```
ryanjulian: * single quotes
* what if the process has more than one MamlPolicy?

garage/tf/policies/maml_policy.py
```diff
@@ -0,0 +1,151 @@
...
+
+        Returns:
+            all_model_outputs: The outputs of rebuilt models.
+            update_opts: Adapted parameter tensors.
+            update_opts_input: Input for calculating adapted parameters.
+        """
+
+        assert not self._initialized, "The MAML policy is initialized and can be initialized once."
+
+        global MAML_VARIABLE_STORE, MAML_TASK_INDEX
```
ryanjulian: do these really need to be global? can it be a class variable?

garage/tf/policies/maml_policy.py
```diff
@@ -0,0 +1,151 @@
...
+                    gradient_i = gradient_var[i]
+
+                    for p, g in zip(params, gradient_i):
+                        adapted_param = p - self._adaptation_step_size * g
+                        name = "maml_policy/{}/{}".format(i, p.name)
+                        self._adapted_param_store[name] = adapted_param
+                        if i == 0:
+                            update_opts.append(adapted_param)
+
+            def maml_get_variable(name, shape=None, **kwargs):
```
ryanjulian: i think you can just make this a provide function in the module, rather than nesting it here

garage/tf/policies/maml_policy.py
```diff
@@ -0,0 +1,151 @@
...
+                scope = tf.get_variable_scope()
+                idx = 0
+                if MAML_TASK_INDEX >= self.n_tasks:
+                    raise ValueError("Invalid TASK_INDEX.")
+                fullname = "maml_policy/{}/{}/{}:{}".format(
```
ryanjulian: perhaps you can make some helper(s) like `_get_name` to replace all these arbitrary string interpolations.

you're actually trying to create an entire system of unique paths. it would be worth it to create a simple helper class/functions which you can test to make sure they satisfy your expectations for uniqueness and other properties.

garage/tf/policies/maml_policy.py
```diff
@@ -0,0 +1,151 @@
...
+                return self._adapted_param_store[fullname]
+
+            # build the model with these parameters
+            model = self.wrapped_policy.model
+
+            # overload the whole tf.get_variable function
+            # this allows us to use an operation as a variable
+            from tensorflow.python.ops import variable_scope
+            original_get_variable = variable_scope.get_variable
+            variable_scope.get_variable = maml_get_variable
```
ryanjulian: is there a reason you hot-patch this rather than using a context manager from `unittest.mock.patch`?

i'm worried this could cause strange (and difficult to find) bugs, especially if this was ever executed in multithreaded code. AFAIK this hot-patches get_variable process-wide

ryanjulian:
closing this for now. we can keep the branch.

Issue #358, opened 2018-10-25
Eliminate garage.misc


ryanjulian:
garage.misc is a dumping ground for code which should mostly be organized somewhere else. There's also a huge list of "utils" modules hiding around everywhere, many duplicating code in the others.

Here we will audit all the code in garage.misc to determined whether it should be
1. Deleted, because it's terrifying (e.g. stub) or badly-designed
1. Replaced by a library dependency
1. Moved to its own top-level module or into some existing module (e.g. garage.logger)
1. Begrudgingly accepted for now, and generate an issue to rewrite it
1. Left alone

We will algo audit all the utils modules to have as few as possible.

naeioi:
@ryanjulian Any suggestion of people I can talk to on stub()? I have no clear idea of what it is used for right now.

ryanjulian:
This is distinct from stub(). I will comment on that issue https://github.com/rlworkgroup/garage/issues/239

ryanjulian:
**[DRAFT]**

# Audit of `garage.misc`

| Module | Description | Notes | Verdict |
| - | - | - | - |
| `autoargs` | Decorator for injecting command line arguments directly into constructors | This is a tempting pattern, but it invites injecting command line arguments into arbitrary call sites. IME this often escalates into entry points with giant numbers of impossible-to-understand command line args. I prefer if we force people to specify their command line args only at the entry point.

ryanjulian:
Updated view

| Module | Description | Verdict | Destination Module | PRs |
| - | - | - | - | - |
| `console` | Console and filesystem utilities, plus some meta-programming nonsense | 🚧 Refactor | `garage.logger` | #464 |
| `ext` | ? | ? | ? |
| `krylov` | ? | ? | ? |
| `overrides` | Decorator for annotating and verifying overided methods | 📚 Replace with library | N/A |
| `special` | ? | ? | ? |
| `tensor_utils` | ? | ? | ? |
| `prog_bar_counter.py` | Paints progress bars to stdout | 🚧 Refactor | `garage.logger` | |

PR #573, opened 2019-03-07
Remove misc.autoargs and usages

### LINKED ISSUE #362: Remove garage.misc.autoargs and usages
While adding command line arguments to arbitrary compilation units is tempting, in practice it leads to lots of pain.

ryanjulian:
Fixes #362
See #358


ryanjulian:
it was previously used a lot in the envs in rlworkgroup/metaworlds

PR #574, opened 2019-03-07
Refactor *MLPPolicy with model

ahtsan:
Custom pickle logic is added, basically remove all unpickleable operations in `__getstate__` and reconstruct those in `__setstate__`.

Also the default model (`model.networks['default']`) will be built in `__setstate__` as well, so after unpickled the policy can be used right away.


garage/tf/policies/base2.py
```diff
@@ -88,6 +94,25 @@ def terminate(self):
         """Clean up operation."""
         pass
 
+    def get_trainable_vars(self, scope=None):
```
ryanjulian: why even have `scope=`?

garage/tf/policies/base2.py
```diff
@@ -88,6 +94,25 @@ def terminate(self):
...
+        """Get trainable vars."""
+        scope = scope if scope else self.name
```
ryanjulian: `self.name` is rarely the full scope, which is actually a path, right?

ryanjulian: e.g. my scope might be `Foo/Bar/Baz/Buz/str(self.name)`, but what you are doing is just `str(self.name)`

garage/tf/policies/base2.py
ahtsan: We can a situation that we only have one policy object, but there is also a target policy. We also need a way to get the variables under the target policy.
For example in DDPG
```python
self.target_policy_output = self.policy.dist_info_sym(target_obs_ph, name="target_policy")

policy_init_ops, policy_update_ops = get_target_ops(
                    self.policy.get_global_vars(),
                    self.policy.get_global_vars("target_policy"), self.tau)
```

garage/tf/policies/base2.py
ahtsan: Do you mean if the policy is wrapped by something else, it won't work as the scope will be different?

tests/garage/tf/policies/test_categorical_mlp_policy_with_model.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class TestCategoricalMLPPolicyWithModel(TfGraphTestCase):
+    def setUp(self):
+        super().setUp()
+        self.env = TfEnv(DummyDiscreteEnv())
+        self.policy = CategoricalMLPPolicyWithModel(env_spec=self.env.spec)
+        self.env.reset()
+
+    def test_categorical_mlp_policy_with_model_get_action(self):
```
ryanjulian: you dont need to include the class names in these methods, e.g. `def test_get_action(self)` is fine

tests/garage/tf/policies/test_continuous_mlp_policy_with_model.py
```diff
@@ -0,0 +1,34 @@
+"""This script creates a unittest that tests qf-derived policy."""
```
ryanjulian: docstrings are neither required nor encouraged in tests

tests/garage/tf/policies/test_categorical_mlp_policy_with_model.py
```diff
@@ -0,0 +1,49 @@
+"""This script creates a unittest that tests qf-derived policy."""
```
ryanjulian: no need for docstrings in tests

tests/garage/tf/policies/test_categorical_mlp_policy_with_model.py
```diff
@@ -0,0 +1,49 @@
...
+        for action in actions:
+            assert self.env.action_space.contains(action)
+
+    def test_categorical_mlp_policy_with_model_is_pickleable(self):
+        with tf.Session(graph=tf.Graph()):
+            p = pickle.dumps(self.policy)
+            policy_pickled = pickle.loads(p)
+            obs, _, _, _ = self.env.step(1)
+            action, _ = policy_pickled.get_action(obs)
+            assert self.env.action_space.contains(action)
```
ryanjulian: you need to assert something stronger here -- that the pre-pickling and post-pickling policy produce identical output for a given input

tests/garage/tf/policies/test_continuous_mlp_policy_with_model.py
```diff
@@ -0,0 +1,34 @@
...
+        for action in actions:
+            assert self.env.action_space.contains(np.asarray([action]))
+
+    def test_continuous_mlp_policy_with_model_is_pickleable(self):
+        with tf.Session(graph=tf.Graph()):
+            p = pickle.dumps(self.policy)
+            policy_pickled = pickle.loads(p)
+            obs, _, _, _ = self.env.step(1)
+            action, _ = policy_pickled.get_action(obs)
+            assert self.env.action_space.contains(np.asarray([action]))
```
ryanjulian: make sure both policies produce identical output for a given input

tests/garage/tf/policies/test_continuous_mlp_policy_with_model.py
```diff
@@ -0,0 +1,34 @@
...
+
+class TestContinuousMLPPolicyWithModel(TfGraphTestCase):
+    def setUp(self):
+        super().setUp()
+        self.env = TfEnv(DummyBoxEnv())
+        self.policy = ContinuousMLPPolicyWithModel(env_spec=self.env.spec)
+        self.env.reset()
+
+    def test_continuous_mlp_policy_with_model_get_action(self):
+        obs, _, _, _ = self.env.step(1)
```
ryanjulian: make sure this doesn't return an obs of all zeros, or your tests are meaningless

tests/garage/tf/policies/test_continuous_mlp_policy_with_model.py
```diff
@@ -0,0 +1,34 @@
...
+        action, _ = self.policy.get_action(obs)
+        assert self.env.action_space.contains(np.asarray([action]))
+        actions, _ = self.policy.get_actions([obs])
+        for action in actions:
```
ryanjulian: isn't this just `action[0]`?

tests/garage/tf/policies/test_deterministic_mlp_policy_with_model.py
```diff
@@ -0,0 +1,34 @@
...
+        for action in actions:
+            assert self.env.action_space.contains(np.asarray([action]))
+
+    def test_deterministic_mlp_policy_with_model_is_pickleable(self):
+        with tf.Session(graph=tf.Graph()):
+            p = pickle.dumps(self.policy)
+            policy_pickled = pickle.loads(p)
+            obs, _, _, _ = self.env.step(1)
+            action, _ = policy_pickled.get_action(obs)
+            assert self.env.action_space.contains(np.asarray([action]))
```
ryanjulian: make sure you get exactly the same output for a given input

tests/garage/tf/policies/test_deterministic_mlp_policy_with_model.py
```diff
@@ -0,0 +1,34 @@
...
+
+class TestDeterministicMLPPolicyWithModel(TfGraphTestCase):
+    def setUp(self):
+        super().setUp()
+        self.env = TfEnv(DummyBoxEnv())
+        self.policy = DeterministicMLPPolicyWithModel(env_spec=self.env.spec)
+        self.env.reset()
+
+    def test_deterministic_mlp_policy_with_model_get_action(self):
+        obs, _, _, _ = self.env.step(1)
```
ryanjulian: make sure obs is not all zeros

tests/garage/tf/policies/test_gaussian_mlp_policy_with_model.py
```diff
@@ -62,28 +62,20 @@ def test_gaussian_mlp_policy_get_action(self):
...
-        assert (self.box_env.action_space.low <= actions3 <=  # noqa: W504
-                self.box_env.action_space.high)
-        assert (self.box_env.action_space.low <= actions4 <=  # noqa: W504
-                self.box_env.action_space.high)
+        assert self.box_env.action_space.contains(np.array(actions1[0]))
+        assert self.box_env.action_space.contains(np.array(actions2[0]))
+        assert self.box_env.action_space.contains(np.array(actions3[0]))
+        assert self.box_env.action_space.contains(np.array(actions4[0]))
 
     def test_gaussian_mlp_policy_kl_sym(self):
```
ryanjulian: no need to put the class names in the test methods

tests/garage/tf/policies/test_deterministic_mlp_policy_with_model.py
```diff
@@ -0,0 +1,34 @@
+"""This script creates a unittest that tests qf-derived policy."""
```
ryanjulian: no docstrings necessary

garage/tf/policies/continuous_mlp_policy_with_model.py
```diff
@@ -0,0 +1,117 @@
...
+                An activation shared by all fc layers.
+            output_nonlinearity(optional):
+                An activation used by the output layer.
+            bn(bool, optional):
+                A bool to indicate whether normalize the layer or not.
+        """
+        assert isinstance(env_spec.action_space, Box)
+        super().__init__(name, env_spec)
+        action_dim = env_spec.action_space.flat_dim
+        if input_include_goal:
```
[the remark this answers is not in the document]
ryanjulian: this case is never tested. is it even used?

garage/tf/policies/continuous_mlp_policy_with_model.py
```diff
@@ -0,0 +1,117 @@
...
+    def _initialize(self):
+        state_input = tf.placeholder(tf.float32, shape=(None, self.obs_dim))
+
+        self.model.build(state_input)
+
+        self._f_prob = tf.get_default_session().make_callable(
+            self.model.networks['default'].output,
+            feed_list=[self.model.networks['default'].input])
+
+    def get_action_sym(self, obs_var, name=None, **kwargs):
```
ryanjulian: untested

garage/tf/policies/deterministic_mlp_policy_with_model.py
```diff
@@ -0,0 +1,100 @@
...
+        self._f_prob = tf.get_default_session().make_callable(
+            self.model.networks['default'].output,
+            feed_list=[self.model.networks['default'].input])
+
+    @property
+    def vectorized(self):
+        """Vectorized or not."""
+        return True
+
+    def get_action_sym(self, obs_var, name):
```
ryanjulian: untested

garage/tf/policies/base2.py
ryanjulian: this example is really confusing to me. if there's a regular policy and a target policy, then why don't i have `self.policy` and `self.target_policy`?

garage/tf/policies/base2.py
ryanjulian: we have no idea which outer scope the policy was created in. there's no reason i can't do

```
with LocalRunner():
    with tf.variable_scope('arbitrary'):
        policy = GaussianMLPPolicy(name='foo')  # scope is `arbitrary/foo`

    PPO(policy=policy, env=env)
```

CatherineSue:
We should probably test the refactored policies with benchmark. It prevents from further debugging the benchmark test.

ryanjulian:
@CatherineSue I agree.

but I think if we can figure out good unit tests for policies, we won't worry so much about messing them up when we refactor.

ryanjulian:
for instance, maybe we can define some unit tests which run basic supervised learning using each model and make sure it produces known-good results.

if GaussianMLPModel can properly fit a `(state, action)` dataset, then it seems unlikely the optimization of `GaussianMLPPolicy` would get messed up

garage/tf/models/continuous_mlp_model.py
```diff
@@ -0,0 +1,70 @@
+"""Discrete MLP Model."""
```
CatherineSue: Should this be Continuous MLP Model?

garage/tf/policies/categorical_mlp_policy_with_model.py
```diff
@@ -0,0 +1,110 @@
...
+        """Symbolic graph of the distribution."""
+        prob = self.model.build(obs_var, name=name)[0]
+        return dict(prob=prob)
+
+    @overrides
+    def dist_info(self, obs, state_infos=None):
+        """Distribution info."""
+        return dict(prob=self._f_prob(obs))
+
+    # The return value is a pair. The first item is a matrix (N, A), where each
```
CatherineSue: Since we are refactoring I think we should delete this incorrect comment. Or change it as the second item is also (N, A). Please correct me if I'm wrong.

garage/tf/policies/continuous_mlp_policy_with_model.py
```diff
@@ -0,0 +1,117 @@
...
+"""
+from akro.tf import Box
+import tensorflow as tf
+
+from garage.misc.overrides import overrides
+from garage.tf.models.continuous_mlp_model import ContinuousMLPModel
+from garage.tf.policies.base2 import Policy2
+
+
+class ContinuousMLPPolicyWithModel(Policy2):
```
CatherineSue: Should this be DeterministicMLPPolicyWithModel? I think in the future we should rename it.

garage/tf/policies/continuous_mlp_policy_with_model.py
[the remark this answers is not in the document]
CatherineSue: It should be used in HER. I agree we should move it to algorithm. This issue will be addressed once the new sampler API is finished.

garage/tf/policies/base2.py
ahtsan: The current design without model is that the policy object contained the network structure, and calling `build_net()` gave us a new network. It reuses the network structure.

With model, we will need to pass two policies to the algorithm in this case.

Can we have something that also allows us to reuse the network structure (model)? Instead of passing n policies, we then just need one policy and build n times.

garage/tf/policies/base2.py
ahtsan: right. Perhaps getting variables under `tf.get_variable_scope` will do the job?

garage/tf/policies/deterministic_mlp_policy_with_model.py
```diff
@@ -0,0 +1,100 @@
...
+        """Symbolic graph to the network."""
+        return self.model.build(obs_var, name=name)
+
+    def _f_prob(self, observations):
```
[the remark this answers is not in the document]
ahtsan: This function should be removed (I forgot to). We need a field to store the make_callable.

garage/tf/policies/base2.py
CatherineSue: Do we need this function with model?

garage/tf/policies/deterministic_mlp_policy_with_model.py
```diff
@@ -0,0 +1,100 @@
+"""DeterministicMLPPolict with model."""
+from akro.tf import Box
+import tensorflow as tf
+
+from garage.misc.overrides import overrides
+from garage.tf.models.discrete_mlp_model import DiscreteMLPModel
+from garage.tf.policies.base2 import Policy2
+
+
+class DeterministicMLPPolicyWithModel(Policy2):
```
CatherineSue: What is the difference of DeterministicMLPPolicyWithModel and ContinuousMLPPolicyWithModel? We should only keep one in the refactoring if they are the same. I also don’t think DeterministicPolicy should only work with Box space.

garage/tf/policies/base2.py
ryanjulian: uhh i'm not sure exactly what you mean, but if you are suggesting we search the variable_scope space then definitely not.

your method as-defined is a strange mix of an instance method and a class method, and i think it's really confusing.

for an instance method, you want to satisfy the signature:
```python
class Policy:

    """
    note there are no arguments (neither optional nor required) because 
    Policy should already know exactly which scope(s) its variables live in. if 
    MyPolicy doesn't know where its variables live, how on earth would i know???
    why would I define a Policy object at all if it's not going to keep track of that?
    """
    def get_trainable_vars(self):

        # magic

p = MyPolicy()
trainable = p.get_trainable_vars() 
```

```python
class Policy:

    """
    here we ask the user to keep track of the scope. this makes sense as a classmethod,
    though i'm not sure why anyone would actually want to use it this way (perhaps as a
    private helper for child classes)

    note scope is *required* because Policy either knows the scope or it doesn't. having
    something like "try to guess the scope if i don't tell you (and barf if you're wrong) but also
    use whatever scope i tell you even if it's wrong" is very surprising behavior.
    """
    @classmethod
    def get_trainable_vars(cls, scope):
        # magic
```

ryanjulian: tl;dr: your policy either needs to keep around a VariableScope object somehow or you need to figure out a different way of doing this

for instance, you can have `Model.get_trainable_vars()` and then have a policy concatenate all the variables of its models:

```python
class MyPolicy:

    def get_trainable_vars(self):
        vars = []
        vars.extend(self._model1.get_trainable_vars())
        vars.extend(self._model2.get_trainable_vars())

        return vars
```

If you wanted this to be magically-implemented by `Policy` for all policies, you would have to force the child classes to maintain some registry of models so that the parent class can query all of them
```python
class Policy:

    def get_trainable_vars(self):
        vars = []
        for m in self._models:
            vars.extend(m.get_trainable_vars())

        return vars

class MyPolicy(Policy):

    def __init__(self):
        model = MyModel()
        self._add_model(model)
```

Alternatively, you can make sure that there's an enclosing scope for each `Policy` in which all Models are built (this makes a lot of sense). You store that VariableScope, and then use it to query for anything your models created:

```python
class Policy:

    def __init__(self, name=None):
        self._variable_scope = tf.variable_scope(name)

    def _initialize(self):
        with self._variable_scope:
            # stuff

    def get_trainable_vars(self):
        return tf.get_collection(tf.GraphKeys.TRAINABLE_VARIABLES, scope=self._variable_scope)
```

this is probably the best option, since it's what will happen once we inherit `Policy` from `Model`

garage/tf/models/continuous_mlp_model.py
ryanjulian: it would be nice of these docstrings actually described the model rather than just rehashing its name

It could just as easily read `"""A model composed only of a multi-layer perceptron (MLP), which maps real-valued inputs to real-valued outputs."""`

garage/tf/policies/continuous_mlp_policy_with_model.py
ahtsan: We do have a DeterministicMLPPolicyWithModel.

I agree that the naming needs to be consistent and more precise. Except having `input_include_goal` as constructor argument, `ContinuousMLPModel` is same as `DiscreteMLPModel` with the output being scaled. 

Should we have a single `MLPModel` with output scale default to be 1? So that we can have the only one `MLPModel` to be used everywhere.

garage/tf/policies/base2.py
ahtsan: Without this function, algorithm would need to have n policy object. Or we can do `model.copy()` when we need duplicate (e.g. target network).

garage/tf/policies/base2.py
ahtsan: It makes a lot of sense. The VariableScope of the `Model` will later become the VariableScope of the `Policy`.

garage/tf/policies/deterministic_mlp_policy_with_model.py
ahtsan: ContinuousMLPPolicyWithModel uses ContinuousMLPModel which scales the output. I suggest we just keep one MLPModel which has default `output_scale` = 1.

I think discussion are needed for how precise we want to have the policy to be. Do we need to construct all the 4 possible policies? i.e. `Discrete/Continuous` * `Deterministic/Stochastic`.

garage/tf/policies/base2.py
ryanjulian: what does `n` equal?

ahtsan: Usually n would be 2 (trainable policy and target policy), e.g. in DDPG and DQN. If we also do the same for QFunction, it would be 2 policies and 2 QF. For TD3, it would be 2 policies and 4 QF.

ryanjulian: Then I think we should copy the model. If an algorithm needs to maintain multiple copies of parameters, it needs multiple copies of the model. 1:1.

garage/tf/policies/deterministic_mlp_policy_with_model.py
CatherineSue: First, the `output_scale` is not important/necessary there because the action value will be scaled in the environment wrapper. 
Second, `ContinuousMLPPolicy` and `DeterministicMLPPolicy` are the same in concept. We have two of them because the `DeterministicMLPPolicy` was implemented long ago and never tested. The goal is to keep only one of them eventually. 
Third, we can follow our current design, based on the type of action space, the stochastic policy will model the corresponding type of distribution. And by default, the deterministic policy is for continuous space.

ahtsan: Let's have only `DeterministicMLPPolicyWithModel` and remove `output_scale`?

garage/tf/policies/base2.py
ryanjulian: which option did you choose?

ahtsan: Second approach, which is to build all models with the same VariableScope and used that to query for anything.

ryanjulian: do you mean that two models will share a variable scope? that would be a violation of the abstraction.

they can share an outer scope, though.

ahtsan: I meant outer scope

ryanjulian: building in the same outer scope is good, but you still need each model to have a unique inner scope. there's no way around it without violating the abstraction.

if i already have a machine for bundling together all the variables inside a single scope (i.e. a `Model`), why would I start trying to look up those variables using scopes/names again? the whole point of making `Model` was to never do that again.

garage/experiment/local_tf_runner.py
```diff
@@ -9,7 +9,7 @@
 import tensorflow as tf
 
 from garage.misc.logger import logger
-from garage.sampler.parallel_sampler import singleton_pool
+from garage.sampler.stateful_pool import singleton_pool
```
ryanjulian: rebase?

garage/tf/policies/base2.py
```diff
@@ -88,6 +95,22 @@ def terminate(self):
...
+
+    def get_global_vars(self):
+        """Get global vars."""
+        return self._variable_scope.global_variables()
+
+    def get_regularizable_vars(self):
+        """Get regularizable vars."""
+        reg_vars = [
+            var for var in self.get_trainable_vars()
+            if 'W' in var.name and 'output' not in var.name
```
ryanjulian: is this a TensorFlow convention?

garage/experiment/local_tf_runner.py
ahtsan: I talked to Keren and this is a bug. It wasn't detected in CI.

tests/garage/tf/policies/test_gaussian_mlp_policy_with_model.py
```diff
@@ -105,57 +100,61 @@ def test_gaussian_mlp_policy_kl_sym(self):
         kl2 = self.sess.run(
             objective2,
             feed_dict={
-                self.policy3.model.networks['default'].input: self.obs,
-                self.policy4.model.networks['default'].input: self.obs
+                self.policy3.model.networks['default'].input: self.zeros_obs,
```
ryanjulian: hmm shouldn't this be true for *any* input, not just the zero input?

tests/garage/tf/policies/test_gaussian_mlp_policy_with_model.py
```diff
@@ -105,57 +100,61 @@ def test_gaussian_mlp_policy_kl_sym(self):
...
-    def test_gaussian_mlp_policy_log_likehihood_sym(self):
+    def test_log_likehihood_sym(self):
         # log_likelihood_sym
         # * Existing way *
         log_prob_sym1 = self.policy1.distribution.log_likelihood_sym(
             self.obs_ph, self.dist1_sym)
         log_prob_func = tensor_utils.compile_function([self.obs_ph],
                                                       log_prob_sym1)
-        log_prob1 = log_prob_func(self.obs)
+        log_prob1 = log_prob_func(self.zeros_obs)
```
ryanjulian: shouldn't this test work for any input?

tests/garage/tf/policies/test_gaussian_mlp_policy_with_model.py
```diff
@@ -105,57 +100,61 @@ def test_gaussian_mlp_policy_kl_sym(self):
...
-    def test_gaussian_mlp_policy_entropy_sym(self):
+    def test_policy_entropy_sym(self):
         # entropy_sym
         # * Existing way *
         entropy_sym1 = self.policy1.distribution.entropy_sym(
             self.dist1_sym, name='entropy_sym1')
         entropy_func = tensor_utils.compile_function([self.obs_ph],
                                                      entropy_sym1)
-        entropy1 = entropy_func(self.obs)  # result
+        entropy1 = entropy_func(self.zeros_obs)  # result
```
ryanjulian: shouldn't this test work for any input?

tests/garage/tf/policies/test_gaussian_mlp_policy_with_model.py
```diff
@@ -105,57 +100,61 @@ def test_gaussian_mlp_policy_kl_sym(self):
...
-    def test_gaussian_mlp_policy_likehihood_sym(self):
+    def test_likehihood_sym(self):
         # likelihood_ratio_sym
         # * Existing way *
         likelihood_ratio_sym1 = self.policy1.distribution.likelihood_ratio_sym(
             self.obs_ph, self.dist1_sym, self.dist2_sym, name='li_ratio_sym1')
         likelihood_ratio_func = tensor_utils.compile_function(
             [self.obs_ph], likelihood_ratio_sym1)
-        likelihood_ratio1 = likelihood_ratio_func(self.obs)
+        likelihood_ratio1 = likelihood_ratio_func(self.zeros_obs)
```
ryanjulian: any input?

garage/tf/policies/base2.py
ryanjulian: can you send me a link to code showing how you solved this?

garage/experiment/local_tf_runner.py
ryanjulian: yeah i just hit this myself

garage/tf/policies/base2.py
```diff
@@ -88,6 +95,22 @@ def terminate(self):
...
 
+    def get_trainable_vars(self):
+        """Get trainable vars."""
+        return self._variable_scope.trainable_variables()
+
+    def get_global_vars(self):
+        """Get global vars."""
+        return self._variable_scope.global_variables()
+
+    def get_regularizable_vars(self):
```
ryanjulian: is this tested?

garage/tf/policies/base2.py
ryanjulian: i.e. what makes `'W'` and `'output'` special?

tests/garage/tf/policies/test_gaussian_mlp_policy_with_model.py
ahtsan: It should be. I wanted the data to have zero mean, since the weights of the mean network is different. 
I think one way is to copy the mean network weights of policy1 to policy3, and policy2 to policy4 so the output for mean network would be the same.

tests/garage/tf/policies/test_gaussian_mlp_policy_with_model.py
ahtsan: yes, I will change that.

tests/garage/tf/policies/test_gaussian_mlp_policy_with_model.py
ahtsan: yes, I will change that.

tests/garage/tf/policies/test_gaussian_mlp_policy_with_model.py
ahtsan: yes, I will change that.

tests/garage/tf/policies/test_gaussian_mlp_policy_with_model.py
ahtsan: yes, I will change that.

tests/garage/tf/policies/test_gaussian_mlp_policy_with_model.py
ahtsan: yes, I will change that.

garage/tf/policies/base2.py
ahtsan: It existed in garage before. I will figure that out.

ryanjulian: now is a good time. it seems like it might be specific to MLPs?

CatherineSue: Yes. I don't think it is the naming convention of TensirFlow. It could be a convention in garage layers.

CatherineSue: Yes. I don't think it is the naming convention of TensirFlow. It could be a convention in garage layers.

CatherineSue: I think in TensorFlow the convention is 'kernel' and 'bias'. 
My thought on this function is, can we add regularizer to the primitives rather than here? I think tf.keras.layer support add_kernel_regularizer.

ahtsan: My thought is if we only care about the regularization loss, can we just get that from `tf.losses.get_regularization_losses` so we don't have to keep track of the regularizable variables? It returns a list of regularization losses as Tensor. (https://www.tensorflow.org/api_docs/python/tf/losses/get_regularization_losses)

CatherineSue: Good idea. I don't have a preference because they serve the same idea. But please test what you decide to use returns the same loss in ddpg.

ahtsan: @CatherineSue I agree. I think we can do `tf.layers.dense(x, kernel_regularizer=tf.contrib.layers.l2_regularizer(0.001))` followed by `loss = ordinary_loss + tf.losses.get_regularization_loss()`. Can I do that in another PR?

CatherineSue: How do you choose l1 or l2 in get_regularization_loss?

ahtsan: We pass the regularizer to the dense layer, e.g. `tf.contrib.layers.l2_regularizer(0.001)` and when we get the loss, it will be l2.

CatherineSue: I see. I think you misunderstand my proposa. By add_regularizer I mean the 'kernel_regularizer' in the argument.

ahtsan: Oh I see. By that we will use the same regularizer within a primitive?

tests/garage/tf/policies/test_gaussian_mlp_policy_with_model.py
ahtsan: Actually can I remove GaussianMLPPolicyWithModel2 at this moment first? At the first step we keep garage.Distribution first, so we will only keep GaussianMLPPolicyWithModel and make sure it works with existing GaussianMLPPolicy.

Later when we start to replace garage.Distribution with tfp.distribution, I will put GaussianMLPPolicyWithModel2 back.

ryanjulian:
@CatherineSue it's your review, so you get to decide what the criteria are.

I think at this stage a couple simple unit tests would be reasonable:

0. Initialize the network weights to a known state
1. Known input/output pair
    Make sure that a known input produces the proper output
2. Empirical mean/variance
    Make sure that a known input produces the expected empirical mean/variance (by sampling n times and calculating mean/variance)

I think a test where we attempt to train each model on a known dataset is a gold standard, but perhaps is too much for this PR.

ryanjulian:
It does seem like the new version of this PR misses coverage on several significant branches:
https://codecov.io/gh/rlworkgroup/garage/compare/c4b49939df36487c49282ddb6bd9af647ea05b87...79f4770dcfd92294f026a12d16125e9e9b6dc601/diff

garage/tf/models/gaussian_mlp_model.py
```diff
@@ -5,7 +5,6 @@
 from garage.experiment import deterministic
 from garage.tf.core.mlp import mlp
 from garage.tf.core.parameter import parameter
-from garage.tf.distributions import DiagonalGaussian
```
CatherineSue: Why delete this?

ahtsan: I thought we might put the distribution object in the policy. But right now I think we should put it in the model as gaussian mlp model will always have a distribution.

CatherineSue: Thank you for the explanation. Putting the distribution in model makes more sense

ahtsan:
My thought about adding unit test is, if we have already tested `MLPModel`, then for any policy that uses `MLPModel`, we only need to test policy-level functions, like `get_action_sym` or `dist_info_sym` and make sure the output is the same as the policy output. The reason is that we already know `MLPModel` works properly.

ryanjulian:
I agree. You can implement this by mocking out the model

tests/garage/tf/models/test_gaussian_mlp_model.py
```diff
@@ -11,99 +12,105 @@ def setUp(self):
...
+            output_dim=self.output_dim, std_share_network=True)
         outputs = model.build(self.input_var)
         out, model_out = self.sess.run(
             [outputs[:-1], model.networks['default'].outputs[:-1]],
             feed_dict={model.networks['default'].input: self.obs})
 
         assert np.array_equal(out, model_out)
-        # output shape should be 2 * output_dim
+
+        # --- output shape ---
```
ryanjulian: if you find yourself creating comment blocks in your tests like this, your test is too long.

please split your tests into more test methods.

tests/garage/tf/models/test_gaussian_mlp_model.py
```diff
@@ -11,99 +12,105 @@ def setUp(self):
...
+    def test_without_std_share_network(self):
+        model = GaussianMLPModel(
+            output_dim=self.output_dim, std_share_network=False, init_std=1.1)
         outputs = model.build(self.input_var)
         out, model_out = self.sess.run(
             [outputs[:-1], model.networks['default'].outputs[:-1]],
             feed_dict={model.networks['default'].input: self.obs})
 
         assert np.array_equal(out, model_out)
+        # --- output shape ---
```
ryanjulian: this test method is way too long. test one property per method, please

tests/garage/tf/policies/test_categorical_mlp_policy_with_model.py
```diff
@@ -1,54 +1,84 @@
...
+
 class TestCategoricalMLPPolicyWithModel(TfGraphTestCase):
     def setUp(self):
         super().setUp()
         self.env = TfEnv(DummyDiscreteEnv())
-        self.policy = CategoricalMLPPolicyWithModel(env_spec=self.env.spec)
         self.env.reset()
+        self.obs, _, _, _ = self.env.step(1)
+        with mock.patch(
+                'garage.tf.policies.categorical_mlp_policy_with_model.MLPModel',  # noqa: E501
```
ryanjulian: there's no need for noqa here. just split the string with

```python
('garage.tf.policies',
 '.categorial_mlp_policy_with_model.MLPModel')
```

tests/garage/tf/policies/test_categorical_mlp_policy_with_model.py
```diff
@@ -1,54 +1,84 @@
...
+        actions, _ = self.policy.get_actions([self.obs])
         for action in actions:
             assert self.env.action_space.contains(action)
 
-    def test_is_pickleable(self):
-        obs, _, _, _ = self.env.step(1)
-        action1, _ = self.policy.get_action(obs)
-        p = pickle.dumps(self.policy)
+    def test_dist_info(self):
+        policy_prob = self.policy.dist_info([self.obs])
```
ryanjulian: i don't understand this test. it seems that you're just verifying that `policy.dist_info` returns the same thing as `self.sess.run(...)` -- but that's a tautology.

you're testing the implementation.

the correct way to test this is to **know ahead-of-time** what the **actual numerical results** should be of `self.policy.dist_info` from `self.obs`. then you can make sure the numbers returned from `dist_info` are what you expet.

tests/garage/tf/policies/test_gaussian_mlp_policy_with_model.py
```diff
@@ -1,55 +1,82 @@
...
-
-        self.obs = [self.box_env.reset()]
+        self.env = TfEnv(DummyBoxEnv())
+        self.env.reset()
+        self.obs, _, _, _ = self.env.step(1)
         self.obs_ph = tf.placeholder(tf.float32, shape=(None, 1))
 
-        self.dist1_sym = self.policy.dist_info_sym(self.obs_ph, name='p1_sym')
+        with mock.patch(
+                'garage.tf.policies.gaussian_mlp_policy_with_model.GaussianMLPModel',  # noqa: E501
```
ryanjulian: just split the string. you don't need noqa

tests/garage/tf/policies/test_gaussian_mlp_policy_with_model.py
```diff
@@ -1,55 +1,82 @@
...
+                new=SimpleGaussianMLPModel):
+            self.policy = GaussianMLPPolicyWithModel(env_spec=self.env.spec)
 
     def test_get_action(self):
-        action, _ = self.policy.get_action(self.obs)
-        assert self.box_env.action_space.contains(np.array(action[0]))
+        action, prob = self.policy.get_action(self.obs)
+        assert self.env.action_space.contains(np.array(action[0]))
+        assert action == prob['mean'] + prob['log_std'] * 0.5
```
ryanjulian: don't test the implementation. use actual numbers.

tests/garage/tf/policies/test_gaussian_mlp_policy_with_model.py
```diff
@@ -1,55 +1,82 @@
...
+
+        actions, probs = self.policy.get_actions([self.obs])
+        assert self.env.action_space.contains(np.array(actions[0]))
+        assert actions == probs['mean'] + probs['log_std'] * 0.5
 
-        actions, _ = self.policy.get_actions(self.obs)
-        assert self.box_env.action_space.contains(np.array(actions[0]))
+    def test_dist(self):
+        assert isinstance(self.policy.distribution, DiagonalGaussian)
```
ryanjulian: this test is not necessary. please remove.

tests/garage/tf/policies/test_gaussian_mlp_policy_with_model.py
```diff
@@ -1,55 +1,82 @@
...
 
     def test_dist_info_sym(self):
         _, dist1 = self.policy.get_action(self.obs)
-        dist2 = self.sess.run(
-            self.dist1_sym, feed_dict={self.obs_ph: self.obs})
 
-        assert np.array_equal(dist1, dist2)
+        dist1_sym = self.policy.dist_info_sym(self.obs_ph, name='p1_sym')
```
ryanjulian: use actual numbers

tests/garage/tf/policies/test_deterministic_mlp_policy_with_model.py
```diff
@@ -1,46 +1,73 @@
...
     def test_get_action_sym(self):
-        obs, _, _, _ = self.env.step(1)
-
         obs_dim = self.env.spec.observation_space.flat_dim
         state_input = tf.placeholder(tf.float32, shape=(None, obs_dim))
         action_sym = self.policy.get_action_sym(state_input, name="action_sym")
 
-        action1, _ = self.policy.get_action(obs)
-        action2 = self.sess.run(action_sym, feed_dict={state_input: [obs]})
+        action1, _ = self.policy.get_action(self.obs)
```
ryanjulian: use actual numbers, not comparison to private methods

tests/garage/tf/policies/test_deterministic_mlp_policy_with_model.py
```diff
@@ -1,46 +1,73 @@
...
+                env_spec=self.env.spec)
 
     def test_get_action(self):
-        obs, _, _, _ = self.env.step(1)
-        action, _ = self.policy.get_action(obs)
+        action, _ = self.policy.get_action(self.obs)
         assert self.env.action_space.contains(np.asarray([action]))
-        actions, _ = self.policy.get_actions([obs])
+
+        out = self.sess.run(
```
ryanjulian: use actual numbers

tests/garage/tf/policies/test_categorical_mlp_policy_with_model.py
```diff
@@ -1,54 +1,84 @@
...
         state_input = tf.placeholder(tf.float32, shape=(None, obs_dim))
         dist1 = self.policy.dist_info_sym(state_input, name="policy2")
 
-        action1, prob1 = self.policy.get_action(obs)
-        prob1_dist_info = self.policy.dist_info([obs])
+        action1, prob1 = self.policy.get_action(self.obs)
+        prob1_dist_info = self.policy.dist_info([self.obs])
 
-        prob2 = self.sess.run(dist1['prob'], feed_dict={state_input: [obs]})
+        prob2 = self.sess.run(
```
ryanjulian: use actual numbers. don't test the implementation.

tests/garage/tf/models/test_gaussian_mlp_model.py
```diff
@@ -11,99 +12,105 @@ def setUp(self):
...
-        model = GaussianMLPModel2(output_dim=2, std_share_network=False)
+    def test_adaptive_std(self):
+        model = GaussianMLPModel(output_dim=self.output_dim, adaptive_std=True)
         outputs = model.build(self.input_var)
         out, model_out = self.sess.run(
             [outputs[:-1], model.networks['default'].outputs[:-1]],
             feed_dict={model.networks['default'].input: self.obs})
 
         assert np.array_equal(out, model_out)
+        # --- output shape ---
```
ryanjulian: please split these into more than one test method

tests/garage/tf/policies/test_categorical_mlp_policy_with_model.py
ahtsan: Yes, I agree that is a tautology. I was wondering should I even test it.
However, since `self.policy.dist_info` is basically same as `self.sess.run(...)`, and `self.sess.run(...)` is actually the output of the underlying model. Numerical correctness is already tested in the tests of that model. Should I test it again here?

ryanjulian: yes. your test has no idea that there's a model inside the policy.

garage/tf/core/mlp.py
```diff
@@ -46,8 +46,8 @@ def mlp(input_var,
                 inputs=l_hid,
                 units=hidden_size,
                 activation=hidden_nonlinearity,
-                kernel_initializer=hidden_w_init,
-                bias_initializer=hidden_b_init,
+                kernel_initializer=hidden_w_init(),
```
ryanjulian: why change from passing the instance to passing the type?

garage/tf/models/gaussian_mlp_model.py
```diff
@@ -108,19 +126,26 @@ def _build(self, state_input):
                     np.zeros(action_dim),
                     np.full(action_dim, self._init_std_param)
                 ], axis=0)  # yapf: disable
-                b = tf.constant_initializer(b)
+
+                def p():
+                    return tf.constant_initializer(b)
```
ryanjulian: just use a lambda

garage/tf/models/gaussian_mlp_model.py
```diff
@@ -96,7 +114,7 @@ def __init__(self,
 
     def network_output_spec(self):
         """Network output spec."""
-        return ['sample', 'mean', 'std', 'std_param', 'dist']
+        return ['sample', 'mean', 'log_std', 'dist']
```
ryanjulian: the reason that this is called `std_param` is that the output is not always `log(stddev)`. there are different parameterizations, which required different procedures to recover stddev

garage/tf/models/gaussian_mlp_model.py
```diff
@@ -57,8 +61,13 @@ def __init__(self,
                  max_std=None,
                  std_hidden_sizes=(32, 32),
                  std_hidden_nonlinearity=tf.nn.tanh,
+                 std_hidden_w_init=tf.contrib.layers.xavier_initializer,
+                 std_hidden_b_init=tf.zeros_initializer,
                  std_output_nonlinearity=None,
-                 std_parameterization='exp'):
+                 std_output_w_init=tf.contrib.layers.xavier_initializer,
```
ryanjulian: please make an issue to split this model into several different ones based on architecture. there are several model architectures here:

| description | mean | std_param |
| --- | --- | --- |
| fixed variance | MLP1 | parameter (trainable or not) |
| split network | MLP1 | MLP2 |
| shared network/multi-headed |  MLP1/head1 | MLP1/head2 |

tests/garage/tf/models/test_gaussian_mlp_model.py
```diff
@@ -11,99 +12,119 @@ def setUp(self):
         super().setUp()
         self.input_var = tf.placeholder(tf.float32, shape=(None, 5))
         self.obs = np.ones((1, 5))
+        self.output_dim = 1
```
ryanjulian: i think you should test over output_dim = 1 and output_dim > 1 (e.g. you can use a parameterized test)

lots of bugs pop up when people fail to consider both cases.

tests/garage/tf/models/test_mlp_model.py
```diff
@@ -0,0 +1,25 @@
...
+from garage.tf.models import MLPModel
+from tests.fixtures import TfGraphTestCase
+
+
+class TestMLPModel(TfGraphTestCase):
+    def setUp(self):
+        super().setUp()
+        self.input_var = tf.placeholder(tf.float32, shape=(None, 5))
+        self.obs = np.ones((1, 5))
+        self.output_dim = 1
```
ryanjulian: test a couple output_dims

tests/garage/tf/policies/test_categorical_mlp_policy_with_model.py
```diff
@@ -0,0 +1,77 @@
...
+            ))
+
+
+class TestCategoricalMLPPolicyWithModel(TfGraphTestCase):
+    def setUp(self):
+        super().setUp()
+        self.env = TfEnv(DummyDiscreteEnv())
+        self.env.reset()
+        self.obs, _, _, _ = self.env.step(1)
+        model_path = 'garage.tf.policies.' + \
```
ryanjulian: split with (), not \

tests/garage/tf/policies/test_deterministic_mlp_policy_with_model.py
```diff
@@ -0,0 +1,66 @@
...
+            ))
+
+
+class TestDeterministicMLPPolicyWithModel(TfGraphTestCase):
+    def setUp(self):
+        super().setUp()
+        self.env = TfEnv(DummyBoxEnv())
+        self.env.reset()
+        self.obs, _, _, _ = self.env.step(1)
+        model_path = 'garage.tf.policies.' + \
```
ryanjulian: split with ()

tests/garage/tf/policies/test_categorical_mlp_policy_with_model.py
```diff
@@ -0,0 +1,77 @@
...
+from garage.tf.models import Model
+from garage.tf.policies import CategoricalMLPPolicyWithModel
+from tests.fixtures import TfGraphTestCase
+from tests.fixtures.envs.dummy import DummyDiscreteEnv
+
+
+class SimpleMLPModel(Model):
+    """Simple MLPModel for testing."""
+
+    def __init__(self, name, output_dim, *args, **kwargs):
```
ryanjulian: why *args and *kwargs?

tests/garage/tf/policies/test_deterministic_mlp_policy_with_model.py
```diff
@@ -0,0 +1,66 @@
...
+import tensorflow as tf
+
+from garage.tf.envs import TfEnv
+from garage.tf.models import Model
+from garage.tf.policies import DeterministicMLPPolicyWithModel
+from tests.fixtures import TfGraphTestCase
+from tests.fixtures.envs.dummy import DummyBoxEnv
+
+
+class SimpleMLPModel(Model):
```
ryanjulian: if you are going to use a fixture repeatedly, please put it in tests/fixtures rather than copying it

tests/garage/tf/policies/test_deterministic_mlp_policy_with_model.py
```diff
@@ -0,0 +1,66 @@
...
+    """Simple MLPModel for testing."""
+
+    def __init__(self, name, output_dim, *args, **kwargs):
```
ryanjulian: why *args and **kwargs?

tests/garage/tf/policies/test_categorical_mlp_policy_with_model.py
```diff
@@ -0,0 +1,77 @@
...
+            1., shape=(
+                1,
+                self.output_dim,
+            ))
+
+
+class TestCategoricalMLPPolicyWithModel(TfGraphTestCase):
+    def setUp(self):
+        super().setUp()
+        self.env = TfEnv(DummyDiscreteEnv())
```
ryanjulian: please test both single- and multi-dimensional obs/action spaces (Dummy defaults to dim=1)

tests/garage/tf/policies/test_deterministic_mlp_policy_with_model.py
```diff
@@ -0,0 +1,66 @@
...
+            1., shape=(
+                1,
+                self.output_dim,
+            ))
+
+
+class TestDeterministicMLPPolicyWithModel(TfGraphTestCase):
+    def setUp(self):
+        super().setUp()
+        self.env = TfEnv(DummyBoxEnv())
```
ryanjulian: please test both single- and multi-dimensional obs/action spaces (Dummy defaults to dim=1)

tests/garage/tf/policies/test_gaussian_mlp_policy_with_model.py
```diff
@@ -1,228 +1,78 @@
...
-
-        assert self.policy1.vectorized == self.policy2.vectorized
-        assert self.policy3.vectorized == self.policy4.vectorized
-
-    def test_gaussian_mlp_policy_get_action(self):
-        action1, _ = self.policy1.get_action(self.obs)
-        action2, _ = self.policy2.get_action(self.obs)
-        action3 = self.policy3.get_action(self.obs)
-        action4 = self.policy4.get_action(self.obs)
+class SimpleGaussianMLPModel(Model):
```
ryanjulian: please put these fixtures into a test library

tests/garage/tf/policies/test_categorical_mlp_policy_with_model.py
```diff
@@ -0,0 +1,77 @@
...
+import tensorflow as tf
+
+from garage.tf.envs import TfEnv
+from garage.tf.models import Model
+from garage.tf.policies import CategoricalMLPPolicyWithModel
+from tests.fixtures import TfGraphTestCase
+from tests.fixtures.envs.dummy import DummyDiscreteEnv
+
+
+class SimpleMLPModel(Model):
```
ryanjulian: please put these fixtures into a test library

tests/garage/tf/models/test_mlp_model.py
```diff
@@ -0,0 +1,25 @@
...
+
+    def test_output_values(self):
+        model = MLPModel(
+            output_dim=self.output_dim,
+            hidden_sizes=(2, ),
```
ryanjulian: what about hidden_size = 1?

in most software, bugs center around the values 0, 1, 2, and 3+. it's good to test all of those cases (or whichever ones are valid for your test)

tests/garage/tf/policies/test_categorical_mlp_policy_with_model.py
```diff
@@ -0,0 +1,77 @@
...
+
+    @mock.patch('numpy.random.rand')
+    def test_get_action(self, mock_rand):
+        mock_rand.return_value = 1
+        action, _ = self.policy.get_action(self.obs)
+        assert self.env.action_space.contains(action)
+        assert action == 0
+
+        actions, _ = self.policy.get_actions([self.obs])
+        for action in actions:
```
ryanjulian: does this loop run for more than one iteration?

tests/garage/tf/policies/test_deterministic_mlp_policy_with_model.py
```diff
@@ -0,0 +1,66 @@
...
+        self.obs, _, _, _ = self.env.step(1)
+        model_path = 'garage.tf.policies.' + \
+                     'deterministic_mlp_policy_with_model.MLPModel'
+        with mock.patch(model_path, new=SimpleMLPModel):
+            self.policy = DeterministicMLPPolicyWithModel(
+                env_spec=self.env.spec)
+
+    def test_get_action(self):
+        action, _ = self.policy.get_action(self.obs)
+        assert self.env.action_space.contains(np.asarray([action]))
```
ryanjulian: why is np.asarray required here? why is get_action returning something that needs to be converted before being used with a space?

tests/garage/tf/policies/test_gaussian_mlp_policy_with_model.py
```diff
@@ -1,228 +1,78 @@
...
 
-        assert (self.box_env.action_space.low <= action1 <=  # noqa: W504
-                self.box_env.action_space.high)
-        assert (self.box_env.action_space.low <= action2 <=  # noqa: W504
-                self.box_env.action_space.high)
-        assert (self.box_env.action_space.low <= action3 <=  # noqa: W504
-                self.box_env.action_space.high)
-        assert (self.box_env.action_space.low <= action4 <=  # noqa: W504
-                self.box_env.action_space.high)
+    def __init__(self, name, output_dim, *args, **kwargs):
```
ryanjulian: why *args and **kwargs?

tests/garage/tf/policies/test_gaussian_mlp_policy_with_model.py
```diff
@@ -1,228 +1,78 @@
...
-        # * New way *
-        # equvaient to distribution.kl_sym()
-        kl_diff_sym2 = self.policy3.model.networks[
-            'default'].dist.kl_divergence(
-                self.policy4.model.networks['default'].dist)
-        objective2 = tf.reduce_mean(kl_diff_sym2)
+class TestGaussianMLPPolicyWithModel(TfGraphTestCase):
+    def setUp(self):
+        super().setUp()
+        self.env = TfEnv(DummyBoxEnv())
```
ryanjulian: please test several output_dim

ryanjulian:
These tests are looking great

Please see my comments about
* Moving your test fixtures into libraries in tests/
* Testing a variety of relevant values, especially for output_dim and hidden_sizes

Other things I noticed:
* Test for DiscreteMLPQFunction is missing
* It seems like you missed some test cases on GaussianMLPModel (see yellow and red lines here https://codecov.io/gh/rlworkgroup/garage/compare/11a265792b57b22211b672541f4f2e62d70b317b...949503e07658aa32dfc64f7e1932d3e493ca93ce/diff)

garage/tf/core/mlp.py
ahtsan: we cannot pickle an instance, but we can pickle the type.

garage/tf/models/gaussian_mlp_model.py
ahtsan: I used lambda in the first place and Flake8 E731 suggested not to assign a lambda function. Should we disable E731 in flake8?

garage/tf/models/gaussian_mlp_model.py
ahtsan: Ok. I will add that back.

garage/tf/models/gaussian_mlp_model.py
ahtsan: So we will have three different GaussianMLPModels?

tests/garage/tf/policies/test_categorical_mlp_policy_with_model.py
ahtsan: when I split with () and put that in `mock.patch()`, it will treat that as a tuple.

tests/garage/tf/policies/test_categorical_mlp_policy_with_model.py
ahtsan: because the policy will pass other arguments like `hidden_sizes` and `hidden_nonlinearity` to the model, but in this `SimpleMLPModel`, I don't want to deal with these.

tests/garage/tf/policies/test_categorical_mlp_policy_with_model.py
ahtsan: Yes if more than one observation is passed to `policy.get_actions`. I will add test to cover those cases.

garage/tf/core/mlp.py
ryanjulian: what if i want to use `tf.constant_initializer(5)`?

garage/tf/models/gaussian_mlp_model.py
ryanjulian: at least, yes. TBD if we should also split based on the stddev parameterization

garage/tf/models/gaussian_mlp_model.py
ryanjulian: i don't think it should complain if you assign the lambda as a kwarg

tests/garage/tf/policies/test_deterministic_mlp_policy_with_model.py
ahtsan: I think it's actually an issue.
e.g. In akro.box, the `contains` function is as following:
```python
    def contains(self, x):
        """Return boolean specifying if x is a valid member of this space."""
        return x.shape == self.shape and (x >= self.low).all() and (
            x <= self.high).all()
```
It assumes the input is a numpy array. 
Because the return action from `policy.get_action` can be a scalar value, it won't work.
Should we force the output of `policy.get_action` to be a numpy array?

tests/garage/tf/policies/test_categorical_mlp_policy_with_model.py
ryanjulian: you don't use commas
```python
(
    'foo'
    'bar'
) == 'foobar'

(
    'foo',
    'bar',
) == ('foo', 'bar')
```

garage/tf/core/mlp.py
ahtsan: we pass a lambda

ryanjulian: is that pickleable?

ahtsan: emm you are right. It's not.

ahtsan: Can we use cloudpickle? It can pickle lambda.

ryanjulian: no, i think we should stick to the standard pickling protocol

garage/sampler/parallel_sampler.py
```diff
@@ -74,6 +73,8 @@ def close():
 
 def _worker_set_seed(_, seed):
     logger.log("Setting seed to %d" % seed)
+    # import here to avoid circular dependency
```
ryanjulian: uhh can we restructure this to avoid a circular dependency instead?

garage/tf/models/gaussian_mlp_model.py
ryanjulian: https://github.com/rlworkgroup/garage/issues/587

tests/garage/tf/policies/test_deterministic_mlp_policy_with_model.py
ryanjulian: it looks like you found a `n = 1` vs `n > 1` bug. good job!

i think that the output of get_action should always pass the test
```python
a = foo.get_action(bar)
assert a in foo.action_space

bs = foo.get_actions(bars)
for b in bs:
    assert b in foo.action_space
```

the question is -- is this a garage bug or an akro bug?

i think it is a garage bug. the output type of a policy should be fixed -- it should't be "a scalar sometimes and an np.ndarray other times." if the value is single-dimensional, i think it should output an ndarray with shape `(1,)`.

of course, i could be persuaded otherwise if this made code elsewhere really nasty.

tests/garage/tf/models/test_gaussian_mlp_model.py
ahtsan: this parameterized?
https://pypi.org/project/parameterized/

CatherineSue: If you don't want to add it to the dep, you can use [parameterized tests in nose2](https://nose2.readthedocs.io/en/latest/params.html).

garage/sampler/parallel_sampler.py
ryanjulian: i think @zhanpenghe fixed this?

garage/tf/models/gaussian_mlp_model.py
```diff
@@ -130,52 +155,67 @@ def _build(self, state_input):
...
+                        output_w_init=self._std_output_w_init,
                         output_b_init=b,
-                        name='std_network')
+                        name='log_std_network',
+                        layer_normalization=self._layer_normalization)
                 else:
-                    p = tf.constant_initializer(self._init_std_param)
-                    std_network = parameter(
+
+                    def p():
```
ryanjulian: just use a lambda

setup.cfg
```diff
@@ -72,7 +72,8 @@ disable =
     useless-suppression,
     wildcard-import,
     not-context-manager,
-    c-extension-no-member
+    c-extension-no-member,
+    wrong-import-order
```
ryanjulian: i would prefer if we just configure pylint properly than disable this. which package did it complain about?

tests/fixtures/envs/dummy/dummy_discrete_pixel_env.py
```diff
@@ -16,8 +16,8 @@ class DummyDiscretePixelEnv(DummyEnv):
     Several properties are made for testing purpose as following:
 
     -Observations are
-        after reset    : np.zeros(self._shape).
-        action 1 (FIRE): np.ones(self._shape).
+        after reset    : np.ones(self._shape).
```
ryanjulian: did this sneak in from a different PR?

ahtsan: I changed it because we don't want to have zeros output

garage/tf/policies/deterministic_mlp_policy_with_model.py
```diff
@@ -0,0 +1,108 @@
...
+    """
+    DeterministicMLPPolicy with model.
+
+    The policy network selects action based on the state of the environment.
+    It uses neural nets to fit the function of pi(s).
+
+    Args:
+        env_spec: Environment specification.
+        name: variable scope of the mlp.
```
CatherineSue: Variable scope

garage/tf/policies/base2.py
```diff
@@ -88,6 +95,14 @@ def terminate(self):
         """Clean up operation."""
         pass
```
CatherineSue: What's this function for? Why pass?

ryanjulian:
Feel free to merge this and we will deal with pylint later

garage/tf/policies/base2.py
ahtsan: I copy from the existing policy base class from `garage.tf.policies.base`. If it is not used anywhere, we can remove that.

PR #575, opened 2019-03-07
Use env_spec in algos

### LINKED ISSUE #513: Use specs for constructing algorithms and primitives, not environments

naeioi:
This PR replaces env with env_spec in algorithm constructor.

This closes #513 .

### DIFF
--- a/examples/tf/her_ddpg_fetchreach.py
+++ b/examples/tf/her_ddpg_fetchreach.py
@@ -51,14 +51,14 @@ def run_task(*_):
             reward_fun=env.compute_reward)
 
         ddpg = DDPG(
-            env,
+            env_spec=env.spec,
             policy=policy,
             policy_lr=1e-3,
             qf_lr=1e-3,
             qf=qf,
             replay_buffer=replay_buffer,
-            plot=False,
             target_update_tau=0.05,
+            n_epoch_cycles=20,
             max_path_length=100,
             n_train_steps=40,
             discount=0.9,
--- a/garage/experiment/local_tf_runner.py
+++ b/garage/experiment/local_tf_runner.py
@@ -128,7 +128,7 @@ class LocalRunner:
 
     def initialize_tf_vars(self):
         """Initialize all uninitialized variables in session."""
-        with tf.name_scope("auto-initialized"):
+        with tf.name_scope("initialize_tf_vars"):
             self.sess.run(
                 tf.variables_initializer([
                     v for v in tf.global_variables()
@@ -174,8 +174,11 @@ class LocalRunner:
             paths: Batch of samples after preprocessed.
 
         """
+        assert self.has_setup
+
         logger.log("Saving snapshot...")
         params = self.algo.get_itr_snapshot(itr)
+        params['env'] = self.env
         if paths:
             params["paths"] = paths
         logger.save_itr_params(itr, params)
--- a/garage/experiment/local_tf_runner.py
+++ b/garage/experiment/local_tf_runner.py
@@ -121,19 +121,20 @@ class LocalRunner:
                 from garage.tf.samplers import OffPolicyVectorizedSampler
                 sampler_cls = OffPolicyVectorizedSampler
 
-        self.sampler = sampler_cls(algo, **sampler_args)
+        self.sampler = sampler_cls(algo, env, **sampler_args)
 
         self.initialize_tf_vars()
         self.has_setup = True
 
     def initialize_tf_vars(self):
         """Initialize all uninitialized variables in session."""
-        self.sess.run(
-            tf.variables_initializer([
-                v for v in tf.global_variables()
-                if v.name.split(':')[0] in str(
-                    self.sess.run(tf.report_uninitialized_variables()))
-            ]))
+        with tf.name_scope("auto-initialized"):
+            self.sess.run(
+                tf.variables_initializer([
+                    v for v in tf.global_variables()
+                    if v.name.split(':')[0] in str(
+                        self.sess.run(tf.report_uninitialized_variables()))
+                ]))
 
     def start_worker(self):
         """Start Plotter and Sampler workers."""
@@ -174,7 +175,7 @@ class LocalRunner:
 
         """
         logger.log("Saving snapshot...")
-        params = self.algo.get_itr_snapshot(itr, paths)
+        params = self.algo.get_itr_snapshot(itr)
         if paths:
             params["paths"] = paths
         logger.save_itr_params(itr, params)
--- a/garage/tf/algos/ddpg.py
+++ b/garage/tf/algos/ddpg.py
@@ -47,7 +47,7 @@ class DDPG(OffPolicyRLAlgorithm):
         Construct class.
 
         Args:
-            env_spec(): Environment specification.
+            env_spec(akro.Space): Environment specification.
             target_update_tau(float): Interpolation parameter for doing the
         soft target update.
             discount(float): Discount factor for the cumulative return.
--- a/garage/tf/algos/ddpg.py
+++ b/garage/tf/algos/ddpg.py
@@ -47,7 +47,7 @@ class DDPG(OffPolicyRLAlgorithm):
         Construct class.
 
         Args:
-            env_spec(akro.Space): Environment specification.
+            env_spec(EnvSpec): Environment specification.
             target_update_tau(float): Interpolation parameter for doing the
         soft target update.
             discount(float): Discount factor for the cumulative return.
--- a/garage/tf/algos/ddpg.py
+++ b/garage/tf/algos/ddpg.py
@@ -28,7 +28,7 @@ class DDPG(OffPolicyRLAlgorithm):
     """
 
     def __init__(self,
-                 env,
+                 env_spec,
                  replay_buffer,
                  target_update_tau=0.01,
                  policy_lr=1e-4,
@@ -47,7 +47,7 @@ class DDPG(OffPolicyRLAlgorithm):
         Construct class.
 
         Args:
-            env(): Environment.
+            env_spec(): Environment specification.
             target_update_tau(float): Interpolation parameter for doing the
         soft target update.
             discount(float): Discount factor for the cumulative return.
@@ -65,7 +65,7 @@ class DDPG(OffPolicyRLAlgorithm):
             max_action(float): Maximum action magnitude.
             name(str): Name of the algorithm shown in computation graph.
         """
-        action_bound = env.action_space.high
+        action_bound = env_spec.action_space.high
         self.max_action = action_bound if max_action is None else max_action
         self.tau = target_update_tau
         self.policy_lr = policy_lr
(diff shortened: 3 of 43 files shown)


garage/experiment/local_tf_runner.py
```diff
@@ -120,19 +120,20 @@ def setup(self, algo, env, sampler_cls=None, sampler_args=None):
...
 
     def initialize_tf_vars(self):
         """Initialize all uninitialized variables in session."""
-        self.sess.run(
-            tf.variables_initializer([
-                v for v in tf.global_variables()
-                if v.name.split(':')[0] in str(
-                    self.sess.run(tf.report_uninitialized_variables()))
-            ]))
+        with tf.name_scope("auto-initialized"):
```
ryanjulian: hmm what prompted this change?

garage/tf/algos/ddpg.py
```diff
@@ -47,7 +47,7 @@ def __init__(self,
         Construct class.
 
         Args:
-            env(): Environment.
+            env_spec(): Environment specification.
```
ryanjulian: does't this have a type?

garage/tf/samplers/on_policy_vectorized_sampler.py
```diff
@@ -13,25 +13,25 @@
...
+        super(OnPolicyVectorizedSampler, self).__init__(algo, env, n_envs)
         self.n_envs = n_envs
 
     @overrides
     def start_worker(self):
         n_envs = self.n_envs
 
-        if getattr(self.algo.env, 'vectorized', False):
-            self.vec_env = self.algo.env.vec_env_executor(
+        if getattr(self.env, 'vectorized', False):
```
ryanjulian: im not sure this case even exists anymore. you can probably delete it.

garage/experiment/local_tf_runner.py
CatherineSue: The current tf graph looks like this: (many report_uninitialized_variables)
image

naeioi:
@ryanjulian Because the snapshot is taken by the algorithm and I removed env from the pickled file, it turns out that there's no way to save env.  My idea is that the saving env is not very important so I just removed it. 

Also, I doubt the experiement resuming really worked previously because `algo.train()` will randomized all variables. It's neither working currently because `run_experiment()` still calls `algo.train()` which is not valid anymore. https://github.com/rlworkgroup/garage/blob/1c299fd56b90f641321cfc0482eca6770256f49c/scripts/run_experiment.py#L174-L178

I plan to resolve this issue together with #516 .

garage/tf/algos/ddpg.py
naeioi: It is `akro.Space`?

garage/experiment/local_tf_runner.py
ryanjulian: ah i see. can you change the scope name to `'initialize_tf_vars'`?

That will make it easy for someone find what code produced this when they see it in TensorBoard.

ryanjulian: also, can  you show the "after" TensorBoard corresponding to Chang's "before"?

garage/experiment/local_tf_runner.py
```diff
@@ -172,7 +173,7 @@ def save_snapshot(self, itr, paths=None):
 
         """
         logger.log("Saving snapshot...")
-        params = self.algo.get_itr_snapshot(itr, paths)
+        params = self.algo.get_itr_snapshot(itr)
         if paths:
             params["paths"] = paths
```
ryanjulian: untested

garage/tf/algos/ddpg.py
ryanjulian: uhh i think it's https://github.com/rlworkgroup/garage/blob/master/garage/envs/env_spec.py for now.

ryanjulian:
@naeioi  it's quite important to serialize the environment, and we can't really merge this without saving the environment.

Look at https://github.com/rlworkgroup/garage/blob/master/examples/sim_policy.py -- this PR breaks it. 

TODO: We should add a test to make sure the snapshotter produces complete .pkl files, so that this PR would fail in the CI immediately.

Reasons:
1. There's no reason `env` can't have meaningful parameters which are saved/restored during pickle/unpickle -- for instance, what if I had a PointMass environment with specifiable mass? Not all envs are just strings from openai/gym. The only interface we impose on `env` that (1) `env == pickle.loads(pickle.dumps(env))` and (2) `isinstance(env, gym.Env)`
2. A `.pkl` file should be a total record of an experiment -- it should have everything I need to evaluate an agent in the environment, continue training the agent in the environment, etc. removing `Env` makes that untrue -- i would have to look at the launcher and be careful to recreate the environment. consider how hard this would be if I decided to procedurally generate my environments!
3. Perhaps (2) isn't the best way to save experiment state, but right now it's what we have.

naeioi:
@ryanjulian That makes sense. I added a test to verify the presence and integrity of both environment and policy in snapshot. As a workaround of algo not having access to the actual environment, the environment is now saved in `save_snapshot()` in the runner.

garage/experiment/local_tf_runner.py
ryanjulian: looks great.

ryanjulian:
awesome!

tests/garage/experiment/test_snapshot.py
```diff
@@ -0,0 +1,70 @@
...
+            env.close()
+
+        tf.reset_default_graph()
+
+        # Read snapshot from self.log_dir
+        # Test the presence and integrity of policy and env
+        with LocalRunner():
+            snapshot = joblib.load(osp.join(self.log_dir, 'params.pkl'))
+            env = snapshot['env']
+            policy = snapshot['policy']
```
ryanjulian: perhaps

```python
assert env
assert policy
```
Just so that any error leads me to this line, instead of some line inside `rollout()`

naeioi: Good point! Fixed.

garage/experiment/local_tf_runner.py
```diff
@@ -171,8 +172,11 @@ def save_snapshot(self, itr, paths=None):
             paths: Batch of samples after preprocessed.
 
         """
+        assert self.has_setup
+
         logger.log("Saving snapshot...")
-        params = self.algo.get_itr_snapshot(itr, paths)
+        params = self.algo.get_itr_snapshot(itr)
+        params['env'] = self.env
```
ryanjulian: hmm can you do this in `save_snapshot` instead? that covers the case where an environment has evolving state.

it would be good to test for that case too, since it would be incredibly hard to detect/debug...

garage/tf/algos/ddpg.py
```diff
@@ -197,7 +197,7 @@ def train_once(self, itr, paths):
         if itr % self.n_epoch_cycles == 0:
             logger.log("Training finished")
             logger.log("Saving snapshot #{}".format(int(epoch)))
-            params = self.get_itr_snapshot(epoch, paths)
+            params = self.get_itr_snapshot(epoch)
```
ryanjulian: is `params` even used?

garage/experiment/local_tf_runner.py
naeioi: Are you saying that I should get the snapshot of env from sampler because the copy of env in the runner is not the actual copy used for sampling? 

This leads to another problem: What to save when there are more than one env in the sampler. We can save all envs, but we dont't actually have a mechanism to restore them in sampler. Or we pick one env and save it, and that will be compatible to the current things we save in snapshot.

garage/tf/algos/ddpg.py
naeioi: No. And the `save_itr_params` is not necessary. I will delete it.

garage/experiment/local_tf_runner.py
ryanjulian: hmm, this presents an interesting problem gap in the interface, but i'm not sure it's one we need to solve right now.

for now i think it's sufficient to make sure there's a copy of `env` in every .pkl file. we can try to handle statefulness later if there's a need.

naeioi: OK. Let's keep an eye on it.

tests/garage/experiment/test_snapshot.py
```diff
@@ -0,0 +1,75 @@
...
+
+            env.close()
+
+        self.tearDown()
+        self.setUp()
+
+        # Read snapshot from self.log_dir
+        # Test the presence and integrity of policy and env
+        with LocalRunner():
+            snapshot = joblib.load(osp.join(self.log_dir, 'params.pkl'))
```
ryanjulian: please add a test case which makes sure that the environment is present in the pkl file generated by every iteration (e.g. 2 or 3)

tests/garage/experiment/test_snapshot.py
```diff
@@ -0,0 +1,75 @@
...
+                max_path_length=100,
+                discount=0.99,
+                max_kl_step=0.01)
+
+            runner.setup(algo, env)
+            runner.train(n_epochs=1, batch_size=4000)
+
+            env.close()
+
+        self.tearDown()
```
ryanjulian: hmm this is pretty confusing to see in a test, and makes you vulnerable to arbitary changes in TfGraphTestCase. I think you should just define local helpers for this.

tests/garage/experiment/test_snapshot.py
naeioi: Added.

tests/garage/experiment/test_snapshot.py
naeioi: Addressed.

examples/tf/her_ddpg_fetchreach.py
```diff
@@ -51,14 +51,14 @@ def run_task(*_):
...
+            env_spec=env.spec,
             policy=policy,
             policy_lr=1e-3,
             qf_lr=1e-3,
             qf=qf,
             replay_buffer=replay_buffer,
-            plot=False,
             target_update_tau=0.05,
+            n_epoch_cycles=20,
             max_path_length=100,
```
CatherineSue: I thought this line is deleted in off policy algorithm?

ryanjulian:
bump. are there any hard blockers on this PR?

examples/tf/her_ddpg_fetchreach.py
naeioi: Actually not. I keep `n_epoch_cycles` so that an algorithm can compute epoch by `iteration/n_epoch_cycles`. This is a little bit awkward but required in DDPG.

naeioi:
I will have this merged after rebase.

examples/tf/her_ddpg_fetchreach.py
CatherineSue: I don't quite understand. I meant is `max_path_length=100` not necessary here?

naeioi:
I disable two pylint checks. 
- `not-context-manager check`
Pylint has long been having bug on checking context manager.
See https://github.com/PyCQA/pylint/issues/782.

- `c-extension-no-member check`
Some C modules don't report their functions and variable in python. In our case, baselines is using mpi4py which makes pylint complain.
See https://travis-ci.com/rlworkgroup/garage/builds/105359651#L690.

Issue #568, opened 2019-03-06
Change VecEnvExecutor to not set done to True in off policy algos


CatherineSue:
`garage.tf.envs.VecEnvExecutor` sets `done = True` when it reaches the `max_path_length`. This behavior blocks the off-policy algorithms to reach the highest performance.

Issue #442, opened 2018-12-19
Turn off special entropy features by default for NPO


zhanpenghe:
We introduced softplus entropy, negative log likelihood (btw, this feature offends an important property of entropy) as entropy into NPO. I think sometimes these features mess up the whole training at least through my observation on training sawyer environments (discovered when training sawyer reacher env with PPO). I suggest that we should turn them off by default.

gautams3:
This was discovered when we tried to train the sawyer reacher env with vanilla PPO

CatherineSue:
If we use `stop_gradient(entropy)`, it won't affect the training in theory, right?

zhanpenghe:
it will.. in theory

CatherineSue:
Could you be more specific here? After you turn off the entropy features, does the training work?

zhanpenghe:
Training did not work without turning them off.

ryanjulian:
Negative log likelihood is by far the most common entropy estimator from the literature, so I'm not sure how it offends anything?

softplus entropy is a hack/tuning knob which I agree should be off by default.

ryanjulian:
The reason we have things like softplus is that PPO is much less stable with entropy than TRPO.

CatherineSue:
This issue is addressed.

PR #579, opened 2019-03-08
Add miscellaneous changes for benchmarking

CatherineSue:
* Algorithm
   * Fix entropy in tf/algos/npo.py. According to the PPO paper, the term
     `policy_ent_coeff * pol_ent` should be added to the policy gradient
     loss instead of the reward. By fixing this, the policy stddev will
     decrease if policy_ent_coeff is very small.
   * Also addressed issue #442.

 * Sampler
   * baselines logs average return of the recent 100 trajectories. Add
     info to garage.tf.BatchSampler. Note this is a temp fix to use
     benchmark.

 * Benchmark
    * Update the current benchmark to work properly. This includes:
      * Add AutoStopEnv wrapper. baselines doesn't set done=True in its
        VecNormEnv. To make the sampler behavior consistent with garage,
        the AutoStopEnv wrapper will set done to True after
        `max_path_length` steps.
      * Replace `trial` typo in test_benchmark_*.py.
      * Move tf.Graph() to parent function as baselines can't run
        multiple times in the same tf graph.
    * Add tensorboard log support to baselines. 
    * Currently, the test_benchmark_ppo and test_benchmark_trpo works.
      test_benchmark_ddpg and test_benchmark_her may not work. For further
      info please refer to issue #568. 

  * Regressor
    * Rename `step_size` to `max_kl_step` in tf/regressors to consistent
      with the naming convention garage uses.
    * Add name to baselines so that NPO can use its name in
     `_fit_baseline`.

Note this PR made some band-aid fix in order to make the benchmark work.
They should be addressed when the sampler and new logger is merged.

### DIFF
--- a/garage/tf/samplers/batch_sampler.py
+++ b/garage/tf/samplers/batch_sampler.py
@@ -1,3 +1,5 @@
+from collections import deque
+
 import numpy as np
 import tensorflow as tf
 
@@ -22,6 +24,7 @@ class BatchSampler(BaseSampler):
     def __init__(self, algo, n_envs):
         super(BatchSampler, self).__init__(algo)
         self.n_envs = n_envs
+        self.eprewmean = deque(maxlen=100)
 
     def start_worker(self):
         assert singleton_pool.initialized, \
@@ -120,6 +123,7 @@ class BatchSampler(BaseSampler):
             [path["returns"][0] for path in paths]))
 
         undiscounted_returns = [sum(path["rewards"]) for path in paths]
+        self.eprewmean.extend(undiscounted_returns)
 
         ent = np.sum(
             self.algo.policy.distribution.entropy(agent_infos) *
@@ -143,6 +147,8 @@ class BatchSampler(BaseSampler):
         logger.record_tabular('AverageDiscountedReturn',
                               average_discounted_return)
         logger.record_tabular('AverageReturn', np.mean(undiscounted_returns))
+        logger.record_tabular('Extras/EpisodeRewardMean',
+                              np.mean(self.eprewmean))
         logger.record_tabular('NumTrajs', len(paths))
         logger.record_tabular('Entropy', ent)
         logger.record_tabular('Perplexity', np.exp(ent))
--- a/tests/benchmarks/test_benchmark_ddpg.py
+++ b/tests/benchmarks/test_benchmark_ddpg.py
@@ -2,8 +2,8 @@
 This script creates a regression test over garage-DDPG and baselines-DDPG.
 
 It get Mujoco1M benchmarks from baselines benchmark, and test each task in
-its trail times on garage model and baselines model. For each task, there will
-be `trail` times with different random seeds. For each trail, there will be two
+its trial times on garage model and baselines model. For each task, there will
+be `trial` times with different random seeds. For each trial, there will be two
 log directories corresponding to baselines and garage. And there will be a plot
 plotting the average return curve from baselines and garage.
 """
@@ -36,6 +36,7 @@ from garage.tf.algos import DDPG
 from garage.tf.envs import TfEnv
 from garage.tf.policies import ContinuousMLPPolicy
 from garage.tf.q_functions import ContinuousMLPQFunction
+from tests.wrappers import AutoStopEnv
 
 # Hyperparams for baselines and garage
 params = {
@@ -66,11 +67,13 @@ class TestBenchmarkDDPG(unittest.TestCase):
         mujoco1m = benchmarks.get_benchmark("Mujoco1M")
 
         timestamp = datetime.datetime.now().strftime("%Y-%m-%d-%H-%M-%S-%f")
-        benchmark_dir = "./benchmark_ddpg/%s/" % timestamp
+        benchmark_dir = "./data/local/benchmarks/ddpg/%s/" % timestamp
 
         for task in mujoco1m["tasks"]:
             env_id = task["env_id"]
             env = gym.make(env_id)
+            baseline_env = AutoStopEnv(
+                env_name=env_id, max_path_length=params["n_rollout_steps"])
             seeds = random.sample(range(100), task["trials"])
 
             task_dir = osp.join(benchmark_dir, env_id)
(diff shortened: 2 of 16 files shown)


CatherineSue:
#PPO Benchmark Result - 1
## Hyperparameters
|Hyperparameter        |Value  |         
|:------------- |:-------------|
| Dicount      | 0.99 |
| GAE parameter | 0.95 |
| Minibatch size | 32 |
| Num. epochs | 10 |
| Horizon | 2048 |
| Policy hidden sizes | (64, 64) |
| Baseline(value function) hidden sizes | (64, 64) |
| Baseline optimizer | PenaltyLbfgsOptimizer |
| Adam optimizer learning rate | 1e-3 |
| Clip range | 0.2 | 

Note that baselines seems to be unstable when the learning rate is 1e-3. This may be related to their policy-vf shared network architecture. Their default lr is 3e-4. I found that `learning_rate=3e-4` causes a slow convergence in garage. Since `learning_rate=1e-3` is the default setting of tf.train.Adamoptimizer, I stick with 1e-3.

## InvertedDoublePendulum-v2

## InvertedPendulum-v2
InvertedPendulum-v2_benchmark

## HalfCheetah-v2
HalfCheetah-v2_benchmark

## Hopper-v2
Hopper-v2_benchmark

## Walker2d-v2
Walker2d-v2_benchmark

Reacher and Swimmer is still running. I'll update later.

CatherineSue:
# PPO Benchmark Result - 2
## Hyperparameters
|Hyperparameter        |Value  |         
|:------------- |:-------------|
| Dicount      | 0.99 |
| GAE parameter | 0.95 |
| Minibatch size | 32 |
| Num. epochs | 10 |
| Horizon | 2048 |
| Policy hidden sizes | (64, 64) |
| Baseline(value function) hidden sizes | (64, 64) |
| Baseline optimizer | **FirstOrderOptimizer** |
| Adam optimizer learning rate | 1e-3 |
| Clip range | 0.2 | 

Note that baselines seems to be unstable when the learning rate is 1e-3. This may be related to their policy-vf shared network architecture. Their default lr is 3e-4. I found that `learning_rate=3e-4` causes a slow convergence in garage. Since `learning_rate=1e-3` is the default setting of tf.train.Adamoptimizer, I stick with 1e-3.

## InvertedDoublePendulum-v2

## InvertedPendulum-v2
InvertedPendulum-v2_benchmark

## HalfCheetah-v2
HalfCheetah-v2_benchmark

## Hopper-v2
Hopper-v2_benchmark

## Walker2d-v2
Walker2d-v2_benchmark

Reacher and Swimmer is still running. I'll update later.

CatherineSue:
## Entropy Test
Screen Shot 2019-03-08 at 1 06 21 PM

ryanjulian:
These look great -- regarding entropy: there are several ways to implement this. The one in the PPO paper is just one. Sometimes it is an augmented loss and sometimes it is augmented reward (maximum entropy vs entropy-regularized). 

I would like to see NPO support both implementations, but it's better right now to have at least one which is known to work.

garage/tf/samplers/batch_sampler.py
```diff
@@ -143,6 +147,7 @@ def process_samples(self, itr, paths):
         logger.record_tabular('AverageDiscountedReturn',
                               average_discounted_return)
         logger.record_tabular('AverageReturn', np.mean(undiscounted_returns))
+        logger.record_tabular('eprewmean', np.mean(self.eprewmean))
```
ryanjulian: if we are going to log this, we should use our own naming convention (e.g. `EpisodeRewardMean`). I think there are already mappings between our names and baseline names in the benchmark scripts.

tests/benchmarks/test_benchmark_ddpg.py
```diff
@@ -2,8 +2,8 @@
 This script creates a regression test over garage-DDPG and baselines-DDPG.
```
ryanjulian: just a thought: perhaps we should add a test case to each benchmark which runs each benchmark for just 1 iteration, to make sure that all the benchmarks stay running using the CI.

we would still exclude the main benchmark from CI

tests/helpers.py
```diff
@@ -1,5 +1,6 @@
 import pickle
```
ryanjulian: tests/wrappers.py?

there will probably be more of these.

garage/tf/samplers/batch_sampler.py
CatherineSue: Yes there are mappings for names. I used this naming to view the regression in tensorboard. I agree we should follow our naming convention. It would be better if we can move this extra info to somewhere else in the future. Because it only serves for benchmark.

tests/benchmarks/test_benchmark_ddpg.py
CatherineSue: Good idea. I believe Kevin has done a similar thing in his branch.

tests/helpers.py
CatherineSue: Good idea. I'll move it.

tests/benchmarks/test_benchmark_ddpg.py
CatherineSue: @cheng-kevin I assume the files under [this folder](https://github.com/rlworkgroup/garage/tree/Automated_Benchmarking_normalized/tests/benchmarks/continuous_integration) are for ci? Could you port them to this branch?

CatherineSue:
I've updated the PPO benchmark results and updated notes. The DDPG benchmark is running.
I also addressed all the comments.

tests/benchmarks/test_benchmark_her.py
```diff
@@ -64,18 +64,18 @@ def test_benchmark_her(self):
...
+                baselines_dir = trial_dir + "/baselines"
 
-                garage_csv = run_garage(env, seed, garage_dir)
+                with tf.Graph().as_default():
+                    garage_csv = run_garage(env, seed, garage_dir)
 
-                CACHED_ENVS.clear()
-                baselines_csv = run_baselines(env_id, seed, baselines_dir)
+                    CACHED_ENVS.clear()
+                    baselines_csv = run_baselines(env_id, seed, baselines_dir)
```
ahtsan: Do we need the AutoStop wrapper here?

tests/wrappers.py
```diff
@@ -0,0 +1,24 @@
+import gym
+
+
+class AutoStopEnv(gym.Wrapper):
+    """A env wrapper that stops rollout at step max_path_length."""
+
+    def __init__(self, env=None, env_name="", max_path_length=100):
+        if env_name:
+            super().__init__(gym.make(env_name))
```
ahtsan: Why do we need the env_name? Why can't we just pass the environment?

garage/tf/samplers/batch_sampler.py
ryanjulian: Perhaps just put them under "Extras/" or "Info/"?

tests/benchmarks/test_benchmark_her.py
CatherineSue: No.

tests/wrappers.py
CatherineSue: You can do both.

CatherineSue: I was referring to [this](https://github.com/rlworkgroup/garage/blob/master/garage/envs/base.py#L44).

ryanjulian: I think Anson was asking why you didn't use the standard `gym.Wrapper` interface [here](https://github.com/openai/gym/blob/master/gym/core.py#L209)

I also think that `gym.Wrapper` is preferable, but it's test code so I won't insist on any changes.

CatherineSue: Thanks for pointing it out. Yes I agree that one is better. I didn't pay much attention to this class because it was implemented before and I just pasted it here. My alter goal is that we can remove this when the new sampler is finished. Then this wrapper won't be required.

ryanjulian: then it doesn't seem we need to talk much about it, anyway :)

PR #582, opened 2019-03-12
Add tf/DQN with dueling support

ahtsan:
- DQN implementation with garage.Model.
- Current for pixel environments we use a bunch of wrappers from baselines. Later we can setup data processing pipeline to make that faster.
- Removed the single-layer mlp in cnn. I think it makes more sense to separate them, so now cnn will return the flattened output. Also modified the unit test for cnn accordingly.
- Used `self.models` to store all the models in QFunction. Sometimes there are multiple models and we need something to keep track of them. I assume the models are stored in order, so input must be `self.models[0].input` and output must be `self.models[-1].output`. 
- Added corresponding functions and property in QFunction to make reference easier, e.g. `def q_vals()` and `input` property.
- Clone method in QFunction to enable object copying (I think there should be better ways). Since all the necessary operations will be done in object construction, we can simply create a new object with a different name (because we want the new object will have a different variable_scope).

Will post benchmark result soon, fixing the scaling.
Will add more tests later.


garage/tf/q_functions/discrete_mlp_q_function.py
```diff
@@ -41,25 +43,62 @@ def __init__(self,
...
+                output_nonlinearity=output_nonlinearity,
+                output_w_init=output_w_init,
+                output_b_init=output_b_init,
+                layer_normalization=layer_normalization)
 
+        self.models.append(model)
         obs_ph = tf.placeholder(tf.float32, (None, ) + obs_dim, name="obs")
 
-        self.model.build(obs_ph)
+        with tf.variable_scope(self._variable_scope):
```
ryanjulian: `with self._variable_scope`

garage/tf/q_functions/discrete_mlp_q_function.py
```diff
@@ -41,25 +43,62 @@ def __init__(self,
...
+            out = obs_ph
+            for model in self.models:
+                out = model.build(out)
+
+    def q_vals(self):
+        return self.models[-1].networks['default'].outputs
```
ryanjulian: why not use a dict?

examples/tf/dqn_breakout.py
```diff
@@ -0,0 +1,92 @@
...
+            time_horizon=max_path_length)
+
+        qf = DiscreteCNNQFunction(
+            env_spec=env.spec,
+            filter_dims=(8, 4, 3),
+            num_filters=(32, 64, 64),
+            strides=(4, 2, 1),
+            dueling=False)
+
+        policy = DiscreteQfDerivedPolicy(env_spec=env, qf=qf)
```
CatherineSue: `env.spec`

garage/tf/algos/dqn.py
```diff
@@ -0,0 +1,282 @@
...
+
+    def __init__(self,
+                 env,
+                 replay_buffer,
+                 max_path_length=200,
+                 num_timesteps=int(1e5),
+                 qf_lr=0.001,
+                 qf_optimizer=tf.train.AdamOptimizer,
+                 discount=1.0,
+                 name=None,
```
CatherineSue: name='DQN'

garage/tf/algos/dqn.py
```diff
@@ -0,0 +1,282 @@
...
+        """
+        Initialize the networks and Ops.
+
+        Assume discrete space for dqn, so action dimension
+        will always be action_space.n
+        """
+        action_dim = self.env.action_space.n
+
+        # build q networks
+        with tf.name_scope(self.name, "DQN"):
```
CatherineSue: I suppose it would be better to use single quotation for new files.

garage/tf/algos/dqn.py
```diff
@@ -0,0 +1,282 @@
...
+                        if grad is not None:
+                            gradients[i] = (tf.clip_by_norm(
+                                grad, self.grad_norm_clipping), var)
+                    self._optimize_loss = optimizer.apply_gradients(gradients)
+                else:
+                    self._optimize_loss = optimizer.minimize(
+                        self._loss, var_list=self.qf.get_trainable_vars())
+
+    @overrides
+    def train(self, sess=None):
```
CatherineSue: @naeioi Could you help @ahtsan to refactor this?

garage/tf/q_functions/discrete_mlp_q_function.py
ahtsan: `self._variable_scope` is a `VariableScope` object, we have to do 
```
with tf.variable_scope(VariableScopeObject):
```
(see https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/variable_scope.py#L1804).
Also `VariableScope` is pickleable, `variable_scope` is not.

garage/tf/algos/dqn.py
```diff
@@ -0,0 +1,282 @@
...
+                    action, _ = self.es.get_action(itr, obs_scaled,
+                                                   self.policy)
+                else:
+                    action, _ = self.policy.get_action(obs_scaled)
+
+                next_obs, reward, done, env_info = self.env.step(action)
+
+                if self.use_atari_wrappers:
+                    next_obs = np.asarray(next_obs)
+                self.replay_buffer.add_transitions(
```
CatherineSue: I think it would be better to add a `add_transition` to replay buffer.

garage/tf/algos/dqn.py
```diff
@@ -0,0 +1,282 @@
...
+                        input("Plotting evaluation run: Press Enter to "
+                              "continue...")
+
+                if self.evaluate:
+                    mean100ep_rewards = round(
+                        np.mean(episode_rewards[-100:]), 1)
+                    mean100ep_qf_loss = np.mean(episode_qf_losses[-100:])
+                    if itr % self.print_freq == 0:
+                        logger.record_tabular('Iteration', itr)
+                        logger.record_tabular('AverageReturn',
```
CatherineSue: I think it would be better to rename this to Episode100RewardMean. Because in garage convention, AverageReturn means the mean of the current rollout's return.

garage/tf/algos/dqn.py
```diff
@@ -0,0 +1,282 @@
...
+                                              mean100ep_rewards)
+                        logger.record_tabular('StdReturn',
+                                              np.std(episode_rewards))
+                        logger.record_tabular('QFunction/AverageQFunctionLoss',
```
CatherineSue: Same here. Please rename it to `'{}/Episode100LossMean'.format(self.qf.name)`

garage/tf/algos/dqn.py
```diff
@@ -0,0 +1,282 @@
...
+        """Optimize network using experiences from replay buffer."""
+        transitions = self.replay_buffer.sample(self.buffer_batch_size)
+
+        observations = transitions["observation"]
+        rewards = transitions["reward"]
+        actions = transitions["action"]
+        next_observations = transitions["next_observation"]
+        dones = transitions["terminal"]
+
+        if len(self.env.spec.observation_space.shape) == 3:
```
CatherineSue: emm, I feel it is strange to have an if clause here. Can we either include it in a util function?

garage/tf/algos/dqn.py
```diff
@@ -0,0 +1,282 @@
...
+                self.action_t_ph: actions,
+                self.reward_t_ph: rewards,
+                self.done_t_ph: dones,
+                self.target_qf.input: next_observations
+            })
+
+        return loss
+
+
+def get_target_ops(variables, target_variables):
```
CatherineSue: If this is required for DQN, please move the same function in DDPG to a module. Perhaps garage/tf/misc?

garage/tf/models/cnn_model.py
```diff
@@ -0,0 +1,44 @@
...
+        hidden_nonlinearity: Activation function for intermediate dense
+            layer(s).
+    """
+
+    def __init__(self,
+                 filter_dims,
+                 num_filters,
+                 strides,
+                 name=None,
+                 padding="SAME",
```
CatherineSue: I think it's better to use single quotations for new files.

garage/tf/models/cnn_model.py
```diff
@@ -0,0 +1,44 @@
...
+        padding: The type of padding algorithm to use, from "SAME", "VALID".
+        hidden_nonlinearity: Activation function for intermediate dense
+            layer(s).
+    """
+
+    def __init__(self,
+                 filter_dims,
+                 num_filters,
+                 strides,
+                 name=None,
```
CatherineSue: name='CNNModel'

garage/tf/q_functions/discrete_cnn_q_function.py
```diff
@@ -0,0 +1,190 @@
...
+        self.models.append(output_model)
+
+        obs_ph = tf.placeholder(tf.float32, (None, ) + obs_dim, name="obs")
+
+        with tf.variable_scope(self._variable_scope):
+            out = obs_ph
+            for model in self.models:
+                out = model.build(out)
+
+    def q_vals(self):
```
CatherineSue: `@property`

garage/tf/q_functions/discrete_mlp_q_function.py
```diff
@@ -41,25 +43,62 @@ def __init__(self,
...
+        self.models.append(model)
         obs_ph = tf.placeholder(tf.float32, (None, ) + obs_dim, name="obs")
 
-        self.model.build(obs_ph)
+        with tf.variable_scope(self._variable_scope):
+            out = obs_ph
+            for model in self.models:
+                out = model.build(out)
+
+    def q_vals(self):
```
CatherineSue: `@property`

tests/garage/tf/algos/test_dqn.py
```diff
@@ -0,0 +1,55 @@
...
+from garage.replay_buffer import SimpleReplayBuffer
+from garage.tf.algos import DQN
+from garage.tf.envs import TfEnv
+from garage.tf.policies import DiscreteQfDerivedPolicy
+from garage.tf.q_functions import DiscreteMLPQFunction
+from tests.fixtures import TfGraphTestCase
+
+
+class TestDQN(TfGraphTestCase):
+    def test_dqn_cartpole(self):
```
CatherineSue: How long does it take to run this test?

garage/tf/algos/dqn.py
```diff
@@ -0,0 +1,282 @@
...
+        # observation. The way to unpack it is to unpack obs as numpy array.
+        # For reference, see
+        # https://github.com/openai/baselines/blob/1259f6ab25c6f7261e33c
+        # 4c3b92df869188f9260/baselines/common/atari_wrappers.py#L202
+        if self.use_atari_wrappers:
+            obs = np.asarray(obs)
+        episode_rewards.append(0.)
+
+        for itr in range(self.num_timesteps):
+            with logger.prefix('Iteration #%d | ' % itr):
```
CatherineSue: 'Timestep' sounds more appropriate to me.

garage/tf/algos/dqn.py
```diff
@@ -0,0 +1,282 @@
...
+                                                   self.policy)
+                else:
+                    action, _ = self.policy.get_action(obs_scaled)
+
+                next_obs, reward, done, env_info = self.env.step(action)
+
+                if self.use_atari_wrappers:
+                    next_obs = np.asarray(next_obs)
+                self.replay_buffer.add_transitions(
+                    observation=[obs],
```
CatherineSue: Could we add normalized obs and next_obs to replay buffer?

garage/tf/models/cnn_model.py
ahtsan: `Model` has a default name if name=None, which will be the class name. How should we do this? If we enforce all derived model class to have name, we don't need the default name at all.

garage/tf/algos/dqn.py
ahtsan: Normalized obs and next_obs will be float. To save memory, we better store uint8 and only normalize them when needed, so I think we could, but we shouldn't.

garage/tf/q_functions/discrete_cnn_q_function.py
```diff
@@ -0,0 +1,190 @@
...
+        self._hidden_w_init = hidden_w_init
+        self._hidden_b_init = hidden_b_init
+        self._output_nonlinearity = output_nonlinearity
+        self._output_w_init = output_w_init
+        self._output_b_init = output_b_init
+        self._layer_normalization = layer_normalization
+        self._dueling = dueling
+
+        obs_dim = self._env_spec.observation_space.shape
+        action_dim = env_spec.action_space.flat_dim
```
CatherineSue: self._env_spec.action_space.flat_dim

garage/tf/q_functions/discrete_cnn_q_function.py
```diff
@@ -0,0 +1,190 @@
+"""Discrete MLP QFunction."""
```
CatherineSue: We should state this CNN network actually supports CNN2MLP. This is not the same as the CNN* primitive in the current garage. If you think the naming is ok, please add more details to this documentation.

garage/tf/q_functions/discrete_mlp_q_function.py
```diff
@@ -69,5 +108,26 @@ def get_qval_sym(self, state_input, name):
...
+            The tf.Tensor output of Discrete MLP QFunction.
         """
-        return self.model.build(state_input, name=name)
+        with tf.variable_scope(self._variable_scope):
+            out = state_input
+            for model in self.models:
+                out = model.build(out, name=name)
+            return out
+
+    def clone(self, name):
```
CatherineSue: Should clone interface be in the base class?

tests/garage/tf/algos/test_dqn.py
```diff
@@ -0,0 +1,55 @@
...
+        max_path_length = 1
+        num_timesteps = 20000
+
+        env = TfEnv(gym.make("CartPole-v0"))
+        replay_buffer = SimpleReplayBuffer(
+            env_spec=env.spec,
+            size_in_transitions=int(5000),
+            time_horizon=max_path_length)
+        qf = DiscreteMLPQFunction(env_spec=env.spec, hidden_sizes=(64, 64))
+        policy = DiscreteQfDerivedPolicy(env_spec=env, qf=qf)
```
CatherineSue: env_spec

garage/tf/algos/dqn.py
```diff
@@ -0,0 +1,282 @@
...
+                 replay_buffer,
+                 max_path_length=200,
+                 num_timesteps=int(1e5),
+                 qf_lr=0.001,
+                 qf_optimizer=tf.train.AdamOptimizer,
+                 discount=1.0,
+                 name=None,
+                 target_network_update_freq=5,
+                 grad_norm_clipping=None,
+                 double_q=False,
```
CatherineSue: Is double_q the same as dueling?

tests/garage/tf/algos/test_dqn.py
```diff
@@ -0,0 +1,55 @@
+"""
+This script creates a test that fails when garage.tf.algos.DDPG performance is
```
CatherineSue: garage.tf.algos.DQN

garage/tf/algos/dqn.py
ahtsan: No, they are different.

garage/tf/q_functions/discrete_mlp_q_function.py
ahtsan: Yes, it will eventually be an interface in Model too.

CatherineSue: Please update it into the base class.

garage/tf/algos/dqn.py
CatherineSue: i c. thanks for explaining it.

garage/tf/models/cnn_model.py
ryanjulian: Seems like you have already violated the `Model` interface by making `name` an kwarg: https://github.com/rlworkgroup/garage/blob/master/garage/tf/models/base.py#L159

tests/garage/tf/algos/test_dqn.py
ahtsan: around 30s.

garage/tf/models/cnn_model.py
ahtsan: oh yes, so a name is actually required.

garage/tf/q_functions/discrete_mlp_q_function.py
ahtsan: you mean store q_vals in a dict? or storing the models in a dict?

ryanjulian: the models. if the order of insertion changes at any time, your code will break.

you can also just keep the models in instance variables (`self.model1` etc.) and *also* add them to the list `self.models`...

ahtsan:
Benchmark with 1M timesteps and random seed=26
SpaceInvadersNoFrameskip-v4_1M
SeaquestNoFrameskip-v4_1M
QbertNoFrameskip-v4_1M
PongNoFrameskip-v4_1M
EnduroNoFrameskip-v4_1M
BreakoutNoFrameskip-v4_1M
BeamRiderNoFrameskip-v4_1M

ahtsan:
Fixed all comments above. I will add more tests.

garage/tf/q_functions/base2.py
```diff
@@ -0,0 +1,44 @@
...
+
+        Args:
+            input_phs (list[tf.Tensor]): Recommended to be positional
+            arguments, e.g. def get_qval_sym(self, state_input, action_input).
+        """
+        raise NotImplementedError
+
+    def clone(self, name):
+        """
+        Return a clone of the Q-function.
```
ryanjulian: does this share parameters of the original, or no?

ahtsan: It doesn't, only copys the configuration. It is usually used for the target network/q function so we don't actually need to also copy the parameters. I will add that in the docstring.

ryanjulian: yeah -- just make sure that the behavior is clear

garage/tf/algos/dqn.py
```diff
@@ -0,0 +1,206 @@
...
+                outputs=[self._loss, self._optimize_loss])
+
+    def train_once(self, itr, paths):
+        """Train the algorithm once."""
+        epoch = itr / self.n_epoch_cycles
+
+        self.episode_rewards.extend(paths['undiscounted_returns'])
+        last_average_return = np.mean(self.episode_rewards)
+        for train_itr in range(self.n_train_steps):
+            if self.replay_buffer.n_transitions_stored >= self.min_buffer_size:  # noqa: E501
```
ryanjulian: this can be split with ()

garage/tf/algos/dqn.py
```diff
@@ -0,0 +1,206 @@
...
+                        axis=1)
+                else:
+                    # r + max_a(Q'(s', _)) - Q(s, a)
+                    future_best_q_val = tf.reduce_max(
+                        self.target_qf.q_vals, axis=1)
+
+                q_best_masked = (1.0 - self.done_t_ph) * future_best_q_val
+                # if done, it's just reward
+                # else reward + discount * future_best_q_val
+                target_q_values = self.reward_t_ph + self.discount * q_best_masked  # noqa: E501
```
ryanjulian: this can be split with ()

garage/tf/misc/tensor_utils.py
```diff
@@ -13,6 +13,23 @@ def run(*input_vals):
...
+    update_ops = []
+    init_ops = []
+    assert len(variables) == len(target_variables)
+    for var, target_var in zip(variables, target_variables):
+        init_ops.append(tf.assign(target_var, var))
+        if tau:
+            update_ops.append(
+                tf.assign(target_var, tau * var + (1.0 - tau) * target_var))
+
+    if tau:
```
ryanjulian: what if `tau` is 0?

garage/tf/models/cnn_model_max_pooling.py
```diff
@@ -0,0 +1,59 @@
...
+from garage.tf.models.base import Model
+
+
+class CNNModelWithMaxPooling(Model):
+    """
+    CNN Model with max pooling.
+
+    Args:
+        filter_dims: Dimension of the filters.
+        num_filters: Number of filters.
```
ryanjulian: needs types

garage/envs/wrappers/atari_env.py
```diff
@@ -0,0 +1,27 @@
...
+        env (gym.Env): The environment to be wrapped.
+    """
+
+    def __init__(self, env):
+        super().__init__(env)
+
+    def step(self, action):
+        """gym.Env step function."""
+        obs, reward, done, info = self.env.step(action)
+        return np.asarray(obs), reward, done, info
```
ryanjulian: is `obs` not an np.ndarray?

`np.asarray()` copies (AFAIK), so you will want to avoid the copy operation if you can.

garage/misc/tensor_utils.py
```diff
@@ -153,3 +154,10 @@ def truncate_tensor_dict(tensor_dict, truncated_len):
         else:
             ret[k] = truncate_tensor_list(v, truncated_len)
     return ret
+
+
+def normalize_pixel_batch(env_spec, observations):
```
ryanjulian: please provide a docstring and a test

garage/replay_buffer/base.py
```diff
@@ -49,15 +50,20 @@ def sample(self, batch_size):
         """Sample a transition of batch_size."""
         raise NotImplementedError
 
-    def add_transitions(self, **kwargs):
+    def add_transition(self, **kwargs):
         """Add one transition into the replay buffer."""
+        transition = {k: [v] for k, v in kwargs.items()}
+        self.add_transitions(**transition)
+
+    def add_transitions(self, **kwargs):
```
ryanjulian: please provide docstrings for this file. if they live in subclasses, please move them to this base class.

is this file well-tested?

garage/replay_buffer/simple_replay_buffer.py
```diff
@@ -25,16 +25,16 @@ def sample(self, batch_size):
             buffer[key] = self._buffer[key][:self._current_size]
```
ryanjulian: you can remove the constructor on this class -- it is equivalent to inheritance with no overriding constructor.

ryanjulian: if `sample` changes the behavior from the base class, please specify in the docstring how it differs.

garage/tf/algos/dqn.py
```diff
@@ -0,0 +1,206 @@
...
+        observations = normalize_pixel_batch(self.env_spec, observations)
+        next_observations = normalize_pixel_batch(self.env_spec,
+                                                  next_observations)
+        loss, _ = self._train_qf(observations, actions, rewards, dones,
+                                 next_observations)
+
+        return loss
+
+
+def huber_loss(x, delta=1.0):
```
ryanjulian: this exists in tensorflow: https://www.tensorflow.org/api_docs/python/tf/losses/huber_loss

is their version not useful?

garage/tf/misc/tensor_utils.py
```diff
@@ -13,6 +13,23 @@ def run(*input_vals):
     return run
 
 
+def get_target_ops(variables, target_variables, tau=None):
```
ryanjulian: please provide a docstring and a test

garage/tf/q_functions/discrete_cnn_q_function.py
```diff
@@ -0,0 +1,221 @@
...
+    def input(self):
+        """Input tf.Tensor of the Q-function."""
+        return self.model.networks['default'].input
+
+    def get_qval_sym(self, state_input, name):
+        """
+        Symbolic graph for q-network.
+
+        Args:
+            state_input: The state input tf.Tensor to the network.
```
ryanjulian: types

tests/garage/tf/models/test_cnn_model.py
```diff
@@ -60,6 +61,54 @@ def test_output_value(self, filter_sizes, in_channels, out_channels,
 
         assert np.array_equal(output, expected_output)
 
+    @params(((1, ), (3, ), (32, ), (1, ), (1, 1),
```
ryanjulian: is this weird formatting from yapf? if so, please make use `# yapf: disable` to stop it

tests/garage/tf/q_functions/test_discrete_cnn_q_function.py
```diff
@@ -0,0 +1,182 @@
...
+from tests.fixtures.models import SimpleMLPModel
+
+
+class TestDiscreteCNNQFunction(TfGraphTestCase):
+    def setUp(self):
+        super().setUp()
+        self.env = TfEnv(DummyDiscretePixelEnv())
+        self.obs = self.env.reset()
+
+    @params(((3, ), (5, ), (1, )), ((3, ), (5, ), (2, )), ((3, 3), (5, 5),
```
ryanjulian: fix weird formatting

tests/garage/tf/q_functions/test_discrete_cnn_q_function.py
```diff
@@ -0,0 +1,182 @@
...
+        action_dim = self.env.action_space.n
+        expected_output = np.full(action_dim, 0.5)
+        outputs = self.sess.run(qf.q_vals, feed_dict={qf.input: [self.obs]})
+        assert np.array_equal(outputs[0], expected_output)
+        outputs = self.sess.run(
+            qf.q_vals, feed_dict={qf.input: [self.obs, self.obs, self.obs]})
+        for output in outputs:
+            assert np.array_equal(output, expected_output)
+
+    @params(((3, ), (5, ), (1, )), ((3, ), (5, ), (2, )), ((3, 3), (5, 5),
```
ryanjulian: fix weird formatting

tests/garage/tf/q_functions/test_discrete_cnn_q_function.py
```diff
@@ -0,0 +1,182 @@
...
+        action_dim = self.env.action_space.n
+        expected_output = np.full(action_dim, 0.5)
+        outputs = self.sess.run(qf.q_vals, feed_dict={qf.input: [self.obs]})
+        assert np.array_equal(outputs[0], expected_output)
+        outputs = self.sess.run(
+            qf.q_vals, feed_dict={qf.input: [self.obs, self.obs, self.obs]})
+        for output in outputs:
+            assert np.array_equal(output, expected_output)
+
+    @params(((3, ), (5, ), (1, ), (1, 1), (1, 1)), ((3, ), (5, ), (2, ),
```
ryanjulian: fix weird formatting

tests/garage/tf/q_functions/test_discrete_cnn_q_function.py
```diff
@@ -0,0 +1,182 @@
...
+        action_dim = self.env.action_space.n
+        expected_output = np.full(action_dim, 0.5)
+        outputs = self.sess.run(qf.q_vals, feed_dict={qf.input: [self.obs]})
+        assert np.array_equal(outputs[0], expected_output)
+        outputs = self.sess.run(
+            qf.q_vals, feed_dict={qf.input: [self.obs, self.obs, self.obs]})
+        for output in outputs:
+            assert np.array_equal(output, expected_output)
+
+    @params(((3, ), (5, ), (1, )), ((3, ), (5, ), (2, )), ((3, 3), (5, 5),
```
ryanjulian: fix weird formatting

garage/replay_buffer/base.py
```diff
@@ -27,6 +27,7 @@ def __init__(self, env_spec, size_in_transitions, time_horizon):
         :param time_horizon: time horizon of rollout
         """
         self._current_size = 0
+        self._current_ptr = 0
```
ryanjulian: it looks like your diff coverage on this file is only 75%. please improve it.

garage/envs/wrappers/atari_env.py
```diff
@@ -0,0 +1,27 @@
+"""Episodic life wrapper for gym.Env."""
+import gym
+import numpy as np
+
+
+class AtariEnv(gym.Wrapper):
```
ryanjulian: this class is barely covered (60%). please test it.

garage/envs/wrappers/atari_env.py
ahtsan: obs is not a `np.ndarray`. It is stated in the docstring that `This wrapper convert the observations returned from baselines wrapped environment, which is a LazyFrames object into numpy arrays.`

garage/tf/algos/dqn.py
```diff
@@ -0,0 +1,206 @@
...
+                # else reward + discount * future_best_q_val
+                target_q_values = self.reward_t_ph + self.discount * q_best_masked  # noqa: E501
+
+                td_error = q_selected - tf.stop_gradient(target_q_values)
+                loss = huber_loss(td_error)
+                self._loss = tf.reduce_mean(loss)
+
+            with tf.name_scope('optimize_ops'):
+                optimizer = self.qf_optimizer(self.qf_lr)
+                if self.grad_norm_clipping is not None:
```
ryanjulian: this mode is not tested at all. please test it or remove.

garage/tf/algos/dqn.py
```diff
@@ -0,0 +1,206 @@
...
+    Args:
+        env_spec (garage.envs.env_spec.EnvSpec): Environment specification.
+        max_path_length (int): Maximum path length. The episode will terminate
+            when length of trajectory reaches max_path_length.
+        qf_lr (float): Learning rate for Q-Function.
+        qf_optimizer (tf.Optimizer): Optimizer for Q-Function.
+        discount (float): Discount factor for rewards.
+        name (str): Name of the algorithm.
+        target_network_update_freq (int): Frequency of updating target network.
+        grad_norm_clipping (float): How much to clip gradient. If None, no
```
ryanjulian: is `0.0` also no clipping -- if not, why not just use 0 as no clipping?
what is the scale of this parameter? is it [0,1], [0, 100], [-1,1]? i don't know.

garage/tf/algos/dqn.py
```diff
@@ -0,0 +1,206 @@
...
+                tabular.record('Epoch', epoch)
+                tabular.record('AverageReturn', np.mean(self.episode_rewards))
+                tabular.record('StdReturn', np.std(self.episode_rewards))
+                tabular.record('Episode100RewardMean', mean100ep_rewards)
+                tabular.record('{}/Episode100LossMean'.format(self.qf.name),
+                               mean100ep_qf_loss)
+        return last_average_return
+
+    @overrides
+    def get_itr_snapshot(self, itr):
```
ryanjulian: i think Keren's latest PR will force you to make this class pickleable, so you might as well do that now.

garage/tf/q_functions/base2.py
```diff
@@ -0,0 +1,47 @@
...
+        """
+        Symbolic graph for q-network.
+
+        All derived classes should implement this function.
+
+        Args:
+            input_phs (list[tf.Tensor]): Recommended to be positional
+            arguments, e.g. def get_qval_sym(self, state_input, action_input).
+        """
+        raise NotImplementedError
```
ryanjulian: please make this an ABC so you can use `pass` instead

garage/tf/q_functions/base2.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def get_qval_sym(self, *input_phs):
+        """
+        Symbolic graph for q-network.
+
+        All derived classes should implement this function.
+
+        Args:
+            input_phs (list[tf.Tensor]): Recommended to be positional
+            arguments, e.g. def get_qval_sym(self, state_input, action_input).
```
ryanjulian: please correct the indentation here

garage/tf/algos/dqn.py
ahtsan: I guess not doing `tf.clip_by_norm` in a for loop will save some time, if we don't do gradient clipping.
```python
                if self.grad_norm_clipping is not None:
                    gradients = optimizer.compute_gradients(
                        self._loss, var_list=self.qf.get_trainable_vars())
                    for i, (grad, var) in enumerate(gradients):
                        if grad is not None:
                            gradients[i] = (tf.clip_by_norm(
                                grad, self.grad_norm_clipping), var)
                    self._optimize_loss = optimizer.apply_gradients(gradients)
                else:
                    self._optimize_loss = optimizer.minimize(
                        self._loss, var_list=self.qf.get_trainable_vars())
```

garage/tf/algos/dqn.py
ahtsan: benchmarks work fine without clipping gradient. Maybe we can just remove it.

ryanjulian: didn't you say it's required to compare to openai/baselines?

ahtsan: openai/baselines use gradient clipping. I tried and benchmarks work comparable with them without gradient clipping. If we want to use exactly the same setting as theirs, we can also do gradient clipping.

ryanjulian: let's leave it for now but make a comment about why it's there. i'd prefer if you added a test which covered it.

garage/tf/algos/dqn.py
ryanjulian: i see, good point.

ryanjulian: please still update the comment with the proper range.

garage/tf/algos/dqn.py
ahtsan: I will use it.

garage/tf/algos/dqn.py
ahtsan: It is pickleable now.

garage/replay_buffer/simple_replay_buffer.py
```diff
@@ -25,16 +21,16 @@ def sample(self, batch_size):
...
 
         # Select which episodes to use
-        time_horizon = buffer["action"].shape[1]
-        rollout_batch_size = buffer["action"].shape[0]
+        time_horizon = buffer['action'].shape[1]
```
CatherineSue: Is it still valid to have `time_horizon` since the replay buffer now stores variable-length episodes? It may beyond the scope of this PR. I just think the interface seems a bit confusing now.

tests/garage/misc/test_tensor_utils.py
```diff
@@ -0,0 +1,26 @@
...
+
+from garage.misc.tensor_utils import normalize_pixel_batch
+from garage.tf.envs import TfEnv
+from tests.fixtures.envs.dummy import DummyBoxEnv
+from tests.fixtures.envs.dummy import DummyDiscretePixelEnv
+
+
+class TestTensorUtil(unittest.TestCase):
+    def test_normalize_pixel_patch(self):
+        env = TfEnv(DummyDiscretePixelEnv())
```
CatherineSue: Could you also test it that a non pixel env won't get normalized?

garage/tf/algos/dqn.py
CatherineSue: I think we can keep it? It is recommended to do gradient clipping with DQN in practice.

garage/replay_buffer/simple_replay_buffer.py
ahtsan: Is it because there is no `max_path_length` anymore?

garage/tf/algos/dqn.py
ahtsan: Sure. I also added a test for gradient clipping.

PR #585, opened 2019-03-19
Replace config_personal with environment variables

ryanjulian:
This allows garage to be more-easily used as a library.

Note that this change should be fairly temporary, since it only affects
the legacy experiment runner, i.e. garage.config will be deleted soon
anyway

* Namespaces all config variables with GARAGE_
* Removes unused config variables
* Attempts to get each config variables from os.environ
* Exempts old experiment runner from flake8 (it's beyond help)


garage/misc/logger/__init__.py
```diff
@@ -1,11 +1,11 @@
 """Global logger module.
 
 This module instantiates a global logger according to configuration.
-Use LOG_TENSORBOARD=True in config.py to enable Tensorboard.
+Use GARAGE_LOG_TENSORBOARD=True in config.py to enable Tensorboard.
```
gonzaiva: "Set the environment variable GARAGE_LOG_TENSORBOARD=True to enable Tensorboard"

PR #589, opened 2019-03-20
Install akro from PyPI

ryanjulian:


krzentner:
Do we want to wait for the akro 0.0.2 release before merging this? Besides that, this LGTM.

ryanjulian:
I don't see much reason to wait for v0.0.2 when v0.0.1 is what's in the repo now (and is already included via git hash).

PR #593, opened 2019-03-22
Refactor CEM under the new interface

### LINKED ISSUE #460: CEM and CMA don't support parallel sampling from TF policy
Currently, both ES are under the original branch. I managed to run them with some minor modifications (added the session initialization) with a single sampler but running with multiple samplers causes problems (that completely undermines the advantages of ES). 
Are there any immediate plans to move both algorithms into the main TF branch?

naeioi:
CEM is an important baseline gradient-free algorithm. It is gradient-free in that it does not rely on machine learning framework (e.g. Tensorflow) to optimize policy. The old CEM implementation uses a custom sampler which has some sampling issues (See #460), and is not compatible with the new runner interface. This PR refactors CEM to support general samplers (`OnPolicyVectorizedSampler` or `BatchSampler`).

This closes #460.


examples/cem_cartpole.py
```diff
@@ -0,0 +1,50 @@
...
+        algo = CEM(
+            env_spec=env.spec,
+            policy=policy,
+            baseline=baseline,
+            best_frac=0.1,
+            max_path_length=100,
+            n_samples=n_samples)
+
+        runner.setup(algo, env, sampler_cls=OnPolicyVectorizedSampler)
+        # NOTE: make sure that n_epoch_cycles == n_samples !
```
ryanjulian: why?

tests/garage/tf/algos/test_cem.py
```diff
@@ -0,0 +1,38 @@
+from garage.algos import CEM
+from garage.baselines import LinearFeatureBaseline
+from garage.experiment import LocalRunner
+from garage.tf.envs import TfEnv
+from garage.tf.policies import CategoricalMLPPolicy
+from garage.tf.samplers import OnPolicyVectorizedSampler
+from tests.fixtures import TfGraphTestCase
+
+
+class TestERWR(TfGraphTestCase):
```
ryanjulian: TestCEM?

ryanjulian:
Does it train?

naeioi:
It trains. I will post the plot of return later.

naeioi:
The plot of `AverageReturn` of cart-pole. This performance should be able to reproduce by `example/cem_cartpole.py`. Parameters: batch_size=1000, n_samples=20, best_frac=0.05

CEM converges to optimal policy quickly on cart-pole. Note that since each epoch runs n_samples iterations, it's true that it trains slower than other advanced algorithms.
cem_rtn

examples/cem_cartpole.py
naeioi: Because CEM samples one policy per epoch cycle, and only update policy distribution at the end of the epoch. `n_samples` specifies the number of policies to be sampled in one epoch.

naeioi: But I think there should be a way to verify this. I just can't figure how to do it.

garage/algos/cem.py
```diff
@@ -1,189 +1,121 @@
...
+        # epoch-wise
+        self.cur_std = self.init_std
+        self.cur_mean = self.policy.get_param_values()
+        # epoch-cycle-wise
+        self.cur_params = self.cur_mean
+        self.all_returns = []
+        self.all_params = [self.cur_mean.copy()]
+        # fixed
+        self.n_best = int(n_samples * best_frac)
+        assert self.n_best >= 1, \
```
CatherineSue: Use parentheses here.

garage/algos/cem.py
```diff
@@ -1,189 +1,121 @@
...
+            rtn = max(self.all_returns)
+            self.all_returns.clear()
+            self.all_params.clear()
+
+        # -- Stage: Generate a new policy for next path sampling
+        self.cur_params = self.sample_params(itr)
+        self.all_params.append(self.cur_params.copy())
+        self.policy.set_param_values(self.cur_params)
+
+        logger.dump_tabular(with_prefix=False)
```
CatherineSue: logger.log(tabular)

garage/algos/cem.py
```diff
@@ -1,189 +1,121 @@
...
+        sample_std = np.sqrt(
+            np.square(self.cur_std) +
+            np.square(self.extra_std) * extra_var_mult)
+        return np.random.standard_normal(
+            self.n_params) * sample_std + self.cur_mean
+
+    def train_once(self, itr, paths):
+        epoch = itr // self.n_samples
+        i_sample = itr - epoch * self.n_samples
+        logger.record_tabular('Epoch', epoch)
```
CatherineSue: tabular.record

examples/cem_cartpole.py
ryanjulian: another way to impose this constraint is to subclass Runner just for CEM and override setup()

CatherineSue: Can we rename n_samples to policy_update_freq? I think n_samples is a bit confusing.

tests/garage/tf/algos/test_cem.py
naeioi: Fixed.

examples/cem_cartpole.py
naeioi: @ryanjulian But this will require picking up an algorithm-specific runner. 
Maybe a simpler way is to do it without subclassing, by checking if the algo is CEM and do verify.

naeioi: @CatherineSue I think `n_samples` is more natural for people familar with CEM, because it is defined as is in the literature.

garage/algos/cem.py
naeioi: Fixed.

examples/cem_cartpole.py
ryanjulian: @naeioi how would that work if CEM and the Runner were not in the same codebase?

You've already defined an algorithm-specific sampler. You just did it in a comment, rather than in code :stuck_out_tongue:

ryanjulian: FWIW it's up to you. Feel free to merge.

Issue #597, opened 2019-03-25
Broken plot in Tensorboard


naeioi:
Logger adds unnecessary evolving prefix to the name of tabular variables. Data points of a single viable are scattered into many plots, each with a single point. 

The name of plots look like the following:
```
_experiment_2019_03_25_16_00_26_0001__epoch__0___AverageDiscountedReturn
_experiment_2019_03_25_16_00_26_0001__epoch__0___AverageReturn
_experiment_2019_03_25_16_00_26_0001__epoch__0___Entropy
...
```

PR #598, opened 2019-03-26
Fix prefix handling in TensorBoardOutput

CatherineSue:
Closes: #597.


garage/logger/tensor_board_output.py
```diff
@@ -56,7 +56,7 @@ def record(self, data, prefix=''):
 
     def _record_tabular(self, data, prefix, step):
         for key, value in data.as_dict.items():
-            self._record_kv(prefix + key, value, step)
+            self._record_kv(key, value, step)
```
ryanjulian: please remove the unused parameter `prefix` from `_record_tabular`

ryanjulian:
i think you need to update the test for TensorBoardOutput as well

PR #599, opened 2019-03-26
Fixup TensorBoardOutput

ryanjulian:
TensorBoardOutput currently uses the `prefix` parameter as a key path
prefix for TensorBoard, which is not how prefix is currently used
around the codebase (e.g. in LocalTfRunner).

This produces bad output to TensorBoard (e.g. every epoch in its own
key space)

This PR makes TensorBoardOutput ignore the `prefix` parameter to
record(). In the future, we should figure out if we can unify the
notion of prefixes and key spaces in the logger.


CatherineSue:
This PR is a duplication of #598. Should I close #598?

ryanjulian:
See #601

PR #602, opened 2019-03-26
Refactor CMA-ES under the new interface

naeioi:
This PR refactors CMA-ES to support the new runner interface. This implementation of CMA-ES learns very slowly, even for a simple task like cart-pole. 

Test will be added later.


examples/cma_es_cartpole.py
```diff
@@ -0,0 +1,50 @@
...
+
+        algo = CMAES(
+            env_spec=env.spec,
+            policy=policy,
+            baseline=baseline,
+            max_path_length=100,
+            n_samples=n_samples)
+
+        runner.setup(algo, env, sampler_cls=OnPolicyVectorizedSampler)
+        # NOTE: make sure that n_epoch_cycles == n_samples !
```
ryanjulian: seems like `n = 2` for the instances which need a special runner constraint. how can we write this down in code instead of comments?

naeioi: Yes. And I want to elaborate on this constraint a little bit.

The `n_epoch_cycles` is defined as `number of calls to algo.train_once() in one epoch`. It controls how large an epoch is (in terms of iterations). It only concerns to **epoch-wise** logging and snapshotting, but not training (at least so far).  

So if someone breaks `n_epoch_cycles == n_samples` constraint, he/she will observe
- Unexpected snapshot frequency.
- Unexpected log of epoch time.

He/she will not observe
- Performance regression. (not learning, learn slowly, etc.)
- Wrong iteration-wise logging. (e.g. AverageReturn logged by sampler)

naeioi: And I am not in favor of subclassing `LocalRunner` for `CEM`  because I fear that once we add one algo-specific runner into the codebase, we end up adding many more sub-runner for new algos. In fact, off-policy algorithms also have this constraint. 

The conflict here is that our encapsulation is leaky.  Today we have constraint on `n_epoch_cycles` by algorithm, some other day in the future we might find a new constaint on some sampler parameters by algorithm. Decoupling is good, but also hard, [explained](https://www.joelonsoftware.com/2002/11/11/the-law-of-leaky-abstractions/).

ryanjulian: alright sgtm

Issue #523, opened 2019-02-21
Refactor GaussianMLP* to use garage.tf.model


ryanjulian:
* [x] GaussianMLPBaseline
* [x] GaussianMLPPolicy
* [x] GaussianMLPRegressor

ahtsan:
Closed by #606.

ryanjulian:
Let's not close this until the old code has been fully replaced.

ahtsan:
closed by #606 and #574

PR #606, opened 2019-03-28
Refactor GaussianMLPRegressor to use models

ahtsan:
New class GaussianMLPRegressorWithModel which does the following:
* Remove Serializable
* It contains a GaussianMLPRegressorModel which inherits from
  GaussianMLPModel

### DIFF
--- a/garage/tf/baselines/gaussian_mlp_baseline_with_model.py
+++ b/garage/tf/baselines/gaussian_mlp_baseline_with_model.py
@@ -1,0 +1,65 @@
+"""This module implements gaussian mlp baseline."""
+import numpy as np
+
+from garage.misc.overrides import overrides
+from garage.np.baselines import Baseline
+from garage.tf.regressors import GaussianMLPRegressorWithModel
+
+
+class GaussianMLPBaselineWithModel(Baseline):
+    """A value function using gaussian mlp network."""
+
+    def __init__(
+            self,
+            env_spec,
+            subsample_factor=1.,
+            num_seq_inputs=1,
+            regressor_args=None,
+            name="GaussianMLPBaselineWithModel",
+    ):
+        """
+        Constructor.
+
+        :param env_spec:
+        :param subsample_factor:
+        :param num_seq_inputs:
+        :param regressor_args:
+        """
+        super().__init__(env_spec)
+        if regressor_args is None:
+            regressor_args = dict()
+
+        self._regressor = GaussianMLPRegressorWithModel(
+            input_shape=(
+                env_spec.observation_space.flat_dim * num_seq_inputs, ),
+            output_dim=1,
+            name=name,
+            **regressor_args)
+        self.name = name
+
+    @overrides
+    def fit(self, paths):
+        """Fit regressor based on paths."""
+        observations = np.concatenate([p["observations"] for p in paths])
+        returns = np.concatenate([p["returns"] for p in paths])
+        self._regressor.fit(observations, returns.reshape((-1, 1)))
+
+    @overrides
+    def predict(self, path):
+        """Predict value based on paths."""
+        return self._regressor.predict(path["observations"]).flatten()
+
+    @overrides
+    def get_param_values(self, **tags):
+        """Get parameter values."""
+        return self._regressor.get_param_values(**tags)
+
+    @overrides
+    def set_param_values(self, flattened_params, **tags):
+        """Set parameter values to val."""
+        self._regressor.set_param_values(flattened_params, **tags)
+
+    @overrides
+    def get_params_internal(self, **tags):
+        """Get internal parameters."""
+        return self._regressor.get_params_internal(**tags)
--- a/garage/tf/baselines/gaussian_mlp_baseline_with_model.py
+++ b/garage/tf/baselines/gaussian_mlp_baseline_with_model.py
@@ -1,4 +1,4 @@
-"""This module implements gaussian mlp baseline."""
+"""A value function (baseline) based on a GaussianMLP model."""
 import numpy as np
 
 from garage.misc.overrides import overrides
@@ -15,7 +15,7 @@ class GaussianMLPBaselineWithModel(Baseline):
             subsample_factor=1.,
             num_seq_inputs=1,
             regressor_args=None,
-            name="GaussianMLPBaselineWithModel",
+            name='GaussianMLPBaselineWithModel',
     ):
         """
         Constructor.
@@ -40,14 +40,14 @@ class GaussianMLPBaselineWithModel(Baseline):
     @overrides
     def fit(self, paths):
         """Fit regressor based on paths."""
-        observations = np.concatenate([p["observations"] for p in paths])
-        returns = np.concatenate([p["returns"] for p in paths])
+        observations = np.concatenate([p['observations'] for p in paths])
+        returns = np.concatenate([p['returns'] for p in paths])
         self._regressor.fit(observations, returns.reshape((-1, 1)))
 
     @overrides
     def predict(self, path):
         """Predict value based on paths."""
-        return self._regressor.predict(path["observations"]).flatten()
+        return self._regressor.predict(path['observations']).flatten()
 
     @overrides
     def get_param_values(self, **tags):
--- a/garage/tf/baselines/gaussian_mlp_baseline_with_model.py
+++ b/garage/tf/baselines/gaussian_mlp_baseline_with_model.py
@@ -7,7 +7,7 @@ from garage.tf.regressors import GaussianMLPRegressorWithModel
 
 
 class GaussianMLPBaselineWithModel(Baseline):
-    """A value function using gaussian mlp network."""
+    """A value function using Gaussian MLP network."""
 
     def __init__(
             self,
--- a/garage/tf/models/__init__.py
+++ b/garage/tf/models/__init__.py
@@ -1,5 +1,9 @@
 from garage.tf.models.base import Model
 from garage.tf.models.gaussian_mlp_model import GaussianMLPModel
+from garage.tf.models.gaussian_mlp_regressor_model import (
+    GaussianMLPRegressorModel)
 from garage.tf.models.mlp_model import MLPModel
 
-__all__ = ["Model", "GaussianMLPModel", "MLPModel"]
+__all__ = [
+    "Model", "GaussianMLPModel", "GaussianMLPRegressorModel", "MLPModel"
+]
--- a/garage/tf/models/__init__.py
+++ b/garage/tf/models/__init__.py
@@ -1,9 +1,5 @@
 from garage.tf.models.base import Model
 from garage.tf.models.gaussian_mlp_model import GaussianMLPModel
-from garage.tf.models.gaussian_mlp_regressor_model import (
-    GaussianMLPRegressorModel)
 from garage.tf.models.mlp_model import MLPModel
 
-__all__ = [
-    "Model", "GaussianMLPModel", "GaussianMLPRegressorModel", "MLPModel"
-]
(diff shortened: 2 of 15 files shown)


ahtsan:
I am adding the remaining tests.

garage/tf/baselines/gaussian_mlp_baseline_with_model.py
```diff
@@ -0,0 +1,65 @@
+"""This module implements gaussian mlp baseline."""
```
ryanjulian: A value function (baseline) based on a GaussianMLP model.

garage/tf/models/__init__.py
```diff
@@ -1,5 +1,9 @@
 from garage.tf.models.base import Model
 from garage.tf.models.gaussian_mlp_model import GaussianMLPModel
+from garage.tf.models.gaussian_mlp_regressor_model import (
```
ryanjulian: i think that this should go in the package `garage.tf.regressors` since it's regressor-specific

ryanjulian:
This looks great and is making a huge difference. One you merge this you'll be able to close #523 , which is a huge milestone because the GaussianMLP primitives are by far the most commonly-used ones.

Some small notes:
* Please strive to have *unit-only* coverage approach 100% for new or highly-revised code. This means we want the tests for "this module only* to cover most of the module, rather than relying on tests from other modules (which might use this module)
You can use this snippet to check:
    ```sh
    # run the tests while measuring coverage
    # replace arguments with (garage/tf/baselines, garage.tf.baselines) to run the same for baselines
    coverage run --source=garage/tf/regressors -m nose2 -c setup.cfg garage.tf.regressors
    coverage report  # prints a report to the terminal
    coverage html   # generates a detailed report as web pages in ./htmlcov/
    xdg-open htmlcov/index.html   # opens the detailed report in your browser
    ```
* When you edit a file in a PR anyway, please take a second to replace double-quoted strings with single-quoted strings. This helps us clean up the codebase as we go.
* Same as above, but don't use `+` or `%` for formatting or concatenating strings -- instead use `str.format()`

garage/tf/models/gaussian_mlp_regressor_model.py
```diff
@@ -0,0 +1,95 @@
...
+    def network_output_spec(self):
+        """Network output spec."""
+        return [
+            'sample', 'means', 'log_stds', 'std_param', 'normalized_means',
+            'normalized_log_stds', 'x_mean', 'x_std', 'y_mean', 'y_std', 'dist'
+        ]
+
+    def _build(self, state_input):
+
+        input_shape = tuple(state_input.get_shape().as_list()[1:])
```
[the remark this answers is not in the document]
ahtsan: perhaps we need a input_spec as a constructor argument?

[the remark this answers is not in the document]
ryanjulian: i think that is a good pattern to use, since policies all have an environment spec, other primitives should have an input/output spec.

that might require you to refactor a lot of code, though, so you can skip it this time (and file an issue instead) if you want to focus on getting this merged.

garage/tf/baselines/gaussian_mlp_baseline_with_model.py
```diff
@@ -0,0 +1,65 @@
+"""A value function (baseline) based on a GaussianMLP model."""
+import numpy as np
+
+from garage.misc.overrides import overrides
+from garage.np.baselines import Baseline
+from garage.tf.regressors import GaussianMLPRegressorWithModel
+
+
+class GaussianMLPBaselineWithModel(Baseline):
+    """A value function using gaussian mlp network."""
```
ryanjulian: Gaussian (generally capitalized)

MLP (capitalized because it's an acronym)

tests/fixtures/regressors/__init__.py
```diff
@@ -0,0 +1,4 @@
+from tests.fixtures.regressors.simple_gaussian_mlp_regressor import (
+    SimpleGaussianMLPRegressor)
+
+__all__ = ["SimpleGaussianMLPRegressor"]
```
ryanjulian: single quotes

tests/fixtures/regressors/simple_gaussian_mlp_regressor.py
```diff
@@ -0,0 +1,30 @@
...
+
+from garage.tf.regressors import StochasticRegressor
+
+
+class SimpleGaussianMLPRegressor(StochasticRegressor):
+    """Simple GaussianMLPModel for testing."""
+
+    def __init__(self, input_shape, output_dim, name, *args, **kwargs):
+        super().__init__(input_shape, output_dim, name)
+        self.param_values = tf.get_variable("{}/params".format(name), [100])
```
ryanjulian: single quotes

garage/tf/regressors/gaussian_mlp_regressor_model.py
```diff
@@ -0,0 +1,98 @@
+"""GaussianMLPRegressorModel."""
+import numpy as np
+import tensorflow as tf
+
+from garage.tf.models import GaussianMLPModel
+
+
+class GaussianMLPRegressorModel(GaussianMLPModel):
+    """
+    GaussianMLPRegressorModel.
```
ryanjulian: Please use a description which is not simply the name of the class

garage/tf/regressors/gaussian_mlp_regressor_with_model.py
```diff
@@ -0,0 +1,239 @@
...
+from garage.logger import tabular
+from garage.tf.misc import tensor_utils
+from garage.tf.optimizers import LbfgsOptimizer, PenaltyLbfgsOptimizer
+from garage.tf.regressors import GaussianMLPRegressorModel
+from garage.tf.regressors import StochasticRegressor
+
+
+class GaussianMLPRegressorWithModel(StochasticRegressor):
+    """
+    GaussianMLPRegressorWithModel.
```
ryanjulian: name as the first line is not necessary

garage/tf/regressors/gaussian_mlp_regressor_model.py
```diff
@@ -0,0 +1,98 @@
+"""GaussianMLPRegressorModel."""
+import numpy as np
+import tensorflow as tf
+
+from garage.tf.models import GaussianMLPModel
+
+
+class GaussianMLPRegressorModel(GaussianMLPModel):
```
CatherineSue: Should this class in model module or regressor? @ryanjulian

tests/garage/tf/baselines/test_gaussian_mlp_baseline.py
```diff
@@ -0,0 +1,65 @@
...
+        box_env = TfEnv(DummyBoxEnv(obs_dim=obs_dim))
+        with mock.patch(('garage.tf.baselines.'
+                         'gaussian_mlp_baseline_with_model.'
+                         'GaussianMLPRegressorWithModel'),
+                        new=SimpleGaussianMLPRegressor):
+            gmb = GaussianMLPBaselineWithModel(env_spec=box_env.spec)
+            new_gmb = GaussianMLPBaselineWithModel(
+                env_spec=box_env.spec, name='GaussianMLPBaselineWithModel2')
+        old_param_values = gmb.get_param_values()
+        old_new_param_values = new_gmb.get_param_values()
```
CatherineSue: I think here `new_param_values` is a fine name?

tests/garage/tf/regressors/test_gaussian_mlp_regressor_with_model.py
```diff
@@ -0,0 +1,103 @@
...
+        expected = [[0], [-1], [-0.707], [0], [0.707], [1], [0]]
+        assert np.allclose(prediction, expected, rtol=0, atol=0.1)
+
+    def test_fit_custom(self):
+        gmr = GaussianMLPRegressorWithModel(
+            input_shape=(1, ),
+            output_dim=1,
+            subsample_factor=0.9,
+            normalize_inputs=False,
+            normalize_outputs=False)
```
CatherineSue: Emm is this test legit? I don’t see [this line](https://github.com/rlworkgroup/garage/pull/606/files#diff-1d908bac267c9cae4bb670633344a07eR193) is checked here. I am not clear the purpose of this test.

tests/garage/tf/regressors/test_gaussian_mlp_regressor_with_model.py
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        prediction = gmr.predict(paths['observations'])
+
+        expected = [[0], [-1], [-0.707], [0], [0.707], [1], [0]]
+        assert np.allclose(prediction, expected, rtol=0, atol=0.1)
+
+    def test_fit_without_trusted_region(self):
+        gmr = GaussianMLPRegressorWithModel(
+            input_shape=(1, ), output_dim=1, use_trust_region=False)
```
CatherineSue: What is the purpose of this test? To test if regressor works properly with `use_trust_region=False`?

tests/garage/tf/regressors/test_gaussian_mlp_regressor_with_model.py
ahtsan: `normalize_inputs` and `normalize_outputs` are by default True. This test is for testing when they are False.

tests/garage/tf/regressors/test_gaussian_mlp_regressor_with_model.py
ahtsan: Yes

tests/garage/tf/baselines/test_gaussian_mlp_baseline.py
ahtsan: Yes that is a fine name, I can change it.

garage/tf/regressors/gaussian_mlp_regressor_model.py
ahtsan: It was in model module. Ryan said it is regressor-specific and better to put it in regressor module.

tests/garage/tf/regressors/test_gaussian_mlp_regressor_with_model.py
CatherineSue: Yes I am aware of it. I meant that I am confused because in this test you don’t check that line or check if the outputs is correctly unnormalized.

ahtsan: I see. The purpose of this test is to see if it can still fit properly. You are right we should assert as much as we can.

tests/garage/tf/regressors/test_gaussian_mlp_regressor_with_model.py
CatherineSue: I am confused. If PenaltyLbfgsOptimizer works well then this test will be correct. But PenaltyLbfgsOptimizer is never tested.

tests/garage/tf/regressors/test_gaussian_mlp_regressor_with_model.py
CatherineSue: I guess a better question is I am not sure what this test is trying to check. A state check or a interaction check? Is the purpose to check the fit result is correct with customized parameters? Then it can be many combinations, why not split it to one unnormalized test and one `subsample_factor=.9`?

ahtsan: Yes it makes sense. Separating into smaller checks, and also assert about something. I will modify that.

garage/tf/regressors/gaussian_mlp_regressor_model.py
CatherineSue: I c. Thanks for pointing it out. I’m making this resolved.

PR #607, opened 2019-03-28
Use scipy.stats distribution types in the logger

ryanjulian:
Replaces tensorflow_probability types used in TensorBoardOutput with
equivalent types from scipy.stats. Also adds a simple mechanism for
logging histograms of raw samples called `garage.logger.Histogram`.

The origin of this PR is a critical performance implication of using
tensorflow_probability types to represent distributions. In practice,
algorithms will need to create a new tfp.Distribution object every time
they log a distribution value. The logger cannot evaluate most ops in
the algorithm's TF graph, because they require a feed_dict to evalaute.

To get around this, most people would just construct a new Distribution
op with their concrete data:
```python
with tf.name_scope('logging'):
    dist = tfp.distributions.MultivariateNormalDiag(
        loc=means, scale_diag=stds)

tabular.record('Normal', dist)
```

Unfortunately, this modified the TF graph which is slow. Repeated
modifications to the graph appear make the runtime of `tf.Session.run`
scale quadratically in the number of modifications. This slows
algorithms to a crawl in a few hundred iterations.

I expect this situation might be fixed in TF 2.0, but for now it is
safest to remove this feature from the logger and use scipy.stats,
which provides just as much expressiveness.

### DIFF
--- a/garage/logger/__init__.py
+++ b/garage/logger/__init__.py
@@ -2,6 +2,7 @@
 
 This module instantiates a global logger singleton.
 """
+from garage.logger.histogram import Histogram
 from garage.logger.logger import Logger, LogOutput
 from garage.logger.simple_outputs import StdOutput, TextOutput
 from garage.logger.tabular_input import TabularInput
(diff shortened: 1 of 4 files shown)


ryanjulian:
image
Time per iteration for the same algorithm, with and without logging using tfp.Distributions types.

PR #612, opened 2019-04-05
Fix LocalRunner hanging for BatchSamplers

gautams3:
Set my tf session to use a config
https://github.com/tensorflow/tensorflow/issues/14442#issuecomment-343501993

### DIFF
--- a/garage/experiment/local_tf_runner.py
+++ b/garage/experiment/local_tf_runner.py
@@ -66,7 +66,7 @@ class LocalRunner:
         if max_cpus > 1:
             from garage.sampler import singleton_pool
             singleton_pool.initialize(max_cpus)
-        self.sess = sess or tf.Session()
+        self.sess = sess or tf.Session(config=tf.ConfigProto(use_per_session_threads=True))
         self.has_setup = False
         self.plot = False
 


naeioi:
Nice catch. Could you provide a case to show this is useful? 

Also, this PR should only be able to fix garage for multithreading, not multiprocessing, as pointed out in https://github.com/tensorflow/tensorflow/issues/14442#issuecomment-343501993.

garage/experiment/local_tf_runner.py
```diff
@@ -66,7 +66,7 @@ def __init__(self, sess=None, max_cpus=1):
         if max_cpus > 1:
             from garage.sampler import singleton_pool
             singleton_pool.initialize(max_cpus)
-        self.sess = sess or tf.Session()
+        self.sess = sess or tf.Session(config=tf.ConfigProto(use_per_session_threads=True))
```
ryanjulian: @naeioi  I think this is testable.

@gautams3 can you provide a minimal launcher which reproduces the problem?

gautams3: This needs a sampler and a LocalRunner for the sampler to call `start_worker`. One option is the [point_reach_embed launcher in embed2learn](https://github.com/ryanjulian/embed2learn-private/blob/388934023af752f5137791319f26164bebdae256/launchers/point_reach_embed.py)

naeioi: Could you grant me access to embed2learn? I cannot open the link above.

ryanjulian: done

ryanjulian:
@gautams3 can you rebase this on master and make it pass the CI?

ryanjulian:
@naeioi can you get this submitted?

naeioi:
Closing this because the issue of freezing of batch sampler is already handled in the local runner. See https://github.com/rlworkgroup/garage/blob/master/garage/experiment/local_tf_runner.py#L54-L63.
I tested embed2learn and didn't see it stuck. Gautam and I discussed on this PR and we concluded that it won't fix anything.

PR #627, opened 2019-04-18
Make off-policy sampler respect batch_size

naeioi:
Some off-policy algorithm such as DDPG and DQN require that environment
is not reset across batches. This PR adds `no_reset` parameter to
off-policy sampler that when it is set to True, the sampler will not
reset environment and the batch_size is strictly respected (previously
the sampler always returns full paths)

**Changelog**
- Add `no_reset` paramter to `OffPolicyVectoriedSampler` constructor. 
When set to `True`, the sampler will not reset environment across batches and 
the `batch_size` is strictly respected. 
- Add a field of boolean list `dones` in the return of `obtrain_samples()`

**Note**
The `return` of a path is currently computed in `process_samples()` in the sampler.
After this PR, the paths returned by `obtrain_samples()` are no longer guaranteed
to be full paths, so the `return` will not be correctly computed. 
The solution is to move `process_samples()` to algorithm and make algorithm responsible 
to compute `return`. This will be in another PR.

### DIFF
--- a/garage/tf/samplers/off_policy_vectorized_sampler.py
+++ b/garage/tf/samplers/off_policy_vectorized_sampler.py
@@ -35,7 +35,11 @@ class OffPolicyVectorizedSampler(BatchSampler):
         super(OffPolicyVectorizedSampler, self).__init__(algo, env, n_envs)
         self.n_envs = n_envs
         self.no_reset = no_reset
-        self.last_obses = None
+
+        self._last_obses = None
+        self._last_uncounted_discount = [0] * n_envs
+        self._last_running_length = [0] * n_envs
+        self._last_success_count = [0] * n_envs
 
     @overrides
     def start_worker(self):
@@ -61,10 +65,10 @@ class OffPolicyVectorizedSampler(BatchSampler):
         :return: A list of paths.
         """
         paths = []
-        if not self.no_reset or self.last_obses is None:
+        if not self.no_reset or self._last_obses is None:
             obses = self.vec_env.reset()
         else:
-            obses = self.last_obses
+            obses = self._last_obses
         dones = np.asarray([True] * self.vec_env.num_envs)
         running_paths = [None] * self.vec_env.num_envs
         n_samples = 0
@@ -90,7 +94,7 @@ class OffPolicyVectorizedSampler(BatchSampler):
                     input_obses)
 
             next_obses, rewards, dones, env_infos = self.vec_env.step(actions)
-            self.last_obses = next_obses
+            self._last_obses = next_obses
             agent_infos = tensor_utils.split_tensor_dict_list(agent_infos)
             env_infos = tensor_utils.split_tensor_dict_list(env_infos)
             n_samples += len(next_obses)
@@ -127,10 +131,28 @@ class OffPolicyVectorizedSampler(BatchSampler):
                                                    env_infos, dones):
                 if running_paths[idx] is None:
                     running_paths[idx] = dict(
-                        rewards=[], env_infos=[], dones=[])
+                        rewards=[],
+                        env_infos=[],
+                        dones=[],
+                        undiscounted_return=self._last_uncounted_discount[idx],
+                        # running_length: Length of path up to now
+                        # Note that running_length is not len(rewards)
+                        # Because a path may not be complete in one batch
+                        running_length=self._last_running_length[idx],
+                        success_count=self._last_success_count[idx])
+
                 running_paths[idx]["rewards"].append(reward)
                 running_paths[idx]["env_infos"].append(env_info)
                 running_paths[idx]["dones"].append(done)
+                running_paths[idx]["running_length"] += 1
+                running_paths[idx]["undiscounted_return"] += reward
+                running_paths[idx]["success_count"] += env_info.get(
+                    "is_success") or 0
+
+                self._last_uncounted_discount[idx] += reward
+                self._last_success_count[idx] += env_info.get(
+                    "is_success") or 0
+                self._last_running_length[idx] += 1
 
                 if done or n_samples >= batch_size:
                     paths.append(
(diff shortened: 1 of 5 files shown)


garage/tf/samplers/off_policy_vectorized_sampler.py
```diff
@@ -22,17 +22,20 @@
...
         :param algo: Algorithms.
         :param n_envs: Number of parallelized sampling envs.
+        :param no_reset: If Ture, don't reset environement after sampling.
         """
         if n_envs is None:
             n_envs = int(algo.rollout_batch_size)
         super(OffPolicyVectorizedSampler, self).__init__(algo, env, n_envs)
         self.n_envs = n_envs
+        self.no_reset = no_reset
+        self.last_obses = None
```
ryanjulian: _private

garage/tf/samplers/off_policy_vectorized_sampler.py
```diff
@@ -22,17 +22,20 @@
...
 
         :param algo: Algorithms.
         :param n_envs: Number of parallelized sampling envs.
+        :param no_reset: If Ture, don't reset environement after sampling.
         """
         if n_envs is None:
             n_envs = int(algo.rollout_batch_size)
         super(OffPolicyVectorizedSampler, self).__init__(algo, env, n_envs)
         self.n_envs = n_envs
+        self.no_reset = no_reset
```
ryanjulian: _private

garage/tf/samplers/off_policy_vectorized_sampler.py
```diff
@@ -118,20 +127,20 @@ def obtain_samples(self, itr, batch_size):
...
-                        rewards=[],
-                        env_infos=[],
-                    )
+                        rewards=[], env_infos=[], dones=[])
                 running_paths[idx]["rewards"].append(reward)
                 running_paths[idx]["env_infos"].append(env_info)
+                running_paths[idx]["dones"].append(done)
 
-                if done:
+                if done or n_samples >= batch_size:
```
CatherineSue: This would generate a `paths` which have an undone path right? So we would have an `undiscounted_returns` like `[98, 2]` (`max_path_length = 100` and `env.reward = 1`)?

naeioi: Yes. `paths` may contain undone paths. However, each path is one element in `paths`. In your example, the `undiscounted_returns` will be like `[ 'undiscounted_returns': [1], 'undiscounted_returns': [1]]`. The first `[1]` is the end of a path. The second `[1]` is the first sample of a path. Because `process_sample()` doesn't have access to previous steps, `return` cannot be correctly computed.

naeioi:
Three fields are added to the return of`obtain_samples()` of `OffPolicyVectoriedSampler`. Theses field are persistent across batches, meaning that they are computed w.r.t the whole path, even though a path returned in one batch might not be complete.
- `running_length`: The full length of the path sampled so far.
- `undiscounted_length`
- `success_count`

garage/tf/samplers/off_policy_vectorized_sampler.py
```diff
@@ -22,17 +22,24 @@
...
     """This class implements OffPolicyVectorizedSampler."""
 
-    def __init__(self, algo, env, n_envs=None):
+    def __init__(self, algo, env, n_envs=None, no_reset=True):
         """
         Construct an OffPolicyVectorizedSampler.
 
         :param algo: Algorithms.
         :param n_envs: Number of parallelized sampling envs.
+        :param no_reset: If Ture, don't reset environement after sampling.
```
ahtsan: True

garage/tf/samplers/off_policy_vectorized_sampler.py
```diff
@@ -151,12 +192,13 @@ def process_samples(self, itr, paths):
         """
         success_history = []
         for path in paths:
```
ahtsan: I think you forgot to remove this block of code. `success_history` is assigned again below.

garage/tf/samplers/off_policy_vectorized_sampler.py
```diff
@@ -151,12 +192,13 @@ def process_samples(self, itr, paths):
...
-            if "is_success" in path["env_infos"]:
-                success = np.array(path["env_infos"]["is_success"])
-                success_rate = np.mean(success)
-                success_history.append(success_rate)
+            success_history.append(
+                path["success_count"] / path["running_length"])
 
-        undiscounted_returns = [sum(path["rewards"]) for path in paths]
+        success_history = [
```
ahtsan: why remove the checking `if 'is_success' in path['env_infos']`?

tests/garage/tf/samplers/test_off_policy_vectorized_sampler.py
```diff
@@ -0,0 +1,68 @@
...
+from garage.replay_buffer import SimpleReplayBuffer
+from garage.tf.algos import DDPG
+from garage.tf.envs import TfEnv
+from garage.tf.policies import ContinuousMLPPolicy
+from garage.tf.q_functions import ContinuousMLPQFunction
+from garage.tf.samplers import OffPolicyVectorizedSampler
+from tests.fixtures import TfGraphTestCase
+
+
+class TestOffPolicyVectorizedSampler(TfGraphTestCase):
```
ryanjulian: can you test this without running an entire algorithm? it seems like you should be able to use a couple test fixtures and/or mocking to write this as a unit test.

naeioi: The sampler needs an algorithm policy to do sampling, but no algorithm is actually run, just to setup sampler.

garage/tf/samplers/off_policy_vectorized_sampler.py
naeioi: `success_history` is used only for monitoring. Previously, if `is_success` is not reported in `env_infos`, the `sucess_rate` being logged will be zero. It is the same result effect as omitting `is_success` here.

garage/tf/samplers/off_policy_vectorized_sampler.py
naeioi: Nice catch!

garage/tf/samplers/off_policy_vectorized_sampler.py
```diff
@@ -22,17 +22,24 @@
...
     """This class implements OffPolicyVectorizedSampler."""
 
-    def __init__(self, algo, env, n_envs=None):
+    def __init__(self, algo, env, n_envs=None, no_reset=True):
         """
         Construct an OffPolicyVectorizedSampler.
 
         :param algo: Algorithms.
         :param n_envs: Number of parallelized sampling envs.
+        :param no_reset: If True, don't reset environement after sampling.
```
ryanjulian: please update these docstrings to Google style.

tests/garage/tf/samplers/test_off_policy_vectorized_sampler.py
ryanjulian: perhaps you can get by with a mock algorithms or policy? anyway, it's not a big deal.

ryanjulian:
if anson is happy, i'm happy.

please try to over some of the special cases missed by your existing tests:
https://codecov.io/gh/rlworkgroup/garage/compare/35889a708a7a7072c6b6ce88dfcbccb055a41b23...9587a380acc6fc97657f30adc9c18614ae768840/diff

also seem my comment about updating docstrings.

otherwise, LGTM

Issue #628, opened 2019-04-19
Estimate entropy in NPO


maxiaoba:
In line 417 of npo.py, should 
policy_neg_log_likeli_valid = self.policy.distribution.log_likelihood_sym(...)
actually be
policy_neg_log_likeli_valid = -self.policy.distribution.log_likelihood_sym(...)?

The same for line 395.

ryanjulian:
@CatherineSue can you take a look?

ryanjulian:
@maxiaoba I think you're right and it might explain why we had to turn off entropy-regularization a few months ago. 

@CatherineSue PTAL

CatherineSue:
Hi @maxiaoba , you are correct that there should be a negative sign. I've fixed this in #632 .

maxiaoba:
Cool. Thanks!

Issue #435, opened 2018-12-18
Replace Enum parameters with strings


amolchanov86:
Hi everyone, 
I would like to share the problem that I encountered that explains why I typically prefer using strings for parameters that are choice-based instead of enums.

I had to copy and re-define all classes of the hierarchy ppo > npo > batch_polopt, but I forgot to import my definition of PGLoss in ppo.py (i.e. I accidentally kept line import garage.tf.algos.npo import PGLoss instead of using the reference to my newly created module) to feed PGLoss.CLIP to the NPO initializer. As a result, PGLoss(Enum) class was used from the original garage module, but then re-defined in my module with the same name. As a result line like PGLoss.CLIP == PGLoss.CLIP would evaluate in False during if statements inside of NPO since those were different PGLoss classes.
That is why I prefer using strings and just give a description of options you have with a string in parameter description. This way, say "LOSS_CLIP" == "LOSS_CLIP" would, obviously, give True, as it should.
On top of that using strings allows storing all parameters in yaml file and feed them directly to corresponding functions/classes after reading yaml to a dictionary. 
Hence, I believe this might a feature request to consider.

PS: Btw, thanks guys that you actually have asserts when if-else runs out of options (the rule of thumb - check and assert inputs whenever possible). It is a good style! Having default silent options often leads to bugs that are very hard to notice.

ryanjulian:
@krzentner what are your thoughts?

@amolchanov86 can you show the code snippet which broke? I've been on the fence about enums vs strings for a while, too.

ryanjulian:
Did your NPO fall through to https://github.com/rlworkgroup/garage/blob/master/garage/tf/algos/npo.py#L329? That is our hope.

Note that we made the Enum string-valued so that you can pass the string instead and it will still work.

I wonder if it would have been better if the Enum were strings instead? I guess you would not have gotten any error, but is that what we want? This only seems to be a problem if you're going to copy NPO, rather than subclass it or use it.

Alas, I wish Python had a proper `switch`--then static analysis could force you to handle all possible switch cases in an Enum. Right now we can only force it in code review.

krzentner:
This report is surprising to me, because I thought that string-valued enums with the same value would compare equal to each other. I guess I shouldn't be surprised that they don't. It also seems like static analysis should catch this, but I don't think that's actually a strong argument, since I don't believe anyone actually uses static analysis in garage (or most of Python). We certainly can't expect our users to be using e.g. mypy.

[...]

PR #632, opened 2019-04-23
Add SOA entropy regularization methods to tf/NPO

CatherineSue:
* Add max entropy method to tf/NPO.
* Fix the neg_log_likeli entropy estimation in tf/NPO.
* Add documentation of sane entropy regularization default.
* Replace Enum params with strings.

### DIFF
--- a/src/garage/tf/algos/ppo.py
+++ b/src/garage/tf/algos/ppo.py
@@ -1,5 +1,4 @@
 from garage.tf.algos.npo import NPO
-from garage.tf.algos.npo import PGLoss
 from garage.tf.optimizers import FirstOrderOptimizer
 
 
@@ -27,18 +26,29 @@ class PPO(NPO):
             conjunction with center_adv the advantages will be
             standardized before shifting.
         fixed_horizon (bool): Whether to fix horizon.
-        pg_loss (str): Objective.
+        pg_loss (str): A string from: 'vanilla', 'surrogate',
+            'surrogate_clip'. The type of loss functions to use.
         lr_clip_range (float): The limit on the likelihood ratio between
             policies, as in PPO.
         max_kl_step (float): The maximum KL divergence between old and new
             policies, as in TRPO.
-        optimizer (float): The optimizer of the algorithm.
-        optimizer_args (dict): Optimizer args.
+        optimizer (object): The optimizer of the algorithm. Should be the
+            optimizers in garage.tf.optimizers.
+        optimizer_args (dict): The arguments of the optimizer.
         policy_ent_coeff (float): The coefficient of the policy entropy.
-        use_softplus_entropy (bool): Whether to use softplus entropy.
-        stop_entropy_gradient (bool): Whether to stop entropy gradient.
+            Setting it to zero would mean no entropy regularization.
+        use_softplus_entropy (bool): Whether to estimate the softmax
+            distribution of the entropy to prevent the entropy from being
+            negative.
+        use_neg_logli_entropy (bool): Whether to estimate the entropy as the
+            negative log likelihood of the action.
+        stop_entropy_gradient (bool): Whether to stop the entropy gradient.
+        entropy_method (str): A string from: 'max', 'regularized',
+            'no_entropy'. The type of entropy method to use. 'max' adds the
+            dense entropy to the reward for each time step. 'regularized' adds
+            the mean entropy to the surrogate objective. See
+            https://arxiv.org/abs/1805.00909 for more details.
         name (str): The name of the algorithm.
-
     """
 
     def __init__(self,
@@ -52,7 +62,7 @@ class PPO(NPO):
                  center_adv=True,
                  positive_adv=False,
                  fixed_horizon=False,
-                 pg_loss=PGLoss.SURROGATE_CLIP,
+                 pg_loss='surrogate_clip',
                  lr_clip_range=0.01,
                  max_kl_step=0.01,
                  optimizer=None,
@@ -61,29 +71,31 @@ class PPO(NPO):
                  use_softplus_entropy=False,
                  use_neg_logli_entropy=False,
                  stop_entropy_gradient=False,
+                 entropy_method='no_entropy',
                  name='PPO'):
         if optimizer is None:
             optimizer = FirstOrderOptimizer
             if optimizer_args is None:
                 optimizer_args = dict()
-        super(PPO, self).__init__(
+        super().__init__(
             env_spec=env_spec,
             policy=policy,
             baseline=baseline,
             scope=scope,
-            gae_lambda=gae_lambda,
             max_path_length=max_path_length,
             discount=discount,
-            pg_loss=pg_loss,
-            lr_clip_range=lr_clip_range,
-            max_kl_step=max_kl_step,
+            gae_lambda=gae_lambda,
             center_adv=center_adv,
             positive_adv=positive_adv,
             fixed_horizon=fixed_horizon,
+            pg_loss=pg_loss,
+            lr_clip_range=lr_clip_range,
+            max_kl_step=max_kl_step,
+            optimizer=optimizer,
+            optimizer_args=optimizer_args,
             policy_ent_coeff=policy_ent_coeff,
             use_softplus_entropy=use_softplus_entropy,
             use_neg_logli_entropy=use_neg_logli_entropy,
             stop_entropy_gradient=stop_entropy_gradient,
-            optimizer=optimizer,
-            optimizer_args=optimizer_args,
+            entropy_method=entropy_method,
             name=name)
--- a/src/garage/tf/algos/tnpg.py
+++ b/src/garage/tf/algos/tnpg.py
@@ -1,4 +1,4 @@
-from garage.tf.algos.npo import NPO, PGLoss
+from garage.tf.algos.npo import NPO
 from garage.tf.optimizers import ConjugateGradientOptimizer
 
 
@@ -26,31 +26,43 @@ class TNPG(NPO):
             conjunction with center_adv the advantages will be
             standardized before shifting.
         fixed_horizon (bool): Whether to fix horizon.
-        pg_loss (str): Objective.
+        pg_loss (str): A string from: 'vanilla', 'surrogate',
+            'surrogate_clip'. The type of loss functions to use.
         lr_clip_range (float): The limit on the likelihood ratio between
-            policies.
+            policies, as in PPO.
         max_kl_step (float): The maximum KL divergence between old and new
-            policies.
-        optimizer (float): The optimizer of the algorithm.
-        optimizer_args (dict): Optimizer args.
+            policies, as in TRPO.
+        optimizer (object): The optimizer of the algorithm. Should be the
+            optimizers in garage.tf.optimizers.
+        optimizer_args (dict): The arguments of the optimizer.
         policy_ent_coeff (float): The coefficient of the policy entropy.
-        use_softplus_entropy (bool): Whether to use softplus entropy.
-        stop_entropy_gradient (bool): Whether to stop entropy gradient.
+            Setting it to zero would mean no entropy regularization.
+        use_softplus_entropy (bool): Whether to estimate the softmax
+            distribution of the entropy to prevent the entropy from being
+            negative.
+        use_neg_logli_entropy (bool): Whether to estimate the entropy as the
+            negative log likelihood of the action.
+        stop_entropy_gradient (bool): Whether to stop the entropy gradient.
+        entropy_method (str): A string from: 'max', 'regularized',
+            'no_entropy'. The type of entropy method to use. 'max' adds the
+            dense entropy to the reward for each time step. 'regularized' adds
+            the mean entropy to the surrogate objective. See
+            https://arxiv.org/abs/1805.00909 for more details.
         name (str): The name of the algorithm.
-
     """
 
     def __init__(self,
                  env_spec,
                  policy,
                  baseline,
+                 scope=None,
                  max_path_length=500,
                  discount=0.99,
                  gae_lambda=0.98,
                  center_adv=True,
                  positive_adv=False,
                  fixed_horizon=False,
-                 pg_loss=PGLoss.SURROGATE,
+                 pg_loss='surrogate',
                  lr_clip_range=0.01,
                  max_kl_step=0.01,
                  optimizer=None,
@@ -59,6 +71,7 @@ class TNPG(NPO):
                  use_softplus_entropy=False,
                  use_neg_logli_entropy=False,
                  stop_entropy_gradient=False,
+                 entropy_method='no_entropy',
                  name='TNPG'):
         if optimizer is None:
             optimizer = ConjugateGradientOptimizer
@@ -67,10 +80,11 @@ class TNPG(NPO):
                 optimizer_args = default_args
             else:
                 optimizer_args = dict(default_args, **optimizer_args)
-        super(TNPG, self).__init__(
+        super().__init__(
             env_spec=env_spec,
             policy=policy,
             baseline=baseline,
+            scope=scope,
             max_path_length=max_path_length,
             discount=discount,
             gae_lambda=gae_lambda,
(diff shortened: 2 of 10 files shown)


CatherineSue:
### Entropy Experiments
I've experimented the `max` and `regularized` entropy methods with both GaussianMLPPolicy and GaussianLSTMPolicy.
Here are some sane defaults:
`entropy_method='max'`, `center_adv=False`, `stop_gradient=True`
`entropy_method='regularized'`, `stop_gradient=False`, `use_neg_logli_entropy=False`

The following curve shows the entropy of GaussianMLPPolicy and GaussianLSTMPolicy. Curves that have increasing entropy have `policy_ent_coeff=1.0`. Curves that have decreasing entropy have `policy_ent_coeff=-1.0`. Note that it takes longer for GaussianLSTMPolicy to train, so I stopped their training before the `max_iter` is reached.

Screen Shot 2019-04-23 at 3 56 04 PM
Screen Shot 2019-04-23 at 4 02 17 PM

garage/tf/algos/npo.py
```diff
@@ -33,14 +33,33 @@ class NPO(BatchPolopt):
     """
     Natural Policy Gradient Optimization.
 
-    Attributes:
-        name(str): The name of the algorithm.
+    Args:
+        pg_loss(enum): The type of loss functions to use.
```
ryanjulian: you can call out the specific enum type (PGLoss) rather than just a generic enum

Note that we have issue #435 open to remove this enum anyway. Perhaps now is a good time to just replace it with the string?

garage/tf/algos/npo.py
```diff
@@ -33,14 +33,33 @@ class NPO(BatchPolopt):
...
+        optimizer_args(dict): The arguments of the optimizer.
+        name(str): The name of the algorithm.
         policy_ent_coeff(float): The coefficient of the policy entropy.
-        optimizer(float): The optimizer of the algorithm.
+        entropy_method(str): A string from: 'max', 'regularized'. The type
+            of entropy method to use. 'max' adds the dense entropy to the
+            reward for each time step. 'regularized' adds the mean entropy
+            to the surrogate objective. See https://arxiv.org/abs/1805.00909
+            for more details.
+            !!NOTE!!: make sure when setting entropy_method to 'max', set
```
ryanjulian: Please put this as a note outside the argument list.

```python
"""
Args:
    foo (str): Stuff

Notes
    Notes about some stuff
"""
```

See example here https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

garage/tf/algos/npo.py
```diff
@@ -33,14 +33,33 @@ class NPO(BatchPolopt):
     """
     Natural Policy Gradient Optimization.
 
-    Attributes:
-        name(str): The name of the algorithm.
+    Args:
```
ryanjulian: please also note which params are optional

garage/tf/algos/npo.py
```diff
@@ -33,14 +33,33 @@ class NPO(BatchPolopt):
...
+        name(str): The name of the algorithm.
         policy_ent_coeff(float): The coefficient of the policy entropy.
-        optimizer(float): The optimizer of the algorithm.
+        entropy_method(str): A string from: 'max', 'regularized'. The type
+            of entropy method to use. 'max' adds the dense entropy to the
+            reward for each time step. 'regularized' adds the mean entropy
+            to the surrogate objective. See https://arxiv.org/abs/1805.00909
+            for more details.
+            !!NOTE!!: make sure when setting entropy_method to 'max', set
+            `center_adv` to False and turn off policy gradient. `center_adv`
```
ryanjulian: long comments like this belong elsewhere in the docstring. argument descriptions should be short

garage/tf/algos/npo.py
```diff
@@ -50,31 +69,39 @@ def __init__(self,
...
             optimizer = LbfgsOptimizer
 
+        if entropy_method == 'max':
+            self._maximum_entropy = True
+            self._entropy_regularzied = False
+        elif entropy_method == 'regularized':
+            self._maximum_entropy = False
+            self._entropy_regularzied = True
+        else:
+            raise NotImplementedError('Unknown entropy_method')
```
ryanjulian: this is a ValueError

garage/tf/algos/npo.py
```diff
@@ -33,14 +33,33 @@ class NPO(BatchPolopt):
...
 
-    Attributes:
-        name(str): The name of the algorithm.
+    Args:
+        pg_loss(enum): The type of loss functions to use.
         lr_clip_range(float): The limit on the likelihood ratio between
             policies, as in PPO.
         max_kl_step(float): The maximum KL divergence between old and new
             policies, as in TRPO.
+        optimizer(object): The optimizer of the algorithm.
```
ryanjulian: is there an optimizer base class we can put here instead?

ryanjulian:
ref #628

garage/tf/algos/npo.py
```diff
@@ -33,14 +33,33 @@ class NPO(BatchPolopt):
...
+        optimizer_args(dict): The arguments of the optimizer.
+        name(str): The name of the algorithm.
         policy_ent_coeff(float): The coefficient of the policy entropy.
-        optimizer(float): The optimizer of the algorithm.
+        entropy_method(str): A string from: 'max', 'regularized'. The type
```
naeioi: Is adding entropy penalty to loss or reward enforced (instead of an option)?

garage/tf/algos/npo.py
CatherineSue: No I don't think there is one. We don't have optimizer base class in garage/tf.

ryanjulian: Ok it's fine. we can just say (object) and note the optimizer package in the description

garage/tf/algos/npo.py
ryanjulian: it's an option. you can also train with no entropy regularization.

also note that entropy is a bonus, not a penalty (we want more of it)

CatherineSue: should we add `None` check to `entropy_method`? my assumption is users can set `policy_ent_coeff=0.0` if they don't want to use entropy regularization

garage/tf/algos/npo.py
CatherineSue: there is no third way to configure the entropy bonus?

ryanjulian:
ref #435

garage/tf/algos/npo.py
ryanjulian: hmm. you should do what you can to prevent the user from providing mutually-incompatible arguments.

e.g. a non-zero `policy_ent_coeff` when `entropy_method` is not `'no_entropy'` should be an error. Note that I recommend that the entropy method for none be `'no_entropy'` rather than the Python symbol `None`, because `None` represents that the user failed to make a choice or provide a value, *not* that the user has *chosen* to not use entropy.

ryanjulian: please add similar checks for other mutually-incompatible arguments, as you have noted in the comments.

only disallow arguments which make no sense together (e.g. guaranteed to fail or logically inconsistent). don't disallow argument combinations which seem useless or unlikely but not harmful.

ryanjulian:
overall this still LGTM, but please see my note on validating arguments. 

it would also be nice if you could push up the coverage on NPO (it stands at 86%). Based on CodeCov, it seems like a couple of the entropy cases might not be fully-covered: https://codecov.io/gh/rlworkgroup/garage/compare/3c49440b2692045c8c9b0db86a1e550d1bca6e84...4d802f847d9cc8ce014d2d2fb975045f40b73e23/diff

Perhaps this is because some combinations of arguments don't make sense, though?

CatherineSue:
This PR is ready for another review.

garage/tf/algos/npo.py
```diff
@@ -22,59 +20,101 @@
...
+        use_neg_logli_entropy (bool, optional): Whether to estimate the
+            entropy as the negative log likelihood of the action.
+        use_softplus_entropy (bool, optional): Whether to estimate the softmax
+            distribution of the entropy to prevent the entropy from being
+            negative.
+        stop_entropy_gradient (bool, optional): Whether to stop the entropy
+            gradient.
+
+    Note:
+        sane defaults for entropy configuration:
```
ryanjulian: can you add some more explicit info on the combinations of entropy arguments which are valid? even i find it a little confusing.

garage/tf/algos/npo.py
ryanjulian: on second thought, the documentation generator makes it clear what's optional, and these notations are just noise. please remove.

i apologize for changing my mind.

ryanjulian:
Are these uncovered branches in CodeCov yellow because they are not covered, or because the missed branches are not valid combinations of configuration?

https://codecov.io/gh/rlworkgroup/garage/compare/8e24d281b3f299663528034191f8f54934d5c6e7...f2ac9b6e0fe1b9673d6d6c8d04cb457bbdce8144/diff#D2-x423

garage/tf/algos/npo.py
CatherineSue: Could you tell me what information you think is missing here?

CatherineSue:
I didn’t test npo with turning on use_softplus_entropy because I didn’t see any negative entropy in experiments. I’m not sure if it is an invalid configuration.
Another yellow line in ppo is because the default arguments of FirstOrderOptimizer has max_epochs=1000. It takes too long to test it. We can change the default max_epochs to a smaller number, like 10 or 20 and then test it.

ryanjulian:
This seems fine to me -- I was just curious.

max_epochs (which should actually be max_minibatches, right?) is pretty big. Feel free to make it more sane if it doesn't change much.

garage/tf/algos/npo.py
CatherineSue: Which documentation generator can make it clear about optional parameters? I tried with pydoc and pdoc but they don’t specify optional parameters.

ryanjulian: For instance, in akro the module documentation looks like this:
https://akro.readthedocs.io/en/latest/_apidoc/akro.html#module-akro.box

You can see the function signature with kwargs, so it's obvious the kwargs are optional and the args are not.

garage/tf/algos/npo.py
ryanjulian: sorry, this seems sufficient to me.

tests/garage/tf/algos/test_npo.py
```diff
@@ -43,9 +43,9 @@ def test_npo_pendulum(self):
 
             env.close()
 
-    def test_npo_unknown_pg_loss(self):
+    def test_npo_with_unknown_pg_loss(self):
         """Test NPO with unkown policy gradient loss."""
-        env = TfEnv(normalize(gym.make("InvertedDoublePendulum-v2")))
+        env = TfEnv(normalize(gym.make('InvertedDoublePendulum-v2')))
```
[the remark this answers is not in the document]
CatherineSue: Done.

tests/garage/tf/algos/test_trpo.py
```diff
@@ -68,8 +68,37 @@ def test_trpo_unknown_kl_constraint(self):
...
+                max_path_length=100,
+                discount=0.99,
+                gae_lambda=0.98,
+                policy_ent_coeff=0.0,
+                kl_constraint='soft')
+            runner.setup(algo, env)
+            last_avg_ret = runner.train(n_epochs=10, batch_size=2048)
+            assert last_avg_ret > 50
+
+            env.close()
```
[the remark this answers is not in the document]
CatherineSue: Done.

garage/tf/algos/npo.py
```diff
@@ -355,7 +399,8 @@ def _build_policy_loss(self, i):
                 else:
                     raise NotImplementedError('Unknown PGLoss')
```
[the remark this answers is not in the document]
CatherineSue: Moved the check to the constructor. I used `in` to check if `pg_loss` is valid because in the constructor we only need to check if an unknown value is passed.

garage/tf/algos/npo.py
```diff
@@ -22,59 +20,101 @@
...
         self._use_neg_logli_entropy = use_neg_logli_entropy
         self._stop_entropy_gradient = stop_entropy_gradient
-
         self._pg_loss = pg_loss
         if optimizer is None:
             if optimizer_args is None:
                 optimizer_args = dict()
             optimizer = LbfgsOptimizer
 
+        if entropy_method == 'max':
```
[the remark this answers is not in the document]
CatherineSue: Moved.

tests/garage/tf/algos/test_trpo_with_model.py
```diff
@@ -36,6 +36,6 @@ def test_trpo_recurrent_cartpole(self):
 
             runner.setup(algo, env)
             last_avg_ret = runner.train(n_epochs=10, batch_size=2048)
-            assert last_avg_ret > 85
+            assert last_avg_ret > 75
```
CatherineSue: 85 seems still to be flaky as it didn't pass the ci check. I changed it to 75.

src/garage/tf/algos/ppo.py
```diff
@@ -61,29 +71,31 @@ def __init__(self,
                  use_softplus_entropy=False,
                  use_neg_logli_entropy=False,
                  stop_entropy_gradient=False,
+                 entropy_method='no_entropy',
                  name='PPO'):
         if optimizer is None:
```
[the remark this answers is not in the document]
CatherineSue: Changed. The reason we didn't add it to default is that in some algorithms the default optimizer can't be decided in `__init__`. It is rare though.

src/garage/tf/algos/tnpg.py
```diff
@@ -65,21 +65,14 @@ def __init__(self,
...
                  stop_entropy_gradient=False,
                  entropy_method='no_entropy',
                  name='TNPG'):
-        if optimizer is None:
-            optimizer = ConjugateGradientOptimizer
-            default_args = dict(max_backtracks=1)
-            if optimizer_args is None:
-                optimizer_args = default_args
-            else:
-                optimizer_args = dict(default_args, **optimizer_args)
```
zequnyu: According to the logic here we want to keep `max_backtracks=1 ` for the `optimizer_args ` even if the user passes their own args. I'm not sure if it's necessary to be this value for this particular algorithm. (I traced back to `ConjugateGradientOptimizer` where the default `max_backtracks` is 15.)

CatherineSue: It is probably necessary. I think setting `max_backtracks` too large will probably affect the training performance of TNPG since it doesn't use the backtrack line search to make sure that the trust region is not broken.

zequnyu: The danger here is that if the user passes some args other than `max_backtracks` it will override the default value. But the previous one (using `default_args`) also has a flaw: `max_backtracks` can not be customised otherwise it raises error for duplicate attribute values.

CatherineSue: No. The previous one won't raise duplicate attribute because python will overwrite the value of an existing key.

CatherineSue: I don't think the previous one is good because the values in `optimizer_args` is likely to overwrite the values in `default_args`.

CatherineSue: I prefer the current because I don't think we should constrain the `max_backtracks=1` or `max_epochs=1` in algorithms. It is unlikely that the algorithm won't be working when `max_epochs=2`. 
If we need fix this, I think set sane defaults in optimizer is better? For instance, the default `max_epochs` in `FirstOrderOptimizer` is 100, that is not a sane default. Even for a small and simple environment, running each iteration in 100 epochs is going to take a lot of time.

CatherineSue: Sorry, on second thought, I don't really have a preference. I think each has its drawbacks. If it is ok with you, I would like to revert this change (both `optimizer` and `optimizer_args`). I think it is not very important for this PR.

zequnyu: Sure. Ok with me.

PR #633, opened 2019-04-23
Fix experiment resuming

naeioi:
The PR fixes experiment resuming. There are 3 ways supported to resume an interrupted experiment.

1. Resume experiment immediately.
```
 with LocalRunner() as runner:
    runner.restore(snapshot_dir)
    runner.resume()
```
2. Resume experiment with modified training arguments.
```
with LocalRunner() as runner:
    runner.restore(snapshot_dir, resume_now=False)
    runner.train(n_epochs)
```

3. Resume via command line.
```
./examples/resume_training.py --snapshot_dir /saved/dir
```


garage/experiment/local_tf_runner.py
```diff
@@ -43,7 +43,7 @@ class LocalRunner:
 
     """
 
-    def __init__(self, sess=None, max_cpus=1):
+    def __init__(self, sess=None, max_cpus=1, restore_from=None):
```
ryanjulian: how about instead there's an alternative constructor called `Runner.restore_from()`?

garage/experiment/local_tf_runner.py
```diff
@@ -171,11 +179,11 @@ def obtain_samples(self, itr, batch_size):
             One batch of samples.
 
         """
-        if self.n_epoch_cycles == 1:
+        if self.train_args["n_epoch_cycles"] == 1:
```
ryanjulian: referencing a dict like this is hacky-looking. why not just store it in `self.n_epoch_cycles`?

garage/experiment/local_tf_runner.py
```diff
@@ -171,11 +179,11 @@ def obtain_samples(self, itr, batch_size):
...
             logger.log('Obtaining samples...')
         return self.sampler.obtain_samples(itr, batch_size)
 
-    def save_snapshot(self, itr, paths=None):
+    def save(self, epoch, paths=None):
```
ryanjulian: snapshot?

ryanjulian:
@krzentner @zhanpenghe PTAL

garage/experiment/experiment_wrapper.py
```diff
@@ -66,7 +66,7 @@ def run_experiment(argv):
     parser.add_argument(
         '--snapshot_mode',
         type=str,
-        default='all',
+        default='last',
```
ryanjulian: please keep `'all'` as default -- i think this is the behavior most people expect.

examples/resume_training.py
```diff
@@ -1,60 +1,37 @@
...
+
     args = parser.parse_args()
-    parent_dir = os.path.dirname(os.path.realpath(args.file))
-    json_file_path = os.path.join(parent_dir, 'params.json')
-    logger.log('Looking for params.json at {}...'.format(json_file_path))
+    params = dict()
+    params['resume_from_dir'] = args.snapshot_dir
+    if args.resume_epoch is not None:
+        params['resume_epoch'] = args.resume_epoch
+    command = to_local_command(
```
ryanjulian: would it be easier to use the experiment wrapper directly in this example? at least there would be fewer moving parts.

ryanjulian: or even just LocalRunner

garage/experiment/local_tf_runner.py
```diff
@@ -226,6 +324,8 @@ def train(self,
             plot: Visualize policy by doing rollout after each epoch.
             store_paths: Save paths in snapshot.
             pause_for_plot: Pause for plot.
+            _start_epoch: (internal) The starting epoch.
```
ryanjulian: this is not very Pythonic.

if you want a private API, create one. For instance you can have a private _train and that will allow to to specify start_epoch. The public train can depend on the private one.

garage/experiment/local_tf_runner.py
```diff
@@ -240,15 +340,23 @@ def train(self,
             else:
                 batch_size = 40 * self.algo.max_path_length
```
ryanjulian: this is a little bit strange. if there is this magic default, can I at least access it through a class property (and have it be properly documented)

garage/experiment/local_tf_runner.py
```diff
@@ -226,6 +324,8 @@ def train(self,
             plot: Visualize policy by doing rollout after each epoch.
```
ryanjulian: please update all docstrings with types.

garage/logger/snapshotter.py
```diff
@@ -78,3 +82,43 @@ def save_itr_params(self, itr, params):
...
+                    raise FileNotFoundError(errno.ENOENT,
+                                            os.strerror(errno.ENOENT),
+                                            '*.pkl file')
+                files.sort(key=lambda f: int(re.findall(r'\d+', f)[0]))
+                snapshot_file = files[0] if itr == 'first' else files[-1]
+                snapshot_file = osp.join(self._snapshot_dir, snapshot_file)
+
+        assert osp.isfile(snapshot_file)
+
+        return pickle.load(open(snapshot_file, 'rb'))
```
ryanjulian: if you are going to always load with pickle, you need to make sure you're not saving with joblib

garage/experiment/experiment_wrapper.py
```diff
@@ -183,11 +193,12 @@ def run_experiment(argv):
     snapshotter.snapshot_gap = args.snapshot_gap
     logger.push_prefix("[%s] " % args.exp_name)
 
-    if args.resume_from is not None:
-        data = joblib.load(args.resume_from)
-        assert 'algo' in data
-        algo = data['algo']
-        algo.train()
+    if args.resume_from_dir is not None:
+        with LocalRunner() as runner:
```
ryanjulian: maybe this is the form we should encourage using in the example?

garage/experiment/local_tf_runner.py
naeioi: That looks better.

garage/experiment/local_tf_runner.py
naeioi: I am grouping arguments for `train()` in a dict, so that it can be returned by `restore()` as a whole.   But true, if I do it this way, it is not so clear what are included in `train_args` by just reading the code.

examples/resume_training.py
naeioi: Using `python -m garage.experiment.experiment_wrapper --restore_from_dir ...` does the same job. But I feel that experiment_wrapper is too cumbersome. This script hides unrelated parts of experiment_wrapper and provides a lightweight access to the resuming feature.

garage/experiment/local_tf_runner.py
naeioi: Make sense.

garage/experiment/local_tf_runner.py
naeioi: What about making `batch_size` a required parameter. I can see a case where a missing `batch_size` makes sense.

garage/logger/snapshotter.py
naeioi: This is to keep consistent with `Snapshotter.save_itr_params()` that uses pickle. I found that we have mixing uses of pickle and joblib to do dump and load in our codebase. This is not good. Since joblib is faster dealing with ndarray, should we just get rid of pickle?

garage/experiment/experiment_wrapper.py
naeioi: I think so. This is the most convenient way to resume an experiment. The programmatical way is more advanced in that it allows more flexibility.

garage/experiment/local_tf_runner.py
naeioi: I found that the interface will be `Runner.restore_from(cls, snapshot_dir, sess=None, max_cpus=1)`. This is a little bit awkward because it cannot be something as simple as `Runner.restore_from(cls, snapshot_dir)`. 

If consider `runner.restore()` a counterpart of `runner.setup()`, then calling either `restore()` or `setup()` is required to set up an experiment. Following this idea, what about just removing support of `restore_from` from constructor.

garage/experiment/local_tf_runner.py
naeioi: I made `batch_size` required.

examples/resume_training.py
ryanjulian: ok sounds fine. we can jus try to make this experience nicer in the future

garage/experiment/experiment_wrapper.py
ryanjulian: okay, can you use this form in the example then?

garage/experiment/local_tf_runner.py
```diff
@@ -100,10 +103,10 @@ def setup(self, algo, env, sampler_cls=None, sampler_args=None):
             policy weights can be loaded before setup().
 
         Args:
-            algo: An algorithm instance.
-            env: An environement instance.
-            sampler_cls: A sampler class.
-            sampler_args: Arguments to be passed to sampler constructor.
+            algo(RLAlgorithm): An algorithm instance.
```
ryanjulian: please use full type signatures if they are not Python primitives

garage/logger/snapshotter.py
ryanjulian: the main problem is that something saved with `joblib.dump()` cannot be loaded with `pickle.load()`.

I think we should stick with vanilla pickle until we have evidence that joblib is necessary. This doesn't mean we can't use joblib (for instance) in the sampler, because those bytes are never committed to disk.

ryanjulian: Alternatively, as long as joblib is used for both save and load, it's fine.

garage/experiment/experiment_wrapper.py
naeioi: Yes. I put it as recommended in the examples.

garage/logger/snapshotter.py
naeioi: joblib is used in unpickling now.

garage/experiment/experiment_wrapper.py
naeioi: Fixed.

garage/experiment/local_tf_runner.py
ryanjulian: ok sgtm

garage/experiment/local_tf_runner.py
ryanjulian: please use the more explicit style. `types.SimpleNameSpace` might be helpful here.

garage/experiment/local_tf_runner.py
```diff
@@ -257,9 +392,11 @@ def train(self,
                     paths = self.sampler.process_samples(itr, paths)
                     last_return = self.algo.train_once(itr, paths)
                     itr += 1
```
zhanpenghe: Hmm for off-policy, should this be `itr += itr * n_epoch_cycles`?

garage/np/algos/base.py
```diff
@@ -1,7 +1,10 @@
+from garage.core import Serializable
+
+
 class Algorithm:
     pass
 
 
-class RLAlgorithm(Algorithm):
+class RLAlgorithm(Algorithm, Serializable):
```
krzentner: This makes me uncomfortable. Inheriting from Serializable has a lot of effects, and many of them are surprising. I would much rather each algorithm that needs this inherit it itself.

I also don't think most algorithms should need to inherit Serializable, since they should already be picklable.

garage/np/algos/cem.py
```diff
@@ -50,6 +50,7 @@ def __init__(self,
                  extra_std=1.,
                  extra_decay_time=100,
                  **kwargs):
+        self.quick_init(locals())
```
krzentner: Please leave a comment listing the fields that cannot be pickled without this.

garage/np/algos/cma_es.py
```diff
@@ -36,6 +36,7 @@ def __init__(self,
                  discount=0.99,
                  sigma0=1.,
                  **kwargs):
+        self.quick_init(locals())
```
krzentner: Please leave a comment listing the fields that cannot be pickled without this.

garage/tf/algos/ddpg.py
```diff
@@ -65,6 +65,7 @@ def __init__(self,
             max_action(float): Maximum action magnitude.
             name(str): Name of the algorithm shown in computation graph.
         """
+        self.quick_init(locals())
```
krzentner: Please leave a comment listing the fields that cannot be pickled without this.

garage/tf/algos/erwr.py
```diff
@@ -28,6 +28,7 @@ def __init__(self,
                  optimizer_args=None,
                  positive_adv=True,
                  **kwargs):
+        self.quick_init(locals())
```
krzentner: Please leave a comment listing the fields that cannot be pickled without this.

garage/tf/algos/npo.py
```diff
@@ -56,6 +56,7 @@ def __init__(self,
                  use_neg_logli_entropy=False,
                  stop_entropy_gradient=False,
                  **kwargs):
+        self.quick_init(locals())
```
krzentner: Please leave a comment listing the fields that cannot be pickled without this.

garage/tf/algos/ppo.py
```diff
@@ -19,6 +19,7 @@ def __init__(self, optimizer=None, optimizer_args=None, **kwargs):
         :param optimizer_args: Optimizer args.
         :param kwargs:
         """
+        self.quick_init(locals())
```
krzentner: Please leave a comment listing the fields that cannot be pickled without this.

garage/tf/algos/reps.py
```diff
@@ -48,8 +48,9 @@ def __init__(self,
                  optimizer_args=dict(max_opt_itr=50),
                  dual_optimizer=scipy.optimize.fmin_l_bfgs_b,
                  dual_optimizer_args=dict(maxiter=50),
-                 name="REPS",
+                 name='REPS',
                  **kwargs):
+        self.quick_init(locals())
```
krzentner: Please leave a comment listing the fields that cannot be pickled without this.

garage/tf/algos/tnpg.py
```diff
@@ -12,6 +12,7 @@ class TNPG(NPO):
     """
 
     def __init__(self, optimizer=None, optimizer_args=None, **kwargs):
+        self.quick_init(locals())
```
krzentner: Please leave a comment listing the fields that cannot be pickled without this.

garage/tf/algos/trpo.py
```diff
@@ -25,13 +25,14 @@ def __init__(self,
                  optimizer=None,
                  optimizer_args=None,
                  **kwargs):
+        self.quick_init(locals())
```
krzentner: Please leave a comment listing the fields that cannot be pickled without this.

garage/tf/algos/vpg.py
```diff
@@ -9,6 +9,7 @@ class VPG(NPO):
     """
 
     def __init__(self, optimizer=None, optimizer_args=None, **kwargs):
+        self.quick_init(locals())
```
krzentner: Please leave a comment listing the fields that cannot be pickled without this.

krzentner: This should not be called here, since it's called in `VPG.__init__`.

garage/tf/algos/trpo.py
krzentner: This should not be called here, since it's called in `NPO.__init__`.

garage/tf/algos/tnpg.py
krzentner: This should not be called here, since it's called in `NPO.__init__`.

garage/tf/algos/ppo.py
krzentner: This should not be called here, since it's called in `NPO.__init__`.

garage/tf/algos/erwr.py
krzentner: This should not be called here, since it's called in `NPO.__init__`.

garage/experiment/local_tf_runner.py
naeioi: Are you suggesting `itr += n_epoch_cycles`? In fact, `itr` is defined as the number of times `algo.train_once()` is called. So `itr += 1` should be ok.

naeioi:
@krzentner The short story is, I did not like and not understand `Serializable`. To make the algorithms pickleable I wrote some helpers. But I later found that I was ending up just rewriting `Serializable`. So I fell back to use it.

To look closer at what is happening, there are three blocking fields. Because of them, none of the algorithms are automatically pickable. 
- `_name_scope`, created via `tf.name_scope(name)`
- `_policy_opt_inputs`, an instance of a wrapper class for tensorflow variables. Not picklable because `namedtuple` is used to create a single-use wrapper class. 
- Several `f_*` which are partially binded function created by `compile_function`. Not pickable because they are functions with closure (not top functions). 

Several possible solutions include,
- Get rid of those unpickleables (avoid using them in the algorithm)
Not practical because it requires redesign of the algorithm.
- Not directly saving them, by saving / loading them manually via `__getstate__` and `__setstate__`
Doable but not trivial. For each unpickable type, I need to come up with a way to store and construct it. And more effort might involve when the implementation of a new algorithm introduces some new weird thing. 
- Not saving them, but reconstructing the whole algorithm instance when restore. 
This requires saving only the arguments of algorithm constructor. And this is what `Serializable` does. Given that the stateful part of an algorithm (mostly policy and value function) is passed via the constructor parameter, I think this is the right way to go.

For the maintenance of `Serializable`, it appears to me to be simple, which implies reliable. After all,  the states of algorithms, which is the most valuable thing because we spend a lot of time training it, are already safely saved by `Parameterized` in the policy.

ryanjulian:
@naeioi  to be clear, a long-term goal in this codebase is to completely delete `Serializable` and `Parameterized`. Of course, this PR does not necessarily need to do that, but I think @krzentner is advocating not making that harder with design choices.

garage/tf/algos/vpg.py
ryanjulian: are we sure? i think have encountered behavior with `Serializable` requiring it be called in the constructor of every type you'd like to pickle (including child classes).

if a child class adds a constructor argument not present in the parent, i don't think Serializable will catch that in the parent. Original rllab code made a habit of peppering *args and **kwargs all over all constructors to catch this flat (I speculate). You can imagine that neither choice is a particularly good one.

ryanjulian:
My overall $0.02:

* It's unreasonable to embark on removing `Serializable` in this PR (out-of-scope)
* It's reasonable to avoid proliferating it (as @krzentner is pointing out)

KR and I discussed this and I don't think it would be terribly bad to implement your desired functionality by adding `__getstate__` and `__setstate__` to each of the concrete algorithm classes. 

Basic roadmap:
1. `del` unpicklable members in `__getstate__`
1. construct the optimizers in `__setstate__` (e.g. https://github.com/rlworkgroup/garage/blob/master/garage/tf/algos/npo.py#L71)
1. call `init_opt()` from both `__init__` and `__setstate__` (populating f_*, etc.)
1. (2) and (3) should repopulate the missing fields you mentioned

ryanjulian:
Please try this and if it turns out to be a total mess, we'll try a different way.

garage/tf/algos/vpg.py
krzentner: To be clear, it doesn't break anything to call `quick_init()` here. Instead, this call silently does nothing, since the *last* call (inside `NPO.__init__`) to `quick_init()` will override it. Read the source of `quick_init()` for why.

The correct thing to do is probably to call `quick_init()` at the *end* of these `__init__` functions, but that breaks the `quick_init()` at start of `__init__` rule, so it's also fragile.

garage/experiment/local_tf_runner.py
```diff
@@ -163,41 +169,132 @@ def obtain_samples(self, itr, batch_size):
...
+            from_epoch(str or int): The epoch to restore from.
+                Can be 'first', 'last' or a number.
+                Not applicable when snapshot_mode='last'.
+            resume_now(bool): Resume training immediately.
+                If False, runner will be set up but not train.
+                And the arguments for train() will be returned as a dict.
+                See examples below.
+
+        Returns:
+            None if resume_now=True.
```
ryanjulian: why not just return the arguments in both cases?

garage/experiment/local_tf_runner.py
```diff
@@ -163,41 +169,132 @@ def obtain_samples(self, itr, batch_size):
...
 
+    def restore(self, snapshot_dir, from_epoch='last', resume_now=True):
+        """Restore experiment from snapshot.
+
+        Args:
+            snapshot_dir(str): Directory of snapshot.
+            from_epoch(str or int): The epoch to restore from.
+                Can be 'first', 'last' or a number.
+                Not applicable when snapshot_mode='last'.
+            resume_now(bool): Resume training immediately.
```
ryanjulian: Hmm this API is kind of strange. `restore` should either always start training or never start training.

What if you made your runner stateful (i.e. it contains everything needed to run train)? Then `runner.restore()` always returns nothing, and you an add `runner.resume()` which uses the arguments acquired during `restore()`. `runner.resume()` can be based on `runner.train()`.

garage/experiment/local_tf_runner.py
```diff
@@ -163,41 +169,132 @@ def obtain_samples(self, itr, batch_size):
...
+            args = self.train_args.copy()
+            args.update(start_epoch=last_epoch + 1)
+            return args
+
     def log_diagnostics(self, pause_for_plot=False):
         """Log diagnostics.
 
         Args:
-            pause_for_plot: Pause for plot.
+            pause_for_plot(bool): Pause for plot.
```
ryanjulian: we don't actually use the `log_diagnostics` API anymore, so feel free to just log wherever you see fit.

garage/tf/algos/reps.py
```diff
@@ -399,19 +414,20 @@ def _dual_opt_input_values(self, samples_data):
...
+            samples_data['agent_infos'][k]
             for k in self.policy.state_info_keys
         ]   # yapf: disable
         policy_old_dist_info_list = [
-            samples_data["agent_infos"][k]
+            samples_data['agent_infos'][k]
             for k in self.policy.distribution.dist_info_keys
         ]
 
+        # pylint: disable=locally-disabled, unexpected-keyword-arg
```
ryanjulian: why the unexpected keyword arg?

garage/np/algos/cma_es.py
ryanjulian: it seems like you missed `__setstate__` and __getstate__` here

garage/np/algos/cem.py
ryanjulian: i think he just missed adding the pickling interface (and removing Serializable) on numpy-based algos

naeioi: Serializable is removed from np algos now.

garage/np/algos/cma_es.py
naeioi: It turns out that they don't know manually saving any state.

garage/np/algos/base.py
naeioi: I removed Serializable from all algorithms.

garage/tf/algos/reps.py
naeioi: pylint complaints that `obs_var` and `action_var` are invalid arguments. I think it is because `_policy_opt_inputs` are generated function and pylint just can't do a correct static analysis on them.

naeioi:
I will make the resume API today. sorry for delay.

garage/experiment/local_tf_runner.py
naeioi: Added now. PTAL.

garage/logger/snapshotter.py
```diff
@@ -78,3 +84,44 @@ def save_itr_params(self, itr, params):
...
+
+        Args:
+            itr(int or string): Iteration to load.
+                Can be an integer, 'last' or 'first'.
+
+        Returns:
+            dict: Loaded snapshot
+
+        """
+        if itr is None:
```
ryanjulian: why would someone pass `None` (when `'last'` is already default), and expect this to work?

garage/experiment/local_tf_runner.py
```diff
@@ -163,41 +170,117 @@ def obtain_samples(self, itr, batch_size):
...
+            2. Resume experiment immediately.
+            with LocalRunner() as runner:
+                runner.restore(snapshot_dir)
+                runner.resume()
+
+            3. Resume experiment with modified training arguments.
+             with LocalRunner() as runner:
+                args = runner.restore(snapshot_dir, resume_now=False)
+                args.n_epochs = 20
+                runner.resume()
```
ryanjulian: how did `runner.resume()` get your new `n_epochs` arg???

tests/garage/experiment/test_resume.py
```diff
@@ -0,0 +1,60 @@
...
+
+        policy_params = fixture_exp()
+        self.tearDown()
+        self.setUp()
+
+        with LocalRunner() as runner:
+            args = runner.restore(folder)
+            assert (runner.policy.get_param_values() == policy_params).all(), \
+                'Policy parameters should persist'
+            assert args.n_epochs == 5, \
```
ryanjulian: please only split with () per PEP8

tests/garage/experiment/test_resume.py
```diff
@@ -0,0 +1,60 @@
...
+                'Snapshot should save training parameters'
+            assert args.start_epoch == 5, \
+                'Last experiment should end at 5th iterations'
+
+            args.n_epochs = 10
```
ryanjulian: are you injecting args by modifying a object returned from `restore()`?

I think this is very surprising in Python.

Perhaps you can return a copy of `args` and then resume can accept an optional new `args` overriding the old one? 

e.g. `resume` just has kwargs which are the same as `train`, but they all default to the original args from the original call to `train`

garage/experiment/local_tf_runner.py
```diff
@@ -210,45 +293,95 @@ def log_diagnostics(self, pause_for_plot=False):
...
     def train(self,
               n_epochs,
+              batch_size,
               n_epoch_cycles=1,
-              batch_size=None,
               plot=False,
               store_paths=False,
-              pause_for_plot=False):
+              pause_for_plot=False,
+              **kwargs):
```
ryanjulian: why is passing `kwargs` here necessary?

garage/experiment/local_tf_runner.py
```diff
@@ -210,45 +293,95 @@ def log_diagnostics(self, pause_for_plot=False):
...
+        return self._train(
+            n_epochs=n_epochs,
+            n_epoch_cycles=n_epoch_cycles,
+            batch_size=batch_size,
+            plot=plot,
+            store_paths=store_paths,
+            pause_for_plot=pause_for_plot,
+            **kwargs)
+
+    def resume(self):
```
ryanjulian: if  you want to modify the training arguments in resume, i think resume should have the same function signature as train (but its default arguments are the ones from the first call to train)

garage/logger/snapshotter.py
naeioi: Fixed.

tests/garage/experiment/test_resume.py
naeioi: Fixed.

garage/experiment/local_tf_runner.py
naeioi: I removed it now.

garage/experiment/local_tf_runner.py
naeioi: What about now. It should look much better.

PR #640, opened 2019-04-26
Fix pickle tests

### LINKED ISSUE #608: Model class unit test flaw
The tests for pickling models, e.g. GaussianMLPModel, do not test the weights pickling correctly. If `pickle.dumps` is called in another session, rather than the session the model is built, it will not be able to get the weights.

TODO: Fix the test and state clearly in the documentation.

ahtsan:
Previously pickling test is done with default parameters. It work even if it does not pickle the weights correctly. This PR fixed the issue by manually assigning weights to some variables so as to make sure the pickling is done correctly. 
This PR will also close #608.

### DIFF
--- a/tests/garage/tf/models/test_model.py
+++ b/tests/garage/tf/models/test_model.py
@@ -164,7 +164,7 @@ class TestModel(TfGraphTestCase):
             # assign bias to all one
             with tf.variable_scope('SimpleModel/state', reuse=True):
                 bias = tf.get_variable('hidden_0/bias')
-            sess.run(tf.assign(bias, tf.ones_like(bias)))
+            bias.load(np.ones_like(bias))
 
             results = sess.run(
                 model.networks['default'].outputs,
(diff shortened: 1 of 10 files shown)


tests/garage/tf/models/test_model.py
```diff
@@ -191,6 +196,12 @@ def test_complicated_model_is_pickleable(self):
             input_var = tf.placeholder(tf.float32, shape=(None, 5))
             outputs = model.build(input_var)
 
+            # assign bias to all one
+            with tf.variable_scope(
+                    'ComplicatedModel/SimpleModel/state', reuse=True):
+                bias = tf.get_variable('hidden_0/bias')
+            sess.run(tf.assign(bias, tf.ones_like(bias)))
```
ryanjulian: i think you can use tf.Variable.load() directly instead of tf.assign()

See the pickler in Model

ahtsan: Would that be a good idea if I create another PR to replace all the `tf.assign()` to `tf.Variable.load()`?

Issue #623, opened 2019-04-16
Eliminate tf.assign operations in regressors


zhanpenghe:
The fit method of regressors sometimes creates new `tf.assign` operations. This will slow down the training in later iterations. These operation should be created only once and cached for future usage.

This PR(#622) fixes this for `GaussianMLPRegressor`.

ryanjulian:
Can we do it without assign? It's not really a recommended way to use TF.

PR #641, opened 2019-04-27
Fix GaussianMLPRegressor and baseline for pickling

### LINKED ISSUE #511: GaussianMLPRegressor is not unpickled appropriately
Sorry I can't come up with a good title for this but I will describe the problem in details here.

GaussianMLPRegressor has variables (e.g. [`y_mean`](https://github.com/rlworkgroup/garage/blob/f33a03ea4d71525eabadbe22f31ad0a91607b134/garage/tf/regressors/gaussian_mlp_regressor.py#L156)) that are created for normalizing the inputs and they are not saved during pickling. This will force user to run a variable initialization when they want to use an unpickled GaussianMLPBaseline. However, running initialization will re-initialize the model variables.

This will affect people who want to use a pickled baseline object to resume training or use them to do further learning.

ahtsan:
* New regressor base class which has the `get_params()` and `set_params()`. In the future, regressor should inherit from `Model` class and those methods should be put into `Model` class. For now I just put it in the regressor base class.
* Make `GaussianMLPRegressorWithModel` and `GaussianMLPBaselinesWithModel` pickleable, which will close #511.
* Modify `SimpleGaussianMLPRegressor` for testing.

### DIFF
--- a/garage/tf/regressors/base2.py
+++ b/garage/tf/regressors/base2.py
@@ -44,6 +44,9 @@ class Regressor2(abc.ABC):
 
         Args:
             xs: Input data.
+
+        Return:
+            The predicted ys.
         """
         raise NotImplementedError
 
@@ -158,9 +161,12 @@ class StochasticRegressor2(Regressor2):
         Symbolic graph of the log likelihood.
 
         Args:
-            x_var: Input tf.Tensor for the input data.
-            y_var: Input tf.Tensor for the label of data.
-            name: Name of the new graph.
+            x_var (tf.Tensor): Input tf.Tensor for the input data.
+            y_var (tf.Tensor): Input tf.Tensor for the label of data.
+            name (str): Name of the new graph.
+
+        Return:
+            tf.Tensor output of the symbolic log likelihood.
         """
         raise NotImplementedError
 
@@ -169,7 +175,10 @@ class StochasticRegressor2(Regressor2):
         Symbolic graph of the distribution.
 
         Args:
-            x_var: Input tf.Tensor for the input data.
-            name: Name of the new graph.
+            x_var (tf.Tensor): Input tf.Tensor for the input data.
+            name (str): Name of the new graph.
+
+        Return:
+            tf.Tensor output of the symbolic distribution.
         """
         raise NotImplementedError
--- a/garage/tf/regressors/base2.py
+++ b/garage/tf/regressors/base2.py
@@ -1,0 +1,175 @@
+"""Regressor base classes without Parameterized."""
+import abc
+
+import tensorflow as tf
+
+from garage.misc.tensor_utils import flatten_tensors
+from garage.misc.tensor_utils import unflatten_tensors
+
+
+class Regressor2(abc.ABC):
+    """
+    Regressor base class without Parameterized.
+
+    Args:
+        input_shape (tuple): Input shape.
+        output_dim (int): Output dimension.
+        name (str): Name of the regressor.
+    """
+
+    def __init__(self, input_shape, output_dim, name):
+        self._input_shape = input_shape
+        self._output_dim = output_dim
+        self._name = name
+        self._variable_scope = None
+        self._cached_params = {}
+        self._cached_param_dtypes = {}
+        self._cached_param_shapes = {}
+        self._cached_assign_ops = {}
+        self._cached_assign_placeholders = {}
+
+    def fit(self, xs, ys):
+        """
+        Fit with input data xs and label ys.
+
+        Args:
+            xs: Input data.
+            ys: Label of input data.
+        """
+        raise NotImplementedError
+
+    def predict(self, xs):
+        """
+        Predict ys based on input xs.
+
+        Args:
+            xs: Input data.
+        """
+        raise NotImplementedError
+
+    def get_params_internal(self, **tags):
+        """
+        Get the list of parameters.
+
+        This internal method does not perform caching, and should
+        be implemented by subclasses.
+        """
+        raise NotImplementedError
+
+    def get_params(self, **tags):
+        """
+        Get the list of parameters, filtered by the provided tags.
+
+        Some common tags include 'regularizable' and 'trainable'
+        """
+        tag_tuple = tuple(sorted(list(tags.items()), key=lambda x: x[0]))
+        if tag_tuple not in self._cached_params:
+            self._cached_params[tag_tuple] = self.get_params_internal(**tags)
+        return self._cached_params[tag_tuple]
+
+    def get_param_dtypes(self, **tags):
+        """Get the list of dtypes for the parameters."""
+        tag_tuple = tuple(sorted(list(tags.items()), key=lambda x: x[0]))
+        if tag_tuple not in self._cached_param_dtypes:
+            params = self.get_params(**tags)
+            param_values = tf.get_default_session().run(params)
+            self._cached_param_dtypes[tag_tuple] = [
+                val.dtype for val in param_values
+            ]
+        return self._cached_param_dtypes[tag_tuple]
+
+    def get_param_shapes(self, **tags):
+        """Get the list of shapes for the parameters."""
+        tag_tuple = tuple(sorted(list(tags.items()), key=lambda x: x[0]))
+        if tag_tuple not in self._cached_param_shapes:
+            params = self.get_params(**tags)
+            param_values = tf.get_default_session().run(params)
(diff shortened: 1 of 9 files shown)


tests/fixtures/models/simple_gaussian_mlp_model.py
```diff
@@ -7,7 +7,11 @@
 class SimpleGaussianMLPModel(Model):
     """Simple GaussianMLPModel for testing."""
 
-    def __init__(self, name, output_dim, *args, **kwargs):
+    def __init__(self,
+                 output_dim,
+                 *args,
+                 name='SimpleGaussianMLPModel',
```
zhanpenghe: Sort arguments?

garage/tf/regressors/base2.py
```diff
@@ -0,0 +1,175 @@
...
+    def predict(self, xs):
+        """
+        Predict ys based on input xs.
+
+        Args:
+            xs: Input data.
+        """
+        raise NotImplementedError
+
+    def get_params_internal(self, **tags):
```
zhanpenghe: Why are we still keeping `get_params_internal`? I thought they are used mainly for `Parameterized`

ahtsan: The optimizers require those API from `Parameterized`, like `get_params`, `get_param_values` and `set_param_values`. So I keep them in order to make it compatible with the optimizers.

zhanpenghe: I see. But since these are not used with the optimizers yet, can we just remove them as other primitive? (Or if other primitives have them, I am ok as long as they are consistent)

tests/fixtures/models/simple_gaussian_mlp_model.py
ahtsan: Will fix it

ahtsan:
* Also replace `tf.assign` by `tf.Variable.load`. This will close #623.

garage/tf/regressors/base2.py
ahtsan: They do. `GaussianMLPBaseline` uses `GaussianMLPRegressor` and optimizers require `GaussianMLPBaseline` to have these API, e.g. in PPO. We already have one test running with `PPO` and `GaussianMLPBaselineWithModel`.

I think other primitives also need them, if they will be used by optimizers.

garage/tf/regressors/base2.py
```diff
@@ -0,0 +1,175 @@
...
+        self._output_dim = output_dim
+        self._name = name
+        self._variable_scope = None
+        self._cached_params = {}
+        self._cached_param_dtypes = {}
+        self._cached_param_shapes = {}
+        self._cached_assign_ops = {}
+        self._cached_assign_placeholders = {}
+
+    def fit(self, xs, ys):
```
ryanjulian: does this have a return?

garage/tf/regressors/base2.py
```diff
@@ -0,0 +1,175 @@
...
+        """
+        Fit with input data xs and label ys.
+
+        Args:
```
ryanjulian: i think this is missing docstring for `Returns`

garage/tf/regressors/base2.py
```diff
@@ -0,0 +1,175 @@
...
+    Args:
+        input_shape (tuple): Input shape.
+        output_dim (int): Output dimension.
+        name (str): Name of the regressor.
+    """
+
+    def __init__(self, input_shape, output_dim, name):
+        super().__init__(input_shape, output_dim, name)
+
+    def log_likelihood_sym(self, x_var, y_var, name=None):
```
ryanjulian: what is the docstring for Returns?

garage/tf/regressors/base2.py
```diff
@@ -0,0 +1,175 @@
...
+        Symbolic graph of the log likelihood.
+
+        Args:
+            x_var: Input tf.Tensor for the input data.
+            y_var: Input tf.Tensor for the label of data.
+            name: Name of the new graph.
+        """
+        raise NotImplementedError
+
+    def dist_info_sym(self, x_var, name=None):
```
ryanjulian: `Returns`?

tests/fixtures/regressors/simple_gaussian_mlp_regressor.py
```diff
@@ -1,30 +1,43 @@
...
-
-    def set_param_values(self, flattened_params, *args, **kwargs):
-        tf.get_default_session().run(
-            tf.assign(self.param_values, flattened_params))
+        if self.ys is None:
+            mean = tf.get_default_session().run(
+                self.model.networks['default'].mean,
+                feed_dict={self.model.networks['default'].input: xs})
+            return np.full((len(xs), 1), mean)
+        else:
```
ryanjulian: you can remove this else

garage/tf/regressors/base2.py
ahtsan: It doesn't return anything. Should we state that explicitly?

tests/fixtures/regressors/simple_gaussian_mlp_regressor.py
ahtsan: The else branch is for testing `self.fit`. If `self.fit(xs, ys)` is called, the prediction should be same as ys. Otherwise it would be the default value of `mean`.

garage/tf/regressors/base2.py
```diff
@@ -0,0 +1,175 @@
...
+            ys: Label of input data.
+        """
+        raise NotImplementedError
+
+    def predict(self, xs):
+        """
+        Predict ys based on input xs.
+
+        Args:
+            xs: Input data.
```
ryanjulian: Return value?

garage/tf/regressors/base2.py
ryanjulian: not necessary

tests/fixtures/regressors/simple_gaussian_mlp_regressor.py
ryanjulian: what i meant was a pattern like this

```python
if special_case:
   # do stuff for special case
   return stuff

# do stuff for non-special case
return other_stuff
```

This is called a guard clause and it helps make code easier to read and refactor

https://medium.com/@scadge/if-statements-design-guard-clauses-might-be-all-you-need-67219a1a981a

ahtsan: got it

garage/tf/regressors/base2.py
```diff
@@ -0,0 +1,184 @@
...
+                                             self.get_param_shapes(**tags))
+            ops = []
+            feed_dict = dict()
+            for param, dtype, value in zip(
+                    self.get_params(**tags), self.get_param_dtypes(**tags),
+                    param_values):
+                if param not in self._cached_assign_ops:
+                    assign_placeholder = tf.placeholder(
+                        dtype=param.dtype.base_dtype)
+                    assign_op = tf.assign(param, assign_placeholder)
```
zhanpenghe: Can we also replace these with variable.load?

ahtsan: We can wrap a variable.load inside a session.make_callable, will that be better than tf.assign?

ahtsan: Yes I think so.

garage/tf/regressors/base2.py
ryanjulian: types, please

ahtsan: Usually the input and output for this kind of general API would be numpy.ndarray?

ryanjulian: i think so. if there's more information on the appropriate shape, please include it in the docstring.

ahtsan: Fixed. If everything looks good, I will squash and merge.

ryanjulian: lgtm

tests/fixtures/models/simple_gaussian_mlp_model.py
ryanjulian: what about this one?

ryanjulian:
this lgtm

tests/fixtures/models/simple_gaussian_mlp_model.py
ahtsan: Fixed. Not sure why it is not shown as Outdated.

PR #642, opened 2019-04-29
LSTM without layers

ahtsan:
* Add `lstm`, `lstm_model`, `categorical_lstm_policy_with_model`.
* Unit tests for the above.
* Test TRPO with `categorical_lstm_policy_with_model`

### DIFF
--- a/examples/tf/trpo_cartpole_recurrent_with_model.py
+++ b/examples/tf/trpo_cartpole_recurrent_with_model.py
@@ -1,0 +1,51 @@
+#!/usr/bin/env python3
+"""
+This is an example to train a task with TRPO algorithm.
+
+It uses an LSTM-based recurrent policy.
+
+Here it runs CartPole-v1 environment with 100 iterations.
+
+Results:
+    AverageReturn: 100
+    RiseTime: itr 13
+"""
+from garage.experiment import LocalRunner, run_experiment
+from garage.np.baselines import LinearFeatureBaseline
+from garage.tf.algos import TRPO
+from garage.tf.envs import TfEnv
+from garage.tf.optimizers import ConjugateGradientOptimizer
+from garage.tf.optimizers import FiniteDifferenceHvp
+from garage.tf.policies import CategoricalLSTMPolicyWithModel
+
+
+def run_task(*_):
+    """Run task."""
+    with LocalRunner() as runner:
+        env = TfEnv(env_name='CartPole-v1')
+
+        policy = CategoricalLSTMPolicyWithModel(
+            name='policy', env_spec=env.spec)
+
+        baseline = LinearFeatureBaseline(env_spec=env.spec)
+
+        algo = TRPO(
+            env_spec=env.spec,
+            policy=policy,
+            baseline=baseline,
+            max_path_length=100,
+            discount=0.99,
+            max_kl_step=0.01,
+            optimizer=ConjugateGradientOptimizer,
+            optimizer_args=dict(
+                hvp_approach=FiniteDifferenceHvp(base_eps=1e-5)))
+
+        runner.setup(algo, env)
+        runner.train(n_epochs=100, batch_size=4000)
+
+
+run_experiment(
+    run_task,
+    snapshot_mode='last',
+    seed=1,
+)
--- a/garage/tf/algos/npo.py
+++ b/garage/tf/algos/npo.py
@@ -234,7 +234,6 @@ class NPO(BatchPolopt):
 
     def _build_policy_loss(self, i):
         pol_dist = self.policy.distribution
-        policy_entropy = self._build_entropy_term(i)
         rewards = i.reward_var
 
         with tf.name_scope('policy_loss'):
@@ -282,6 +281,10 @@ class NPO(BatchPolopt):
                     policy_dist_info_flat,
                     i.flat.valid_var,
                     name='policy_dist_info_valid')
+
+                policy_dist_info = policy_dist_info_valid
+
+            policy_entropy = self._build_entropy_term(i, policy_dist_info)
 
             # Calculate loss function and KL divergence
             with tf.name_scope('kl'):
@@ -384,14 +387,9 @@ class NPO(BatchPolopt):
 
             return loss, pol_mean_kl
 
-    def _build_entropy_term(self, i):
+    def _build_entropy_term(self, i, policy_dist_info):
         with tf.name_scope('policy_entropy'):
             if self.policy.recurrent:
-                policy_dist_info = self.policy.dist_info_sym(
-                    i.obs_var,
-                    i.policy_state_info_vars,
-                    name='policy_dist_info_entropy')
-
                 policy_neg_log_likeli = self.policy.distribution.log_likelihood_sym(  # noqa: E501
                     i.action_var,
                     policy_dist_info,
@@ -404,26 +402,16 @@ class NPO(BatchPolopt):
                         policy_dist_info)
 
             else:
-                policy_dist_info_flat = self.policy.dist_info_sym(
-                    i.flat.obs_var,
-                    i.flat.policy_state_info_vars,
-                    name='policy_dist_info_flat_entropy')
-
-                policy_dist_info_valid = filter_valids_dict(
-                    policy_dist_info_flat,
-                    i.flat.valid_var,
-                    name='policy_dist_info_valid')
-
                 policy_neg_log_likeli_valid = self.policy.distribution.log_likelihood_sym(  # noqa: E501
                     i.valid.action_var,
-                    policy_dist_info_valid,
+                    policy_dist_info,
                     name='policy_log_likeli')
 
                 if self._use_neg_logli_entropy:
                     policy_entropy = policy_neg_log_likeli_valid
                 else:
                     policy_entropy = self.policy.distribution.entropy_sym(
-                        policy_dist_info_valid)
+                        policy_dist_info)
 
             # This prevents entropy from becoming negative for small policy std
             if self._use_softplus_entropy:
--- a/garage/tf/algos/npo.py
+++ b/garage/tf/algos/npo.py
@@ -390,7 +390,7 @@ class NPO(BatchPolopt):
                 policy_dist_info = self.policy.dist_info_sym(
                     i.obs_var,
                     i.policy_state_info_vars,
-                    name='policy_dist_info')
+                    name='policy_dist_info_entropy')
 
                 policy_neg_log_likeli = self.policy.distribution.log_likelihood_sym(  # noqa: E501
                     i.action_var,
--- a/garage/tf/algos/npo.py
+++ b/garage/tf/algos/npo.py
@@ -234,6 +234,7 @@ class NPO(BatchPolopt):
 
     def _build_policy_loss(self, i):
         pol_dist = self.policy.distribution
+        policy_entropy = self._build_entropy_term(i)
         rewards = i.reward_var
 
         with tf.name_scope('policy_loss'):
@@ -283,8 +284,6 @@ class NPO(BatchPolopt):
                     name='policy_dist_info_valid')
 
                 policy_dist_info = policy_dist_info_valid
-
-            policy_entropy = self._build_entropy_term(i, policy_dist_info)
 
             # Calculate loss function and KL divergence
             with tf.name_scope('kl'):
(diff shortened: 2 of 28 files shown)


ryanjulian:
It seems you may have forgotten to cover the `state_include_action` case for LSTM:

https://codecov.io/gh/rlworkgroup/garage/commit/07d0d3155ce9961affb152589e9dc3545f487831#D12-207

garage/tf/core/layers.py
```diff
@@ -49,12 +49,12 @@ def as_tuple(x, n, t=None):
         x = (x, ) * n
 
     if (t is not None) and not all(isinstance(v, t) for v in x):
-        raise TypeError("expected a single value or an iterable "
-                        "of {0}, got {1} instead".format(t.__name__, x))
+        raise TypeError('expected a single value or an iterable '
```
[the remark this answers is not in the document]
ahtsan: It's done by the pre-push checks.

garage/tf/algos/npo.py
```diff
@@ -390,7 +390,7 @@ def _build_entropy_term(self, i):
                 policy_dist_info = self.policy.dist_info_sym(
                     i.obs_var,
                     i.policy_state_info_vars,
-                    name='policy_dist_info')
+                    name='policy_dist_info_entropy')
```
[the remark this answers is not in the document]
ahtsan: I renamed it because we can't call `dist_info_sym` with the same name twice (because the underlying model will build twice with the same name, which is not allowed). I think Chang is going to fix it by changing the `name_scope` to `variable_scope`. @CatherineSue

garage/tf/policies/base2.py
```diff
@@ -1,6 +1,8 @@
 """Policy base classes without Parameterized."""
 import tensorflow as tf
 
+from garage.misc.tensor_utils import flatten_tensors, unflatten_tensors
```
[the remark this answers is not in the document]
ahtsan: After discussing with Zhanpeng and ran some test, I think `tf.VariableScope` is working properly.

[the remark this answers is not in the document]
ahtsan: I will fix it in another PR since modifying `Policy2` will affect other policies as well.

garage/tf/algos/npo.py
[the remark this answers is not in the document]
CatherineSue: Changing it from `name_scope` to `variable_scope` won't fix the error. The cause is in Model instead of here.

garage/tf/policies/categorical_lstm_policy_with_model.py
```diff
@@ -0,0 +1,255 @@
...
+            prev_action_var = tf.cast(prev_action_var, tf.float32)
+            all_input_var = tf.concat(
+                axis=2, values=[obs_var, prev_action_var])
+        else:
+            all_input_var = obs_var
+
+        with tf.variable_scope(self._variable_scope):
+            outputs, _, _, _, _, _ = self.model.build(
+                all_input_var,
+                *self.model.networks['default'].inputs[1:4],
```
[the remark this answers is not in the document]
ryanjulian: yeah this is pretty crazy. what even is `inputs[1:4]`?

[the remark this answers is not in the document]
ahtsan: It is crazy. I will fix it.

examples/tf/trpo_cartpole_recurrent_with_model.py
```diff
@@ -0,0 +1,51 @@
...
+            baseline=baseline,
+            max_path_length=100,
+            discount=0.99,
+            max_kl_step=0.01,
+            optimizer=ConjugateGradientOptimizer,
+            optimizer_args=dict(
+                hvp_approach=FiniteDifferenceHvp(base_eps=1e-5)))
+
+        runner.setup(algo, env)
+        runner.train(n_epochs=100, batch_size=4000)
```
ryanjulian: this is a lot of iters. how long does this test take?

garage/tf/algos/npo.py
```diff
@@ -384,14 +387,9 @@ def _build_policy_loss(self, i):
 
             return loss, pol_mean_kl
 
-    def _build_entropy_term(self, i):
+    def _build_entropy_term(self, i, policy_dist_info):
```
ryanjulian: please don't make these functions dependent on anything but `i`

garage/tf/models/base.py
```diff
@@ -188,8 +188,7 @@ def build(self, *inputs, name=None):
         if not self._networks:
             # First time building the model, so self._networks are empty
             # We store the variable_scope to reenter later when we reuse it
-            with tf.variable_scope(
-                    self._name, reuse=False) as self._variable_scope:
+            with tf.variable_scope(self._name) as self._variable_scope:
```
ryanjulian: please explicitly assign the new a scope into a member variable -- do not capture it as a coincidence of a context manager

garage/tf/models/base.py
```diff
@@ -188,8 +188,7 @@ def build(self, *inputs, name=None):
         if not self._networks:
```
ryanjulian: don't forget to update the types in the docstring

garage/tf/models/base.py
```diff
@@ -210,20 +209,29 @@ def build(self, *inputs, name=None):
                     network = Network()
                     network._inputs = inputs
                     network._outputs = self._build(*inputs, name)
-        spec = self.network_output_spec()
-        if spec:
-            c = namedtuple(network_name,
-                           [*spec, 'input', 'output', 'inputs', 'outputs'])
+        in_spec = self.network_input_spec()
```
ryanjulian: this logic about `spec` is getting kind of complicated and i'm not sure i understand what's going on.

can you explain what the purpose of this code is? perhaps it would be good to break it into a well-documented helper function.

can you also help me understand the purpose of the input and output specs?

tests/helpers.py
```diff
@@ -87,6 +87,77 @@ def convolve(_input, filter_weights, filter_bias, strides, filter_sizes,
...
+    w_ho = np.full((num_units, num_units), w_h_init)
+    b_o = np.full((num_units, ), b_init)
+
+    w_x_ifco = np.concatenate([w_xi, w_xf, w_xc, w_xo], axis=1)
+    w_h_ifco = np.concatenate([w_hi, w_hf, w_hc, w_ho], axis=1)
+
+    x_ifco = np.matmul(input_val, w_x_ifco)
+    h_ifco = np.matmul(step_hidden, w_h_ifco)
+
+    x_i, x_f, x_c, x_o = np.split(x_ifco, 4, axis=1)  # noqa: E501, pylint: disable=unbalanced-tuple-unpacking
```
ryanjulian: what is the justifcation for disabling E501 and the pylint warning?

garage/tf/algos/npo.py
ryanjulian: it is very easy to make a loss function which is a tangle of calls to other functions. we instead make sure all components of the loss function are only dependent on `i` (the input spec, basically) which makes it easy to follow.

this will be much easier once the loss function is a model.

ahtsan: I see. I modified that since `_build_entropy_term` actually reuses the same `policy_dist_info` symbolic tensor with `_build_policy_loss`. I will come up with another way.

garage/tf/models/base.py
ahtsan: Forgot to modify it here. Will fix.

tests/helpers.py
ahtsan: Since I can't find out what's the reason pylint is complaining about that error, so I disable the pylint checking and since it is too long, I disabled E501 as well. I will dig into it and see what pylint is talking about.

garage/tf/models/base.py
ahtsan: Basically the `specs` tells us what the inputs and outputs are. There are 4 cases:
### Case 1: Single input/output model
Then we don't need any specs, since we don't need extra attribute for the networks. We only need `input,inputs,output,outputs` which are the default ones in network.
### Case 2: Single input/Multiple output model
We set `network_output_spec()` so that we can have assign "names" to the outputs, e.g. 
```python
def network_output_spec(self):
    return ['state', 'action']
```
Then we can do `model.networks['default'].state` and `model.networks['default'].action`
In this case, we don't need `network_input_spec()` since there is only one input. What we need is only `input`, which are already included by default.
### Case 3: Multiple input/Single output model
Same as above, but instead of having `network_output_spec`, we have `network_input_spec` this time.
### Case 4: Multiple input/output model
We have both `network_output_spec` and `network_input_spec`. Then we can have sth like
`model.networks['default'].state` and `model.networks['default'].action` as the outputs and `model.networks['default'].state_input` and `model.networks['default'].action_input` as the inputs, to make it more readable.

tests/helpers.py
ryanjulian: perhaps is one of your return values itself a tuple or list?

garage/tf/models/base.py
ryanjulian: this seems pretty complex -- is there a way we can always have an input/output spec, even for multi -input and -output models?

i think *always* having a spec would simplify the logic significantly.

ahtsan: Do you mean we make `specs` not optional?
Right now by default the specs are a empty list, which means networks will only have `input,inputs,output,outputs`.

examples/tf/trpo_cartpole_recurrent_with_model.py
ahtsan: Settings are the same as the existing `trpo_cartpole_recurrent`.`trpo_cartpole_recurrent` takes  ~400s while `trpo_cartpole_recurrent_with_model` takes ~460s.

garage/tf/models/base.py
ryanjulian: i mean that you internally maintain the `specs` data structure regardless of whether the user has chosen to do multi input/output or not. then the single-input and single-output cases don't need to be handled with so much special logic.

the outer interface can stay the same.

examples/tf/trpo_cartpole_recurrent_with_model.py
ryanjulian: oops i thought this was a test. it is concerning that the model version takes 10% longer. i wonder why, but we don't need to worry much about it now.

garage/tf/algos/npo.py
ryanjulian: you can just not reuse it...

ahtsan: We have to reuse it if we don't do dist info sym with a different name. Because each network name is unique. That's why I add "_entropy" after that.

ahtsan: We can just append a "_2" to the name in the second call of dist info sym.

ryanjulian: sure

tests/helpers.py
ahtsan: no... They are all `numpy.ndarray`.

ahtsan: Could be because interpreter cannot ensure the right hand side can be split into 4. Is there any other way to pass pylint for this?

ryanjulian: generally pylint rules try to encourage being very explicit and writing code so that it can be statically analyzed for errors (e.g. using pylint).

in this case i think pylint has good point, because the `np.split` operation is totally static (all inputs are known at compile time), but it has been written as a dynamic operation unnecessarily.

the rationale for this is that code is read much more often than it is written, so it's usually better to make it longer and easier to read rather than shorted and easier to write.

maybe it would be clearer if you just wrote out. this way a reader doesn't have to look up documentation for np.split to find out how your array is unpacked.

```python
x_i, x_f, x_c, x_o = x_ifco[:, :4], x_ifco[:, 4:8], x_ifco[:, 8:12], x_ifco[:, 12:16]
# or
x_i = x_ifco[:, :4]
x_f = x_ifco[:, 4:8]
x_c = x_ifco[:, 8:12]
x_o = x_ifco[:, 12:16]
```

ahtsan: Got it.

tests/garage/tf/core/test_lstm.py
```diff
@@ -0,0 +1,105 @@
+import numpy as np
+import tensorflow as tf
+
+from garage.tf.core.lstm import lstm
+from tests.fixtures import TfGraphTestCase
+
+
+class TestLSTM(TfGraphTestCase):
```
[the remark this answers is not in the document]
ryanjulian: i think overall this testing strategy is okay, but please add some simple checks too, e.g.

* output length vs input length
* gradient path (or lack thereof) between pairs of inputs/outputs
* non-zero hidden states
* etc.

perhaps you can take a look as the tests from lasagne for some ideas: https://github.com/Lasagne/Lasagne/blob/master/lasagne/tests/layers/test_recurrent.py

garage/tf/models/base.py
ryanjulian: bump

PR #644, opened 2019-04-29
Fix variable scope reentering

### LINKED ISSUE #609: Use tf.variable_scope instead of tf.VariableScope for policies using models
Using `tf.VariableScope` will not add a new scope to the graph. This introduces bugs to `model.parameters`.

ahtsan:
Previously a `tf.VariableScope` object is created and cached. However, it will not create a `tf.variable_scope` object and add the scope to the graph.
This PR fixes variable scope reentering by following the TensorFlow suggested way.
(Reference: https://github.com/tensorflow/tensorflow/blob/r1.13/tensorflow/python/ops/variable_scope.py#L1985)

This will close #609.

### DIFF
--- a/garage/tf/policies/categorical_conv_policy_with_model.py
+++ b/garage/tf/policies/categorical_conv_policy_with_model.py
@@ -107,8 +107,8 @@ class CategoricalConvPolicyWithModel(StochasticPolicy2):
     def _initialize(self):
         state_input = tf.placeholder(tf.float32, shape=(None, ) + self.obs_dim)
 
-        with tf.variable_scope(
-                name=self.name, reuse=False) as self._variable_scope:
+        with tf.variable_scope(self.name) as vs:
+            self._variable_scope = vs
             self.model.build(state_input)
 
         self._f_prob = tf.get_default_session().make_callable(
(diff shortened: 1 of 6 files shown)


garage/tf/policies/categorical_conv_policy_with_model.py
```diff
@@ -107,7 +107,8 @@ def __init__(self,
     def _initialize(self):
         state_input = tf.placeholder(tf.float32, shape=(None, ) + self.obs_dim)
 
-        with tf.variable_scope(self._variable_scope):
+        with tf.variable_scope(
```
ryanjulian: IMO this is a really confusing way to assign `self._variable_scope`

why not just do it in the constructor?

```python
self._variable_scope = tf.variable_scope(...)
```

ahtsan: After some testing, I think `tf.VariableScope` is working fine. The pros of using `tf.VariableScope` is that it's pickleable and can be used for reentering.

ahtsan:
After some testing, I think tf.VariableScope is working fine. The pros of using `tf.VariableScope` is that it's pickleable and can be used for reentering. The bug Zhanpeng is caused by passing wrong constructor argument to `tf.VariableScope`, which is fixed. 
If it sounds good, I will close this PR.

garage/tf/policies/categorical_conv_policy_with_model.py
ahtsan: How about just store the name and everytime we do
```
with tf.variable_scope(self.name, reuse=True):
    ....
```
?

Issue #646, opened 2019-05-08
Move process_samples() from sampler to algorithm


naeioi:
`process_samples()` takes raw batch data and generates fields that are useful in algorithm training. Currently, it resides in sampler and runner uses it to process batch data before piping to the algorithm. This is problematic for several reasons, 
- An algorithm may need a different preprocessing procedure but has not control of it.
- For off-policy sampler, the last path in one batch might not be complete. This will make the logging of return not accurate. (Only logging of **complete** path is accurate because it reflects the performance of an algorithm). 
Currently, the `process_samples()` of off-policy sampler wipes out the `done` from raw data. If we make the algorithm accepts raw data and do preprocessing itself, it can test if a path is complete and avoid logging it if not.

We can fix these issues by moving process_samples() from sampler to algorithm. Right now all on-policy algorithms share the same preprocessing procedure. We can make it at the base class, and only override it in a new algorithm implementation when necessary.

zequnyu:
@naeioi One quick question: do all the tf/np algorithms need process_samples()? Then I suppose I should put the method to `garage.np.algos.RLAlgorithm`. Or just tf algorithms because I saw no samplers in the np folder.

naeioi:
> do all the tf/np algorithms need process_samples()?

Yes. 

The fact is that currently, np algorithms use sampler under `garage.tf.samplers`. Check out https://github.com/rlworkgroup/garage/blob/master/examples/np/cem_cartpole.py

This is because even though np algorithms do not use a tensorflow optimizer, they still need a policy, and we have only polices implemented in tensorflow. 

> I should put the method to garage.np.algos.RLAlgorithm

I think an abstract `process_samples()` should be defined in `garage.np.algos.RLAlgorithm`. And `process_samples()` in `garage.tf.samplers.BatchSampler` should be moved to `BatchPolopt` in `gararge.np/tf.algos`, also `process_samples()` in `garage.tf.samplers.OffPolicyVectorizedSampler` should be moved to `garage.tf.algos.OffPolicyRLAlgorithm`.

PR #652, opened 2019-05-11
Move process_samples into algorithms

zequnyu:
This PR will do the refactoring of #646 as well as addressing #671.

### DIFF
--- a/garage/experiment/local_tf_runner.py
+++ b/garage/experiment/local_tf_runner.py
@@ -238,6 +238,7 @@ class LocalRunner:
 
             When resume programmatically, snapshot directory should be
             specify manually or through run_experiment() interface.
+
         """
         snapshotter.snapshot_dir = snapshot_dir
         saved = snapshotter.load(from_epoch)
--- a/garage/experiment/local_tf_runner.py
+++ b/garage/experiment/local_tf_runner.py
@@ -399,7 +399,6 @@ class LocalRunner:
             with logger.prefix('epoch #%d | ' % epoch):
                 for cycle in range(n_epoch_cycles):
                     paths = self.obtain_samples(itr, batch_size)
-                    paths = self.sampler.process_samples(itr, paths)
                     last_return = self.algo.train_once(itr, paths)
                     itr += 1
                 self.save(epoch, paths if store_paths else None)
--- a/garage/np/algos/base.py
+++ b/garage/np/algos/base.py
@@ -2,8 +2,7 @@ import abc
 
 
 class RLAlgorithm(abc.ABC):
-    """Base class for all the algorithms.
-    """
+    """Base class for all the algorithms."""
 
     @abc.abstractmethod
     def train_once(self, itr, paths):
@@ -13,5 +12,6 @@ class RLAlgorithm(abc.ABC):
         Args:
             itr (int): Iteration number.
             paths (list[dict]): A list of collected paths.
+
         """
         pass
--- a/garage/np/algos/base.py
+++ b/garage/np/algos/base.py
@@ -1,7 +1,17 @@
-class Algorithm:
-    pass
+import abc
 
 
-class RLAlgorithm(Algorithm):
-    def train(self):
-        raise NotImplementedError
+class RLAlgorithm(abc.ABC):
+    """Base class for all the algorithms.
+    """
+
+    @abc.abstractmethod
+    def train_once(self, itr, paths):
+        """Performs one step of policy optimization
+        given one batch of samples.
+
+        Args:
+            itr (int): Iteration number.
+            paths (list[dict]): A list of collected paths.
+        """
+        pass
--- a/garage/np/algos/batch_polopt.py
+++ b/garage/np/algos/batch_polopt.py
@@ -1,135 +1,99 @@
-from dowel import logger, tabular
+import collections
 
-from garage.experiment import snapshotter
+from dowel import tabular
+import numpy as np
+
+from garage.misc import special, tensor_utils
 from garage.np.algos.base import RLAlgorithm
-from garage.plotter import Plotter
-from garage.sampler import BatchSampler
 
 
 class BatchPolopt(RLAlgorithm):
-    """
-    Base class for batch sampling-based policy optimization methods.
-    This includes various policy gradient methods like vpg, npg, ppo, trpo,
-    etc.
+    """A batch-based algorithm interleaves sampling and policy optimization.
+    In one round of training, the runner will first instruct the sampler to do
+    environment rollout and the sampler will collect a given number of samples
+    (in terms of environment interactions). The collected paths are then
+    absorbed by `RLAlgorithm.train_once()` and an algorithm performs one step
+    of policy optimization.
+    The updated policy will then be used in the next round of sampling.
+
+    Args:
+        policy (garage.tf.policies.base.Policy): Policy.
+        baseline (garage.tf.baselines.Baseline): The baseline.
+        discount (float): Discount.
+        max_path_length (int): Maximum length of a single rollout.
     """
 
-    def __init__(self,
-                 env,
-                 policy,
-                 baseline,
-                 scope=None,
-                 n_itr=500,
-                 start_itr=0,
-                 batch_size=5000,
-                 max_path_length=500,
-                 discount=0.99,
-                 gae_lambda=1,
-                 plot=False,
-                 pause_for_plot=False,
-                 center_adv=True,
-                 positive_adv=False,
-                 store_paths=False,
-                 sampler_cls=None,
-                 sampler_args=None,
-                 **kwargs):
-        """
-        :param env: Environment
-        :param policy: Policy
-        :type policy: Policy
-        :param baseline: Baseline
-        :param scope: Scope for identifying the algorithm. Must be specified if
-         running multiple algorithms
-        simultaneously, each using different environments and policies
-        :param n_itr: Number of iterations.
-        :param start_itr: Starting iteration.
-        :param batch_size: Number of samples per iteration.
-        :param max_path_length: Maximum length of a single rollout.
-        :param discount: Discount.
-        :param gae_lambda: Lambda used for generalized advantage estimation.
-        :param plot: Plot evaluation run after each iteration.
-        :param pause_for_plot: Whether to pause before contiuing when plotting.
-        :param center_adv: Whether to rescale the advantages so that they have
-         mean 0 and standard deviation 1.
-        :param positive_adv: Whether to shift the advantages so that they are
-         always positive. When used in
-        conjunction with center_adv the advantages will be standardized before
-         shifting.
-        :param store_paths: Whether to save all paths data to the snapshot.
-        """
-        self.env = env
+    def __init__(self, policy, baseline, discount, max_path_length):
         self.policy = policy
         self.baseline = baseline
-        self.scope = scope
-        self.n_itr = n_itr
-        self.current_itr = start_itr
-        self.batch_size = batch_size
-        self.max_path_length = max_path_length
         self.discount = discount
-        self.gae_lambda = gae_lambda
-        self.plot = plot
-        self.pause_for_plot = pause_for_plot
-        self.center_adv = center_adv
-        self.positive_adv = positive_adv
-        self.store_paths = store_paths
-        if sampler_cls is None:
-            sampler_cls = BatchSampler
-        if sampler_args is None:
-            sampler_args = dict()
-        self.sampler = sampler_cls(self, **sampler_args)
+        self.max_path_length = max_path_length
 
-    def start_worker(self):
-        self.sampler.start_worker()
+        self.episode_reward_mean = collections.deque(maxlen=100)
 
-    def shutdown_worker(self):
-        self.sampler.shutdown_worker()
+    def process_samples(self, itr, paths):
+        """Return processed sample data based on the collected paths.
 
-    def train(self):
-        plotter = Plotter()
-        if self.plot:
-            plotter.init_plot(self.env, self.policy)
-        self.start_worker()
-        self.init_opt()
-        for itr in range(self.current_itr, self.n_itr):
-            with logger.prefix('itr #{} | '.format(itr)):
-                paths = self.sampler.obtain_samples(itr)
-                samples_data = self.sampler.process_samples(itr, paths)
-                self.log_diagnostics(paths)
-                self.optimize_policy(itr, samples_data)
-                logger.log('Saving snapshot...')
-                params = self.get_itr_snapshot(itr, samples_data)
-                self.current_itr = itr + 1
-                params['algo'] = self
-                if self.store_paths:
-                    params['paths'] = samples_data['paths']
-                snapshotter.save_itr_params(itr, params)
-                logger.log('saved')
-                logger.log(tabular)
-                if self.plot:
-                    plotter.update_plot(self.policy, self.max_path_length)
-                    if self.pause_for_plot:
-                        input('Plotting evaluation run: Press Enter to '
-                              'continue...')
+        Args:
+            itr (int): Iteration number.
+            paths (list[dict]): A list of collected paths
 
-        plotter.close()
-        self.shutdown_worker()
-
-    def log_diagnostics(self, paths):
-        self.policy.log_diagnostics(paths)
-        self.baseline.log_diagnostics(paths)
-
-    def init_opt(self):
+        Returns:
+            dict: Processed sample data, with key
+                * average_return: (float)
         """
-        Initialize the optimization procedure. If using tf, this may
-        include declaring all the variables and compiling functions
-        """
-        raise NotImplementedError
+        baselines = []
+        returns = []
 
-    def get_itr_snapshot(self, itr, samples_data):
-        """
-        Returns all the data that should be saved in the snapshot for this
-        iteration.
-        """
-        raise NotImplementedError
+        max_path_length = self.max_path_length
 
-    def optimize_policy(self, itr, samples_data):
-        raise NotImplementedError
+        if hasattr(self.baseline, 'predict_n'):
+            all_path_baselines = self.baseline.predict_n(paths)
+        else:
+            all_path_baselines = [
+                self.baseline.predict(path) for path in paths
+            ]
+
+        for idx, path in enumerate(paths):
+            # baselines
+            path['baselines'] = all_path_baselines[idx]
+            baselines.append(path['baselines'])
+
+            # returns
+            path['returns'] = special.discount_cumsum(path['rewards'],
+                                                      self.discount)
+            returns.append(path['returns'])
+
+        agent_infos = [path['agent_infos'] for path in paths]
+        agent_infos = tensor_utils.stack_tensor_dict_list([
+            tensor_utils.pad_tensor_dict(p, max_path_length)
+            for p in agent_infos
+        ])
+
+        valids = [np.ones_like(path['returns']) for path in paths]
+        valids = tensor_utils.pad_tensor_n(valids, max_path_length)
+
+        average_discounted_return = (np.mean(
+            [path['returns'][0] for path in paths]))
+
+        undiscounted_returns = [sum(path['rewards']) for path in paths]
+        self.episode_reward_mean.extend(undiscounted_returns)
+
+        ent = np.sum(self.policy.distribution.entropy(agent_infos) *
+                     valids) / np.sum(valids)
+
+        samples_data = dict(average_return=np.mean(undiscounted_returns))
+
+        tabular.record('Iteration', itr)
+        tabular.record('AverageDiscountedReturn', average_discounted_return)
+        tabular.record('AverageReturn', np.mean(undiscounted_returns))
+        tabular.record('Extras/EpisodeRewardMean',
+                       np.mean(self.episode_reward_mean))
+        tabular.record('NumTrajs', len(paths))
+        tabular.record('Entropy', ent)
+        tabular.record('Perplexity', np.exp(ent))
+        tabular.record('StdReturn', np.std(undiscounted_returns))
+        tabular.record('MaxReturn', np.max(undiscounted_returns))
+        tabular.record('MinReturn', np.min(undiscounted_returns))
+
+        return samples_data
--- a/garage/np/algos/batch_polopt.py
+++ b/garage/np/algos/batch_polopt.py
@@ -9,18 +9,20 @@ from garage.np.algos.base import RLAlgorithm
 
 class BatchPolopt(RLAlgorithm):
     """A batch-based algorithm interleaves sampling and policy optimization.
+
     In one round of training, the runner will first instruct the sampler to do
     environment rollout and the sampler will collect a given number of samples
     (in terms of environment interactions). The collected paths are then
     absorbed by `RLAlgorithm.train_once()` and an algorithm performs one step
-    of policy optimization.
-    The updated policy will then be used in the next round of sampling.
+    of policy optimization. The updated policy will then be used in the
+    next round of sampling.
 
     Args:
         policy (garage.tf.policies.base.Policy): Policy.
         baseline (garage.tf.baselines.Baseline): The baseline.
         discount (float): Discount.
         max_path_length (int): Maximum length of a single rollout.
+
     """
 
     def __init__(self, policy, baseline, discount, max_path_length):
@@ -41,6 +43,7 @@ class BatchPolopt(RLAlgorithm):
         Returns:
             dict: Processed sample data, with key
                 * average_return: (float)
+
         """
         baselines = []
         returns = []
--- a/garage/np/algos/cem.py
+++ b/garage/np/algos/cem.py
@@ -1,10 +1,10 @@
 from dowel import logger, tabular
 import numpy as np
 
-from garage.np.algos.base import RLAlgorithm
+from garage.np.algos import BatchPolopt
 
 
-class CEM(RLAlgorithm):
+class CEM(BatchPolopt):
     r"""Cross Entropy Method.
 
     CEM works by iteratively optimizing a gaussian distribution of policy.
@@ -23,16 +23,17 @@ class CEM(RLAlgorithm):
         This implementation leverage n_epoch_cycles to do rollouts for a single
         policy in an epoch cycle.
 
-    Attributes:
-        env_spec(garage.envs.EnvSpec): Environment specification.
-        policy(garage.np.policies.Policy): Action policy.
-        baseline(): Baseline for GAE (Generalized Advantage Estimation).
-        n_samples(int): Number of policies sampled in one epoch.
-        max_path_length(int):  Maximum length of a single rollout.
-        discount(float): Environment reward discount.
-        init_std(float): Initial std for policy param distribution.
-        extra_std(float): Decaying std added to param distribution.
-        extra_decay_time(float): Epochs that it takes to decay extra std.
+    Args:
+        env_spec (garage.envs.EnvSpec): Environment specification.
+        policy (garage.np.policies.Policy): Action policy.
+        baseline(garage.np.baselines.Baseline): Baseline for GAE
+            (Generalized Advantage Estimation).
+        n_samples (int): Number of policies sampled in one epoch.
+        max_path_length (int): Maximum length of a single rollout.
+        discount (float): Environment reward discount.
+        init_std (float): Initial std for policy param distribution.
+        extra_std (float): Decaying std added to param distribution.
+        extra_decay_time (float): Epochs that it takes to decay extra std.
     """
 
     def __init__(self,
@@ -40,24 +41,20 @@ class CEM(RLAlgorithm):
                  policy,
                  baseline,
                  n_samples,
-                 gae_lambda=1,
-                 max_path_length=500,
                  discount=0.99,
+                 max_path_length=500,
                  init_std=1,
                  best_frac=0.05,
                  extra_std=1.,
                  extra_decay_time=100):
+        super().__init__(policy, baseline, discount, max_path_length)
         self.env_spec = env_spec
-        self.policy = policy
-        self.baseline = baseline
         self.n_samples = n_samples
-        self.extra_decay_time = extra_decay_time
-        self.extra_std = extra_std
-        self.best_frac = best_frac
+
         self.init_std = init_std
-        self.discount = discount
-        self.gae_lambda = gae_lambda
-        self.max_path_length = max_path_length
+        self.best_frac = best_frac
+        self.extra_std = extra_std
+        self.extra_decay_time = extra_decay_time
 
         # epoch-wise
         self.cur_std = self.init_std
@@ -81,6 +78,8 @@ class CEM(RLAlgorithm):
             self.n_params) * sample_std + self.cur_mean
 
     def train_once(self, itr, paths):
+        paths = self.process_samples(itr, paths)
+
         epoch = itr // self.n_samples
         i_sample = itr - epoch * self.n_samples
         tabular.record('Epoch', epoch)
--- a/garage/np/algos/cem.py
+++ b/garage/np/algos/cem.py
@@ -5,7 +5,7 @@ from garage.np.algos import BatchPolopt
 
 
 class CEM(BatchPolopt):
-    r"""Cross Entropy Method.
+    """Cross Entropy Method.
 
     CEM works by iteratively optimizing a gaussian distribution of policy.
 
@@ -29,11 +29,13 @@ class CEM(BatchPolopt):
         baseline(garage.np.baselines.Baseline): Baseline for GAE
             (Generalized Advantage Estimation).
         n_samples (int): Number of policies sampled in one epoch.
-        max_path_length (int): Maximum length of a single rollout.
         discount (float): Environment reward discount.
+        max_path_length (int): Maximum length of a single rollout.
+        best_frac (float): The best fraction.
         init_std (float): Initial std for policy param distribution.
         extra_std (float): Decaying std added to param distribution.
         extra_decay_time (float): Epochs that it takes to decay extra std.
+
     """
 
     def __init__(self,
@@ -47,7 +49,7 @@ class CEM(BatchPolopt):
                  best_frac=0.05,
                  extra_std=1.,
                  extra_decay_time=100):
-        super().__init__(policy, baseline, discount, max_path_length)
+        super(CEM, self).__init__(policy, baseline, discount, max_path_length)
         self.env_spec = env_spec
         self.n_samples = n_samples
 
--- a/garage/np/algos/cma_es.py
+++ b/garage/np/algos/cma_es.py
@@ -6,7 +6,7 @@ from garage.np.algos import BatchPolopt
 
 
 class CMAES(BatchPolopt):
-    r"""Covariance Matrix Adaptation Evolution Strategy.
+    """Covariance Matrix Adaptation Evolution Strategy.
 
     Note:
         The CMA-ES method can hardly learn a successful policy even for
@@ -19,9 +19,10 @@ class CMAES(BatchPolopt):
         baseline (garage.np.baselines.Baseline): Baseline for GAE
             (Generalized Advantage Estimation).
         n_samples (int): Number of policies sampled in one epoch.
-        max_path_length (int): Maximum length of a single rollout.
         discount (float): Environment reward discount.
+        max_path_length (int): Maximum length of a single rollout.
         sigma0 (float): Initial std for param distribution.
+
     """
 
     def __init__(self,
@@ -32,7 +33,8 @@ class CMAES(BatchPolopt):
                  discount=0.99,
                  max_path_length=500,
                  sigma0=1.):
-        super().__init__(policy, baseline, discount, max_path_length)
+        super(CMAES, self).__init__(policy, baseline, discount,
+                                    max_path_length)
         self.env_spec = env_spec
         self.policy = policy
         self.n_samples = n_samples
--- a/garage/np/algos/cma_es.py
+++ b/garage/np/algos/cma_es.py
@@ -2,10 +2,10 @@ import cma
 from dowel import logger, tabular
 import numpy as np
 
-from garage.np.algos.base import RLAlgorithm
+from garage.np.algos import BatchPolopt
 
 
-class CMAES(RLAlgorithm):
+class CMAES(BatchPolopt):
     r"""Covariance Matrix Adaptation Evolution Strategy.
 
     Note:
@@ -13,16 +13,15 @@ class CMAES(RLAlgorithm):
         simple task. It is still maintained here only for consistency with
         original rllab paper.
 
-    Attributes:
-        env_spec(EnvSpec): Environment specification.
-        policy(Policy): Action policy.
-        baseline(): Baseline for GAE (Generalized Advantage Estimation).
-        n_samples(int): Number of policies sampled in one epoch.
-        gae_lambda(float): Lambda used for generalized advantage estimation.
-        max_path_length(int):  Maximum length of a single rollout.
-        discount(float): Environment reward discount.
-        sigma0(float): Initial std for param distribution.
-
+    Args:
+        env_spec (garage.envs.EnvSpec): Environment specification.
+        policy (garage.np.policies.Policy): Action policy.
+        baseline (garage.np.baselines.Baseline): Baseline for GAE
+            (Generalized Advantage Estimation).
+        n_samples (int): Number of policies sampled in one epoch.
+        max_path_length (int): Maximum length of a single rollout.
+        discount (float): Environment reward discount.
+        sigma0 (float): Initial std for param distribution.
     """
 
     def __init__(self,
@@ -30,18 +29,15 @@ class CMAES(RLAlgorithm):
                  policy,
                  baseline,
                  n_samples,
-                 gae_lambda=1,
-                 max_path_length=500,
                  discount=0.99,
+                 max_path_length=500,
                  sigma0=1.):
+        super().__init__(policy, baseline, discount, max_path_length)
         self.env_spec = env_spec
         self.policy = policy
-        self.baseline = baseline
         self.n_samples = n_samples
+
         self.sigma0 = sigma0
-        self.discount = discount
-        self.gae_lambda = gae_lambda
-        self.max_path_length = max_path_length
 
         init_mean = self.policy.get_param_values()
         self.es = cma.CMAEvolutionStrategy(init_mean, sigma0)
@@ -55,6 +51,8 @@ class CMAES(RLAlgorithm):
         return self.es.ask(self.n_samples)
 
     def train_once(self, itr, paths):
+        paths = self.process_samples(itr, paths)
+
         epoch = itr // self.n_samples
         i_sample = itr - epoch * self.n_samples
 
--- a/garage/sampler/base.py
+++ b/garage/sampler/base.py
@@ -1,195 +1,39 @@
-from dowel import logger, tabular
-import numpy as np
-
-from garage.misc import special, tensor_utils
-from garage.sampler import utils
+import abc
 
 
-class Sampler:
+class Sampler(abc.ABC):
+    @abc.abstractmethod
     def start_worker(self):
+        """Initialize the sampler,
+        e.g. launching parallel workers if necessary.
         """
-        Initialize the sampler, e.g. launching parallel workers if necessary.
-        """
-        raise NotImplementedError
+        pass
 
+    @abc.abstractmethod
     def obtain_samples(self, itr):
-        """
-        Collect samples for the given iteration number.
-        :param itr: Iteration number.
-        :return: A list of paths.
-        """
-        raise NotImplementedError
+        """Collect samples for the given iteration number.
 
-    def process_samples(self, itr, paths):
-        """
-        Return processed sample data (typically a dictionary of concatenated
-        tensors) based on the collected paths.
-        :param itr: Iteration number.
-        :param paths: A list of collected paths.
-        :return: Processed sample data.
-        """
-        raise NotImplementedError
+        Args:
+            itr (int): Iteration number.
 
+        Returns:
+            list[dict]: A list of paths.
+        """
+        pass
+
+    @abc.abstractmethod
     def shutdown_worker(self):
+        """Terminate workers if necessary.
         """
-        Terminate workers if necessary.
-        """
-        raise NotImplementedError
+        pass
 
 
 class BaseSampler(Sampler):
     def __init__(self, algo, env):
         """
-        :type algo: BatchPolopt
+        Args:
+            algo (garage.np.algos.RLAlgorithm): The algorithm.
+            env (gym.Env): The environment.
         """
         self.algo = algo
         self.env = env
-
-    def process_samples(self, itr, paths):
-        baselines = []
-        returns = []
-
-        if hasattr(self.algo.baseline, 'predict_n'):
-            all_path_baselines = self.algo.baseline.predict_n(paths)
-        else:
-            all_path_baselines = [
-                self.algo.baseline.predict(path) for path in paths
-            ]
-
-        for idx, path in enumerate(paths):
-            path_baselines = np.append(all_path_baselines[idx], 0)
-            deltas = path['rewards'] + \
-                self.algo.discount * path_baselines[1:] - path_baselines[:-1]
-            path['advantages'] = special.discount_cumsum(
-                deltas, self.algo.discount * self.algo.gae_lambda)
-            path['returns'] = special.discount_cumsum(path['rewards'],
-                                                      self.algo.discount)
-            baselines.append(path_baselines[:-1])
-            returns.append(path['returns'])
-
-        ev = special.explained_variance_1d(
-            np.concatenate(baselines), np.concatenate(returns))
-
-        if not self.algo.policy.recurrent:
-            observations = tensor_utils.concat_tensor_list(
-                [path['observations'] for path in paths])
-            actions = tensor_utils.concat_tensor_list(
-                [path['actions'] for path in paths])
-            rewards = tensor_utils.concat_tensor_list(
-                [path['rewards'] for path in paths])
-            returns = tensor_utils.concat_tensor_list(
-                [path['returns'] for path in paths])
-            advantages = tensor_utils.concat_tensor_list(
-                [path['advantages'] for path in paths])
-            env_infos = tensor_utils.concat_tensor_dict_list(
-                [path['env_infos'] for path in paths])
-            agent_infos = tensor_utils.concat_tensor_dict_list(
-                [path['agent_infos'] for path in paths])
-
-            if self.algo.center_adv:
-                advantages = utils.center_advantages(advantages)
-
-            if self.algo.positive_adv:
-                advantages = utils.shift_advantages_to_positive(advantages)
-
-            average_discounted_return = \
-                np.mean([path['returns'][0] for path in paths])
-
-            undiscounted_returns = [sum(path['rewards']) for path in paths]
-
-            ent = np.mean(self.algo.policy.distribution.entropy(agent_infos))
-
-            samples_data = dict(
-                observations=observations,
-                actions=actions,
-                rewards=rewards,
-                returns=returns,
-                advantages=advantages,
-                env_infos=env_infos,
-                agent_infos=agent_infos,
-                paths=paths,
-            )
-        else:
-            max_path_length = max([len(path['advantages']) for path in paths])
-
-            # make all paths the same length (pad extra advantages with 0)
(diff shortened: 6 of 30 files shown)


garage/experiment/local_tf_runner.py
```diff
@@ -211,6 +209,7 @@ def save(self, epoch, paths=None):
 
         logger.log('Saved')
 
+
```
CatherineSue: Only one line between class methods. Please delete this.

garage/np/algos/base.py
```diff
@@ -5,3 +5,13 @@ class Algorithm:
 class RLAlgorithm(Algorithm):
     def train(self):
         raise NotImplementedError
+
+    def process_samples(self, itr, paths):
+        """
+        Return processed sample data (typically a dictionary of concatenated
+        tensors) based on the collected paths.
+        :param itr: Iteration number.
```
CatherineSue: I think we are going to use Google's Python docstring style. Please refer to this [template](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html).

garage/np/algos/batch_polopt.py
```diff
@@ -115,6 +118,155 @@ def log_diagnostics(self, paths):
...
+        if hasattr(self.baseline, 'predict_n'):
+            all_path_baselines = self.baseline.predict_n(paths)
+        else:
+            all_path_baselines = [
+                self.baseline.predict(path) for path in paths
+            ]
+
+        for idx, path in enumerate(paths):
+            path_baselines = np.append(all_path_baselines[idx], 0)
+            deltas = path['rewards'] + \
```
CatherineSue: I think brackets are preferred to break lines here.

garage/np/algos/batch_polopt.py
```diff
@@ -115,6 +118,155 @@ def log_diagnostics(self, paths):
...
+            )
+        else:
+            max_path_length = max([len(path['advantages']) for path in paths])
+
+            # make all paths the same length (pad extra advantages with 0)
+            obs = [path['observations'] for path in paths]
+            obs = tensor_utils.pad_tensor_n(obs, max_path_length)
+
+            if self.center_adv:
+                raw_adv = np.concatenate(
```
CatherineSue: Why not use `utils.center_adv` too?

garage/np/algos/batch_polopt.py
```diff
@@ -115,6 +118,155 @@ def log_diagnostics(self, paths):
...
+                [path['advantages'] for path in paths])
+            env_infos = tensor_utils.concat_tensor_dict_list(
+                [path['env_infos'] for path in paths])
+            agent_infos = tensor_utils.concat_tensor_dict_list(
+                [path['agent_infos'] for path in paths])
+
+            if self.center_adv:
+                advantages = utils.center_advantages(advantages)
+
+            if self.positive_adv:
```
CatherineSue: Why recurrent policies don't have this check?

CatherineSue:
Thanks for the great work. I only have some minor comments.

garage/np/algos/batch_polopt.py
```diff
@@ -115,6 +118,155 @@ def log_diagnostics(self, paths):
...
+                rewards=rewards,
+                returns=returns,
+                valids=valids,
+                agent_infos=agent_infos,
+                env_infos=env_infos,
+                paths=paths,
+            )
+
+        logger.log('fitting baseline...')
+        if hasattr(self.baseline, 'fit_with_samples'):
```
zhanpenghe: Why do we fit baseline in preprocess?

garage/np/algos/cem.py
```diff
@@ -81,6 +87,8 @@ def sample_params(self, epoch):
             self.n_params) * sample_std + self.cur_mean
 
     def train_once(self, itr, paths):
+        paths = self.process_samples(itr, paths)
```
zhanpenghe: Can we move this to the runner?

garage/np/algos/batch_polopt.py
zequnyu: Thank you for comments! As @naeioi told me this function is outdated and never being used now. So we may update it later completely?

ryanjulian:
It looks like you may have accidentally introduced a merge commit and are now stuck with an unclean branch.

See a tutorial here https://git-rebase.io/ on how to rebase.

I recommend you rebase on master and squash all your commits into one to avoid this problem.

garage/np/algos/batch_polopt.py
ryanjulian: this was probably just copied from old code, but yes it should be fixes.

garage/np/algos/cem.py
ryanjulian: hmm interesting design question. @naeioi what do you think?

garage/np/algos/base.py
```diff
@@ -5,3 +5,17 @@ class Algorithm:
...
     def train(self):
         raise NotImplementedError
+
+    def process_samples(self, itr, paths):
+        """Return processed sample data (typically a dictionary of concatenated
+        tensors) based on the collected paths.
+
+        Args:
+        itr (int): Iteration number.
+        paths (str): A list of collected paths.
```
ryanjulian: paths is not a str.

garage/np/algos/batch_polopt.py
ryanjulian: which function do you mean, specifically?

garage/np/algos/batch_polopt.py
ryanjulian: some of this is implemented in tensorflow for NPO now, and parts of process_samples may be redundant.

garage/np/algos/cem.py
```diff
@@ -113,5 +121,105 @@ def train_once(self, itr, paths):
         logger.log(tabular)
         return rtn
 
+    def process_samples(self, itr, paths):
```
ryanjulian: is all of this processing actually required for CEM? which of these data does it actually use?

garage/np/algos/cma_es.py
```diff
@@ -80,5 +88,105 @@ def train_once(self, itr, paths):
         logger.log(tabular)
         return rtn
 
+    def process_samples(self, itr, paths):
+        baselines = []
```
ryanjulian: does CMA actually use all of these?

garage/sampler/base.py
```diff
@@ -20,16 +17,6 @@ def obtain_samples(self, itr):
         """
         raise NotImplementedError
```
ryanjulian: please make this an ABC and use `pass` instead

garage/tf/algos/batch_polopt.py
```diff
@@ -50,9 +56,14 @@ def __init__(self,
         self.center_adv = center_adv
         self.positive_adv = positive_adv
         self.fixed_horizon = fixed_horizon
+
+        self.eprewmean = deque(maxlen=100)
```
ryanjulian: please use a better variable name for this -- we don't pay for code hosting by the character :P. `self.episode_reward_mean`

garage/tf/algos/batch_polopt.py
```diff
@@ -63,6 +74,106 @@ def log_diagnostics(self, paths):
         self.policy.log_diagnostics(paths)
         self.baseline.log_diagnostics(paths)
 
+    def process_samples(self, itr, paths):
```
ryanjulian: please audit this function to make sure all of this data is still used by the algorithm. some of it might not be. some might be used only for logging (keep those)

garage/tf/algos/ddpg.py
```diff
@@ -192,10 +192,20 @@ def __setstate__(self, state):
...
     def train_once(self, itr, paths):
+        paths = self.process_samples(itr, paths)
+
         epoch = itr / self.n_epoch_cycles
 
-        self.episode_rewards.extend(paths['undiscounted_returns'])
-        self.success_history.extend(paths['success_history'])
+        paths_c = paths['complete']
+        self.episode_rewards.extend([
+            paths['undiscounted_returns'][i] for i in range(len(paths_c))
```
ryanjulian: i think you want to use `enumerate` for these

garage/tf/algos/ddpg.py
```diff
@@ -192,10 +192,20 @@ def __setstate__(self, state):
         self.init_opt()
 
     def train_once(self, itr, paths):
+        paths = self.process_samples(itr, paths)
+
         epoch = itr / self.n_epoch_cycles
 
-        self.episode_rewards.extend(paths['undiscounted_returns'])
-        self.success_history.extend(paths['success_history'])
+        paths_c = paths['complete']
```
ryanjulian: this would be clearer without the helper variable.

garage/tf/algos/ddpg.py
```diff
@@ -192,10 +192,20 @@ def __setstate__(self, state):
...
+        self.episode_rewards.extend([
+            paths['undiscounted_returns'][i] for i in range(len(paths_c))
+            if paths_c[i]
+        ])
+        self.success_history.extend([
+            paths['success_history'][i] for i in range(len(paths_c))
```
ryanjulian: i think this is equivalent
```python
[p for p, complete in zip(paths['success_history'], paths['complete']) if complete]
```

garage/tf/algos/ddpg.py
```diff
@@ -192,10 +192,20 @@ def __setstate__(self, state):
...
 
     def train_once(self, itr, paths):
+        paths = self.process_samples(itr, paths)
+
         epoch = itr / self.n_epoch_cycles
 
-        self.episode_rewards.extend(paths['undiscounted_returns'])
-        self.success_history.extend(paths['success_history'])
+        paths_c = paths['complete']
+        self.episode_rewards.extend([
```
ryanjulian:
```python
[p for p, complete in zip(paths['undiscounted_returns'], paths['complete']) if complete]
```

garage/tf/algos/ddpg.py
```diff
@@ -192,10 +192,20 @@ def __setstate__(self, state):
...
+            paths['undiscounted_returns'][i] for i in range(len(paths_c))
+            if paths_c[i]
+        ])
+        self.success_history.extend([
```
ryanjulian:
```python
[p for p, complete in zip(paths['success_history'], paths['complete']) if complete]
```

garage/tf/algos/off_policy_rl_algorithm.py
```diff
@@ -53,6 +53,31 @@ def log_diagnostics(self, paths):
         self.policy.log_diagnostics(paths)
         self.qf.log_diagnostics(paths)
 
+    def process_samples(self, itr, paths):
+        """Return processed sample data based on the collected paths.
+
+        Args:
+            itr(int): Iteration number.
+            paths(list): A list of collected paths.
```
ryanjulian: list of what?

garage/tf/algos/off_policy_rl_algorithm.py
```diff
@@ -53,6 +53,31 @@ def log_diagnostics(self, paths):
...
+
+        Returns:
+            list: Processed sample data.
```
ryanjulian: list of what?

ryanjulian: this function as written pretty clearly returns a dict...

garage/tf/samplers/batch_sampler.py
```diff
@@ -24,11 +18,10 @@ class BatchSampler(BaseSampler):
     def __init__(self, algo, env, n_envs):
         super(BatchSampler, self).__init__(algo, env)
         self.n_envs = n_envs
-        self.eprewmean = deque(maxlen=100)
 
     def start_worker(self):
         assert singleton_pool.initialized, \
```
ryanjulian: please use () to split lines

garage/np/algos/batch_polopt.py
zequnyu: It is the process_samples() in garage.sampler.BaseSampler in master branch, which I'm trying to move to garage.np.algos.BatchPolopt.

zequnyu:
I will try to address all the comments and squash it into a single commit.

garage/tf/algos/off_policy_rl_algorithm.py
zequnyu: Don't really get your point here. Not "A list of collected paths"?

garage/tf/algos/batch_polopt.py
zequnyu: Currently `advantages` and `env_infos` are not in use. Can I delete them?

garage/np/algos/base.py
```diff
@@ -5,3 +5,17 @@ class Algorithm:
 class RLAlgorithm(Algorithm):
     def train(self):
         raise NotImplementedError
+
```
ryanjulian: please make `RLAlgorithm` an ABC and use `pass` and `@abstractmethod` to define these.

garage/np/algos/base.py
```diff
@@ -5,3 +5,17 @@ class Algorithm:
...
+    def process_samples(self, itr, paths):
+        """Return processed sample data (typically a dictionary of concatenated
+        tensors) based on the collected paths.
+
+        Args:
+        itr (int): Iteration number.
+        paths (list): A list of collected paths.
```
ryanjulian: what is the type inside the list? is it `list[dict]` or `list[str]` or `list[list]`, etc?

garage/np/algos/batch_polopt.py
```diff
@@ -115,6 +118,155 @@ def log_diagnostics(self, paths):
         self.policy.log_diagnostics(paths)
         self.baseline.log_diagnostics(paths)
 
+    def process_samples(self, itr, paths):
+        baselines = []
+        returns = []
+
+        if hasattr(self.baseline, 'predict_n'):
```
ryanjulian: hmm it is probably more appropriate to calculate the baseline in the training loop. @CatherineSue what do you think?

garage/sampler/base.py
```diff
@@ -1,40 +1,8 @@
...
-    def shutdown_worker(self):
-        """
-        Terminate workers if necessary.
-        """
-        raise NotImplementedError
+class Sampler(ABC):
+    pass
 
 
 class BaseSampler(Sampler):
```
ryanjulian: please update this docstring to use the new format

garage/sampler/base.py
```diff
@@ -1,40 +1,8 @@
...
 
-
-class Sampler:
-    def start_worker(self):
-        """
-        Initialize the sampler, e.g. launching parallel workers if necessary.
-        """
-        raise NotImplementedError
-
-    def obtain_samples(self, itr):
```
ryanjulian: why remove all these abstract methods from Sampler? this seems to be the sampler interface.

garage/tf/algos/batch_polopt.py
ryanjulian: env_infos no, advantages yes

garage/tf/algos/off_policy_rl_algorithm.py
ryanjulian: when the type in a doctstring is a collection, we usually include the inner type. e.g. a list of strings is `list[str]`

garage/tf/algos/off_policy_rl_algorithm.py
```diff
@@ -53,6 +53,29 @@ def log_diagnostics(self, paths):
         self.policy.log_diagnostics(paths)
         self.qf.log_diagnostics(paths)
 
+    def process_samples(self, itr, paths):
+        """Return processed sample data based on the collected paths.
+        Args:
+            itr(int): Iteration number.
+            paths(list): A list of collected paths.
+        Returns:
+            dict: Processed sample data.
```
ryanjulian: can you say a little bit more about the shape of the data? what keys does it have and what are their types?

garage/tf/algos/batch_polopt.py
ryanjulian: please provide a docstring for this function

garage/np/algos/cem.py
naeioi: I prefer to give an algorithm more control over the batch data. That is, let the algorithm decides when and how to do preprocessing, or even not doing it at all.

garage/tf/algos/off_policy_rl_algorithm.py
```diff
@@ -53,6 +53,35 @@ def log_diagnostics(self, paths):
...
+
+        Args:
+            itr (int): Iteration number.
+            paths (list[dict]): A list of collected paths.
+
+        Returns:
+            dict: Processed sample data.
+            {'undiscounted_returns': list[float],
+            'success_history': list[float],
+            'complete': list[boolean]}
```
naeioi: Use `bool` instead of `boolean`.

garage/np/algos/base.py
```diff
@@ -1,7 +1,95 @@
...
 
-class RLAlgorithm(Algorithm):
-    def train(self):
-        raise NotImplementedError
+from garage.logger import tabular
+from garage.misc import special
+from garage.tf.misc import tensor_utils
+
+
+class RLAlgorithm():
```
CatherineSue: `class RLAlgorithm:`?

garage/sampler/base.py
```diff
@@ -1,195 +1,40 @@
...
-        raise NotImplementedError
+        pass
 
 
 class BaseSampler(Sampler):
     def __init__(self, algo, env):
         """
-        :type algo: BatchPolopt
+        Args:
+            algo (BatchPolopt): The algorithm to use.
```
CatherineSue: missing argument documentation

garage/np/algos/base.py
```diff
@@ -1,7 +1,95 @@
...
+from collections import deque
 
+import numpy as np
 
-class RLAlgorithm(Algorithm):
-    def train(self):
-        raise NotImplementedError
+from garage.logger import tabular
+from garage.misc import special
+from garage.tf.misc import tensor_utils
```
CatherineSue: This looks weird to me. Why do we need tf module dependency in `garage.np`?

garage/np/algos/batch_polopt.py
CatherineSue: Sorry just saw this. I agree.

garage/np/algos/base.py
zequnyu: I discussed with KR yesterday about it and this is the way he indicated. The `cem` and `cam_es` algorithms inherit from it instead of `np.BatchPolopt`. Actually `np.BatchPolopt` is pretty much not be used by anyone except some existing tests. By looking up the test for `cem` and `cam_es`, I found that it actually use our `OnPolicyVectorizedSampler`, which is inside tf. And also our `LocalRunner` is now a tf runner. I put the process_samples() here (`np.algos.RLAlgorithm`) for `cem` and `cam_es` use only, and rewrote some code to let them share this method without unnecessary duplicate implementation, as `cem` and `cam_es` use the same data to train.

To me it's actually more comfortable to make `RLAlgorithm` abstract and put the process_samples() for `cem` and `cam_es` into `np.BatchPolopt` and let `cem` and `cam_es` inherit from it. But that means I will need to remove some tests, such like `tests.fixtures.algos.InstrumentedBatchPolopt`. So my concern really is, am I allowed to do so?

I'm pretty new to the policy. So please correct me if I misunderstood something.

garage/np/algos/base.py
ryanjulian: i think the root cause here is that the samplers should probably not be in `tf`

zequnyu: It seems that the function used here is the same as in garage.misc.tensor_utils. I will update to use it to avoid tf dependency. Thanks for pointing out.

garage/np/algos/cem.py
zhanpenghe: Hmm I think the algorithms are defining how to optimize the models for the system while the runner are defining the flow of everything. (e.g. sampling data -> preprocessing them -> feed them to the algorithms for optimization).

If the workflow of a training procedure is different, it sounds like a different runner for me.

zequnyu: Moving the pre-processing to the runner makes more sense to me: it's better that `train_once()` do only training.

ryanjulian: This can always be refactored easily, so for now let's just go with Keren's judgement and see how it pans out.

garage/np/algos/base.py
```diff
@@ -1,7 +1,91 @@
...
+from dowel import tabular
+import numpy as np
 
-class RLAlgorithm(Algorithm):
-    def train(self):
-        raise NotImplementedError
+from garage.misc import special, tensor_utils
+
+
+class RLAlgorithm():
```
ryanjulian: i think chang means you should remove the () from `class RLAlgorithm:`

garage/np/algos/base.py
```diff
@@ -1,7 +1,91 @@
...
+        self.episode_reward_mean = deque(maxlen=100)
+
+    def process_samples(self, itr, paths):
+        """Return processed sample data based on the collected paths.
+        This method is currently used by cem and cam_es.
+        Args:
+            itr (int): Iteration number.
+            paths (list[dict]): A list of collected paths.
+        Returns:
+            dict: Processed sample data. Only {'average_return': float]}
```
ryanjulian: what do you mean by "Only {'average_return': float]}"?

garage/sampler/base.py
```diff
@@ -1,195 +1,41 @@
...
-from garage.sampler import utils
 
-
-class Sampler:
+class Sampler(ABC):
+    @abstractmethod
     def start_worker(self):
+        """Initialize the sampler,
+        e.g. launching parallel workers if necessary.
+
```
ryanjulian: remove extra newline

garage/sampler/base.py
```diff
@@ -1,195 +1,41 @@
...
+        pass
 
 
 class BaseSampler(Sampler):
     def __init__(self, algo, env):
         """
-        :type algo: BatchPolopt
+        Args:
+            algo (garage.np.algos.RLAlgorithm): The algorithm.
+            env (garage.envs.EnvSpec): The environment.
```
ryanjulian: is it an env or an EnvSpec? I think it is an env. the type is then `gym.Env`

garage/tf/algos/batch_polopt.py
```diff
@@ -64,6 +75,112 @@ def log_diagnostics(self, paths):
...
+        """Return processed sample data based on the collected paths.
+
+        Args:
+            itr (int): Iteration number.
+            paths (list[dict]): A list of collected paths.
+
+        Returns:
+            dict: Processed sample data.
+
```
ryanjulian: please remove extra newline

garage/tf/algos/batch_polopt.py
```diff
@@ -64,6 +75,112 @@ def log_diagnostics(self, paths):
...
 
+    def process_samples(self, itr, paths):
+        """Return processed sample data based on the collected paths.
+
+        Args:
+            itr (int): Iteration number.
+            paths (list[dict]): A list of collected paths.
+
+        Returns:
+            dict: Processed sample data.
```
ryanjulian: is this a dict of lists, a dict of np.ndarrays, etc?

tests/garage/tf/algos/test_dqn.py
```diff
@@ -55,7 +55,7 @@ def test_dqn_cartpole(self):
                 n_epochs=n_epochs,
                 n_epoch_cycles=n_epoch_cycles,
                 batch_size=sampler_batch_size)
-            assert last_avg_ret > 20
+            assert last_avg_ret > 15
```
ryanjulian: did you check with @ahtsan about this?

lowering the expected performance of an algorithm is a pretty serious change to a test. we need to make sure the new changes didn't cause the performance to drop.

garage/tf/algos/off_policy_rl_algorithm.py
```diff
@@ -53,6 +53,35 @@ def log_diagnostics(self, paths):
...
+    def process_samples(self, itr, paths):
+        """Return processed sample data based on the collected paths.
+
+        Args:
+            itr (int): Iteration number.
+            paths (list[dict]): A list of collected paths.
+
+        Returns:
+            dict: Processed sample data.
+            {'undiscounted_returns': list[float],
```
ryanjulian: this is probably best displayed as a list.

tests/garage/tf/algos/test_dqn.py
ahtsan: Yes we discussed. We both confirm it is returning ~30 but sometimes ~20 on Travis. I believe it is another flaky test.

garage/np/algos/base.py
```diff
@@ -1,7 +1,91 @@
...
+        average_discounted_return = (np.mean(
+            [path['returns'][0] for path in paths]))
+
+        undiscounted_returns = [sum(path['rewards']) for path in paths]
+        self.episode_reward_mean.extend(undiscounted_returns)
+
+        ent = np.sum(self.policy.distribution.entropy(agent_infos) *
+                     valids) / np.sum(valids)
+
+        samples_data = dict(average_return=np.mean(undiscounted_returns))
```
ryanjulian: if you only return `average_return`, why calculate all this other stuff?

ryanjulian: if you are going to do these calculations/logging in the base class, make sure that the inheriting classes can reuse them, rather than having to recalculate everything again

garage/tf/algos/batch_polopt.py
```diff
@@ -64,6 +75,112 @@ def log_diagnostics(self, paths):
...
+            valids=valids,
+            agent_infos=agent_infos,
+            env_infos=env_infos,
+            paths=paths,
+            average_return=np.mean(undiscounted_returns),
+        )
+
+        tabular.record('Iteration', itr)
+        tabular.record('AverageDiscountedReturn', average_discounted_return)
+        tabular.record('AverageReturn', np.mean(undiscounted_returns))
```
ryanjulian: couldn't it get most of this by calling `super().process_samples(itr, paths)` first?

tests/garage/tf/algos/test_dqn.py
ryanjulian: checked with anson

garage/np/algos/base.py
ryanjulian: oh i see. please use the format i suggested below.

garage/np/algos/base.py
ryanjulian: this class is missing a docstring -- please add one

i'm surprised this passes the CI...

garage/np/algos/base.py
```diff
@@ -1,7 +1,91 @@
...
+        self.gae_lambda = gae_lambda
+        self.discount = discount
+        self.max_path_length = max_path_length
+
+        self.episode_reward_mean = deque(maxlen=100)
+
+    def process_samples(self, itr, paths):
+        """Return processed sample data based on the collected paths.
+        This method is currently used by cem and cam_es.
+        Args:
```
ryanjulian: missing newline. shouldn't flake8 catch this?

garage/np/algos/base.py
```diff
@@ -1,7 +1,91 @@
...
+            itr (int): Iteration number.
+            paths (list[dict]): A list of collected paths.
+        Returns:
```
ryanjulian: missing newline. please study the example carefully.

garage/np/algos/cem.py
```diff
@@ -26,7 +26,8 @@ class CEM(RLAlgorithm):
     Attributes:
```
ryanjulian: there are args, not attributes. please fix.

garage/np/algos/cma_es.py
```diff
@@ -14,9 +14,10 @@ class CMAES(RLAlgorithm):
         original rllab paper.
 
     Attributes:
```
ryanjulian: these are Args, not attributes. please fix.

garage/tf/algos/batch_polopt.py
```diff
@@ -23,23 +28,24 @@ def __init__(self,
...
-        :param max_path_length: Maximum length of a single rollout.
-        :param discount: Discount.
-        :param gae_lambda: Lambda used for generalized advantage estimation.
-        :param center_adv: Whether to rescale the advantages so that they have
-         mean 0 and standard deviation 1.
-        :param positive_adv: Whether to shift the advantages so that they are
-         always positive. When used in conjunction with center_adv the
-         advantages will be standardized before shifting.
-        :return:
+        Args:
```
ryanjulian: this docstring is missing a description of the class

garage/tf/algos/batch_polopt.py
ryanjulian: please use the format example i gave below to document the keys in the dict

garage/np/algos/base.py
```diff
@@ -1,7 +1,91 @@
...
+import numpy as np
 
-class RLAlgorithm(Algorithm):
-    def train(self):
-        raise NotImplementedError
+from garage.misc import special, tensor_utils
+
+
+class RLAlgorithm():
+    def __init__(self,
```
ryanjulian: i don't think we want a constructor or methods on `RLAlgorithm` -- it should only be an interface class. 

plus, lots of these constructor arguments and process_samples looks very specific to BatchPolopt.

@naeioi please advise

garage/tf/algos/batch_polopt.py
zequnyu: The examples Chang gave (https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) add new line.

zequnyu: I will delete anyway.

ryanjulian: never mind -- my apology

garage/np/algos/base.py
zhanpenghe: @zequnyu You dont need the parentheses.

zequnyu:
Please see if it's better now.

garage/np/algos/base.py
```diff
@@ -1,7 +1,26 @@
...
 
 
-class RLAlgorithm(Algorithm):
-    def train(self):
-        raise NotImplementedError
+class RLAlgorithm(ABC):
+    """Base class for all the algorithms.
+    """
+
+    @abstractmethod
```
ryanjulian: i think you can remove this from the interface, because it's not called from outside RLAlgorithm

garage/np/algos/base.py
```diff
@@ -1,7 +1,26 @@
...
+
+        Args:
+            itr (int): Iteration number.
+            paths (list[dict]): A list of collected paths.
+        """
+        pass
+
+    @abstractmethod
+    def train_once(self, itr, paths):
+        """Training once.
```
ryanjulian: please leave a much more informative description of this interface. if you are unclear on its purpose, ask Keren for an explanation.

garage/np/algos/batch_polopt.py
```diff
@@ -1,135 +1,84 @@
...
 
-    def get_itr_snapshot(self, itr, samples_data):
-        """
-        Returns all the data that should be saved in the snapshot for this
-        iteration.
+        self.episode_reward_mean = deque(maxlen=100)
+
+    def process_samples(self, itr, paths):
+        """Return processed sample data based on the collected paths.
+        This method is currently used by cem and cam_es.
```
ryanjulian: no need to note where it's used. that's really hard to maintain.

garage/np/algos/batch_polopt.py
```diff
@@ -1,135 +1,84 @@
...
+        samples_data = dict(average_return=np.mean(undiscounted_returns))
+
+        tabular.record('Iteration', itr)
+        tabular.record('AverageReturn', np.mean(undiscounted_returns))
+
+        return samples_data
 
-    def optimize_policy(self, itr, samples_data):
-        raise NotImplementedError
+    def train_once(self, itr, paths):
```
ryanjulian: make BatchPolopt an ABC as well, so you can declare this method abstract (or skip implementing it--idk what the rules are for one ABC inheriting from another)

garage/tf/algos/batch_polopt.py
```diff
@@ -64,6 +74,121 @@ def log_diagnostics(self, paths):
...
+
+        Args:
+            itr (int): Iteration number.
+            paths (list[dict]): A list of collected paths.
+
+        Returns:
+            dict: Processed sample data, with key
+                * observations: (list[list])
+                * actions: (list[list]))
+                * rewards: (list[list[float]])
```
ryanjulian: most of this data definitely ends up as np.ndarray, not list-of-lists...

garage/sampler/base.py
```diff
@@ -1,195 +1,39 @@
-from dowel import logger, tabular
-import numpy as np
+from abc import ABC, abstractmethod
```
ryanjulian: please only import packages from stdlib. this make it easier to spot where dependencies come form.

e.g.

```python
import abc

class Sampler(abc.ABC):
    # etc...
```

garage/tf/algos/batch_polopt.py
```diff
@@ -64,6 +74,121 @@ def log_diagnostics(self, paths):
...
+                * baselines: (list[list[float]])
+                * returns: (list[list[float]])
+                * valids: (list[list[bool]])
+                * agent_infos: (dict)
+                * agent_infos: (dict)
```
ryanjulian: repeated agent_infos

garage/np/algos/batch_polopt.py
ryanjulian: i think we should file an issue for this one rather than asking @zequnyu to fix it in this PR

garage/np/algos/batch_polopt.py
```diff
@@ -1,135 +1,84 @@
...
+                 env_spec,
                  policy,
                  baseline,
-                 scope=None,
-                 n_itr=500,
-                 start_itr=0,
-                 batch_size=5000,
-                 max_path_length=500,
-                 discount=0.99,
+                 n_samples,
```
ryanjulian: were all of these constructor arguments unused?

garage/np/algos/batch_polopt.py
```diff
@@ -1,135 +1,84 @@
...
+            path['returns'] = special.discount_cumsum(path['rewards'],
+                                                      self.discount)
+            returns.append(path['returns'])
+
+        undiscounted_returns = [sum(path['rewards']) for path in paths]
+
+        samples_data = dict(average_return=np.mean(undiscounted_returns))
+
+        tabular.record('Iteration', itr)
+        tabular.record('AverageReturn', np.mean(undiscounted_returns))
```
ryanjulian: all of the metrics logged in tf.algos.BatchPolopt (StdReturn, etc.) are useful and should still be logged here.

garage/np/algos/batch_polopt.py
```diff
@@ -1,135 +1,84 @@
...
-from garage.plotter import Plotter
-from garage.sampler import BatchSampler
 
 
 class BatchPolopt(RLAlgorithm):
     """
-    Base class for batch sampling-based policy optimization methods.
-    This includes various policy gradient methods like vpg, npg, ppo, trpo,
-    etc.
+    Base class for cem and cam_es.
```
ryanjulian: Please leave a more descriptive comments. Consult Chang and Keren for advice on how to explain the purpose of this class.

garage/np/algos/batch_polopt.py
```diff
@@ -1,135 +1,84 @@
...
+
+    Args:
+        env_spec (garage.envs.EnvSpec): Environment specification.
+        policy (garage.tf.policies.base.Policy): Policy.
```
ryanjulian: does it actually use policies and baselines from garage.tf?

garage/tf/algos/batch_polopt.py
```diff
@@ -22,25 +46,6 @@ def __init__(self,
                  positive_adv=False,
                  fixed_horizon=False,
                  **kwargs):
```
ryanjulian: please don't swallow kwargs

garage/np/algos/batch_polopt.py
zequnyu: Yes.

garage/np/algos/batch_polopt.py
zequnyu:
```
ent = np.sum(self.policy.distribution.entropy(agent_infos) *
                     valids) / np.sum(valids)
```
> if you only return average_return, why calculate all this other stuff?

I kept all the logging used variables last time. Would you tell me which one is not necessary to calculate?
```
        tabular.record('Iteration', itr)
        tabular.record('AverageDiscountedReturn', average_discounted_return)
        tabular.record('AverageReturn', np.mean(undiscounted_returns))
        tabular.record('Extras/EpisodeRewardMean',
                       np.mean(self.episode_reward_mean))
        tabular.record('NumTrajs', len(paths))
        tabular.record('Entropy', ent)
        tabular.record('Perplexity', np.exp(ent))
        tabular.record('StdReturn', np.std(undiscounted_returns))

```

garage/tf/algos/batch_polopt.py
zequnyu: I didn't get your point. I didn't make any concrete change here.

garage/np/algos/batch_polopt.py
zequnyu: Seems so. The current implementation of cem and cam_es on master has all these arguments though.

garage/np/algos/base.py
zequnyu: Remove @abstractmethod?

garage/tf/algos/batch_polopt.py
naeioi: It might be more involved than expected, because several fields are produced by `tensor_utils`, but the np version doesn't.

naeioi: One thing we can try is 
- make a full version of np `process_samples()` (compute and log everything).
- inherit tf `BatchPolopt` from np `BatchPolopt`.
- call `super().process_samples(itr, paths)` to get processed sample.
- cast processed sample fields from ndarray to tensor.

zequnyu: Do you want me to implement this way? I think we are trying to separate np and tf algorithms. Put the full version of process_samples() to np seems confusing.

garage/np/algos/batch_polopt.py
zequnyu: If you mean tf.algos.BatchPolopt, it does have all the things. For np it seems only AverageReturn is needed.

garage/tf/algos/batch_polopt.py
zequnyu: Now many algorithms are using `kwargs`. I worked on it for a while. I think we may need another issue/PR to update all of them.

zequnyu:
For comments not mark resolved, I left my comments and may need some help to move forward.

garage/tf/algos/batch_polopt.py
ryanjulian: i think this is fair. please file an issue for auditing this throughout the codebase (you don't need to assign yourself).

however, please do fix it for the classes you touch in this PR. we often fix small issues like this as we go using the rule "if you touch it, you have to fix it."

garage/np/algos/batch_polopt.py
ryanjulian: if an argument is not used in a class, it should not be in the constructor. if instead, those arguments are used by child classes, the arguments should appear in the child constructor, not the parent constructor.

can you remove the arguments not used by `BatchPolopt`, and add them to CEM and CMAES as needed?

garage/np/algos/batch_polopt.py
ryanjulian: i guess what i'm saying is that all on-policy algorithms need to log these statistics, so in that sense, many of these calculations are "needed," even if they aren't used during calculations in the algorithms.

garage/np/algos/base.py
ryanjulian: i don't see a fix here

ryanjulian: generally we wait for the reviewer to resolve any comments, not the author :)

garage/np/algos/batch_polopt.py
ryanjulian: @zequnyu did you file an issue for this?

garage/sampler/base.py
ryanjulian: this isn't updated. did you forget to push?

garage/tf/algos/batch_polopt.py
```diff
@@ -1,13 +1,37 @@
-from dowel import logger
+from collections import deque
```
ryanjulian: please only import stdlib dependencies using their full package, e.g.

```python
import collections

q = collections.deque(5)
```

this make it easier to understand what types are from dependencies when people are reading the code.

garage/tf/algos/batch_polopt.py
ryanjulian: i agree -- they should be separate.

perhaps it would be worth it to create a shared helper function for the most common parts of process samples (e.g. calculating common stats like StdReturn, Entropy, etc.)?

garage/tf/samplers/batch_sampler.py
ryanjulian: this is not fixed yet.

garage/np/algos/base.py
zequnyu: I think this is an outdated version. This is fixed.

garage/np/algos/batch_polopt.py
zequnyu: Filed

garage/np/algos/base.py
```diff
@@ -1,7 +1,18 @@
-class Algorithm:
-    pass
+import abc
+from abc import abstractmethod
```
ryanjulian: please just use `@abc.abstractmethod`

garage/sampler/base.py
```diff
@@ -1,195 +1,40 @@
-from dowel import logger, tabular
-import numpy as np
+import abc
+from abc import abstractmethod
```
ryanjulian: please just use `@abc.abstractmethod`

garage/tf/algos/batch_polopt.py
zequnyu: Updated for all on policy algorithms.

garage/tf/algos/batch_polopt.py
zequnyu: I don't come out a good idea of this. Since I remember you said you are against the utility things. And it seems not worth here to me. @naeioi What do you think?

garage/np/algos/batch_polopt.py
ryanjulian: let's keep all of them for now. just file an issue to extract the common rollout statistics into a library, rather than copy/pasting them between classes

garage/tf/algos/batch_polopt.py
ryanjulian: let's fix it in a different PR

garage/tf/algos/ddpg.py
```diff
@@ -20,50 +11,78 @@
...
+                 name='DDPG'):
         """
-        Construct class.
-
         Args:
-            env_spec(EnvSpec): Environment specification.
-            target_update_tau(float): Interpolation parameter for doing the
-        soft target update.
+            env_spec (EnvSpec): Environment specification.
+            policy (garage.tf.policies.base.Policy): Policy.
```
ryanjulian: please document constructor arguments in the class docstring, as mentioned in `CONTRIBUTING.md`

garage/tf/algos/ddpg.py
```diff
@@ -20,50 +11,78 @@
 
 
 class DDPG(OffPolicyRLAlgorithm):
-    """
-    A DDPG model based on https://arxiv.org/pdf/1509.02971.pdf.
+    """A DDPG model based on https://arxiv.org/pdf/1509.02971.pdf.
```
ryanjulian: the original position of this documentation (module) was fine. but this is fine too. just please be consistent.

garage/tf/algos/dqn.py
```diff
@@ -17,37 +9,67 @@
...
-        qf_optimizer (tf.Optimizer): Optimizer for Q-Function.
-        discount (float): Discount factor for rewards.
-        name (str): Name of the algorithm.
-        target_network_update_freq (int): Frequency of updating target network.
-        grad_norm_clipping (float): Maximum clipping value for clipping tensor
-            values to a maximum L2-norm. It must be larger than 0. If None,
-            no gradient clipping is done. For detail, see docstring for
-            tf.clip_by_norm.
-        double_q (bool): Bool for using double q-network.
+    """DQN from https://arxiv.org/pdf/1312.5602.pdf.
```
ryanjulian: the original position of this documentation was fine. please just make sure all the algorithms are consistent with where they put the documentation (either in module or class).

garage/tf/algos/dqn.py
```diff
@@ -17,37 +9,67 @@
...
                  target_network_update_freq=5,
                  grad_norm_clipping=None,
                  double_q=False,
-                 **kwargs):
+                 reward_scale=1.,
+                 input_include_goal=False,
+                 smooth_return=True,
+                 name='DQN'):
+        """
+        Args:
```
ryanjulian: please document the constructor in the class docstring, as mentioned here https://github.com/rlworkgroup/garage/blob/master/CONTRIBUTING.md#documentation

garage/tf/algos/erwr.py
```diff
@@ -20,20 +20,83 @@ class ERWR(VPG):
...
                  optimizer=None,
                  optimizer_args=None,
-                 positive_adv=True,
-                 **kwargs):
+                 policy_ent_coeff=0.0,
+                 use_softplus_entropy=False,
+                 use_neg_logli_entropy=False,
+                 stop_entropy_gradient=False,
+                 name='ERWR'):
+        """
```
ryanjulian: please document in the class docstring

garage/tf/algos/npo.py
```diff
@@ -32,30 +31,61 @@ class PGLoss(Enum):
...
-                 name='NPO',
-                 policy=None,
                  policy_ent_coeff=0.0,
                  use_softplus_entropy=False,
                  use_neg_logli_entropy=False,
                  stop_entropy_gradient=False,
-                 **kwargs):
+                 name='NPO'):
+        """
+        Args:
```
ryanjulian: these go in the class docstring

garage/tf/algos/off_policy_rl_algorithm.py
```diff
@@ -1,34 +1,50 @@
...
-This module implements a class for off-policy rl algorithms.
-
-Off-policy algorithms such as DQN, DDPG can inherit from it.
-"""
 from garage.np.algos import RLAlgorithm
 
 
 class OffPolicyRLAlgorithm(RLAlgorithm):
-    """This class implements OffPolicyRLAlgorithm."""
+    """This class implements OffPolicyRLAlgorithm for off-policy rl algorithms.
```
ryanjulian: RL

garage/tf/algos/off_policy_rl_algorithm.py
```diff
@@ -1,34 +1,50 @@
...
+                 n_train_steps=50,
+                 buffer_batch_size=64,
+                 min_buffer_size=int(1e4),
+                 rollout_batch_size=1,
+                 reward_scale=1.,
+                 input_include_goal=False,
+                 smooth_return=True,
+                 exploration_strategy=None):
+        """
+        Args:
```
ryanjulian: class docstring

garage/tf/algos/ppo.py
```diff
@@ -11,21 +10,81 @@ class PPO(NPO):
...
+                 use_neg_logli_entropy=False,
+                 stop_entropy_gradient=False,
+                 name='PPO'):
         """
-        Construct class.
-
-        :param optimizer: Optimizer.
-        :param optimizer_args: Optimizer args.
-        :param kwargs:
+        Args:
```
ryanjulian: class docstring

garage/tf/algos/reps.py
```diff
@@ -27,28 +26,53 @@ class REPS(BatchPolopt):
...
                  l2_reg_loss=0.,
                  optimizer=LbfgsOptimizer,
                  optimizer_args=dict(max_opt_itr=50),
                  dual_optimizer=scipy.optimize.fmin_l_bfgs_b,
                  dual_optimizer_args=dict(maxiter=50),
-                 name='REPS',
-                 **kwargs):
+                 name='REPS'):
+        """
+        Args:
```
ryanjulian: class docstring

garage/tf/algos/tnpg.py
```diff
@@ -11,16 +10,81 @@ class TNPG(NPO):
...
+                 max_kl_step=0.01,
+                 optimizer=None,
+                 optimizer_args=None,
+                 policy_ent_coeff=0.0,
+                 use_softplus_entropy=False,
+                 use_neg_logli_entropy=False,
+                 stop_entropy_gradient=False):
+        """
+        Args:
+            env_spec (garage.envs.EnvSpec): Environment specification.
```
ryanjulian: class docstring

garage/tf/algos/trpo.py
```diff
@@ -20,11 +20,61 @@ class TRPO(NPO):
...
+            optimizer=None,
+            optimizer_args=None,
+            policy_ent_coeff=0.0,
+            use_softplus_entropy=False,
+            use_neg_logli_entropy=False,
+            stop_entropy_gradient=False,
+            kl_constraint=KLConstraint.HARD,
+    ):
+        """
+        Args:
```
ryanjulian: class docstring

garage/tf/algos/vpg.py
```diff
@@ -8,7 +8,60 @@ class VPG(NPO):
...
+                 max_kl_step=0.01,
+                 optimizer=None,
+                 optimizer_args=None,
+                 policy_ent_coeff=0.0,
+                 use_softplus_entropy=False,
+                 use_neg_logli_entropy=False,
+                 stop_entropy_gradient=False,
+                 name='VPG'):
+        """
+        Args:
```
ryanjulian: class docstring

zequnyu:
Updated.

garage/sampler/batch_sampler.py
```diff
@@ -11,7 +10,9 @@ def __init__(self, algo):
         """
         Init function.
 
-        :type algo: BatchPolopt
+        Args:
```
ryanjulian: please document the constructor in the class docstring

garage/sampler/base.py
```diff
@@ -1,195 +1,41 @@
...
-        raise NotImplementedError
+        """Terminate workers if necessary."""
+        pass
 
 
 class BaseSampler(Sampler):
     def __init__(self, algo, env):
         """
-        :type algo: BatchPolopt
+        Args:
```
ryanjulian: please document the constructor in the class docstring

garage/tf/algos/batch_polopt.py
```diff
@@ -1,13 +1,39 @@
...
 from garage.np.algos import RLAlgorithm
+from garage.tf.misc import tensor_utils
 
 
 class BatchPolopt(RLAlgorithm):
-    """
-    Base class for batch sampling-based policy optimization methods.
+    """Base class for batch sampling-based policy optimization methods.
+
     This includes various policy gradient methods like vpg, npg, ppo, trpo,
```
ryanjulian: capitalize VPG, NPG, etc.

garage/tf/algos/dqn.py
```diff
@@ -17,37 +9,67 @@
...
+    environments. To deal with pixel environments, numbers of tricks are
+    usually needed, e.g. skipping frames and stacking frames as single
+    observation.
 
     Args:
         env_spec (garage.envs.env_spec.EnvSpec): Environment specification.
-        max_path_length (int): Maximum path length. The episode will terminate
-            when length of trajectory reaches max_path_length.
+        policy (garage.tf.policies.base.Policy): Policy.
+        qf (): The q value network.
```
ryanjulian: please add a type

garage/tf/algos/erwr.py
```diff
@@ -1,11 +1,10 @@
...
 from garage.tf.algos.vpg import VPG
 from garage.tf.optimizers import LbfgsOptimizer
 
 
 class ERWR(VPG):
-    """
-    Episodic Reward Weighted Regression [1].
+    """Episodic Reward Weighted Regression [1].
 
     Notes
```
ryanjulian: please change this to conform to the Google comment style, e.g.

```python
"""
Notes:
    My note here.
"""

garage/tf/algos/dqn.py
zequnyu: No shared base class so I just left `object` here.

ryanjulian: okay

PR #661, opened 2019-05-14
Upgrade mujoco_py to 2.0, gym to 0.12.4, and dm_control to latest

### LINKED ISSUE #634: Upgrade mujoco_py to 2.0
This was released a few weeks ago. It will allow us to get rid of the split versions of MuJoCo in the installer.

### LINKED ISSUE #639: Upgrade gym to 0.12

gitanshu:
Closes #634, closes #639


docker/Dockerfile.base
```diff
@@ -58,18 +58,13 @@ RUN git clone https://github.com/glfw/glfw.git && \
...
-  rm mujoco.zip
-ENV LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/root/.mujoco/mjpro150/bin
-
 # MuJoCo 2.0 (for dm_control)
 RUN mkdir -p /root/.mujoco && \
   wget https://www.roboti.us/download/mujoco200_linux.zip -O mujoco.zip && \
   unzip mujoco.zip -d $HOME/.mujoco && \
-  rm mujoco.zip
+  rm mujoco.zip && \
+  cp -r $HOME/.mujoco/mujoco200_linux $HOME/.mujoco/mujoco200
```
ryanjulian: if `mujoco200_linux` is the new canonical path used by MuJoCo then let's keep that throughout. Lots of libraries may hardcode into that path looking for MuJoCo.

scripts/setup_colab.sh
```diff
@@ -228,7 +219,7 @@ cp "${_arg_mjkey}" "${HOME}/.mujoco/mjkey.txt"
   pip install -e .[all]
   pip install -e .[dev]
   pip install git+https://github.com/rlworkgroup/metaworlds.git@master#egg=metaworlds
```
ryanjulian: please ensure that you also upgrade the packages at rlworkgroup/metaworlds and  that the tests in metaworlds still pass. they should be runnable inside the garage docker/conda environments.

scripts/setup_colab.sh
```diff
@@ -116,7 +116,7 @@ if [[ -z "`command -v file`" ]]; then
   apt-get install -y -q file libmagic-dev
```
ryanjulian: please try this script on colab and verify it works before submitting

scripts/setup_linux.sh
```diff
@@ -169,16 +169,7 @@ if [[ "${_arg_modify_bashrc}" = on ]]; then
   echo -e "\n# Added by the garage installer" >> "${BASH_RC}"
```
ryanjulian: please try this script on a clean machine (or VM) and verify it works before submitting

scripts/setup_macos.sh
```diff
@@ -192,16 +192,7 @@ if [[ "${_arg_modify_bashrc}" = on ]]; then
   echo -e "\n# Added by the garage installer" >> "${BASH_PROF}"
```
ryanjulian: please try this script on the cleanest mac you can find (or VM) and verify it works before submitting

setup.py
```diff
@@ -15,12 +15,12 @@
     # command again and the install succeeds because absl-py has been
     # installed. This is stupid, but harmless.
     'dm_control @ https://api.github.com/repos/deepmind/dm_control/tarball/92f9913013face0468442cd0964d5973ea2089ea',  # noqa: E501
-    'gym[all]==0.10.11',
+    'gym[all]==0.12.0',
```
ryanjulian: why not 0.12.1?

scripts/setup_colab.sh
```diff
@@ -217,8 +208,8 @@ cp "${_arg_mjkey}" "${HOME}/.mujoco/mjkey.txt"
   MUJOCO_150="$(mktemp -d)/mujoco_150"
   git clone --branch 1.50.1.1 --depth 1 https://github.com/openai/mujoco-py.git $MUJOCO_150
```
zhanpenghe: I think this is still using old version of mujoco_py.

docker/Dockerfile.base
gitanshu: I think dm_control was complaining about not finding mujoco200, and gym wanted mujoco200_linux, so i kept both around for now before i test on a clean install. Also, this PR is still a WIP as the title suggests, submitted it just to see the tests on CI

scripts/setup_colab.sh
gitanshu: Great catch. Missed it probably because everywhere else it was called mjpro150. Will fix this

setup.py
gitanshu: 0.12.1 was throwing off all the tests because of a change in the way a property was exposed. Still looking into it. Will do that before submitting

setup.py
```diff
@@ -16,11 +16,11 @@
     # installed. This is stupid, but harmless.
     'dm_control @ https://api.github.com/repos/deepmind/dm_control/tarball/92f9913013face0468442cd0964d5973ea2089ea',  # noqa: E501
```
naeioi: I don't remember why we locked dm_control. Can you try the latest version and see if it fails?

ryanjulian: dm_control doesn't publish on pip/PyPI (last I checked), so the only way to retrieve it is using git. Pinning to master is a recipe for disaster, so I used the latest commit hash (at the time) instead.

setup.py
ryanjulian: i think they changed from using properties to member variables for `observation_space` and `action_space`. this can create problems when an inheriting class tries to use a property.

tldr: you will probably have to change some properties to member variables in garage

ryanjulian: we might as well just do it now

setup.py
gitanshu: Yeah they don't publish. Should I try bringing it to the latest commit then?

docker/Dockerfile.base
gitanshu: I mixed them up. Mujoco-py is the one that looks for mujoco200 while the canonical name is mujoco200_<linux/macos>. So i ended up creating a symlink mujoco200 to mujoco200_linux

scripts/setup_colab.sh
gitanshu: Tested on colab and fixed. Works now.
https://colab.research.google.com/drive/1O_hDHDJqxkp2L1Lw4sHknKGFRgne20nE

scripts/setup_linux.sh
gitanshu: Done

ryanjulian:
@gitanshu is this ready?

gitanshu:
> @gitanshu is this ready?

@ryanjulian Haven't been able to successfully run the setup script on a mac 10.13.6, even from master branch. Works fine on 10.14.4 & .5. Everything else is ready.

ryanjulian:
what error are you getting on 10.13.6?

setup.py
ryanjulian: yes

gitanshu:
@ryanjulian it pretty much fails to build mujoco-py without much info 

[a paste of the log is here](https://pastebin.com/R97aW7Nz)

Stepping through the build script narrowed it down to [this line](https://github.com/openai/mujoco-py/blob/1aeb02e2df429860aaba6783d61fc382c58bf885/mujoco_py/builder.py#L124) where ipdb exited with illegal instruction

ryanjulian:
your log from OSX 10.13 is still using gcc8 to compile. can you try installing gcc@7 or gcc@6 for 10.13? you might need different GCC version for different OSX versions.

gitanshu:
I tried both gcc@7 and gcc@6 yesterday. Same error

ryanjulian:
can you send error logs for the versions using gcc@7 and gcc@6?

gitanshu:
[gcc-6 logs](https://pastebin.com/4tGLbjJ2)

[gcc-7 logs](https://pastebin.com/NAZchGEm)

scripts/setup_macos.sh
```diff
@@ -192,29 +194,21 @@ if [[ "${_arg_modify_bashrc}" = on ]]; then
...
-if [[ ! -d "${HOME}/.mujoco/mjpro150" ]]; then
-  mkdir -p "${HOME}"/.mujoco
-  MUJOCO_ZIP="$(mktemp -d)/mujoco.zip"
-  wget https://www.roboti.us/download/mjpro150_osx.zip -O "${MUJOCO_ZIP}"
-  unzip -u "${MUJOCO_ZIP}" -d "${HOME}"/.mujoco
-else
-  print_warning "MuJoCo is already installed"
-fi
-# Set up MuJoCo 2.0 (for dm_control)
+# Set up MuJoCo 2.0 (for gym and dm_control)
```
ryanjulian: line 162 `xcode-select --install` fails if command line tools are already installed.

ryanjulian: i think you can use the same `set +e`/`set -e` trick used during brew install to fix this

ryanjulian:
FYI gym got bumped up a couple versions over the weekend.

ryanjulian:
@gitanshu i tracked this down to AVX/AVX2 instructions not being exposed to some virtualbox VMs -- MuJoCo requires AVX instructions to compile. some virtualbox VMs have a flaw where AVX instructions are not exposed to the guest, even if the host has them.

my laptop runs 10.13.6, so i will run your updated script on it to see if it works.

ryanjulian:
@gitanshu this installed fine on my macOS 10.13.6 machine with gcc@8, so i think you are in the clear

setup.py
ryanjulian: does it work?

setup.py
```diff
@@ -16,11 +16,11 @@
...
     'dowel==0.0.2',
-    'gym[all]==0.10.11',
+    'gym[all]==0.12.1',
```
ryanjulian: they are now on 0.12.4 https://github.com/openai/gym/releases/tag/0.12.4

ryanjulian:
make sure to run the tests marked "nightly" manually -- these are tests which make sure dm_control and gym still work well with garage, but they take a long time so they are not run in the CI

This should do the trick
```sh
make run-headless RUN_CMD="make ci-job-nightly"
```

gitanshu:
> make sure to run the tests marked "nightly" manually -- these are tests which make sure dm_control and gym still work well with garage, but they take a long time so they are not run in the CI
> 
> This should do the trick
> 
> ```shell
> make run-headless RUN_CMD="make ci-job-nightly"
> ```

Done. All tests are working fine.

PR #662, opened 2019-05-15
Refactor DeterministicMLP_ to use garage.tf.Model

ahtsan:
This PR refactors `DeterministicMLPBaseline` and `DeterministicMLPRegressor` with `garage.tf.Model`. Tests are added accordingly. It is also tested with `PPO` to make sure it works with existing algorithms.

### DIFF
--- a/garage/tf/baselines/__init__.py
+++ b/garage/tf/baselines/__init__.py
@@ -1,16 +1,15 @@
 """Baseline estimators for TensorFlow-based algorithms"""
+from garage.tf.baselines.continuous_mlp_baseline import ContinuousMLPBaseline
 from garage.tf.baselines.continuous_mlp_baseline_with_model import (
     ContinuousMLPBaselineWithModel)
-from garage.tf.baselines.deterministic_mlp_baseline import (
-    DeterministicMLPBaseline)
 from garage.tf.baselines.gaussian_conv_baseline import GaussianConvBaseline
 from garage.tf.baselines.gaussian_mlp_baseline import GaussianMLPBaseline
 from garage.tf.baselines.gaussian_mlp_baseline_with_model import (
     GaussianMLPBaselineWithModel)
 
 __all__ = [
+    'ContinuousMLPBaseline',
     'ContinuousMLPBaselineWithModel',
-    'DeterministicMLPBaseline',
     'GaussianConvBaseline',
     'GaussianMLPBaseline',
     'GaussianMLPBaselineWithModel',
--- a/garage/tf/baselines/__init__.py
+++ b/garage/tf/baselines/__init__.py
@@ -1,16 +1,16 @@
 """Baseline estimators for TensorFlow-based algorithms"""
+from garage.tf.baselines.continuous_mlp_baseline_with_model import (
+    ContinuousMLPBaselineWithModel)
 from garage.tf.baselines.deterministic_mlp_baseline import (
     DeterministicMLPBaseline)
-from garage.tf.baselines.deterministic_mlp_baseline_with_model import (
-    DeterministicMLPBaselineWithModel)
 from garage.tf.baselines.gaussian_conv_baseline import GaussianConvBaseline
 from garage.tf.baselines.gaussian_mlp_baseline import GaussianMLPBaseline
 from garage.tf.baselines.gaussian_mlp_baseline_with_model import (
     GaussianMLPBaselineWithModel)
 
 __all__ = [
+    'ContinuousMLPBaselineWithModel',
     'DeterministicMLPBaseline',
-    'DeterministicMLPBaselineWithModel',
     'GaussianConvBaseline',
     'GaussianMLPBaseline',
     'GaussianMLPBaselineWithModel',
--- a/garage/tf/baselines/__init__.py
+++ b/garage/tf/baselines/__init__.py
@@ -1,6 +1,8 @@
 """Baseline estimators for TensorFlow-based algorithms"""
 from garage.tf.baselines.deterministic_mlp_baseline import (
     DeterministicMLPBaseline)
+from garage.tf.baselines.deterministic_mlp_baseline_with_model import (
+    DeterministicMLPBaselineWithModel)
 from garage.tf.baselines.gaussian_conv_baseline import GaussianConvBaseline
 from garage.tf.baselines.gaussian_mlp_baseline import GaussianMLPBaseline
 from garage.tf.baselines.gaussian_mlp_baseline_with_model import (
@@ -8,6 +10,7 @@ from garage.tf.baselines.gaussian_mlp_baseline_with_model import (
 
 __all__ = [
     'DeterministicMLPBaseline',
+    'DeterministicMLPBaselineWithModel',
     'GaussianConvBaseline',
     'GaussianMLPBaseline',
     'GaussianMLPBaselineWithModel',
--- a/garage/tf/baselines/deterministic_mlp_baseline_with_model.py
+++ b/garage/tf/baselines/deterministic_mlp_baseline_with_model.py
@@ -1,0 +1,71 @@
+"""A value function (baseline) based on a GaussianMLP model."""
+import numpy as np
+
+from garage.misc.overrides import overrides
+from garage.np.baselines import Baseline
+from garage.tf.regressors import DeterministicMLPRegressorWithModel
+
+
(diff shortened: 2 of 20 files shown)


garage/tf/regressors/deterministic_mlp_regressor_model.py
```diff
@@ -0,0 +1,72 @@
+"""DeterministicMLPRegressorModel."""
+import numpy as np
+import tensorflow as tf
+
+from garage.tf.models import MLPModel
+
+
+class DeterministicMLPRegressorModel(MLPModel):
```
ryanjulian: isn't this more accurately `NormalizedInputMLPModel`? I bet you could also make an outer model `NormalizeInputs` which wraps any model and does this.

garage/tf/regressors/deterministic_mlp_regressor_with_model.py
```diff
@@ -0,0 +1,172 @@
+"""A regressor based on a GaussianMLP model."""
```
ryanjulian: based on...?

garage/tf/baselines/deterministic_mlp_baseline_with_model.py
```diff
@@ -0,0 +1,71 @@
+"""A value function (baseline) based on a GaussianMLP model."""
+import numpy as np
+
+from garage.misc.overrides import overrides
+from garage.np.baselines import Baseline
+from garage.tf.regressors import DeterministicMLPRegressorWithModel
```
ryanjulian: do the Deterministic* primitives all work with both continuous and discrete data?

I ask because Gaussian/Categorical/DeterministicThing all make sense as names, unless there needs to be two different deterministic primitives (discrete and continuous). The naming scheme so far is not very consistent. I'd like to pick one going forward and stick with it.

garage/tf/regressors/deterministic_mlp_regressor_with_model.py
```diff
@@ -0,0 +1,172 @@
...
+        tabular.record('{}/dLoss'.format(self._name), loss_before - loss_after)
+
+    def predict(self, xs):
+        """
+        Predict y based on input xs.
+
+        Args:
+            xs (numpy.ndarray): Input data.
+
+        Returns:
```
ryanjulian: missing returns

garage/tf/regressors/deterministic_mlp_regressor_with_model.py
```diff
@@ -0,0 +1,172 @@
...
+        """
+        return self._f_predict(xs)
+
+    def predict_sym(self, xs, name=None):
+        """
+        Symbolic graph of the prediction.
+
+        Args:
+            xs (tf.Tensor): Input tf.Tensor for the input data.
+            name (str): Name of the new graph.
```
ryanjulian: missing returns

garage/tf/regressors/deterministic_mlp_regressor_model.py
```diff
@@ -0,0 +1,72 @@
...
+        output_b_init (callable): Initializer function for the bias
+            of output dense layer(s). The function should return a
+            tf.Tensor.
+    """
+
+    def __init__(self,
+                 input_shape,
+                 output_dim,
+                 name='DeterministicMLPRegressorModel',
+                 **kwargs):
```
ryanjulian: please don't capture **kwargs unless you really need to.

garage/tf/regressors/deterministic_mlp_regressor_model.py
ahtsan: I agree with the naming. However, the `network_input_spec` and `network_output_spec` for different model is different. We can't have a single model `NormalizeInputs` which represents all the model with normalized inputs since their inputs/outputs are different. Would it be too complicated if we create a base model `NormalizeInputs` and have each model inherit from it and add specific `network_input_spec` and `network_output_spec`?

garage/tf/baselines/deterministic_mlp_baseline_with_model.py
ahtsan: I think a `MLP` would always be continuous (since output value is a real value), and if we wrap the `MLP` in a primitive to discretize it, we would call that discrete (e.g. `DiscreteMLPPolicy`/`DiscreteMLPQFunction`). I agree we should not have `Deterministic` and have `Discrete` and `Continuous` instead.

ryanjulian: alright as long as you're thinking about it. just pick a rule and be consistent going forward. what are you going to name things in this PR?

ahtsan: This PR refactor the existing `DeterministicMLPBaseline` which is continuous, I think we should call it `ContinuousMLPBaselineWithModel`?

ryanjulian: SGTM

garage/tf/regressors/deterministic_mlp_regressor_model.py
ryanjulian: SGTM

garage/tf/baselines/deterministic_mlp_baseline_with_model.py
CatherineSue: I agree Continuous is a better name

garage/tf/regressors/deterministic_mlp_regressor_model.py
nish21: +1 on `NormalizedInputMLPModel`. I was refactoring `CategoricalMLPRegressor` and it uses the exact same subclass as this.

tests/garage/tf/baselines/test_deterministic_mlp_baseline_with_model.py
```diff
@@ -0,0 +1,97 @@
...
+
+class TestDeterministicMLPBaselineWithModel(TfGraphTestCase):
+    @params([1], [2], [1, 1], [2, 2])
+    def test_fit(self, obs_dim):
+        box_env = TfEnv(DummyBoxEnv(obs_dim=obs_dim))
+        with mock.patch(('garage.tf.baselines.'
+                         'deterministic_mlp_baseline_with_model.'
+                         'DeterministicMLPRegressorWithModel'),
+                        new=SimpleMLPRegressor):
+            gmb = DeterministicMLPBaselineWithModel(env_spec=box_env.spec)
```
CatherineSue: minor thing, is `dmb` a better name, I guess `gmb` stands for GaussianMLPBaseline?

tests/garage/tf/regressors/test_deterministic_mlp_regressor_with_model.py
```diff
@@ -0,0 +1,105 @@
...
+import tensorflow as tf
+
+from garage.tf.optimizers import LbfgsOptimizer
+from garage.tf.regressors import DeterministicMLPRegressorWithModel
+from tests.fixtures import TfGraphTestCase
+
+
+class TestDeterministicMLPRegressorWithModel(TfGraphTestCase):
+    def test_fit_normalized(self):
+        gmr = DeterministicMLPRegressorWithModel(
```
CatherineSue: still a minor comment, dmr?

garage/tf/regressors/deterministic_mlp_regressor_with_model.py
CatherineSue: Please remember to address this.

tests/garage/tf/baselines/test_deterministic_mlp_baseline_with_model.py
ahtsan: Yes, I will change it to `cmb`.

garage/tf/baselines/__init__.py
```diff
@@ -7,6 +9,7 @@
     GaussianMLPBaselineWithModel)
 
 __all__ = [
+    'ContinuousMLPBaselineWithModel',
     'DeterministicMLPBaseline',
```
ryanjulian: is the new name for this one `ContinuousMLPBaseline` or `DiscreteMLPBaseline`?

ahtsan: This is the new name for the original `DiscreteMLPBaseline` because there was no `ContinuousMLPBaseline` in garage before.

PR #665, opened 2019-05-15
Move garage package to src/

ryanjulian:
This simplifies packaging for pip, and complies with Python packaging
best practices.


krzentner:
It's probably best to choose a time to submit this when no one else has any code in flight, since it's likely to wreck every other pull request.

PR #667, opened 2019-05-16
Test fix test_stateful_pool

zequnyu:
As discussed with Ryan and K.R. this is only the fix for the test.


naeioi:
Why does this test fail anyway?

krzentner:
> Why does this test fail anyway?

Basically, `StatefullPool.run_collect` keeps calling a function in a pool of workers, until it has collected `threshold` results. However, there may be multiple workers that are still running when `threshold` is reached, and they will all submit results.

Although this is technically a race condition, the documentation doesn't say that this won't happen, and the rest of the system handles it, so there's no need to fix it right now.

PR #668, opened 2019-05-17
Refactor GRU with garage.tf.models.Model

ahtsan:
Added GRU, GRUModel and CategoricalGRUPolicyWithModel.
It aims to replace the existing GRUNetwork and GRULayer,
which are based on garage.tf.core.layers.

Added test for TRPO with CategoricalGRUPolicyWithModel.

Apart from testing functionality of CategoricalGRUPolicyWithModel
in test_categorical_gru_policy_with_model.py, transitions from the
old model (CategoricalGRUPolicy) to the new model
(CategoricalGRUPolicyWithModel) are also tested in
test_categorical_gru_policy_with_model_transit.py, to make sure
they have the same API.

Existing GRU implementation in GRULayer is not exactly the same as
TensorFlow implementation (from original paper), and is modified in
this PR.


tests/garage/tf/core/test_gru.py
```diff
@@ -0,0 +1,384 @@
...
+        (1, 3, 1, 0, 0),        # yapf: disable
+        (3, 1, 1, 0, 0),        # yapf: disable
+        (3, 3, 1, 0, 0),        # yapf: disable
+        (3, 3, 3, 0, 0),        # yapf: disable
+        (1, 1, 1, 0.5, 0.5),    # yapf: disable
+        (1, 1, 3, 0.5, 0.5),    # yapf: disable
+        (1, 3, 1, 0.5, 0.5),    # yapf: disable
+        (3, 1, 1, 0.5, 0.5),    # yapf: disable
+        (3, 3, 1, 0.5, 0.5),    # yapf: disable
+        (3, 3, 3, 0.5, 0.5))    # yapf: disable
```
ryanjulian: i think you should be able to just put one yapf_disable right after the `@params` statement and it should ignore the whole block.

if it doesn't, you can use this form
```python
# yapf: disable
stuff()
more_stuff()
# yapf: enable
```

tests/garage/tf/core/test_gru.py
```diff
@@ -0,0 +1,384 @@
...
+    def test_output_shapes(self, time_step, input_dim, output_dim, hidden_init,
+                           cell_init):
+        obs_inputs = np.full((self.batch_size, time_step, input_dim), 1.)
+        obs_input = np.full((self.batch_size, input_dim), 1.)
+
+        _input_var = tf.placeholder(
```
ryanjulian: why the _private?

ahtsan: Not sure. I will remove the `_`

tests/garage/tf/core/test_gru.py
```diff
@@ -0,0 +1,390 @@
...
+    # yapf: enable
+    def test_output_shapes(self, time_step, input_dim, output_dim, hidden_init,
+                           cell_init):
+        obs_inputs = np.full((self.batch_size, time_step, input_dim), 1.)
+        obs_input = np.full((self.batch_size, input_dim), 1.)
+
+        input_var = tf.placeholder(
+            tf.float32, shape=(None, None, input_dim), name='input')
+        step_input_var = tf.placeholder(
+            tf.float32, shape=(None, input_dim), name='input')
```
CatherineSue: Why do the two placeholders have the same name?

garage/tf/core/layers.py
```diff
@@ -1052,7 +1052,7 @@ class GRULayer(Layer):
     Reset gate:        r(t) = f_r(x(t) @ W_xr + h(t-1) @ W_hr + b_r)
     Update gate:       u(t) = f_u(x(t) @ W_xu + h(t-1) @ W_hu + b_u)
     Cell gate:         c(t) = f_c(x(t) @ W_xc + r(t) * (h(t-1) @ W_hc) + b_c)
-    New hidden state:  h(t) = (1 - u(t)) * h(t-1) + u_t * c(t)
+    New hidden state:  h(t) = (1 - u(t)) * c(t) + u_t * h(t-1)
```
CatherineSue: I think the original update is correct?

garage/tf/core/gru.py
```diff
@@ -0,0 +1,62 @@
...
+        hidden_state_init=tf.zeros_initializer(),
+        hidden_state_init_trainable=False):
+    """
+    Gated Recurrent Unit (GRU).
+
+    Args:
+        name (str): Name of the variable scope.
+        gru_cell (tf.keras.layers.Layer): GRU cell used to generate
+            outputs.
+        all_input_var (tf.Tensor): Place holder for entire time-seried inputs.
```
CatherineSue: time-series

garage/tf/core/gru.py
```diff
@@ -0,0 +1,62 @@
...
+        output_nonlinearity (callable): Activation function for output dense
+            layer. It should return a tf.Tensor. Set it to None to
+            maintain a linear activation.
+        hidden_state_init (callable): Initializer function for the
+            initial hidden state. The functino should return a tf.Tensor.
+        hidden_state_init_trainable (bool): Bool for whether the initial
+            hidden state is trainable.
+
+    Return:
+        outputs (tf.Tensor): Entire time-seried outputs.
```
CatherineSue: do you mean time-series? please ignore if time-seried is a word.

tests/garage/tf/core/test_gru.py
```diff
@@ -0,0 +1,390 @@
...
+            assert output.shape == (self.batch_size, output_dim)
+            assert hidden.shape == (self.batch_size, self.hidden_dim)
+
+        full_output = self.sess.run(
+            outputs_t, feed_dict={input_var: obs_inputs})
+
+        assert full_output.shape == (self.batch_size, time_step, output_dim)
+
+    # yapf: disable
+    @params(
```
CatherineSue: Could you add a test of variable length inputs to GRU and LSTM? It seems we don't have any tests using off-policy algorithms with recurrent policies. And since DQN doesn't set a max_path_length, I assume the episodes would have variable lengths. For recurrent policies, we should pad them. Do we pad it somewhere? Please point it to me if we do.

garage/tf/core/layers.py
ahtsan: Original update is not the same as the origin paper.

CatherineSue: Please check? I think the old one is consistent with the paper and the wikipedia.

ahtsan: emm I think so. TensorFlow is not, in this case. Since we are using TensorFlow GRUCell, I wanted to make the existing and new `GRU` give the same output. Right now they are like inverse of each other, one of them is learning `x` while another one is learning `1-x`.

ahtsan: If we don't have to care about making them having the same output, I will stick with the original implementation.

CatherineSue: They are both okay to me and they can both be seen as the original paper. They used the tf notation in 1406.1078v3 then used our old notation in 1412.3555v1. If we need to make sure the new one and the old one has the same implementation, we should probably be consistent with tf implementation. @ryanjulian What do you think?

ahtsan: I talked with Ryan and he agreed we should stick with tf implementation.

tests/garage/tf/core/test_gru.py
ahtsan: If we pad the inputs, then we don't have to worry about variable length inputs?

tests/garage/tf/core/test_gru.py
ahtsan: Will fix

garage/tf/core/gru.py
ahtsan: Will fix

garage/tf/core/gru.py
ahtsan: Will fix

tests/garage/tf/core/test_gru.py
CatherineSue: Sorry just saw this. 
I think I might have used an incorrect example. It is uncommon to use LSTMNetwork in DQN. Besides, during training, the shape of samples (from replay buffer) would be `(sample_size, 1, obs_dim)`. During sampling, the shape would be `(1, time_steps, obs_dim)`. So it might not be a problem in the current off-policy algorithms. 
If we pad them, we just need to indicate that `<pad>` is a condition to stop the traverse. `tf.while_loop` or `tf.scan`(what we are using) should have an argument to pass the condition (please check, I am not familiar with `tf.scan`).

CatherineSue: I think this problem seems to beyond the pr's scope. If you think we should address this later or we don't have the situation to deal with variable-length inputs, feel free to ignore this issue.

PR #672, opened 2019-05-20
Refactor CategoricalMLPRegressor to use models

nish21:
Refactor the `CategoricalMLPRegressor` module to use `garage.tf.models`.

Added unit tests for the module `CategoricalMLPRegressorWithModel`.

### DIFF
--- a/garage/tf/regressors/categorical_mlp_regressor_with_model.py
+++ b/garage/tf/regressors/categorical_mlp_regressor_with_model.py
@@ -1,0 +1,255 @@
+"""A regressor based on MLP with Normalized Inputs."""
+from dowel import tabular
+import numpy as np
+import tensorflow as tf
+
+from garage.tf.distributions import Categorical
+from garage.tf.misc import tensor_utils
+from garage.tf.models import NormalizedInputMLPModel
+from garage.tf.optimizers import ConjugateGradientOptimizer, LbfgsOptimizer
+from garage.tf.regressors import StochasticRegressor2
+
+
+class CategoricalMLPRegressorWithModel(StochasticRegressor2):
+    """
+    CategoricalMLPRegressor with garage.tf.models.NormalizedInputMLPModel.
+
+    A class for performing regression (or classification, really) by fitting
+    a Categorical distribution to the outputs. Assumes that the output will
+    always be a one hot vector
+
+    Args:
+        input_shape (tuple[int]): Input shape of the training data.
+        output_dim (int): Output dimension of the model.
+        name (str): Model name, also the variable scope.
+        hidden_sizes (list[int]): Output dimension of dense layer(s) for
+            the MLP for the network. For example, (32, 32) means the MLP
+            consists of two hidden layers, each with 32 hidden units.
+        hidden_nonlinearity (callable): Activation function for intermediate
+            dense layer(s). It should return a tf.Tensor. Set it to
+            None to maintain a tanh activation.
+        hidden_w_init (callable): Initializer function for the weight
+            of intermediate dense layer(s). The function should return a
+            tf.Tensor. Default is Glorot uniform initializer.
+        hidden_b_init (callable): Initializer function for the bias
+            of intermediate dense layer(s). The function should return a
+            tf.Tensor. Default is zero initializer.
+        output_nonlinearity (callable): Activation function for output dense
+            layer. It should return a tf.Tensor. Set it to None to
+            maintain a softmax activation.
+        output_w_init (callable): Initializer function for the weight
+            of output dense layer(s). The function should return a
+            tf.Tensor. Default is Glorot uniform initializer.
+        output_b_init (callable): Initializer function for the bias
+            of output dense layer(s). The function should return a
+            tf.Tensor. Default is zero initializer.
+        optimizer (tf.Optimizer): Optimizer for minimizing the negative
+            log-likelihood.
+        optimizer_args (dict): Arguments for the optimizer. Default is None,
+            which means no arguments.
+        tr_optimizer (tf.Optimizer): Optimizer for trust region approximation.
+        tr_optimizer_args (dict): Arguments for the trust region optimizer.
+            Default is None, which means no arguments.
+        use_trust_region (bool): Whether to use trust region constraint.
+        max_kl_step (float): KL divergence constraint for each iteration.
+        normalize_inputs (bool): Bool for normalizing inputs or not.
+        layer_normalization (bool): Bool for using layer normalization or not.
+        no_initial_trust_region (bool): Bool to specify if trust region
+            constraint has been used initially.
+    """
+
+    def __init__(self,
+                 input_shape,
+                 output_dim,
+                 name='CategoricalMLPRegressorWithModel',
+                 hidden_sizes=(32, 32),
+                 hidden_nonlinearity=tf.nn.tanh,
+                 hidden_w_init=tf.glorot_uniform_initializer(),
+                 hidden_b_init=tf.zeros_initializer(),
+                 output_nonlinearity=tf.nn.softmax,
+                 output_w_init=tf.glorot_uniform_initializer(),
+                 output_b_init=tf.zeros_initializer(),
+                 optimizer=None,
+                 optimizer_args=None,
+                 tr_optimizer=None,
+                 tr_optimizer_args=None,
+                 use_trust_region=True,
+                 max_kl_step=0.01,
+                 normalize_inputs=True,
+                 layer_normalization=False,
+                 no_initial_trust_region=True):
+
+        super().__init__(input_shape, output_dim, name)
+        self._use_trust_region = use_trust_region
+        self._no_initial_trust_region = no_initial_trust_region
+        self._max_kl_step = max_kl_step
+        self._normalize_inputs = normalize_inputs
+        self._first_optimized = not no_initial_trust_region
+
+        with tf.variable_scope(
+                self._name, reuse=False) as self._variable_scope:
+            if optimizer is None:
+                optimizer = LbfgsOptimizer()
+            if tr_optimizer is None:
+                tr_optimizer = ConjugateGradientOptimizer()
+            self._optimizer = optimizer
+            self._tr_optimizer = tr_optimizer
+
+        self.model = NormalizedInputMLPModel(
+            input_shape,
+            output_dim,
+            hidden_sizes=hidden_sizes,
+            hidden_nonlinearity=hidden_nonlinearity,
+            hidden_w_init=hidden_w_init,
+            hidden_b_init=hidden_b_init,
+            output_nonlinearity=output_nonlinearity,
+            output_w_init=output_w_init,
+            output_b_init=output_b_init,
+            layer_normalization=layer_normalization)
+
+        self._initialize()
+
+    def _initialize(self):
+        input_var = tf.placeholder(
+            tf.float32, shape=(None, ) + self._input_shape)
+
+        with tf.variable_scope(self._variable_scope):
+            self.model.build(input_var)
+
+            ys_var = tf.placeholder(
+                dtype=tf.float32, name='ys', shape=(None, self._output_dim))
+
+            old_prob_var = tf.placeholder(
+                dtype=tf.float32,
+                name='old_prob',
+                shape=(None, self._output_dim))
+
+            y_hat = self.model.networks['default'].y_hat
+
+            old_info_vars = dict(prob=old_prob_var)
+            info_vars = dict(prob=y_hat)
+
+            self._dist = Categorical(self._output_dim)
+            mean_kl = tf.reduce_mean(
+                self._dist.kl_sym(old_info_vars, info_vars))
+
+            loss = -tf.reduce_mean(
+                self._dist.log_likelihood_sym(ys_var, info_vars))
+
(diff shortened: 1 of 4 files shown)


garage/tf/regressors/categorical_mlp_regressor_with_model.py
```diff
@@ -0,0 +1,255 @@
...
+            of output dense layer(s). The function should return a
+            tf.Tensor. Default is Glorot uniform initializer.
+        output_b_init (callable): Initializer function for the bias
+            of output dense layer(s). The function should return a
+            tf.Tensor. Default is zero initializer.
+        optimizer (tf.Optimizer): Optimizer for minimizing the negative
+            log-likelihood.
+        optimizer_args (dict): Arguments for the optimizer. Default is None,
+            which means no arguments.
+        tr_optimizer (tf.Optimizer): Optimizer for trust region approximation.
```
ryanjulian: is this actually a tf.Optimizer or is it a garage optimizer?

garage/tf/regressors/categorical_mlp_regressor_with_model.py
```diff
@@ -0,0 +1,255 @@
...
+        output_nonlinearity (callable): Activation function for output dense
+            layer. It should return a tf.Tensor. Set it to None to
+            maintain a softmax activation.
+        output_w_init (callable): Initializer function for the weight
+            of output dense layer(s). The function should return a
+            tf.Tensor. Default is Glorot uniform initializer.
+        output_b_init (callable): Initializer function for the bias
+            of output dense layer(s). The function should return a
+            tf.Tensor. Default is zero initializer.
+        optimizer (tf.Optimizer): Optimizer for minimizing the negative
```
ryanjulian: same question as below

garage/tf/regressors/categorical_mlp_regressor_with_model.py
```diff
@@ -0,0 +1,255 @@
...
+            info_vars = dict(prob=y_hat)
+
+            self._dist = Categorical(self._output_dim)
+            mean_kl = tf.reduce_mean(
+                self._dist.kl_sym(old_info_vars, info_vars))
+
+            loss = -tf.reduce_mean(
+                self._dist.log_likelihood_sym(ys_var, info_vars))
+
+            predicted = tensor_utils.to_onehot_sym(
```
ryanjulian: isn't there a TF API for this?

garage/tf/regressors/categorical_mlp_regressor_with_model.py
```diff
@@ -0,0 +1,255 @@
...
+
+        Args:
+            xs (numpy.ndarray): Input data.
+            ys (numpy.ndarray): Input labels in one hot representation.
+
+        Return:
+            The predicted log likelihoods.
+
+        """
+        prob = self._f_prob(np.asarray(xs))
```
ryanjulian: why np.asarray? you already said xs is an np.ndarray

garage/tf/regressors/categorical_mlp_regressor_with_model.py
CatherineSue: I think it should be a garage.tf.Optimizer? Is there a tf second order optimizer?

tests/garage/tf/regressors/test_categorical_mlp_regressor_with_model.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+def get_labels(input):
+    label = [0, 0]
+    label[0 if np.sin(input) <= 0 else 1] = 1
+    return label
+
+
+class TestCategoricalMLPRegressorWithModel(TfGraphTestCase):
+    def test_fit_normalized(self):
```
ryanjulian: please also test the fit methods with a range of reasonable configuration values (catch 0-vs-1, 1-vs-2, 2-vs-many errors, etc.)

garage/tf/regressors/categorical_mlp_regressor_with_model.py
nish21: It should be `garage.tf.Optimizer`, will change that.

garage/tf/regressors/categorical_mlp_regressor_with_model.py
```diff
@@ -0,0 +1,255 @@
...
+        optimizer_args (dict): Arguments for the optimizer. Default is None,
+            which means no arguments.
+        tr_optimizer (tf.Optimizer): Optimizer for trust region approximation.
+        tr_optimizer_args (dict): Arguments for the trust region optimizer.
+            Default is None, which means no arguments.
+        use_trust_region (bool): Whether to use trust region constraint.
+        max_kl_step (float): KL divergence constraint for each iteration.
+        normalize_inputs (bool): Bool for normalizing inputs or not.
+        layer_normalization (bool): Bool for using layer normalization or not.
+        no_initial_trust_region (bool): Bool to specify if trust region
```
ryanjulian: i don't understand this explanation

garage/tf/regressors/categorical_mlp_regressor_with_model.py
ahtsan: Can you also update the same thing for `ContinuousMLPRegressorWithModel` and `GaussianMLPRegressorWithModel`?

garage/tf/regressors/categorical_mlp_regressor_with_model.py
ryanjulian: if so, please remove this (also remove it from tensor_utils and replace usages elsewhere with the TF API)

garage/tf/regressors/categorical_mlp_regressor_with_model.py
```diff
@@ -0,0 +1,255 @@
...
+        super().__init__(input_shape, output_dim, name)
+        self._use_trust_region = use_trust_region
+        self._no_initial_trust_region = no_initial_trust_region
+        self._max_kl_step = max_kl_step
+        self._normalize_inputs = normalize_inputs
+        self._first_optimized = not no_initial_trust_region
+
+        with tf.variable_scope(
+                self._name, reuse=False) as self._variable_scope:
+            if optimizer is None:
```
ryanjulian: looks like the default case is missed in tests. do your default optimizers actually work? how do you know?

garage/tf/regressors/categorical_mlp_regressor_with_model.py
```diff
@@ -0,0 +1,255 @@
...
+                optimizer = LbfgsOptimizer()
```
ryanjulian: i can't use optimizer_args if i don't set an optimizer in the constructor?

garage/tf/regressors/categorical_mlp_regressor_with_model.py
nish21: There is a TF API, and this usage is just a wrapper to TF function. Will probably replace it with TF API here and separately address other usages in a different PR.

garage/tf/regressors/categorical_mlp_regressor_with_model.py
```diff
@@ -0,0 +1,255 @@
...
+
+        super().__init__(input_shape, output_dim, name)
+        self._use_trust_region = use_trust_region
+        self._no_initial_trust_region = no_initial_trust_region
+        self._max_kl_step = max_kl_step
+        self._normalize_inputs = normalize_inputs
+        self._first_optimized = not no_initial_trust_region
+
+        with tf.variable_scope(
+                self._name, reuse=False) as self._variable_scope:
```
ahtsan: Avoid assigning the return of the context manager to an instance attribute. Do
```python
with tf.variable_scope(self.name, reuse=False) as vs:
    self._variable_scope = vs
```
Please also update the same thing for `ContinuousMLPRegressorWithModel` and `GaussianMLPRegressorWithModel`.

garage/tf/regressors/categorical_mlp_regressor_with_model.py
nish21: missed that, will update

garage/tf/regressors/categorical_mlp_regressor_with_model.py
nish21: Sure

garage/tf/regressors/categorical_mlp_regressor_with_model.py
nish21: I think the default case is tested but the case where optimizer is explicitly passed to the constructor is not tested. Will address it shortly.

garage/tf/regressors/categorical_mlp_regressor_with_model.py
ryanjulian: lgtm

garage/tf/regressors/categorical_mlp_regressor_with_model.py
```diff
@@ -0,0 +1,262 @@
...
+
+        Return:
+            The predicted log likelihoods.
+
+        """
+        prob = self._f_prob(xs)
+        return self._dist.log_likelihood(ys, dict(prob=prob))
+
+    def dist_info_sym(self, x_var, name=None):
+        """Symbolic graph of the distribution."""
```
ahtsan: docstring missing arguments

tests/garage/tf/regressors/test_categorical_mlp_regressor_with_model.py
```diff
@@ -0,0 +1,244 @@
...
+    elif input_shape == (2, ):
+        paths = {'observations': [[0, 0], [0, 1], [1, 0], [1, 1]]}
+        expected = [[1, 0], [0, 1], [0, 1], [1, 0]]
+
+    return paths, expected
+
+
+class TestCategoricalMLPRegressorWithModel(TfGraphTestCase):
+    @params(((1, ), 2), ((2, ), 2))
+    def test_fit_normalized(self, input_shape, output_dim):
```
ahtsan: What happen if input shape is (1,1)?

nish21: Ryan and I discussed about this. Since the underlying model for this regressor is a MLP, a multi-dimensional input does not make much sense and it is assumed that the input is flattened. Currently none of the regressors implicitly flatten the input. This needs to be noted down somewhere that the input is expected to be of shape (x, ).

garage/tf/regressors/categorical_mlp_regressor_with_model.py
nish21: will change that

tests/garage/tf/regressors/test_categorical_mlp_regressor_with_model.py
ahtsan: I see. I think adding a checking / assert would be better in this case, for all MLP regressors.

nish21: That's a good idea.

ahtsan: Currently all MLPBaselines assume single dimension for input by passing `env_spec.observation_space.flat_dim` as the `input_shape`. However, they do not flatten the inputs in `fit()` and `predict()`. If a MLPBaseline is used in a multi-dimensional environment, would that be a problem? @ryanjulian

ryanjulian: i think it is fine. you just need to make it clear in the docs that the interface assumes flattened inputs. in the future i think we should flatten them automatically.

nish21: Alright, I've updated the docs to include the note on the flattened inputs.

garage/tf/regressors/categorical_mlp_regressor_with_model.py
```diff
@@ -0,0 +1,273 @@
...
+                 optimizer=None,
+                 optimizer_args=None,
+                 tr_optimizer=None,
+                 tr_optimizer_args=None,
+                 use_trust_region=True,
+                 max_kl_step=0.01,
+                 normalize_inputs=True,
+                 layer_normalization=False):
+
+        super().__init__(input_shape, output_dim, name)
```
ahtsan: Maybe it will save us some time in the future, if we add a checking / assert here for input_shape.

ryanjulian: what check, specifically? i guess that it has 2 dimensions only?

ahtsan: Shouldn't it be single dimension? Since the inputs are flattened, like `len(input_shape) == 1`?

ryanjulian: does it support a batch dimension?

ahtsan: I think we usually do `shape=(None, ) + input_shape` like at line 123, so batch dimension is expected.

nish21: Yes, it supports batch dimension (also tested in tests), restriction is just that individual data points should be one dimensional

ryanjulian: okay -- we can improve the interface for primitives once we get the models transition done.

Issue #654, opened 2019-05-13
Master branch can't pass make test


zequnyu:
Current master branch can't pass `make test`. However the failed tests will pass when running unittest separately.

```
======================================================================
ERROR: test_dm_control_tf_policy (tests.garage.envs.dm_control.test_dm_control_tf_policy.TestDmControlTfPolicy)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/root/code/garage/tests/garage/envs/dm_control/test_dm_control_tf_policy.py", line 38, in test_dm_control_tf_policy
    runner.train(n_epochs=1, batch_size=10)
  File "/root/code/garage/garage/experiment/local_tf_runner.py", line 321, in train
    start_epoch=0)
  File "/root/code/garage/garage/experiment/local_tf_runner.py", line 407, in _train
    self.save(epoch, paths if store_paths else None)
  File "/root/code/garage/garage/experiment/local_tf_runner.py", line 210, in save
    snapshotter.save_itr_params(epoch, params)
  File "/root/code/garage/garage/logger/snapshotter.py", line 85, in save_itr_params
    with open(file_name, 'wb') as file:
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmpgplyc983/params.pkl'

======================================================================
ERROR: test_cem_cartpole (tests.garage.np.algos.test_cem.TestCEM)
Test CEM with Cartpole-v1 environment.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/root/code/garage/tests/garage/np/algos/test_cem.py", line 35, in test_cem_cartpole
    n_epochs=5, batch_size=2000, n_epoch_cycles=n_samples)
  File "/root/code/garage/garage/experiment/local_tf_runner.py", line 321, in train
    start_epoch=0)
  File "/root/code/garage/garage/experiment/local_tf_runner.py", line 407, in _train
    self.save(epoch, paths if store_paths else None)
  File "/root/code/garage/garage/experiment/local_tf_runner.py", line 210, in save
    snapshotter.save_itr_params(epoch, params)
  File "/root/code/garage/garage/logger/snapshotter.py", line 85, in save_itr_params
    with open(file_name, 'wb') as file:
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmpgplyc983/params.pkl'

======================================================================
ERROR: test_cma_es_cartpole (tests.garage.np.algos.test_cma_es.TestCMAES)
[...]

PR #674, opened 2019-05-22
Fix `make test`

zequnyu:
This PR will fix [#654](https://github.com/rlworkgroup/garage/issues/654).

### DIFF
--- a/tests/fixtures/fixtures.py
+++ b/tests/fixtures/fixtures.py
@@ -4,7 +4,7 @@ import unittest
 from dowel import logger
 import tensorflow as tf
 
-from garage.experiment import deterministic
+from garage.experiment import deterministic, snapshotter
 from tests.fixtures.logger import NullOutput
 
 
@@ -34,6 +34,8 @@ class TfGraphTestCase(unittest.TestCase):
         if tf.get_default_session() is self.sess:
             self.sess.__exit__(None, None, None)
         self.sess.close()
+
+        snapshotter.reset()
         # These del are crucial to prevent ENOMEM in the CI
         # b/c TensorFlow does not release memory explicitly
         del self.graph


ryanjulian:
Did you determine why this failed in `make test` but not in the CI?

zequnyu:
> Did you determine why this failed in `make test` but not in the CI?

My guess is that because the order of tests is different on different machine (which explains the difference errors info from me and gitanshu). If test_resume.py runs after the algorithms tests, it won't throw errors.

Is there any way to test in the CI without push a commit? @ryanjulian

ryanjulian:
@zequnyu you can got to travisci.com and trigger a build by choosing your branch.

zequnyu:
@ryanjulian Just checked the CI tests order `test_resume` (test 654) was the very latter one. So CI passed because of the lucky order.

ryanjulian:
awesome thanks for checking.

is there a way we could write the tests so that it fails in the CI regardless of order?

zequnyu:
Our unit tests should be independent to each others. So the order should not be a matter. The key problem I think is the design of `shapshotter`. Even if we want a global `snapshotter`, instantiating it inside its implementation file and import everywhere might not be a good idea.

tests/garage/experiment/test_resume.py
```diff
@@ -58,4 +58,5 @@ def test_resume(self):
 
             runner.resume(n_epochs=10)
 
+        snapshotter.reset()
```
ryanjulian: these two lines should be in teardown

ryanjulian:
I agree that the current implementation is problematic, but we should still be able to test it reliably. 

I think we need to reset the snapshotter in the tearDown methods for all tests that use LocalRunner...

tests/garage/experiment/test_resume.py
zequnyu: Yes... Let me make the changes and see if it breaks anything.

zequnyu: Now `snapshotter.reset()` in `tearDown()`. Maybe not the create and delete of temp folder since it's only used in test_resume() now.

ryanjulian:
this lgtm

PR #675, opened 2019-05-22
Fix flake8 failure by removing per-file-ignores

krzentner:
At least on my system, without this change I get this error:
```
./garage/misc/krylov.py:0:1: X100 Superfluous per-file-ignores for N802,N803,N806
```

It's possible that this fix won't work either, but if the CI says this change is good then I think we can submit it.

### DIFF
--- a/scripts/travisci/check_flake8.sh
+++ b/scripts/travisci/check_flake8.sh
@@ -143,7 +143,6 @@ check_flake8() {
   flake8 --isolated \
          --import-order-style=google \
          --application-import-names="${garage_packages}" \
-         --per-file-ignores="./src/garage/misc/krylov.py:N802,N803,N806" \
          "$@"
   status="$((${status} | ${?}))"
 }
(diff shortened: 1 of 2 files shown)


ryanjulian:
does this work if you run flake8 on the entire codebase? `flake8 garage`?

ryanjulian:
it seems to have failed?

ryanjulian:
its this obseleted by #700?

krzentner:
It's basically obsoleted by #700. I think the failure caused here is due to a version mismatch between my environment and travis. I'll close this PR once I get the version out of travis.

krzentner:
Okay, this happened because I had an old flake8 version, and installed the `flake8-per-file-ignores` plugin, instead of upgrading flake8. This shouldn't be a concern for anyone else.

krzentner:
For reference, this is the version string TravisCI has:
`3.7.7 (flake8-docstrings: 1.3.0, pydocstyle: 3.0.0, import-order: 0.18.1, mccabe: 0.6.1, naming: 0.7.0, pycodestyle: 2.5.0, pyflakes: 2.1.1) CPython 3.6.6 on Linux`

I have no idea where to get `naming: 0.7.0`, since it doesn't seem to have anything to do with the `naming` module on [pypi](https://pypi.org/project/naming/), which has latest version `0.3.0`. At any rate, the other plugins can all be installed, and everything seems to pass now, so this issue is resolved.

PR #676, opened 2019-05-23
Fix some code inconsistencies

nish21:
This commit fixes the following things.

1. Removes `tensor_utils.to_onehot_sym` and replaces the calls with `tf.one_hot`. The idea was to replace all existing usages of the former API and use the TF API. Turns out there was only one file using that function.

2. There were a few instances were return value of a context manager was assigned to instance attributes which are now assigned to a temporary variable which is then assigned to the instance attribute.

### DIFF
--- a/garage/tf/regressors/gaussian_mlp_regressor_with_model.py
+++ b/garage/tf/regressors/gaussian_mlp_regressor_with_model.py
@@ -41,7 +41,7 @@ class GaussianMLPRegressorWithModel(StochasticRegressor2):
         output_b_init (callable): Initializer function for the bias
             of output dense layer(s). The function should return a
             tf.Tensor.
-        optimizer (tf.Optimizer): Optimizer for minimizing the negative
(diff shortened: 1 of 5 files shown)


garage/tf/regressors/gaussian_mlp_regressor_with_model.py
```diff
@@ -41,7 +41,7 @@ class GaussianMLPRegressorWithModel(StochasticRegressor2):
         output_b_init (callable): Initializer function for the bias
             of output dense layer(s). The function should return a
             tf.Tensor.
-        optimizer (tf.Optimizer): Optimizer for minimizing the negative
+        optimizer (garage,tf.Optimizer): Optimizer for minimizing the negative
```
ahtsan: garage.tf.Optimizer

nish21: fixed it

PR #677, opened 2019-05-23
GaussianLSTMPolicy with model

ahtsan:
Added GaussianLSTMModel and GaussianLSTMPolicyWithModel.

Added test for PPO with GaussianLSTMPolicyWithModel.

Apart from testing functionality of GaussianLSTMPolicyWithModel
in test_gaussian_lstm_policy_with_model.py, transitions from the
old policy (GaussianLSTMPolicy) to the new policy
(GaussianLSTMPolicyWithModel) are also tested in
test_gaussian_lstm_policy_with_model_transit.py, to make sure
they have the same API.


garage/tf/policies/gaussian_lstm_policy_with_model.py
```diff
@@ -0,0 +1,295 @@
+"""GaussianLSTMPolicy with GaussianLSTMModel."""
+from akro.tf import Box
```
ryanjulian: `import akro.tf`

garage/tf/policies/gaussian_lstm_policy_with_model.py
```diff
@@ -0,0 +1,295 @@
...
+                 cell_state_init=tf.zeros_initializer(),
+                 cell_state_init_trainable=False,
+                 forget_bias=True,
+                 learn_std=True,
+                 std_share_network=False,
+                 init_std=1.0,
+                 layer_normalization=False,
+                 state_include_action=True):
+        if not isinstance(env_spec.action_space, Box):
+            raise ValueError('GaussianLSTMPolicy only works'
```
ryanjulian: maybe also print the user what they passed, to help them understand the difference

garage/tf/policies/gaussian_lstm_policy_with_model.py
```diff
@@ -0,0 +1,295 @@
...
+
+    def reset(self, dones=None):
+        """
+        Reset the policy.
+
+        Args:
+            dones (numpy.ndarray): Bool that indicates terminal state(s).
+
+        """
+        if dones is None:
```
ryanjulian: `if not dones`

garage/tf/policies/gaussian_lstm_policy_with_model.py
```diff
@@ -0,0 +1,295 @@
...
+            dones = [True]
+        dones = np.asarray(dones)
```
ryanjulian: why np.asarray? you already specified the input is an np.ndarray

garage/tf/policies/gaussian_lstm_policy_with_model.py
```diff
@@ -0,0 +1,295 @@
...
+                name=name)
+
+        return dict(mean=mean_var, log_std=log_std_var)
+
+    def reset(self, dones=None):
+        """
+        Reset the policy.
+
+        Args:
+            dones (numpy.ndarray): Bool that indicates terminal state(s).
```
ryanjulian: it seems this need to be a numpy.ndarray of bools, right?

garage/tf/policies/gaussian_lstm_policy_with_model.py
```diff
@@ -0,0 +1,295 @@
...
+    def get_actions(self, observations):
+        """
+        Get multiple actions from this policy for the input observations.
+
+        Args:
+            observations (numpy.ndarray): Observations from environment.
+
+        Returns:
+            actions (numpy.ndarray): Predicted actions.
+            agent_infos (dict): Mean and log std of the distributions
```
ryanjulian: dict of lists? list of dicts?

ahtsan: I think it's `dict(numpy.ndarray)`. Will fix

garage/tf/policies/gaussian_lstm_policy_with_model.py
ahtsan: Should we print the class? like
```python
            raise ValueError('GaussianLSTMPolicy only works with '
                             'akro.tf.Box action space, but not {}'
                             .format(type(env_spec.action_space)))
```

garage/tf/policies/gaussian_lstm_policy_with_model.py
ahtsan: Yes, do you mean we should state `dones (numpy.ndarray[Bool])`?

garage/tf/policies/gaussian_lstm_policy_with_model.py
ryanjulian: i think printing just `env_spec.action_space` should produce a nice output, no?

garage/tf/policies/gaussian_lstm_policy_with_model.py
ryanjulian: meh, on second thought let's not go this deep with the typing yet.

garage/tf/policies/gaussian_lstm_policy_with_model.py
ahtsan: It gives us Discrete(2) while type gives us the full class name, i.e. akro.tf.Discrete

ryanjulian: ok lgtm

src/garage/tf/models/gaussian_lstm_model.py
```diff
@@ -0,0 +1,234 @@
...
+        hidden_w_init (callable): Initializer function for the weight
+            of intermediate dense layer(s). The function should return a
+            tf.Tensor.
+        hidden_b_init (callable): Initializer function for the bias
+            of intermediate dense layer(s). The function should return a
+            tf.Tensor.
+        recurrent_nonlinearity (callable): Activation function for recurrent
+            layers. It should return a tf.Tensor. Set it to None to
+            maintain a linear activation.
+        recurrent_w_init (callable): Initializer function for the weight
```
zhanpenghe: Hmm I believe it is not callable right? As far as I saw it is an initializer instance

ryanjulian: tf initializers are callables which return tensors

garage/tf/policies/gaussian_lstm_policy_with_model.py
```diff
@@ -0,0 +1,295 @@
...
+    def get_action(self, observation):
+        """
+        Get single action from this policy for the input observation.
+
+        Args:
+            observation (numpy.ndarray): Observation from environment.
+
+        Returns:
+            action (numpy.ndarray): Predicted action.
+            agent_info (dict[numpy.ndarray]): Mean and log std of the
```
ryanjulian: maybe be explicit about the contents of the dict, e.g.

https://github.com/rlworkgroup/garage/blob/master/src/garage/tf/algos/batch_polopt.py#L85

garage/tf/policies/gaussian_lstm_policy_with_model.py
```diff
@@ -0,0 +1,295 @@
...
+    def get_actions(self, observations):
+        """
+        Get multiple actions from this policy for the input observations.
+
+        Args:
+            observations (numpy.ndarray): Observations from environment.
+
+        Returns:
+            actions (numpy.ndarray): Predicted actions.
+            agent_infos (dict[numpy.ndarray]): Mean and log std of the
```
ryanjulian: be explicit about keys in the dict https://github.com/rlworkgroup/garage/blob/master/src/garage/tf/algos/batch_polopt.py#L85

src/garage/tf/policies/gaussian_lstm_policy_with_model.py
```diff
@@ -0,0 +1,295 @@
...
+    def get_action(self, observation):
+        """
+        Get single action from this policy for the input observation.
+
+        Args:
+            observation (numpy.ndarray): Observation from environment.
+
+        Returns:
+            action (numpy.ndarray): Predicted action.
+            agent_info (dict[numpy.ndarray]): Mean and log std of the
```
ryanjulian: be explicit
https://github.com/rlworkgroup/garage/blob/master/src/garage/tf/algos/batch_polopt.py#L85

zhanpenghe:
Can you please make it multi layer recurrent? Single layer sometimes does not work well for hard problem..

ahtsan:
@zhanpenghe I think another PR will be more appropriate.

zhanpenghe:
File an issue then.

PR #678, opened 2019-05-23
Update CONTRIBUTING.md

ryanjulian:
Improvements to the garage-specific style section:
* Clearer formatting
* Examples of do's and don'ts
* Adds rule requiring use of `abc` for base and interface classes
* Adds rule requiring absolute module imports for external dependencies

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,9 +34,135 @@ The Python code in garage conforms to the [PEP8](https://www.python.org/dev/peps
 
 ### garage-specific Python style
 These are garage-specific rules which are not part of the aforementioned style guides.
-* Python package imports should be sorted alphabetically within their PEP8 groupings. The sorting is alphabetical from left to right, ignoring case and Python keywords (i.e. `import`, `from`, `as`). Notable exceptions apply in `__init__.py` files, where sometimes this rule will trigger a circular import.
-* We prefer single-quoted strings (`'foo'`) over double-quoted strings (`"foo"`), unless there is a compelling escape or formatting reason for using single quotes (e.g. a single quote appears inside the string).
-* Add convenience imports in `__init__.py` of a package for shallow first-level repetitive imports, but not for subpackages, even if that subpackage is defined in a single `.py` file. For instance, if an import line reads `from garage.foo.bar import Bar` then you should add `from garage.foo.bar import Bar` to `garage/foo/__init__.py` so that users may instead write `from garage.foo import Bar`. However, if an import line reads `from garage.foo.bar.stuff import Baz`, *do not* add `from garage.foo.bar.stuff import Baz` to `garage/foo/__init__.py`, because that obscures the `stuff` subpackage.
+* Python package imports should be sorted alphabetically within their PEP8 groupings.
+
+    The sorting is alphabetical from left to right, ignoring case and Python keywords (i.e. `import`, `from`, `as`). Notable exceptions apply in `__init__.py` files, where sometimes this rule will trigger a circular import.
+
+* Prefer single-quoted strings (`'foo'`) over double-quoted strings (`"foo"`).
+
+    Double-quoted strings can be used if there is a compelling escape or formatting reason for using single quotes (e.g. a single quote appears inside the string).
+
+* Add convenience imports in `__init__.py` of a package for shallow first-level repetitive imports, but not for subpackages, even if that subpackage is defined in a single `.py` file.
+
+    For instance, if an import line reads `from garage.foo.bar import Bar` then you should add `from garage.foo.bar import Bar` to `garage/foo/__init__.py` so that users may instead write `from garage.foo import Bar`. However, if an import line reads `from garage.foo.bar.stuff import Baz`, *do not* add `from garage.foo.bar.stuff import Baz` to `garage/foo/__init__.py`, because that obscures the `stuff` subpackage.
+
+    *Do*
+
+    `garage/foo/__init__.py`:
+    ```python
+    """Foo package."""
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -34,9 +34,116 @@ The Python code in garage conforms to the [PEP8](https://www.python.org/dev/peps
...
+    This applies to both packages from the standard library and 3rd-party dependencies. If a package has a long or cumbersome full path, or is used very frequently (e.g. `numpy`, `tensorflow`), you may use the keyword `as` to create a file-specific name which makes sense.
+    
+    *Do*
+    ```python
+    import collections
+    
+    import gym.spaces
+    import tensorflow as tf
+    
+    from garage.tf.models import MLPModel
```
ahtsan: Why shouldn't we do `import garage.tf.models` here and do `garage.tf.models.MLPModel` in code? Is it because it's too long?

ryanjulian: The purpose of this rule is to make it easy to see anywhere in the code where we're using external vs internal dependencies.

Forcing people to write out long names for external deps makes it obvious that something is from an external dependency, even if the import is 1000 lines up, e.g. `q = collections.deque(10)` vs `q = deque(10)`.

```python
import collections

from garage.tf.model import MLPModel

q = collections.deque(10)
m = MLPModel()
```

IMO internal dependencies are used so frequently that applying this rule to internal dependencies (e.g. `MLPModel`) would just make the code *harder* to read and for little benefit. A variation on this style is practiced in some places, though, e.g. Google Python style practices a variation where you can import modules but not objects, e.g.

```python
import collections

from garage.tf import models

q = collections.deque(10)
m = models.MLPModel()  # not bad
```

I don't know anyone who practices full paths for both internal and external dependencies, but I'm sure someone does it for some reason:
```python
import collections

import garage.tf.models

q = collections.deque(10)
m = garage.tf.models.MLPModel(m)  # ick
```

The assumption is that internal code will be used more frequently than external code,

with a few notable exceptions, which can be abbreviated with standard notations (e.g. tf, np, torch) to shorten them.

ryanjulian: on reason behind the google variation is that their codebase is so big that, past a couple parent packages, everything might as well be considered "external" code, even if it's owned by Google (because they all live in a single package namespace).

We don't have this problem, so I'm not sure forcing people to include the module as a stem (e.g. `models.MLPModel()`) really achieves much.

ryanjulian: one argument in favor of the google style, though, if that you're virtually guaranteed that `models` is defined either in `models.py` or a directory `models/`, whereas `MLPModel` could be anywhere.

krzentner: Yeah, this all sounds about right to me. One note is that I've sometimes seen styles where specific external modules do allow importing classes / functions individually (ironically, collections itself being the most common example). I could see this either way, especially because garage has been "exploded" into other libraries controlled by the rlworkgroup. For example, no one seems keen to write `dowel.logger.log(...)` everywhere. We could probably keep this list fairly short, and it would still be helpful.

On a related note, by convention PyTorch actually has three imports: `torch`, `torch.nn as nn`, and `torch.nn.functional as F`.

CONTRIBUTING.md
```diff
@@ -34,9 +34,116 @@ The Python code in garage conforms to the [PEP8](https://www.python.org/dev/peps
...
+    ```python
+    import collections
+
+    import gym.spaces
+    import tensorflow as tf
+
+    from garage.tf.models import MLPModel
+
+    q = collections.deque(10)
+    d = gym.spaces.Discrete(5)
```
krzentner: Could you make these examples shorter? They're a little distracting.
Also, I don't think `gym.spaces` exists anymore, which makes it more confusing.

ryanjulian:
Does anyone have more specific feedback before I submit this? I will apply it across all the rlworkgroup projects once accepted.

@krzentner @ahtsan @CatherineSue

CONTRIBUTING.md
ryanjulian: I will add a list of known exceptions

CONTRIBUTING.md
```diff
@@ -34,9 +34,116 @@ The Python code in garage conforms to the [PEP8](https://www.python.org/dev/peps
...
+    The sorting is alphabetical from left to right, ignoring case and Python keywords (i.e. `import`, `from`, `as`). Notable exceptions apply in `__init__.py` files, where sometimes this rule will trigger a circular import.
+
+* Prefer single-quoted strings (`'foo'`) over double-quoted strings (`"foo"`).
+
+    Double-quoted strings can be used if there is a compelling escape or formatting reason for using single quotes (e.g. a single quote appears inside the string).
+
+* Add convenience imports in `__init__.py` of a package for shallow first-level repetitive imports, but not for subpackages, even if that subpackage is defined in a single `.py` file.
+
+    For instance, if an import line reads `from garage.foo.bar import Bar` then you should add `from garage.foo.bar import Bar` to `garage/foo/__init__.py` so that users may instead write `from garage.foo import Bar`. However, if an import line reads `from garage.foo.bar.stuff import Baz`, *do not* add `from garage.foo.bar.stuff import Baz` to `garage/foo/__init__.py`, because that obscures the `stuff` subpackage.
+
```
zhanpenghe: Maybe add inter-package importing too? They need to be full path instead of short imports to avoid circular imports.

CONTRIBUTING.md
ryanjulian: `gym.spaces` exists. are you thinking of `garage.spaces`?

Issue #377, opened 2018-10-31
Audit garage.contrib


ryanjulian:

ryanjulian:
### contrib/bichengcao
**Description:** Example experiments using openai/gym.
**Comment:** Duplicates what's already in examples.
**Verdict:** Remove

### contrib/rllab_hyperopt
**Description:** Using hyperopt with rllab (garage) for hyperparameter searches.
**Comment:** This is highly-reliant on the old rllab-style EC2 interface, which is uniformly pretty bad. I think the better approach here would be to use ray tune or something similar.
**Verdict:** Remove

### contrib/alexbeloi
**Description:** Implements a special sampler which uses importance sampling to return historical samples
**Comment:** I can't find a citation for this method which makes it difficult to assess if it's been implemented correctly. Keeping it doesn't seem to hurt, as long as it's not a huge amount of refactoring.
**Verdict:** Attempt to merge

PR #684, opened 2019-05-24
Remove contrib.rllab_hyperopt

ryanjulian:
This is heavily-reliant on the old EC2 integration, which we are working
on removing and replacing with an exteran job runner.


ryanjulian:
ref #377

PR #686, opened 2019-05-26
Separate large tests cases into their own CI job

ahtsan:
Currently CI contains too many test cases. This PR adjusts some test case passing threshold and assigns attribute to some huge test cases so that those tests can be handled by another bigger CI.

### DIFF
--- a/tests/garage/np/algos/test_cem.py
+++ b/tests/garage/np/algos/test_cem.py
@@ -11,10 +11,10 @@ class TestCEM(TfGraphTestCase):
     def test_cem_cartpole(self):
         """Test CEM with Cartpole-v1 environment."""
         with LocalRunner() as runner:
-            env = TfEnv(env_name="CartPole-v1")
+            env = TfEnv(env_name='CartPole-v1')
 
             policy = CategoricalMLPPolicy(
-                name="policy", env_spec=env.spec, hidden_sizes=(32, 32))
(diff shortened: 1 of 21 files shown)


tests/garage/np/algos/test_cem.py
```diff
@@ -32,7 +32,9 @@ def test_cem_cartpole(self):
 
             runner.setup(algo, env, sampler_cls=OnPolicyVectorizedSampler)
             rtn = runner.train(
-                n_epochs=5, batch_size=2000, n_epoch_cycles=n_samples)
+                n_epochs=10, batch_size=2048, n_epoch_cycles=n_samples)
             assert rtn > 40
 
             env.close()
+
+    test_cem_cartpole.huge = True
```
ryanjulian: can you mark these as "large" instead? we reserve huge for benchmarks right now

tests/garage/tf/algos/test_ppo_with_model_baseline.py
```diff
@@ -67,3 +67,6 @@ def test_ppo_pendulum_recurrent_continuous_baseline(self):
             assert last_avg_ret > 30
 
             env.close()
+
+    test_ppo_pendulum_continuous_baseline.huge = True
```
ryanjulian: preferably put this right next to the test definition

Issue #680, opened 2019-05-24
Eliminate the use of process-global snapshotter


zequnyu:
We should have the `LocalRunner` create a `snapshotter` instance it owns.

ryanjulian:
@zequnyu is this fixed?

zequnyu:
Closed via #699.

PR #699, opened 2019-05-29
Refactor snapshotter

zequnyu:
This PR will address #680.
- removed the global Snapshotter
- updated the Snapshotter API and its usage
- updated tests and examples correspondingly


examples/tf/dqn_cartpole.py
```diff
@@ -15,44 +15,43 @@
...
-            n_train_steps=500,
-            n_epoch_cycles=n_epoch_cycles,
-            target_network_update_freq=1,
-            buffer_batch_size=32)
+    n_epochs = 10
+    n_epoch_cycles = 10
+    sampler_batch_size = 500
+    num_timesteps = n_epochs * n_epoch_cycles * sampler_batch_size
+    env = TfEnv(gym.make('CartPole-v0'))
+    replay_buffer = SimpleReplayBuffer(
```
ryanjulian: can you add some whitespace to make this file a little easier to read?

src/garage/experiment/experiment_wrapper.py
```diff
@@ -73,6 +74,16 @@ def run_experiment(argv):
         type=int,
         default=1,
         help='Gap between snapshot iterations.')
+    parser.add_argument(
```
ryanjulian: AFAICT it's impossible to pass a tensorflow session as a command line argument

src/garage/experiment/experiment_wrapper.py
```diff
@@ -182,24 +193,32 @@ def run_experiment(argv):
...
+        snapshot_dir=log_dir,
+        snapshot_mode=args.snapshot_mode,
+        snapshot_gap=args.snapshot_gap)
+
     if args.resume_from_dir is not None:
-        with LocalRunner() as runner:
-            runner.restore(args.resume_from_dir, from_epoch=args.resume_epoch)
+        with LocalRunner(
+                snapshotter=snapshotter,
+                sess=args.tf_session,
```
ryanjulian: hmm this might be a problem. people have real reasons to want to use their own TF session...

src/garage/experiment/local_tf_runner.py
```diff
@@ -44,26 +62,12 @@ class LocalRunner:
...
-
-        Args:
-            max_cpus(int): The maximum number of parallel sampler workers.
-            sess(tf.Session): An optional tensorflow session.
-                  A new session will be created immediately if not provided.
+    def __init__(self, snapshotter=None, sess=None, max_cpus=1):
+        if snapshotter is not None:
+            self.snapshotter = snapshotter
+        else:
+            self.snapshotter = Snapshotter()
```
ryanjulian: hmm what if we just don't create any snapshots in this case? where will a snapshotter with no arguments even put its snapshots???

zequnyu: Can we assume that, if user doesn't pass `snapshotter` to `LocalRunner`, then they don't want to snapshot at all?

ryanjulian: i think so. or perhaps we should just have them pass `snapshot_dir` and other optional arguments (mode, etc.), since there is only one Snapshotter class that we can think of.

@krzentner what do you think?

zequnyu:
My thought is: if we cannot remove context manager due to the tf session, then we might also put the `snapshotter` into `__enter__` in `LocalRunner` for now. `SnapshotConfig` also serves as a sign to indicate whether the user want to snapshot or not.

If this is acceptable, I'll make some tests to finalise it.

examples/tf/trpo_swimmer.py
```diff
@@ -22,7 +22,7 @@ def run_task(*_):
             baseline=baseline,
             max_path_length=500,
             discount=0.99,
-            step_size=0.01)
+            max_kl_step=0.01)
```
krzentner: Why is this change here?

src/garage/experiment/local_tf_runner.py
```diff
@@ -78,10 +80,18 @@ def __init__(self, sess=None, max_cpus=1):
...
 
+        Also, snapshotter is created if SnapshotConfig is passed.
+
         Returns:
             This local runner.
 
         """
+
+        if isinstance(self.snapshot_config, SnapshotConfig):
+            self.snapshotter = Snapshotter(self.snapshot_config.snapshot_dir,
```
krzentner: Instead of passing every field of `snapshot_config` into the `Snapshotter`, why not just make `snapshot_config` a constructor function for `Snapshotter`? e.g. this line would become:
```python
self.snapshotter = self.snapshot_config()
```

examples/tf/trpo_swimmer.py
zequnyu: Because we renamed `step_size` to `max_kl_step` in the algorithm.

zequnyu: Some parameters in `/examples` are outdated. Which one you think is unrelated change?

ryanjulian: it's true -- we changed this param to `max_kl_step` some time ago.

src/garage/experiment/local_tf_runner.py
zequnyu: Calling a `config` to create a `snapshotter` is a kind of surprising behaviour to me. Also, I don't quite get your point of:

> users can pass in a new type if they would like

I think our concern here is to know whether user want to snapshot or not. What type would they pass into `LocalRunner` potentially except `SnapshotConfig`?

zequnyu: I just checked the doc that both `SimpleNamespace` and `namedtuple` don't have the `__call__` attribute.

krzentner: What I was trying to get at is that users might have a different snapshotter class that they want to use instead of our snapshotter.

Since "calling" classes in Python constructs them, a simple way of making it possible to pass in a new snapshotter type is to make the way that you use the config class be to call it. Then, the user could just pass their own class in place of the config, and everything would just work.

Notably, you can still define `__call__` for a custom named tuple (for example, like this):
```python
import typing

class SnapshotterConfig(typing.NamedTuple):
    snapshot_dir: str
    snapshot_mode: str
    snapshot_gap: int

    def __call__(self):
        return Snapshotter(**self._asdict())
```

Having said that, custom Snapshotters seem like a niche concern, so we can drop it.

src/garage/experiment/local_tf_runner.py
```diff
@@ -225,12 +232,12 @@ def restore(self, snapshot_dir, from_epoch='last'):
         Examples:
             1. Resume experiment immediately.
             with LocalRunner() as runner:
-                runner.restore(snapshot_dir)
```
naeioi: Please update documentation of `restore()` on `snapshot_dir`.

naeioi: Please update documentation of `restore()` on `snapshot_dir`.

src/garage/experiment/local_tf_runner.py
```diff
@@ -241,8 +248,11 @@ def restore(self, snapshot_dir, from_epoch='last'):
             specify manually or through run_experiment() interface.
 
         """
-        snapshotter.snapshot_dir = snapshot_dir
-        saved = snapshotter.load(from_epoch)
+        assert isinstance(self.snapshot_config, SnapshotConfig), (
+            'You must initialize SnapshotConfig in LocalRunner before '
```
naeioi: This does not give instruction on where to load the snapshot. What about 'You have to specify saved snapshot folder in construction via SnapshotConfig'.

src/garage/experiment/local_tf_runner.py
```diff
@@ -193,24 +199,25 @@ def save(self, epoch, paths=None):
...
-            params['paths'] = paths
-        params['last_epoch'] = epoch
-        snapshotter.save_itr_params(epoch, params)
+            # Save states
+            params['env'] = self.env
+            params['algo'] = self.algo
+            if paths:
+                params['paths'] = paths
+            params['last_epoch'] = epoch
+            self.snapshotter.save_itr_params(epoch, params)
```
naeioi: This line breaks when there's no snapshotter.

src/garage/experiment/local_tf_runner.py
```diff
@@ -193,24 +199,25 @@ def save(self, epoch, paths=None):
         """
         assert self.has_setup
 
-        logger.log('Saving snapshot...')
+        if isinstance(self.snapshot_config, SnapshotConfig):
+            logger.log('Saving snapshot...')
```
naeioi: I suggest adding a warning of non-resumable training when `SnapshotConfig` is missing.

src/garage/experiment/local_tf_runner.py
zequnyu: This will not break as we checked `isinstance(self.snapshot_config, SnapshotConfig)` before. Though I don't know if there is a better way to tell the Runner whether user want to snapshot or not.

src/garage/experiment/local_tf_runner.py
zequnyu: The case is not possible if we use `experiment_wrapper` to run experiment. Now the `SnapshotConfig` will be passed to the Runner with default option even users don't want to snapshot. I just realised this is a problem.

zequnyu: After some discussion with KR, I made some changes and added a warning in `__init__`. Please see if it resolves your concern.

ryanjulian:
@zequnyu are you still working on this?

examples/tf/dqn_cartpole.py
```diff
@@ -22,17 +22,23 @@ def run_task(*_):
         n_epoch_cycles = 10
         sampler_batch_size = 500
         num_timesteps = n_epochs * n_epoch_cycles * sampler_batch_size
+
```
ryanjulian: this additional spacing appears unrelated to the stated change -- please remove it

src/garage/experiment/__init__.py
```diff
@@ -3,15 +3,11 @@
...
 from garage.experiment.experiment import variant
 from garage.experiment.experiment import VariantGenerator
-from garage.experiment.snapshotter import Snapshotter
-
-# LocalRunner needs snapshotter to be imported, so we have to use a strange
-# import order here
-snapshotter = Snapshotter()
+from garage.experiment.snapshotter import SnapshotConfig, Snapshotter
 
 from garage.experiment.local_tf_runner import LocalRunner  # noqa: I100,E402,E501,I202 pylint: disable=wrong-import-position
```
ryanjulian: please remove all these linting exceptions and attempt to place this at the correct position (you may need to re-apply some)

src/garage/experiment/experiment_wrapper.py
```diff
@@ -120,7 +120,8 @@ def run_experiment(argv):
         type=ast.literal_eval,
         default=False,
         help='Print only the tabular log information (in a horizontal format)')
-    parser.add_argument('--seed', type=int, help='Random seed for numpy')
+    parser.add_argument(
+        '--seed', type=int, default=1, help='Random seed for numpy')
```
ryanjulian: this is actually used for numpy, tensorflow, pytorch, random, and any other RNG in the system.

src/garage/experiment/experiment_wrapper.py
```diff
@@ -120,7 +120,8 @@ def run_experiment(argv):
         type=ast.literal_eval,
         default=False,
         help='Print only the tabular log information (in a horizontal format)')
-    parser.add_argument('--seed', type=int, help='Random seed for numpy')
+    parser.add_argument(
```
ryanjulian: I think this should default to `None` so that the system runs a time-based seed by default.

src/garage/experiment/local_tf_runner.py
```diff
@@ -44,26 +64,17 @@ class LocalRunner:
...
-        Note:
-            The local runner will set up a joblib task pool of size max_cpus
-            possibly later used by BatchSampler. If BatchSampler is not used,
-            the processes in the pool will remain dormant.
-
-            This setup is required to use tensorflow in a multiprocess
-            environment before a tensorflow session is created
-            because tensorflow is not fork-safe.
+    def __init__(self, snapshot_config=None, sess=None, max_cpus=1):
+        if isinstance(snapshot_config, SnapshotConfig):
```
ryanjulian: why not just `if snapshot_config`?

src/garage/experiment/local_tf_runner.py
```diff
@@ -44,26 +64,17 @@ class LocalRunner:
...
+            self.snapshotter = Snapshotter(snapshot_config.snapshot_dir,
+                                           snapshot_config.snapshot_mode,
+                                           snapshot_config.snapshot_gap)
+        else:
+            if snapshot_config is not None:
```
ryanjulian: perhaps you mean `if snapshot_config is None`?

src/garage/experiment/local_tf_runner.py
ryanjulian: i think you should restructure this using a guard clause:

https://blog.mafr.de/2009/06/12/a-case-for-guard-clauses/

```python
self.snapshotter = None
if snapshot_config:
    self.snapshotter = Snapshotter(...)
else:
    warnings.warn('No snapshotter created...etc.')
```

src/garage/experiment/local_tf_runner.py
```diff
@@ -193,28 +204,28 @@ def save(self, epoch, paths=None):
         """
         assert self.has_setup
 
-        logger.log('Saving snapshot...')
+        if self.snapshotter is not None:
```
ryanjulian: `if self.snapshotter:`

src/garage/experiment/local_tf_runner.py
ryanjulian: i don't think it will break because you checked that `if self.snaphotter:`

src/garage/experiment/local_tf_runner.py
```diff
@@ -182,7 +193,7 @@ def obtain_samples(self, itr, batch_size):
             logger.log('Obtaining samples...')
         return self.sampler.obtain_samples(itr, batch_size)
 
-    def save(self, epoch, paths=None):
+    def _save(self, epoch, paths=None):
```
ryanjulian: i recommend you rename this to `_maybe_save` since this is not guaranteed to save a snapshot.

do we really want to never save a snapshot? is this a use case we want to support? it seems like the code would be simpler if no snapshots was an error.

@krzentner WDYT?

ryanjulian: hmm why is save a private API and restore a public API? is there a problem with save being public?

src/garage/experiment/local_tf_runner.py
```diff
@@ -241,8 +252,11 @@ def restore(self, snapshot_dir, from_epoch='last'):
             specify manually or through run_experiment() interface.
 
         """
-        snapshotter.snapshot_dir = snapshot_dir
-        saved = snapshotter.load(from_epoch)
+        assert self.snapshotter is not None, (
+            'You must specify SnapshotConfig in LocalRunner construction '
```
ryanjulian: this should be an exception, not an assert.

i think a more apt error is "You cannot restore this LocalRunner because a snapshot is required to do so. You can ensure your experiment produces a snapshot directory by providing a SnapshotConfig to LocalRunner during training."

with that said, I think a lot of your code could be simpler if you just disallowed the no-snapshots case, and I recommend you do that. Snapshots don't take up much space, especially if your snapshot_mode is 'last'.

src/garage/experiment/snapshotter.py
```diff
@@ -1,4 +1,4 @@
-"""Allows the taking of snapshots."""
+import collections
 import errno
 import os
 from os import listdir
```
ryanjulian: import os

src/garage/experiment/local_tf_runner.py
zequnyu: I'm trying to prevent a situation that user passes something which is not a correct `SnapshotConfig`, then we warn them.

zequnyu:
I think this is ready for another review.

Now if Nothing is passed to `LocalRunner`, it will create its `snapshotter` with default path set in `snapshotter.__init__()` It will by default save the last epoch. This way we aren't forced to pass the config when adding tests using `LocalRunner`.

Downside is we'd better do everything example like this:
```python3
    with LocalRunner(snapshot_config=snapshot_config):
        ...
        ...
```
Otherwise the customised settings in `run_experiment()` will be silently ignored. Bad example is:
```python3
def run_task(*_):
    with LocalRunner():
        # snapshot mode is still by default 'last' even if the intention is 'all'
        # because `snapshot_config` is not passed
        ...
        ...
run_experiment(run_task, snapshot_mode='all', seed=1)
```
But this is probably a worth cost, otherwise we have to define a `snapshot_config` every time if `LocalRunner` is needed in tests, which seems completely unnecessary.

src/garage/experiment/local_tf_runner.py
zequnyu: `save()` is only used inside. What's the reason to set it to public?

examples/tf/her_ddpg_fetchreach.py
```diff
@@ -71,7 +71,7 @@ def run_task(*_):
 
         runner.setup(algo=ddpg, env=env)
 
-        runner.train(n_epochs=50, n_epoch_cycles=20)
+        runner.train(n_epochs=50, n_epoch_cycles=20, batch_size=100)
 
 
 run_experiment(run_task, snapshot_mode='last', seed=1)
```
ryanjulian: why would run_experiment still require a `snapshot_mode` param if LocalRunner requires a `snapshot_config`?

src/garage/experiment/local_tf_runner.py
```diff
@@ -136,7 +143,7 @@ def setup(self, algo, env, sampler_cls=None, sampler_args=None):
         logger.log(self.sess.graph)
         self.has_setup = True
 
-        self.setup_args = SimpleNamespace(
+        self.setup_args = types.SimpleNamespace(
```
ryanjulian: any reason setup_args is public?

src/garage/experiment/local_tf_runner.py
```diff
@@ -206,16 +213,17 @@ def save(self, epoch, paths=None):
         if paths:
             params['paths'] = paths
         params['last_epoch'] = epoch
-        snapshotter.save_itr_params(epoch, params)
+        self.snapshotter.save_itr_params(epoch, params)
```
ryanjulian: any reason snapshotter is public?

src/garage/experiment/local_tf_runner.py
```diff
@@ -206,16 +213,17 @@ def save(self, epoch, paths=None):
...
 
         logger.log('Saved')
 
-    def restore(self, snapshot_dir, from_epoch='last'):
+    def restore(self, resume_from_dir, resume_from_epoch='last'):
```
ryanjulian: maybe just `from_dir` and `from_epoch`

src/garage/experiment/local_tf_runner.py
```diff
@@ -206,16 +213,17 @@ def save(self, epoch, paths=None):
...
         """Restore experiment from snapshot.
 
         Args:
-            snapshot_dir(str): Directory of snapshot.
-            from_epoch(str or int): The epoch to restore from.
+            resume_from_dir (str): Directory of the pickle file
+                to resume experiment from.
+            resume_from_epoch (str or int): The epoch to restore from.
```
ryanjulian: how about -1 for last? this is consistent with the rest of python.

src/garage/experiment/local_tf_runner.py
```diff
@@ -311,11 +319,10 @@ def train(self,
             The average return in last epoch cycle.
 
         """
-        assert self.has_setup, ('Use Runner.setup() to setup runner before '
-                                'training.')
+        assert self.has_setup, 'Use setup() to setup runner before training.'
```
ryanjulian: if you are detecting a user error, rather than asserting an invariant, raise and exception.

src/garage/experiment/snapshotter.py
```diff
@@ -1,127 +1,125 @@
...
 import os
-from os import listdir
-from os import path as osp
 import pickle
 import re
 
 import joblib
 
-from garage.misc.console import mkdir_p
+from garage.experiment.experiment import ensure_dir
```
ryanjulian: See https://docs.python.org/3.5/library/pathlib.html#pathlib.Path.mkdir instead (and delete the one from misc.console)

src/garage/experiment/snapshotter.py
```diff
@@ -1,127 +1,125 @@
...
+SnapshotConfig = collections.namedtuple(
+    'SnapshotConfig', ['snapshot_dir', 'snapshot_mode', 'snapshot_gap'])
 
-class Snapshotter:
-    """This class handles the creation of snapshots."""
 
-    def __init__(self):
-        self.reset()
+class Snapshotter:
+    """The Snapshotter allows the taking of snapshots.
```
ryanjulian: Please provide a more relevant description of the role this has in the system.

src/garage/experiment/snapshotter.py
```diff
@@ -1,127 +1,125 @@
...
-            elif self._snapshot_mode == 'gap':
-                if itr % self._snapshot_gap == 0:
-                    file_name = osp.join(self._snapshot_dir,
+        file_name = None
+
+        if self._snapshot_mode == 'all':
+            file_name = os.path.join(self._snapshot_dir, 'itr_%d.pkl' % itr)
+        elif self._snapshot_mode == 'last':
+            # override previous params
+            file_name = os.path.join(self._snapshot_dir, 'params.pkl')
```
ryanjulian: last_params.pkl? or latest_params.pkl?

src/garage/experiment/snapshotter.py
```diff
@@ -1,127 +1,125 @@
...
+            pass
+        else:
+            raise ValueError('Invalid snapshot mode {}'.format(
+                self._snapshot_mode))
+
+        if file_name:
+            with open(file_name, 'wb') as file:
+                pickle.dump(params, file)
+
+    def load(self, resume_from_dir, itr='last'):
```
ryanjulian: `from_dir`? we don't know we are resuming.

src/garage/experiment/snapshotter.py
```diff
@@ -1,127 +1,125 @@
...
+        """Load one snapshot of parameters from disk.
 
         Args:
-            itr(int or string): Iteration to load.
+            resume_from_dir (str): Directory of the pickle file
```
ryanjulian: this is API leakage. we don't know we are resuming anything. this component only loads things.

src/garage/experiment/snapshotter.py
```diff
@@ -1,127 +1,125 @@
...
                                             '*.pkl file')
                 files.sort(key=lambda f: int(re.findall(r'\d+', f)[0]))
-                snapshot_file = files[0] if itr == 'first' else files[-1]
-                snapshot_file = osp.join(self._snapshot_dir, snapshot_file)
+                resume_from_file = files[0] if itr == 'first' else files[-1]
+                resume_from_file = os.path.join(resume_from_dir,
+                                                resume_from_file)
 
-        assert osp.isfile(snapshot_file)
+        assert os.path.isfile(resume_from_file)
```
ryanjulian: if you are detecting errors from user input, raise an exception, don't use assert.

src/garage/experiment/snapshotter.py
```diff
@@ -1,127 +1,125 @@
...
+            resume_from_file = os.path.join(resume_from_dir,
+                                            'itr_{}.pkl'.format(itr))
         else:
             if itr not in ('last', 'first'):
                 raise ValueError(
                     "itr should be an integer or 'last' or 'first'")
 
-            snapshot_file = osp.join(self._snapshot_dir, 'params.pkl')
-            if not osp.isfile(snapshot_file):
+            resume_from_file = os.path.join(resume_from_dir, 'params.pkl')
```
ryanjulian: load_from_file

src/garage/experiment/snapshotter.py
```diff
@@ -1,127 +1,125 @@
...
+            if not os.path.isfile(resume_from_file):
                 files = [
-                    f for f in listdir(self._snapshot_dir)
+                    f for f in os.listdir(resume_from_dir)
```
ryanjulian: pathlib might make several of these manipulations simpler

src/garage/experiment/snapshotter.py
```diff
@@ -1,127 +1,125 @@
...
                     if f.endswith('.pkl')
                 ]
                 if not files:
                     raise FileNotFoundError(errno.ENOENT,
                                             os.strerror(errno.ENOENT),
                                             '*.pkl file')
```
ryanjulian: can you provide more helpful info than `*.pkl`?

src/garage/experiment/snapshotter.py
```diff
@@ -1,127 +1,125 @@
...
                 files.sort(key=lambda f: int(re.findall(r'\d+', f)[0]))
```
ryanjulian: just use sorted. i'm not even usre what this regex is doing

tests/garage/experiment/test_experiment_wrapper.py
```diff
@@ -0,0 +1,27 @@
...
+import cloudpickle
+
+from garage.experiment import SnapshotConfig
+from garage.experiment.experiment_wrapper import run_experiment
+
+
+class TestExperimentWrapper(unittest.TestCase):
+    @staticmethod
+    def method_call(snapshot_config, variant_data):
+        assert isinstance(snapshot_config, SnapshotConfig)
```
ryanjulian: it it doesn't need `self`, you can just make this a method in the module (outside the class)

tests/garage/experiment/test_experiment_wrapper.py
```diff
@@ -0,0 +1,27 @@
+import base64
+import unittest.mock
+
+import cloudpickle
+
+from garage.experiment import SnapshotConfig
+from garage.experiment.experiment_wrapper import run_experiment
+
+
+class TestExperimentWrapper(unittest.TestCase):
```
ryanjulian: tests no longer need to inherit from `unittest.TestCase` -- please remove

tests/garage/experiment/test_experiment_wrapper.py
ryanjulian: can you assert anything more about variant_data and snapshot_config?

tests/garage/experiment/test_resume.py
```diff
@@ -36,18 +35,17 @@ def fixture_exp():
...
-        snapshotter.snapshot_dir = folder
-        snapshotter.snapshot_mode = 'last'
+        temp_dir = tempfile.TemporaryDirectory()
+
+        snapshot_config = SnapshotConfig(
+            snapshot_dir=temp_dir.name, snapshot_mode='last', snapshot_gap=1)
+        policy_params = fixture_exp(snapshot_config)
 
-        policy_params = fixture_exp()
         self.teardown_method()
```
ryanjulian: please make these separate test cases, rather than calling teardown/setup manually .that's an abuse of the test framework.

tests/garage/experiment/test_resume.py
```diff
@@ -36,18 +35,17 @@ def fixture_exp():
 
 class TestResume(TfGraphTestCase):
     def test_resume(self):
-        # Manually create and remove temp folder
-        # Otherwise, tempfile unexpectedly removes folder in child folder
-        folder = tempfile.mkdtemp()
-        snapshotter.snapshot_dir = folder
-        snapshotter.snapshot_mode = 'last'
+        temp_dir = tempfile.TemporaryDirectory()
```
ryanjulian: this setup stuff should be in setup_method

tests/garage/experiment/test_resume.py
```diff
@@ -72,4 +70,4 @@ def test_resume(self):
             assert runner.train_args.store_paths
             assert not runner.train_args.pause_for_plot
 
-        shutil.rmtree(folder)
+        temp_dir.cleanup()
```
ryanjulian: should be in teardown_method

tests/garage/experiment/test_snapshot.py
```diff
@@ -27,24 +27,18 @@ def reset_tf(cls):
...
-    def teardown_class(cls):
-        snapshotter.snapshot_dir = cls.prev_log_dir
-        snapshotter.snapshot_mode = cls.prev_mode
+    def tearDownClass(cls):
         logger.remove_all()
-        cls.log_dir.cleanup()
 
     def test_snapshot(self):
-        with LocalRunner() as runner:
+        temp_dir = tempfile.TemporaryDirectory()
```
ryanjulian: should be in setup_method/class

tests/garage/experiment/test_snapshot.py
```diff
@@ -27,24 +27,18 @@ def reset_tf(cls):
...
-        snapshotter.snapshot_dir = cls.log_dir.name
-        snapshotter.snapshot_mode = 'all'
-
         logger.add_output(NullOutput())
 
     @classmethod
-    def teardown_class(cls):
-        snapshotter.snapshot_dir = cls.prev_log_dir
-        snapshotter.snapshot_mode = cls.prev_mode
+    def tearDownClass(cls):
```
ryanjulian: please don't use the old style camelCase

tests/garage/experiment/test_snapshot.py
```diff
@@ -80,3 +74,5 @@ def test_snapshot(self):
                 assert algo.policy
 
                 rollout(env, algo.policy, animated=False)
+
+        temp_dir.cleanup()
```
ryanjulian: should be in teardown_method

tests/garage/experiment/test_snapshotter.py
```diff
@@ -19,41 +24,78 @@
...
+
+    def test_load(self):
+        snapshot_config = SnapshotConfig(
+            snapshot_dir=self.temp_dir.name,
+            snapshot_mode='all',
+            snapshot_gap=1)
+        sample_task(snapshot_config)
+        snapshotter = Snapshotter()
+
+        super().teardown_method()
```
ryanjulian: please make these different test cases, rather than using teardown/setup manually. that is an abuse of the test framework.

tests/garage/experiment/test_snapshotter.py
```diff
@@ -19,41 +24,78 @@
...
     def test_invalid_snapshot_mode(self):
         with pytest.raises(ValueError):
-            self.snapshotter.snapshot_dir = self.snapshot_dir.name
-            self.snapshotter.snapshot_mode = 'invalid'
-            self.snapshotter.save_itr_params(2, {'testparam': 'invalid'})
+            snapshotter = Snapshotter(
+                snapshot_dir=self.temp_dir.name, snapshot_mode='invalid')
+            snapshotter.save_itr_params(2, {'testparam': 'invalid'})
+
+    def test_load(self):
```
ryanjulian: please make sure all the modes are covered here (none, gap, last, gap_and_last, etc.). you can use pytest.mark.parametrize to make it easier.

tests/garage/experiment/test_snapshotter.py
```diff
@@ -19,41 +24,78 @@
 }), ('none', {})]
 
 
-class TestSnapshotter:
+def sample_task(snapshot_config):
```
ryanjulian: is there a way to make this task lighter? perhaps with a simple fixture or a mock?

examples/np/cem_cartpole.py
```diff
@@ -16,9 +16,9 @@
 from garage.tf.samplers import OnPolicyVectorizedSampler
 
 
-def run_task(*_):
+def run_task(snapshot_config, *_):
```
ryanjulian: for running a task, the LocalRunner is inside your experiment method. for resuming a task, the LocalRunner is outside your experiment method. how do you plan on resolving this API in the future?

tests/garage/experiment/test_snapshotter.py
```diff
@@ -19,41 +24,78 @@
...
+            snapshotter = Snapshotter(
+                snapshot_dir=self.temp_dir.name, snapshot_mode='invalid')
+            snapshotter.save_itr_params(2, {'testparam': 'invalid'})
+
+    def test_load(self):
+        snapshot_config = SnapshotConfig(
+            snapshot_dir=self.temp_dir.name,
+            snapshot_mode='all',
+            snapshot_gap=1)
+        sample_task(snapshot_config)
```
ryanjulian: this seems to be nearly the same integration test as test_snapshot.py -- can you keep this file as unit tests and keep only integration tests in test_snapshotter?

tests/garage/experiment/test_snapshot.py
```diff
@@ -5,7 +5,7 @@
 import joblib
```
ryanjulian: can you rename this test_snapshotter_integration.py? this is an integration test and it's kind of confusing when there are unit tests in test_snapshotter

tests/garage/experiment/test_snapshotter.py
```diff
@@ -19,41 +24,78 @@
...
+            baseline=baseline,
+            max_path_length=100,
+            discount=0.99,
+            optimizer_args=dict(tf_optimizer_args=dict(learning_rate=0.01, )))
+
+        runner.setup(algo, env)
+        runner.train(n_epochs=5, batch_size=10)
+
+
+class TestSnapshotter(TfGraphTestCase):
```
ryanjulian: please write this as a unit test which doesn't rely on classes outside Snapshotter

examples/np/cem_cartpole.py
zequnyu: I plan to move the LocalRunner for resume inside run_experiment(). I already finished it, but I'd better bring up after this.

ryanjulian: ok i think this is outside the scope of this PR, but we definitely need to sit down and plan this API for the future.

ryanjulian:
we can skip the API design issues i mentioned for this PR, but please address the comments about tests and code issues.

after this is done, let's sit down and figure out a plan for the relationship between launcher/experiment_wrapper/runner/run_experiment -- there are way too many layers and they interact in surprising ways.

examples/tf/her_ddpg_fetchreach.py
zequnyu: Because `snapshot_config` is constructed inside `experiment_wrapper`. And we pass the parameters (str, mode, gap...) to it using `run_experiment`.

examples/np/cem_cartpole.py
ryanjulian: okay great -- then ignore my comments about the API for now.

zequnyu:
> after this is done, let's sit down and figure out a plan for the relationship between launcher/experiment_wrapper/runner/run_experiment -- there are way too many layers and they interact in surprising ways.

Agree.

tests/garage/experiment/test_experiment_wrapper.py
zequnyu: Probably not as these are the only two return values.

ryanjulian: umm can't you assert what's inside SnapshotConfig?

src/garage/experiment/local_tf_runner.py
zequnyu: Are you suggesting we should let users do `from_epoch=0, from_epoch=-1` instead of `from_epoch='first', from_epoch='last'`?

tests/garage/experiment/test_snapshotter.py
```diff
@@ -19,41 +20,37 @@
...
-    def test_set_snapshot_dir(self):
-        self.snapshotter.snapshot_dir = self.snapshot_dir.name
-        assert self.snapshotter.snapshot_dir == self.snapshot_dir.name
+        super().teardown_method()
+        self.temp_dir.cleanup()
 
     @pytest.mark.parametrize('mode, files', [*configurations])
     def test_snapshotter(self, mode, files):
-        self.snapshotter.snapshot_dir = self.snapshot_dir.name
+        dir = self.temp_dir
```
ryanjulian: dir is a python reserved keyword. i'm surprised pylint is not unhappy.

zequnyu:
@naeioi @krzentner Please review this when you are free. I want to merge this before bring my next PR up.

ryanjulian:
@zequnyu please send your next PR so we can review it in parallel

tests/garage/experiment/test_snapshotter.py
```diff
@@ -19,41 +20,37 @@
 }), ('none', {})]
 
 
-class TestSnapshotter:
+class TestSnapshotter(TfGraphTestCase):
```
zhanpenghe: Does this test involve using a tensorflow graph?

zequnyu: Seems not. Updated.

Issue #587, opened 2019-03-19
Split GaussianMLPModel into several variants


ryanjulian:
The current GaussianMLPModel actually represents several architecture variants, which makes it difficult to use and to test. We should split it into several variants based on architecture (and possibly parameterization)

| description | mean | std_param |
| --- | --- | --- |
| fixed variance | MLP1 | parameter (trainable or not) |
| split network | MLP1 | MLP2 |
| shared network/multi-headed |  MLP1/head1 | MLP1/head2 |

PR #707, opened 2019-05-30
init commit for pytorch

yonghyuc:
Create directories for pytorch module


naeioi:
I have a working torch implementation of MLP and VPG a while ago. It was before the major architectural overhaul on experiment runner and `garage.tf.Model` but hopefully it might help you a bit as a reference. https://github.com/rlworkgroup/garage/pull/474/commits/606a55e495d16c37a5d134dc235eb69e4f181b5f

src/garage/torch/distributions/base.py
```diff
@@ -0,0 +1,37 @@
+class Distribution:
```
zhanpenghe: We don't need this class for pytorch. We can just use pytorch distributions. https://pytorch.org/docs/stable/distributions.html

src/garage/torch/distributions/diagonal_gaussian.py
```diff
@@ -0,0 +1,78 @@
+import torch
+from math import pi, e
+from garage.torch.distributions.base import Distribution
+
+
+class DiagonalGaussian(Distribution):
```
zhanpenghe: As the previous comment, we just need to use PyTorch distribution.

src/garage/torch/modules/mlp_module.py
```diff
@@ -0,0 +1,91 @@
...
+        output_layer = nn.Linear(in_dim, output_dim)
+        output_w_init(output_layer.weight)
+        output_b_init(output_layer.bias)
+        self.layers.append(output_layer)
+
+        if output_nonlinearity is not None:
+            self.layers.append(output_nonlinearity())
+
+    def forward(self, inputs):
+        return nn.Sequential(*self.layers)(inputs)
```
zhanpenghe: Please do not create a new module to do this. In this case a new module will be created for every forward pass.

src/garage/torch/modules/gaussian_mlp_module.py
```diff
@@ -0,0 +1,228 @@
...
+            hidden_nonlinearity=self._std_hidden_nonlinearity,
+            hidden_w_init=self._std_hidden_w_init,
+            hidden_b_init=self._std_hidden_b_init,
+            output_nonlinearity=self._std_output_nonlinearity,
+            output_w_init=self._std_output_w_init,
+            output_b_init=self._init_adaptive_b,
+            layer_normalization=self._layer_normalization)
+
+    def _init_adaptive_b(self, b):
+        return nn.init.constant_(b, self._init_std_param.item())
```
zhanpenghe: Why not just use `nn. new_full`.

src/garage/torch/distributions/base.py
yonghyuc: Thank you for letting me know.
So This means that I don't need to return "dist" from GaussianMLPModule?

src/garage/torch/modules/gaussian_mlp_module.py
yonghyuc: I just tried to be similar with the nn.init methods (zeros_, ones_, ...)
(https://pytorch.org/docs/stable/_modules/torch/nn/init.html)

Do yo think using nn.new_full is better?

src/garage/torch/distributions/base.py
yonghyuc: I'll remove the Distribution class and use Pytorch distributions.

src/garage/torch/distributions/diagonal_gaussian.py
yonghyuc: I'll remove the Distribution class and use Pytorch distributions.

src/garage/torch/modules/mlp_module.py
yonghyuc: I'll remove the MlpModule and use Linda's MlpModule

src/garage/torch/modules/gaussian_mlp_module.py
zhanpenghe: Oh sorry. I think I made a mistake. `newfull` actually does not serve our purpose here. We need to be able to initialize b to arbitrary vector.

src/garage/torch/distributions/base.py
zhanpenghe: Oh sorry for the confusion. I just meant that we don't need to write our own `Distribution` class in garage and we should be able to use Pytorch's directly and only implement those in-exist in PyTorch.

src/garage/torch/modules/mlp_module.py
```diff
@@ -81,7 +81,9 @@ def forward(self, input_val):
         """Forward method."""
         x = input_val
         for layer in self._layers[:-1]:
-            x = self._hidden_nonlinearity(layer(x))
+            x = layer(x)
+            if self._output_nonlinearity is not None:
```
zhanpenghe: Why is this checked?  I think @lywong92 is also trying to fix the output non-linearity too.

src/garage/torch/policies/gaussian_mlp_policy.py
```diff
@@ -0,0 +1,64 @@
+import torch
+from torch import nn
+from garage.torch.modules import GaussianMLPModule
+import akro
+
+class GaussianMLPPolicy(nn.Module):
```
zhanpenghe: This should inherit from `garage.np.Policy`

src/garage/torch/modules/gaussian_mlp_module.py
```diff
@@ -0,0 +1,222 @@
...
+        self._hidden_b_init = hidden_b_init
+        self._output_nonlinearity = output_nonlinearity
+        self._output_w_init = output_w_init
+        self._output_b_init = output_b_init
+        self._layer_normalization = layer_normalization
+
+        # Tranform std arguments to parameterized space
+        self._init_std_param, self._min_std_param, self._max_std_param = None, None, None
+
+        init_std = torch.Tensor([init_std])
```
zhanpenghe: Can we just pass in a single value instead of a list? What is the advantage to make it 1-dimensional instead of simple 0-dimensional?

zhanpenghe: You can use `torch.tensor` instead of `torch.Tensor` if you want to just do it in 0 dimensional.

tests/garage/torch/modules/test_gaussian_mlp_module.py
```diff
@@ -0,0 +1,319 @@
...
+            std_share_network=True,
+            hidden_nonlinearity=None,
+            std_parameterization='exp',
+            hidden_w_init=nn.init.ones_,
+            output_w_init=nn.init.ones_
+        )
+
+        action, mean, log_std, std, dist = module(torch.ones(5))
+
+        assert mean.equal(5 * (torch.Tensor(hidden_sizes).prod() + torch.zeros(output_dim)))
```
zhanpenghe: Why are we adding zeros? If they are not useful, we can just remove them.

src/garage/torch/modules/gaussian_mlp_module.py
yonghyuc: I will change torch.Tensor to torch.tensor

src/garage/torch/modules/mlp_module.py
yonghyuc: I talked with Linda about this, and I think we have some confusion about this part.
I would revert this and wait for Linda's PR

src/garage/torch/policies/gaussian_mlp_policy.py
yonghyuc: I make new Policy class under the src/garage/torch/policies/base.py
and GaussianMLPPolicy inherit the Policy class

tests/garage/torch/modules/test_gaussian_mlp_module.py
yonghyuc: It is more like np.full
torch.Tensor(hidden_sizes).prod() is scalar and torch.zeros(output_dim) is matrix,
so it works like element-wise addition.
But I think this is quite confusing.. so I would change this

src/garage/torch/modules/mlp_module.py
yonghyuc: I need this changes for the testcases of the GaussianMLPModule.
So I think I would wait until @lywong92 update this code

src/garage/torch/policies/base.py
```diff
@@ -0,0 +1,53 @@
+
+
+class Policy:
```
krzentner: The rest of garage uses the `abc` module to define base classes, please do so here as well.
An example is [here](https://github.com/rlworkgroup/garage/blob/8b1bac869ac2bb167aa9982c80ee968e662b9f5a/src/garage/sampler/base.py).

src/garage/torch/policies/base.py
```diff
@@ -0,0 +1,53 @@
...
+    def __init__(self, env_spec):
+        self._env_spec = env_spec
+        self._variable_scope = None
+
+    # Should be implemented by all policies
+    def get_action(self, observation):
+        """Get action given observation."""
+        raise NotImplementedError
+
+    def get_actions(self, observations):
```
krzentner: Why not define this to just call `get_action` on each sub-array?
That way it will work, even on non-vectorized policies.

src/garage/torch/policies/base.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+    Policy base class without Parameterzied.
+
+    Args:
+        env_spec (garage.envs.env_spec.EnvSpec): Environment specification.
+
+    """
+    def __init__(self, env_spec):
+        self._env_spec = env_spec
+        self._variable_scope = None
```
zhanpenghe: Hmm there is no variable scope in pytorch..

src/garage/torch/policies/gaussian_mlp_policy.py
```diff
@@ -0,0 +1,73 @@
...
+            std_hidden_nonlinearity=std_hidden_nonlinearity,
+            std_output_nonlinearity=std_output_nonlinearity,
+            std_parameterization=std_parameterization,
+            layer_normalization=layer_normalization
+        )
+
+    def forward(self, inputs):
+        return self._model(inputs)
+
+    def get_action(self, observations):
```
zhanpenghe: I think here we want to be clear whether these should be symbolic or not. IMO we should handle all the symbolic stuff with `forward` so I think these should be calling `detach()` to stop any gradients.

@krzentner  what do you think.

krzentner: Oh yeah, these should actually return `numpy` arrays, so they should call `.detach().numpy()`.

src/garage/torch/policies/base.py
```diff
@@ -0,0 +1,53 @@
+
```
krzentner: Actually, this class should probably go in src/garage/np/policies/base.py. Perhaps we can put it there, and call it `PolicyBase`. @zhanpenghe, what do you think?

We're fairly close to deleting `Parameterized`, so the old `Policy` class shouldn't really be needed. And Tensorflow specific stuff in that class (e.g. the `_variable_scope`) should probably be moved to the `tf` branch.

src/garage/torch/policies/base.py
yonghyuc: I am just wondering there is no abc.ABC in [Policy](https://github.com/rlworkgroup/garage/blob/8b1bac869ac2bb167aa9982c80ee968e662b9f5a/src/garage/tf/policies/base.py#L4) and [Policy2](https://github.com/rlworkgroup/garage/blob/8b1bac869ac2bb167aa9982c80ee968e662b9f5a/src/garage/tf/policies/base2.py#L7) class
Is there any difference with others?

src/garage/torch/policies/base.py
yonghyuc: Oh. Ok. I will change the code

src/garage/torch/policies/base.py
yonghyuc: I will remove the _variable_scope

src/garage/torch/policies/base.py
krzentner: I meant this:
```python
return np.stack([self.get_action(obs) for obs in observations])
```

src/garage/torch/policies/base.py
krzentner: I think that the main reason is just that those two are rather old, before we started using `abc`.
One could also argue that's there's a lot of code in `Policy2`, and only a few methods to overwrite, so it's not *really* a base class.

I think we're moving away from using `garage.tf.policies.Policy`, since it uses `Parameterized`.

src/garage/torch/policies/base.py
yonghyuc: Do I need to move and rename the policy class?
@krzentner @zhanpenghe

src/garage/torch/policies/base.py
yonghyuc: ok thanks

src/garage/torch/policies/base.py
yonghyuc: thanks! I will put abc.ABC and "@abc.abstractclassmethod"

src/garage/torch/modules/gaussian_mlp_module.py
```diff
@@ -0,0 +1,222 @@
...
+            else:
+                broadcast_shape = list(inputs.shape[:-1]) + [self._action_dim]
+                x = torch.zeros(*broadcast_shape)
+                std = x + self._init_std_param
+
+        if self._min_std_param or self._max_std_param:
+            std = std.clamp(min=to_scalar(self._min_std_param), max=to_scalar(self._max_std_param))
+
+        if self._std_parameterization == 'exp':
+            log_std_var = std
```
zhanpenghe: Shouldn't this be `std.log()`?

src/garage/torch/modules/gaussian_mlp_module.py
```diff
@@ -0,0 +1,222 @@
...
+        else:
+            log_std_var = std.exp().add(1.).log().log()
```
zhanpenghe: What is this actually doing?
Please use `elif` to be more specific then raise `NotImplementedError` for all other cases.

src/garage/torch/modules/gaussian_mlp_module.py
yonghyuc: I followed tensorflow [GaussianMLPModule](https://github.com/rlworkgroup/garage/blob/3bf0b2b40c3db30c66f95460f0280b442457cc7c/src/garage/tf/models/gaussian_mlp_model.py#L215).
but I think you are right.. could you check the tensorflow version too?

src/garage/torch/modules/gaussian_mlp_module.py
yonghyuc: I think we already check the std_parameterization at __init__
and also followed tensorflow [GaussianMLPModule](https://github.com/rlworkgroup/garage/blob/3bf0b2b40c3db30c66f95460f0280b442457cc7c/src/garage/tf/models/gaussian_mlp_model.py#L219).
If you think it needs to be fixed, I will change the code

src/garage/torch/modules/gaussian_mlp_module.py
zhanpenghe: I think there is a difference here. They assume that the neural network output is always log std but here your output is a std. I think we should also do that too. But to make the naming more clear: name the output of neural network as `log_std` and name the after-parameterized variable as `std_param`. We should just change to use `std_param` everywhere since it does not have to be a log std.

The naming in the tf tree is a bit confusing. @nish21 @ahtsan, I suggest fix this in the future?

zhanpenghe: Oh sorry, I made a mistake. Didn't notice that the second log in the softplus...

@yonghyuc ok the naming is fine. Just change the naming of model output to log_std..

src/garage/torch/modules/gaussian_mlp_module.py
zhanpenghe: Ok thats fine

src/garage/torch/modules/gaussian_mlp_module.py
yonghyuc: you mean change the name "log_std_var" to "log_std"?

src/garage/torch/policies/gaussian_mlp_policy.py
```diff
@@ -0,0 +1,81 @@
...
+            layer_normalization=layer_normalization
+        )
+
+    def forward(self, inputs):
+        return self._model(inputs)
+
+    def get_action(self, observation):
+        action_var, mean, log_std_var, std, dist = self.forward(observation)
+
+        return action_var[0].detach().cpu().numpy(), \
```
zhanpenghe: Hmm lets not force it to use cpu..

src/garage/torch/policies/gaussian_mlp_policy.py
```diff
@@ -0,0 +1,81 @@
...
+               dict(
+                   mean=mean[0].detach().cpu().numpy(),
+                   log_std=log_std_var[0].detach().cpu().numpy()
+               )
+
+    def get_actions(self, observations):
+        action_var, mean, log_std_var, std, dist = self.forward(observations)
+
+        return action_var.detach().cpu().numpy(),\
```
zhanpenghe: Don't force it to use cpu

src/garage/torch/policies/gaussian_mlp_policy.py
```diff
@@ -0,0 +1,81 @@
...
+               dict(
+                   mean=mean.detach().cpu().numpy(),
```
zhanpenghe: Dont force it to use cpu

src/garage/torch/modules/gaussian_mlp_module.py
zhanpenghe: std -> log_std

yonghyuc: and.. if the output from std_network is log_std..
in the softplus part, isn't the equation should be` log ( log ( 1 + e^ (e^x) ) )`?
because x is log_std not std so do we need to exponentiate the x?

src/garage/torch/policies/gaussian_mlp_policy.py
yonghyuc: I am just wondering, we don't need to consider cuda variable?
(https://discuss.pytorch.org/t/convert-to-numpy-cuda-variable/499)

zhanpenghe: We do. But we want to handle this globally so that when changing device, we dont need to replace cpu to cuda everywhere.

yonghyuc: oh ok got it!

src/garage/torch/policies/gaussian_mlp_policy.py
yonghyuc: fixed

src/garage/torch/policies/gaussian_mlp_policy.py
yonghyuc: fixed

src/garage/torch/modules/gaussian_mlp_module.py
yonghyuc: Talked with Zhanpeng and decided to change the code into `log ( log ( 1 + e^ (e^x) ) )`

src/garage/torch/modules/gaussian_mlp_module.py
```diff
@@ -0,0 +1,223 @@
...
+                 input_dim,
+                 output_dim,
+                 hidden_sizes=(32, 32),
+                 hidden_nonlinearity=torch.tanh,
+                 hidden_w_init=nn.init.xavier_uniform_,
+                 hidden_b_init=nn.init.zeros_,
+                 output_nonlinearity=None,
+                 output_w_init=nn.init.xavier_uniform_,
+                 output_b_init=nn.init.zeros_,
+                 learn_std=True,
```
krzentner: This argument is ignored. As far as I can tell, it should determine if `_init_std` is a `torch.nn.Parameter` or just a `torch.tensor`?

src/garage/torch/modules/gaussian_mlp_module.py
```diff
@@ -0,0 +1,223 @@
...
+        self._init_std_param, self._min_std_param, self._max_std_param = None, None, None
+
+        init_std = torch.tensor(init_std, dtype=torch.float)
+        if min_std:
+            min_std = torch.tensor(min_std, dtype=torch.float)
+        if max_std:
+            max_std = torch.tensor(max_std, dtype=torch.float)
+
+        if self._std_parameterization == 'exp':
+            self._init_std_param = init_std.log()
```
krzentner: Since this is a `torch.tensor`, not a `torch.nn.Parameter`, it won't be trained, even if `learn_std` is True.

src/garage/torch/modules/gaussian_mlp_module.py
```diff
@@ -0,0 +1,223 @@
...
+            self._min_std_param = min_std.log() if min_std else None
+            self._max_std_param = max_std.log() if max_std else None
+
+        elif self._std_parameterization == 'softplus':
+            self._init_std_param = init_std.exp().add(-1).log()
```
krzentner: Same here.

src/garage/torch/policies/base.py
krzentner: As far as I can tell, the policy class in [src/garage/np/policies/base.py](https://github.com/rlworkgroup/garage/blob/master/src/garage/np/policies/base.py) is only ever used in a few tests, and doesn't use `Parameterized` at all. We can move it in a later PR though, in case doing so causes other breakages.

src/garage/torch/modules/gaussian_mlp_module.py
```diff
@@ -0,0 +1,223 @@
...
+        else:
+            log_std_var = log_std.exp().exp().add(1.).log().log()
+
+        rnd = torch.rand(list(mean.shape)[1:])
+        action_var = rnd * log_std_var.exp() + mean
+
+        cov = (log_std_var.exp() ** 2).diag_embed()
+        dist = MultivariateNormal(mean, cov)
+
+        return action_var, mean, log_std_var, log_std, dist
```
krzentner: Please drop the `_var` postfix from `action_var` and `log_std_var`, since that postfix only really made sense in Tensorflow. It's confusing, since it might refer to "variance" but does not. The original `log_std` probably isn't needed.

Alternatively, you could rename `log_std` to `log_std_uncentered` or something.

src/garage/torch/policies/gaussian_mlp_policy.py
```diff
@@ -0,0 +1,81 @@
...
+            std_output_nonlinearity=std_output_nonlinearity,
+            std_parameterization=std_parameterization,
+            layer_normalization=layer_normalization
+        )
+
+    def forward(self, inputs):
+        return self._model(inputs)
+
+    def get_action(self, observation):
+        action_var, mean, log_std_var, std, dist = self.forward(observation)
```
krzentner: Please drop the `_var` postfix.

src/garage/torch/policies/gaussian_mlp_policy.py
```diff
@@ -0,0 +1,81 @@
...
+
+        return action_var[0].detach().numpy(), \
+               dict(
+                   mean=mean[0].detach().numpy(),
+                   log_std=log_std_var[0].detach().numpy()
+               )
+
+    def get_actions(self, observations):
+        action_var, mean, log_std_var, std, dist = self.forward(observations)
```
krzentner: Please drop the `_var` prefix.

src/garage/torch/modules/gaussian_mlp_module.py
yonghyuc: hmm I assume init_std is float type. like other min_std and max_std.
Do I need to get init_std as torch object(Parameter or tensor)?

src/garage/torch/modules/gaussian_mlp_module.py
yonghyuc: thanks! fixed!

src/garage/torch/policies/gaussian_mlp_policy.py
yonghyuc: fixed

src/garage/torch/policies/gaussian_mlp_policy.py
yonghyuc: fixed

src/garage/torch/modules/gaussian_mlp_module.py
yonghyuc: Got it!

src/garage/torch/modules/gaussian_mlp_module.py
yonghyuc: fixed

src/garage/torch/modules/gaussian_mlp_module.py
yonghyuc: fixed

src/garage/torch/modules/gaussian_mlp_module.py
yonghyuc: Make default value of init_std as torch.nn.Parameter because in TF version, default value if  learn_std is True

krzentner: I would just wrap `init_std` when you assign it to `self._init_std_param`, but otherwise yeah.

src/garage/torch/policies/gaussian_mlp_policy.py
krzentner: BTW, `std` here should probably be renamed, since it's not the standard deviation.

src/garage/torch/modules/gaussian_mlp_module.py
ryanjulian: actually, please only return the dist (it should be sufficient for anyone's purposes). we are working on that refactoring in TF -- no need to add technical debt in torch.

src/garage/torch/modules/gaussian_mlp_module.py
```diff
@@ -0,0 +1,223 @@
...
+
+
+def to_scalar(tensor):
```
ryanjulian: this should be in a utility library or not at all. is this not a helper in torch already?

ryanjulian: oh i see. i think it is actually clearer to write this out. it doesn't convert a tensor to a scalar, it implements a default value.

src/garage/torch/policies/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""Pytorch modules."""
```
ryanjulian: PyTorch

src/garage/torch/policies/base.py
```diff
@@ -0,0 +1,54 @@
...
+    def get_action(self, observation):
+        """Get action given observation."""
+        pass
+
+    def get_actions(self, observations):
+        """Get actions given observations."""
+        return np.stack([self.get_action(obs) for obs in observations])
+
+    @property
+    def vectorized(self):
```
ryanjulian: i think we should remove this flag and assume all policies implement get_actions(). if they can't be vectorized naturally, they can implement it with a for loop.

src/garage/torch/policies/base.py
```diff
@@ -0,0 +1,54 @@
...
+        """Policy action space."""
+        return self._env_spec.action_space
+
+    @property
+    def env_spec(self):
+        """Policy environment specification."""
+        return self._env_spec
+
+    @property
+    def recurrent(self):
```
ryanjulian: hmm is this API really needed? should it be called stateful instead? what do we do differently if something is recurrent?

src/garage/torch/policies/base.py
```diff
@@ -0,0 +1,54 @@
...
+        """Observation space."""
+        return self._env_spec.observation_space
+
+    @property
+    def action_space(self):
+        """Policy action space."""
+        return self._env_spec.action_space
+
+    @property
+    def env_spec(self):
```
ryanjulian: is this API actually used?

src/garage/torch/policies/gaussian_mlp_policy.py
```diff
@@ -0,0 +1,81 @@
...
+            layer_normalization=layer_normalization
+        )
+
+    def forward(self, inputs):
+        return self._model(inputs)
+
+    def get_action(self, observation):
+        action_var, mean, log_std_var, std, dist = self.forward(observation)
+
+        return action_var[0].detach().numpy(), \
```
ryanjulian: please split by wrapping the line in (), per PEP8, rather than using \

tests/garage/torch/modules/test_gaussian_mlp_module.py
```diff
@@ -0,0 +1,368 @@
...
+from unittest import mock
+import torch
+from torch import nn
+from nose2.tools.params import params
+
+
+class TestGaussianMLPmodule(unittest.TestCase):
+
+    @mock.patch('torch.rand')
+    @params((5, 1, (1,)), (5, 1, (2,)), (5, 2, (3,)), (5, 2, (1, 1)), (5, 3, (2, 2)))
```
ryanjulian: for parameterized tests, please format these to have one set of arguments (tuple) per line. you may need to use `# yapf: disable`/`# yapf: enable`

tests/garage/torch/modules/test_gaussian_mlp_module.py
```diff
@@ -0,0 +1,368 @@
+import pickle
+from garage.torch.modules.gaussian_mlp_module import GaussianMLPModule
+import unittest
+from unittest import mock
+import torch
+from torch import nn
+from nose2.tools.params import params
```
ryanjulian: please see the [updated CONTRIBUTING.md](https://github.com/rlworkgroup/garage/pull/678) for conventions on imports. these external imports should use full paths.

ryanjulian: your imports also need to be grouped properly. this is also covered in CONTRIBUTING.md

src/garage/torch/policies/__init__.py
yonghyuc: fixed

src/garage/torch/modules/gaussian_mlp_module.py
yonghyuc: change name of the method to 'to_scalar_if_not_none'

src/garage/torch/policies/gaussian_mlp_policy.py
yonghyuc: fixed

src/garage/torch/policies/gaussian_mlp_policy.py
yonghyuc: fixed

src/garage/torch/modules/gaussian_mlp_module.py
yonghyuc: fixed

src/garage/torch/policies/gaussian_mlp_policy.py
yonghyuc: fixed

src/garage/torch/modules/gaussian_mlp_module.py
yonghyuc: Is it ok to use requires_grad?
like
`self._init_std_param = torch.tensor(init_std, dtype=torch.float, requires_grad=self._learn_std)`

krzentner: That might break other things in the future, since it e.g. won't be listed in the `state_dict` just use:
```python
self._init_std_param = nn.Parameter(init_std)
```

src/garage/torch/policies/base.py
krzentner: Yeah, I agree that's the right way to go.

src/garage/torch/policies/base.py
krzentner: Yeah, I agree we can delete this, but now I realize we probably need a `reset()` method on this type.

tests/garage/torch/modules/test_gaussian_mlp_module.py
yonghyuc: fixed

tests/garage/torch/modules/test_gaussian_mlp_module.py
yonghyuc: fixed

src/garage/torch/policies/gaussian_mlp_policy.py
yonghyuc: fixed

src/garage/torch/policies/base.py
yonghyuc: fixed

src/garage/torch/modules/gaussian_mlp_module.py
yonghyuc: ok, fixed

src/garage/torch/policies/base.py
yonghyuc: removed

src/garage/torch/policies/base.py
yonghyuc: removed recurrent function

src/garage/torch/modules/gaussian_mlp_module.py
yonghyuc: only return dist, this dist contains mean and variance. Also we can get actions using rsample.

src/garage/torch/modules/gaussian_mlp_module.py
```diff
@@ -0,0 +1,231 @@
...
+        output_w_init (callable): Initializer function for the weight
+            of output dense layer(s). The function should return a
+            torch.Tensor.
+        output_b_init (callable): Initializer function for the bias
+            of output dense layer(s). The function should return a
+            torch.Tensor.
+        learn_std (bool): Is std trainable.
+        init_std (float): Initial value for std.
+            > std_parameterization
+                'exp' : plain value (not log or exponentiated)
```
ryanjulian: this parameter should accept an unparameterized value for std, and convert to the appropriate parameterization.

src/garage/torch/modules/gaussian_mlp_module.py
```diff
@@ -0,0 +1,231 @@
...
+        adaptive_std (bool): Is std a neural network. If False, it will be a
+            parameter.
+        std_share_network (bool): Boolean for whether mean and std share
+            the same network.
+        std_hidden_sizes (list[int]): Output dimension of dense layer(s) for
+            the MLP for std. For example, (32, 32) means the MLP consists
+            of two hidden layers, each with 32 hidden units.
+        min_std (float): If not None, the std is at least the value of min_std,
+            to avoid numerical issues (normal value).
+            > std_parameterization
```
ryanjulian: again, these values should be converted automatically

src/garage/torch/modules/gaussian_mlp_module.py
```diff
@@ -0,0 +1,231 @@
...
+            hidden_sizes=self._hidden_sizes,
+            hidden_nonlinearity=self._hidden_nonlinearity,
+            hidden_w_init=self._hidden_w_init,
+            hidden_b_init=self._hidden_b_init,
+            output_nonlinearity=self._output_nonlinearity,
+            output_w_init=self._output_w_init,
+            output_b_init=self._output_b_init,
+            layer_normalization=self._layer_normalization)
+
+    def _get_adaptive_std_network(self):
```
ryanjulian: `get` is usually used for properties, perhaps you can use the verb `build` or `make`?

src/garage/torch/modules/gaussian_mlp_module.py
```diff
@@ -0,0 +1,231 @@
...
+        elif self._std_parameterization == 'softplus':
+            self._init_std_param = self._init_std_param.exp().add(-1).log().log()
+            self._min_std_param = min_std.exp().add(-1).log().log() if min_std else None
+            self._max_std_param = max_std.exp().add(-1).log().log() if max_std else None
+
+        else:
+            raise NotImplementedError
+
+        if std_share_network:
+            self.shared_mean_log_std_network = self._get_shared_mean_std_network()
```
ryanjulian: is there a reason that this attribute is public?

src/garage/torch/modules/gaussian_mlp_module.py
```diff
@@ -0,0 +1,231 @@
...
+        else:
+            self.mean_module = self._get_mean_network()
```
ryanjulian: is there a reason this is public?

src/garage/torch/modules/gaussian_mlp_module.py
```diff
@@ -0,0 +1,231 @@
...
+
+            if adaptive_std:
+                self.adaptive_log_std_module = self._get_adaptive_std_network()
```
ryanjulian: why public?

src/garage/torch/modules/gaussian_mlp_module.py
yonghyuc: You mean this module have to accept only plain std value?

src/garage/torch/modules/gaussian_mlp_module.py
yonghyuc: fixed

src/garage/torch/modules/gaussian_mlp_module.py
yonghyuc: no, fixed.

src/garage/torch/modules/gaussian_mlp_module.py
yonghyuc: fixed

src/garage/torch/modules/gaussian_mlp_module.py
yonghyuc: fixed

src/garage/torch/modules/gaussian_mlp_module.py
ryanjulian: @yonghyuc if this is true, it is a bug. @ahtsan can you check?

if the init_std/min_std/max_std are being processed wrong in TF, please file an issue and assign it to Anson. It should be easy to fix.

tests/garage/torch/modules/test_gaussian_mlp_module.py
```diff
@@ -0,0 +1,362 @@
...
+    module_pickled = pickle.loads(h)
+
+    dist_p = module_pickled(input)
+
+    assert dist.mean.equal(dist_p.mean)
+    assert dist.variance.equal(dist_p.variance)
+    assert dist.rsample().shape == dist_p.rsample().shape
+
+
+@pytest.mark.xfail(raises=NotImplementedError)
```
ryanjulian: can you instead use pytest.raises?

src/garage/torch/policies/gaussian_mlp_policy.py
```diff
@@ -0,0 +1,139 @@
+import torch
+from torch import nn
+
+from garage.torch.modules import GaussianMLPModule
+from garage.torch.policies import Policy
+
+class GaussianMLPPolicy(nn.Module, Policy):
```
ryanjulian: this class needs to expose its underlying distribution so that we can access differentiable paths through dist.entropy, dist.log_likelihood, etc.

tests/garage/torch/modules/test_gaussian_mlp_module.py
```diff
@@ -0,0 +1,362 @@
...
+        std_share_network=True,
+        hidden_nonlinearity=None,
+        std_parameterization='exp',
+        hidden_w_init=nn.init.ones_,
+        output_w_init=nn.init.ones_
+    )
+
+    dist = module(torch.ones(5))
+
+    exp_mean = torch.full((output_dim, ), 5 * (torch.Tensor(hidden_sizes).prod().item()))
```
ryanjulian: shouldn't this be np.full, etc? the mean and variance are numpy types.

src/garage/torch/modules/gaussian_mlp_module.py
ahtsan: @ryanjulian @yonghyuc We talked and agreed that it is true and it is a bug. I will file an issue about it.

src/garage/torch/policies/gaussian_mlp_policy.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def forward(self, inputs):
+        return self._model(inputs)
+
+    def get_action(self, observation):
+        dist = self.forward(observation)
+
+        return (dist.rsample()[0].detach().numpy(),
+                dict(
+                    mean=dist.mean[0].detach().numpy(),
```
ryanjulian: this style of returning the mean and log_std in get_action is an outdated part of TF which we are removing. please return only the action.

src/garage/torch/policies/gaussian_mlp_policy.py
```diff
@@ -0,0 +1,139 @@
...
+                    log_std=dist.variance[0].sqrt().detach().numpy()
+                ))
+
+    def get_actions(self, observations):
+        dist = self.forward(observations)
+
+        return (dist.rsample().detach().numpy(),
+                dict(
+                    mean=dist.mean.detach().numpy(),
+                    log_std=dist.variance.sqrt().detach().numpy()
```
ryanjulian: please remove mean/log_std.

src/garage/torch/policies/gaussian_mlp_policy.py
ryanjulian: please provide a more modular way of defining (several types) of GaussianMLPModels.

if you look closely, this module implements 3-5 different models which switch between configurations based on the constructor. instead, we can just have 3-5 different classes which mostly rely on the same base class.

@lywong92 has a similar idea going on in #764. I want to avoid creating technical debt in torch which we are trying to fix in https://github.com/rlworkgroup/garage/issues/587

your implementation can serve as a template for the TF version.

tests/garage/torch/modules/test_gaussian_mlp_module.py
```diff
@@ -0,0 +1,362 @@
+import pickle
+
+import pytest
+import torch
+from torch import nn
+
+from garage.torch.modules.gaussian_mlp_module import GaussianMLPModule
+
+plain_settings = [
+    (5, 1, (1,)),
```
ryanjulian: please try a larger range of input_dims. problems often occur around the values 1, 2, and more than 2.

ryanjulian: i think the ranges used in https://github.com/rlworkgroup/garage/blob/master/tests/garage/tf/models/test_gaussian_mlp_model.py should be sufficient.

tests/garage/torch/modules/test_gaussian_mlp_module.py
```diff
@@ -0,0 +1,362 @@
...
+
+    dist = module(torch.ones(5))
+
+    exp_variance = max_value ** 2
+
+    assert dist.variance.equal(torch.full((output_dim, ), exp_variance))
+
+
+@pytest.mark.parametrize("input_dim, output_dim, hidden_sizes", plain_settings)
+def test_std_network_output_values_pickable(input_dim, output_dim, hidden_sizes):
```
ryanjulian: these pickleable tests may not be necessary if pickling is provided by the torch API.

@krzentner what do you think?

src/garage/torch/policies/base.py
ryanjulian: i agree -- please just correct the one in np.

src/garage/torch/policies/gaussian_mlp_policy.py
```diff
@@ -0,0 +1,139 @@
...
+            std_hidden_nonlinearity=std_hidden_nonlinearity,
+            std_output_nonlinearity=std_output_nonlinearity,
+            std_parameterization=std_parameterization,
+            layer_normalization=layer_normalization
+        )
+
+    def forward(self, inputs):
+        return self._model(inputs)
+
+    def get_action(self, observation):
```
ryanjulian: this has a very predictable implementation for pretty much all vectorized policies -- can it just be in a base class?

```python
    return np.squeeze(self.get_actions(np.expand_dims(observation, axis=0)))
```

tests/garage/torch/policies/test_gaussian_mlp_policy.py
```diff
@@ -0,0 +1,100 @@
...
+    policy_pickled = pickle.loads(h)
+
+    sample_p, dist_info_p = policy_pickled.get_actions(input)
+
+    assert np.array_equal(sample.shape, sample_p.shape)
+    assert np.array_equal(dist_info['mean'], dist_info_p['mean'])
+    assert np.array_equal(dist_info['log_std'], dist_info_p['log_std'])
+
+
+class TestEnv():
```
ryanjulian: there are existing fixtures like this in tests/fixtures. you can also use Mock (preferable)

tests/garage/torch/policies/test_gaussian_mlp_policy.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+@pytest.mark.parametrize(
+    "input_dim,output_dim,hidden_sizes",
+    [(5, 1, (1,)),
+     (5, 1, (2,)),
+     (5, 2, (3,)),
+     (5, 2, (1, 1)),
+     (5, 3, (2, 2))])
+def test_policy_is_picklable(input_dim, output_dim, hidden_sizes):
```
ryanjulian: not sure this is necessary @krzentner WDYT?

tests/garage/torch/policies/test_gaussian_mlp_policy.py
```diff
@@ -0,0 +1,100 @@
...
+
+import numpy as np
+import pytest
+import torch
+
+from garage.torch.policies import GaussianMLPPolicy
+
+
+@mock.patch('garage.torch.policies.gaussian_mlp_policy.GaussianMLPModule')
+def test_policy_get_actions(mock_model):
```
ryanjulian: please parameterize all your tests cases. you can use https://github.com/rlworkgroup/garage/blob/master/tests/garage/tf/policies/test_gaussian_mlp_policy_with_model.pyy as a guide.

tests/garage/torch/modules/test_gaussian_mlp_module.py
yonghyuc: chane to pytest.raises

src/garage/torch/modules/gaussian_mlp_module.py
ryanjulian: ok. please file the issue. preferably fix before release if it's not huge.

tests/garage/torch/modules/test_gaussian_mlp_module.py
yonghyuc: the mean and variance from the distribution are torch.tensor type.
I think mean and variance from the GaussianMLPPolicy was numpy type.(currently removed)
So, I compare the mean and variance with torch.tensor type expected value

src/garage/torch/policies/gaussian_mlp_policy.py
yonghyuc: removed

src/garage/torch/policies/gaussian_mlp_policy.py
yonghyuc: removed

tests/garage/torch/modules/test_gaussian_mlp_module.py
yonghyuc: fixed

tests/garage/torch/policies/test_gaussian_mlp_policy.py
yonghyuc: I tried to use the Mock, but if I used mock object as a test_env, 
I got the error 
_pickle.PicklingError: Can't pickle <class 'unittest.mock.MagicMock'>: it's not the same object as unittest.mock.MagicMock

src/garage/torch/policies/gaussian_mlp_policy.py
yonghyuc: Now, users can get the distribution using forward function.
Do you mean the Policy should cache the distribution and return the information?

yonghyuc: So... you means dividing the current GaussianMLPModule into 3 different classes?
for example, GaussianSharedMlpModule, GaussianAdaptiveMlpModule, GaussianMlpModule?

src/garage/torch/policies/gaussian_mlp_policy.py
yonghyuc: Currently, I do opposite way

`
    def get_actions(self, observations):
        """Get actions given observations."""
        return np.stack([self.get_action(obs) for obs in observations])
`

do you think we need to change this code?

src/garage/torch/modules/gaussian_mlp_module.py
yonghyuc: fixed

src/garage/torch/modules/gaussian_mlp_module.py
yonghyuc: fixed

tests/garage/torch/policies/test_gaussian_mlp_policy.py
yonghyuc: fixed

src/garage/torch/policies/gaussian_mlp_policy.py
ryanjulian: thanks for pointing that out. let's try with the forward-only implementation for now.

ryanjulian: correct. but you should also provide concrete implementation which takes parameters, and produces a policy by constructing the correct modules and passing them.

src/garage/torch/policies/gaussian_mlp_policy.py
ryanjulian: yes -- you want to use a vectorized feed to get_actions, not a for loop

@krzentner will torch do this automatically or does he need to feed a batch?

tests/garage/torch/modules/test_gaussian_mlp_module.py
ryanjulian: this is a lot of ranges (a lot more than TF) -- perhaps pick some covering subset? we are pushing how long the CI can run right now.

tests/garage/torch/policies/test_gaussian_mlp_policy.py
```diff
@@ -0,0 +1,104 @@
...
+    (1, 2, (2,)),
+    (1, 3, (3,)),
+    (1, 1, (1, 2)),
+    (1, 2, (2, 1)),
+    (1, 3, (4, 5)),
+    (2, 1, (1,)),
+    (2, 2, (2,)),
+    (2, 3, (3,)),
+    (2, 1, (1, 2)),
+    (2, 2, (2, 1)),
```
ryanjulian: be careful about how long the test takes to run.

yonghyuc: on my local machine, it was just taking a few seconds.
so I think it would be fine

tests/garage/torch/modules/test_gaussian_mlp_module.py
yonghyuc: on my local machine, it was just taking a few seconds.
so I think it would be fine

src/garage/torch/policies/gaussian_mlp_policy.py
yonghyuc: I changed the GaussianMLPModules (GaussianMLPModule, GaussianMLPAdaptiveModule, GaussianMLPSharedModule) and GaussianMLPPolicy.

Could you check the changes?

src/garage/torch/modules/gaussian_mlp_module.py
```diff
@@ -0,0 +1,255 @@
...
+
+class GaussianMLPSharedModule(GaussianMLPBaseModule):
+    """
+    GaussianMLPModule which has only one mean network.
+    Std will be a parameter
+    """
+    def __init__(self,
+                 input_dim,
+                 output_dim,
+                 **kwargs):
```
ryanjulian: please never use **kwargs except in very special code. all args you pass along should be declared.

src/garage/torch/modules/gaussian_mlp_module.py
```diff
@@ -0,0 +1,255 @@
...
+
+
+class GaussianMLPModule(GaussianMLPBaseModule):
+    """
+    GaussianMLPModule that mean and std share the same network.
+    """
+    def __init__(self,
+                 input_dim,
+                 output_dim,
+                 **kwargs):
```
ryanjulian: please don't swallow kwargs

src/garage/torch/modules/gaussian_mlp_module.py
```diff
@@ -0,0 +1,255 @@
...
+            std = log_std_uncentered.exp()
+        else:
+            std = log_std_uncentered.exp().exp().add(1.).log()
+
+        cov = (std**2).diag_embed()
+        dist = MultivariateNormal(mean, cov)
+
+        return dist
+
+    def to_scalar_if_not_none(self, tensor):
```
ryanjulian: @krzentner WDTY of this?

src/garage/torch/modules/gaussian_mlp_module.py
```diff
@@ -0,0 +1,255 @@
+import abc
+
+import torch
+from torch import nn
+from torch.distributions import MultivariateNormal
+
+from garage.torch.modules.mlp_module import MLPModule
+
+
+class GaussianMLPBaseModule(abc.ABC, nn.Module):
```
ryanjulian: what about the configuration where std is just a flat parameter?

src/garage/torch/modules/gaussian_mlp_module.py
```diff
@@ -0,0 +1,255 @@
...
+            output_w_init=self._output_w_init,
+            output_b_init=self._init_shared_b,
+            layer_normalization=self._layer_normalization
+        )
+
+    def _init_shared_b(self, b):
+        nn.init.zeros_(b[..., :self._action_dim])
+        nn.init.constant_(b[..., self._action_dim:], self._init_std_param.item())
+
+    def _get_mean_and_log_std(self, inputs):
```
ryanjulian: can you actually implement this as a two-headed MLP? the TF implementation (which i wrote) isn't very nice.

basically what you want is an MLP module with `len(hidden_sizes)` hidden layers, but 2 different output layers. you can implement `MultiHeadedMLPModule` to do this. it will take in a parameter `n_heads` and return `n` outputs, with an output layer for each n.

this will be really useful in the future for many architectures.

--then you can just do
```python
def __init__(self, ...):
    self._shared_mean_log_std_network =  MultiHeadedMLPModule(n_heads=2, ...)

def _get_mean_and_log_std(self, input):
    return self._shared_mean_log_std_network(input)
```

ryanjulian: the topology for `n` hidden layers and `k` output heads is
```
[input] --> [hidden_1] --> ... --> [hidden_n] --> | --> [output_head_1]
                                                  |...
                                                  | --> [output_head_k]
```

note that you need to support a different `output_b_init` and `output_w_init` for each output head. that's pretty easy -- just have inputs `output_b_inits` and `output_w_inits`. for simplicity, you can still support `output_b_init`/`output_w_init` and use that for all heads (but only allow the user to specify one at at time)

src/garage/torch/policies/gaussian_mlp_policy.py
ryanjulian: checked

src/garage/torch/modules/__init__.py
```diff
@@ -1,4 +1,5 @@
 """Pytorch modules."""
 from garage.torch.modules.mlp_module import MLPModule
+from garage.torch.modules.gaussian_mlp_module import GaussianMLPModule, GaussianMLPAdaptiveModule, GaussianMLPSharedModule
 
-__all__ = ['MLPModule']
+__all__ = ['MLPModule', 'GaussianMLPModule', 'GaussianMLPAdaptiveModule', 'GaussianMLPSharedModule']
```
ryanjulian: I suggest `GaussianMLPIndependentStdModule` and `GaussianMLPTwoHeadedModule`

tests/garage/torch/policies/test_gaussian_mlp_policy.py
```diff
@@ -0,0 +1,105 @@
...
+
+    h = pickle.dumps(policy)
+    policy_pickled = pickle.loads(h)
+
+    sample_p = policy_pickled.get_actions(input)
+
+    assert np.array_equal(sample.shape, sample_p.shape)
+
+
+class EnvForTest:
```
ryanjulian: we have existing fixtures for this in tests/fixtures. I also think mocks would probably work fine.

src/garage/torch/modules/gaussian_mlp_module.py
yonghyuc: fixed

src/garage/torch/modules/gaussian_mlp_module.py
yonghyuc: fixed

src/garage/torch/modules/__init__.py
yonghyuc: fixed

src/garage/torch/policies/base.py
zhanpenghe: I agree too.

src/garage/torch/modules/gaussian_mlp_module.py
yonghyuc: Implemented MultiHeadedMLPModule

src/garage/torch/policies/gaussian_mlp_policy.py
yonghyuc: fixed

tests/garage/torch/policies/test_gaussian_mlp_policy.py
krzentner: I think the test is a good idea, but it really only needs one set of parameters. If it pickles at one size, it will pickle for all sizes.

src/garage/torch/modules/gaussian_mlp_module.py
krzentner: I would just make it a private function of the module. Otherwise I'm fine with it.

It would definitely be a code smell in Java, but using small intermediate functions is relatively normal in Python.

yonghyuc: fixed

tests/garage/torch/policies/test_gaussian_mlp_policy.py
yonghyuc: fixed

src/garage/torch/policies/base.py
yonghyuc: Is it ok to do this job on new PR?
I am worried that this change could affect other codes

krzentner: Sure, implementing this change in a later PR sounds fine.

tests/garage/torch/modules/test_gaussian_mlp_module.py
krzentner: I generally don't care if there are "extra" tests, as long as they don't seem excessive.
Having said that, this test *should* be redundant.

Basically, I don't care, but please don't spend time writing these tests, since they're not needed.

yonghyuc: you mean there are too many test data in this test cases?
or this is not needed because PyTorch provides pickling?

krzentner: This one test just isn't needed because PyTorch provides the picking support.

yonghyuc: Oh.. ok then can I delete other pickling testcases?
I think I have one more pickling testcase in gaussian mlp policy

krzentner: Sure.

yonghyuc: delete every pickable tests

tests/garage/torch/policies/test_gaussian_mlp_policy.py
yonghyuc: delete this code because the pickable testcase is removed

tests/garage/torch/policies/test_gaussian_mlp_policy.py
yonghyuc: delete this code because the pickable testcase is removed

src/garage/torch/modules/multi_headed_mlp_module.py
```diff
@@ -0,0 +1,125 @@
+import torch.nn as nn
+
+
+class MultiHeadedMLPModule(nn.Module):
+    ERROR_MESSAGE = '{} should be either an integer or a collection of length n_heads ({}), but {} provided.'
```
ryanjulian: this is not necessary. please just put it right in the code.

src/garage/torch/modules/multi_headed_mlp_module.py
```diff
@@ -0,0 +1,125 @@
...
+
+    Return:
+        The list of output torch. Tensor of the MLP
+    """
+    def __init__(self,
+                 n_heads,
+                 input_dim,
+                 output_dims,
+                 hidden_sizes,
+                 hidden_nonlinearity=nn.ReLU,
```
ryanjulian: uhhhh usually RL uses tanh

src/garage/torch/modules/multi_headed_mlp_module.py
```diff
@@ -0,0 +1,125 @@
...
+        self._layers = nn.ModuleList()
+
+        if n_heads < 1:
+            raise ValueError("n_head should be greater than or equal to 1 but {} provided.".format(n_heads))
+
+        if isinstance(output_dims, (list, tuple)):
+            if len(output_dims) != 1 and len(output_dims) != n_heads:
+                raise ValueError(self.ERROR_MESSAGE.format('output_dim', n_heads, output_dims))
+        else:
+            output_dims = [output_dims]
```
ryanjulian: shouldn't this be `output_dims = [output_dims] * n_heads`?

src/garage/torch/modules/multi_headed_mlp_module.py
```diff
@@ -0,0 +1,125 @@
...
+            self._layers.append(layer)
+            prev_size = size
+
+        self._output_layers = nn.ModuleList()
+        for i in range(n_heads):
+            layer = nn.Sequential()
+
+            output_dim = output_dims[0 if len(output_dims) == 1 else i]
+            linear_layer = nn.Linear(prev_size, output_dim)
+            output_w_inits[0 if len(output_w_inits) == 1 else i](linear_layer.weight)
```
ryanjulian: this is incredibly difficult to read. please see my suggestion above (just initialize it to be length n_heads)

src/garage/torch/modules/multi_headed_mlp_module.py
```diff
@@ -0,0 +1,125 @@
...
+                 output_b_inits=nn.init.zeros_,
+                 layer_normalization=False):
+        super().__init__()
+
+        self._layers = nn.ModuleList()
+
+        if n_heads < 1:
+            raise ValueError("n_head should be greater than or equal to 1 but {} provided.".format(n_heads))
+
+        if isinstance(output_dims, (list, tuple)):
```
ryanjulian: can you put this repeated incantation into a helper function? it would help you share the error message, too

src/garage/torch/modules/multi_headed_mlp_module.py
ryanjulian: see my comment below. multiplying it here makes the code a lot simpler.

src/garage/torch/modules/multi_headed_mlp_module.py
yonghyuc: I tried to make the default value similar with [tf.models.MLPModel](https://github.com/rlworkgroup/garage/blob/47d388949b322e29c17958c0f75e79cd46741666/src/garage/tf/models/mlp_model.py#L48) and[ torch.model.MLPModule](https://github.com/rlworkgroup/garage/blob/47d388949b322e29c17958c0f75e79cd46741666/src/garage/torch/modules/mlp_module.py#L50)
If you think this need to change, I would fix the code

src/garage/torch/modules/multi_headed_mlp_module.py
yonghyuc: if there is only one output_dim, then the module use only first value.
so I think this would be problem, but if you think this is needed to change, I would fix the code

src/garage/torch/modules/multi_headed_mlp_module.py
yonghyuc: fixed

src/garage/torch/modules/multi_headed_mlp_module.py
yonghyuc: fixed

src/garage/torch/modules/multi_headed_mlp_module.py
yonghyuc: fixed

src/garage/torch/modules/multi_headed_mlp_module.py
ahtsan: I think MLPModel uses relu while *MLPModel uses tanh as default values for tf modules.

PR #714, opened 2019-06-03
Fix plot bug 

### LINKED ISSUE #697: Cannot plot during training
```
from garage.experiment import LocalRunner, run_experiment

def run_task(*_):
    with LocalRunner() as runner:
        env = TfEnv(env_name='CartPole-v1')

        policy = CategoricalMLPPolicy(
            name='policy', env_spec=env.spec, hidden_sizes=(32, 32))

        algo = TRPO(
            env_spec=env.spec,
            policy=policy,
            baseline=baseline,
            max_path_length=100,
            discount=0.99,
            max_kl_step=0.01)

        runner.setup(algo, env)
        runner.train(n_epochs=100, batch_size=4000,plot=True)


run_experiment(
    run_task,
    snapshot_mode='last',
    seed=4,
    n_parallel=4,
    plot=True,
    use_tf=False,
    use_gpu=False
)
```
##########################################################
3) Why we removed viskit? I cannot find it in garage. 

Thanks！ I really like rllab and garage.

zequnyu:
This will fix #697.

### DIFF
--- a/src/garage/experiment/local_tf_runner.py
+++ b/src/garage/experiment/local_tf_runner.py
@@ -388,6 +388,7 @@ class LocalRunner:
             pause_for_plot=pause_for_plot,
             start_epoch=start_epoch)
 
+        self.plot = plot
         self.start_worker()
 
         self.start_time = time.time()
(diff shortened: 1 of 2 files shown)


ryanjulian:
Can you add a test which would have failed when this bug wasn't fixed?

PR #715, opened 2019-06-03
Benchmark use all cpus

cheng-kevin:


ryanjulian:
Looks like you might need to rebase?

ahtsan:
I think DQN will run forever on cpu.

tests/benchmarks/test_benchmark_ddpg.py
```diff
@@ -152,8 +153,8 @@ def run_garage(env, seed, log_dir):
     :return:
     '''
     deterministic.set_seed(seed)
-
-    with LocalRunner() as runner:
+    ncpu = max(multiprocessing.cpu_count() // 2, 1)
```
ryanjulian: hmm this is pretty conservative. does this pin your CPUs to 100% load average?

cheng-kevin: Do you suggest setting it to the max with increased nice-ness?

ryanjulian: i suggest you try it on your local machine with default priority and see if you can max out the CPUs there. hyperthreading makes it difficult to predict exactly how many workers will actually max out the CPU.

ryanjulian: this heuristic is basically saying "set ncpu to the number of physical cores and ignore hyperthreading," which is pretty conservative. Also, you might need different numbers of workers for TF vs the sampler, since they have different usage profiles.

cheng-kevin: I tested TF vs Sampler, and this heuristic maxes out the cpu of my local machine in both cases with load averages of about 1.5-2.0

ryanjulian: if you machine has 12 logical CPUs (e.g. 12 are shown in `htop`), then your target load average should be 12. is that what you're seeing?

PR #716, opened 2019-06-03
Auto generate docs from docstrings

### LINKED ISSUE #647: Fix documentation generation
See https://github.com/rlworkgroup/akro for an example.

A lot of the existing documentation template is kind of non-applicable, so it would be fine to delete the existing templates (not the documentation itself) and start over with the template from akro.

gitanshu:
closes #647


ryanjulian:
hmm this doesn't seem to be totally working. for instance, `cnn_model` doesn't actually have any docs: https://garage.readthedocs.io/en/fix_api_doc/_apidoc/garage.tf.models.cnn_model.html

gitanshu:
Yeah something's off. A lot of them are turning up empty. Will look into it

ryanjulian:
Yeah, it looks like everything passes CI, but the API docs are still missing for many modules.

ryanjulian:
Unfortunately the automated API docs are still incomplete. Also, see #755 which is new.

I propose we close both issues using this PR.

gitanshu:
This is being done on fix_doc_gen branch now, opening a PR for that

Issue #704, opened 2019-05-29
Refactor Runner to allow Algorithm to own training loop


krzentner:
Currently, the [`LocalTfRunner._train`](https://github.com/rlworkgroup/garage/blob/8b1bac869ac2bb167aa9982c80ee968e662b9f5a/src/garage/experiment/local_tf_runner.py#L352) method owns the outer training loops, because it uses them to provide services (such as snapshotting), to the Algorithm. However, all of those services could be provided by a "magic iterator" over the epoch.

Concretely, this would mean that the `LocalTfRunner._train` method would look like this:
```python
last_return = None
itr = 0
for epoch in self.step_epochs():
    for cycle in range(n_epoch_cycles):
        paths = self.obtain_samples(itr, batch_size)
        last_return = self.algo.train_once(itr, paths)
        itr += 1
return last_return
```

Once `LocalTfRunner._train` looks like this (since it will then use public APIs), we can move it to `BaseAlgorithm` (and perhaps rename it). We should probably also consider changing the sampler API, so that the Algorithm has direct access to the sampler, and can pass it a policy to obtain samples with.

naeioi:
I don't quite get the point of this proposal. Can you elaborate a bit? In particular, if we were to place the training loop in algorithm, and give an algorithm access to sampler, isn't it essentially reverting to the garage before runner came to exist? 

To me, the idea of runner owning training loop is to decouple algorithm from experiment orchestration. This includes sampling, logging, snapshotting, and the data piping between components in garage. So that algorithm cares only about updating policy given a batch of samples.

krzentner:
For most deep RL algorithms, this change won't make a big difference, since they are easy to write using the loop described above by defining `train_once`. However, for algorithms that operate in other ways (such as a collection of policies), this is really awkward. For example, evolutionary strategies algorithms need to switch out the policy each `train_once` step, and match up the `paths` it receives with the previous `policy`. We would like it to be easy to express algorithms like that. However, we still want the runner to own all of other "services" like sampling, logging, and snapshotting.

I think that the refactoring we're doing here will let us do that, by exposing a simple interface for all of those other services. What we're giving up in the process is that the runner no longer controls data pipelining between the sampler and the algorithm, and instead the algorithm interacts with the sampler through the sampler interface. This seems sufficient to me, since I don't see what else we're likely to do, and the dataflow here is just two lines of code, so it's fine if the algorithm owns it.

naeioi:
[...]

PR #719, opened 2019-06-05
Refactor LocalRunner training loop

zequnyu:
#704.

### DIFF
--- a/src/garage/experiment/local_tf_runner.py
+++ b/src/garage/experiment/local_tf_runner.py
@@ -187,7 +187,7 @@ class LocalRunner:
 
         Args:
             itr(int): Index of iteration (epoch).
-            paths(dict): Batch of samples after preprocessed.
+            paths(None or dict): Batch of samples after preprocessed.
 
         """
         assert self.has_setup
@@ -310,7 +310,11 @@ class LocalRunner:
             The average return in last epoch cycle.
 
         """
-        return self._train(
+        assert self.has_setup, ('Use Runner.setup() to setup runner before '
+                                'training.')
+
+        # Save arguments for restore
+        self.train_args = SimpleNamespace(
             n_epochs=n_epochs,
             n_epoch_cycles=n_epoch_cycles,
             batch_size=batch_size,
@@ -318,6 +322,60 @@ class LocalRunner:
             store_paths=store_paths,
             pause_for_plot=pause_for_plot,
             start_epoch=0)
+
+        self.plot = plot
+
+        return self._train()
+
+    def step_epochs(self):
+        """Generator for training.
+
+        This function serves as a generator. It is used to separate
+        services such as snapshotting, sampler control from the actual
+        training loop.
+
+        There are two fields both used by `step_epochs()` and `_train()`:
+        `self.step_itr` and `self.step_path`. This function initializes
+        the two. However, it has to be updated in each epoch in `_train()`
+        manually, as the example shows below.
+
+        Yields:
+            int: The next training epoch.
+
+        Examples:
+            for epoch in self.step_epochs():
+                for cycle in range(n_epoch_cycles):
+                    self.step_path = self.obtain_samples(...)
+                    self.algo.train_once(...)
+                self.step_itr += 1
+
+        """
+        try:
+            self.start_worker()
+
+            self.start_time = time.time()
+
+            self.step_itr = (
+                self.train_args.start_epoch * self.train_args.n_epoch_cycles)
+            self.step_path = None
+
+            for epoch in range(self.train_args.start_epoch,
+                               self.train_args.n_epochs):
+                self.itr_start_time = time.time()
+
+                with logger.prefix('epoch #%d | ' % epoch):
+                    yield epoch
+
+                    save_path = (self.step_path
+                                 if self.train_args.store_paths else None)
+                    self.save(epoch, save_path)
+
+                    self.log_diagnostics(self.train_args.pause_for_plot)
+                    logger.dump_all(self.step_itr)
+                    tabular.clear()
+
+        finally:
+            self.shutdown_worker()
 
     def resume(self,
                n_epochs=None,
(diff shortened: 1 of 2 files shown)


src/garage/experiment/local_tf_runner.py
```diff
@@ -319,6 +323,39 @@ def train(self,
...
+                yield
+
+                self.save(
+                    epoch, self.step_epochs_args.path
+                    if self.train_args.store_paths else None)
+                self.log_diagnostics(self.train_args.pause_for_plot)
+                logger.dump_all(self.step_epochs_args.itr)
+                tabular.clear()
+
+        self.shutdown_worker()
```
krzentner: It would be nice for `shutdown_worker()` to be called in a finally clause, so that the workers will get shutdown even if the loop using this generator throws an exception. This can be in a later PR though.

Please either make this change in this PR, or create a github issue.

src/garage/experiment/local_tf_runner.py
```diff
@@ -340,73 +377,36 @@ def resume(self,
...
-    def _train(self,
-               n_epochs,
-               n_epoch_cycles,
-               batch_size,
-               plot,
-               store_paths,
-               pause_for_plot,
-               start_epoch=0):
+        if n_epochs:
+            self.train_args.n_epochs = n_epochs
```
krzentner: The decrease in code coverage is entirely due to these lines never being tested.
They're probably correct, but it's still a good practice to test them (in my opinion).
Out of principle, please add a test to `tests/garage/experiment/test_local_tf_runner.py` which sets all of these parameters.

src/garage/experiment/local_tf_runner.py
```diff
@@ -340,73 +377,36 @@ def resume(self,
...
-                    itr += 1
-                self.save(epoch, paths if store_paths else None)
-                self.log_diagnostics(pause_for_plot)
-                logger.dump_all(itr)
-                tabular.clear()
-
-        self.shutdown_worker()
+        for _ in self._step_epochs():
+            for cycle in range(self.train_args.n_epoch_cycles):
+                self.step_epochs_args.path = self.obtain_samples(
```
krzentner: It's true that `path` (and `itr`) are used by `step_epochs()`, but using a namespace is fairly overkill, and feels really strange.

Please just make `path` and `itr` fields of the runner instead.

@ryanjulian Do you have other opinions on how to fix this API? Maybe these should be fields on the Algorithm?

src/garage/experiment/local_tf_runner.py
```diff
@@ -319,6 +323,39 @@ def train(self,
...
+        self.step_epochs_args = SimpleNamespace(
+            itr=self.train_args.start_epoch * self.train_args.n_epoch_cycles,
+            path=None)
+
+        for epoch in range(self.train_args.start_epoch,
+                           self.train_args.n_epochs):
+            self.itr_start_time = time.time()
+
+            with logger.prefix('epoch #%d | ' % epoch):
+                yield
```
krzentner: Please yield `epoch` here, since it's useful to the user (and is intuitive from the method name).

src/garage/experiment/local_tf_runner.py
```diff
@@ -319,6 +323,39 @@ def train(self,
             pause_for_plot=pause_for_plot,
             start_epoch=0)
 
+        return self._train()
+
+    def _step_epochs(self):
```
krzentner: This should be a public API, so call the method `step_epochs()` (without the underscore).
The most important thing to describe in this docstring is the surprising behaviour. In particular, the fact that this method uses the `step_epoch_args` (or after renaming, the `itr` and `path` fields), at the end of each epoch.

The docstring should also describe that it yields each epoch.

src/garage/experiment/local_tf_runner.py
```diff
@@ -340,73 +377,36 @@ def resume(self,
...
-
-    def _train(self,
-               n_epochs,
-               n_epoch_cycles,
-               batch_size,
-               plot,
-               store_paths,
-               pause_for_plot,
-               start_epoch=0):
+        if n_epochs:
```
ryanjulian: you can use the python idiom
`self.train_args.n_epochs = n_epochs or self.train_args.n_epochs`

src/garage/experiment/local_tf_runner.py
```diff
@@ -319,6 +323,39 @@ def train(self,
             pause_for_plot=pause_for_plot,
             start_epoch=0)
 
+        return self._train()
+
+    def _step_epochs(self):
+        """Magic generator for training.
```
ryanjulian: we should probably not call this magical

src/garage/experiment/local_tf_runner.py
```diff
@@ -319,6 +323,39 @@ def train(self,
...
+
+        This function serves as a generator. It is used to separate
+        services such as snapshotting, sampler control from the actual
```
ryanjulian: I think an example would make this docstring 100% more clear

src/garage/experiment/local_tf_runner.py
ryanjulian: i was just going to comment on this. the `SimpleNamespace` was intended only for convenient storage (pickling) and passing back to _train(), not as objects fields. it should never be used as a container for object fields, so I think we should refactor this back to use only fields.

as for `path` and `itr`: why is `path` saved anyway? the return from `obtain_samples` could be Very Large, and we don't necessarily want to save it.

i'm not really offended by `itr` because the runner needs that information (e.g. for logging, saving snapshots). it should be a field of the runner and it will also be a field in the experiment pkl file.

src/garage/experiment/local_tf_runner.py
```diff
@@ -340,73 +377,36 @@ def resume(self,
...
-               start_epoch=0):
+        if n_epochs:
+            self.train_args.n_epochs = n_epochs
+        if batch_size:
+            self.train_args.batch_size = batch_size
+        if n_epoch_cycles:
+            self.train_args.n_epoch_cycles = n_epoch_cycles
+        if plot:
+            self.train_args.plot = plot
+        if store_paths:
```
krzentner: This is a boolean, so you need to explicitly check that `store_paths is not None`.

src/garage/experiment/local_tf_runner.py
krzentner: Sometimes it's useful to store `path`. Right now, `path` is only saved if `store_paths == True` (it defaults to `False`), which addresses the concern with it potentially being Very Large. It should probably be on the same object as `itr`? 

Alternatively, we could have some mechanism for explicitly passing paths to be saved, which would probably be cleaner. That's out of scope for this PR though.

zequnyu:
Close and reopen due to endless "Waiting for status to be reported".

zequnyu:
This is ready for another review. Sadly Codecov report is never back even if I reopen the PR.

zequnyu:
Continue at #728.

PR #722, opened 2019-06-06
ContinuousMLPPolicy with model

ahtsan:
This PR added `continuous_mlp_policy_with_model` (which will replace the existing `continuous_mlp_policy` and `deterministic_mlp_policy_with_model` as they all are the same).

Existing `continuous_mlp_policy` scales the action according to the environment action space bounds. Instead of scaling in the policy, this PR scales the action output from policy in the sampler, which encourages the policy to learn outputs within [-1, 1] (in theory the policy will learn better).

Attached the plot for `continuous_mlp_policy` and `continuous_mlp_policy_with_model` on the environment `InvertedPedulum`, which has action space bounds [-3, 3]. Three runs in total.

`continuous_mlp_policy`:
old_ddpg

`continuous_mlp_policy_with_model`:
new_ddpg


src/garage/tf/samplers/off_policy_vectorized_sampler.py
```diff
@@ -98,7 +99,13 @@ def obtain_samples(self, itr, batch_size):
                 actions, agent_infos = self.algo.policy.get_actions(
                     obs_normalized)
 
-            next_obses, rewards, dones, env_infos = self.vec_env.step(actions)
+            if (isinstance(self.env.action_space, Box)):
+                scaled_actions = np.multiply(actions,
```
zhanpenghe: Why are actions scaled by the upperbound of the action space?

src/garage/tf/samplers/off_policy_vectorized_sampler.py
```diff
@@ -98,7 +99,13 @@ def obtain_samples(self, itr, batch_size):
                 actions, agent_infos = self.algo.policy.get_actions(
                     obs_normalized)
 
-            next_obses, rewards, dones, env_infos = self.vec_env.step(actions)
+            if (isinstance(self.env.action_space, Box)):
```
zhanpenghe: Can we have an option not to scale them?

ahtsan: Yes we can. Are there some cases where it's better for the policy to learn the exact unnormalized output values?

zhanpenghe: It is very easy for people to ignore the scaling if we put this in a sampler. Once the policy is trained, if scaling is ignored, the visualization will never be correct and will be confusing. I would suggest either putting this as an option of the policy or just have an environment wrapper to do the job.

ahtsan: Ok. Made the scaling an option.

src/garage/tf/samplers/off_policy_vectorized_sampler.py
```diff
@@ -23,18 +24,27 @@ class OffPolicyVectorizedSampler(BatchSampler):
...
 
     Args:
-        algo(garage.np.RLAlgorithm): Algorithm.
-        env(garage.envs.GarageEnv): Environment.
-        n_envs(int): Number of parallel environments managed by sampler.
+        algo (garage.np.RLAlgorithm): Algorithm.
+        env (garage.envs.GarageEnv): Environment.
+        n_envs (int): Number of parallel environments managed by sampler.
+        no_reset (bool): Reset environment between samples or not.
+        scale_outputs (bool): Scale output by action space upper bound or
```
ryanjulian: i don't think this option should exist because, for instance, a gaussian policy cannot be meaningfully scaled. It is not the role of the sampler to modify the output of a policy.

ahtsan: We probably need an environment wrapper in this case?

ahtsan: How about make the algorithm responsible to scale?

zhanpenghe: I don't think the algorithm should scale the actions either. Either the policy or the environment should do it.

ahtsan: How about we do it in `normalized_env` as an option?

ryanjulian: I think an environment wrapper makes the most sense. What is the application for this?

CatherineSue: I thought `normalized_env` scaled actions. [see here](https://github.com/rlworkgroup/garage/blob/master/src/garage/envs/normalized_env.py#L95)

ahtsan: Yes I just found out `normalized_env` already did that. I guess then we don't have to do any scaling anywhere else but we have to document clearly `normalized_env` is suggested to be used but also not to use if not appropriate, e.g. Gaussian policy on a Box env.

I will go ahead and remove the scaling.

ahtsan:
CI should pass after akro bumped to >=0.0.3

DDPG result on `InvertedPendulum:
Screen Shot 2019-06-22 at 11 56 17 PM

src/garage/tf/envs/base.py
```diff
@@ -1,7 +1,7 @@
-from akro.tf import Box
-from akro.tf import Dict
-from akro.tf import Discrete
-from akro.tf import Tuple
+from akro import Box
```
ryanjulian: `import akro`

src/garage/tf/policies/categorical_conv_policy.py
```diff
@@ -1,4 +1,4 @@
-from akro.tf import Discrete
+from akro import Discrete
```
ryanjulian: import akro

src/garage/tf/policies/categorical_conv_policy_with_model.py
```diff
@@ -1,5 +1,5 @@
 """CategoricalConvPolicy with model."""
-from akro.tf import Discrete
+from akro import Discrete
```
ryanjulian: import akro

src/garage/tf/policies/categorical_gru_policy.py
```diff
@@ -1,4 +1,4 @@
-from akro.tf import Discrete
+from akro import Discrete
```
ryanjulian: import akro

src/garage/tf/policies/categorical_gru_policy_with_model.py
```diff
@@ -1,5 +1,5 @@
 """CategoricalGRUPolicy with model."""
-from akro.tf import Discrete
+from akro import Discrete
```
ryanjulian: import akro

src/garage/tf/policies/categorical_lstm_policy.py
```diff
@@ -1,4 +1,4 @@
-from akro.tf import Discrete
+from akro import Discrete
```
ryanjulian: import akro

src/garage/tf/policies/categorical_lstm_policy_with_model.py
```diff
@@ -1,5 +1,5 @@
 """CategoricalLSTMPolicy with model."""
-from akro.tf import Discrete
+from akro import Discrete
```
ryanjulian: import akro

src/garage/tf/policies/categorical_mlp_policy.py
```diff
@@ -1,4 +1,4 @@
-from akro.tf import Discrete
+from akro import Discrete
```
ryanjulian: import akro

src/garage/tf/policies/categorical_mlp_policy_with_model.py
```diff
@@ -1,5 +1,5 @@
 """CategoricalMLPPolicy with model."""
-from akro.tf import Discrete
+from akro import Discrete
```
ryanjulian: import akro

src/garage/tf/policies/continuous_mlp_policy.py
```diff
@@ -5,7 +5,7 @@
 algorithms. It accepts an observation of the environment and predicts an
 action.
 """
-from akro.tf import Box
+from akro import Box
```
ryanjulian: import akro

src/garage/tf/policies/gaussian_lstm_policy.py
```diff
@@ -1,4 +1,4 @@
-from akro.tf import Box
+from akro import Box
```
ryanjulian: import akro

src/garage/tf/policies/gaussian_gru_policy.py
```diff
@@ -1,4 +1,4 @@
-from akro.tf import Box
+from akro import Box
```
ryanjulian: import akro

src/garage/tf/policies/discrete_qf_derived_policy.py
```diff
@@ -3,7 +3,7 @@
 
 This policy chooses the action that yields to the largest q-value.
 """
-from akro.tf import Discrete
+from akro import Discrete
```
ryanjulian: import akro

src/garage/tf/policies/deterministic_mlp_policy.py
```diff
@@ -1,4 +1,4 @@
-from akro.tf import Box
+from akro import Box
```
ryanjulian: import akro

src/garage/tf/policies/gaussian_mlp_policy.py
```diff
@@ -1,4 +1,4 @@
-from akro.tf import Box
+from akro import Box
```
ryanjulian: import akro

src/garage/tf/policies/gaussian_mlp_policy_with_model.py
```diff
@@ -1,5 +1,5 @@
 """GaussianMLPPolicy with GaussianMLPModel."""
-from akro.tf import Box
+from akro import Box
```
ryanjulian: import akro

src/garage/tf/envs/base.py
```diff
@@ -1,7 +1,4 @@
-from akro.tf import Box
-from akro.tf import Dict
-from akro.tf import Discrete
-from akro.tf import Tuple
+import akro
 from cached_property import cached_property
 from gym.spaces import Box as GymBox
```
ryanjulian: `import gym.spaces` here -- sorry for missing it before

ahtsan:
Benchmarks

HalfCheetah-v2_benchmark
Hopper-v2_benchmark
InvertedPendulum-v2_benchmark
Reacher-v2_benchmark
Swimmer-v2_benchmark
Walker2d-v2_benchmark

CatherineSue:
The benchmark looks good. Thanks for the great work. 
Could you send me the source code you used to run the benchmark? It looks like baselines is not learning. But garage has different sampling behaviors than baselines. Is baselines learning but not shown in the plot?
Since we are testing primitives, in my opinion, it would be more direct if you compare the performance of old primitive vs new primitives instead of new primitives vs. baselines.

ahtsan:
It's on the branch `continuous_mlp_policy_with_model`.

The benchmarks should be efficient, per our discussion. This PR will pass the CI test and able to submit once akro version is bumped.

ryanjulian:
@ahtsan if akro is the only thing blocking this PR, can you get #740 ready for submit and then submit this one?

PR #725, opened 2019-06-11
Add pytorch

lywong92:


setup.py
```diff
@@ -24,6 +24,7 @@
     'psutil',
     'pyprind',
     'python-dateutil',
+    'pytorch-cpu torchvision-cpu -c pytorch',
```
ryanjulian: is there a reason these each can't be different entries?

tests/garage/torch/modules/test_mlp_module.py
```diff
@@ -0,0 +1,48 @@
...
+            hidden_w_init=nn.init.ones_,
+            output_w_init=nn.init.ones_)
+        output1 = module.forward(input_val)
+
+        h = pickle.dumps(module)
+        model_pickled = pickle.loads(h)
+        output2 = model_pickled.forward(input_val)
+
+        assert np.array_equal(torch.all(torch.eq(output1, output2)), True)
+
```
ryanjulian: note the missing newline. the CI will ding you on this.

you can configure your editor to add these (also to strip trailing whitespace, use spaces instead of tabs, etc.). see CONTRIBUTING.md for details. there is a .editorconfig if you'd like to use that plug-in.

lywong92: Oh right I totally forgot about that. Just installed the hooks

setup.py
lywong92: No I will separate them

tests/garage/torch/modules/test_mlp_module.py
```diff
@@ -0,0 +1,47 @@
...
+            (5, 3, (2, 2)))
+    def test_output_values(self, input_dim, output_dim, hidden_sizes):
+        input_val = torch.ones([1, 5], dtype=torch.float32)
+        module = MLPModule(
+            input_dim=input_dim,
+            output_dim=output_dim,
+            hidden_sizes=hidden_sizes,
+            hidden_w_init=nn.init.ones_,
+            output_w_init=nn.init.ones_)
+        output = module.forward(input_val)
```
zhanpenghe: We don't need to directly call forward here. I think you just need to do `output = module(input_val)`

tests/garage/torch/modules/test_mlp_module.py
```diff
@@ -0,0 +1,47 @@
...
+            (5, 3, (2, 2)))
+    def test_is_pickleable(self, input_dim, output_dim, hidden_sizes):
+        input_val = torch.ones([1, 5], dtype=torch.float32)
+        module = MLPModule(
+            input_dim=input_dim,
+            output_dim=output_dim,
+            hidden_sizes=hidden_sizes,
+            hidden_w_init=nn.init.ones_,
+            output_w_init=nn.init.ones_)
+        output1 = module.forward(input_val)
```
zhanpenghe: Same as above.

tests/garage/torch/modules/test_mlp_module.py
```diff
@@ -0,0 +1,47 @@
...
+
+        h = pickle.dumps(module)
+        model_pickled = pickle.loads(h)
+        output2 = model_pickled.forward(input_val)
```
zhanpenghe: Same as above.

src/garage/torch/modules/mlp_module.py
```diff
@@ -0,0 +1,92 @@
...
+                 hidden_nonlinearity=func.relu,
+                 hidden_w_init=nn.init.xavier_normal_,
+                 hidden_b_init=nn.init.zeros_,
+                 output_nonlinearity=None,
+                 output_w_init=nn.init.xavier_normal_,
+                 output_b_init=nn.init.zeros_,
+                 layer_normalization=False):
+        super(MLPModule, self).__init__()
+
+        self.input_dim = input_dim
```
zhanpenghe: Maybe make these variables private? e.g. `self._input_dim = input_dim`.

zhanpenghe:
In order to pass the CI, the first letter of your commit message needs to be capitalized. You can squash them and force-push this branch to resolve it.

setup.py
```diff
@@ -24,10 +24,12 @@
     'psutil',
     'pyprind',
     'python-dateutil',
+    'torch',
```
zhanpenghe: Can we pin the version of pytorch? This is a very important dependency.

src/garage/torch/modules/mlp_module.py
```diff
@@ -0,0 +1,92 @@
...
+    def forward(self, input_val):
+        """Forward method."""
+        x = input_val
+        for layer in self._layers[:-1]:
+            x = self._hidden_nonlinearity(layer(x))
+            if self._layer_normalization:
+                x = nn.LayerNorm(x.shape[1])(x)
+
+        x = self._layers[-1](x)
+        if self._output_nonlinearity:
```
zhanpenghe: `if self._output_nonlinearity is not None:`

src/garage/torch/modules/mlp_module.py
```diff
@@ -0,0 +1,92 @@
...
+                 output_dim,
+                 hidden_sizes,
+                 hidden_nonlinearity=func.relu,
+                 hidden_w_init=nn.init.xavier_normal_,
+                 hidden_b_init=nn.init.zeros_,
+                 output_nonlinearity=None,
+                 output_w_init=nn.init.xavier_normal_,
+                 output_b_init=nn.init.zeros_,
+                 layer_normalization=False):
+        super(MLPModule, self).__init__()
```
zhanpenghe: I think you only need `super().__init__()`.

src/garage/torch/modules/mlp_module.py
```diff
@@ -0,0 +1,92 @@
+"""
+MLP Module.
+
+A Pytorch module composed only of a multi-layer perceptron (MLP), which maps
+real-valued inputs to real-valued outputs.
+"""
+
+from torch import nn as nn
+from torch.nn import functional as func
```
zhanpenghe: Let's follow the pytorch documentations and do `from torch.nn import functional as F`

setup.py
lywong92: Will do!

Issue #713, opened 2019-06-03
test_batch_sampler() stuck silently


zequnyu:
The `test_batch_sampler()` got stuck silently even tested independently. However, while tracking the commit history, I found that when adding 
```
def setUp(self):
    logger.add_output(NullOutput())
def tearDown(self):
    logger.remove_all()
```
to the test class, it could pass.

naeioi:
I suspect logger does some tensorflow initialization and that hangs batch_sample. could you try running examples/tf/trpo_cartpole_batch_sampler.py and see if it stucks?

zequnyu:
The example works.

zequnyu:
I also have a [test](https://github.com/rlworkgroup/garage/pull/717/files#diff-87b617e336912cbf3885a20466709120) got stuck silently. This time it could pass local unittest and nose2, but not CI. I feel like both due to the multi processing part of `singleton_pool`, which is pretty tricky to debug.

krzentner:
The fact that the change at you describe fixes the problem is a pretty big hint.

@zequnyu If you add a `dowel.StdOutput` instead of the `NullOutput`, is there any particularly interesting log information?

naeioi:
You are not likely to fix this bug. This is yet another bug rooted in tensorflow not being fork-safe. Your workaround posted at the very beginning works because it overrides `setUp()` of `TfGraphTestCase`, where a tensorflow session is created, and which will lead to a freeze in a multiprocess context. 

My suggest is either we remove this test, or place it under an isolated test class, which does not inherit `TfGraphTestCase`.

zhanpenghe:
I have not taken a look at the test yet but I think it is not related to tensorflow. Currently dowel is hanging if  logger.remove _all() is not called. I encountered this too in my research project.

zequnyu:
Closed by #727.

PR #727, opened 2019-06-12
Update test_batch_sampler

zequnyu:
- renamed and relocated the test
- updated it to avoid `singleton_pool` hanging problem

This could allow us to test it locally without hanging due to tensorflow not being fork-safe. 
#713


zequnyu:
It hangs due to tf sessions. But other tests in test_local_tf_runner need that. I actually followed your [suggestion](https://github.com/rlworkgroup/garage/issues/713#issuecomment-499363316). Back to the time I didn't realise setting `setUp()` and `tearDown()` will override the setting of tf sessions in tf base test class, so I incorrectly thought it's a logger problem.

PR #729, opened 2019-06-17
Add pytest deps

gitanshu:


ryanjulian:
Hmm so it looks like your first job failed because it had to rebuild the docker container. I think you can fix that by submitting the dependency update in a separate PR first 

1. PR which adds pytest and friends
1. PR which adds pytest-backed tests
1. PR which removes nose2 and friends

Are you sure filtering tests by attribute (or pytest's version of attributes/filters) is working for all of the jobs?

It's also interesting that execution of the large job takes 8min longer. Is this reproducible on your local machine?

gitanshu:
> Hmm so it looks like your first job failed because it had to rebuild the docker container. I think you can fix that by submitting the dependency update in a separate PR first
> 
> 1. PR which adds pytest and friends
> 2. PR which adds pytest-backed tests
> 3. PR which removes nose2 and friends

On it.

> Are you sure filtering tests by attribute (or pytest's version of attributes/filters) is working for all of the jobs?

Yes, the filtering is working fine. Pytest uses annotations (@pytest.mark.large etc) for this. 

> It's also interesting that execution of the large job takes 8min longer. Is this reproducible on your local machine?

No, pytest and nose2 take almost the same time on my local machine (15-20 sec diff)

PR #733, opened 2019-06-19
Pinned akro and fixed torch mlp module

lywong92:


krzentner:
The coverage failure is just noise, and the dockercloud results shouldn't be affected, so I'm going to merge this in.

Issue #732, opened 2019-06-19
Bump akro to 0.0.4


zhanpenghe:

ryanjulian:
@krzentner this is finished right?

ryanjulian:
fixed by #740

PR #735, opened 2019-06-20
Remove Parameterized from GarageEnv

zhanpenghe:
It also remove redundant codes in `TfEnv`.

### DIFF
--- a/src/garage/tf/envs/base.py
+++ b/src/garage/tf/envs/base.py
@@ -20,12 +20,6 @@ class TfEnv(GarageEnv):
         env (gym.Env): the env that will be wrapped
     """
 
-    def __init__(self, env=None, env_name=''):
-        super().__init__(env, env_name)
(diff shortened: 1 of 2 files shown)


src/garage/tf/envs/base.py
```diff
@@ -20,12 +20,6 @@ class TfEnv(GarageEnv):
         env (gym.Env): the env that will be wrapped
     """
 
-    def __init__(self, env=None, env_name=''):
```
ryanjulian: hmm if you are removing this, why not remove GarageEnv totally? It is a TODO item anyway.

zhanpenghe:
@ahtsan yea, I opened an issue for this already. https://github.com/rlworkgroup/garage/issues/732

krzentner:
> akro.tf no longer exists so I guess TfEnv._to_akro_space() can also be removed. But that should probably be done in the "refactor-for-akro-change" PR.

Yeah, I'm working on that now.

PR #745, opened 2019-06-23
Make FiniteDifferenceHvp pickleable

ahtsan:
Currently all policies using `FiniteDifferenceHvp` in `examples/tf/` don't work because `FiniteDifferenceHvp` is not pickleable.

This PR fixes the issue, and also update outdated arguments in some TRPO usage.


ryanjulian:
@ahtsan can you add a test which would have caught this bug? i think you can just run one step of TRPO with the LSTM policy.

src/garage/tf/optimizers/conjugate_gradient_optimizer.py
```diff
@@ -133,6 +133,16 @@ def eval(x):
 
         return eval
 
+    def __getstate__(self):
+        """Object.__getstate__."""
+        new_dict = self.__dict__.copy()
+        del new_dict['opt_fun']
+        return new_dict
+
+    def __setstate__(self, state):
```
krzentner: Do you need to re-create `opt_fun` here?
If you *don't*, then you should be able to delete this `__setstate__` implementation, since this is the "default."
Note that you'll still have a `__getstate__` implementation then, despite not having a `__setstate__` implementation.

ahtsan: Good catch, will remove `__getstate__` since `opt_fun` is recreated somewhere else.

PR #763, opened 2019-07-02
[WIP] Test ubuntu 18.04 on CI

gitanshu:


docker/Dockerfile.base
```diff
@@ -2,7 +2,7 @@
 # and scripts/setup_macos.sh, if there's any changes applied to this file,
 # specially regarding the installation of dependencies, apply those same
 # changes to the mentioned files.
-ARG PARENT_IMAGE=ubuntu:16.04
+ARG PARENT_IMAGE=ubuntu:18.04
```
ryanjulian: can you branch these into Dockerfile.base.18.04, Dockerfile.base.16.04, etc? we will want to keep both configs around for a while. we can have the Makefile and CI use 18.04 by default.

gitanshu: Yes, that was the plan. This was just a quick push to check if things were working while i fix my local environment.

ryanjulian: there will be several changes between the 18.04 and 16.04 Dockerfiles -- the 18.04 file will likely be shorter (e.g. no need to recompile GLFW from source). Ask Aravind for what he does for 18.04. The xvfb config may or may not change. 

Try to remove as much as you can, since everything makes the Dockerfile bigger and we have probably accumulated things we don't need.

See: https://github.com/openai/gym/blob/0.11.0/test.dockerfile.18.04

gitanshu: About GLFW, we still need to recompile it from source because ubuntu 18.04 ships with glfw 3.2.1 which still has the same issue (https://github.com/glfw/glfw/issues/1004) and glfw 3.3 is not available in ubuntu software repos yet

docker/Dockerfile.base.18.04
```diff
@@ -0,0 +1,102 @@
...
+    # libglfw3 \
+    libglew-dev \
+    libosmesa6-dev \
+    patchelf \
+    # OpenAI baselines
+    libopenmpi-dev && \
+  apt-get clean && \
+  rm -rf /var/lib/apt/lists/*
+
+# Build GLFW because the Ubuntu 16.04 version is too old
```
ryanjulian: 18.04?

gitanshu:
@ryanjulian we'd have to add/modify the dockerhub repos for this. I don't have enough access to do that.

PR #764, opened 2019-07-02
Added DeterministicPolicy and ContinuousNNQFunction in torch

lywong92:


src/garage/torch/policies/continuous_mlp_policy.py
```diff
@@ -0,0 +1,70 @@
...
+    def __init__(self,
+                 env_spec,
+                 nn_module,
+                 name='ContinuousMLPPolicy',
+                 input_include_goal=False):
+        """
+        Initialize class with multiple attributes.
+
+        Args:
+            env_spec():
```
krzentner: Please delete lines of documentations which don't add anything (in this case, the `env_spec():` and `nn_module():` lines.

src/garage/torch/policies/continuous_mlp_policy.py
```diff
@@ -0,0 +1,70 @@
+"""
```
krzentner: So this class as written isn't actually specific to MLPs. Please rename the file and class to `ContinuousNNPolicy` (or some other name that you feel is appropriate).

src/garage/torch/policies/continuous_mlp_policy.py
```diff
@@ -0,0 +1,70 @@
...
+    This class implements a policy network.
+
+    The policy network selects action based on the state of the environment.
+    It uses neural nets to fit the function of pi(s).
+    """
+
+    def __init__(self,
+                 env_spec,
+                 nn_module,
+                 name='ContinuousMLPPolicy',
```
krzentner: I'm not convinced that we actually need a name parameter. In the tensorflow branch, it's needed to interact with the graph, but pytorch doesn't need it. Let's remove it for now, and we'll add it back if we need it in the future.

src/garage/torch/policies/continuous_mlp_policy.py
```diff
@@ -0,0 +1,70 @@
...
+        """Forward method."""
+        x = torch.FloatTensor(self._nn_module(input_val))
+        return torch.distributions.Uniform(x, x)
+
+    @overrides
+    def get_action(self, observation):
+        """Return a single action."""
+        with torch.no_grad():
+            d = self.forward(observation.unsqueeze(0))
+            return d.sample().detach().numpy()
```
krzentner: There should be a `[0]` somewhere here, since we want to return a single action (not a batch).
Alternatively you could use `.squeeze(0)` for symmetry's sake.

src/garage/torch/q_functions/continuous_mlp_q_function.py
```diff
@@ -0,0 +1,40 @@
+"""This modules creates a continuous MLP Q fcuntion network."""
+
+
+class ContinuousMLPQFunction:
```
krzentner: Like above, please rename MLP to NN.

src/garage/torch/policies/continuous_mlp_policy.py
```diff
@@ -0,0 +1,70 @@
...
+import akro
+import torch
+
+from garage.misc.overrides import overrides
+from garage.tf.policies.base import Policy
+
+
+class ContinuousMLPPolicy(Policy):
+    """
+    This class implements a policy network.
```
krzentner: It would be good to mention here that this policy is deterministic, and intended for use in algorithms which can train deterministic policies.

src/garage/torch/q_functions/continuous_mlp_q_function.py
```diff
@@ -0,0 +1,40 @@
...
+    This class implements a q value network.
+
+    It predicts the q value for all actions based on the input state. It uses
+    an MLP to fit the function of Q(s, a).
+    """
+
+    def __init__(self,
+                 env_spec,
+                 nn_module,
+                 name='ContinuousMLPQFunction',
```
krzentner: Again, I don't *think* we need the name parameter, so let's delete it for now.

tests/garage/torch/policies/test_continuous_mlp_policy.py
```diff
@@ -0,0 +1,65 @@
...
+
+from garage.torch.modules import MLPModule
+from garage.torch.policies import ContinuousMLPPolicy
+from tests.fixtures.envs.dummy import DummyBoxEnv
+
+
+class TestContinuousPolicies(unittest.TestCase):
+    def test_output_1d(self):
+        env_spec = MagicMock(return_value=DummyBoxEnv())
+        env_spec.action_space = MagicMock(
```
krzentner: Why do you need `MagicMock` here? Is the existing test fixture in `DummyBoxEnv` missing some functionality?

.gitignore
```diff
@@ -19,6 +19,7 @@ lib64/
 parts/
 sdist/
 var/
+.vscode/
```
ryanjulian: this is fine, but to make it easier to keep track of what we added (above the GitHub default), can you add this to one of the sections below https://github.com/rlworkgroup/garage/pull/764/files#diff-a084b794bc0759e7a6b77810e01874f2R120

you might consider sending a PR to https://github.com/github/gitignore

src/garage/torch/policies/continuous_mlp_policy.py
```diff
@@ -0,0 +1,70 @@
...
+                A str contains the name of the policy.
+        """
+        assert isinstance(env_spec.action_space, akro.Box)
+
+        super().__init__(env_spec)
+
+        self._env_spec = env_spec
+        self._nn_module = nn_module
+        self._name = name
+        if input_include_goal:
```
ryanjulian: this is some technical debt from the HER implementation in TF -- please remove and do it the right way in Torch (the goal should be added to the input before it gets to this module)

src/garage/torch/policies/continuous_mlp_policy.py
```diff
@@ -0,0 +1,70 @@
...
+                 name='ContinuousMLPPolicy',
+                 input_include_goal=False):
+        """
+        Initialize class with multiple attributes.
+
+        Args:
+            env_spec():
+            nn_module():
+                A PyTorch module.
+            name(str, optional):
```
ryanjulian: no need to use the `optional` tags, since our documentation generator detects this automagically

src/garage/torch/policies/continuous_mlp_policy.py
ryanjulian: per the [example](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html#example-google), please put a space between the param name and type: `param (type)`

ryanjulian: make sure to add types to all these params!

src/garage/torch/policies/continuous_mlp_policy.py
ryanjulian: i think we are removing `name` parameters, since they are not necessary in torch (and soon not in TF2.0)

src/garage/torch/policies/continuous_mlp_policy.py
ryanjulian: to be consistent with the TF API, please also create a (very small) class which is actually `ContinuousMLPPolicy`. This is a really nice helper class though.

src/garage/torch/policies/continuous_mlp_policy.py
```diff
@@ -0,0 +1,70 @@
...
+        return torch.distributions.Uniform(x, x)
+
+    @overrides
+    def get_action(self, observation):
+        """Return a single action."""
+        with torch.no_grad():
+            d = self.forward(observation.unsqueeze(0))
+            return d.sample().detach().numpy()
+
+    @overrides
```
ryanjulian: no need to use @overrides -- we have a TODO to remove it.

src/garage/torch/q_functions/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""Pytorch q functions."""
```
ryanjulian:
```python
"""PyTorch Q-functions"""
```

src/garage/torch/q_functions/continuous_mlp_q_function.py
```diff
@@ -0,0 +1,40 @@
+"""This modules creates a continuous MLP Q fcuntion network."""
+
+
+class ContinuousMLPQFunction:
+    """
+    This class implements a q value network.
```
ryanjulian: No need for "This class implements" -- you can always start with "Implements a Q-Value network..."

src/garage/torch/q_functions/continuous_mlp_q_function.py
```diff
@@ -0,0 +1,40 @@
...
+
+    It predicts the q value for all actions based on the input state. It uses
```
ryanjulian: Q-value

src/garage/torch/q_functions/continuous_mlp_q_function.py
```diff
@@ -0,0 +1,40 @@
...
+                 nn_module,
+                 name='ContinuousMLPQFunction',
+                 input_include_goal=False):
+        """
+        Initialize class with multiple attributes.
+
+        Args:
+            env_spec():
+            nn_module():
+                A PyTorch module.
```
ryanjulian: I don't think these need to be newline/indented? don't do that unless you go over the column limit

tests/garage/torch/policies/test_continuous_mlp_policy.py
ryanjulian: honestly i think Mocks are better than the fixtures, unless it is overly-cumbersome to use mock for this case.

in any case, yes, it makes no sense to mock a fixture.

tests/garage/torch/policies/test_continuous_mlp_policy.py
```diff
@@ -0,0 +1,65 @@
...
+import numpy as np
+import torch
+from torch import nn
+
+from garage.torch.modules import MLPModule
+from garage.torch.policies import ContinuousMLPPolicy
+from tests.fixtures.envs.dummy import DummyBoxEnv
+
+
+class TestContinuousPolicies(unittest.TestCase):
```
ryanjulian: the test format has changed and not longer relies on `unittest.TestCase`

tests/garage/torch/q_functions/test_continuous_mlp_q_function.py
```diff
@@ -0,0 +1,65 @@
...
+import numpy as np
+import torch
+from torch import nn
+
+from garage.torch.modules import MLPModule
+from garage.torch.q_functions import ContinuousMLPQFunction
+from tests.fixtures.envs.dummy import DummyBoxEnv
+
+
+class TestContinuousMLPQFunction(unittest.TestCase):
```
ryanjulian: please add more comprehensive test cases which test many parameter variations and input/output pairs.

you can see https://github.com/rlworkgroup/garage/blob/master/tests/garage/tf/q_functions/test_continuous_mlp_q_function_with_model.py and https://github.com/rlworkgroup/garage/pull/722 for samples

src/garage/torch/policies/__init__.py
```diff
@@ -0,0 +1,4 @@
+"""Pytorch policies."""
```
ryanjulian:
```python
"""PyTorch policies."""
```

src/garage/torch/policies/continuous_mlp_policy.py
lywong92: I talked to @zhanpenghe yesterday, and we talked about making the this policy class module agnostic and adding a parameter that stores the policy type (deterministic or stochastic). If type is deterministic, then get_action simply returns the forward call. Otherwise, it samples from the distribution returned by the forward call. Does that sound okay to everyone?

Also, we talked about not following the TF implementation since torch does things differently, so do we still want a ContinuousMLPPolicy class in this case? Lmk what you guys think.

krzentner: I think we should have a `ContinuousMLPPolicy` helper class, but it can wait for a later PR.

ryanjulian: changing your type behavior based on constructor parameters is an antipattern, even if someone else did it in their algorithm. 

i think it's fine to wait for `ContinuousMLPPolicy` in a different PR, but in that case please update the commit message and PR description (and file an issue, assign it to yourself).

src/garage/torch/policies/continuous_mlp_policy.py
ryanjulian: i agree -- just remove it. does torch even accept name params?

zhanpenghe: No.

src/garage/torch/q_functions/continuous_mlp_q_function.py
zhanpenghe: Can we also make this a module so that we don't need to do something like `qf.module(observations, actions)`? Instead, we can do `qf(observations, actions)` to build a graph?

src/garage/torch/q_functions/continuous_mlp_q_function.py
zhanpenghe: Hmm all actions? What if the action space is continuous instead of discrete?

src/garage/torch/q_functions/continuous_mlp_q_function.py
```diff
@@ -0,0 +1,40 @@
...
+            self._obs_dim = env_spec.observation_space.flat_dim_with_keys(
+                ['observation', 'desired_goal'])
+        else:
+            self._obs_dim = env_spec.observation_space.flat_dim
+        self._action_dim = env_spec.action_space.flat_dim
+        self._action_bound = env_spec.action_space.high
+
+    def get_qval(self, observations):
+        """Return Q value."""
+        return self._nn_module.forward(observations)
```
zhanpenghe: detach?

src/garage/torch/q_functions/continuous_mlp_q_function.py
```diff
@@ -0,0 +1,40 @@
...
+        self._name = name
+        if input_include_goal:
+            self._obs_dim = env_spec.observation_space.flat_dim_with_keys(
+                ['observation', 'desired_goal'])
+        else:
+            self._obs_dim = env_spec.observation_space.flat_dim
+        self._action_dim = env_spec.action_space.flat_dim
+        self._action_bound = env_spec.action_space.high
+
+    def get_qval(self, observations):
```
zhanpenghe: I think q value should condition on observations and actions.

src/garage/torch/policies/continuous_mlp_policy.py
ryanjulian: easy fix, then

src/garage/torch/q_functions/continuous_mlp_q_function.py
ryanjulian: i agree. this is the design we are working towards in TF

src/garage/torch/policies/continuous_mlp_policy.py
ryanjulian: basically, you just have to remove this case and not consider it.

src/garage/torch/policies/continuous_nn_policy.py
```diff
@@ -0,0 +1,70 @@
+"""
+This modules creates a continuous policy network.
```
ryanjulian: continuous-valued

src/garage/torch/policies/continuous_nn_policy.py
```diff
@@ -0,0 +1,70 @@
...
+
+A continuous neural network can be used as policy method in different RL
```
ryanjulian: continuous-valued

src/garage/torch/policies/continuous_nn_policy.py
```diff
@@ -0,0 +1,70 @@
...
+    Implements a module-agnostic policy network.
+
+    The policy network selects action based on the state of the environment.
+    It uses a PyTorch neural network module to fit the function of pi(s).
+    """
+
+    def __init__(self,
+                 env_spec,
+                 nn_module,
+                 policy_type='Deterministic',
```
ryanjulian: please use lowercase

src/garage/torch/policies/continuous_nn_policy.py
```diff
@@ -0,0 +1,70 @@
...
+                 input_include_goal=False):
+        """
+        Initialize class with multiple attributes.
+
+        Args:
+            env_spec (garage.envs.env_spec.EnvSpec): Environment specification.
+            nn_module (nn.Module): Neural network module in PyTorch.
+            policy_type (str): Type of policy, deterministic or stochastic.
+        """
+        assert isinstance(env_spec.action_space, akro.Box)
```
ryanjulian: please check/raise a ValueError, rather than asserting

src/garage/torch/policies/continuous_nn_policy.py
```diff
@@ -0,0 +1,70 @@
...
+
+        super().__init__(env_spec)
+
+        self._env_spec = env_spec
+        self._nn_module = nn_module
+        self._policy_type = policy_type
+        if input_include_goal:
```
ryanjulian: you can remove this -- it was implemented hastily in TF and should be done differently in torch.

src/garage/torch/policies/continuous_nn_policy.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def forward(self, input_val):
+        """Forward method."""
+        return self._nn_module(input_val)
+
+    def get_action(self, observation):
+        """Return a single action."""
+        with torch.no_grad():
+            x = self.forward(observation.unsqueeze(0))
+            if self._policy_type == 'Deterministic':
```
ryanjulian: when dispatching on an enum like policy type, *always* explicitly test all valid values, and raise an error for a bad one:

```python
if self_policy_type == 'deterministic':
    # stuff
elif self_policy_type == 'stochastic':
   # stuff
else:
    # It's a RuntimeError because it should have been validated in the constructor, so we should never expect this to be wrong here.
    raise RuntimeError('Unkown policy_type, valid values are "stochastic" or "deterministic", but got value "{}"'.format(self._policy_type))
```

src/garage/torch/policies/continuous_nn_policy.py
```diff
@@ -0,0 +1,70 @@
...
+        """
+        Initialize class with multiple attributes.
+
+        Args:
+            env_spec (garage.envs.env_spec.EnvSpec): Environment specification.
+            nn_module (nn.Module): Neural network module in PyTorch.
+            policy_type (str): Type of policy, deterministic or stochastic.
+        """
+        assert isinstance(env_spec.action_space, akro.Box)
+
```
ryanjulian: you need to validate `policy_type` here and raise a ValueError if it's not a valid type.

src/garage/torch/policies/continuous_nn_policy.py
ryanjulian: is this really cleaner than inheritance? @krzentner

ryanjulian: alternatively, we can make all policies stochastic and use a deterministic (dirac delta) distribution to implement determinism, but i'm not sure if that would mess up things like log_likelihood (probably would)

src/garage/torch/q_functions/continuous_nn_q_function.py
```diff
@@ -0,0 +1,34 @@
...
+        Initialize class with multiple attributes.
+
+        Args:
+            env_spec (garage.envs.env_spec.EnvSpec): Environment specification.
+            nn_module (nn.Module): Neural network module in PyTorch.
+        """
+        self._env_spec = env_spec
+        self._nn_module = nn_module
+        if input_include_goal:
+            self._obs_dim = env_spec.observation_space.flat_dim_with_keys(
```
ryanjulian: please remove this goal-conditioned stuff. this is the wrong layer to do it in.

src/garage/torch/policies/continuous_nn_policy.py
krzentner: Yeah, I was always intending us to use a "deterministic" distribution to implement this, and so that the API is always the same. I'm not sure where this `policy_type` idea came from, and I don't really like it, since it will makes API types "change at a distance."

src/garage/torch/policies/continuous_nn_policy.py
krzentner: I really think it's easier to just not have `policy_type` be a thing.

src/garage/torch/policies/continuous_nn_policy.py
krzentner: Can you please just delete `policy_type` and put it back to how it was? (i.e. forward returning a `Uniform` distribution covering only one value.)

src/garage/torch/policies/continuous_nn_policy.py
```diff
@@ -0,0 +1,70 @@
...
+            self._obs_dim = env_spec.observation_space.flat_dim_with_keys(
+                ['observation', 'desired_goal'])
+        else:
+            self._obs_dim = env_spec.observation_space.flat_dim
+        self._action_dim = env_spec.action_space.flat_dim
+        self._action_bound = env_spec.action_space.high
+
+    def forward(self, input_val):
+        """Forward method."""
+        return self._nn_module(input_val)
```
krzentner: Can you wrap the module's output in a `Uniform` distribution here, as we discussed?

src/garage/torch/policies/deterministic_policy.py
```diff
@@ -0,0 +1,48 @@
...
+        self._action_bound = env_spec.action_space.high
+
+    def forward(self, input_val):
+        """Forward method."""
+        return self._nn_module(input_val)
+
+    def get_action(self, observation):
+        """Return a single action."""
+        with torch.no_grad():
+            x = self.forward(observation.unsqueeze(0))
```
krzentner: You can just return `x` from this function. There's no need to wrap and unwrap in the same function.

src/garage/torch/policies/deterministic_policy.py
```diff
@@ -0,0 +1,48 @@
...
+            d = torch.distributions.Uniform(x, x)
+            return d.rsample().numpy().squeeze(0)
+
+    def get_actions(self, observations):
+        """Return multiple actions."""
+        with torch.no_grad():
+            x = self.forward(observations)
+            d = torch.distributions.Uniform(x, x)
```
krzentner: Same as above, just return `x`.

src/garage/tf/core/parameter.py
```diff
@@ -1,7 +1,6 @@
 """Parameter layer in TensorFlow."""
 
 import tensorflow as tf
-from tensorflow.python.ops.gen_array_ops import broadcast_to
```
krzentner: Why is this file in this change? Please undo the changes to it (`git checkout origin/master src/garage/tf/core/parameter.py`).

ryanjulian:
In the future, I recommend against ever creating merge commits. Yes we squash them when we merge the PR, but they can really mess with the CI infrastructure and you might be strange results.

PR #765, opened 2019-07-03
Refactor GaussianConvRegressor

nish21:
This PR refactors GaussianConvRegressor to use garage.tf.models.
Also adds unit tests for the classes.

### DIFF
--- a/src/garage/tf/baselines/gaussian_conv_baseline_with_model.py
+++ b/src/garage/tf/baselines/gaussian_conv_baseline_with_model.py
@@ -1,0 +1,67 @@
+"""Gaussian Conv Baseline."""
+import numpy as np
+
+from garage.misc.overrides import overrides
+from garage.np.baselines import Baseline
+from garage.tf.regressors import GaussianConvRegressorWithModel
+
+
+class GaussianConvBaselineWithModel(Baseline):
+    """A value function using Gaussian Conv network."""
+
+    def __init__(
+            self,
+            env_spec,
+            subsample_factor=1.,
+            regressor_args=None,
+            name='GaussianConvBaselineWithModel',
+    ):
+        """
+        Gaussian Conv Baseline with Model.
+
+        It fits the input data to a gaussian distribution estimated by
+        a Conv.
+
+        Args:
+            env_spec (garage.envs.env_spec.EnvSpec): Environment specification.
+            subsample_factor (float): The factor to subsample the data. By
+                default it is 1.0, which means using all the data.
+            regressor_args (dict): Arguments for regressor.
+        """
+        super().__init__(env_spec)
+        if regressor_args is None:
+            regressor_args = dict()
+
+        self._regressor = GaussianConvRegressorWithModel(
+            input_shape=(env_spec.observation_space.shape),
+            output_dim=1,
+            name=name,
+            **regressor_args)
+        self.name = name
+
+    @overrides
+    def fit(self, paths):
+        """Fit regressor based on paths."""
+        observations = np.concatenate([p['observations'] for p in paths])
+        returns = np.concatenate([p['returns'] for p in paths])
+        self._regressor.fit(observations, returns.reshape((-1, 1)))
+
+    @overrides
+    def predict(self, path):
+        """Predict value based on paths."""
+        return self._regressor.predict(path['observations']).flatten()
+
+    @overrides
+    def get_param_values(self, **tags):
+        """Get parameter values."""
+        return self._regressor.get_param_values(**tags)
+
+    @overrides
+    def set_param_values(self, flattened_params, **tags):
+        """Set parameter values to val."""
+        self._regressor.set_param_values(flattened_params, **tags)
+
+    @overrides
+    def get_params_internal(self, **tags):
+        """Get parameter values."""
+        return self._regressor.get_params_internal(**tags)
--- a/src/garage/tf/core/parameter.py
+++ b/src/garage/tf/core/parameter.py
@@ -9,7 +9,8 @@ def parameter(input_var,
               initializer=tf.zeros_initializer(),
               dtype=tf.float32,
               trainable=True,
-              name='parameter'):
+              name='parameter',
+              pixel_env=False):
     """
     Parameter layer.
 
@@ -26,6 +27,7 @@ def parameter(input_var,
         dtype: Data type of the variables (default is tf.float32).
         trainable (bool): Whether these variables are trainable.
         name (str): Variable scope of the variables.
+        pixel_env (bool): Whether the environment is pixel based or not.
 
     Return:
         A tensor of the broadcasted variables.
@@ -38,7 +40,11 @@ def parameter(input_var,
             initializer=initializer,
             trainable=trainable)
 
-        broadcast_shape = tf.concat(
-            axis=0, values=[tf.shape(input_var)[:-1], [length]])
+        if pixel_env:
+            batch_dim = [tf.shape(input_var)[0]]
+        else:
+            batch_dim = tf.shape(input_var)[:-1]
+
+        broadcast_shape = tf.concat(axis=0, values=[batch_dim, [length]])
         p_broadcast = broadcast_to(p, shape=broadcast_shape)
         return p_broadcast
--- a/src/garage/tf/core/parameter.py
+++ b/src/garage/tf/core/parameter.py
@@ -27,7 +27,6 @@ def parameter(input_var,
         dtype: Data type of the variables (default is tf.float32).
         trainable (bool): Whether these variables are trainable.
         name (str): Variable scope of the variables.
-        pixel_env (bool): Whether the environment is pixel based or not.
 
     Return:
         A tensor of the broadcasted variables.
--- a/src/garage/tf/core/parameter.py
+++ b/src/garage/tf/core/parameter.py
@@ -6,11 +6,11 @@ from tensorflow.python.ops.gen_array_ops import broadcast_to
 
 def parameter(input_var,
               length,
+              batch_dim=None,
               initializer=tf.zeros_initializer(),
               dtype=tf.float32,
               trainable=True,
-              name='parameter',
-              pixel_env=False):
+              name='parameter'):
     """
     Parameter layer.
 
@@ -40,11 +40,8 @@ def parameter(input_var,
             initializer=initializer,
             trainable=trainable)
 
-        if pixel_env:
-            batch_dim = [tf.shape(input_var)[0]]
-        else:
+        if batch_dim is None:
             batch_dim = tf.shape(input_var)[:-1]
-
         broadcast_shape = tf.concat(axis=0, values=[batch_dim, [length]])
         p_broadcast = broadcast_to(p, shape=broadcast_shape)
         return p_broadcast
--- a/src/garage/tf/models/__init__.py
+++ b/src/garage/tf/models/__init__.py
@@ -1,7 +1,7 @@
 from garage.tf.models.base import Model
 from garage.tf.models.cnn_model import CNNModel
 from garage.tf.models.cnn_model_max_pooling import CNNModelWithMaxPooling
-from garage.tf.models.gaussian_conv_model import GaussianConvModel
+from garage.tf.models.gaussian_cnn_model import GaussianCNNModel
 from garage.tf.models.gaussian_gru_model import GaussianGRUModel
 from garage.tf.models.gaussian_lstm_model import GaussianLSTMModel
 from garage.tf.models.gaussian_mlp_model import GaussianMLPModel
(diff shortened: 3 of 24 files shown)


src/garage/tf/baselines/gaussian_conv_baseline_with_model.py
```diff
@@ -0,0 +1,67 @@
...
+            self,
+            env_spec,
+            subsample_factor=1.,
+            regressor_args=None,
+            name='GaussianConvBaselineWithModel',
+    ):
+        """
+        Gaussian Conv Baseline with Model.
+
+        It fits the input data to a gaussian distribution estimated by
```
ryanjulian: Gaussian

src/garage/tf/baselines/gaussian_conv_baseline_with_model.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def __init__(
+            self,
+            env_spec,
+            subsample_factor=1.,
+            regressor_args=None,
+            name='GaussianConvBaselineWithModel',
+    ):
+        """
+        Gaussian Conv Baseline with Model.
```
ryanjulian: GaussianConvBaseline

src/garage/tf/baselines/gaussian_conv_baseline_with_model.py
```diff
@@ -0,0 +1,67 @@
...
+
+        It fits the input data to a gaussian distribution estimated by
+        a Conv.
```
ryanjulian: a convolutional neural network (CNN).

src/garage/tf/core/parameter.py
```diff
@@ -9,7 +9,8 @@ def parameter(input_var,
               initializer=tf.zeros_initializer(),
               dtype=tf.float32,
               trainable=True,
-              name='parameter'):
+              name='parameter',
+              pixel_env=False):
```
ryanjulian: I don't think this is an appropriate API. it breaks several levels of abstraction. there's no way a parameter layer needs to know whether or no the environment uses pixels.

src/garage/tf/baselines/gaussian_conv_baseline_with_model.py
```diff
@@ -0,0 +1,67 @@
+"""Gaussian Conv Baseline."""
+import numpy as np
+
+from garage.misc.overrides import overrides
+from garage.np.baselines import Baseline
+from garage.tf.regressors import GaussianConvRegressorWithModel
+
+
+class GaussianConvBaselineWithModel(Baseline):
+    """A value function using Gaussian Conv network."""
```
ryanjulian: which fits to a Gaussian model whose parameters are estimated by a CNN.

src/garage/tf/core/parameter.py
```diff
@@ -38,7 +40,11 @@ def parameter(input_var,
             initializer=initializer,
             trainable=trainable)
 
-        broadcast_shape = tf.concat(
-            axis=0, values=[tf.shape(input_var)[:-1], [length]])
+        if pixel_env:
+            batch_dim = [tf.shape(input_var)[0]]
+        else:
+            batch_dim = tf.shape(input_var)[:-1]
```
ryanjulian: Can you please explain why this is necessary? are you automatically flattening 2D inputs?

src/garage/tf/models/__init__.py
```diff
@@ -15,7 +16,7 @@
 
 __all__ = [
     'CNNModel', 'CNNModelWithMaxPooling', 'LSTMModel', 'Model',
-    'GaussianGRUModel', 'GaussianLSTMModel', 'GaussianMLPModel', 'GRUModel',
-    'MLPDuelingModel', 'MLPMergeModel', 'MLPModel', 'NormalizedInputMLPModel',
-    'Sequential'
+    'GaussianConvModel', 'GaussianGRUModel', 'GaussianLSTMModel',
```
ryanjulian: Can you take this opportunity to rename all of the *Conv* primitives to *CNN* instead? it's more consistent with there rest of the codebase.

src/garage/tf/models/gaussian_conv_model.py
```diff
@@ -0,0 +1,301 @@
...
+from garage.tf.models.base import Model
+
+
+class GaussianConvModel(Model):
+    """
+    GaussianConvModel.
+
+    Args:
+        filter_dims(tuple[int]): Dimension of the filters. For example,
+            (3, 5) means there are two convolutional layers. The filter
```
ryanjulian: Can I have 2 filters which are (5x7) followed by (3x3)? why not? how would I specify that? `((5, 7), (3, 3))`?

src/garage/tf/models/gaussian_conv_model.py
```diff
@@ -0,0 +1,301 @@
...
+                 output_w_init=tf.glorot_uniform_initializer(),
+                 output_b_init=tf.zeros_initializer(),
+                 learn_std=True,
+                 adaptive_std=False,
+                 std_share_network=False,
+                 init_std=1.0,
+                 min_std=1e-6,
+                 max_std=None,
+                 std_filter_dims=[],
+                 std_num_filters=[],
```
ryanjulian: doesn't filter_dims tell me how many filters there should be?

src/garage/tf/models/gaussian_conv_model.py
```diff
@@ -0,0 +1,301 @@
...
+                        layer_normalization=self._layer_normalization)
+                else:
+                    log_std_network = parameter(
+                        state_input,
+                        length=action_dim,
+                        initializer=tf.constant_initializer(
+                            self._init_std_param),
+                        trainable=self._learn_std,
+                        name='log_std_network',
+                        pixel_env=True)
```
ryanjulian: um -- why can't i just have a 2D parameter?

src/garage/tf/models/gaussian_conv_model.py
```diff
@@ -0,0 +1,301 @@
...
+                std_param = tf.minimum(std_param, self._max_std_param)
+
+        with tf.variable_scope('std_parameterization'):
+            # build std_var with std parameterization
+            if self._std_parameterization == 'exp':
+                log_std_var = std_param
+            else:  # we know it must be softplus here
+                log_std_var = tf.log(tf.log(1. + tf.exp(std_param)))
+
+        dist = DiagonalGaussian(self._output_dim)
```
ryanjulian: please also use a tensorflow_probability.distributions.MultivariateNormDiag, as in GaussianMLPModel

src/garage/tf/core/parameter.py
ahtsan: @ryanjulian I discussed with Nisanth and realized that how batch size is obtained from the input.shape is different under different circumstances, e.g. shape[0] for pixel environment while shape[:-1] for recurrent policies. I am currently investigating if parameter class is correctly implemented for recurrent policies too as GaussianLSTMPolicyWithModel seems not working in benchmark. 

Maybe a better way is to create another `parameter_conv` class that handles pixel environments specifically? @nish21

src/garage/tf/models/gaussian_conv_model.py
ahtsan: I think currently we are still using `DiagonalGaussian` for `Gaussian*Model`.

src/garage/tf/core/parameter.py
nish21: Yes, I believe @ahtsan covered most of it. The parameter class broadcasts a variable of shape `(L, )` to shape `(batch_dim, L)`. And the batch dimensions to pick is different based on whether a pixel based input is being used or a recurrent policy is being used. 
One other way to implement this would be for the class creating the parameter to specify which dimensions of the input is batch_dim.
So the API would then be 
```
parameter(*some_inputs, batch_dim=input_shape[:0])
and 
def parameter(...):
    ...
    broadcast_shape = tf.concat(values=[batch_dim, [length]])
```

src/garage/tf/baselines/gaussian_conv_baseline_with_model.py
nish21: will change

src/garage/tf/core/parameter.py
nish21: No, I'm not flattening it. Like I explained in the earlier comment, if we have an input_var of shape `(100, 10, 10, 3)` that is 100 images of shape `(10, 10, 3)`, and my variable of shape 
`(1, )` needs to be `(100, 1)`. I'm picking the batch_dim from the input shape and using that to broadcast the variable. For recurrent policies it needs to be `input_shape[:-1]`

src/garage/tf/models/__init__.py
nish21: sure, will do that

src/garage/tf/models/gaussian_conv_model.py
ryanjulian: you are right. please ignore.

src/garage/tf/models/gaussian_conv_model.py
nish21: `filter_dims` specifies the dimension of the filters, `num_filters` specifies the number of filters to use of that dimension, `std_*` arguments specify the same parameters for the `std` CNN.

src/garage/tf/core/parameter.py
ryanjulian: i think the right thing to do is figure out what is going on and file an issue.

for now you can add a parameter specifying which input dimension is the batch dimension.

for the issue
1. we should figure out if we can keep the batch dimension in the same position for all data (preferably at `shape[0]`)
1. if there are pressing performance reasons why we can't keep the batch_dim at 0, implement a consistent fix across the codebase
1. make sure LSTM is not broken (and fix it if it is, add tests to catch the bug)

src/garage/tf/models/gaussian_conv_model.py
nish21: I don't think the implementation of CNN in garage supports filters of different dimensions like `(5,7)`.

ahtsan: It doesn't but I agree we should have support for that. I suggest filing an issue and making that as a feature to be implemented later as that's out of the scope of this PR.

nish21: Yes, will file an issue.

src/garage/tf/models/gaussian_conv_model.py
ryanjulian: are you saying that std_filter_dims and std_num_filters is a run-length encoding? i.e. the total number of filters is `sum(std_num_filters)`?

nish21: yes, i guess so. `num_filters = (3, 6)` and `filter_dims = (5, 3)` implies 3 filters of  shape `(5x5)` and 6 filters of shape `(3x3)`

ryanjulian: ok that is a strange way of doing it and hard to test/validate. can you file an issue to change the API to just use a list of lists/tuple of tuples?

it can be the same issue as the comment above about filter_dims

nish21: Sure, I'll file one now

src/garage/tf/core/parameter.py
```diff
@@ -26,6 +27,7 @@ def parameter(input_var,
         dtype: Data type of the variables (default is tf.float32).
         trainable (bool): Whether these variables are trainable.
         name (str): Variable scope of the variables.
+        pixel_env (bool): Whether the environment is pixel based or not.
```
ryanjulian: i think you forgot to update the documentation here.

nish21: oops, will change that.

tests/garage/tf/models/test_gaussian_cnn_model.py
```diff
@@ -0,0 +1,630 @@
...
+        input_shape = self.obs.shape[1:]  # height, width, channel
+        self._input_ph = tf.placeholder(
+            tf.float32, shape=(None, ) + input_shape, name='input')
+
+    @mock.patch('tensorflow.random.normal')
+    @pytest.mark.parametrize(
+        'filter_sizes, in_channels, out_channels, '
+        'strides, output_dim, hidden_sizes',
+        [
+            ((1, ), (3, ), (3, ), (1, ), 1, (1, )),  # noqa: E122
```
ryanjulian: why not follow E122? i think you can still have block formatting...

tests/garage/tf/regressors/test_gaussian_cnn_regressor_with_model.py
```diff
@@ -0,0 +1,280 @@
...
+            output_dim=1,
+            adaptive_std=False,
+            use_trust_region=False)
+
+        new_input_var = tf.placeholder(
+            tf.float32, shape=(None, ) + input_shape)
+        new_ys_var = tf.placeholder(
+            dtype=tf.float32, name='ys', shape=(None, output_dim))
+
+        data = np.random.random(size=input_shape)
```
ryanjulian: were you not able to find a way to test without random data?

nish21: I didn't realize I was using random data in this test case, I'll fix it.

ryanjulian:
great job!

Issue #444, opened 2018-12-20
Remove conda from the setup scripts


amolchanov86:
As the title says, it would be better if the installation script would identify if the user already has anaconda running and simply add env to it, rather than installing miniconda and adding env there.
Right now, I hack it by providing a symbolic link from _~/miniconda/envs/garage_ to _~/anaconda2/envs/garage_  but the proper fix would be appropriate.

ryanjulian:
This is actually intentional because we only test with miniconda2, and we don't want to provide installation routines for configurations which we don't test.

I didn't see it as a bug, since anaconda/miniconda can coexist peacefully.

What do you think about instead providing an arg to the setup script which allows the user to specify an existing conda to use (along with a warning message that we haven't tested your conda)?

amolchanov86:
Btw, one good update to documentation would be to add an explicit note that "if you are using anaconda, please follow the Manual installation" :)
I have just noticed that I shouldn't have used the express :)

ryanjulian:
Okay, let's make that part of the scope for this issue.

PR #770, opened 2019-07-06
Remove f-expression from CEM

ryanjulian:
This is not Python 3.5 compatible.


ryanjulian:
This needs to wait on #444

ryanjulian:
will be fixed in #774

PR #771, opened 2019-07-06
Pin pyglet version to the latest compatible with gym 0.12.4

### LINKED ISSUE #768: Nightly builds are failing
https://travis-ci.com/rlworkgroup/garage/jobs/213465246

I think it may have been a API change in pyglet. we may need to pin that dep.

gitanshu:
Closes #768

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,6 +21,10 @@ required = [
     'mujoco-py<2.1,>=2.0',
     'numpy==1.14.5',
     'psutil',
+    # Pyglet 1.4.0 introduces some api change which breaks some
+    # gym environments
+    # See: https://github.com/openai/gym/issues/1588
+    'pyglet<1.4.0,>=1.3.0',
     'pyprind',
     'python-dateutil',
     'torch==1.1.0',


setup.py
```diff
@@ -21,6 +21,7 @@
     'mujoco-py<2.1,>=2.0',
     'numpy==1.14.5',
     'psutil',
+    'pyglet<1.4.0,>=1.3.0',
```
ryanjulian: can you leave a comment about why this is pinned?

krzentner:
I'm confused. The changes on this branch do a few things, but don't actually seem to pin pyglet. Did I miss something?

gitanshu:
> I'm confused. The changes on this branch do a few things, but don't actually seem to pin pyglet. Did I miss something?

It started with pinning the pyglet version, then Ryan found that gym started pinning the pyglet version in the latest release, so he switched to that. But gym also updated atari-py in 0.13.x, and so now when we do spec.make() on atari envs, it hangs up at a [native call here](https://github.com/openai/atari-py/blob/7649e3891f45d99cab3174548bf59f2040a358a2/atari_py/ale_python_interface.py#L149).

So I'd say we rollback to pinning pyglet version for now. What do you think?

ryanjulian:
I am so tired of fixing this bug.

@gitanshu after the release, can you send a PR to openai/gym fixing this?

ryanjulian:
ok please roll this back to the original fix and merge it, and file an issue to update gym.

ryanjulian:
great work, thanks for the triage.

PR #772, opened 2019-07-07
Set pydocstyle version to less than 4.0.0

gitanshu:
pydocstyle 4.0.0 was released 18 hours ago and it breaks flake8-docstrings 1.3.0 which in turn is breaking our CI builds

https://gitlab.com/pycqa/flake8-docstrings/issues/36


ryanjulian:
Can you generate the same PR for akro, metaworlds, viskit, and dowel?

PR #773, opened 2019-07-07
Stop enforcing W504; Remove check_flake8.sh, use flake8 directly

### LINKED ISSUE #758: Remove check_flake8.sh
This was created to help us check new code for flake8 compliance, even though the codebase at the time was really bad.

We now have brought the codebase up-to-date, and check_flake8 is no longer needed.

This can be done as soon as #409 is closed.

It will require:
* Switching .pre-commit-config.yaml to use the default flake8 check
* Verifying that flake8 still catches problems in the CI (in test, examples, src, and root)
* Verifying that flake8 still catches missing documentation in the CI for new code

gitanshu:
Closes #758


src/garage/envs/base.py
```diff
@@ -94,9 +93,11 @@ def _close_viewer_window(self):
                     glfw.destroy_window(self.env.viewer.window)
             elif any(package in self.env.spec._entry_point
                      for package in KNOWN_GYM_NOT_CLOSE_VIEWER):
-                if (hasattr(self.env, 'viewer') and
-                    (isinstance(self.env.viewer, Viewer)
-                     or isinstance(self.env.viewer, SimpleImageViewer))):
+                # yapf: disable
```
ryanjulian: hmm -- if this version of YAPF still making the wrong splits?

src/garage/envs/base.py
```diff
@@ -94,9 +93,11 @@ def _close_viewer_window(self):
...
+                if (hasattr(self.env, 'viewer')
+                    and (isinstance(self.env.viewer, Viewer)
+                         or isinstance(self.env.viewer, SimpleImageViewer))):
+                    # yapf: enable
```
ryanjulian: is YAPF not behaving?

src/garage/tf/algos/dqn.py
```diff
@@ -185,8 +184,10 @@ def train_once(self, itr, paths):
         self.episode_rewards.extend(paths['undiscounted_returns'])
         last_average_return = np.mean(self.episode_rewards)
         for train_itr in range(self.n_train_steps):
-            if (self.replay_buffer.n_transitions_stored >=
-                    self.min_buffer_size):
+            # yapf: disable
```
ryanjulian: it seems like YAPF is misbehaving for comparison or logical operators?

src/garage/tf/algos/dqn.py
```diff
@@ -80,22 +80,21 @@ def __init__(self,
...
-            n_epoch_cycles=n_epoch_cycles,
-            buffer_batch_size=buffer_batch_size,
-            rollout_batch_size=rollout_batch_size,
-            replay_buffer=replay_buffer,
-            max_path_length=max_path_length,
-            discount=discount,
-            reward_scale=reward_scale,
-            input_include_goal=input_include_goal,
-            smooth_return=smooth_return)
+        super(DQN, self).__init__(env_spec=env_spec,
```
zequnyu: `super().__init__()`

src/garage/tf/algos/dqn.py
gitanshu: It works fine at some places for logical operators like https://github.com/rlworkgroup/garage/pull/773/files#diff-eec0e5bc5f0646f0458d535037a7bcf0R363 & https://github.com/rlworkgroup/garage/pull/773/files#diff-eec0e5bc5f0646f0458d535037a7bcf0R363
but goes crazy at others. And yapf doesn't have a flag that covers comparison operators at all

gitanshu: It also messed up around some arithmetic operators sometimes like if the an arithmetic expression is passed to a function call such as [here](https://github.com/rlworkgroup/garage/pull/773/files#diff-bb9dfa32db8aec7c259bc87d036131adR53)

ryanjulian: i think comparison operators are considered logical operators -- but yes every time i've tried this, YAPF has had terrible W504 support.

you may also want to try `split_before_bitwise_operator`

you can increase `split_penalty_arithmetic_operator`, `split_penalty_bitwise_operator` and `split_penalty_bitwise_operator` if that doesn't work.  the defaults are here: https://github.com/google/yapf/blob/45cf63a4538a2d5790db5182b84aca8c1b3bc491/yapf/yapflib/style.py#L341 -- just try doubling them until it behaves correctly.

gitanshu: [yapf source](https://github.com/google/yapf/blob/27fe4ad04fd61979ae8b734d13a569a29a3b79bd/yapf/yapflib/unwrapped_line.py#L520) has only 'and' and 'or' as logical operators. I don't see comparison operators mentioned at all.

Increasing the penalties helps 1 or 2 cases of logical operators in terms of line break but still not the best for readability. It largely does no good for arithmetic operators, and instead increasing it too much for arithmetic operators causes unexpected changes. 

TLDR: even with all these knobs and high penalties, yapf doesn't produce expected output

ryanjulian:
can you also address #758 in this PR?

gitanshu:
I tried switching to the default pre-commit flake8 hook [as described here](https://flake8.pycqa.org/en/latest/user/using-hooks.html) but it runs into the [pydocstyle version issue](https://gitlab.com/pycqa/flake8-docstrings/issues/36). I am guessing pre-commit runs the cloned repos in it's own environment which is why the version of pydocstyle we pin in setup.py doesn't get picked up here

ryanjulian:
@gitanshu to fix the pydocstyle issue, look at how we run pylint in .pre-commit-config.yaml

gitanshu:
> @gitanshu to fix the pydocstyle issue, look at how we run pylint in .pre-commit-config.yaml

Yeah I realised that after my last comment here. Addressed it in https://github.com/rlworkgroup/garage/pull/773/commits/19a0594d86cf129f6abbabb10e8c83fada4955f1

ryanjulian:
What triggered all the changes to hanging indentation? Was it your tweaks to the YAPF config?

I am starting to think the simple solution for now is to disable W504 in the flake8 config. I don't think our tools are ready to support it yet.

gitanshu:
@ryanjulian it could be the version bump from 0.25 to 0.27. There might be a knob to disable it in yapf. I could look into it. But I think there's still gonna be a number of yapf disables in future code too.

As for disabling W504, I guess we could just leave/close this PR and revisit it later when tools support it better.

ryanjulian:
i think removing enforcement of W504 (and any custom YAPF stuff you had to add) + check_flake8.sh would be a net win for the project, because it would simplify the CI pipeline and unify how code is checked for style (the rules in check_flake8 are a little complicated and it tends to break).

.pre-commit-config.yaml
```diff
@@ -9,9 +9,11 @@ repos:
         stages: [commit-msg]
     -   id: check-flake8
         name: Check Flake8
-        language: script
-        entry: scripts/travisci/check_flake8.sh
+        language: system
```
ryanjulian: is it not sufficient to pin the flake8 version both here and in setup.py?

gitanshu: Do you mean without mentioning a language? It's a mandatory key for pre-commit and `system` seems like the most relevant value. `python` could work but it tries to install flake-docstrings in a virtual env and fails on the pydocstyle issue.

ryanjulian: i mean not using system at all (just the default pre-commit rule), but pinning the version to match ours.

gitanshu: We don't pin flake8 version yet, but we could do something like this

```

-   repo: https://gitlab.com/pycqa/flake8
    rev: '3.7.8'
    hooks:
    -   id: flake8
        stages: [commit, push]
        additional_dependencies: ['flake8-import-order', 'flake8-docstrings==1.3.0', 'pydocstyle<4.0.0']
```

Is this what you meant?

.pre-commit-config.yaml
```diff
@@ -9,9 +9,11 @@ repos:
         stages: [commit-msg]
     -   id: check-flake8
         name: Check Flake8
-        language: script
-        entry: scripts/travisci/check_flake8.sh
+        entry: flake8
```
ryanjulian: is there a reason we need to add our own `system` entry, instead of just using the default pre-commit flake8 rule.

ryanjulian: i.e. here https://github.com/rlworkgroup/garage/pull/773/files#diff-82c3b36709926de3a1b27bea0d8258ccL70

.pre-commit-config.yaml
```diff
@@ -9,9 +9,11 @@ repos:
         stages: [commit-msg]
```
ryanjulian: please take a moment to remove the commented-out entries in this file. they are no longer necessary

ryanjulian: the `stages` key is now deprecated, and defaults to our config, so you can remove all of these `stages` keys.

ryanjulian: (except for the commit-msg one, which is not default)

ryanjulian: okay i am now less sure of this change, so please test it, but i think it would help clean up this file a bit.

ryanjulian: ok i confirmed the behavior -- it is save to remove the stages keys.

setup.py
```diff
@@ -70,7 +70,7 @@
     'recommonmark',
     'sphinx',
     'sphinx_rtd_theme',
-    'yapf',
+    'yapf==0.25.0',
```
gitanshu: @ryanjulian If we are setting yapf to 0.28.0 in pre-commit, this should also be 0.28.0, right?

ryanjulian: yes -- we should strive to have them all match.

gitanshu:
Moved to #844

PR #774, opened 2019-07-08
Remove conda from Dockerfiles and setup scripts

### LINKED ISSUE #444: Remove conda from the setup scripts

### LINKED ISSUE #666: Remove conda environment from CI/Dockerfiles
Now that we use setup.py for tracking dependencies, there's no need be creating conda environments inside the Dockerfile and CI. We can replace them with a vanilla virtualenv for greatest compatibility.

I suggest we add tests which ensure that the package can still be successfully solved and installed by conda and pipenv. If they don't take too long (<2min) then we can run them in the main CI, otherwise they will be cron_job tests (which are run only nightly).

ryanjulian:
This PR removes conda environments from the Dockerfiles and setup
scripts.
    
For the CI, we use a vanilla virtualenv to ensure that the garage
package installs in the lowest common-denonenator environment. We also
add a job to TravisCI to verify that conda can successfully build.
    
For the setup scripts, we remove the environment creation altogether and
assume that the user will install garage in her own environment of
choice. The scripts serve to only setup pre-requisities, not the Python
environment itself
    
Summary of changes:
* Remove conda environment from Dockerfiles and setup scripts
* Add virtualenv to Dockerfiles
* Add TravisCI job to build the conda environment
* Add pytest-xdist for parallel tests
* Update some tests and code to be Python 3.5 compliant

Fixes #444 
Fixes #666

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: python
 
-python: "3.6.6"
+python: "3.5"
 
 services:
   - docker
@@ -22,6 +22,8 @@ jobs:
         - DEPLOY_FROM_THIS_JOB="true"
     - name: "Large tests"
       env: JOB_RUN_CMD="make ci-job-large"
+    - name: "Verify conda and pipenv installations"
+      env: JOB_RUN_CMD="make ci-job-verify-envs"
     - if: type = cron
       name: "Nightly tests"
       env: JOB_RUN_CMD="make ci-job-nightly"
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,7 +21,11 @@ After following the standard garage setup steps, make sure to run to install the
 
 To setup pre-commit in your repo:
 ```sh
-conda activate garage
+# make sure your Python environment is activated, e.g.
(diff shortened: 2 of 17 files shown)


zhanpenghe:
Hmm why not just use `pipenv` instead of `virtualenv`?

ryanjulian:
ppipenv is just one environment manager, and it does dependency conflict resolution, and other things which vanilla pip/setuptools installations don't do. conda/pipenv are also slower than just installing into a virtualenv.

conda remains very popular with researchers, and it would be a mistake to just assume we're not going to support using it.

We need to verify that the package can be installed into a vanilla Python environment successfully using pip, because that is the promise of using PyPI. We are also going the extra mile and verifying that conda and pipenv can both successfully find a solution to the dependency list.

PR #775, opened 2019-07-09
Use tf.broadcast_to to pass pylint tests

zhanpenghe:
pylint cannot import from tensorflow internal modules and we should use tensorflow external API anyways.


zhanpenghe:
No he just asks me for help. I asked him to run regression for the models before running benchmarking.

Issue #31, opened 2018-06-11
Create ray sampler


ryanjulian:
Right now rllab uses parallelism in an ad-hoc manner through multiprocessing library, mostly in the sampler. If we use a principled parallelism library (e.g. mpi, ray, others), we can probably clean up the code while avoiding tricky multiprocessing bugs in the future.

Imported from https://github.com/ryanjulian/rllab/issues/82

avnishn:
What I've learned is that in order to do a batch sampler with ray:

- share agent parameters with workers in between sampler iterations, and use those shared parameters to update the model inside the individual workers. 
- This unfortunately means that we can't make a sampler that is agnostic of the library that we are using, as we will need to have custom ways of updating an agents parameters between sampler iterations (e.g. one set of parameter getter and setters for pytorch, and one for tensorflow

- The most efficient way of communicating objects with ray is to put them into numpy array objects. Ray still supports somewhat fast serialization of Primitive types: ints, floats, longs, bools, strings, unicode, and numpy arrays. Any list, dictionary, or tuple whose elements can be serialized by Ray. **But definitely not arbitrary python objects**
- I have noticed that it is easy to get the parameters of torch models using model.state_dict(), and set the parameters using model.load_state_dict()
- With tensorflow, we can get all the variables, and set them by following the same methods that the people who wrote the distributed sgd Ray code did.
- The distributed sgd ray code chooses to pass around generic python lists instead of dictionaries, or numpy arrays, when moving parameters. Its probably worth trying out using dictionaries, or numpy arrays, or lists to pass around data to see which way provides the fastest serialization. Perhaps converting lists to arrays is too slow of an operation on larger models.

avnishn:
Further discoveries:

- Garage tf policies have an option to get and set their parameters using numpy arrays, making the job of passing parameters even easier
- This approach again will make it ultimately hard to separate the sampler from the entire library as a standalone library, assuming that the sampler class will allow for users to use an arbitrary agent with their sampler.

krzentner:
I spent some time looking into and working on this problem last week. I ended up writing a sampler which appears to be faster than any of our previous attempts. I'll put it in a garage branch today so we can discuss the details, but basically it uses a simple state machine per worker and queues directly. This appears to be a better abstraction for our purposes than a pool, which the rllab sampler uses.

[...]

PR #776, opened 2019-07-09
Distributed Ray Sampler

avnishn:
- Built for parallel or distributed on policy batch sampling.
- Useful for if we want to run experiments with distributed sampling.
- Operates in a data parallel model (each processing element does the
  same operation, but with different inputs).
- Also works for local experiments well, but there is an overhead that
  presents itself when using tensorflow nn policies when sampling, as
  described in the next point.
- The first iteration of sampling is slightly slower than the on_policy_vectorized
  sampler on some examples (trpo_swimmer) because of a startup overhead that
  occurs because the first time we run new tf graphs for collecting
  samples. After the first iteration of the sampler, the
  sampler collects trajectories at the same speed as the on_policy_vectorized sampler.
  In the case of the trpo_swimmer example, there is an extra 10 second
  overhead due to the construction of new tensorflow graphs for
  sampling.

Reference to #31

### DIFF
--- a/src/garage/sampler/ray_batched_sampler.py
+++ b/src/garage/sampler/ray_batched_sampler.py
@@ -1,0 +1,215 @@
+"""This is an implementation of an on policy batch sampler.
+
+Uses a data parallel design.
+Included is a sampler that deploys sampler workers.
+
+The sampler workers must implement some type of set agent parameters
+function, and a rollout function
+"""
+from collections import defaultdict
+import pickle
+
+import numpy as np
+import psutil
+import ray
+
+from garage.misc import tensor_utils
+from garage.misc.prog_bar_counter import ProgBarCounter
+from garage.sampler.base import BaseSampler
+
+
+class RaySampler(BaseSampler):
+    """Collects Policy Rollouts in a data parallel fashion.
+
+    Args:
+        - algo: A garage algo object
+        - env: A gym/akro env object
+        - should_render(bool): should the sampler render the trajectories
+        - sampler_worker_cls: If none, uses the default SamplerWorker
+            class
+
+    """
+
+    def __init__(self,
+                 algo,
+                 env,
+                 should_render=False,
+                 num_processors=None,
+                 sampler_worker_cls=None):
+        self.SamplerWorker = ray.remote(SamplerWorker if sampler_worker_cls is
+                                        None else sampler_worker_cls)
+
+        self.env = env
+        self.algo = algo
+        self.max_path_length = self.algo.max_path_length
+        self.should_render = should_render
+        if not ray.is_initialized():
+            ray.init(ignore_reinit_error=True)
+        self.num_workers = num_processors if num_processors \
+            else psutil.cpu_count(logical=False)
+        self.all_workers = defaultdict(None)
+        self.active_workers = []
+        self.active_worker_ids = []
+
+    def start_worker(self):
+        """Initialize a new ray worker."""
+        env_pkl = pickle.dumps(self.env)
+        agent_pkl = pickle.dumps(self.algo.policy)
+        env_pkl_id, agent_pkl_id = ray.put(env_pkl), ray.put(agent_pkl)
+        for worker_id in range(self.num_workers):
+            self.all_workers[worker_id] = (self.SamplerWorker.remote(
+                worker_id, env_pkl_id, agent_pkl_id, self.max_path_length,
+                self.should_render))
+        self.idle_worker_ids = list(range(self.num_workers))
+
+    def obtain_samples(self, itr, num_samples):
+        """Sample the policy for new trajectories.
+
+        Args:
+            - itr(int): iteration number
+            - num_samples(int):number of steps the the sampler should collect
+        """
+        pbar = ProgBarCounter(num_samples)
+        completed_samples = 0
+        traj = []
+        updating_workers = []
+        self.idle_worker_ids = list(range(self.num_workers))
+
+        curr_policy_params = self.algo.policy.get_param_values()
+        params_id = ray.put(curr_policy_params)
+        while self.idle_worker_ids:
+            worker_id = self.idle_worker_ids.pop()
+            worker = self.all_workers[worker_id]
+            updating_workers.append(worker.set_agent.remote(params_id))
+
+        while completed_samples < num_samples:
+            updated, updating_workers = ray.wait(
+                updating_workers, num_returns=1, timeout=0.1)
+            upd = [ray.get(up) for up in updated]
+            self.idle_worker_ids.extend(upd)
+            while self.idle_worker_ids:
+                idle_worker_id = self.idle_worker_ids.pop()
+                self.active_worker_ids.append(idle_worker_id)
+                worker = self.all_workers[idle_worker_id]
+                self.active_workers.append(worker.rollout.remote())
+
+            ready, not_ready = ray.wait(
+                self.active_workers, num_returns=1, timeout=0.001)
+            self.active_workers = not_ready
+            for result in ready:
+                trajectory, num_returned_samples = self._process_trajectory(
+                    result)
+                completed_samples += num_returned_samples
+                pbar.inc(num_returned_samples)
+
+                traj.append(trajectory)
+        pbar.stop()
+        return traj
+
+    def shutdown_worker(self):
+        """Shuts down the worker."""
+        ray.shutdown()
+
+    def _process_trajectory(self, result):
+        trajectory = ray.get(result)
+        ready_worker_id = trajectory[0]
+        self.active_worker_ids.remove(ready_worker_id)
+        self.idle_worker_ids.append(ready_worker_id)
+        trajectory = dict(
+            observations=self.algo.env_spec.observation_space.flatten_n(
+                trajectory[1]),
+            actions=self.algo.env_spec.action_space.flatten_n(trajectory[2]),
+            rewards=tensor_utils.stack_tensor_list(trajectory[3]),
+            agent_infos=trajectory[4],
+            env_infos=trajectory[5])
+        num_returned_samples = len(trajectory['observations'])
+        return trajectory, num_returned_samples
+
+
+class SamplerWorker:
+    """Constructs a single sampler worker.
+
+    The worker can have its parameters updated, and sampler its policy for
+    trajectories or rollouts.
+
+    Args:
+        - worker_id(int): the id of the sampler_worker
+        - env: gym or akro env object
+        - max_path_length(int): max trajectory length
+        - should_render(bool): if true, renders trajectories after
+            sampling them
+
+    """
+
+    def __init__(self,
+                 worker_id,
+                 env,
+                 agent,
+                 max_path_length,
+                 should_render=False):
+        self.worker_id = worker_id
+        self.env = pickle.loads(env)
+        self.agent = pickle.loads(agent)
+        self.max_path_length = max_path_length
+        self.should_render = should_render
+        self.agent_updates = 0
+
+    def set_agent(self, flattened_params):
+        """Set the agent params.
+
+        Args:
+            - flattened_params(): model parameters in numpy format
+        """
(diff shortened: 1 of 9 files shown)


src/garage/sampler/ray_batched_sampler.py
```diff
@@ -0,0 +1,215 @@
...
+from garage.misc.prog_bar_counter import ProgBarCounter
+from garage.sampler.base import BaseSampler
+
+
+class RaySampler(BaseSampler):
+    """Collects Policy Rollouts in a data parallel fashion.
+
+    Args:
+        - algo: A garage algo object
+        - env: A gym/akro env object
```
zhanpenghe: Hmm I think akro only has spaces not envs.

tests/fixtures/policies/ray_sampler_utils.py
```diff
@@ -0,0 +1,49 @@
...
+    'H' or 'o': hole (terminates episode)
+    'G' : goal
+    [2,2,1,0,3,1,1,1,2,2,1,1,1,2,2,1]
+    """
+
+    def __init__(self):
+        self.scriptedActions = [2, 2, 1, 0, 3, 1, 1, 1, 2, 2, 1, 1, 1, 2, 2, 1]
+
+    def set_param_values(self, params):
+        return params
```
zhanpenghe: pass?

tests/fixtures/policies/ray_sampler_utils.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def get_param_values(self):
+        return 'ScriptedPolicy'
+
+    def reset(self, dones=None):
+        pass
+
+    def get_action(self, obs):
+        return self.scriptedActions[obs], {}
```
zhanpenghe: Agent info should be a dictionary.

src/garage/sampler/ray_batched_sampler.py
```diff
@@ -0,0 +1,215 @@
...
+        - should_render(bool): should the sampler render the trajectories
+        - sampler_worker_cls: If none, uses the default SamplerWorker
+            class
+
+    """
+
+    def __init__(self,
+                 algo,
+                 env,
+                 should_render=False,
```
ryanjulian: hmm shouldn't this be a property of the worker class? who would want to render _all_ of their workers?

src/garage/sampler/ray_batched_sampler.py
```diff
@@ -0,0 +1,215 @@
...
+                 num_processors=None,
+                 sampler_worker_cls=None):
+        self.SamplerWorker = ray.remote(SamplerWorker if sampler_worker_cls is
```
ryanjulian: please use PEP8-style naming

src/garage/sampler/ray_batched_sampler.py
```diff
@@ -0,0 +1,215 @@
...
+                                        None else sampler_worker_cls)
+
+        self.env = env
```
ryanjulian: please make these all _private, unless there's a good reason for them to be public.

src/garage/sampler/ray_batched_sampler.py
```diff
@@ -0,0 +1,215 @@
...
+        self.algo = algo
+        self.max_path_length = self.algo.max_path_length
+        self.should_render = should_render
+        if not ray.is_initialized():
+            ray.init(ignore_reinit_error=True)
+        self.num_workers = num_processors if num_processors \
+            else psutil.cpu_count(logical=False)
+        self.all_workers = defaultdict(None)
```
ryanjulian: perhaps just `self._workers`

src/garage/sampler/ray_batched_sampler.py
```diff
@@ -0,0 +1,215 @@
...
+        self.SamplerWorker = ray.remote(SamplerWorker if sampler_worker_cls is
+                                        None else sampler_worker_cls)
+
+        self.env = env
+        self.algo = algo
+        self.max_path_length = self.algo.max_path_length
+        self.should_render = should_render
+        if not ray.is_initialized():
+            ray.init(ignore_reinit_error=True)
+        self.num_workers = num_processors if num_processors \
```
ryanjulian: please split lines using `()`, as in PEP8

src/garage/tf/samplers/__init__.py
```diff
@@ -3,7 +3,10 @@
     OffPolicyVectorizedSampler)
 from garage.tf.samplers.on_policy_vectorized_sampler import (
     OnPolicyVectorizedSampler)
+from garage.tf.samplers.ray_batched_sampler import (RaySamplerTF,
+                                                    SamplerWorkerTF)
 
 __all__ = [
-    'BatchSampler', 'OffPolicyVectorizedSampler', 'OnPolicyVectorizedSampler'
+    'BatchSampler', 'OffPolicyVectorizedSampler', 'OnPolicyVectorizedSampler',
+    'RaySamplerTF', 'SamplerWorkerTF'
```
ryanjulian: there's no reason to name these `FooTF`, because they are already in the tree `garage.tf`. 

I didn't see a anything TF-specific in these classes. Why not put them in `garage.sampler`?

src/garage/tf/samplers/ray_batched_sampler.py
```diff
@@ -0,0 +1,46 @@
...
+
+    """
+
+    def __init__(self,
+                 worker_id,
+                 env,
+                 agent,
+                 max_path_length,
+                 should_render=False):
+        self.sess = tf.InteractiveSession()
```
ryanjulian: we don't use interactive sessions in garage. please use `tf.get_default_session()` for now.

tests/fixtures/policies/ray_sampler_utils.py
```diff
@@ -0,0 +1,49 @@
+class MockAlgo():
+    def __init__(self, env_spec, policy, max_path_length=16):
+        self.env_spec = env_spec
+        self.policy = policy
+        self.max_path_length = max_path_length
+
+
+class ScriptedPolicy():
```
ryanjulian: can you attempt to use the mock library directly, rather than creating more test fixtures? i think they are overused in our test suite.

you should be able to use mock.side_effect to achieve what you want here.

tests/fixtures/policies/ray_sampler_utils.py
```diff
@@ -0,0 +1,49 @@
+class MockAlgo():
```
ryanjulian: can you try just using `unittest.mock` instead?

ryanjulian: you don't need `()` on your class

tests/garage/sampler/test_ray_sampler.py
```diff
@@ -0,0 +1,49 @@
...
+import ray
+
+from garage.sampler import RaySampler, SamplerWorker
+from garage.tf.envs import TfEnv
+from garage.tf.samplers.on_policy_vectorized_sampler \
+    import OnPolicyVectorizedSampler
+from tests.fixtures.policies.ray_sampler_utils import MockAlgo, ScriptedPolicy
+
+
+class TestSampler():
```
ryanjulian: no need for `()

tests/garage/sampler/test_ray_sampler.py
```diff
@@ -0,0 +1,49 @@
...
+        trajs2 = sampler2.obtain_samples(0, 1)[0]
+
+        assert (trajs1['observations'].shape == np.array(
+            trajs2['observations']).shape == (6, 16))
+        traj2_action_shape = np.array(trajs2['actions']).shape
+        assert (trajs1['actions'].shape == traj2_action_shape == (6, 4))
+        assert (sum(trajs1['rewards']) == sum(trajs2['rewards']) == 1)
+
+
+if __name__ == '__main__':
```
ryanjulian: please remove this. it's unnecessary using pytest.

tests/garage/sampler/test_ray_sampler.py
```diff
@@ -0,0 +1,49 @@
...
+from garage.tf.envs import TfEnv
+from garage.tf.samplers.on_policy_vectorized_sampler \
+    import OnPolicyVectorizedSampler
+from tests.fixtures.policies.ray_sampler_utils import MockAlgo, ScriptedPolicy
+
+
+class TestSampler():
+    @staticmethod
+    def test_ray_batch_sampler():
+        ray.init(
```
ryanjulian: this should be in setup_method or setup_class

tests/garage/sampler/test_ray_sampler.py
```diff
@@ -0,0 +1,49 @@
...
+            ignore_reinit_error=True,
+            object_store_memory=10000000,
+            redis_max_memory=10000000)
+        gym.envs.register(
```
ryanjulian: there is no need for this. just construct and pass the environment directly.

tests/garage/sampler/test_ray_sampler.py
```diff
@@ -0,0 +1,49 @@
...
+        env = TfEnv(gym.make('GridWorldEnv-v0'))
+
+        policy = ScriptedPolicy()
+
+        algo = MockAlgo(env.spec, policy)
+        sampler1 = RaySampler(
+            algo, env, num_processors=1, sampler_worker_cls=SamplerWorker)
+        sampler1.start_worker()
+
+        sampler2 = OnPolicyVectorizedSampler(algo, env)
```
ryanjulian: this is a nice test to verify what you've done is working, but to commit this to master you need to add unit tests which cover your classes individually, instead of just comparing the output of a single sampling process to the current sampler. see tests/garage/tf/models for some good unit tests.

tests/garage/sampler/test_ray_sampler.py
avnishn: will do!

avnishn:
@zhanpenghe one random example that I ran to benchmark: 

on my laptop, trpo_swimmer takes 41 seconds with 4 envs (number of cores on my computer) with the batch sampler, and 46 seconds on my computer with ray sampler. 

I suspect that some of the slowdown in the ray sampler comes from the warning that is thrown by mujoco that says it doesn't know which glfw to use, and the warning that tensorflow throws that says that it hasn't been compiled to run on intel cpu's (I have pip installed tensorflow on my laptop). 

I noticed that when running on one of the linux towers in the lab, this slowdown is not as prevalent. 

I'll run more examples with on different hyper parameters to further test the speed

tests/fixtures/policies/ray_sampler_utils.py
krzentner: I'm not particularly opposed to test fixtures, as long as they're simple.
Having said that, in this case the only method which needs a non-trivial mock is `get_action`, so it shouldn't be too hard to use the Mocking library directly.

krzentner: It's worth mentioning that a specific disadvantage of `MagicMock` (and `PickleableMagicMock`), is that we don't have a way of pickling the mocking information itself (e.g. method return values). In comparison, text fixtures are generally trivially pickleable.

tests/fixtures/policies/ray_sampler_utils.py
krzentner: Wait, it is a dictionary in the code here, right? Did I miss something?

avnishn:
I think we might actually have an issue with the way in which we report the time it takes for garage experiments to run. 

I ran a trpo_swimmer example along side the linux time utility using both the ray batched sampler, and the current batch sampler, and I found these to be the average results when running for 40 training iterations, where 4000 samples are collected in each training experiment:

**Ray Sampler:**
real    1m4.790s
user    0m48.018s
sys     0m13.804s

**Batch Sampler**
real    1m3.566s
user    4m41.187s
sys     0m39.524s

I ran both experiments 5 times, and these numbers didn’t really change
————-
What I’ve gathered is that they actually take the same time in real time, however the ray sampler uses less cpu time (is more efficient) on a single machine.

Issue #87, opened 2018-06-15
Eliminate pickled vs non-pickled mode if possible


ryanjulian:
Having pickled and non-pickled modes makes `garage` behave quite differently under the hood, adding a dimension to the testing matrix. Primitives in non-pickled mode are simple python objects, but the serialization implied by pickled mode actually changes the semantics of `garage` quite a bit, in unexpected ways.

Worse still, currently pickling behavior is actually just a side effect of `n_parallel > 1`, so it's a hidden testing dimension. It should at least be an explicit configuration parameter, or the only mode offered by `run_experiment`.

Ideally, parallel and non-parallel modes should operate using the same code, just a different number of workers. Pickling adds non-zero overhead (just how much is not clear), so it's understandable that some people would want to turn it off.

naeioi:
It is worth noting that when `n_parallel > 1` garage will setup parallel samplers, with each sampler worker having its own copy of the policy, but the policy optimization is still on a single process. Ideally, this semantic should have no impact on performance (not in terms of execution but reward)

ryanjulian:
I think it's okay to skip pickling as an optimization if `n_parallel == 1`, but we can still catch pickling bugs in this configuration by insisting on pickling/unpickling the primitives before using them. This optimization can also be internal to the sampler, rather than implemented at the runner level.

@naeioi is this what you do right now?

zequnyu:
I'm confused why we initialise `singleton_pool` in `LocalRunner`. Could you explain to me what's the difference between `n_parallel` in `run_experiment()` and `max_cpus` in `LocalRunner`?

naeioi:
@zequnyu There is no difference. Both `n_parallel` and `max_cpus` do the same job - Initialize `singleton_pool` so that `BatchSampler` works properly. So this is redundancy and one of them should be deprecated. I will argue to remove `n_parallel` from `run_experiment()` because we are likely to remove `run_experiment()` anyway.

naeioi:
Regarding this issue. @ryanjulian Is there a problematic reproducer? Otherwise, as you have pointed out, some people will want to run in the non-pickled mode for performance.

ryanjulian:
The problem is structural. Some bugs appear when pickling which do not appear if you don't pickle. We don't want to have to create a test for both configurations for every test.

My suggested fix is to pickle/unpickle (and use only the pickled version) everything which would be transmitted to a remote process, even if you are in single-process mode. This could be done once at the beginning of training.

[...]

PR #777, opened 2019-07-10
Unify resume via experiment API

zequnyu:
This PR will address https://github.com/rlworkgroup/garage/issues/87#issuecomment-505220260.
- Unified resume and train usage in experiment_wrapper
- Updated command line resume and code example

### DIFF
--- a/scripts/garage
+++ b/scripts/garage
@@ -1,10 +1,7 @@
 #!/usr/bin/env python3
-import os
-import subprocess
-
 import click
 
-from garage.experiment import to_local_command
+from garage.experiment import LocalRunner, run_experiment
 
 
 @click.group()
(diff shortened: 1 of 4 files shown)


scripts/garage
```diff
@@ -14,27 +11,32 @@ def cli():
 
 @cli.command()
 @click.option(
-    '--snapshot_dir',
-    required=True,
-    help='Directory of the pickle file to resume experiment from.')
+    '--resume_from_dir', required=True, help='Path to resume experiment from.')
 @click.option(
-    '--resume_epoch',
+    '--resume_from_epoch',
```
ryanjulian: this command will appear as `garage resume` so i think `--from_epoch` and `--from_dir` are still good arguments names, e.g.

`garage resume --from_dir=/data/foo/bar --from_epoch=last`

scripts/garage
```diff
@@ -14,27 +11,32 @@ def cli():
...
+        log_dir = resume_from_dir
+
+    run_experiment(
+        run_task,
+        log_dir=log_dir,
+        resume_from_dir=resume_from_dir,
+        resume_from_epoch=resume_from_epoch)
+
+
+def run_task(_snapshot_config, _variant_data, from_dir, from_epoch):
```
ryanjulian: hmm perhaps just define this inside `resume()`, unless it's used elsewhere. if it is used elsewhere, make it private.

scripts/garage
krzentner: Yeah, I agree that this would be much more ergonomic.

ryanjulian: actually, making from_dir positional is probably a better interface:

`garage resume data/foo/bar --from_epoch=last`

naeioi: I second on your last line. handy and pretty.

zequnyu:
This might need force-push. I don't think we can increase coverage of `experiment_wrapper` under current architecture.

ryanjulian:
Please rebase it and let me know when it's ready for force-push (i.e. the CI passed)

zequnyu:
@ryanjulian This is ready.

PR #785, opened 2019-07-11
Fix recurrent policy bug

ahtsan:
Checking whether dones is None by `if not dones` is error-prone, which makes
GaussianLSTMPolicyWithModel and GaussianGRUPolicyWithModel not working.

Also some tests should have higher thresholds in order to make sure the policies
are working properly.

### DIFF
--- a/src/garage/tf/policies/gaussian_gru_policy_with_model.py
+++ b/src/garage/tf/policies/gaussian_gru_policy_with_model.py
@@ -175,6 +175,10 @@ class GaussianGRUPolicyWithModel(StochasticPolicy2):
         """
         Reset the policy.
 
+        If dones is None, it will be by default np.array([True]) which implies
+        the policy will not be "vectorized", i.e. number of parallel
+        environments for training data sampling = 1.
+
         Args:
             dones (numpy.ndarray): Bool that indicates terminal state(s).
 
--- a/src/garage/tf/policies/gaussian_gru_policy_with_model.py
+++ b/src/garage/tf/policies/gaussian_gru_policy_with_model.py
@@ -179,7 +179,7 @@ class GaussianGRUPolicyWithModel(StochasticPolicy2):
             dones (numpy.ndarray): Bool that indicates terminal state(s).
 
         """
-        if not dones:
+        if dones is None:
             dones = np.array([True])
         if self.prev_actions is None or len(dones) != len(self.prev_actions):
             self.prev_actions = np.zeros((len(dones),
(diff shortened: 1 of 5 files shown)


src/garage/tf/policies/gaussian_gru_policy_with_model.py
```diff
@@ -179,7 +179,7 @@ def reset(self, dones=None):
             dones (numpy.ndarray): Bool that indicates terminal state(s).
 
         """
-        if not dones:
+        if dones is None:
```
ryanjulian: hmm why is `dones=None` an acceptable input for the RNN?

krzentner: I think it's more that `reset()` (without any parameters) is how the RNN is how the policy is restarted.
But `np.array` is mutable, so putting that as the default argument would be bad practice.

krzentner: Frustratingly, this only passed before because the `np.array` only had one element, otherwise it would have crashed with this exception:
```
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

ahtsan: Should we make dones an argument without default value? Currently the experiment runner always provides dones as a numpy array.

krzentner: I would prefer to be able to call `reset()` without any arguments. It's fairly obvious what the input should be anyways.

ryanjulian: wow i wish there was a static analysis rule which would catch this.

@ahtsan are you sure that the `rollout` function doesn't call `reset()` without arguments?

ahtsan: There's one hidden assumption though. When dones is None, it currently assigns dones to be np.array([True]), which implies vec_env=1.

krzentner: Perhaps it should be `np.repeat(True, self.vec_env)`? Either way, I don't actually know what the point of this argument is. I can't find any case in the codebase where we don't pass in an array of True values for this argument, but there are many places where we don't pass in anything.

krzentner: Ah, I see, I spoke too soon. We do actually pass this argument in the `OnPolicyVectorizedSampler`. I still think it's fine for it to have a default value, although I understand that this implies that the policy basically won't be "vectorized" then.

ahtsan: I suggest we keep the current interface but state that policy won't be "vectorized" by default if dones is not passed.

PR #786, opened 2019-07-11
Fix gaussian recurrent policies implementation

ahtsan:
When modeling `log_std` as a parameter, instead of creating two separate parameters for
`log_std` and `step_log_std`, they should share the same tf.Variable.

This PR fixed the bug and refactored parameter class to serve the purpose.

Attached the result of running `PPO` with `GaussianLSTMPolicy` (blue) and
`GaussianLSTMPolicyWithModel` (orange) on `InvertedDoublePendulum`.

Screen Shot 2019-07-11 at 11 22 21 AM

### DIFF
--- a/src/garage/tf/core/parameter.py
+++ b/src/garage/tf/core/parameter.py
@@ -3,13 +3,7 @@
 import tensorflow as tf
 
 
-def parameter(input_var,
-              length,
-              batch_dim=None,
-              initializer=tf.zeros_initializer(),
-              dtype=tf.float32,
-              trainable=True,
-              name='parameter'):
+class Parameter:
     """
     Parameter layer.
 
@@ -21,6 +15,8 @@ def parameter(input_var,
     Args:
         input_var (tf.Tensor): Input tf.Tensor.
         length (int): Integer dimension of the variables.
+        param (tf.Tensor): tf.Tensor to be reused. If None, a new tf.Tensor
+            will be created.
         initializer (callable): Initializer of the variables. The function
             should return a tf.Tensor.
         dtype: Data type of the variables (default is tf.float32).
@@ -30,16 +26,36 @@ def parameter(input_var,
     Return:
         A tensor of the broadcasted variables.
     """
-    with tf.variable_scope(name):
-        p = tf.get_variable(
-            'parameter',
-            shape=(length, ),
-            dtype=dtype,
-            initializer=initializer,
-            trainable=trainable)
 
-        if batch_dim is None:
-            batch_dim = tf.shape(input_var)[:-1]
-        broadcast_shape = tf.concat(axis=0, values=[batch_dim, [length]])
-        p_broadcast = tf.broadcast_to(p, shape=broadcast_shape)
-        return p_broadcast
+    def __init__(self,
+                 input_var,
+                 length,
+                 param=None,
+                 batch_dim=None,
+                 initializer=tf.zeros_initializer(),
+                 dtype=tf.float32,
+                 trainable=True,
+                 name='parameter'):
+        with tf.variable_scope(name):
+            if param is None:
+                self._param = tf.get_variable(
+                    'parameter',
+                    shape=(length, ),
+                    dtype=dtype,
+                    initializer=initializer,
+                    trainable=trainable)
+            else:
+                self._param = param
+            if batch_dim is None:
+                batch_dim = tf.shape(input_var)[:-1]
+            broadcast_shape = tf.concat(axis=0, values=[batch_dim, [length]])
+            self._reshaped_param = tf.broadcast_to(
+                self._param, shape=broadcast_shape)
+
+    @property
+    def reshaped_param(self):
+        return self._reshaped_param
+
+    @property
+    def param(self):
+        return self._param
--- a/src/garage/tf/core/parameter.py
+++ b/src/garage/tf/core/parameter.py
@@ -3,10 +3,7 @@
 import tensorflow as tf
 
 
-def parameter(input_var,
-              length,
-              param=None,
-              batch_dim=None,
+def parameter(length,
               initializer=tf.zeros_initializer(),
               dtype=tf.float32,
               trainable=True,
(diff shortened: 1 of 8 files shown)


src/garage/tf/core/parameter.py
```diff
@@ -31,15 +34,14 @@ def parameter(input_var,
...
-        p = tf.get_variable(
-            'parameter',
-            shape=(length, ),
-            dtype=dtype,
-            initializer=initializer,
-            trainable=trainable)
-
+        if param is None:
+            param = tf.get_variable(
+                'parameter',
```
krzentner: Are you sure this shouldn't be `name`? TBH, I'm not totally sure how this class is used.

ahtsan: I should have given more context. Currently we use this parameter function for modeling a "parameter" - a reshaped tf.Variable so that it matches the input placeholder shape. Therefore we always call the tf.Variable `parameter`, similar to that the weights of a MLP is called `kernel`.

However, sometimes the underlying tf.Variable (not the reshaped tf.Variable) is needed to be reused, and should be returned as well. Similarly, the parameter function should have an argument for the incoming tf.Variable if it needs to be reused.

ryanjulian: i'm not sure what's going on here. are you using this as a helper function to broadcast an existing param? why are there now two return values?

ahtsan: I am using to create a new param and also broadcast it. In some cases the same param is needed for another parameter so I return it as well.

ryanjulian: this is a little bit strange. all of the rest of the primitive functions imply resource allocation. suddenly this one may or may not allocate resources and also for some reason has 2 return values.

is it possible to refactor it into the original (which allocates) and a helper (which broadcasts)? are those the two relevant parts here?

if you supply me with the parameter, i don't need to return it back to you. this is python, you already have it. do you also need access to the allocated params?

ahtsan: It actually makes more sense to separate into two different helper functions. Will do.

src/garage/tf/misc/tensor_utils.py
```diff
@@ -13,6 +13,22 @@ def run(*input_vals):
     return run
 
 
+def broadcast(param, input_var, batch_dim=None):
```
ryanjulian: perhaps `broadcast_with_batch`? otherwise it's not clear what this is doing different from regular `tf.broadcast_to`

src/garage/tf/models/gaussian_cnn_model.py
```diff
@@ -270,14 +271,14 @@ def _build(self, state_input, name=None):
...
-                        state_input,
+                    log_std_param = parameter(
                         length=action_dim,
-                        batch_dim=[batch_dim],
                         initializer=tf.constant_initializer(
                             self._init_std_param),
                         trainable=self._learn_std,
                         name='log_std_network')
+                    log_std_network = broadcast(
+                        log_std_param, state_input, batch_dim=[batch_dim])
```
ryanjulian: why would batch_dim be a list?

src/garage/tf/misc/tensor_utils.py
ahtsan: Makes sense

src/garage/tf/models/gaussian_cnn_model.py
ahtsan: The helper function expects batch_dim to be a list since it can be a list (for recurrent policies).

ahtsan: We can make it accept an integer, and turn into a list tho

ahtsan: How about this?

```
def broadcast_with_batch(param, input_var, is_time_series=False):
    if is_time_series:
        batch_dim = tf.shape(input_var)[:-1]
    else:
        batch_dim = tf.shape(input_var)[0]
    broadcast_shape = tf.concat(axis=0, values=[batch_dim, tf.shape(param)])
    return tf.broadcast_to(param, shape=broadcast_shape)
```

I think we only have this two cases.

ahtsan: @ryanjulian Is it better?

src/garage/tf/misc/tensor_utils.py
ryanjulian: i don't think you addressed this?

src/garage/tf/models/gaussian_cnn_model.py
ahtsan: If this looks ok, I will also rename it to `broadcast_with_batch`

ryanjulian: i suggest you add a param `batch_dim` instead. switching function behavior based on a type-specific bool is an anti-pattern.

either we should have two different functions for two different types (time series batches and non time-series batches), or we should parameterize the possible cases and ignore the types (e.g. `batch_dim`).

it would be nice if we could just unify the batch position for all data. is there really a need for having time series reverse their batch position? in any case, fixing that globally is a better candidate for an issue than this PR.

ryanjulian: err i see. you actually implemented my suggestion above, but provided a different one as example.

ahtsan: Should we define `batch_dim` as the first `batch_dim` dimension as the batch dimension? Then it becomes
```python
def broadcast_with_batch(param, input_var, batch_dim=1):
        assert batch_dim > 0
        batch_shape = tf.shape(input_var)[:batch_dim]
    broadcast_shape = tf.concat(axis=0, values=[batch_shape, tf.shape(param)])
    return tf.broadcast_to(param, shape=broadcast_shape)
``` 
For non-recurrent primitives, we have `batch_dim=1` while for recurrent policies, we will have `batch_dim=2`.

ryanjulian: i think batch dims should be a tuple or list of dimensions which are considered part of the batch. for recurrent, you told me there is more than one batch dimension, so you would list those in batch_dims

e.g.
```python
# if dimensions 3 and 4 are considered batch
foo = broadcast_with_batch(bar, bux, batch_dims=(3,4))
```

you might also consider for convenience allowing non-iterable inputs for convenience
```
# dimension 0 is the batch
foo = broadcast_with_batch(bar, bux, batch_dims=0)
```

by the way, your example above uses default `batch_dim=1` -- aren't dimensions zero-indexed? what data uses 1 as the batch dimension?

src/garage/tf/misc/tensor_utils.py
```diff
@@ -13,6 +13,22 @@ def run(*input_vals):
...
+    """
+    Broadcast an existing variable to match input batch dimension.
+
+    Args:
+        param (tf.Tensor): Variable to broadcast.
+        input_var (tf.Tensor): Input batch variable.
+        batch_dim (int): Define the first `batch_dim` dimension as
+            Batch dimension. If None, default to be 1.
+    """
+    assert batch_dim > 0
```
ryanjulian: use an if statement and raise a ValueError. asserts are not for validating user input.

src/garage/tf/core/parameter.py
```diff
@@ -3,43 +3,31 @@
 import tensorflow as tf
 
 
-def parameter(input_var,
-              length,
-              batch_dim=None,
+def parameter(length,
```
ryanjulian: at this point this is just a wrapped call to `tf.Variable`, which is not very useful.

I know we've gone back-and-forth on this. maybe the better solution is to leave `parameter` as-is and just make a special helper function for recurrent policies? it seems that the recurrent case has 2 big differences:

1. you need to broadcast differently
1. you need to reuse the variable

perhaps you can just have `recurrent_parameter()` which handles the broadcasts automatically and returns two outputs. this also helps us encode the proper batch positions for recurrent networks.

ahtsan: so `parameter()` will create a new parameter and broadcast, assuming the 0th dim is batch dim and returned the broadcasted parameter while `recurrent_parameter()` will create a new parameter and broadcast, assuming the 0th and 1st dim are batch dims, and return both the broadcasted parameter and the original variable?

Does it sound good?

ryanjulian: sounds good

ahtsan: I found that eventually we always want to broadcast again for step inputs using the original parameter, in recurrent case. Why not do it both at the same time inside `recurrent_parameter`?

ryanjulian: that sounds good to me.

PR #787, opened 2019-07-11
Update TensorFlow to 1.14.0

gitanshu:
This commit updates a lot of deprecated imports to tf.compat.v1 imports, and some other method updates like (tf.log -> tf.math.log) and more. There are still more deprecations, but merging it as-is is still useful since it also brings python 3.7 support.

### DIFF
--- a/docs/user/implement_algo_advanced.rst
+++ b/docs/user/implement_algo_advanced.rst
@@ -92,7 +92,7 @@ extra diagnostic information as well as supporting recurrent policies):
             'action',
             batch_dims=1,
         )
-        advantage_var = tf.placeholder('advantage')
+        advantage_var = tf.compat.v1.placeholder('advantage')
         dist = self.policy.distribution
         old_dist_info_vars = {
             k: TT.matrix('old_%s' % k)
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,7 @@ repos:
     -   id: pylint
         name: pylint
         entry: pylint
-        args: ['-j 0', '--rcfile=setup.cfg']
+        args: ['-j 2', '--rcfile=setup.cfg']
         stages: [commit, push]
         language: system
         files: \.py$
--- a/docs/user/implement_algo_basic.rst
+++ b/docs/user/implement_algo_basic.rst
@@ -207,7 +207,7 @@ First, we construct symbolic variables for the input data:
         name='actions',
         batch_dims=1
     )
-    returns_var = tf.placeholder(name='returns')
+    returns_var = tf.compat.v1.placeholder(name='returns')
 
 Note that we can transform the policy gradient formula as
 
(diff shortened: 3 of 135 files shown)


docs/user/implement_algo_advanced.rst
```diff
@@ -92,7 +92,7 @@ extra diagnostic information as well as supporting recurrent policies):
             'action',
             batch_dims=1,
         )
-        advantage_var = tf.placeholder('advantage')
+        advantage_var = tf.compat.v1.placeholder('advantage')
```
ryanjulian: does it throw deprecation warnings if you don't use the compat.v1 modules?

gitanshu: Yes

ryanjulian: alright

ryanjulian:
you might need to wait for #774 for this one, because it is exceeding travis max runtime (because it has to rebuild the docker).

ryanjulian:
do the tests pass locally?

gitanshu:
> do the tests pass locally?

Yup

zequnyu:
It seems I had the same issue with #777.

ryanjulian:
#774 should fix it

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: python
 
-python: "3.5"
+python: "3.7"
```
ryanjulian: this doesn't affect the tested python version, because evaluations happen in docker not in travis

zequnyu:
Do we support tensorflow 1.13? It seems pip doesn't have `tensorflow<1.13,>=1.12.0` now.

ryanjulian:
@zequnyu can you post the error message you got which suggests PyPI doesn't have TF1.13 anymore?

The objective of this PR is to upgrade to TF1.14. We only support one version of TF at a time (the API moves too quickly to do anything else).

gitanshu:
@ryanjulian this happens on python 3.7 only, since 3.7 isn't added as supported version on TF Pypi before version 1.14
Merging this PR should fix that.

ryanjulian:
@gitanshu  sorry -- what happens only in Py37? -- i didn't follow Zequn's original question

gitanshu:
Basically if you do `pip install tensorflow<1.13.1` on python 3.7, pypi says `No matching distribution found`. 1.13.1 was the first version to officially support python 3.7 through pypi

ryanjulian:
@gitanshu can you push this over the finish line today? I think #773 will make it a gnarly refactor if we don't.

ryanjulian:
@gitanshu @zequnyu ah. great -- so it sounds like this PR will allow us to support Py3.5-3.7?

gitanshu:
> @gitanshu @zequnyu ah. great -- so it sounds like this PR will allow us to support Py3.5-3.7?

py36-py37 -> yes. For py35, I am seeing a problem in local where the resolved version of atari-py (0.1.15) is unavailable for py35, but somehow seems to work fine on CI with the same version of python 3.5.6

gitanshu:
This PR is ready to go though, as soon as CI starts behaving again

Issue #778, opened 2019-07-10
Remove cloudpickle


zequnyu:
It seems `cloudpickle` is no longer needed in garage and therefore should be removed and replaced by `pickle`.

ryanjulian:
Should be a fast one, right?

zequnyu:
Yes, sorry I'm away and my network is not good. I'll do it ASAP.

zequnyu:
Close via #791.

ryanjulian:
great work!

PR #791, opened 2019-07-13
Remove cloudpickle from experiment API

zequnyu:
#778.
- removed cloudpickle from experiment API
- updated setup and tests correspondingly

### DIFF
--- a/src/garage/experiment/experiment.py
+++ b/src/garage/experiment/experiment.py
@@ -218,7 +218,6 @@ def run_experiment(method_call=None,
                    variant=None,
                    use_tf=False,
                    use_gpu=False,
-                   use_cloudpickle=None,
                    pre_commands=None,
                    **kwargs):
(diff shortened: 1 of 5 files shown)


src/garage/experiment/experiment.py
```diff
@@ -241,20 +240,17 @@ def run_experiment(method_call=None,
...
 
-    if use_cloudpickle is None:
-        for task in (batch_tasks or [method_call]):
-            assert hasattr(task, '__call__')
-            use_cloudpickle = True
-            # ensure variant exists
-            if variant is None:
-                variant = dict()
+    for task in (batch_tasks or [method_call]):
+        assert hasattr(task, '__call__')
```
ryanjulian: please instead raise a ValueError with a useful message

ryanjulian:
@zequnyu is there a reason this is still waiting?

PR #793, opened 2019-07-15
Ray sampler

avnishn:
- Built for parallel or distributed on policy batch sampling.
- Useful for if we want to run experiments with distributed sampling.
- Operates in a data parallel model (each processing element does the
   same operation, but with different inputs).
- Also works for local experiments well, but there is an overhead that
  presents itself when using tensorflow nn policies when sampling, as
  described in the next point.
- The first iteration of sampling is slightly slower than the on_policy_vectorized
   sampler on some examples (trpo_swimmer) because of a startup overhead that
   occurs because the first time we run new tf graphs for collecting
   samples. After the first iteration of the sampler, the
   sampler collects trajectories at the same speed as the on_policy_vectorized sampler.
- In the case of the trpo_swimmer example, there is an extra 10 second
   overhead due to the construction of new tensorflow graphs for
   sampling.
Reference to #31

### Update
I think we might actually have an issue with the way in which we report the time it takes for garage experiments to run.

I ran a **trpo_swimmer example** along side the **linux time utility** using both the **ray batched sampler vs the current batch sampler**, and I found these to be the average results when running for 40 training iterations, where 4000 samples are collected in each training experiment:

**Ray Sampler:**
real 1m4.790s
user 0m48.018s
sys 0m13.804s

**Batch Sampler:**
real 1m3.566s
user 4m41.187s
sys 0m39.524s

I ran both experiments 5 times, and these numbers didn’t really change
————-
What I’ve gathered is that they actually take the same time in real time, however the ray sampler uses less cpu time (is more efficient) on a single machine.


src/garage/tf/samplers/ray_batched_sampler.py
```diff
@@ -0,0 +1,71 @@
...
+    is in a separate process.
+    """
+
+    def __init__(self,
+                 worker_id,
+                 env,
+                 agent,
+                 seed,
+                 max_path_length,
+                 should_render=False):
```
zhanpenghe: I don't think we should render environment in a sampler. We should just leave this to a plotter.

src/garage/sampler/ray_batched_sampler.py
```diff
@@ -0,0 +1,216 @@
...
+        self._sampler_worker = ray.remote(SamplerWorker if sampler_worker_cls
+                                          is None else sampler_worker_cls)
+        self._env = env
+        self._algo = algo
+        self._seed = seed
+        deterministic.set_seed(self._seed)
+        self._max_path_length = self._algo.max_path_length
+        self._should_render = should_render
+        if not ray.is_initialized():
+            ray.init(ignore_reinit_error=True)
```
zhanpenghe: Hmm why ignoring this error? I think this is unnecessary if `not ray.is_initialized()` is true?

src/garage/tf/samplers/ray_batched_sampler.py
```diff
@@ -0,0 +1,71 @@
...
+        self.sess = tf.Session()
+        self.sess.__enter__()
```
zhanpenghe: Is this necessary?

avnishn: Yeah I have to create a new session for all of the new workers, since they sit on separate processes

src/garage/sampler/ray_batched_sampler.py
avnishn: So the argument `ignore_reinit_error=True` is supposed to avoid the sampler crashing if Ray has already been initialized on that machine, but I discovered that it didn’t work for whatever reason(likely a bug), so I wrapped the code to initialize Ray in a function that checks if ray has already been initialized. I’ll go ahead and remove the argument, as you’re right, it shouldn’t serve any function.

src/garage/tf/samplers/ray_batched_sampler.py
avnishn: I can change that, but I think @krzentner was saying that it might be useful to render one of the sampler workers for debugging. KR can you comment on this?

krzentner: Yeah, it's sometimes pretty useful to be able to render the environment in the sampler for debugging purposes. I know the conventional way of watching observing the policy is to use a separate script, but that doesn't always give the same information.

As long as doing this isn't a huge burden, I would prefer we keep it.

src/garage/tf/samplers/ray_batched_sampler.py
zhanpenghe: Oh I am just wondering if we have to call `__enter__` here.

src/garage/sampler/ray_batched_sampler.py
```diff
@@ -0,0 +1,216 @@
...
+        self._num_workers = num_processors if num_processors \
```
ryanjulian: split with (), per PEP8

src/garage/sampler/ray_batched_sampler.py
```diff
@@ -0,0 +1,216 @@
...
+        - sampler_worker_cls: If none, uses the default SamplerWorker
+            class
+
+    """
+
+    def __init__(self,
+                 algo,
+                 env,
+                 seed,
+                 should_render=False,
```
ryanjulian: sampler workers should not render ever.

src/garage/sampler/ray_batched_sampler.py
```diff
@@ -0,0 +1,216 @@
...
+        self._num_workers = num_processors if num_processors \
+            else psutil.cpu_count(logical=False)
+        self._all_workers = defaultdict(None)
+
+    def start_worker(self):
+        """Initialize a new ray worker."""
+        env_pkl_id = ray.put(pickle.dumps(self._env))
+        agent_pkl_id = ray.put(pickle.dumps(self._algo.policy))
+        for worker_id in range(self._num_workers):
+            self._all_workers[worker_id] = (self._sampler_worker.remote(
```
ryanjulian: is there a reason the () here is necessary?

src/garage/sampler/ray_batched_sampler.py
```diff
@@ -0,0 +1,216 @@
...
+        self._should_render = should_render
+        if not ray.is_initialized():
+            ray.init(ignore_reinit_error=True)
+        self._num_workers = num_processors if num_processors \
+            else psutil.cpu_count(logical=False)
+        self._all_workers = defaultdict(None)
+
+    def start_worker(self):
+        """Initialize a new ray worker."""
+        env_pkl_id = ray.put(pickle.dumps(self._env))
```
ryanjulian: doesn't `ray.put()` pickle the argument for you?

src/garage/sampler/ray_batched_sampler.py
```diff
@@ -0,0 +1,216 @@
...
+        traj = []
+        updating_workers = []
+        self._idle_worker_ids = list(range(self._num_workers))
+        curr_policy_params = self._algo.policy.get_param_values()
+        params_id = ray.put(curr_policy_params)
+        while self._idle_worker_ids:
+            worker_id = self._idle_worker_ids.pop()
+            worker = self._all_workers[worker_id]
+            updating_workers.append(worker.set_agent.remote(params_id))
+        while completed_samples < num_samples:
```
ryanjulian: perhaps can you break this logic into some private helper functions? it's a bit hard to follow if someone is just scanning.

splitting up the code into some logical blocks (i.e. with vertical whitespace) and adding comments would also help.

this is the heart of this class. it should be easy to read/understand.

src/garage/sampler/ray_batched_sampler.py
```diff
@@ -0,0 +1,216 @@
...
+                pbar.inc(num_returned_samples)
+                traj.append(trajectory)
+        pbar.stop()
+        return traj
+
+    def shutdown_worker(self):
+        """Shuts down the worker."""
+        ray.shutdown()
+
+    def _process_trajectory(self, result):
```
ryanjulian: can you add a docstring about what this is doing? honestly i'm not sure why it's necessary.

src/garage/tf/samplers/ray_batched_sampler.py
```diff
@@ -0,0 +1,71 @@
...
+            env,
+            seed,
+            should_render=False,
+            num_processors=None,
+            sampler_worker_cls=SamplerWorkerTF)
+
+    def shutdown(self):
+        """Shuts down the worker."""
+        temp = []
+        for worker in self._all_workers.values():
```
ryanjulian: is there a reason this isn't simply super().shutdown()?

src/garage/tf/samplers/ray_batched_sampler.py
```diff
@@ -0,0 +1,71 @@
...
+Currently the same as garage.samplers.RaySampler but includes
+support for Tensorflow sessions
+"""
+import ray
+import tensorflow as tf
+
+from garage.sampler import RaySampler, SamplerWorker
+
+
+class RaySamplerTF(RaySampler):
```
ryanjulian: @krzentner do you think the right pattern for this is inheritance? i guess it will be moot once we transition to TF2.0, and sessions don't really exist.

tests/fixtures/policies/scripted_policy.py
```diff
@@ -0,0 +1,36 @@
+class ScriptedPolicy:
```
ryanjulian: can you add this class to `garage.np.policies` instead? it would be useful outside of tests.

tests/fixtures/policies/scripted_policy.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+
+    def __init__(self, scripted_actions, agent_env_infos=None):
+        self.scripted_actions = scripted_actions
+        self.agent_env_infos = agent_env_infos
+
+    def set_param_values(self, params):
+        return params
+
+    def get_param_values(self):
```
ryanjulian: I don't think this follows the get_param_values contract. shouldn't it be a list or something?

tests/fixtures/policies/scripted_policy.py
```diff
@@ -0,0 +1,36 @@
...
+
+    Args:
+        - scripted actions(list or dictionary): data structure indexed by
+            obervation, returns a corresponding action
+        - agent_env_infos(list or dictionary): data structure indexed by
+            obervation, returns a corresponding agent_env_info
+    """
+
+    def __init__(self, scripted_actions, agent_env_infos=None):
+        self.scripted_actions = scripted_actions
```
ryanjulian: both of these should be private.

tests/garage/sampler/test_ray_sampler.py
```diff
@@ -0,0 +1,82 @@
...
+
+from garage.envs.grid_world_env import GridWorldEnv
+from garage.sampler import RaySampler, SamplerWorker
+from garage.tf.envs import TfEnv
+from garage.tf.samplers.on_policy_vectorized_sampler \
+    import OnPolicyVectorizedSampler
+from tests.fixtures.policies.scripted_policy import ScriptedPolicy
+
+
+class TestSampler():
```
ryanjulian: no need for ()

tests/garage/sampler/test_ray_sampler.py
```diff
@@ -0,0 +1,82 @@
...
+    'S' : starting point
+    'F' or '.': free space
+    'W' or 'x': wall
+    'H' or 'o': hole (terminates episode)
+    'G' : goal
+    [2,2,1,0,3,1,1,1,2,2,1,1,1,2,2,1]
+    """
+
+    def setup_method(self):
+        ray.init(local_mode=True, ignore_reinit_error=True)
```
ryanjulian: is this required if i want to use RaySampler in a regular launcher?

tests/garage/tf/samplers/test_off_policy_vectorized_sampler.py
```diff
@@ -18,6 +19,7 @@
 
 
 class TestOffPolicyVectorizedSampler(TfGraphTestCase):
+    @pytest.mark.flaky
```
ryanjulian: this should be fixed now

src/garage/tf/samplers/ray_batched_sampler.py
krzentner: I think inheritance is fine, in this cases.

src/garage/tf/samplers/ray_batched_sampler.py
avnishn: So the decision here was, I manually enter a tensorflow session globally for all of my worker processes, on each one of their processes, as opposed to using a with statement to enter a tensorflow session. Does it matter if after I reap these processes, I exit all of the tensorflow sessions that I entered. If not, then I can get rid of this.

src/garage/sampler/ray_batched_sampler.py
avnishn: It was so I could pass pep8. I had too many nested loops, and also, it then made sense to have a function that processes trajectories returned from workers, after they are deposited in the object store.

src/garage/sampler/ray_batched_sampler.py
avnishn: I can do that sure thing!

src/garage/sampler/ray_batched_sampler.py
avnishn: I believe it does, I can go ahead and change this, and make sure that it doesn’t effect speed for any reasons.

src/garage/sampler/ray_batched_sampler.py
avnishn: @krzentner said he might need the option to render a worker, if I understood him correctly.

tests/garage/tf/samplers/test_off_policy_vectorized_sampler.py
avnishn: will remove this

src/garage/sampler/ray_batched_sampler.py
krzentner: It's certainly rare to need to render workers, but there's vaguely two reasons it seems convenient to me:
1. If you're debugging some behavior that you only see in the sampler, being able to render some of the environments can be a really useful debugging tool.
2. It's sometimes easier to make one of the workers render than to start a totally separate program to do rendering.

tests/garage/sampler/test_ray_sampler.py
avnishn: Its not required for a regular launch. The reason I do this is because while running tests, there is a cap on the amount available to the tests, and the ray needs at least 10mb extra of memory to run. When I run it like this, I am running it locally, so that it runs without creating its 10mb object store, and also without creating sampling workers on separate processes.

avnishn: The sampler will run serially this way

src/garage/sampler/ray_batched_sampler.py
```diff
@@ -0,0 +1,216 @@
+"""This is an implementation of an on policy batch sampler.
```
krzentner: Please rename this file to `ray_sampler.py` (or rename the class to `RayBatchSampler`) to keep the file and class names consistent.

src/garage/tf/samplers/ray_batched_sampler.py
avnishn: Well we have to enter the session at some point in order to sample from the policy. We can do this with with statements that open and close the policy for us, or since each of the sampler workers won’t have any other sessions running on them, we can enter the session upon initialization, and exit when we’re shutting down the workers.

tests/fixtures/policies/scripted_policy.py
avnishn: added

tests/fixtures/policies/scripted_policy.py
avnishn: fixed

avnishn:
This PR has commit has passed both the travis, and codecov tests:

https://codecov.io/gh/rlworkgroup/garage/pull/793
https://travis-ci.com/rlworkgroup/garage/builds/120906807

can an admin override the CI tests?

ryanjulian:
I would like to wait until we get codecov fixed. Hopefully that will be today.

src/garage/sampler/ray_batched_sampler.py
avnishn: resolved

src/garage/sampler/ray_batched_sampler.py
avnishn: will remove

src/garage/sampler/ray_batched_sampler.py
avnishn: resolved

src/garage/tf/samplers/ray_batched_sampler.py
avnishn: additionally, ray itself has a shutdown routine that needs to be called (line 41)

src/garage/sampler/ray_batched_sampler.py
avnishn: resolved

src/garage/sampler/ray_batched_sampler.py
avnishn: resolved

src/garage/sampler/ray_batched_sampler.py
avnishn: So it does pickle, but it tries to use cloud pickle by default when we don’t manually pickle the policy, and because the policy can’t be cloudpickled, that causes the sampler to fail.

src/garage/sampler/ray_batched_sampler.py
avnishn: resolved.

src/garage/sampler/ray_batched_sampler.py
avnishn: fixed

avnishn:
https://codecov.io/gh/rlworkgroup/garage/pull/793

avnishn:
https://codecov.io/gh/rlworkgroup/garage/pull/793

I’m passing the code coverage test at the moment, can we go ahead and merge this with master?
@krzentner

krzentner:
Yeah, I don't know why codecov is broken here. Dockerfile.intel is broken and about to be deleted. I'll just merge this.

examples/tf/trpo_swimmer_ray_sampler.py
```diff
@@ -0,0 +1,50 @@
...
+from garage.tf.envs import TfEnv
+from garage.tf.policies import GaussianMLPPolicyWithModel
+from garage.tf.samplers import RaySamplerTF
+
+seed = 100
+
+
+def run_task(*_):
+    """."""
+    with LocalRunner() as runner:
```
zequnyu: We need to rebase this to use `LocalTFRunner`

examples/tf/trpo_swimmer_ray_sampler.py
```diff
@@ -0,0 +1,50 @@
...
+
+        algo = TRPO(
+            env_spec=env.spec,
+            policy=policy,
+            baseline=baseline,
+            max_path_length=500,
+            discount=0.99,
+            max_kl_step=0.01)
+
+        # runner.setup(algo, env)
```
zequnyu: This line seems useless.

examples/tf/trpo_swimmer_ray_sampler.py
krzentner: Addressed in #830

examples/tf/trpo_swimmer_ray_sampler.py
krzentner: Addressed in #830

PR #809, opened 2019-07-20
Backport v2018.10.2

krzentner:
These are mostly the same as the v2019.02 backports, although I squashed the first two together to make them clearer.

They probably work. Unfortunately, I'm not actually sure how we run tests on this release. @ryanjulian ?


ryanjulian:
Basically, it's manually running the tests locally. We can create an issue for fixing the CI for releases (which will require backporting some CI stuff), but for now I say we push it if we are reasonably-confident it works.

krzentner:
AFAICT, it's no longer (easily) possible to install the packages needed to run tests for this version locally. Specifically, conda gets stuck on broken dependencies in mkl and tensorflow.
I'd rather not merge these if they're not known to work and no one is known to be using this release. If anyone is using it, then we can look into merging these backports.

ryanjulian:
I will make an effort to find the last stable depenency graph for the branch and lock it there, then apply these changes. According to our [maintenance plan](https://github.com/rlworkgroup/garage/releases/tag/v2018.10.0), June 2019 is EOL for this release, so I'd like to lock the dependencies, apply these backports and move on.

PR #813, opened 2019-07-22
Fix incorrect usage of snapshot_config

zequnyu:


scripts/garage
```diff
@@ -26,7 +26,7 @@ def resume(from_dir, from_epoch, log_dir):
         log_dir = from_dir
 
     def run_task(_snapshot_config, _variant_data, _from_dir, _from_epoch):
```
krzentner: By the way, why do these arguments begin with underscores?

zequnyu: `_variant_data` is not used, `_from_dir`, `_from_epoch` differ from the outer function. Also I'm trying to indicate the `run_task` should not be touched by users.

Issue #736, opened 2019-06-20
Refactor CEM and CMA_ES to use new train method in Algorithm


zequnyu:
To refactor CEM and CMA_ES to conform to new interface introduced by #728, so that they don't depend on itr to determine the stage.

zequnyu:
> CEM and CMA still need refactoring to conform to this new interface, so that they don't depend on itr to determine the stage, and n_samples should be gone after refactoring.

After some attempts I feel our current code is probably ok. `train_once()` and `process_samples()` need `itr`, and I think `train_once()` is a good place to determine the stage.
For `n_epoch_cycles`, it's hard to remove from `LocalRunner`. We can, however, remove it from the algorithms, which will result in more integration between our runner and algos. Is this what you are looking for? @naeioi

ryanjulian:
@zequnyu will this be covered by your new PR to remove n_epoch_cycles?

zequnyu:
Closed by #1093.

PR #815, opened 2019-07-25
Implemented DDPG and task in PyTorch

lywong92:


src/garage/torch/algos/ddpg.py
```diff
@@ -0,0 +1,300 @@
...
+        self.policy_optimizer.zero_grad()
+        action_loss.backward()
+        self.policy_optimizer.step()
+
+        # update target networks
+        self.update_target()
+        return (qval_loss.detach(), y_target, qval.detach(),
+                action_loss.detach())
+
+    def update_target(self):
```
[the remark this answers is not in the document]
zhanpenghe: I think this function should be fine in DDPG but we should move `np_to_torch` and `torch_to_np` to utils.

src/garage/torch/algos/ddpg.py
```diff
@@ -0,0 +1,300 @@
...
+
+        if self.input_include_goal:
+            goals = transitions['goal']
+            next_inputs = np.concatenate((next_observations, goals), axis=-1)
+            inputs = np.concatenate((observations, goals), axis=-1)
+        else:
+            next_inputs = next_observations
+            inputs = observations
+        with torch.no_grad():
+            next_actions = self.target_policy._nn_module(next_inputs)
```
zhanpenghe: Can we not call `._nn_module` and instead just do `self.target_policy(next_inputs)`. (I assume policies are modules?)

src/garage/torch/algos/ddpg.py
```diff
@@ -0,0 +1,300 @@
...
+        self.qf_optimizer = optimizer(
+            self.qf._nn_module.parameters(), lr=self.qf_lr)
+
+    def train_once(self, itr, paths):
+        """Perform one iteration of training."""
+        paths = self.process_samples(itr, paths)
+
+        epoch = itr / self.n_epoch_cycles
+
+        self.episode_rewards.extend([
```
zhanpenghe: I think we should move this to `process_samples`

src/garage/torch/algos/ddpg.py
```diff
@@ -0,0 +1,300 @@
...
+                self.episode_policy_losses = []
+                self.episode_qf_losses = []
+                self.epoch_ys = []
+                self.epoch_qs = []
+
+            self.success_history.clear()
+
+        return last_average_return
+
+    def optimize_policy(self, itr, samples_data):
```
zhanpenghe: I think `samples_data` is not used at all. Should we move the `replay_buffer.sample()` out of this function and use that as samples_data?

examples/torch/ddpg_pendulum.py
```diff
@@ -0,0 +1,92 @@
...
+            hidden_nonlinearity=F.relu,
+            output_nonlinearity=torch.tanh)
+        target_policy = DeterministicPolicy(env.spec, target_policy_module)
+
+        target_qf_module = MLPModule(
+            input_dim=input_dim,
+            output_dim=1,
+            hidden_sizes=[64, 64],
+            hidden_nonlinearity=F.relu)
+        target_qf = ContinuousNNQFunction(env.spec, target_qf_module)
```
zhanpenghe: I think a module can be easily copied (e.g. `copy.deepcopy(module)`) so we don't meed to reinitialize one with hyperparameters. 

Having that said, I also noticed that there are complaints about this is not working somehow. Please verify.

src/garage/torch/algos/ddpg.py
```diff
@@ -0,0 +1,300 @@
...
+import torch
+
+from garage.np.algos.off_policy_rl_algorithm import OffPolicyRLAlgorithm
+
+
+class DDPG(OffPolicyRLAlgorithm):
+    """A DDPG model implemented with PyTorch.
+
+    DDPG, also known as Deep Deterministic Policy Gradient, uses actor-critic
+    method to optimize the policy and reward prediction. It uses a supervised
```
zhanpenghe: q-function instead of reward prediction?

src/garage/torch/algos/ddpg.py
```diff
@@ -0,0 +1,300 @@
...
+        optimizer (torch.optimizer): Optimizer for training policy network
+            and Q-vaule network.
+        criterion (torch.nn.modules.loss): Loss function for Q-value newtork.
+        clip_pos_returns (bool): Whether or not clip positive returns.
+        clip_return (float): Clip return to be in [-clip_return,
+            clip_return].
+        max_action (float): Maximum action magnitude.
+        reward_scale (float): Reward scale.
+        input_include_goal (bool): Whether input includes goal.
+        smooth_return (bool): Whether to smooth the return.
```
zhanpenghe: `Whether to smooth the return for logging.`

This option was introduced just because baseline used this and we want to compare against baseline. We should consider removing this at some point.

src/garage/torch/algos/ddpg.py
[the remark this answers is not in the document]
ryanjulian: hi, "some people" here -- is this really not a feature of torch?

src/garage/torch/algos/ddpg.py
```diff
@@ -0,0 +1,300 @@
...
+        observations = transitions['observation']
+        rewards = transitions['reward']
+        actions = transitions['action']
+        next_observations = transitions['next_observation']
+        terminals = transitions['terminal']
+
+        rewards = rewards.reshape(-1, 1)
+        terminals = terminals.reshape(-1, 1)
+
+        if self.input_include_goal:
```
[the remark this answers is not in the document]
ryanjulian: please remove this and use a wrapper or preprocessor for it instead.

it doesn't really help for us to intentionally create more technical debt around HER.

src/garage/torch/algos/ddpg.py
```diff
@@ -0,0 +1,300 @@
...
+        Returns:
+           Tuple of data in numpy arrays.
+
+        """
+        value_out = []
+        for v in value_in:
+            value_out.append(v.numpy())
+        return tuple(value_out)
+
+    def init_opt(self):
```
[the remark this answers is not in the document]
ryanjulian: i think they should just plain be removed in torch.  get_itr_snapshot is even deprecated (you should just pickle the algorithm).

examples/torch/ddpg_pendulum.py
```diff
@@ -0,0 +1,92 @@
...
+from garage.torch.algos import DDPG
+from garage.torch.modules import MLPModule
+from garage.torch.policies import DeterministicPolicy
+from garage.torch.q_functions import ContinuousNNQFunction
+
+
+def run_task(*_):
+    """Set up environment and algorithm and run the task."""
+    with LocalRunner() as runner:
+        env = TfEnv(normalize(gym.make('InvertedDoublePendulum-v2')))
```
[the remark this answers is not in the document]
ryanjulian: i think there's a missing wrapper which converts the gym spaces to akro spaces. perhaps we should move this into akro?

src/garage/torch/algos/ddpg.py
ryanjulian: we will still be benchmarking against baselines for the forseeable future, but if we want to do this a different way (e.g. with inheritance or a preprocessor or a wrapper), that's fine too

src/garage/torch/algos/ddpg.py
ryanjulian: Q-function

src/garage/torch/algos/ddpg.py
```diff
@@ -0,0 +1,300 @@
...
+            of the policy network.
+        qf_weight_decay (float): L2 weight decay factor for parameters
+            of the q value network.
+        optimizer (torch.optimizer): Optimizer for training policy network
+            and Q-vaule network.
+        criterion (torch.nn.modules.loss): Loss function for Q-value newtork.
+        clip_pos_returns (bool): Whether or not clip positive returns.
+        clip_return (float): Clip return to be in [-clip_return,
+            clip_return].
+        max_action (float): Maximum action magnitude.
```
ryanjulian: there does this come into plan in DDPG?

examples/torch/ddpg_pendulum.py
```diff
@@ -0,0 +1,92 @@
...
+
+        policy_module = MLPModule(
+            input_dim=env.spec.observation_space.flat_dim,
+            output_dim=env.spec.action_space.flat_dim,
+            hidden_sizes=[64, 64],
+            hidden_nonlinearity=F.relu,
+            output_nonlinearity=torch.tanh)
+        policy = DeterministicPolicy(env.spec, policy_module)
+
+        input_dim = (env.spec.observation_space.flat_dim +
```
ryanjulian: this is a real pain and error-prone. i'd prefer if we define `ContinuousMLPQFunction` which infers this directly from the env_spec.

examples/torch/ddpg_pendulum.py
```diff
@@ -0,0 +1,92 @@
...
+
+        action_noise = OUStrategy(env.spec, sigma=0.2)
+
+        policy_module = MLPModule(
+            input_dim=env.spec.observation_space.flat_dim,
+            output_dim=env.spec.action_space.flat_dim,
+            hidden_sizes=[64, 64],
+            hidden_nonlinearity=F.relu,
+            output_nonlinearity=torch.tanh)
+        policy = DeterministicPolicy(env.spec, policy_module)
```
ryanjulian: same case here with inferring from env_spec.

src/garage/torch/algos/ddpg.py
```diff
@@ -0,0 +1,300 @@
...
+        policy_lr (float): Learning rate for training policy network.
+        qf_lr (float): Learning rate for training Q-value network.
+        discount(float): Discount factor for the cumulative return.
+        policy_weight_decay (float): L2 weight decay factor for parameters
+            of the policy network.
+        qf_weight_decay (float): L2 weight decay factor for parameters
+            of the q value network.
+        optimizer (torch.optimizer): Optimizer for training policy network
+            and Q-vaule network.
+        criterion (torch.nn.modules.loss): Loss function for Q-value newtork.
```
ryanjulian: can you call this the Q-function loss instead?

is there an application for this argument? does anybody know of a use for it?

src/garage/torch/algos/ddpg.py
```diff
@@ -0,0 +1,300 @@
...
+        target_policy (garage.torch.policies.base.Policy): Copy of policy.
+        qf (object): Q-value network.
+        target_qf (object): Copy of Q-value network.
+        replay_buffer (garage.replay_buffer.ReplayBuffer): Replay buffer.
+        n_epoch_cycles (int): Number of train_once calls per epoch.
+        n_train_steps (int): Training steps.
+        max_path_length (int): Maximum path length. The episode will
+            terminate when length of trajectory reaches max_path_length.
+        min_buffer_size (int): The minimum buffer size for replay buffer.
+        rollout_batch_size (int): Roll out batch size.
```
ryanjulian: isn't this a parameter of the sampler?

src/garage/torch/algos/ddpg.py
```diff
@@ -0,0 +1,300 @@
...
+        policy (garage.torch.policies.base.Policy): Policy.
+        target_policy (garage.torch.policies.base.Policy): Copy of policy.
+        qf (object): Q-value network.
+        target_qf (object): Copy of Q-value network.
+        replay_buffer (garage.replay_buffer.ReplayBuffer): Replay buffer.
+        n_epoch_cycles (int): Number of train_once calls per epoch.
+        n_train_steps (int): Training steps.
+        max_path_length (int): Maximum path length. The episode will
+            terminate when length of trajectory reaches max_path_length.
+        min_buffer_size (int): The minimum buffer size for replay buffer.
```
ryanjulian: shouldn't this be a parameter of the trainer?

src/garage/torch/algos/ddpg.py
```diff
@@ -0,0 +1,300 @@
...
+    networks involved to stabilize the training process.
+
+    Args:
+        env_spec (EnvSpec): Environment specification.
+        policy (garage.torch.policies.base.Policy): Policy.
+        target_policy (garage.torch.policies.base.Policy): Copy of policy.
+        qf (object): Q-value network.
+        target_qf (object): Copy of Q-value network.
+        replay_buffer (garage.replay_buffer.ReplayBuffer): Replay buffer.
+        n_epoch_cycles (int): Number of train_once calls per epoch.
```
ryanjulian: isn't this a parameter of the runner?

src/garage/torch/algos/ddpg.py
```diff
@@ -0,0 +1,300 @@
...
+    DDPG, also known as Deep Deterministic Policy Gradient, uses actor-critic
+    method to optimize the policy and reward prediction. It uses a supervised
+    method to update the critic network and policy gradient to update the actor
+    network. And there are exploration strategy, replay buffer and target
+    networks involved to stabilize the training process.
+
+    Args:
+        env_spec (EnvSpec): Environment specification.
+        policy (garage.torch.policies.base.Policy): Policy.
+        target_policy (garage.torch.policies.base.Policy): Copy of policy.
```
ryanjulian: i think that this copy should be maintained by the algorithm (derived from policy), rather than expecting the user to provide an exact copy of policy.

src/garage/torch/algos/ddpg.py
```diff
@@ -0,0 +1,300 @@
...
+        qf (object): Q-value network.
+        target_qf (object): Copy of Q-value network.
+        replay_buffer (garage.replay_buffer.ReplayBuffer): Replay buffer.
+        n_epoch_cycles (int): Number of train_once calls per epoch.
+        n_train_steps (int): Training steps.
+        max_path_length (int): Maximum path length. The episode will
```
ryanjulian: runner

src/garage/torch/algos/ddpg.py
```diff
@@ -0,0 +1,300 @@
...
+            of the q value network.
+        optimizer (torch.optimizer): Optimizer for training policy network
+            and Q-vaule network.
+        criterion (torch.nn.modules.loss): Loss function for Q-value newtork.
+        clip_pos_returns (bool): Whether or not clip positive returns.
+        clip_return (float): Clip return to be in [-clip_return,
+            clip_return].
+        max_action (float): Maximum action magnitude.
+        reward_scale (float): Reward scale.
+        input_include_goal (bool): Whether input includes goal.
```
ryanjulian: please move this out and into a wrapper instead.

src/garage/torch/algos/ddpg.py
```diff
@@ -0,0 +1,300 @@
...
+        qf_weight_decay (float): L2 weight decay factor for parameters
+            of the q value network.
+        optimizer (torch.optimizer): Optimizer for training policy network
+            and Q-vaule network.
+        criterion (torch.nn.modules.loss): Loss function for Q-value newtork.
+        clip_pos_returns (bool): Whether or not clip positive returns.
+        clip_return (float): Clip return to be in [-clip_return,
+            clip_return].
+        max_action (float): Maximum action magnitude.
+        reward_scale (float): Reward scale.
```
ryanjulian: what is the purpose of this parameter?

src/garage/torch/algos/ddpg.py
```diff
@@ -0,0 +1,300 @@
...
+
+        Returns:
+           Dictionary of data in PyTorch tensors.
+
+        """
+        for key, value in dict.items():
+            dict[key] = torch.FloatTensor(value)
+        return dict
+
+    def torch_to_np(self, value_in):
```
ryanjulian: i think this should be a feature of akro

src/garage/torch/algos/ddpg.py
```diff
@@ -0,0 +1,300 @@
...
+                 qf_weight_decay=0,
+                 optimizer=torch.optim.Adam,
+                 qf_criterion=torch.nn.MSELoss(),
+                 clip_pos_returns=False,
+                 clip_return=np.inf,
+                 max_action=None,
+                 reward_scale=1.,
+                 input_include_goal=False,
+                 smooth_return=True,
+                 name='DDPG'):
```
ryanjulian: not needed in torch, AFAIK

src/garage/torch/algos/ddpg.py
```diff
@@ -0,0 +1,300 @@
...
+
+        # update target networks
+        self.update_target()
+        return (qval_loss.detach(), y_target, qval.detach(),
+                action_loss.detach())
+
+    def update_target(self):
+        """Update parameters in the target policy and Q-value network."""
+        for t_param, param in zip(self.target_qf._nn_module.parameters(),
+                                  self.qf._nn_module.parameters()):
```
krzentner: The fact that we have to reach into the q function like this reinforces my view that the q function should just be an appropriately shaped neural network. This doesn't need to be addressed in this change, but it would be good to fix soon.

src/garage/torch/algos/ddpg.py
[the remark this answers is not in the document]
krzentner: I don't believe it is a feature of torch. rlkit defines a similar helper function to do this.
If torch did have such a function, it would definitely belong in the `torch.nn.utils` module. However, no function to do this is mentioned [in the documentation](https://pytorch.org/docs/stable/nn.html#utilities).

examples/torch/ddpg_pendulum.py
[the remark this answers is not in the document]
lywong92: That's what I thought, too, but I see the wrappers in GarageEnv now. I can use GarageEnv instead of TfEnv.

src/garage/torch/algos/ddpg.py
lywong92: Will remove.

src/garage/torch/algos/ddpg.py
lywong92: Will remove.

examples/torch/ddpg_pendulum.py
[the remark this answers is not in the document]
zhanpenghe: Sounds good to me.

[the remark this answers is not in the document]
ryanjulian: I think GarageEnv is abstract and therefore can't be instantiated. 

feel free to create TorchEnv and put it in the tree. it's probably the easiest way out of this right now. in the future i think the right place to do this is in the Runner anyway (and possibly moving the classes to akro), but that's out-of-scope here. just work with that you've got.

src/garage/torch/algos/ddpg.py
ryanjulian: hmm i think there's a lot to be said for just making policies modules (is-a) vs policies having modules (has-a). why don't you try it and see what it looks like?

examples/torch/ddpg_pendulum.py
ryanjulian: alternatively you could do `pickle.loads(pickle.dumps(module))`

src/garage/torch/algos/ddpg.py
ryanjulian: it might be an optimization added in baselines, which we keep to we can compare apples-to-apples.

i think it's fine, though it might be more appropriate in the policy module. it depends on how it's applied.

src/garage/torch/algos/ddpg.py
ryanjulian: i think that is a good idea until we have a use case for a param. when we do have a use case, the param is very easy to add. but having many which are not used everywhere is a lot of cognitive load.

src/garage/torch/algos/ddpg.py
ryanjulian: hmm can you confirm with @zequnyu and @naeioi ?

this seems strange to me. i was under the impression we refactored this logic out of the algorithms.

ryanjulian:
first of all, great job on this! it's exciting to see the first torch algorithm.

there's some technical debt in here because you started with the TF version which isn't totally cleaned up. that's not your fault.

this algorithm needs to be benchmarked against the TF and baselines versions anyway (and often that leads to some studies and bug fixing), so while you do that (and write the tests), perhaps we can all work on clearing up some of the lingering design issues?

if we get a very clean version of this algorithm, it will serve as a good template for other torch algorithms. it is also a worthy investment because this will make writing other algorithms in this family (SAC and TD3) pretty easy.

src/garage/torch/algos/ddpg.py
zequnyu: If we want to remove `n_epoch_cycles` from algorithm, there is a way to access it from algorithm.train() via `runner.train_args`, but I feel it's too complicated. We had an issue in #736 however we haven't yet made the change.

src/garage/torch/algos/ddpg.py
krzentner: That sounds out of scope for this PR. We should definitely discuss it though.

src/garage/torch/algos/ddpg.py
zhanpenghe: I think this is not used at all. Could you confirm it in the TF tree?

lywong92: Seems like it's used in obtain_samples in off_policy_vectorized_sample.py:

if self.algo.input_include_goal:
                ...
else:
    self.algo.replay_buffer.add_transitions(
         observation=obses,
         action=actions,
         reward=rewards * self.algo.reward_scale,
         terminal=dones,
         next_observation=next_obses,
         )

zhanpenghe: I see. Let's keep it for now then.

src/garage/torch/policies/deterministic_policy.py
```diff
@@ -6,16 +6,19 @@
 """
 import torch
 
+from garage.torch.modules import MLPModule
+from garage.torch.policies import Policy
 
-class DeterministicPolicy:
+
+class DeterministicPolicy(MLPModule, Policy):
```
zhanpenghe: `DeterministicMLPolicy` since this is always an MLP now?

src/garage/torch/q_functions/continuous_nn_q_function.py
```diff
@@ -2,16 +2,18 @@
 
 import torch
 
+from garage.torch.modules import MLPModule
 
-class ContinuousNNQFunction:
+
+class ContinuousNNQFunction(MLPModule):
```
zhanpenghe: `DeterministicMLPPolicy`?

tests/garage/torch/policies/test_deterministic_policy.py
```diff
@@ -6,88 +6,84 @@
...
             hidden_sizes=hidden_sizes,
             hidden_w_init=nn.init.ones_,
             output_w_init=nn.init.ones_)
 
-        policy = DeterministicPolicy(env_spec, nn_module)
         expected_output = np.full([1, act_dim],
                                   fill_value=obs_dim * np.prod(hidden_sizes),
                                   dtype=np.float32)
-        assert np.array_equal(policy.get_action(obs), expected_output)
+        assert np.array_equal(policy.get_action(obs)[0], expected_output)
```
zhanpenghe: If this is `get_action` instead of `get_actions`, shouldn't the dimension of the output be `(action_dim,)` already?

src/garage/torch/algos/ddpg.py
krzentner: Oh, I think we refactored `n_epoch_cycles` back in, since it's only used by `OffPolicyAlgorithm`?

zequnyu: It's also used in cem and cma_es. it's called `n_samples`.

tests/garage/torch/policies/test_deterministic_policy.py
lywong92: Yes it is. The get_action method returns the action and an empty dictionary, and that's why the [0] is there

zhanpenghe: Oh I see. It makes sense.

src/garage/torch/algos/ddpg.py
zhanpenghe: I am pretty sure these parameters were introduced back into the algorithms in a PR for local runner recently @ryanjulian .

zhanpenghe:
I rebased this against master and help resolve some comments to push this further. 

Thanks to @lywong92 I think this is ready for another look. The policy now has a module and it looks in the same style as the TF tree. @ryanjulian @krzentner 

@lywong92 Could you post your benchmarking here instead of on slack? I think the benchmarking is pretty good but I have minor concerns since it looks consistently better than the TF tree. Otherwise, this looks good to me as long as the benchmarking is fair comparison.

lywong92:
Here's the benchmark result:

HalfCheetah-v2_benchmark
Swimmer-v2_benchmark
Hopper-v2_benchmark
Walker2d-v2_benchmark
InvertedPendulum-v2_benchmark
Reacher-v2_benchmark

ryanjulian:
These are looking pretty good. One concern I have is that baselines consistently performs very poorly. Looking at the graphs, it seems possible that there's some constant scalar difference between baselines and everything else.

Can you verify that the metrics here are calculated exactly the same way in all three codebases?

lywong92:
I used the existing ddpg benchmark script and added Torch ddpg. All 3 of them seem to be using the same scalar parameters. The differences I see are that there are extra params in baseline ddpg called critic_lr_reg being set to 0 and nb_eval_step being set to 100, but I don't think nb_eval_step gets used anyway because eval_env is set to None.

lywong92:
@zhanpenghe I'm using 0.27.0

lywong92:
@ryanjulian @krzentner Does this look okay to you guys?

ryanjulian:
The code all looks fine, but I still want to solve the mystery of the really low performance from openai/baselines. It's plausible, but would be very surprising because many people use that code.

ryanjulian:
Okay, fair. LGTM.

Nisanth is already doing a lot of benchmarking, anyway.

src/garage/torch/q_functions/continuous_mlp_q_function.py
```diff
@@ -2,16 +2,18 @@
 
 import torch
 
+from garage.torch.modules import MLPModule
 
-class ContinuousNNQFunction:
+
+class ContinuousMLPQFunction(MLPModule):
     """
     Implements a module-agnostic Q-value network.
```
zhanpenghe: Hmm btw this is not module agnostic anymore.

PR #816, opened 2019-07-25
Separate LocalRunner and LocalTFRunner

zequnyu:
- separated tf sessions from LocalRunner to LocalTFRunner
- updated docs and tests correspondingly

### DIFF
--- a/src/garage/experiment/local_runner.py
+++ b/src/garage/experiment/local_runner.py
@@ -1,27 +1,14 @@
-"""
-The local runner for tensorflow algorithms.
-
-A runner setup context for algorithms during initialization and
-pipelines data between sampler and algorithm during training.
-"""
 import copy
 import time
 import types
 
 from dowel import logger, tabular
-import tensorflow as tf
 
 from garage.experiment.snapshotter import Snapshotter
 
-# Note: Optional module should be imported ad hoc to break circular dependency.
-
 
 class LocalRunner:
-    """This class implements a local runner for tensorflow algorithms.
-
-    A local runner provides a default tensorflow session using python context.
-    This is useful for those experiment components (e.g. policy) that require a
-    tensorflow session during construction.
+    """Base class of local runner.
 
     Use Runner.setup(algo, env) to setup algorithm and environement for runner
     and Runner.train() to start training.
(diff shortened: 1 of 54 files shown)


src/garage/experiment/local_runner.py
```diff
@@ -1,69 +1,24 @@
...
 class LocalRunner:
-    """This class implements a local runner for tensorflow algorithms.
-
-    A local runner provides a default tensorflow session using python context.
-    This is useful for those experiment components (e.g. policy) that require a
-    tensorflow session during construction.
-
-    Use Runner.setup(algo, env) to setup algorithm and environement for runner
-    and Runner.train() to start training.
+    """Base class of local runner.
```
krzentner: I feel like this docstring should provide more context of what the runner is for.
Specifically, that the runner is the interface which controls the lifecycle of the algorithm, and manages services within that lifecycle (specifically, snapshotting, resuming, and sampling).

ryanjulian: Yes -- please provide a nice long description. this is a major interface.

src/garage/tf/experiment/__init__.py
```diff
@@ -0,0 +1,4 @@
+"""TF Experiment functions."""
```
ryanjulian: No need to abbreviate TensorFlow here. we don't pay for code storage by the character :P

src/garage/tf/experiment/local_tf_runner.py
```diff
@@ -0,0 +1,133 @@
...
+                env=env,
+                policy=policy,
+                baseline=baseline,
+                max_path_length=100,
+                discount=0.99,
+                max_kl_step=0.01)
+            runner.setup(algo, env)
+            runner.train(n_epochs=100, batch_size=4000)
+
+        // to resume immediately.
```
ryanjulian: // cannot be used for comments in python...

ryanjulian:
I'm not sure why the Docker cache is not working in the CI (which is why your build didn't pass). I restarted the job -- ping me if that doesn't fix it. You may also need to rebase on master for the cache to work properly.

krzentner:
I'm merging this, since we're about to delete the Dockerfile.intel image anyways, and the PyTorch folks would benefit from this change.

PR #818, opened 2019-07-25
Don't overwrite mjkey.txt if it exists

krzentner:
This is primarily useful for re-running the setup script, since it requires mjkey be specified, and fails if you give the mjkey which is already installed.


krzentner:
> does this make the setup scripts idempotent? that's a nice wish list item.

I believe as long as `--modify-bashrc` isn't passed, the scripts are idempotent. At least I ran `scripts/setup_linux.sh` multiple times without any problems.

PR #819, opened 2019-07-25
Fix test_experiment.py when run in parallel

krzentner:
Two different tests in this file use the default test log directory. As far as I can tell, they are the only tests to do so.

By moving the one not testing for the default directory to a different directory, the test can safely be run in parallel.


zequnyu:
Maybe add some comments about this parallel issue, or I think we can just delete this test. `test_experiment_with_variant()` isn't really important. I added it only to pass the 90% diff coverage.

Issue #790, opened 2019-07-12
VectorizedSamplers should have a higher default n_envs


ryanjulian:
Currently, the default n_envs for VectorizedSamplers is 1. I think it should be at least `max_cpus` as passed to LocalRunner. With VectorizedSampler, CPU load does not appear to scale linearly with number of envs, so this could perhaps be even higher (until we reach marginal returns).

zequnyu:
Emmm would you tell more about how can we find the number of envs for marginal returns?

zequnyu:
@ryanjulian  would you tell more about how can we find the number of envs for marginal returns?

ryanjulian:
@zequnyu I'm not sure I understand how marginal returns are related. Can you explain?

zequnyu:
> With VectorizedSampler, CPU load does not appear to scale linearly with number of envs, so this could perhaps be even higher (until we reach marginal returns).

I'm not sure how to find the maximum number of `n_envs` so that it maximize CPU load. Did I understand your comment correct?

ryanjulian:
Oh, sorry. By "marginal returns" i actually meant "diminishing marginal returns" -- i.e. until it stops making it faster to add environments.

how to determine the ratio? basically, just run some tests. you want to find a multiplier `c` such that when you set `n_envs = c * max_cpus` maxes out the machine (i.e. load average = 1.3-1.5x the number of logical CPUs on a hyperthreaded machine).

essentially, we want to find `c` such that `n_envs = c` consumes ~1 CPU.

zequnyu:
Closed by #825.

PR #825, opened 2019-07-30
Set VectorizedSamplers default n_envs

zequnyu:
Address #790.
When cpu is 1, tested 3 times for `n_envs` in [1...8] using:
```
with LocalRunner(sess=self.sess) as runner:
    env = TfEnv(gym.make('CartPole-v0'))

    policy = CategoricalMLPPolicy(
        env_spec=env.spec, hidden_sizes=[32, 32])

    algo = REPS(
        env_spec=env.spec,
        policy=policy,
        baseline=baseline,
        max_path_length=100,
        discount=0.99)

    runner.setup(algo, env)

    last_avg_ret = runner.train(n_epochs=30, batch_size=4000)
    assert last_avg_ret > 5
```
Screen Shot 2019-07-29 at 12 02 21

Set default `n_envs` to 4 * `max_cpus`.

### DIFF
--- a/examples/resume_training.py
+++ b/examples/resume_training.py
@@ -1,12 +1,11 @@
 #!/usr/bin/env python3
-"""
-This is an example to resume training programmatically.
-"""
+"""This is an example to resume training programmatically."""
 from garage.experiment import run_experiment
 from garage.tf.experiment import LocalTFRunner
 
 
 def run_task(snapshot_config, *_):
+    """Run task."""
     with LocalTFRunner(snapshot_config=snapshot_config) as runner:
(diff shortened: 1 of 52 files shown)


ryanjulian:
Nice work! What happens if you set n_cpu > 1? 

Does the scaling still hold? What is the curve for this experiment for n = 1, 2, ...`echo  $(nproc)`?

ryanjulian:
derp, i just realized you don't need the CPUs dimension.

how many CPUs does your machine have?

zequnyu:
Tests for up to 72 CPUs: https://pastebin.com/raw/d9UqgsEx. I think 4 is good as default value.

PR #826, opened 2019-07-30
Update .codecov.yml

ryanjulian:


krzentner:
What does this change do exactly? It's not really self-describing, and the commit message isn't clear either.

PR #827, opened 2019-07-30
Remove intel package and Dockerfile

gitanshu:
This commit removes the intel-tensorflow dep and the related
dockerfiles since TF1.14 provides the optimizations by
default


krzentner:
I believe we might need to delete (this image on dockerhub)[https://hub.docker.com/r/rlworkgroup/garage-intel] to prevent all of our PRs from failing after merging this.

@ryanjulian I believe you are the only one with permission to do so? Although I have a Docker Hub account linked with my Github account, I have no access to the Docker Hub organization.

PR #828, opened 2019-08-01
Make dm_control an extra dependency

krzentner:
Because dm_control needs to be installed from vcs or tarball, garage
cannot be directly installed in pipenv (see [this
issue](https://github.com/pypa/pipenv/issues/3396)).

I've copied the example that used our dm_control wrapper, modified
the original example to use gym, and modified the copy to clarify the
extra dependency.

I've also modified the `garage.env.dm_control` module to report a
helpful error message if dm_control isn't found.

`pip install garage[all]` will still install dm_control.

### DIFF
--- a/examples/step_dm_control_env.py
+++ b/examples/step_dm_control_env.py
@@ -1,9 +1,12 @@
 #!/usr/bin/env python3
-"""Example of how to load, step, and visualize an environment."""
-import gym
+"""Example of how to load, step, and visualize an environment.
+
+This example requires that garage[dm_control] be installed.
+"""
+from garage.envs.dm_control import DmControlEnv
 
 # Construct the environment
-env = gym.make('MountainCar-v0')
+env = DmControlEnv.from_suite('walker', 'run')
 
 # Reset the environment and launch the viewer
 env.reset()
--- a/setup.py
+++ b/setup.py
@@ -20,11 +20,6 @@ required = [
     'click',
     'cloudpickle',
     'cma==1.1.06',
-    # dm_control throws an error during install about not being able to
-    # find a build dependency (absl-py). Later pip executes the `install`
-    # command again and the install succeeds because absl-py has been
-    # installed. This is stupid, but harmless.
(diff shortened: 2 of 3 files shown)


ryanjulian:
It can be installed with pipenv, but requires the `--pre` flag. pipenv refuses to match dependencies <1.0 by default.

If this is the only package which requires the flag, then I'm happy to move it into an optional. If the mainline packages require `--pre` anyway, then I want to avoid running in circles around the issue.

krzentner:
Ah, I see. This is indeed the only package which requires `--pre`. As far as I can tell, versions before `1.0` are still considered release versions, but versions like e.g. `2.0rc1` aren't.

gitanshu:
I don't need to use --pre for anything with pipenv if I install pipenv from master. OTOH, if I use the last release of pipenv (2018.11.26), even --pre doesn't work for dm_control.

krzentner:
> I don't need to use --pre for anything with pipenv if I install pipenv from master. OTOH, if I use the last release of pipenv (2018.11.26), even --pre doesn't work for dm_control.

Ah, thanks for testing this. I just know that I couldn't get it to work with the latest release. If a pipenv release fixes this soon, then I'd be happy to drop this change.

ryanjulian:
I think the best move is probably to advise people to use pre-release pipenv. I don't see anything on their GitHub indicating a release is imminent, but working around this is kind of a PITA for a normal user.

ref: https://github.com/pypa/pipenv/issues/2596

gitanshu:
I opened a PR for that https://github.com/rlworkgroup/garage/pull/839

krzentner:
After experimenting with pre-release pipenv for a bit, I don't think advising people to use it is a good idea. It (and prerelease pip) break sometimes, and are incredibly difficult to fix afterwards.

ryanjulian:
Okay, let's roll forward with this fix then.

krzentner:
Okay, I'll merge this then. It's worth mentioning that all of garage's functionality can still be used in pipenv by directly installing the required `dm_control` version. This works since `pipenv` (at least since version 2018.11) supports installing packages from url, it just doesn't support dependencies of those packages being url / version control.

Issue #829, opened 2019-08-05
Add default sampler reference to algorithms


zequnyu:
Currently our `LocalRunner` decides sampler class for algorithms.
https://github.com/rlworkgroup/garage/blob/f4a6271edd0f9c280c306d1f0bbf4bc1591ab85e/src/garage/experiment/local_tf_runner.py#L128-L139
The plan is to add reference of default sampler to algorithms so that we can remove the above code from `LocalRunner`.

zequnyu:
Close via #832.

PR #832, opened 2019-08-06
Add default sampler class to algorithms 

zequnyu:
#829.

### DIFF
--- a/src/garage/experiment/local_runner.py
+++ b/src/garage/experiment/local_runner.py
@@ -87,20 +87,8 @@ class LocalRunner:
 
         if sampler_args is None:
             sampler_args = {}
-
         if sampler_cls is None:
-            from garage.tf.algos.batch_polopt import BatchPolopt
-            if isinstance(algo, BatchPolopt):
-                if self.policy.vectorized:
-                    from garage.tf.samplers import OnPolicyVectorizedSampler
-                    sampler_cls = OnPolicyVectorizedSampler
-                else:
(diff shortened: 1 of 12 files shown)


krzentner:
Please also note in the `RLAlgorithm` (in `src/garage/np/algos/base.py`) docstring that this field is required if `sampler_cls` isn't specified to the runner.

Unfortunately, abstract properties are a huge pain to use in Python, since they actually require using `@property`, so we probably shouldn't use one.

krzentner:
Also, yeah. The fact that our generic `np` algorithm classes import TensorFlow samplers indicates to me that we need to do more to separate the TensorFlow setup logic from the samplers. That's out of scope for this change though.

zequnyu:
Yes, after looking at our tf-samplers, I think we could move them all to general `samplers/`, except the recently added Ray sampler.

zequnyu:
Actually, current implementation of policies use a lot of `@property`, would you explain why we shouldn't use it? @krzentner

krzentner:
We do use `@property` a lot, and that's fine. The issue I have with `@abc.abstractproperty` is that it requires you to use `@property` in the implementation class, even if you don't want to.

On the topic of the samplers being TF specific, as far as I can tell the only reason they're TF specific is that they set up a `tf.Session`. However, the logic for how to correctly set up global state like that is also owned by the runner. I feel like we could solve all of this if the runner gave the sampler a "set up global environment" callback to invoke in its workers. That's probably out of scope for this PR, but should be pretty easy to implement.

PR #833, opened 2019-08-07
Implemented VPG in PyTorch

yonghyuc:
Implemented vanilla policy gradient in PyTorch

### DIFF
--- a/src/garage/torch/algos/vpg.py
+++ b/src/garage/torch/algos/vpg.py
@@ -1,0 +1,225 @@
+"""Vanilla Policy Gradient (REINFORCE)."""
+import collections
+
+from dowel import tabular
+import numpy as np
+import torch
+import torch.nn.functional as F
+
+from garage.misc import special
+from garage.np.algos import BatchPolopt
+from garage.torch.algos import loss_function_utils as loss_utils
+
+
+class VPG(BatchPolopt):
+    """Vanilla Policy Gradient (REINFORCE).
+
+    Args:
+        env_spec (garage.envs.EnvSpec): Environment specification.
+        policy (garage.torch.policies.base.Policy): Policy.
+        baseline : The baseline.
+        max_path_length (int): Maximum length of a single rollout.
+        policy_lr (float): Learning rate for training policy network.
+        n_samples (int): Number of train_once calls per epoch.
+        discount (float): Discount.
+        gae_lambda (float): Lambda used for generalized advantage
+            estimation.
+        center_adv (bool): Whether to rescale the advantages
+            so that they have mean 0 and standard deviation 1.
+        positive_adv (bool): Whether to shift the advantages
+            so that they are always positive. When used in
+            conjunction with center_adv the advantages will be
+            standardized before shifting.
+        optimizer (object): The optimizer of the algorithm. Should be the
+            optimizers in torch.optim.
+        policy_ent_coeff (float): The coefficient of the policy entropy.
+            Setting it to zero would mean no entropy regularization.
+        use_softplus_entropy (bool): Whether to estimate the softmax
+            distribution of the entropy to prevent the entropy from being
+            negative.
+        stop_entropy_gradient (bool): Whether to stop the entropy gradient.
+        entropy_method (str): A string from: 'max', 'regularized',
+            'no_entropy'. The type of entropy method to use. 'max' adds the
+            dense entropy to the reward for each time step. 'regularized' adds
+            the mean entropy to the surrogate objective. See
+            https://arxiv.org/abs/1805.00909 for more details.
+    """
+
+    def __init__(
+            self,
+            env_spec,
+            policy,
+            baseline,
+            max_path_length=500,
+            policy_lr=1e-2,
+            n_samples=1,
+            discount=0.99,
+            gae_lambda=1,
+            center_adv=True,
+            positive_adv=False,
+            optimizer=None,
+            policy_ent_coeff=0.0,
+            use_softplus_entropy=False,
+            stop_entropy_gradient=False,
+            entropy_method='no_entropy',
+    ):
+        self.env_spec = env_spec
+        self.policy_lr = policy_lr
+        self.gae_lambda = gae_lambda
+        self.center_adv = center_adv
+        self.positive_adv = positive_adv
+        self.policy_ent_coeff = policy_ent_coeff
+        self.use_softplus_entropy = use_softplus_entropy
+        self.stop_entropy_gradient = stop_entropy_gradient
+        self.entropy_method = entropy_method
+        self.eps = 1e-8
+
+        self.maximum_entropy = (entropy_method == 'max')
+        self.entropy_regularzied = (entropy_method == 'regularized')
+        self._check_entropy_configuration(entropy_method, center_adv,
+                                          stop_entropy_gradient,
+                                          policy_ent_coeff)
+        self._episode_reward_mean = collections.deque(maxlen=100)
+        self.optimizer = optimizer(policy.parameters(), lr=policy_lr)
+
+        super().__init__(policy=policy,
+                         baseline=baseline,
+                         discount=discount,
+                         max_path_length=max_path_length,
+                         n_samples=n_samples)
+
+    def _check_entropy_configuration(self, entropy_method, center_adv,
+                                     stop_entropy_gradient, policy_ent_coeff):
+        if entropy_method not in ('max', 'regularized', 'no_entropy'):
+            raise ValueError('Invalid entropy_method')
+
+        if entropy_method == 'max':
+            if center_adv:
+                raise ValueError('center_adv should be False when '
+                                 'entropy_method is max')
+            if not stop_entropy_gradient:
+                raise ValueError('stop_gradient should be True when '
+                                 'entropy_method is max')
+        if entropy_method == 'no_entropy':
+            if policy_ent_coeff != 0.0:
+                raise ValueError('policy_ent_coeff should be zero '
+                                 'when there is no entropy method')
+
+    def train_once(self, itr, paths):
+        """Perform one step of policy optimization."""
+        samples_data = self.process_samples(itr, paths)
+
+        self._optimize_policy(itr, paths)
+        self.baseline.fit(paths)
+
+        return samples_data['average_return']
+
+    def _optimize_policy(self, itr, paths):
+        self.optimizer.zero_grad()
+
+        policy_entropies = torch.stack([
+            self._add_padding(
+                self.policy.get_entropy(torch.Tensor(path['observations'])),
+                self.max_path_length) for path in paths
+        ])
+
+        valids = [len(path['actions']) for path in paths]
+
+        baselines = torch.stack([
+            self._add_padding(self._get_baselines(path), self.max_path_length)
+            for path in paths
+        ])
+
+        rewards = torch.stack([
+            self._add_padding(torch.Tensor(path['rewards']),
+                              self.max_path_length) for path in paths
+        ])
+
+        if self.maximum_entropy:
+            rewards += self.policy_ent_coeff * policy_entropies
+
+        advantages = loss_utils.compute_advantages(self.discount,
+                                                   self.gae_lambda,
+                                                   self.max_path_length,
+                                                   baselines, rewards)
+
+        if self.center_adv:
+            means, vars = list(
+                zip(*[(valid_adv.mean(), valid_adv.var())
+                      for valid_adv in self._filter_valids(advantages, valids)]
+                    ))
+            advantages = F.batch_norm(advantages.t(),
+                                      torch.Tensor(means),
+                                      torch.Tensor(vars),
+                                      eps=self.eps).t()
+
+        if self.positive_adv:
+            advantages -= advantages.min()
+
+        log_likelihoods = torch.stack([
+            self._add_padding(
+                self.policy.log_likelihood(torch.Tensor(path['observations']),
+                                           torch.Tensor(path['actions'])),
+                self.max_path_length) for path in paths
+        ])
+
+        outputs = log_likelihoods * advantages
+
+        if self.entropy_regularzied:
+            outputs += self.policy_ent_coeff * policy_entropies
+
+        loss = torch.cat(self._filter_valids(outputs, valids))
+        # using a negative because optimizers use gradient descent,  whilst we
+        # want gradient ascent.
+        (-loss).mean().backward()
+
+        self.optimizer.step()
+
+    def _add_padding(self, tensor, length):
+        padding_length = max(length - tensor.shape[-1], 0)
+        return F.pad(tensor, (0, padding_length))
+
+    def _get_policy_entropy(self, obs):
+        policy_entropy = self.policy.get_entropy(obs).sum()
+
+        if self.stop_entropy_gradient:
+            policy_entropy.requires_grad = False
+
+        # This prevents entropy from becoming negative for small policy std
+        if self.use_softplus_entropy:
+            policy_entropy = F.softplus(policy_entropy)
+
+        return policy_entropy
+
+    def _get_baselines(self, path):
+        if hasattr(self.baseline, 'predict_n'):
+            return torch.Tensor(self.baseline.predict_n(path))
+        else:
+            return torch.Tensor(self.baseline.predict(path))
+
+    def _filter_valids(self, tensor, valids):
+        return [tensor[i][:valids[i]] for i in range(len(valids))]
+
+    def process_samples(self, itr, paths):
+        """Process sample data based on the collected paths."""
+        for path in paths:
+            path['returns'] = special.discount_cumsum(path['rewards'],
+                                                      self.discount)
+        average_discounted_return = (np.mean(
+            [path['returns'][0] for path in paths]))
+        undiscounted_returns = [sum(path['rewards']) for path in paths]
+        average_return = np.mean(undiscounted_returns)
+
+        self._episode_reward_mean.extend(undiscounted_returns)
+
+        tabular.record('Iteration', itr)
+        tabular.record('AverageDiscountedReturn', average_discounted_return)
+        tabular.record('AverageReturn', average_return)
+        tabular.record('Extras/EpisodeRewardMean',
+                       np.mean(self._episode_reward_mean))
+        tabular.record('NumTrajs', len(paths))
+        tabular.record('StdReturn', np.std(undiscounted_returns))
+        tabular.record('MaxReturn', np.max(undiscounted_returns))
+        tabular.record('MinReturn', np.min(undiscounted_returns))
+
+        return dict(average_return=average_return)
--- a/src/garage/torch/policies/base.py
+++ b/src/garage/torch/policies/base.py
@@ -8,7 +8,6 @@ class Policy(abc.ABC):
 
     Args:
         env_spec (garage.envs.env_spec.EnvSpec): Environment specification.
-
     """
 
     def __init__(self, env_spec):
@@ -33,3 +32,8 @@ class Policy(abc.ABC):
     def action_space(self):
         """Policy action space."""
         return self._env_spec.action_space
+
+    @property
+    def vectorized(self):
+        """Vectorized or not."""
+        return False
--- a/src/garage/torch/policies/gaussian_mlp_policy.py
+++ b/src/garage/torch/policies/gaussian_mlp_policy.py
@@ -1,12 +1,11 @@
 """GaussianMLPPolicy."""
-import numpy as np
 import torch
 
 from garage.torch.modules import GaussianMLPModule
 from garage.torch.policies import Policy
 
 
-class GaussianMLPPolicy(GaussianMLPModule, Policy):
+class GaussianMLPPolicy(Policy, GaussianMLPModule):
     """
     GaussianMLPPolicy.
 
@@ -18,14 +17,13 @@ class GaussianMLPPolicy(GaussianMLPModule, Policy):
         module : GaussianMLPModule to make prediction based on a gaussian
         distribution.
     :return:
-
     """
 
     def __init__(self, env_spec, **kwargs):
-        self._env_spec = env_spec
         self._obs_dim = env_spec.observation_space.flat_dim
         self._action_dim = env_spec.action_space.flat_dim
 
+        Policy.__init__(self, env_spec)
         GaussianMLPModule.__init__(self,
                                    input_dim=self._obs_dim,
                                    output_dim=self._action_dim,
@@ -33,21 +31,40 @@ class GaussianMLPPolicy(GaussianMLPModule, Policy):
 
     def forward(self, inputs):
         """Forward method."""
-        return super().forward(inputs)
+        return super().forward(torch.Tensor(inputs))
 
     def get_action(self, observation):
         """Get a single action given an observation."""
         with torch.no_grad():
             observation = observation.unsqueeze(0)
             dist = self.forward(observation)
-            std = dist.variance**0.5
-            return dist.rsample().squeeze(0).numpy(), dict(
-                mean=dist.mean.squeeze(0), log_std=np.log(std).squeeze(0))
+            return (dist.rsample().squeeze(0).numpy(),
+                    dict(mean=dist.mean.squeeze(0).numpy(),
+                         log_std=(dist.variance**.5).log().squeeze(0).numpy()))
 
     def get_actions(self, observations):
         """Get actions given observations."""
         with torch.no_grad():
             dist = self.forward(observations)
-            std = dist.variance**0.5
-            return dist.rsample().detach().numpy(), dict(mean=dist.mean,
-                                                         log_std=np.log(std))
+            return (dist.rsample().numpy(),
+                    dict(mean=dist.mean.numpy(),
+                         log_std=(dist.variance**.5).log().numpy()))
+
+    def log_likelihood(self, observation, action):
+        """Get log likelihood given observations and action."""
+        dist = self.forward(observation)
+        return dist.log_prob(action)
+
+    def get_entropy(self, observation):
+        """Get entropy given observations."""
+        dist = self.forward(observation)
+        return dist.entropy()
+
+    def reset(self, dones=None):
+        """Reset the environment."""
+        pass
+
+    @property
+    def vectorized(self):
+        """Vectorized or not."""
+        return True
--- a/tests/benchmarks/test_benchmark_vpg.py
+++ b/tests/benchmarks/test_benchmark_vpg.py
@@ -1,0 +1,257 @@
+"""
+This script creates a regression test over
+garage-PyTorch-VPG, garage-TF-VPG and baselines-VPG.
+It get Mujoco1M benchmarks from baselines benchmark, and test each task in
+its trial times on garage model and baselines model. For each task, there will
+be `trial` times with different random seeds. For each trial, there will be two
+log directories corresponding to baselines and garage. And there will be a plot
+plotting the average return curve from baselines and garage.
+"""
+
+import datetime
+import json
+import os.path as osp
+import random
+
+from baselines.bench import benchmarks
+import dowel
+import dowel.logger as dowel_logger
+import gym
+import matplotlib.pyplot as plt
+import pandas as pd
+import pytest
+import tensorflow as tf
+import torch
+
+from garage.envs import normalize
+from garage.experiment import deterministic, LocalRunner
+from garage.np.baselines import LinearFeatureBaseline
+from garage.tf.algos import VPG as TF_VPG
+from garage.tf.envs import TfEnv
+from garage.tf.experiment import LocalTFRunner
+from garage.tf.policies import GaussianMLPPolicy as TF_GMP
+from garage.torch.algos import VPG as PyTorch_VPG
+from garage.torch.policies import GaussianMLPPolicy as PyTorch_GMP
+from tests import helpers as Rh
+from tests.fixtures import snapshot_config
+
+hyper_parameters = {
+    'hidden_sizes': [64, 64],
+    'center_adv': True,
+    'learning_rate': 1e-2,
+    'discount': 0.99,
+    'n_epochs': 250,
+    'max_path_length': 500
+}
+
+
+def run_garage_pytorch(env, seed, log_dir):
+    """Create garage PyTorch VPG model and training."""
+    env = TfEnv(normalize(env))
+
+    deterministic.set_seed(seed)
+
+    runner = LocalRunner(snapshot_config)
+
+    policy = PyTorch_GMP(
+        env.spec,
+        hidden_sizes=hyper_parameters['hidden_sizes'],
+        hidden_nonlinearity=torch.tanh,
+        # init_std=1e-6,
+        output_nonlinearity=None)
+
+    baseline = LinearFeatureBaseline(env_spec=env.spec)
+
+    algo = PyTorch_VPG(env_spec=env.spec,
+                       policy=policy,
+                       optimizer=torch.optim.Adam,
+                       baseline=baseline,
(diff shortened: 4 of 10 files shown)


yonghyuc:
Currently, I am benchmarking the PyTorch VPG and TensorFlow VPG.

The problem is that with "InvertedDoublePendulum-v2" environment, 
when I trained TensorFlow VPG, I got reward 300 ~ 400 on average,
but when I trained PyTorch VPG, I got reward 150 ~ 170 on average.

I checked TensorFlow VPG code and most part looks similar. 
(using advantage function, Adam optimizer, ...)

How could I improve the PyTorch VPG?

yonghyuc:
[TF_VPG_LOG.txt](https://github.com/rlworkgroup/garage/files/3478975/TF_VPG_LOG.txt)
[TORCH_VPG_LOG.txt](https://github.com/rlworkgroup/garage/files/3478976/TORCH_VPG_LOG.txt)

I attached log files from TF and PyTorch VPG.

src/garage/torch/algos/vpg.py
```diff
@@ -0,0 +1,286 @@
...
+import torch
+import torch.nn.functional as F  # noqa
+
+from garage.misc import special
+from garage.np.algos import RLAlgorithm
+from garage.tf.misc import tensor_utils
+
+
+class VPG(RLAlgorithm):
+    """Vanilla Policy Gradient.
```
krzentner: It would be good to mention that this algorithm is also known as REINFORCE.

krzentner:
Oh, I just saw the benchmarking comment. That's a pretty big difference, so we definitely need to figure out what's causing it. Unfortunately, I don't have a good guess for where to start. I would recommend you also compare to Catalyst's implementation, which should give you some sense of if there's some mistake in this version or if there's some trick that the TF branch has.

avnishn:
@yonghyuc what happens when you visualize the learned policies? Does the agent complete the task when using the torch policy?

src/garage/torch/algos/vpg.py
```diff
@@ -0,0 +1,286 @@
...
+                eps=self.eps)
+            advantages = advantages.squeeze()
+
+        if self.positive_adv:
+            advantages -= advantages.min()
+
+        log_likelihood = self.policy.get_log_likelihood(
+            torch.Tensor(path['observations']), torch.Tensor(path['actions']))
+
+        grad = (log_likelihood * advantages).sum()
```
zhanpenghe: Hmm this does not looks very correct to me. I think what you want is actually take the mean across the whole batch so this should not be a sum. Having that said, I think you should firstly flatten the batch dimension before calculating the loss instead of just pass each path to calculate the loss. Though, they are very similar, but not 100% the same in some cases.

src/garage/torch/algos/vpg.py
```diff
@@ -0,0 +1,286 @@
...
+
+        if self.entropy_regularzied:
+            grad += self.policy_ent_coeff * policy_entropy
+
+        return -grad.unsqueeze(0)  # Maximize
```
zhanpenghe: I think you want to minimize this?

src/garage/torch/algos/vpg.py
zhanpenghe: Also let's just rename this variable to something else other than grad? This is not really gradients.

src/garage/torch/algos/vpg.py
```diff
@@ -0,0 +1,286 @@
...
+        policy_entropy = self._get_policy_entropy(obs)
+
+        if self.maximum_entropy:
+            rewards = [
+                reward + self.policy_ent_coeff * policy_entropy
+                for reward in rewards
+            ]
+
+        baselines = self._get_baselines(path)
+        path['baselines'] = baselines.detach().numpy()
```
zhanpenghe: I think we should always make `baseline.predict` detach so we don't need to detach here.

src/garage/torch/algos/vpg.py
```diff
@@ -0,0 +1,286 @@
...
+
+
+class VPG(RLAlgorithm):
+    """Vanilla Policy Gradient.
+
+    Args:
+        env_spec (garage.envs.EnvSpec): Environment specification.
+        policy (garage.torch.policies.base.Policy): Policy.
+        baseline : The baseline.
+        scope (str): Scope for identifying the algorithm.
```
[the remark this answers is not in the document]
yonghyuc: fixed

src/garage/torch/policies/gaussian_mlp_policy.py
```diff
@@ -1,10 +1,11 @@
 """GaussianMLPPolicy."""
-from torch import nn
 
-from garage.torch.policies import Policy
+import torch
 
+from garage.torch.policies.base import Policy
 
-class GaussianMLPPolicy(nn.Module, Policy):
```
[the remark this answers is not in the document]
yonghyuc: I just think we use get_action (or get_actions) method more instead of forward.
So I assume this GaussianMLPPolicy is not a module.

If we need to put `nn.Module` back, I will fix this

src/garage/torch/algos/vpg.py
```diff
@@ -0,0 +1,286 @@
...
+        self.max_kl_step = max_kl_step
+        self.policy_ent_coeff = policy_ent_coeff
+        self.use_softplus_entropy = use_softplus_entropy
+        self.stop_entropy_gradient = stop_entropy_gradient
+        self.entropy_method = entropy_method
+        self.eps = 1e-8
+
+        self.maximum_entropy = (entropy_method == 'max')
+        self.entropy_regularzied = (entropy_method == 'regularized')
+        #     Entropy regularization, max entropy rl
```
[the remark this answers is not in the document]
yonghyuc: fixed - deleted

src/garage/torch/algos/vpg.py
yonghyuc: fixed

src/garage/torch/algos/vpg.py
yonghyuc: image

I thought we need to sum the (log likelihood * advantage) and then get mean value across the sum values.

Could you explain more about this part?

yonghyuc: I think the `tf.dynamic_partition` has an important role in "[validating](https://github.com/rlworkgroup/garage/blob/1d649d61b52830fd8303b52d0a3430af94e5b92b/src/garage/tf/misc/tensor_utils.py#L62)" true values from paddings. However, I cannot find PyTorch function similar to `tf.dynamic_partition`.

Also, for [computing advantage value](https://github.com/rlworkgroup/garage/blob/1d649d61b52830fd8303b52d0a3430af94e5b92b/src/garage/tf/misc/tensor_utils.py#L250), (I am not sure about this but,) TensorFlow Conv1d is different from PyTorch Conv1d. (PyTorch doesn't have a filter.. (instead it has 'kernel size').
And, I can't use discount factor in PyTorch cumsum or cumprod.

Do you have any idea?

src/garage/torch/algos/vpg.py
yonghyuc: I think I need to maximize the value and `Adam` optimizer try to minimize the value (loss).
This is why I put the "-" on the "grad" value.

Am I understanding correctly?

src/garage/torch/algos/vpg.py
zhanpenghe: The `D_k` here should be referring to the sampled batch so this is divided the size of the whole batch. You can see it as taking a mean across the whole batch. Please verify since I did not take a look at the paper but I believe it is the right thing to do.

src/garage/torch/algos/vpg.py
zhanpenghe: Yes you are right. I think you can just remove the comment since it is a bit confusing.

src/garage/torch/algos/vpg.py
zhanpenghe: I think you can actually just do the validation by multiply everything with the `valids`.

yonghyuc:
Could anyone explain [log_likelihood](https://github.com/rlworkgroup/garage/blob/1d649d61b52830fd8303b52d0a3430af94e5b92b/src/garage/tf/distributions/diagonal_gaussian.py#L73) in diagonal_gaussian.py?

```
    def log_likelihood(self, xs, dist_info):
        means = dist_info['mean']
        log_stds = dist_info['log_std']
        zs = (xs - means) / np.exp(log_stds)
        return - np.sum(log_stds, axis=-1) - \
            0.5 * np.sum(np.square(zs), axis=-1) - \
            0.5 * self.dim * np.log(2 * np.pi)
```

this looks quite different from [Pytorch log_prob](https://pytorch.org/docs/stable/_modules/torch/distributions/normal.html#Normal.log_prob)

src/garage/torch/policies/gaussian_mlp_policy.py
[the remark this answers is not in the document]
avnishn: can we have garage.torch.policies.base.Policy inherit from not only abstract base class, but also nn.Module?
Then we can have gaussianMLPPolicy, and all other new policies inherit from only garage.torch.policies.base.Policy.

src/garage/torch/algos/vpg.py
avnishn: @yonghyuc I agree with zhanpenghe. At some point, you'll need to divide the reward/-loss by the number of trajectories provided by the sampler. Because the gradient is a linear operation, you can do that here on line 162.

src/garage/torch/algos/vpg.py
```diff
@@ -0,0 +1,286 @@
...
+        return np.mean([sum(path['rewards']) for path in paths])
+
+    def _optimize_policy(self, itr, paths):
+        loss = [self._get_policy_loss(path) for path in paths]
+
+        self.optimizer.zero_grad()
+        torch.cat(loss).mean().backward()
+        self.optimizer.step()
+
+    def _get_policy_loss(self, path):
```
avnishn: can we call this instead _get_policy_gradient, or something along those lines. I think it's a bit confusing to use this terminology because the original paper does not describe this as the loss.

src/garage/torch/algos/vpg.py
yonghyuc: fixed - delete comment

src/garage/torch/policies/gaussian_mlp_policy.py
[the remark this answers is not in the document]
zhanpenghe: Hmm I think we have followed the tf tree style to implement DDPG, which I think looks pretty good. Could get back to that? I am sorry for the back-and-forth design decision.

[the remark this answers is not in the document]
zhanpenghe: Please see #815 .

src/garage/torch/algos/vpg.py
```diff
@@ -0,0 +1,231 @@
...
+        filter = torch.full((1, 1, 1, max_path_length), discount * gae_lambda)
+        self.filter = torch.cumprod(filter, dim=-1)
+
+        super().__init__(policy=policy,
+                         baseline=baseline,
+                         discount=discount,
+                         max_path_length=max_path_length,
+                         n_samples=n_samples)
+
+    def _check_entropy_configuration(self, entropy_method, center_adv,
```
avnishn: Do we need this in a vpg implementation? When might someone to use an entropy gradient with vpg?

avnishn: @krzentner

krzentner: Maximum entropy RL is useful (and meaningful) even with VPG, right? That requires the entropy gradient.

ryanjulian:
@avnishn yes to all.

yonghyuc:
benchmark result between PyTorch VPG and TF VPG

Swimmer-v2
Swimmer

Half Cheetah-v2
HalfCheetah

InvertedPendulum-v2
InvertedPendulum(1)

src/garage/torch/policies/base.py
```diff
@@ -35,3 +35,8 @@ def observation_space(self):
     def action_space(self):
         """Policy action space."""
         return self._env_spec.action_space
+
+    @property
+    def vectorized(self):
```
krzentner: I think we're just going to delete the method, and assume that all policies are vectorized (i.e. that all policies implement `get_actions`). You can make this change in a later PR though.

ryanjulian:
Can you run this on the full Mujoco1M suite and report the results? Please also include the random seeds you used for each trial (on your plots).

zhanpenghe:
I would suggest also include your codes for benchmarking into the benchmark test.

ryanjulian:
Yes! It should go in tests/benchmarks

yonghyuc:
Benchmark results:


HalfCheetah-v2_benchmark

Hopper-v2_benchmark

InvertedPendulum-v2_benchmark

Reacher-v2_benchmark

Swimmer-v2_benchmark

Walker2d-v2_benchmark

src/garage/torch/policies/base.py
yonghyuc: I tried to but I checked that `BatchPolopt` use the `vectorized` property to decide `sampler_cls`
[(code)](https://github.com/rlworkgroup/garage/blob/7bb784efdefc7fd8f160718c1d989144ba91e07f/src/garage/np/algos/batch_polopt.py#L40)
Do I need to change the `BatchPolopt`?

yonghyuc:
> Yes! It should go in tests/benchmarks

Done! Could you check the graph and review the changes please?

src/garage/torch/policies/base.py
krzentner: You don't need to change `BatchPolopt` in this PR, but it would be good for this function to always return True, and have a default implementation of `get_actions` in terms of `get_action`.

Once every policy in the code base is always vectorized, then we'll modify `BatchPolopt`.

src/garage/torch/algos/vpg.py
```diff
@@ -0,0 +1,230 @@
...
+        rewards = torch.stack([
+            self._add_padding(torch.Tensor(path['rewards']),
+                              self.max_path_length)  # noqa: E501
+            for path in paths
+        ])
+
+        if self.maximum_entropy:
+            rewards += self.policy_ent_coeff * policy_entropies
+
+        deltas = (rewards + self.discount * F.pad(baselines,
```
ryanjulian: there is a helper function for this calculation in TF, because doing it with convolution can be quite tricky.

can you pull the advantage calculation out into a helper function and write a unit test for it? https://github.com/rlworkgroup/garage/blob/master/src/garage/tf/misc/tensor_utils.py#L201

you can also use your helper function to verify that you get the same results as the TF version for the same inputs.

src/garage/torch/algos/vpg.py
```diff
@@ -0,0 +1,230 @@
...
+            for path in paths
+        ])
+
+        outputs = advantages * log_likelihoods
+
+        if self.entropy_regularzied:
+            outputs += self.policy_ent_coeff * policy_entropies
+
+        loss = torch.cat(self._filter_valids(outputs, valids))
+        (-loss).mean().backward()
```
ryanjulian: this seems like an important step. perhaps you should add a comment for it.

krzentner:
This looks pretty good. The graphs mostly show this implementation performing well, but the half-cheetah benchmark still clearly shows the tf implementation performing better. Do we have any idea why that might be? I'm not really sure what tricks are in the tf implementation that might have been missed here.

src/garage/torch/policies/gaussian_mlp_policy.py
```diff
@@ -18,36 +17,54 @@ class GaussianMLPPolicy(GaussianMLPModule, Policy):
...
         with torch.no_grad():
             dist = self.forward(observations)
-            std = dist.variance**0.5
-            return dist.rsample().detach().numpy(), dict(mean=dist.mean,
-                                                         log_std=np.log(std))
+            return (dist.rsample().numpy(),
+                    dict(mean=dist.mean.numpy(),
+                         log_std=(dist.variance**.5).log().numpy()))
+
+    def get_log_likelihood(self, observation, action):
```
ryanjulian: these functions are differentiable right? in that case, i think you can omit the `get` part of the name. we tend to use `get` to distinguish non-differentiable functions.

tests/benchmarks/test_benchmark_vpg.py
```diff
@@ -0,0 +1,292 @@
...
+    'center_adv': True,
+    'learning_rate': 1e-2,
+    'discount': 0.99,
+    'n_epochs': 250,
+    'max_path_length': 500
+}
+
+
+def run_garage_pytorch(env, seed, log_dir):
+    '''
```
ryanjulian: while you're here, please update these comments to conform to CONTRIBUTING.md

src/garage/torch/policies/gaussian_mlp_policy.py
yonghyuc: [Pytorch description and sample code](https://pytorch.org/docs/stable/distributions.html) shows they are differentiable.
I will change the name.

tests/benchmarks/test_benchmark_vpg.py
yonghyuc: change the comments

src/garage/torch/algos/vpg.py
yonghyuc: add comments 
-----------
using a negative because optimizers use gradient descent, whilst we want gradient ascent.
-----------

yonghyuc:
Fix few bugs in VPG and graph looks better.

HalfCheetah-v2_benchmark

Hopper-v2_benchmark

InvertedPendulum-v2_benchmark

Reacher-v2_benchmark

Swimmer-v2_benchmark

Walker2d-v2_benchmark

src/garage/torch/algos/vpg.py
yonghyuc: extract `compute_advantages` function and make test case for the function

src/garage/torch/misc/tensor_utils.py
```diff
@@ -0,0 +1,15 @@
+"""utility functions for PyTorch tensors."""
```
ryanjulian: i know this is the location for TF but we are trying to actually get rid of generic misc/utils modules.

Can you put this into a module inside torch.algos instead? You can call it `loss_function_utils.py`

tests/garage/torch/misc/test_tensor_utils.py
```diff
@@ -0,0 +1,46 @@
...
+import garage.tf.misc.tensor_utils as tf_utils
+import garage.torch.misc.tensor_utils as torch_utils
+from tests.fixtures import TfGraphTestCase
+
+
+class TestTensorUtil(TfGraphTestCase):
+    # yapf: disable
+    @pytest.mark.parametrize('rewards_val, baselines_val', [
+        (np.ones((4, 6)), np.zeros((4, 6))),
+        (np.random.randn(4, 6), np.random.randn(4, 6)),
```
ryanjulian: please don't use randomness in tests. if you need to, you can instead generate a static random data set, or use alternatives like digits of pi or fibonacci sequences.

src/garage/torch/algos/vpg.py
```diff
@@ -0,0 +1,226 @@
...
+        valids = [len(path['actions']) for path in paths]
+
+        baselines = torch.stack([
+            self._add_padding(self._get_baselines(path), self.max_path_length)
+            for path in paths
+        ])
+
+        rewards = torch.stack([
+            self._add_padding(torch.Tensor(path['rewards']),
+                              self.max_path_length)  # noqa: E501
```
ryanjulian: i don't understand why this is here, because this line is not over length.

src/garage/torch/algos/vpg.py
```diff
@@ -0,0 +1,226 @@
...
+
+        return samples_data['average_return']
+
+    def _optimize_policy(self, itr, paths):
+        self.optimizer.zero_grad()
+
+        policy_entropies = torch.stack([
+            self._add_padding(
+                self.policy.get_entropy(torch.Tensor(path['observations'])),
+                self.max_path_length)  # noqa: E501
```
ryanjulian: this line is not over lenght?

src/garage/torch/algos/vpg.py
```diff
@@ -0,0 +1,226 @@
...
+
+        advantages = utils.compute_advantages(self.discount, self.gae_lambda,
+                                              self.max_path_length, baselines,
+                                              rewards)
+
+        if self.center_adv:
+            means, vars = list(
+                zip(*[(valid_adv.mean(), valid_adv.var())
+                      for valid_adv in self._filter_valids(advantages, valids)]
+                    ))  # noqa: E501
```
ryanjulian: this line is not over length?

tests/benchmarks/test_benchmark_vpg.py
```diff
@@ -0,0 +1,258 @@
...
+        algo = TF_VPG(
+            env_spec=env.spec,
+            policy=policy,
+            baseline=baseline,
+            max_path_length=hyper_parameters['max_path_length'],
+            discount=hyper_parameters['discount'],
+            center_adv=hyper_parameters['center_adv'],
+            optimizer_args=dict(
+                tf_optimizer_args=dict(
+                    learning_rate=hyper_parameters['learning_rate'], ),  # noqa
```
ryanjulian: why noqa?

tests/garage/torch/misc/test_tensor_utils.py
```diff
@@ -0,0 +1,46 @@
...
+    def test_compute_advantages(self, rewards_val, baselines_val):
+        discount = 0.99
+        gae_lambda = 1.
+        max_len = rewards_val.shape[-1]
+
+        torch_advs = torch_utils.compute_advantages(
+            discount,
+            gae_lambda,
+            max_len,
+            torch.Tensor(baselines_val),  # NOQA
```
ryanjulian: why noqa?

src/garage/torch/misc/tensor_utils.py
```diff
@@ -0,0 +1,15 @@
...
+import torch
+import torch.nn.functional as F  # noqa
```
ryanjulian: which QA rule is this violating? perhaps we make add a blanket exception to the configs.

please only add exceptions for specific rules. it helps us track which rules are creating problems, and also allows us to quickly remove exceptions when we find a solution.

tests/garage/torch/misc/test_tensor_utils.py
```diff
@@ -0,0 +1,46 @@
...
+import pytest
+import tensorflow as tf
+import torch
+
+import garage.tf.misc.tensor_utils as tf_utils
+import garage.torch.misc.tensor_utils as torch_utils
+from tests.fixtures import TfGraphTestCase
+
+
+class TestTensorUtil(TfGraphTestCase):
```
ryanjulian: in the future i'll definitely request a test which uses known inputs/outputs rather than just comparing output to some other function, but this will suffice for this PR.

src/garage/torch/misc/tensor_utils.py
yonghyuc: done

tests/garage/torch/misc/test_tensor_utils.py
yonghyuc: done

src/garage/torch/algos/vpg.py
yonghyuc: removed

src/garage/torch/algos/vpg.py
yonghyuc: removed

tests/benchmarks/test_benchmark_vpg.py
yonghyuc: removed

src/garage/torch/misc/tensor_utils.py
yonghyuc: removed

tests/garage/torch/misc/test_tensor_utils.py
yonghyuc: removed

src/garage/torch/algos/vpg.py
yonghyuc: removed

tests/benchmarks/test_benchmark_vpg.py
```diff
@@ -0,0 +1,256 @@
...
+
+        algo = TF_VPG(env_spec=env.spec,
+                      policy=policy,
+                      baseline=baseline,
+                      max_path_length=hyper_parameters['max_path_length'],
+                      discount=hyper_parameters['discount'],
+                      center_adv=hyper_parameters['center_adv'],
+                      optimizer_args=dict(tf_optimizer_args=dict(
+                          learning_rate=hyper_parameters['learning_rate']),
+                                          verbose=True))  # noqa E126
```
yonghyuc: I put `noqa E126` because yapf keep reformatting this code and Flake8 says "E126 continuation line over-indented for hanging indent"

src/garage/torch/algos/vpg.py
ryanjulian: I don't see the term REINFORCE anywhere in this docstring

tests/benchmarks/test_benchmark_vpg.py
ryanjulian: ah. please stick with what flake8 prefers.

please format it how you think it looks best (and satisfies flake8) and then use `# yapf: disable` to prevent YAPF from messing with the code.

tests/garage/torch/algos/test_loss_function_utils.py
```diff
@@ -0,0 +1,38 @@
...
+import garage.tf.misc.tensor_utils as tf_utils
+import garage.torch.algos.loss_function_utils as torch_loss_utils
+from tests.fixtures import TfGraphTestCase
+
+
+class TestLossFunctionUtils(TfGraphTestCase):
+
+    def test_compute_advantages(self):
+        discount = 0.99
+        gae_lambda = 1.
```
ryanjulian: test this with something other than 1, since this `gae_lambda` won't produce a meaningful decay.

tests/garage/torch/algos/test_loss_function_utils.py
```diff
@@ -0,0 +1,38 @@
...
+        rewards_val, baselines_val = np.ones((4, 6)), np.zeros((4, 6))
```
ryanjulian: please use a non-zero value for the baselines. otherwise, an entire component of this computation is inactive.

ryanjulian: here are some ideas for non-trivial sequences which are not-random:
`np.arange`, the fibonacci sequence, digits of `np.pi`, digits of e, etc. to get non-huge numbers you can divide many of these by some constant multiple.

tests/benchmarks/test_benchmark_vpg.py
yonghyuc: thanks! fixed

tests/garage/torch/algos/test_loss_function_utils.py
yonghyuc: fixed

tests/garage/torch/algos/test_loss_function_utils.py
yonghyuc: fixed

src/garage/torch/algos/vpg.py
ryanjulian: @yonghyuc please address this comment...

src/garage/torch/policies/base.py
ryanjulian: So is the right thing to do here to return `True`?

tests/garage/torch/algos/test_loss_function_utils.py
```diff
@@ -0,0 +1,52 @@
...
+    @pytest.mark.parametrize('gae_lambda, rewards_val, baselines_val', [
+        (0.4,
+         np.ones((4, 6)),
+         np.zeros((4, 6))),
+        (0.8,
+         np.repeat(np.expand_dims([3, 1, 4, 1, 5, 9], axis=0), repeats=4, axis=0),  # noqa: E501
+         np.repeat(np.expand_dims(np.arange(6), axis=0), repeats=4, axis=0)),
+        (1.2,
+         np.ones((4, 6)),
+         np.repeat(np.expand_dims([1, 1, 2, 3, 5, 8], axis=0), repeats=4, axis=0)),  # noqa: E501
```
ryanjulian: this is very difficult to read and runs afoul of the style checker.

can you make these cases constants at the top of the file and reference them in the parameterize case?

it may actually be simpler and easier to read to just include them as large static numpy arrays at the top of the file than constructing them with `np.repeat`/`np.expand_dims`. As written, it would take a couple minutes to interpret what the actual inputs are here.

yonghyuc: fixed!

src/garage/torch/policies/base.py
krzentner: Yeah, this method should return `True`, and `get_actions` should be implemented in terms of `get_action`.

src/garage/torch/algos/vpg.py
yonghyuc: sorry, 
I just added "REINFORCE" word in the docstring

tests/garage/torch/algos/test_loss_function_utils.py
```diff
@@ -0,0 +1,54 @@
...
+import pytest
+import tensorflow as tf
+import torch
+
+import garage.tf.misc.tensor_utils as tf_utils
+import garage.torch.algos.loss_function_utils as torch_loss_utils
+from tests.fixtures import TfGraphTestCase
+
+
+def stack(arr, d):
```
ryanjulian: how does this differ from np.stack?

tests/garage/torch/algos/test_loss_function_utils.py
```diff
@@ -0,0 +1,54 @@
...
+pi_digits = [3, 1, 4, 1, 5, 9]
+e_digits = [2, 7, 1, 8, 2, 8]
+fibs = [1, 1, 2, 3, 5, 8]
+
+
+class TestLossFunctionUtils(TfGraphTestCase):
+    # yapf: disable
+    @pytest.mark.parametrize('gae_lambda, rewards_val, baselines_val', [
+        (0.4, np.ones((4, 6)), np.zeros((4, 6))),
+        (0.8, stack(pi_digits, 4), stack(arrange, 4)),
```
ryanjulian: is this really simpler than just defining them as static arrays at the top of the file?

tests/garage/torch/algos/test_loss_function_utils.py
```diff
@@ -0,0 +1,54 @@
...
+import garage.tf.misc.tensor_utils as tf_utils
+import garage.torch.algos.loss_function_utils as torch_loss_utils
+from tests.fixtures import TfGraphTestCase
+
+
+def stack(arr, d):
+    return np.repeat(np.expand_dims(arr, axis=0), repeats=d, axis=0)
+
+
+arrange = np.arange(6)
```
ryanjulian: please use all caps for constants.

tests/garage/torch/algos/test_loss_function_utils.py
yonghyuc: np.stack join a sequence of arrays, and this is doing kind of copying the array d times and join the arrays.

tests/garage/torch/algos/test_loss_function_utils.py
yonghyuc: fixed

tests/garage/torch/algos/test_loss_function_utils.py
yonghyuc: fixed

PR #834, opened 2019-08-08
Add gpu version of pip package

gitanshu:


gitanshu:
> This looks good to me, as long as you've tested it. The `tensorflow-gpu` package doesn't conflict with the `tensorflow` package, right?

The packages don't conflict. And I did see python processes in nvidia-smi when I ran a couple examples. Not sure if that's enough to say it works.

PR #835, opened 2019-08-08
Cleanup EC2/cluster docs and examples

gitanshu:


docs/user/experiments.rst
```diff
@@ -110,7 +110,7 @@ You should see some log messages like the following:
...
 
-The benefit for launching experiment this way is that we separate the configuration of experiment parameters and the actual execution of the experiment. `run_experiment` supports multiple ways of running the experiment, either locally, locally in a docker container, or remotely on ec2 (see the section on :ref:`cluster`). Multiple experiments with different hyper-parameter settings can be quickly constructed and launched simultaneously on multiple ec2 machines using this abstraction.
+The benefit for launching experiment this way is that we separate the configuration of experiment parameters and the actual execution of the experiment. `run_experiment` supports multiple ways of running the experiment, either locally, locally in a docker container, or remotely on ec2. Multiple experiments with different hyper-parameter settings can be quickly constructed and launched simultaneously on multiple ec2 machines using this abstraction.
```
gitanshu: Not sure if / how I should delete the 2 references to ec2 here?

zhanpenghe: Yes please do that. 

@krzentner do we have any plan to add these features back? Although I think they are not used frequently (or at all) recently.

krzentner: So Zequn and I had a fairly long discussion on Thursday about this topic, as well as the design of `run_experiment` generally. The main takeaway was that we feel like the current `run_experiment` infrastructure adds a lot of complexity without actually adding much value.

Specifically, all of the functionality which generates the experiment command is really complicated, but has some large limitations. Firstly, it basically pickles `run_task` and its arguments and then base64 encodes them into arguments. This only works as long as those values can be fit in 4KiB after encoding, since that's the limit on the length of command arguments on Linux. It also is only sufficient if the rest of the code the experiment depends on has been copied to where this command will be run. A more reliable method would be to save this pickle data to a file and then invoke that. If we're running the experiment remotely, then we can just copy this file with all of the source files we need to copy anyways. This method should be substantially easier to implement. I also propose that `run_experiment` should be completely optional in using garage. In other words, we should eliminate any other global setup it currently does (such as creating the data directory / configuring the GPU device) and move it into the runner (or some component of the runner).

Generally, I do think having an easy way of running garage remotely is a good idea, although I'm not particularly tied to EC2 per se. I also don't particularly feel like our use case here is particularly unusual. Basically, we want to copy a subset of our current directory to a remote host, install any dependent python packages, then run a command there. We could fairly easily handle the second step, and the rest feels like operations so typical that some python library should know how to do them automatically. In the spirit of making garage use standard / off the shelf components, I propose we choose one of those libraries to be the recommended way of remotely running experiments.

@zequnyu , @ryanjulian What do you think about this topic?

zhanpenghe: I think this is a good idea. According to my experience of running garage on remote servers (gcp and resl's), I think this functionality is useful especially for hyper parameter sweep. I think we can target this to be just sending an experiment to a fresh linux machine since that make things more general and I think our docker now it pretty good for that.

zhanpenghe: I believe that we have also implemented utilities for running an experiment with a specific commit. That could be very useful too.

ryanjulian: tl;dr:
I'm pretty much in agreement with what KR has said. I think the ability to run on a cluster is generally useful, but that the current design has shown that trying to bake that logic directly into the main experiment code is fraught. I think ripping out the current "cluster" support is long-overdue.

I'd rather see garage experiments concern themselves with being the best at being a *single training process*, which is *well-behaved* with respect to its neighbors (e.g. logging into the correct directory prefixes, taking up the number of resources its instructed to use, etc.), and make it *easy to expose tunable parameters* via the command line.

I think if we have those three ingredients, it's easy to wrap garage with something which decides how to deploy/schedule processes onto a cluster -- whether that be something we author or some other tool such as an HPC cluster scheduler. The things most cluster schedulers usually have in common is that their interface is basically "give me a path to a process and a set of arguments for that process, and I will give you back a directory full of the output files for that process". It's incredibly effective, and easier to use than what a garage has going on now.

ryanjulian: To expand, there are plenty of tools which will make an AWS/Azure/GCP account look like "give me a Docker container, a command, and a storage path, and I will figure out how to run that for you using x/y/z resources."  If we can just design garage properly, we can provide good examples for how to use those tools (and make sure to support them well), rather than haphazardly implementing a bad version of Kubernetes ourselves.

zequnyu: Yes, considering the current stage, I think it's good to make an independent `LocalRunner` as the next step, which would make our experiment API optional.

Issue #702, opened 2019-05-29
Refactor garage.tf.envs into garage.sampler


ryanjulian:
* These are not environment code nor wrappers
* They don't actually contain tf-specific implementations

zequnyu:
Close via #836.

PR #836, opened 2019-08-09
Move tf vec_env executor to garage.sampler

zequnyu:
#702. 
Also serves as the step before refactoring tf_samplers.


krzentner:
I've added the new / more active contributors, since I want them to start reviewing more. You only need one LGTM from among them.

PR #838, opened 2019-08-10
Add a new ReplayBuffer, "PathBuffer"

### LINKED ISSUE #837: Fix ReplayBuffer / HER for variable lengths paths
The current ReplayBuffer stores transitions in fixed length blocks called "episodes". One would expect each of those transitions to be a path. However, the implementations of `ReplayBuffer` and `OffPolicyVectorizedSampler` do not ensure that is the case, and instead merely collect samples from many separate paths, splitting them into episodes after a fixed number of time steps across all paths. This means if any paths are not exactly `time_horizon` transitions long, then episodes will actually contain transitions from multiple paths. Then, when HerReplayBuffer samples random episodes and post-processes them it produces invalid results since the episode can contain transitions from multiple paths.

krzentner:
This is a replay buffer implementation that differs from the existing
replay buffer implementation in that it can efficiently store variable
length paths. Ultimately, it should replace the existing replay buffer
and allow us to fix #837.

It may also be useful for as yet unimplemented off algorithms which want
to process whole paths.

### DIFF
--- a/src/garage/replay_buffer/path_buffer.py
+++ b/src/garage/replay_buffer/path_buffer.py
@@ -1,0 +1,139 @@
+"""A replay buffer that efficiently stores and can sample whole paths."""
+import collections
+
+import numpy as np
+
+
+class PathBuffer:
+    """A replay buffer that stores and can sample whole paths.
+
+    This buffer only stores valid steps, and doesn't require paths to
+    have a maximum length.
+    Args:
+        capacity_in_steps (int): total memory allocated for the buffer
+    """
+
+    def __init__(self, capacity_in_transitions):
+        self._capacity = capacity_in_transitions
+        self._transitions_stored = 0
+        self._first_idx_of_next_path = 0
+        # Each path in the buffer has a tuple of two ranges in
+        # self._path_segments. If the path is stored in a single contiguous
+        # region of the buffer, the second range will be range(0, 0).
+        # The "left" side of the deque contains the oldest path.
+        self._path_segments = collections.deque()
+        self._buffer = {}
+
+    def add_path(self, path):
+        """Add a path to the buffer.
+
+        Args:
+            path(dict): A dict of array of shape (path_len, flat_dim)
+        """
+        for key, buf_arr in self._buffer.items():
+            path_array = path.get(key, None)
+            if path_array is None:
+                raise ValueError('Key {} missing from path.'.format(key))
+            elif (len(path_array.shape) != 2
+                  or path_array.shape[1] != buf_arr.shape[1]):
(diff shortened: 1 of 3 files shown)


krzentner:
Note that I intend to do any related refactoring in later PRs.

src/garage/replay_buffer/path_buffer.py
```diff
@@ -0,0 +1,132 @@
...
+        # Remove paths which will overlap with this one.
+        while (self._path_segments and self._segments_overlap(
+                first_seg, self._path_segments[0][0])):
+            self._path_segments.popleft()
+        while (self._path_segments and self._segments_overlap(
+                second_seg, self._path_segments[0][0])):
+            self._path_segments.popleft()
+        self._path_segments.append((first_seg, second_seg))
+        for key, array in path.items():
+            buf_arr = self._get_or_allocate_key(key, array)
```
avnishn: This logic seems similar to using a collections.defaultdict data structure. Maybe that is best used here.

krzentner: The problem is that there's no type we can set for the `defaultdict` that will work, since the array's shape depends on what we're trying to insert into it.

src/garage/replay_buffer/path_buffer.py
```diff
@@ -0,0 +1,132 @@
...
+                length_key = key
+            elif len(value) != length:
+                raise ValueError('path has inconsistent lengths between '
+                                 '{!r} and {!r}.'.format(length_key, key))
+        if not length:
+            raise ValueError('Nothing in path')
+        return length
+
+    @staticmethod
+    def _segments_overlap(seg_a, seg_b):
```
avnishn: Could be more clear with included docstrings what this function does

krzentner:
So the replay doesn't act exactly like a fifo queue, in that there's no way to pop data from it, and hence no need to manage garbage / uninitialized memory. Most of the complexity here comes from something hardware fifo's don't support: the elements in the buffer are variable size.

I think a slightly simpler implementation might be possible, by always "spilling" any paths that would need to be split across the end of the buffer back to the start, and decreasing `self._transitions_stored` to match. This would result in slightly less efficient memory usage. Having said that, the current implementation appears to be correct and implements all of the functionality we need, so I don't see a good reason to go chasing the simplest possible implementation.

I'll merge this one the CI troubles get fixed.

avnishn:
Agreed, I didn't consider the variable sizes of paths in the buffer. This would make it challenging to keep track of the segments that make up a path, which you do in this pr by using the path segments deque.

Issue #701, opened 2019-05-29
Refactor garage.tf.samplers into garage.sampler


ryanjulian:

zequnyu:
Do we still have other work to do apart from #840? @ryanjulian

ryanjulian:
Hmm actually no. I looked and I think this is about as minimal as it gets right now. Good work everyone!

PR #840, opened 2019-08-13
Move vectorized samplers from tf to garage.sampler

zequnyu:
#701.
- refactor tests accordingly
- fix tf_vars initialize bug in `examples/np`

### DIFF
--- a/examples/np/cem_cartpole.py
+++ b/examples/np/cem_cartpole.py
@@ -11,10 +11,10 @@ Results:
 from garage.experiment import run_experiment
 from garage.np.algos import CEM
 from garage.np.baselines import LinearFeatureBaseline
+from garage.sampler import OnPolicyVectorizedSampler
 from garage.tf.envs import TfEnv
 from garage.tf.experiment import LocalTFRunner
 from garage.tf.policies import CategoricalMLPPolicy
-from garage.tf.samplers import OnPolicyVectorizedSampler
 
 
 def run_task(snapshot_config, *_):
@@ -26,6 +26,8 @@ def run_task(snapshot_config, *_):
             name='policy', env_spec=env.spec, hidden_sizes=(32, 32))
 
         baseline = LinearFeatureBaseline(env_spec=env.spec)
+
+        runner.initialize_tf_vars()
 
         n_samples = 20
 
--- a/src/garage/sampler/batch_sampler.py
+++ b/src/garage/sampler/batch_sampler.py
@@ -13,7 +13,7 @@ class BatchSampler(BaseSampler):
     """
 
     def __init__(self, algo, env):
-        super(BatchSampler, self).__init__(algo, env)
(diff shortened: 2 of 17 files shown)


zequnyu:
Also considering move `LocalTFRunner.initialize_tf_vars()` to `__enter__()` and make it private. Since in cem and cma_es it actually gets initialised twice. @naeioi What do you think?

examples/np/cem_cartpole.py
```diff
@@ -27,6 +27,8 @@ def run_task(snapshot_config, *_):
 
         baseline = LinearFeatureBaseline(env_spec=env.spec)
 
+        runner.initialize_tf_vars()
```
zhanpenghe: Why is `runner.setup()` not working here?

zequnyu: Because CategoricalMLPPolicy is a tf policy and requires variables before `setup()`.

src/garage/sampler/batch_sampler.py
```diff
@@ -12,8 +12,9 @@ class BatchSampler(BaseSampler):
 
     """
 
-    def __init__(self, algo, env):
-        super(BatchSampler, self).__init__(algo, env)
+    def __init__(self, algo, env, n_envs):
+        super().__init__(algo, env)
+        self.n_envs = n_envs
```
zhanpenghe: This variable does not seem to be used?

zequnyu: Ohh yes, updated. Thanks for pointing out.

examples/np/cem_cartpole.py
zhanpenghe: Hmm can you send me a permalink of the variables?

zequnyu: So it is used here:
https://github.com/rlworkgroup/garage/blob/2c23600100250344bf0d9a36f89f47d5e944c9e7/src/garage/np/algos/cem.py#L63
We should probably do some refactoring.

zhanpenghe: Yes please file an issue for this. I think we can just initialize it as `None` at the beginning and only set it when the training start.

zhanpenghe:
Nice work. This looks good to me!

naeioi:
> 
> 
> Also considering move `LocalTFRunner.initialize_tf_vars()` to `__enter__()` and make it private. Since in cem and cma_es it actually gets initialised twice. @naeioi What do you think?

This will not work. Because when runner is entered, policy is not declared yet. 
That being said, I wonder why CEM does not requires manually initialization before. 
Any clues on this? @zequnyu 

I am hectic with the last week of my internship. Starting from next week I will resume working on garage. Looks like I have a lot to follow!

zequnyu:
@naeioi #842 should solve this.

Issue #841, opened 2019-08-13
Refactor get_param_values() in CEM and CMA_ES


zequnyu:
https://github.com/rlworkgroup/garage/blob/2c23600100250344bf0d9a36f89f47d5e944c9e7/src/garage/np/algos/cem.py#L63
https://github.com/rlworkgroup/garage/blob/2c23600100250344bf0d9a36f89f47d5e944c9e7/src/garage/np/algos/cma_es.py#L43

`get_param_values()` requires `LocalTFRunner.initialize_tf_vars()`, which is currently called in Runner's `setup()`. This means current examples and tests of CEM anad CMA_ES must initialize the variables twice. We could either set `get_param_values()` when start training or move `initialize_tf_vars()` to `__enter()__`.

zhanpenghe:
Hmm lets not moving `initialize_tf_vars` since most of the models are not initialized at that stage.

zequnyu:
Are you suggesting we should do something like this in algo, for example CEM? @zhanpenghe 

```python3
def train():
    # epoch-wise
    self.cur_std = self.init_std
    self.cur_mean = self.policy.get_param_values()
    # epoch-cycle-wise
    self.cur_params = self.cur_mean
    self.all_returns = []
    self.all_params = [self.cur_mean.copy()]
    # fixed
    self.n_best = int(n_samples * best_frac)
    assert self.n_best >= 1, (
        'n_samples is too low. Make sure that n_samples * best_frac >= 1')
    self.n_params = len(self.cur_mean)

    super().train()
```

zhanpenghe:
Yes.

PR #842, opened 2019-08-16
Resolve duplicate variables initialization in CEM and CMA_ES

zequnyu:
#841.
Moved some variables (requiring tf_vars) initialization just before `algo.train()` so that `LocalTFRunner` doesn't need to call `initialize_tf_vars()` twice.


src/garage/np/algos/cma_es.py
```diff
@@ -40,16 +40,22 @@ def __init__(self,
...
 
         self.all_params = self.sample_params()
         self.cur_params = self.all_params[0]
         self.policy.set_param_values(self.cur_params)
         self.all_returns = []
 
-    def sample_params(self):
-        return self.es.ask(self.n_samples)
+    def train(self, runner, batch_size):
+        self._initialize_vars()
```
naeioi: I suspect doing so will fail resume mechanism.

tests/garage/np/algos/test_cem.py
```diff
@@ -21,8 +21,6 @@ def test_cem_cartpole(self):
 
             baseline = LinearFeatureBaseline(env_spec=env.spec)
 
-            runner.initialize_tf_vars()
```
naeioi: We have already provided policies with tf session by runner, but looks like sometimes they want more. @ahtsan Do you think this problem will recur in the future? If so, we may need a mechanism more fundamental.

src/garage/np/algos/cma_es.py
zhanpenghe: This is actually fine. Please take a look at what the initialization is saying. It is not initializing tf tensors' value. @naeioi

zhanpenghe: I think to avoid confusion. Let's just put the content of `_initialize_vars` to `train` or rename this method. I am pretty bad at naming stuff so please give it a good name :).

src/garage/np/algos/cem.py
```diff
@@ -66,18 +79,14 @@ def __init__(self,
...
 
-    def sample_params(self, epoch):
-        extra_var_mult = max(1.0 - epoch / self.extra_decay_time, 0)
-        sample_std = np.sqrt(
-            np.square(self.cur_std) +
-            np.square(self.extra_std) * extra_var_mult)
-        return np.random.standard_normal(
-            self.n_params) * sample_std + self.cur_mean
+    def train(self, runner, batch_size):
+        self._initialize_vars()
```
zhanpenghe: Same as my below comment.

src/garage/np/algos/cem.py
```diff
@@ -66,18 +79,14 @@ def __init__(self,
         self.all_returns = []
         self.all_params = [self.cur_mean.copy()]
         # fixed
```
zhanpenghe: Also this comment is not saying anything I believe? Please remove or add information.

src/garage/np/algos/cma_es.py
zequnyu: If naming is the problem, we can just put the content inside. (Being said, I guess `_initialize_vars` is good here unless someone can figure out a better one)

naeioi: Ah true. It's me who wrote this function and I even forget how it works! Thanks for pointing out @zhanpenghe .

src/garage/np/algos/cem.py
naeioi: It actually means 'constant', in comparison to previous comments '(updated) epoch-wise' and '(updated) epoch-cycle-wise'. Please help me fix the bad wording @zequnyu

tests/garage/np/algos/test_cem.py
zhanpenghe: I think it is fine for now. The runner experience is pretty ok for single task RL.

zequnyu:
I'm having trouble with the pre-commit push. yapf is forcing docs in every module, class and public method. Can you look into it? @gitanshu

ryanjulian:
YAPF doesn't force anything about docs, so I think you are probably referring to flake8 or pylint.

In any case, make sure you are fully rebased, since the tools check only modified files. However, any difference between your branch and the base (even if it is from someone else's commit) is considered "modified".

zequnyu:
Yeah it's flake8. But having all the docs seems unnecessary for some cases (tests for example). Is this the new standard we should follow? @ryanjulian

ryanjulian:
The docs requirement should not be active for tests. Perhaps that got misconfigured?

ryanjulian:
Ah I see the root cause. @gitanshu can you reconfigure `.pre-commit-config.yaml` to ignore docs rules (I think `--ignore=D`) for /tests?

@zequnyu you can also just make the change yourself and I'll +1 it. I think @naeioi and @gitanshu both have the experience with pre-commit to show you how to do it.

zhanpenghe:
I think @lywong92 is having the same issue as yours.
@zequnyu

zequnyu:
Not sure `--ignore=D` enough to resolve this. For `/src`, do we need both `D100 Missing docstring in public module` and `D101 Missing docstring in public class`?  Also, not every method needs `D102 Missing docstring in public method`.

ryanjulian:
We require full docstrings in src/, per https://github.com/rlworkgroup/garage/blob/master/CONTRIBUTING.md#documentation

tests/ is an exception because tests don't really need autogenerated documentation or type hints.

various stuff outside src/ and test/ (e.g. examples/) is harder to pin down. I would err on the side of requiring docstrings, since that code gets read most by non-maintainers.

ryanjulian:
If you don't want to document a method or class, the solution is simple: don't make it part of the public API.

naeioi:
I suggest using `git push --no-verify` to make your change visible and we can see how flake8 breaks it.

gitanshu:
@zequnyu this should work for ignoring docstrings checks in tests directory
https://github.com/rlworkgroup/garage/commit/d85a31ec14feda48d8be709db2ae361c05110c5c#diff-380c6a8ebbbce17d55d50ef17d3cf906R13

zequnyu:
Thank you! I have one more problem with D204. Please see my comments under the commit.

PR #845, opened 2019-08-20
Enable numpy & google style docstrings in sphinx

gitanshu:
Before
before

After
after

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -82,6 +82,7 @@ extensions = [
     'sphinx.ext.coverage',
     'sphinx.ext.mathjax',
     'sphinx.ext.viewcode',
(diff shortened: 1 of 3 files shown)


ryanjulian:
Can you link to a branch build of the docs so I can compare before/after?

Or perhaps you can just attach screenshots so reviewers can understand the change.

gitanshu:
I jumped the gun on opening the PR and then couldn't figure out why rtfd won't let me access branch docs with private access. Anyway, added screenshots now. Notice the difference in how method parameters are presented. The _before_ Args section is also kinda broken.

gitanshu:
Will do! Also, just noticed the `Note` looks so nice now

PR #848, opened 2019-08-23
Move mujoco-py dependency to extras

krzentner:
This should finally allow users to `pip install garage` without a Mujoco license.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,6 @@ required = [
     'gym[all]==0.12.4',
     'joblib<0.13,>=0.12',
     'matplotlib',
-    'mujoco-py<2.1,>=2.0',
     'numpy==1.14.5',
     'psutil',
     # Pyglet 1.4.0 introduces some api change which breaks some
@@ -45,6 +44,10 @@ required = [
 
 # Dependencies for optional features
 extras = {}
+
+extras['mujoco'] = [
+    'mujoco-py<2.1,>=2.0',
+]
 
 extras['dm_control'] = [
     # dm_control throws an error during install about not being able to


zhanpenghe:
BTW, I think garage still need numpy to install first to make a one step installation.

krzentner:
> 
> 
> BTW, I think garage still need numpy to install first to make a one step installation.

Yes, I think that's the case until we upgrade pycma.

gitanshu:
@krzentner if we are trying to make mujoco optional, it should be declared after extras[all], so that it isn't part of all, right?

krzentner:
Nope. The point of `all` is that it should include all optional extras. That's generally the convention throughout the python ecosystem.

gitanshu:
Right. I don't get how this change allows `pip install garage` without mjkey. I tried to pip install garage either using a .whl file or a source archive in a new environment, and it complains about mjkey either way

ryanjulian:
Can you post a log to the error? We don't actually deploy to `pip` yet, so `pip install garage` will probably get you garbage. `pip install .` should work, though.

krzentner:
@gitanshu Is correct. I should have tested this more carefully. Because we still install `gym[all]`, we install `gym[mujoco]`, which still tries to build `mujoco_py`. I'll have a second PR up that actually fixes this.

gitanshu:
> Can you post a log to the error? We don't actually deploy to `pip` yet, so `pip install garage` will probably get you garbage. `pip install .` should work, though.

I wasn't trying to install from pip. I tried by building a wheel and installing that, using a source archive and also trying to install from the local folder directly. It's the same log as usual when we don't have mjkey, but install mujoco_py. As @krzentner said, the dependency was coming from gym[all].

ryanjulian:
can you find the line where something tries to import mujoco_py? that will be the culprit.

PR #854, opened 2019-08-30
Replace GaussianMLPPolicy and Baseline

nish21:


ryanjulian:
Where do we stand on the benchmarks?

Issue #509, opened 2019-02-05
Relaxing Mujoco Key install requirement


ancorso:
I am currently using garage for an RL project where I am running Travis-CI. Currently, garage requires a MuJoCo key to install, so I am having to include a key with my repo so that Travis-CI has access to it for the fresh install on each build. My project doesn't use MuJoCo so it would be great if the MuJoCo key requirement could be relaxed for installation.

Thanks!

ryanjulian:
Hi Anthony -- thanks for the FR!

Previously mujoco_py required a key for installation, but that was recently relaxed to only checking that a key is present.

If you take a look at our most recent [Dockerfiles](https://github.com/rlworkgroup/garage/tree/master/docker) (which we use in our own TravisCI), we recently edited them to use an empty `mjkey.txt` for building the file. I think you should be able to do the same with the setup script (just pass it an empty mjkey).

I'll keep this open as a housecleaning item to edit the setup scripts. We'd love it if you could send us a PR with the change you're suggesting.

For running garage in TravisCI, I highly recommend your checkout our [.travis.yml](https://github.com/rlworkgroup/garage/blob/master/.travis.yml) and [Dockerfiles](https://github.com/rlworkgroup/garage/tree/master/docker). Many hours have been sacrificed to get this whole suite running in CI. The installation script will probably burn a lot of your build minute quota, so I recommend running with Docker and retrieving the pre-built Dockerfile cache. You should be able to find our latest prebuilts on [Docker Hub](https://hub.docker.com/u/rlworkgroup).

ancorso:
This is great, thanks! I will try out your suggestions and submit a PR if I can get it working.

ancorso:
Hi Ryan, 

I've made the changes to the setup scripts but I'm new to using PRs and I'm not sure how to submit one. I read through the CONTRIBUTING.md explanation and here is what I did:

- Installed all of the pre-commit checks 
- Created a new branch called "features/setup-without-mjkey"
- Made the changes
- Committed the changes (all of the pre-commit checks passed)
- Tried to push the changes

But then I got a "permission denied" response from the server. Do I need permissions to submit a new branch or have I done something else wrong?

Thanks! (and sorry for my naivete on this!)

ryanjulian:
Hi Anthony,

Generally PRs are submitted from one fork to another. What you need to do is create your own fork of garage (use the fork button on GitHub), then commit and push to your account, and then submit a PR from your fork to ours.

[...]

PR #858, opened 2019-08-30
Fix indirect mujoco requirement

krzentner:
This change *actually* makes it possible to `pip install garage`
(assuming numpy==1.14.5 is already installed) without mujoco installed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ required = [
     'cloudpickle',
     'cma==1.1.06',
     'dowel==0.0.2',
-    'gym[all]==0.12.4',
+    'gym[box2d,atari]==0.12.4',
     'joblib<0.13,>=0.12',
     'matplotlib',
     'numpy==1.14.5',
@@ -47,6 +47,7 @@ extras = {}
 
 extras['mujoco'] = [
     'mujoco-py<2.1,>=2.0',
+    'gym[all]==0.12.4',
 ]
 
 extras['dm_control'] = [
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ required = [
     'cloudpickle',
     'cma==1.1.06',
     'dowel==0.0.2',
-    'gym[box2d,atari]==0.12.4',
+    'gym[atari,box2d,classic_control]==0.12.4',
     'joblib<0.13,>=0.12',
     'matplotlib',
     'numpy==1.14.5',
--- a/setup.py
+++ b/setup.py
@@ -12,6 +12,7 @@ except ImportError:
         'https://github.com/rlworkgroup/garage/issues/800 for more info.')
 
 TF_VERSION = '<1.15,>=1.14.0'
+GYM_VERSION = '==0.12.4'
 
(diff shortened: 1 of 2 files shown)


krzentner:
See the [gym setup.py](https://github.com/openai/gym/blob/master/setup.py) for why this is necessary.

setup.py
```diff
@@ -22,7 +22,7 @@
     'cloudpickle',
     'cma==1.1.06',
     'dowel==0.0.2',
-    'gym[all]==0.12.4',
+    'gym[box2d,atari]==0.12.4',
```
gitanshu: we can include classic_control too here, right? It doesn't depend on mujoco either acc to [gym setup.py](https://github.com/openai/gym/blob/master/setup.py)

gitanshu: Also maybe we could extract gym version as a constant

krzentner: Sure, I can do that. It won't do anything though, since the extra list for it is empty.

gitanshu:
There's [one place in our code](https://github.com/rlworkgroup/garage/blob/a6f696b5f48adc007b0f44fa320b9a26bf72e6b1/src/garage/envs/base.py#L89) where we import mujoco_py. We'd also need to catch the import error and tell user to install `garage[mujoco]`

krzentner:
> 
> 
> There's [one place in our code](https://github.com/rlworkgroup/garage/blob/a6f696b5f48adc007b0f44fa320b9a26bf72e6b1/src/garage/envs/base.py#L89) where we import mujoco_py. We'd also need to catch the import error and tell user to install `garage[mujoco]`

I assumed that wouldn't be necessary, since at that point we know the user has passed in a class which requires mujoco to construct.

setup.py
```diff
@@ -22,7 +23,7 @@
     'cloudpickle',
     'cma==1.1.06',
     'dowel==0.0.2',
-    'gym[all]==0.12.4',
+    'gym[atari,box2d,classic_control]' + GYM_VERSION,
```
ryanjulian: any reason you we didn't use `[all]`?

ryanjulian:
catching the ImportError would be nice, and probably at least prevent some people from filing github issues.

src/garage/envs/base.py
```diff
@@ -86,7 +86,12 @@ def _close_viewer_window(self):
                    for package in KNOWN_GYM_NOT_CLOSE_MJ_VIEWER):
                 # This import is not in the header to avoid a MuJoCo dependency
                 # with non-MuJoCo environments that use this base class.
-                from mujoco_py.mjviewer import MjViewer
+                try:
+                    from mujoco_py.mjviewer import MjViewer
+                except ImportError:
```
krzentner: @ryanjulian Is this what you had in mind? Or should I re-throw the exception?

ryanjulian: seems fair. I guess the rest is handled by gym.

gitanshu:
I guess we could close https://github.com/rlworkgroup/garage/issues/509 now, right?

Issue #853, opened 2019-08-29
Make snapshot_config a required argument


zhanpenghe:
Currently, if we don't pass a `snapshot_config` to the runner like this (sorry, I am an old school garage user so I did not catch up with the snapshotter):
```
def run_task(*_):
    with LocalTFRunner() as runner:
```
the snapshots and log data could locate at different locations SILENTLY when we specify a log directory or log prefix.

I suggest that we should make this argument (i.e. the `snapshot_config` of a local runner) required so doing this will at least raise an error. Or, alternatively, I think we should handle this by forcing the snapshotter to use the log directory of dowel and remove the `snapshot_dir` from snapshotter. What do you think?

@krzentner @naeioi @ryanjulian @zequnyu

ryanjulian:
This should either be required, or the default behavior should be sane. It should not locate the logs at different positions.

zhanpenghe:
@ryanjulian I second this too.

zequnyu:
I agree it is dangerous to have a silent bug: https://github.com/rlworkgroup/garage/pull/699#issuecomment-507883750.

It is interesting to force `snapshotter` to use `dower.logger` directory, since as I studied the source code, it's kind of hard to get `log_dir` from `logger`. For this, we might need to refactor logger's API? @zhanpenghe Also I think this way we need to think more about `snapshotter`'s default `log_dir` when we are not using experiment API (when doing tests).

Alternatively as you pointed out, we can make `config` required for Runner, and add a temporary `config` to test fixtures for the use of every test.

It's hard to foresee which one is better, but I can try both to figure out.

zequnyu:
Close by #860.

PR #860, opened 2019-09-03
Make snapshot_config required for Runner

zequnyu:
Address #853.

- make snapshot_config as required to `Runner`
- fix wrong use in `example/torch`

### DIFF
--- a/examples/torch/ddpg_pendulum.py
+++ b/examples/torch/ddpg_pendulum.py
@@ -20,9 +20,9 @@ from garage.torch.policies import DeterministicMLPPolicy
 from garage.torch.q_functions import ContinuousMLPQFunction
 
 
-def run_task(*_):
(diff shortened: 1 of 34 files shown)


naeioi:
Good job. I think making snapshot_config explicit really helps reduce confusion when setting up garage. Here I want to propose an alternative solution.

In #853 , the main concern is that snapshots are siliently saved to a path different to log. While in most cases snapshot and log should be in the same path, we now have seperate paths to configure them.

What do you think of the idea of a single, unique **workspace** for each experiment, provided by runner, that dowel and snapshottor both rely on. This way we don't have to setup logger and snapshottor independently.

zequnyu:
I think this is a design problem of the `run_experiment` API. For now let's just go with this solution. We should definitely discuss more of simplifying the `run_experiment` API as it's apparently over-complicated.

PR #861, opened 2019-09-04
Update pycma to 2.7.0

### LINKED ISSUE #800: Upgrade pycma so that garage can install without numpy
The current version of pycma we use requires numpy to run its setup.py. This breaks most automated install processes. Later versions of pycma don't have this problem, but they have an API change which requires updating `garage.np.algos.cmaes`.

krzentner:
With this change, it should be possible to `pip install garage` without
any other steps. This fixes #800.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,6 @@ ci-verify-pipenv: export VIRTUAL_ENV=
 ci-verify-pipenv:
 	pip3 install --upgrade pipenv setuptools
 	pipenv --three
-	pipenv install numpy==1.14.5  # cma is dumb and requires numpy to execute setup.py
 	pipenv install .[all,dev]
 	pipenv graph
 	# pylint will verify all imports work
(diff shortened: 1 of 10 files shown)


krzentner:
I feel like we should still require some minimal numpy version. Should I just change it to "numpy>=1.14.5"?

ryanjulian:
Pip-driven minumums are fine. The current version is probably a good guide.

I'm just suggesting you remove the install guard in setup.py which ensures that numpy be installed *before* setup.py is executed (which is not compliant with a standard pip setup). i.e. I'm suggesting we remove this: https://github.com/rlworkgroup/garage/blob/80a5951efbbdeb7557140531391213612fa218c0/setup.py#L5

krzentner:
> I'm just suggesting you remove the install guard in setup.py 
Oh right, I totally forgot that was there. I tested in a virtual env with `--no-site-packages` and no numpy installed, but that didn't trigger the guard for some reason.

Anyways, it's fixed now. If CI says this change is good, I'll submit it.

PR #864, opened 2019-09-10
Supress ray output

avnishn:
This PR is to suppress all output from sampler workers to the console. This allows us to avoid having the console filled with repeat warnings about tensorflow and mujoco, that are already outputted by the parent process that the training loop is running on. Also we removed GaussianMLPPolicy with model from the codebase, and I updated the ray sampler example to reflect this.


ryanjulian:
What happens when one of the processes has an error? Perhaps you can suppress the output from all but 1 process? Or only suppress stdout but not stderr?

avnishn:
> What happens when one of the processes has an error? Perhaps you can suppress the output from all but 1 process? Or only suppress stdout but not stderr?

@ryanjulian, I just checkout out how the output suppressing feature in ray works. It doesn't suppress errors, so errors from processes will still be outputted to the console :) .

ryanjulian:
lgtm

PR #881, opened 2019-09-16
Propose new Sampler API and port Ray Sampler

krzentner:
I'm uploading this pull request now to discuss this design, not with the intent of merging it right now.

However, I've ported the ray sampler to the new design to demonstrate that this port is not difficult from the sampler side. The ray unit test still passes (after also being ported to the new API).

### DIFF
--- a/src/garage/sampler/ray_sampler.py
+++ b/src/garage/sampler/ray_sampler.py
@@ -13,66 +13,96 @@ import numpy as np
 import psutil
 import ray
 
-from garage.experiment import deterministic
 from garage.misc.prog_bar_counter import ProgBarCounter
-from garage.sampler.base import BaseSampler
+from garage.sampler.sampler import Sampler
 
 
-class RaySampler(BaseSampler):
+class RaySampler(Sampler):
     """Collects Policy Rollouts in a data parallel fashion.
 
     Args:
-        algo (garage.np.algo.RLAlgorithm): A garage algo object
-        env (gym.Env): A gym/akro env object
-        seed (int): Random seed.
-        should_render (bool): If the sampler render the trajectories.
-        num_processors (int): Number of processors to be used.
-        sampler_worker_cls (garage.sampler.ray_sampler.SamplerWorker):
-            If none, uses the default SamplerWorker class
+        worker_factory(garage.sampler.WorkerFactory): Used for worker behavior.
+        agents(list[garage.Policy]): Agents to distribute across workers.
+        envs(list[gym.Env]): Environments to distribute across workers.
+        num_processors(int): Number of workers processes to spawn. If none,
+            defaults to number of physical CPU cores.
+        sampler_worker_cls(type): If None, uses the default SamplerWorker
+            class.
 
     """
 
     def __init__(self,
-                 algo,
-                 env,
-                 seed,
-                 should_render=False,
+                 worker_factory,
+                 agents,
+                 envs,
                  num_processors=None,
                  sampler_worker_cls=None):
-        super().__init__(algo, env)
+        # pylint: disable=super-init-not-called
         self._sampler_worker = ray.remote(SamplerWorker if sampler_worker_cls
                                           is None else sampler_worker_cls)
-        self._seed = seed
-        deterministic.set_seed(self._seed)
-        self._max_path_length = self.algo.max_path_length
-        self._should_render = should_render
+        self._worker_factory = worker_factory
+        self._agents = agents
+        self._envs = self._worker_factory.prepare_worker_messages(envs)
         if not ray.is_initialized():
             ray.init(log_to_driver=False)
         self._num_workers = (num_processors if num_processors else
                              psutil.cpu_count(logical=False))
         self._all_workers = defaultdict(None)
-        self._idle_worker_ids = list(range(self._num_workers))
-        self._active_worker_ids = []
+        self._workers_started = False
+        self.start_worker()
+
+    @classmethod
+    def from_worker_factory(cls, worker_factory, agents, envs):
+        """Construct this sampler.
+
+        Args:
+            worker_factory(WorkerFactory): Pickleable factory for creating
+                workers. Should be transmitted to other processes / nodes where
+                work needs to be done, then workers should be constructed
+                there.
+            agents(Agent or List[Agent]): Agent(s) to use to perform rollouts.
+                If a list is passed in, it must have length exactly
+                `worker_factory.n_workers`, and will be spread across the
+                workers.
+            envs(gym.Env or List[gym.Env]): Environment rollouts are performed
+                in. If a list is passed in, it must have length exactly
+                `worker_factory.n_workers`, and will be spread across the
+                workers.
+
+        Returns:
+            Sampler: An instance of `cls`.
+
+        """
+        return cls(worker_factory, agents, envs)
 
     def start_worker(self):
         """Initialize a new ray worker."""
-        # Pickle the environment once, instead of once per worker.
-        env_pkl = pickle.dumps(self.env)
-        # We need to pickle the agent so that we can e.g. set up the TF Session
+        if self._workers_started:
+            return
+        self._workers_started = True
+        # We need to pickle the agent so that we can e.g. set up the TF.Session
         # in the worker *before* unpicling it.
-        agent_pkl = pickle.dumps(self.algo.policy)
+        agent_pkls = self._worker_factory.prepare_worker_messages(
+            self._agents, pickle.dumps)
         for worker_id in range(self._num_workers):
             self._all_workers[worker_id] = self._sampler_worker.remote(
-                worker_id, env_pkl, agent_pkl, self._seed,
-                self._max_path_length, self._should_render)
+                worker_id, self._envs[worker_id], agent_pkls[worker_id],
+                self._worker_factory)
 
-    # pylint: disable=arguments-differ
-    def obtain_samples(self, itr, num_samples):
+    def obtain_samples(self, itr, num_samples, agent_update, env_update=None):
         """Sample the policy for new trajectories.
 
         Args:
-            itr (int): Iteration number.
-            num_samples (int): Number of steps the the sampler should collect.
+            itr(int): Iteration number.
+            num_samples(int): Number of steps the the sampler should collect.
+            agent_update(object): Value which will be passed into the
+                `agent_update_fn` before doing rollouts. If a list is passed
+                in, it must have length exactly `factory.n_workers`, and will
+                be spread across the workers.
+            env_update(object): Value which will be passed into the
+                `env_update_fn` before doing rollouts. If a list is passed in,
+                it must have length exactly `factory.n_workers`, and will be
+                spread across the workers.
 
         Returns:
             list[dict]: Sample paths, each path with key
@@ -83,8 +113,9 @@ class RaySampler(BaseSampler):
                 * env_infos: (dict)
 
         """
-        _active_workers = []
-        self._active_worker_ids = []
+        active_workers = []
+        active_worker_ids = []
+        idle_worker_ids = list(range(self._num_workers))
         pbar = ProgBarCounter(num_samples)
         completed_samples = 0
         traj = []
@@ -92,12 +123,16 @@ class RaySampler(BaseSampler):
 
         # update the policy params of each worker before sampling
         # for the current iteration
-        curr_policy_params = self.algo.policy.get_param_values()
-        params_id = ray.put(curr_policy_params)
-        while self._idle_worker_ids:
-            worker_id = self._idle_worker_ids.pop()
+        idle_worker_ids = list(range(self._num_workers))
+        param_ids = self._worker_factory.prepare_worker_messages(
+            agent_update, ray.put)
+        env_ids = self._worker_factory.prepare_worker_messages(
+            env_update, ray.put)
+        while idle_worker_ids:
+            worker_id = idle_worker_ids.pop()
             worker = self._all_workers[worker_id]
-            updating_workers.append(worker.set_agent.remote(params_id))
+            updating_workers.append(
+                worker.update.remote(param_ids[worker_id], env_ids[worker_id]))
 
         while completed_samples < num_samples:
             # if there are workers still being updated, check
@@ -109,26 +144,26 @@ class RaySampler(BaseSampler):
                                                      num_returns=1,
                                                      timeout=0.1)
                 upd = [ray.get(up) for up in updated]
-                self._idle_worker_ids.extend(upd)
+                idle_worker_ids.extend(upd)
 
             # if there are idle workers, use them to collect trajectories
             # mark the newly busy workers as active
-            while self._idle_worker_ids:
-                idle_worker_id = self._idle_worker_ids.pop()
-                self._active_worker_ids.append(idle_worker_id)
+            while idle_worker_ids:
+                idle_worker_id = idle_worker_ids.pop()
+                active_worker_ids.append(idle_worker_id)
                 worker = self._all_workers[idle_worker_id]
-                _active_workers.append(worker.rollout.remote())
+                active_workers.append(worker.rollout.remote())
 
             # check which workers are done/not done collecting a sample
             # if any are done, send them to process the collected trajectory
             # if they are not, keep checking if they are done
-            ready, not_ready = ray.wait(_active_workers,
+            ready, not_ready = ray.wait(active_workers,
                                         num_returns=1,
                                         timeout=0.001)
-            _active_workers = not_ready
+            active_workers = not_ready
             for result in ready:
-                trajectory, num_returned_samples = self._process_trajectory(
-                    result)
+                trajectory, num_returned_samples = _process_trajectory(
+                    result, active_worker_ids, idle_worker_ids)
                 completed_samples += num_returned_samples
                 pbar.inc(num_returned_samples)
                 traj.append(trajectory)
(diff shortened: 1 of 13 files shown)


src/garage/sampler/base.py
```diff
@@ -1,43 +1,48 @@
...
+    def __init__(self, algo, env):
+        """Construct a Sampler from an Algorithm.
+
+        Using this method is deprecated.
+        """
+        self.algo = algo
+        self.env = env
+
+    @classmethod
+    def construct(cls, config: SamplerConfig, agent, env) -> 'Sampler':
```
avnishn: if we have this construct that takes a config, can we get rid of passing the algo to the init @krzentner

src/garage/sampler/base.py
```diff
@@ -1,43 +1,48 @@
...
         """
-        pass
 
     @abc.abstractmethod
-    def obtain_samples(self, itr):
-        """Collect samples for the given iteration number.
-
-        Args:
-            itr (int): Iteration number.
+    def obtain_samples(self, itr, num_samples, agent_update, env_update=None):
```
avnishn: can you write in the comments what agent_update and env_update mean

src/garage/sampler/base.py
avnishn: NVM

src/garage/sampler/base.py
avnishn: are these functions for updating the env and updating the agent? if so can we rename to agent_update_fn and env_update_fn

krzentner: These are data that will be passed into the `agent_update_fn` and `env_update_fn`. I clearly need to write more documentation for these and how they are used.

The general idea here is that we can specify one config type that works for a large number of algorithms, and then only need to specify the data to feed that config. For example, all Tensorflow policies will use the same `worker_init_fn` and `agent_update_fn`, so we can define a SamplerConfig that has the right functions for those, then specify the parameter values in each algorithm (or in one of the Algorithm base classes).

src/garage/sampler/base.py
krzentner: Yeah, the goal of this change is actually to let us avoid the `Sampler`s ever having direct access to `Algorithm`s. They only get `SamplerConfig`s.

src/garage/sampler/ray_sampler.py
```diff
@@ -155,91 +160,30 @@ def _process_trajectory(self, result):
...
-        """Set the agent params.
-
-        Args:
-            - flattened_params(): model parameters in numpy format
-        """
-        self.agent.set_param_values(flattened_params)
-        self.agent_updates += 1
+        self.config = config
+
+    def update(self, agent_update, env_update):
```
krzentner: @avnishn This function describes fair succinctly the relationship between the sampler, `config`, `agent_update_fn`, `env_update_fn`, `agent`, `agent_update`, `env`, and `env_update`.

src/garage/sampler/base.py
krzentner: > ... if the init is deprecated, are you saying that construct is going to be the new init of the sampler?

The intent is actually to have algorithms ask the runner to create samplers. This is because the runner generally knows the correct / default values for `init_worker_fn` and `update_agent_fn`. But yeah, the runner will construct Samplers by calling `construct`, at least for now. In the future, we *could* refactor things so that `__init__` is used by the runner instead, but we might not want to do that, since there's a general aversion to making `__init__` part of APIs.

src/garage/sampler/base.py
krzentner: Done.

src/garage/sampler/base.py
```diff
@@ -1,43 +1,71 @@
...
+    def __init__(self, algo, env):
+        """Construct a Sampler from an Algorithm.
+
+        Calling this method is deprecated.
+        """
+        self.algo = algo
+        self.env = env
+
+    @classmethod
+    def construct(cls, config, agent, env):
```
ryanjulian: do you envision replacing the constructor with this after the API change is finished?

krzentner:
> Would this design allow us to get rid of framework-specific samplers classes (e.g. those in tf/samplers)?

I designed it with that in mind. For the case of off-policy algorithms, we still need a better way of incorporating exploration strategies, but that could be handled in this design too (by using a different rollout function).

> How well would this generalize to a multi-environment sampler, assuming we don't want to just wrap multiple environments in one round-robin environment? Do you have some thoughts on what that would look like?

The environment update function provides a reasonably simple way of handling multiple environments. It receives the worker number, which means we could also handle different workers managing different environments for efficiency. The only major assumption of this design which might introduce efficiency or ergonomics issues is that we transmit the same environment update to all workers. I don't think this is a major concern, and there's ways we can work around it if necessary.

>  Are there better names than SamplerConfig? Perhaps not?

The main reason I don't like the term config here is that this object really provides additional behavior (mostly) not data. This pattern (of having an object with a bunch of functions in it to act like a poor-man's trait pattern) has been used in other places. In JavaScript, it's sometimes called the [revealing constructor pattern](https://blog.domenic.me/the-revealing-constructor-pattern/) (although JavaScript has object literals, making it somewhat more natural). We could also think of this as injecting new behavior into the Sampler, since it decouples the Sampler from the agent and environment types. Despite these related techniques, I don't really know what else to call this. `SamplerExtension`, `SamplerBehavior`, and `SamplerOperations` are reasonable options, I suppose, but not better than `SamplerConfig`. We could focus more on how this object interacts with workers, and call it a `WorkerConfig`, `UnderlyingWorker`, `WorkerBehavior`, or `WorkerImpl`.

Alternatively, we could turn this more into a "proper" worker abstraction, albeit one that doesn't know how to distribute itself. Then we could think of this as an inversion of control in the worker interface (and call this a `Worker`, `InnerWorker`, `WorkerImpl`, or `WorkerCore`). I'll think about these options a little more.

src/garage/sampler/base.py
krzentner: Maybe? Having the constructor be part of the API is generally avoided in languages where types aren't first class (since it's not meaningful there), so it's considered somewhat unusual in other languages. However, it seems fairly natural in Python. I don't have a strong preference here.

src/garage/sampler/base.py
```diff
@@ -1,43 +1,79 @@
+"""Base sampler class."""
+
 import abc
+import copy
 
 
 class Sampler(abc.ABC):
-    @abc.abstractmethod
+    """Base class of all samplers."""
```
ryanjulian: Definitely a TODO here is to describe the sampler design in a little more detail

src/garage/sampler/base.py
```diff
@@ -1,43 +1,79 @@
...
+
+    @classmethod
+    def construct(cls, config, agents, envs):
+        """Construct this sampler from a config.
+
+        Args:
+            config(SamplerConfig): Configuration which specifies how to
+                intialize workers, update agents and environments, and perform
+                rollouts.
+            agents(Agent or [Agent]): Agent(s) to use to perform rollouts. It
```
ryanjulian: we try to use PEP484-ish notation here, so it would be `Agent or List[Agent]`

src/garage/sampler/config.py
```diff
@@ -0,0 +1,217 @@
...
+
+def identity_function(x):
+    """Do nothing.
+
+    This function can exists so it can be pickled.
+    """
+    return x
+
+
+class SamplerConfig:
```
ryanjulian: SamplingStrategy? SamplingPlan?

src/garage/sampler/config.py
```diff
@@ -0,0 +1,217 @@
...
+
+from collections import defaultdict
+
+import numpy as np
+import psutil
+
+from garage.experiment import deterministic
+
+
+def rollout(config, worker_number, agent, env, max_path_length):
```
ryanjulian: default_rollout_fn?

src/garage/sampler/config.py
```diff
@@ -0,0 +1,217 @@
...
+
+    Args:
+        seed(int): The seed to use to intialize random number generators.
+        n_workers(int): The number of workers to use.
+        max_path_length(int): The maximum length paths which will be sampled.
+        worker_init_fn((int, SamplerConfig) -> None): Function to run in
+            workers before constructing any agents or environments in them.
+            The first argument should be the worker number, and the second
+            argument should be this config.
+        env_update_fn((int, gym.Env, object) -> gym.Env): Function to run to
```
ryanjulian: why include the current agent and the current environment as arguments to `env_update_fn`/`agent_update_fn`? presumably they either already have these or don't care.

is your intent to avoid the sentinel case?

src/garage/sampler/config.py
```diff
@@ -0,0 +1,217 @@
...
+    functions to this one). Not that this object must be picklable, since it
+    may be passed to workers. However, its fields individually need not be.
+
+    All arguments to this type must be passed by keyword.
+
+    Args:
+        seed(int): The seed to use to intialize random number generators.
+        n_workers(int): The number of workers to use.
+        max_path_length(int): The maximum length paths which will be sampled.
+        worker_init_fn((int, SamplerConfig) -> None): Function to run in
```
ryanjulian: worker number is an odd thing to be carrying around/passing everywhere, but I understand the intent. passing `SamplerConfig` to `worker_init_fn` and `rollout`, sometimes in addition to `worker_number` seems an odd asymmetry to me.

more importantly, who gets a `SamplerConfig` and who gets a `worker_number` (or both) i find a little hard to predict from just reading the API.

what if you split the roles here into `SamplerConfig` which is global/initialization and `SamplingContext` which is generated by `Sampler` and created/managed per-worker. for instance, the context could include the random seed this worker should use (rather than expecting the worker to derive it from the global seed), the worker number, perhaps which resources this worker is allowed to use, etc.

src/garage/sampler/config.py
krzentner: Those functions definitely don't otherwise have access to those objects, since those objects might have been created (e.g. by unpickling) after the functions were. I think if you read how the ray sampler uses these functions it's fairly obvious why they need to be passed in.

src/garage/sampler/config.py
krzentner: WorkerFactory? The name is still up for discussion, of course.

src/garage/sampler/config.py
krzentner: I ended up doing something like this (`WorkerFactory` and `Worker`). I do think it makes sense to rename these though, since they're not exactly workers in the sense most people probably expect.

src/garage/sampler/base.py
```diff
@@ -1,43 +1,80 @@
...
-        """Terminate workers if necessary."""
-        pass
+    def start_worker(self):
+        """Initialize the sampler.
 
+        i.e. launching parallel workers if necessary.
 
-class BaseSampler(Sampler):
-    """Base class for sampler.
+        This method is deprecated, please launch workers in construct instead.
```
ryanjulian: perhaps you can raise a warning here for some extra nagging?

src/garage/sampler/base.py
```diff
@@ -1,43 +1,80 @@
+"""Base sampler class."""
```
ryanjulian: i want to start moving away from `base.py` everywhere. perhaps you can put this in `sampler.py`?

src/garage/sampler/ray_sampler.py
```diff
@@ -9,63 +9,71 @@
...
 
-class RaySampler(BaseSampler):
+class RaySampler(Sampler):
     """Collects Policy Rollouts in a data parallel fashion.
 
     Args:
         - algo: A garage algo object
-        - env: A gym/akro env object
-        - should_render(bool): should the sampler render the trajectories
+        - agents: Agent(s) to run.
```
ryanjulian: might as well update this docstring to be compliant (i.e. no - bullets, types)

src/garage/sampler/ray_sampler.py
```diff
@@ -155,91 +166,27 @@ def _process_trajectory(self, result):
...
-    trajectories or rollouts.
-
     Args:
-        - worker_id(int): the id of the sampler_worker
-        - env_pkl: A pickled gym or akro env object
-        - agent_pkl: A pickled agent
-        - max_path_length(int): max trajectory length
-        - should_render(bool): if true, renders trajectories after
-            sampling them
+        - worker_id(int): The id of the sampler_worker
```
ryanjulian: fix docstrings

src/garage/sampler/worker.py
```diff
@@ -0,0 +1,155 @@
...
+
+import numpy as np
+
+from garage.experiment import deterministic
+
+
+class Worker(abc.ABC):
+    """Worker class used in all Samplers."""
+
+    def __init__(self, seed, max_path_length, worker_number):
```
ryanjulian: this is a nit, but the order i would probably expect in this API for `worker_number` to come first.

+1 making seed required.  I think that this is a good practice for reproducibility because it forces the implementer to at least choose a seed to pass (instead of silent repro bugs because of a missing seed).

src/garage/sampler/worker.py
```diff
@@ -0,0 +1,155 @@
...
+    """Initialize a worker.
+
+    Args:
+        seed(int): The seed to use to intialize random number generators.
+        max_path_length(int): The maximum length paths which will be sampled.
+        worker_number(int): The number of the worker this update is
+            occurring in. This argument is used to set a different seed for
+            each worker.
+
+    Should create fields the following fields:
```
ryanjulian: typo

src/garage/sampler/worker.py
```diff
@@ -0,0 +1,155 @@
...
+        agent(Policy or None): The worker's initial agent.
+        env(gym.Env or None): The worker's environment.
+    """
+
+    def __init__(self, seed, max_path_length, worker_number):
+        self.seed = seed
```
ryanjulian: can most of these be _private?

src/garage/sampler/worker.py
```diff
@@ -0,0 +1,155 @@
...
+
+        Args:
+            worker_number(int): The number of the worker this update is
+                occurring in. This argument is used to set a different seed for
+                each worker.
+        """
+        deterministic.set_seed(self.seed + self.worker_number)
+
+    def update_agent(self, agent_update):
+        """Update an agent, assuming it implements garage.tf.policies.Policy2.
```
ryanjulian: i think that `Policy` will probably be deleted and replaced by `Policy2` by the time you merge this.

src/garage/sampler/worker.py
```diff
@@ -0,0 +1,155 @@
...
+        Args:
+            env_update(gym.Env or None): The environment to replace the
+                existing env with.
+
+        Returns:
+            The updated environment.
+
+        """
+        if env_update is not None:
+            self.env = env_update
```
ryanjulian: you may need to call `env.close()` on the old env for this it be properly garbage-collected.

src/garage/sampler/worker_factory.py
```diff
@@ -0,0 +1,80 @@
...
+    """
+
+    def __init__(
+            self,
+            *,  # Require passing by keyword.
+            seed,
+            max_path_length,
+            n_workers=psutil.cpu_count(logical=False),
+            worker_class=DefaultWorker):
+        self.seed = seed
```
ryanjulian: can most of these by _private?

src/garage/sampler/worker_factory.py
```diff
@@ -0,0 +1,80 @@
...
+                of length self.n_workers.
+            preprocess(function): Function to call on each single object before
+                creating the list.
+
+        Returns:
+            List[object]: A list of length self.n_workers.
+
+        """
+        if isinstance(objs, list):
+            print('objs', objs)
```
ryanjulian: did you forget to delete this?

src/garage/sampler/worker_factory.py
```diff
@@ -0,0 +1,80 @@
...
+            if len(objs) != self.n_workers:
+                raise ValueError(
+                    "Length of list doesn't match number of workers")
```
ryanjulian: nit: single-quotes

src/garage/sampler/worker_factory.py
```diff
@@ -0,0 +1,80 @@
...
+            seed,
+            max_path_length,
+            n_workers=psutil.cpu_count(logical=False),
+            worker_class=DefaultWorker):
+        self.seed = seed
+        self.n_workers = n_workers
+        self.max_path_length = max_path_length
+        self.worker_class = worker_class
+
+    def get_worker_broadcast(self, objs, preprocess=identity_function):
```
ryanjulian: i was initially confused by the name of this API because it suggested to me that it's going to return a `Worker`. perhaps it can just be `broadcast` or `broadcast_to_n_workers`?

ryanjulian:
Overall this is looking great and I think a definite improvement over what we currently have.

What are your thoughts on a rollout strategy? Should we this now and refactor the existing as we go? Or run one giant refactor PR? I look a look at our sampler coverage and it's pretty OK, so I think either could be viable options.

src/garage/sampler/worker.py
krzentner: I'm just gonna make them require being passed by keyword. They're all integers, so they're easy to mix up.

src/garage/sampler/base.py
krzentner: I don't think that's quite warranted yet, since it's going to nag all the time.

krzentner:
> What are your thoughts on a rollout strategy? Should we this now and refactor the existing as we go? Or run one giant refactor PR? I look a look at our sampler coverage and it's pretty OK, so I think either could be viable options.
Definitely several PRs. We need to at least make the other samplers minimally follow this API before we refactor the runner and algorithms (which is the *real* target of this refactoring).

src/garage/sampler/ray_sampler.py
```diff
@@ -9,68 +9,76 @@
...
 
     Args:
-        - algo: A garage algo object
-        - env: A gym/akro env object
-        - should_render(bool): should the sampler render the trajectories
-        - sampler_worker_cls: If none, uses the default SamplerWorker
-            class
+        worker_factory(garage.sampler.WorkerFactory): Used for worker behavior.
+        agents(list[garage.Policy]): Agents to distribute across workers.
+        envs(gym.Env): Environments to distribute across workers.
```
ryanjulian: `list[gym.Env]`?

src/garage/sampler/sampler.py
```diff
@@ -0,0 +1,91 @@
...
+    def __init__(self, algo, env):
+        """Construct a Sampler from an Algorithm.
+
+        Calling this method is deprecated.
+        """
+        self.algo = algo
+        self.env = env
+
+    @classmethod
+    def construct(cls, worker_factory, agents, envs):
```
ryanjulian: consider naming this `from_worker_factory`, which is a convention for non-default constructors in python. I suppose is could be `from_worker_factor_agents_and_envs` but that seems a little excessive.

src/garage/sampler/worker.py
```diff
@@ -0,0 +1,215 @@
...
+
+class Worker(abc.ABC):
+    """Worker class used in all Samplers."""
+
+    def __init__(self, seed, max_path_length, worker_number):
+        """Initialize a worker.
+
+        Args:
+            seed(int): The seed to use to intialize random number generators.
+            max_path_length(int): The maximum length paths which will be
```
ryanjulian: is `inf` a valid input here? what if I never want to stop (e.g. some off-policy envs)?

src/garage/sampler/worker.py
```diff
@@ -0,0 +1,215 @@
...
+                sampled.
+            worker_number(int): The number of the worker this update is
+                occurring in. This argument is used to set a different seed for
+                each worker.
+
+        Should create fields the following fields:
```
ryanjulian: typo.

I think you can use an `Attributes:` section to document these

src/garage/sampler/sampler.py
```diff
@@ -0,0 +1,91 @@
...
+                it must have length exactly `factory.n_workers`, and will be
+                spread across the workers.
+
+        Returns:
+            List[Dict[str, np.array]]: A list of paths.
+
+        """
+
+    @abc.abstractmethod
+    def shutdown_worker(self):
```
ryanjulian: is this similarly-deprecated?

src/garage/sampler/worker.py
```diff
@@ -0,0 +1,215 @@
...
+        """Update the worker's agent, using agent_update."""
+
+    def update_env(self, env_update):
+        """Update the worker's env, using env_update."""
+
+    def rollout(self):
+        """Sample a single rollout of the agent in the environment.
+
+        Returns:
+            observations(np.array): Non-flattened array of observations.
```
ryanjulian: please include shape information (or make references to the env_spec)

src/garage/sampler/worker.py
```diff
@@ -0,0 +1,215 @@
...
+            agent_infos(Dict[str, np.array]): Dictionary of stacked,
+                non-flattened `agent_info`s.
+            env_infos(Dict[str, np.array]): Dictionary of stacked,
+                non-flattened `env_info`s.
+
+        """
+
+    def __getstate__(self):
+        """Refuse to be pickled."""
+        raise ValueError('Workers are not pickleable. '
```
ryanjulian: nice

src/garage/sampler/worker.py
```diff
@@ -0,0 +1,215 @@
...
+
+        Args:
+            worker_number(int): The number of the worker this update is
+                occurring in. This argument is used to set a different seed for
+                each worker.
+        """
+        deterministic.set_seed(self._seed + self._worker_number)
+
+    def update_agent(self, agent_update):
+        """Update an agent, assuming it implements garage.tf.policies.Policy.
```
ryanjulian: `garage.Policy`?

src/garage/sampler/worker.py
```diff
@@ -0,0 +1,215 @@
...
+
+        Args:
+            agent_update(Dict[str, np.array]): Parameters to agent, which
+                should have been generated by calling
+                `policy.get_param_values`.
```
ryanjulian: i think Anson was planning on deleting `get_param_values` since `Parameterized` no longer exists. I guess this will be whatever ends up in the interface `garage.Policy`

src/garage/sampler/worker.py
```diff
@@ -0,0 +1,215 @@
...
+                occurring in. This argument is used to set a different seed for
+                each worker.
+        """
+        deterministic.set_seed(self._seed + self._worker_number)
+
+    def update_agent(self, agent_update):
+        """Update an agent, assuming it implements garage.tf.policies.Policy.
+
+        Args:
+            agent_update(Dict[str, np.array]): Parameters to agent, which
```
ryanjulian: you specified `object` in the interface, so does this `DefaultWorker` specifically do agent updates with dicts?

src/garage/sampler/worker.py
```diff
@@ -0,0 +1,215 @@
...
+        """Use any non-None env_update as a new environment.
+
+        A simple env update function. If env_update is not None, it should be
+        the complete new environment.
+
+        This allows changing environments by passing the new environment as
+        `env_update` into `obtain_samples`.
+
+        Args:
+            env_update(gym.Env or None): The environment to replace the
```
ryanjulian: presumably, for some other worker, this could actually pass data which updates an env which you won't throw away?

src/garage/sampler/worker.py
```diff
@@ -0,0 +1,215 @@
...
+
+        Returns:
+            The updated environment.
+
+        """
+        if env_update is not None:
+            self.env.close()
+            self.env = env_update
+
+    def start_rollout(self):
```
ryanjulian: if there a reason for the following 3 methods to be public? though I suppose it's a little moot because these objects are usually remote, anyway.

src/garage/sampler/worker.py
```diff
@@ -0,0 +1,215 @@
...
+        else:
+            return True
+
+    def collect_rollout(self):
+        """Collect the current rollout, clearing the internal buffer.
+
+        Returns:
+            observations(np.array): Non-flattened array of observations.
+            actions(np.array): Non-flattened array of actions.
+            rewards(np.array): Array of rewards of shape (timesteps, 1).
```
ryanjulian: use shape documentation here or reference the EnvSpec

src/garage/sampler/worker.py
```diff
@@ -0,0 +1,215 @@
...
+        return (np.array(observations), np.array(actions), np.array(rewards),
+                dict(agent_infos), dict(env_infos))
+
+    def rollout(self):
+        """Sample a single rollout of the agent in the environment.
+
+        Returns:
+            observations(np.array): Non-flattened array of observations.
+            actions(np.array): Non-flattened array of actions.
+            rewards(np.array): Array of rewards of shape (timesteps, 1).
```
ryanjulian: Use shape documentation here or use the EnvSpec

ryanjulian:
@avnishn using the "request changes" button blocks this PR until you approve the updated PR. please re-review or remove your request.

src/garage/sampler/worker.py
krzentner: Yep. I've added a comment.

src/garage/sampler/worker.py
krzentner: Yep. I've added a comment.

src/garage/sampler/worker.py
krzentner: Yeah, some samplers may want to single-step workers, for various reasons, so this is intentionally part of the public `Worker` API.

Issue #653, opened 2019-05-12
Ensure that the CI works on PRs from forks


ryanjulian:

ryanjulian:
@gitanshu this is fixed, right?

gitanshu:
Closed by #1220

PR #885, opened 2019-09-17
Deterministic rollout

jonasnm:
Adding an option for taking deterministic actions in rollouts. This can be useful for testing the trained policies.

### DIFF
--- a/src/garage/sampler/utils.py
+++ b/src/garage/sampler/utils.py
@@ -11,7 +11,18 @@ def rollout(env,
             max_path_length=np.inf,
             animated=False,
             speedup=1,
-            always_return_paths=False):
+            always_return_paths=False,
+            deterministic=False):
+    """ Generate a sample from a policy.
+
+    Args:
+        deterministic (bool): Boolean variable indicating whether a
+        stochastic or deterministic action should be taken during the
+        rollout. This is False (stochastic actions) by default.
+
+    Returns:
+
+    """
     observations = []
     actions = []
     rewards = []
@@ -24,6 +35,8 @@ def rollout(env,
         env.render()
     while path_length < max_path_length:
         a, agent_info = agent.get_action(o)
+        if deterministic:
+            a = agent_info['mean']
         next_o, r, d, env_info = env.step(a)
         observations.append(env.observation_space.flatten(o))
         rewards.append(r)


ryanjulian:
@krzentner what do you think?

krzentner:
I think this is generally a good idea. This implementation does assume that `agent_info` contains "mean", which we've never guaranteed. This is probably always the case for stochastic policies right now, and I believe would be fairly easy to guarantee in the future (since both `torch.distributions.Distribution`and `tf.distributions.Distribution` have `mean` fields).

It's probably also worth mentioning that in the near future I intend to merge this rollout function with the rollout functions used by the samplers.

src/garage/sampler/utils.py
```diff
@@ -11,7 +11,8 @@ def rollout(env,
             max_path_length=np.inf,
             animated=False,
             speedup=1,
-            always_return_paths=False):
+            always_return_paths=False,
+            deterministic=False):
```
krzentner: Please document this new parameter (you don't need to document the old parameters).

src/garage/sampler/utils.py
```diff
@@ -24,7 +35,13 @@ def rollout(env,
...
     while path_length < max_path_length:
         a, agent_info = agent.get_action(o)
-        next_o, r, d, env_info = env.step(a)
+
+        if deterministic == False:
+            action = a
+        else:
+            action = agent_info['mean']
+
+        next_o, r, d, env_info = env.step(action)
```
thomasjo: Seems like this block of code can be simplified to make it a bit more pythonic.

```python
if deterministic:
    a = agent_info['mean']

next_o, r, d, env_info = env.step(a)
```

One possible benefit of this change is that it fixes what seems like a bug below, where the "old" action `a` is appended to the action space instead of the "new" one.

src/garage/sampler/utils.py
```diff
@@ -49,7 +66,6 @@ def rollout(env,
         env_infos=tensor_utils.stack_tensor_dict_list(env_infos),
     )
 
-
```
thomasjo: You should undo this whitespace removal.

ryanjulian:
ref https://github.com/rlworkgroup/garage/issues/653

ryanjulian:
@krzentner PTAL

ryanjulian:
@jonasnm I'm happy to merge this once the CI is passing. You will likely get CI errors about a missing MuJoCo license (which we will ignore).

You may find it easier to pass the commit message check by squashing all of your commits into one and rewriting the commit message to satisfy the style guide.

See:
https://github.com/rlworkgroup/garage/blob/master/CONTRIBUTING.md#commit-message-format
https://medium.com/@slamflipstrom/a-beginners-guide-to-squashing-commits-with-git-rebase-8185cf6e62ec

jonasnm:
> @jonasnm I'm happy to merge this once the CI is passing. You will likely get CI errors about a missing MuJoCo license (which we will ignore).
> 
> You may find it easier to pass the commit message check by squashing all of your commits into one and rewriting the commit message to satisfy the style guide.
> 
> See:
> https://github.com/rlworkgroup/garage/blob/master/CONTRIBUTING.md#commit-message-format
> https://medium.com/@slamflipstrom/a-beginners-guide-to-squashing-commits-with-git-rebase-8185cf6e62ec

Cool! I will fix it today. Sorry for the delay.

krzentner:
@jonasnm Thanks for getting this into a good state. The CI is a bit too strict in this case, so I'll take it from here.

krzentner:
This has effectively been merged in #896.

thomasjo:
@jonasnm Congratulations on getting your first pull request merged! :tada:

Issue #812, opened 2019-07-22
Fix the CI for active release branches


ryanjulian:
Our release branches fail the CI because we have some infrastructure which assumes that all CI jobs are based on master. If we fix these assumptions, we can keep the CI running on active release branches to make sure backports and the like are still working.

gitanshu:
Closed by #889

PR #889, opened 2019-09-20
Fix CI for branch builds; Update flake8-docstrings

gitanshu:
Currently check_precommit.sh fails for release branches because it checks all commits that are different between master and branch.

However, not all commits in master follow the commit message guidelines since those messages are made through github squash and merge interface. Also, there's no point to checking commits that are already in master.

Unlike suggested in https://github.com/rlworkgroup/garage/issues/812, this PR still assumes that branches (hence CI jobs) are based off master, but I think it's a valid assumption for out workflow.

This commit modifies check_precommit.sh to pick only commits in the branch that are not in master, by taking the commits in branch that were made after the last common (merge-base) commit of master and branch.

Test run in a branch cut from release-2019.02:
5d4ce23
https://travis-ci.com/rlworkgroup/garage/builds/128575672

This PR also updates flake8-docstrings to current functioning version 1.4.0 (1.3.0 breaks current active release branches too), and updates and restricts pydocstyle to 4.0.x

### DIFF
--- a/scripts/travisci/check_precommit.sh
+++ b/scripts/travisci/check_precommit.sh
@@ -24,7 +24,8 @@ if [[ "${TRAVIS_PULL_REQUEST}" != "false" && "${TRAVIS}" == "true" ]]; then
 else
   git remote set-branches --add origin master
   git fetch
-  pre-commit run --source origin/master --origin ${TRAVIS_BRANCH}
+  merge_base=$(git merge-base origin/master ${TRAVIS_BRANCH})
+  pre-commit run --source ${merge_base} --origin ${TRAVIS_BRANCH}
   status="$((${status} | ${?}))"
 
   # Check commit messages
@@ -41,7 +42,7 @@ else
       cat "${commit_msg}"
     fi
 
-  done < <(git rev-list origin/master..."${TRAVIS_BRANCH}")
+  done < <(git rev-list ^origin/master "${TRAVIS_BRANCH}")
 fi
 
 exit "${status}"
--- a/setup.py
+++ b/setup.py
@@ -60,15 +60,13 @@ extras['dev'] = [
     # Please keep alphabetized
     'baselines @ https://api.github.com/repos/openai/baselines/tarball/f2729693253c0ef4d4086231d36e0a4307ec1cb3',  # noqa: E501
     'flake8',
-    'flake8-docstrings==1.3.0',
+    'flake8-docstrings==1.4.0',
     'flake8-import-order',
     'gtimer',
     'pandas',
     'pep8-naming==0.7.0',
     'pre-commit',
-    # pydocstyle 4.0.0 breaks flake8-docstrings 1.3.0
-    # See https://gitlab.com/pycqa/flake8-docstrings/issues/36
-    'pydocstyle<4.0.0',
+    'pydocstyle<4.1,>=4.0',
     'pylint==1.9.2',
     'pytest>=3.6',  # Required for pytest-cov on Python 3.6
     'pytest-cov',


setup.py
```diff
@@ -60,14 +60,12 @@
...
-    'flake8-docstrings==1.3.0',
+    'flake8-docstrings==1.4.0',
     'flake8-import-order',
     'pandas',
     'pep8-naming==0.7.0',
     'pre-commit',
-    # pydocstyle 4.0.0 breaks flake8-docstrings 1.3.0
-    # See https://gitlab.com/pycqa/flake8-docstrings/issues/36
-    'pydocstyle<4.0.0',
+    'pydocstyle<4.1,>=4.0',
```
krzentner: Shouldn't this be `<4.1,>4.0.0`?

scripts/travisci/check_precommit.sh
```diff
@@ -41,7 +42,7 @@ else
       cat "${commit_msg}"
     fi
 
-  done < <(git rev-list origin/master..."${TRAVIS_BRANCH}")
+  done < <(git rev-list ^origin/master "${TRAVIS_BRANCH}")
```
krzentner: You checked that this doesn't include the head of `origin/master`, right?

setup.py
gitanshu: The updated flake8-docstrings works fine with pydocstyle 4.0.0, so `>=` should be fine

scripts/travisci/check_precommit.sh
gitanshu: Yep

setup.py
gitanshu: Oh okay I see what I did there. The PR says I updated pydocstyle to 4.0.1 while I actually set it to >=4.0.0 before submitting. 4.0.0 works fine though, so I'll update the PR description.

ryanjulian:
Looking at the updated pre-commit documentation, here a couple convenient options for us:

1. https://pre-commit.com/#meta-hooks
1. https://pre-commit.com/#updating-hooks-automatically
1. https://pre-commit.com/#automatically-enabling-pre-commit-on-repositories

My first impression is that we could reduce friction significantly by:
* Adding a meta-hook (1) which checks to make sure pre-commit is up-to-date (and usually updating using (2)
* Adding (3) to our repos so that everyone who clones is already setup with pre-commit.

@krzentner @naeioi @gitanshu WDYT?

gitanshu:
> Wow, this took quite some sleuthing. Thank you. **Can you make sure this fix gets propagated to the other rlworkgroup projects using pre-commit?**

Will do. Also needs to be merged in release-2019.02
 >Since it's been a while, perhaps we should ask whether using pre-commit as the single-source-of-truth for presubmit checks has been a net benefit.

My general sentiment is pre-commit is simple and the config is convenient, compared to rolling out our own checks, with these options covering most of our issues.

>Looking at the updated pre-commit documentation, here a couple convenient options for us:

>https://pre-commit.com/#meta-hooks
>https://pre-commit.com/#updating-hooks-automatically
>https://pre-commit.com/#automatically-enabling-pre-commit-on-repositories
>My first impression is that we could reduce friction significantly by:

>Adding a meta-hook (1) which checks to make sure pre-commit is up-to-date (and usually updating >using (2)

Looks like there's only 3 pre-defined meta-hooks, none of which check for pre-commit being up to date, and there is no option to define our own. It might be possible to do this with a custom hook (not meta) though. (2) should be enough anyway. 

I was slightly skeptical about auto-updating in the sense that what if pre-commit, say, removes a hook. 

For example, flake8 now has it's own hook too, and pre-commit will someday deprecate/remove it from pre-commit-hooks repo (https://github.com/pre-commit/pre-commit-hooks/issues/344) but looks like they handle it well, so at least we'd know. So, it's all good.

Individual hooks might still break something on auto-update, but this might be far less frequent compared to the current issues. 

So, I think this is a good idea overall.

>Adding (3) to our repos so that everyone who clones is already setup with pre-commit.

I really like this one.

Another thing, since we run flake8 and friends inside pre-commit, should we still keep them in dev dependencies where they go out of sync regularly with the pre-commit version?

Also, this discussion reminded me that I should update pydocstyle in .pre-commit-config.yaml too

ryanjulian:
#### "Update me!" meta-hook
While it's not pre-defined, it would be easy to define a hook which checks that you are up-to-date and shows a big fat warning otherwise (it needn't block the commit). A hook is just a shell script.

#### Auto-install
I created an issue https://github.com/rlworkgroup/garage/issues/894 to track this and the above idea. Try it when you have some free time.

### Keeping versions in dev dependencies
That's a good question. I think yes `[dev]` is supposed to have everything needed for developing the software. In principle, pre-commit checks things, it doesn't fix them. It would be weird if pre-commit checked my code with yapf, but I can't run yapf in the environment (e.g. using my editor of choice) to automatically run YAPF on save (because YAPF is trapped in the pre-commit venv).

AFAICT this problem only applies to flake8 and yapf, our two most frequent problem children. We should probably add comments to setup.py/.pre-commit-config.yaml reminding people to update them together, though.

naeioi:
I have a strong inclination to keep using pre-commit than writing our own `make check`. It will be hard to make an easily configurable tool. 

> A local pre-commit installation (as run during git commit) is not guaranteed to return the same results as the CI, because the CI always uses the most up-to-date pre-commit environment

Versions of pre-commit hooks are already pinned in `.pre-commit-config.yaml`. I think as long as we also pinned pre-commit itself in `setup.py` this will not be a problem.

krzentner:
Echoing what everyone else has said here, I think that pre-commit has served us quite well, even if occasionally a version mis-match has made things difficult. I definitely don't think we should make installing docker at prerequisite of developing garage. I feel that would roughly double (or more) the amount of setup required, which is far too much.

I think an "Update Me!" meta-hook is probably a good idea, and would save time overall.

ryanjulian:
I think you may need to rebase in order for codecov to successfully report (and make this PR green)

Issue #529, opened 2019-02-21
Refactor BernoulliMLPRegressor to use garage.tf.Model


ryanjulian:
* [x] BernoulliMLPRegressor

ahtsan:
closed by #705

naeioi:
Closed by #890

PR #890, opened 2019-09-21
Replace BernoulliMLPRegressor

naeioi:
This PR replaces the old `BernoulliMLPRegressor` with the new one implemented using `garage.tf.models`.

Transit test can be found in `benchmark_bernoulli_mlp_regressor` branch. Travis log can be found [here](https://travis-ci.com/rlworkgroup/garage/jobs/237741815). 

No integration test is added because there is no `BernoulliMLPBaseline` that uses this regressor. Such a baseline will not be practical to train an algorithm, because the value function to be fit is always a real-value function, not a binary one. I believe the same logic applies to `CategoricalMLPRegressor`.


ahtsan:
After some offline discussion, I think we agree that we should preserve existing interface from the old primitives, so please add back the missing functions.

avnishn:
@ahtsan can you explain what you mean by 

> preserve existing interface from the old primitives

Thanks!

ahtsan:
> @ahtsan can you explain what you mean by
> 
> > preserve existing interface from the old primitives
> 
> Thanks!

I meant to keep all the functions that exist in the old primitives, e.g. `sample_predict` which exists in old primitive but not in the model-version. We should keep that in the model-version too.

ryanjulian:
@naeioi makes a good point about this regressor not being a very valuable baseline.

We can only speculate as to why it was originally introduced. I looked through the garage/rllab history and didn't find any clues. I can imagine two reasons: as a baseline for a sparse-reward environment only (where you know the reward is only 0 or 1) and as an estimator of some sort for a meta-RL algorithm.

If we don't find a use for it in the next release or two, I think it it safe to delete.

For this PR, I think a transition test is sufficient.

tests/benchmarks/test_benchmark_ppo.py
```diff
@@ -1,233 +0,0 @@
-'''
```
ryanjulian: why did you delete this file?

src/garage/tf/regressors/bernoulli_mlp_regressor.py
```diff
@@ -1,167 +1,242 @@
...
+        tabular.record('{}/LossAfter'.format(self._name), loss_after)
+        tabular.record('{}/dLoss'.format(self._name), loss_before - loss_after)
+        self._first_optimized = True
 
     def predict(self, xs):
-        return self.f_predict(np.asarray(xs))
+        """
+        Predict ys based on input xs.
 
-    def sample_predict(self, xs):
```
ahtsan: Please add back `sample_predict` and `predict_log_likelihood` in the model version, and tests for the functions.

tests/garage/tf/regressors/test_bernoulli_mlp_regressor.py
```diff
@@ -75,8 +75,8 @@ class TestBernoulliMLPRegressorWithModel(TfGraphTestCase):
     ])
     # yapf: enable
     def test_fit_normalized(self, input_shape, output_dim):
-        bmr = BernoulliMLPRegressorWithModel(
-            input_shape=input_shape, output_dim=output_dim)
+        bmr = BernoulliMLPRegressorWithModel(input_shape=input_shape,
```
ahtsan: should also be renamed? also for other occurrence in this file

naeioi:
Transit tests can be found in `benchmark_bernoulli_mlp_regressor` branch. Travis report is [here](https://travis-ci.com/rlworkgroup/garage/jobs/240924068). 

Close #529 .

PR #891, opened 2019-09-22
Replace CategoricalLSTMPolicy with Model

naeioi:
This PR replaces `CategoricalLSTMPolicy` with the one implemented using `garage.tf.models`. 

Benchmark script is in `benchmark_categorical_lstm_policy` branch, modified from `benchmark_ppo_catogorical.py`. 

There seems to be some randomness in the benchmark. In some environments, the trails of the old and the new policies do not match very well. When this happends, usually the new one performs worse. See the figures below. 

Raw data for tensorboard. 
[2019-09-21-16-09-09-112853.zip](https://github.com/rlworkgroup/garage/files/3640255/2019-09-21-16-09-09-112853.zip)

**Assault-ramDeterministic-v4**
assault_seed_42
assault_seed_70
assault_seed_96
**Breakout-ramDeterministic-v4**
breakout_seed_40
breakout_seed_47
breakout_seed_86
**ChopperCommand-ramDeterministic-v4**
chopper_seed_3
chopper_seed_71
chopper_seed_99
**LunarLander-v2**
lunarLander_seed_23
lunarLander_seed_24
lunarLander_seed_35
**Tutankham-ramDeterministic-v4/trial_1_seed_41**
tutankham_seed_41
tutankham_seed_65
tutankham_seed_79


ahtsan:
Except `Breakout-ramDeterministic-v4`, benchmarks looks good. Do you mind running a few more trials for `Breakout-ramDeterministic-v4`?

naeioi:
> Did you run these on PPO? You might also consider running them on some of the continuous control benchmarks which use discrete actions.
Yes. These benchmarks used PPO.

I am going to take Anson's advice to rerun Breakout to see if the divergence was due to some uncontrolled randomness. And if not, look into the implementation of CategoricalLSTMPolicy.

krzentner:
Honestly, with single runs like this it's really hard to tell if we're just looking at noise or not. Generally without at least 3 runs / env it's hard to know if there's a real difference.

naeioi:
> Honestly, with single runs like this it's really hard to tell if we're just looking at noise or not. Generally without at least 3 runs / env it's hard to know if there's a real difference.

Every env is indeed run for 3 trails, but each with different seed. I think this is also how we proceed with other primitives.

ryanjulian:
@naeioi have you tried again? found anything new?

naeioi:
I rerun Breakout environment several times and the result is consistent. So the performance difference is not due to randomness but implementation. 

Screenshot from 2019-10-01 14-20-56

The old implementation is also more consistent w.r.t seed. It almost always converges to an average return of 1 regardless of seed, while the new one is more sensitive to seed and often converges to a suboptimal policy.

I am looking into the implementation detail.

ryanjulian:
i think that you may need to run more seeds to really tell. see the kind of plots Utkarsh is generating in https://github.com/rlworkgroup/garage/pull/903

naeioi:
> 
> 
> i think that you may need to run more seeds to really tell. see the kind of plots Utkarsh is generating in #903

Just want to take a note here: In the last figure, I was benchmarking with the same seeds that were observed to give a bad performance for the policy with the model. I rerun with these seeds and the bad results are consistent.

ryanjulian:
@naeioi @ahtsan do you have any hypothesis about what is causing this? @nish21 is this similar to the problems you see with GaussianGRU?

nish21:
I didn't find any issues with GRU implementation. The issue with GaussianGRUPolicy was that I forgot to replace `tf.random.normal` with `np.random.normal` while generating actions.

naeioi:
I will try using `np.random.normal` to see if it gives better results.

naeioi:
OK.  I found that `tf.random.normal` is not relevant because it is not used in categorical policies...

ryanjulian:
@naeioi what do the results look like with 10-30 seeds on Breakout? Do you need a bigger computer to run them?

naeioi:
@ryanjulian @ahtsan The averaged result of breakout with 10 seeds looks pretty good.
Breakout-ramDeterministic-v4_benchmark

I indeed need a destop to do benchmarking. The results I posted here were all done on my laptop. 

I was planning to write transit test for `RecurrentCategorical` and `LSTMModel`, but this new result convinced me that it is not necessary.

naeioi:
Will do. Thanks all for your help and effort to make this PR ready.

ryanjulian:
@naeioi is there anything blocking this merge?

naeioi:
Nothing. I am closing this now.

PR #892, opened 2019-09-23
Replace ContinuousMLPQFunction

### LINKED ISSUE #526: Refactor ContinuousMLPPolicy & ContinuousMLPQFunction to use garage.tf.Model
* [x] ContinuousMLPPolicy
* [x] ContinuousMLPQFunction

ahtsan:
This PR does the following:
* Remove all occurrence of ContinuousMLPQFunction (old one)
* Rename ContinuousMLPQFunctionWithModel to ContinuousMLPQFunction
* Remove corresponding transition test (test_continuous_mlp_q_function_with_model_transit.py)

Summary: In most cases model-version performs better or at least as good as the old one.

For benchmark scripts, checkout the branch origin/benchmark_continuous_mlp_q_function.

Benchmark results:
image
image
image
image
image
image
image

This and #888 will close #526.


ryanjulian:
Overall this LGTM.

I'm a little concerned about a couple of the results on HalfCheetah and Walker2d (the hardest environments). It a couple cases seem inconsistent. I don't think it's indicative of a problem with the primitive, but instead it may be the algorithm.

*Can you check that we're using the same exploration strategy for both the baseline and our algo?* I think @utkarshjp7 may have also unearthed some subtle differences between our DDPG and the baselines one, which might explain some differences.

ahtsan:
> Overall this LGTM.
> 
> I'm a little concerned about a couple of the results on HalfCheetah and Walker2d (the hardest environments). It a couple cases seem inconsistent. I don't think it's indicative of a problem with the primitive, but instead it may be the algorithm.
> 
> _Can you check that we're using the same exploration strategy for both the baseline and our algo?_ I think @utkarshjp7 may have also unearthed some subtle differences between our DDPG and the baselines one, which might explain some differences.

Utkarsh actually thought DDPG is fine, he just found something for TD3. I will have another pass on it and run benchmark on these two harder environments and see if it persist.

ahtsan:
> Overall this LGTM.
> 
> I'm a little concerned about a couple of the results on HalfCheetah and Walker2d (the hardest environments). It a couple cases seem inconsistent. I don't think it's indicative of a problem with the primitive, but instead it may be the algorithm.
> 
> _Can you check that we're using the same exploration strategy for both the baseline and our algo?_ I think @utkarshjp7 may have also unearthed some subtle differences between our DDPG and the baselines one, which might explain some differences.
I ran benchmark again and it looks better. Perhaps it's a matter of random seed?
image
image

ryanjulian:
Yes I suspect it probably has something to do with these environments sensitivity to initialization and/or exploration.

Here's a way you can check (which will help people in the future running benchmarks):

Run the same 2 experiments here but using 30 random seeds. Plots the mean results (across all 30 seeds) for garage and garage_models (that's 2 curves) with a confidence interval. This is very easy to do [with `seaborn`](https://seaborn.pydata.org/generated/seaborn.lineplot.html) and `pandas`.

I suspect we will find that the confidence interval is wider than the variation we are seeing on the plots.

ahtsan:
I ran for 10 trials and model version outperforms the old version.

**HalfCheetah-v2:**
Old
image
Model
image
**Walker2d-v2**
Old
image
Model
image

Code is in `origin/benchmark_continuous_mlp_q_function`

ryanjulian:
Awesome! Feel free to merge this.

If you can somehow add the code you used to produce these plots to master (and put both curves on the same plot), I think it would be super helpful for future benchmarking. At least post a branch or a gist for the people currently working on the models transition.

PR #893, opened 2019-09-23
Replace CategoricalMLPRegressor

zequnyu:
- Deleted CategoricalMLPRegressor
- Renamed CategoricalMLPRegressorWithModel to CategoricalMLPRegressor and did minor style/name update
- Updated test accordingly

[Transition test](https://github.com/rlworkgroup/garage/blob/catmlp_regressor_transition_test/tests/garage/tf/regressors/test_categorical_mlp_regressor_with_model_transit.py)

[Result](https://travis-ci.com/rlworkgroup/garage/builds/128789446)

No integration test and benchmarking needed.


src/garage/tf/regressors/__init__.py
```diff
@@ -21,8 +17,7 @@
 from garage.tf.regressors.gaussian_mlp_regressor import GaussianMLPRegressor
 
 __all__ = [
-    'BernoulliMLPRegressor', 'BernoulliMLPRegressorWithModel',
```
ahtsan: Why `BernoulliMLPRegressorWithModel` is also removed?

zequnyu: Might be a mistake...

ryanjulian:
Can you explain why benchmarks are not needed?

zequnyu:
> Can you explain why benchmarks are not needed?

As discussed with Anson, this should be similar to [BernoulliMLPRegressor](https://github.com/rlworkgroup/garage/pull/890#issue-319992039) right?

ryanjulian:
Ah I see. Yes, it doesn't appear benchmarks are needed.

PR #897, opened 2019-09-26
Replace GaussianCNNRegressor/Baseline

ahtsan:
This PR does the following:
* Remove all occurrence of GaussianConvRegressor and GaussianConvBaseline
* Rename GaussianCNNRegressorWithModel to GaussianCNNRegressor
* Rename GaussianCNNBaselineWithModel to GaussianCNNBaseline

**Benchmarking**: 10 trial on the environments `MemorizeDigits-v0` and `CubeCrash-v0`.

**Summary**: Model version performs better.

`CubeCrash-v0`:
old
image
model
image

`MemorizeDigits-v0`:
old
image
model
image

For benchmark scripts, checkout the branch [`origin/benchmark_gaussian_cnn_baseline.`](https://github.com/rlworkgroup/garage/tree/benchmark_gaussian_cnn_baseline)


nish21:
These results look good. But should we test them on some more environments?

ahtsan:
> These results look good. But should we test them on some more environments?

Yes, the rest would be Atari, which I am not sure if it's possible to train with OnPolicyAlgorithm (I keep having ResourceExhausted on my pc).

nish21:
What about `CarRacing-v0` and `CubeCrashScreenBecomesBlack-v0`? The input sizes are listed here. https://github.com/openai/gym/wiki/Table-of-environments

ahtsan:
> What about `CarRacing-v0` and `CubeCrashScreenBecomesBlack-v0`? The input sizes are listed here. https://github.com/openai/gym/wiki/Table-of-environments



> What about `CarRacing-v0` and `CubeCrashScreenBecomesBlack-v0`? The input sizes are listed here. https://github.com/openai/gym/wiki/Table-of-environments

Yes we can run them too. I will attach the results later.

tests/garage/tf/baselines/test_baselines.py
```diff
@@ -4,36 +4,23 @@
...
 class TestTfBaselines(TfGraphTestCase):
+
     def test_baseline(self):
         """Test the baseline initialization."""
         box_env = TfEnv(DummyBoxEnv())
         deterministic_mlp_baseline = ContinuousMLPBaseline(env_spec=box_env)
         gaussian_mlp_baseline = GaussianMLPBaseline(env_spec=box_env)
 
-        discrete_env = TfEnv(Resize(DummyDiscrete2DEnv(), width=64, height=64))
-        gaussian_conv_baseline = GaussianConvBaseline(
```
nish21: Shouldn't this be changed to `GaussianCNNBaseline` instead of removing?

ahtsan: This file actually tests nothing. It simply calls the constructor and functions for coverage. The baselines should be tested in their own file, e.g. `test_gaussian_cnn_baseline.py` for `GaussianCNNBaseline`.

ahtsan:
> What about `CarRacing-v0` and `CubeCrashScreenBecomesBlack-v0`? The input sizes are listed here. https://github.com/openai/gym/wiki/Table-of-environments

Looks like we don't have a cnn policy for continuous action space so we can't do `CarRacing-v0`.

ahtsan:
> It would be good to run e.g. `Breakout-v0` (or one of the other easier atari environments) with PPO or TRPO. If that works we should be good to submit this change.

I will run on `Pong` and attach result later.

ahtsan:
After 1000 epochs, it learns to classify and able to obtain positive rewards @krzentner 
Here is the result of 3 trials on `MemorizeDigits-v0`
image
image

krzentner:
(For posterity, we discussed offline using Atari to benchmark this change, but this baseline has never been used for Atari, and we didn't want to have to do the hyper-parameter search to use it.)

PR #900, opened 2019-09-26
Added helper function for plotting mean curves

ahtsan:
Plot mean curve using `seaborn`.

Benchmarking on n environments and y trials will generate
* n `<env-id>_benchmark_.png` which plot curves for each trial
* n `<env_id>_benchmark_mean.png` which plot one averaged curve over y trials

Example usage: https://github.com/rlworkgroup/garage/blob/633a965ab1403b81ec6154690a4d8d1c3568069e/tests/benchmarks/test_benchmark_ddpg.py

Example:
Running benchmarks for DDPG on 1 environment and 2 trials generates
image
image


setup.py
```diff
@@ -29,6 +29,7 @@
     'ray',
     'scikit-image',
     'scipy',
+    'seaborn',
```
krzentner: Can you add it to the `dev` dependencies instead? It should only be needed when running benchmarking.

ahtsan: Fixed

PR #903, opened 2019-10-01
Replace CategoricalMLPPolicy

utkarshjp7:
This PR replaces the layers implementation of CategoricalMLPPolicy with the model implementation.

Here are the benchmark results on `'LunarLander-v2` and `CartPole-v1`. Plots show mean over three trials. 

LunarLander-v2_benchmark_mean
CartPole-v1_benchmark_mean


src/garage/tf/policies/categorical_mlp_policy.py
```diff
@@ -1,103 +1,146 @@
...
         return action, dict(prob=prob)
 
     def get_actions(self, observations):
+        """Return multiple actions."""
         flat_obs = self.observation_space.flatten_n(observations)
         probs = self._f_prob(flat_obs)
         actions = list(map(self.action_space.weighted_sample, probs))
         return actions, dict(prob=probs)
 
+    def get_regularizable_vars(self):
```
ahtsan: Should we add this function to the base class as well?

ahtsan:
Benchmark is doing pretty good on these two environments. It would be nice to have it ran on a few more harder environment, possibly Atari-ram envs. @nish21 can you provide some good environments for this case?

src/garage/tf/policies/categorical_mlp_policy.py
utkarshjp7: `CNN` and `GRU` policies will have different implementation of this function since they use different naming convention for parameters. I think we should leave this here.

utkarshjp7:
> Benchmark is doing pretty good on these two environments. It would be nice to have it ran on a few more harder environment, possibly Atari-ram envs. @nish21 can you provide some good environments for this case?

Yes, @nish21 provided me a list of Atrai-ram envs. I will update once its done.

utkarshjp7:
Breakout-ramDeterministic-v4_benchmark_mean
ChopperCommand-ramDeterministic-v4_benchmark_mean
Tutankham-ramDeterministic-v4_benchmark_mean

utkarshjp7:
> Please add your benchmark (and possibly an example) before submitting.

Do you mean benchmark against `baselines` or any other external library, or just the garage implementation itself without comparing with anyone?

There is already an example in `examples/tf/vpg_cartpole.py` which works without having to change anything.

tests/garage/experiment/test_resume.py
```diff
@@ -25,6 +26,8 @@ def teardown_method(self):
         super().teardown_method()
 
     def test_resume(self):
+        self.sess = tf.compat.v1.Session(graph=tf.Graph())
```
ryanjulian: shouldn't this be handled by TfGraphTestCase? I think this is our most buggy fixture, TBH.

ryanjulian:
@naeioi @zequnyu please take a look at the LocalTFRunner changes before this is merged.

ryanjulian:
@utkarshjp7 I think it's sufficient to just add your benchmark script, but without the old primitives. The way I see it is that writing those scripts took time, and they will probably come in handy next time someone changes this code (e.g. for tf2 conversion or removing our custom distributions, for instance).

tests/garage/experiment/test_resume.py
ahtsan: This is a hot fix - The test tried to initialize the same set of parameters twice in the same graph and restore the weights, and at the second time the parameters' name are all appended with `_1`, causing the restoring to fail. This fix essentially creates a new graph when restoring. I imagine in a real use case we won't really save and restore the test with the same session? even so, it's doable with this fix.

PR #905, opened 2019-10-01
Implemented PPO in PyTorch

yonghyuc:
Implemented Proximal Policy Optimization in PyTorch

### DIFF
--- a/src/garage/torch/algos/__init__.py
+++ b/src/garage/torch/algos/__init__.py
@@ -1,5 +1,6 @@
 """PyTorch algorithms."""
 from garage.torch.algos.ddpg import DDPG
 from garage.torch.algos.vpg import VPG
+from garage.torch.algos.ppo import PPO  # noqa: I100
 
-__all__ = ['DDPG', 'VPG']
+__all__ = ['DDPG', 'VPG', 'PPO']
--- a/src/garage/torch/algos/loss_function_utils.py
+++ b/src/garage/torch/algos/loss_function_utils.py
@@ -62,3 +62,50 @@ def compute_advantages(discount, gae_lambda, max_path_length, baselines,
 
     advantages = F.conv2d(deltas, adv_filter, stride=1).squeeze()
     return advantages
+
+
+def pad_to_last(nums, total_length, axis=-1, val=0):
+    """Pad val to last in nums in given axis.
+
+    length of the result in given axis should be total_length.
+
+    Raises:
+      IndexError: If the input axis value is out of range of the nums array
+
+    Args:
+        nums (numpy.ndarray): The array to pad.
+        total_length (int): The final width of the Array.
+        axis (int): Axis along which a sum is performed.
+        val (int): The value to set the padded value.
+
+    Returns:
+        torch.Tensor: Padded array
+
+    """
+    tensor = torch.Tensor(nums)
+    axis = (axis + len(tensor.shape)) if axis < 0 else axis
+
+    if len(tensor.shape) <= axis:
+        raise IndexError('axis {} is out of range {}'.format(
+            axis, tensor.shape))
+
+    padding_config = [0, 0] * len(tensor.shape)
+    padding_idx = abs(axis - len(tensor.shape)) * 2 - 1
+    padding_config[padding_idx] = max(total_length - tensor.shape[axis], val)
+    return F.pad(tensor, padding_config)
+
+
+def filter_valids(tensor, valids):
+    """Filter out tensor using valids (last index of valid tensors).
+
+    valids contains last indices of each rows.
+
+    Args:
+        tensor (torch.Tensor): The tensor to filter
+        valids (list[int]): Array of length of the valid values
+
+    Returns:
+        torch.Tensor: Filtered Tensor
+
+    """
+    return [tensor[i][:valids[i]] for i in range(len(valids))]
--- a/src/garage/torch/algos/ppo.py
+++ b/src/garage/torch/algos/ppo.py
@@ -1,0 +1,108 @@
+"""Proximal Policy Optimization (PPO)."""
+import copy
+
+import torch
+
+from garage.torch.algos import VPG
+
+
+class PPO(VPG):
+    """Proximal Policy Optimization (PPO).
+
+    Args:
+        env_spec (garage.envs.EnvSpec): Environment specification.
+        policy (garage.torch.policies.base.Policy): Policy.
+        baseline (garage.np.baselines.Baseline): The baseline.
+        max_path_length (int): Maximum length of a single rollout.
+        policy_lr (float): Learning rate for training policy network.
+        lr_clip_range (float): The limit on the likelihood ratio between
+            policies.
+        n_samples (int): Number of train_once calls per epoch.
+        discount (float): Discount.
+        gae_lambda (float): Lambda used for generalized advantage
+            estimation.
+        center_adv (bool): Whether to rescale the advantages
+            so that they have mean 0 and standard deviation 1.
+        positive_adv (bool): Whether to shift the advantages
+            so that they are always positive. When used in
+            conjunction with center_adv the advantages will be
+            standardized before shifting.
+        optimizer (object): The optimizer of the algorithm. Should be the
+            optimizers in torch.optim.
+        policy_ent_coeff (float): The coefficient of the policy entropy.
+            Setting it to zero would mean no entropy regularization.
+        use_softplus_entropy (bool): Whether to estimate the softmax
+            distribution of the entropy to prevent the entropy from being
+            negative.
+        stop_entropy_gradient (bool): Whether to stop the entropy gradient.
+        entropy_method (str): A string from: 'max', 'regularized',
+            'no_entropy'. The type of entropy method to use. 'max' adds the
+            dense entropy to the reward for each time step. 'regularized' adds
+            the mean entropy to the surrogate objective. See
+            https://arxiv.org/abs/1805.00909 for more details.
+
+    """
+
+    def __init__(self,
+                 env_spec,
+                 policy,
+                 baseline,
+                 max_path_length=500,
+                 policy_lr=3e-4,
+                 lr_clip_range=2e-1,
+                 n_samples=1,
+                 discount=0.99,
+                 gae_lambda=0.97,
+                 center_adv=True,
+                 positive_adv=False,
+                 optimizer=None,
+                 policy_ent_coeff=0.0,
+                 use_softplus_entropy=False,
+                 stop_entropy_gradient=False,
+                 entropy_method='no_entropy'):
+        super().__init__(env_spec, policy, baseline, max_path_length,
+                         policy_lr, n_samples, discount, gae_lambda,
+                         center_adv, positive_adv, optimizer, policy_ent_coeff,
+                         use_softplus_entropy, stop_entropy_gradient,
+                         entropy_method)
+
+        self._lr_clip_range = lr_clip_range
+        self._old_policy = copy.deepcopy(self.policy)
+
+    def _compute_objective(self, advantages, valids, obs, actions, rewards):
+        """Compute objective using surrogate value and clipped surrogate value.
+
+        Args:
+            advantages (torch.Tensor): Expected rewards over the actions
+            valids (list[int]): length of the valid values for each path
+            obs (torch.Tensor): Observation from the environment.
+            actions (torch.Tensor): Predicted action.
+            rewards (torch.Tensor): Feedback from the environment.
+
+        Returns:
+            torch.Tensor: Calculated objective values
+
+        """
+        # pylint: disable=unused-argument
+        # Compute constraint
+        with torch.no_grad():
+            old_ll = self._old_policy.log_likelihood(obs, actions)
+        new_ll = self.policy.log_likelihood(obs, actions)
+
+        likelihood_ratio = (new_ll - old_ll).exp()
+
+        # Memorize the policy state_dict
+        self._old_policy.load_state_dict(self.policy.state_dict())
+
+        # Calculate surrogate
+        surrogate = likelihood_ratio * advantages
+
+        # Clipping the constraint
+        likelihood_ratio_clip = torch.clamp(likelihood_ratio,
+                                            min=1 - self._lr_clip_range,
+                                            max=1 + self._lr_clip_range)
+
+        # Calculate surrotate clip
+        surrogate_clip = likelihood_ratio_clip * advantages
+
+        return torch.min(surrogate, surrogate_clip)
--- a/src/garage/torch/algos/vpg.py
+++ b/src/garage/torch/algos/vpg.py
@@ -8,7 +8,7 @@ import torch.nn.functional as F
 
 from garage.misc import tensor_utils
 from garage.np.algos import BatchPolopt
-from garage.torch.algos import loss_function_utils as loss_utils
+from garage.torch.algos import loss_function_utils
 
 
 class VPG(BatchPolopt):
@@ -83,7 +83,9 @@ class VPG(BatchPolopt):
                                           stop_entropy_gradient,
                                           policy_ent_coeff)
         self._episode_reward_mean = collections.deque(maxlen=100)
-        self._optimizer = optimizer(policy.parameters(), lr=policy_lr)
+        self._optimizer = optimizer(policy.parameters(),
+                                    lr=policy_lr,
+                                    eps=1e-5)
 
         super().__init__(policy=policy,
                          baseline=baseline,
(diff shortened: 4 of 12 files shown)


yonghyuc:
benchmark result between PyTorch PPO, TF PPO, and baseline PPO2

HalfCheetah-v2_benchmark
HalfCheetah-v2_benchmark

Hopper-v2_benchmark
Hopper-v2_benchmark

InvertedPendulum-v2_benchmark
InvertedPendulum-v2_benchmark

Reacher-v2_benchmark
Reacher-v2_benchmark

Swimmer-v2_benchmark
Swimmer-v2_benchmark

Walker2d-v2_benchmark
Walker2d-v2_benchmark

yonghyuc:
In src/garage/torch/algos/__init__.py, 
when I put the `from garage.torch.algos.ppo import PPO` above the `from garage.torch.algos.vpg import VPG` (to follow the flake8 rule - I100), I got an error 
`    
from garage.torch.algos import VPG
E   ImportError: cannot import name 'VPG'
`

Do I just need to put the `# yapf: disable` and ignore the flake8 warning message or is there any other solution to solve this?

ryanjulian:
Can you provide benchmark plots averaged over 10 seeds, similar to the plots here? https://github.com/rlworkgroup/garage/pull/903

src/garage/torch/algos/vpg.py
```diff
@@ -107,34 +123,28 @@ def _check_entropy_configuration(self, entropy_method, center_adv,
 
     def train_once(self, itr, paths):
         """Perform one step of policy optimization."""
-        samples_data = self.process_samples(itr, paths)
+        valids, obs, actions, rewards, average_return = self.process_samples(
+            itr, paths)  # noqa: E501
```
ryanjulian: why is there a noqa here? the line isn't too long.

src/garage/torch/algos/vpg.py
```diff
@@ -197,14 +194,26 @@ def _get_baselines(self, path):
...
     def process_samples(self, itr, paths):
         """Process sample data based on the collected paths."""
         for path in paths:
             path['returns'] = special.discount_cumsum(path['rewards'],
                                                       self.discount)
+
+        valids, obs, actions, rewards = [], [], [], []
+        for path in paths:
+            valids.append(len(path['actions']))
+            obs.append(
```
ryanjulian:
```python
obs = [_add_padding_2d(path['observations'], self.max_path_length) for path in paths]
```
...and similarly for returns, actions, rewards, etc.

src/garage/torch/algos/vpg.py
```diff
@@ -11,6 +11,22 @@
 from garage.torch.algos import loss_function_utils as loss_utils
```
ryanjulian: is shortening this really necessary?

src/garage/torch/algos/ppo.py
```diff
@@ -0,0 +1,96 @@
...
+                 use_softplus_entropy=False,
+                 stop_entropy_gradient=False,
+                 entropy_method='no_entropy'):
+        super().__init__(env_spec, policy, baseline, max_path_length,
+                         policy_lr, n_samples, discount, gae_lambda,
+                         center_adv, positive_adv, optimizer, policy_ent_coeff,
+                         use_softplus_entropy, stop_entropy_gradient,
+                         entropy_method)
+
+        self._constraint_clip_range = lr_clip_range
```
ryanjulian: please don't rename member variables which originate at constructor params, i.e. this should be `self._lr_clip_range`

src/garage/torch/algos/ppo.py
```diff
@@ -0,0 +1,96 @@
...
+        current_policy_state_dict = self.policy.state_dict()
+        self.policy.load_state_dict(self._old_policy_state_dict)
+
+        with torch.no_grad():
+            old_ll = self.policy.log_likelihood(obs, actions)
+
+        self.policy.load_state_dict(current_policy_state_dict)
+        new_ll = self.policy.log_likelihood(obs, actions)
+
+        constraint = (new_ll - old_ll).exp()
```
ryanjulian: usually the word "constraint" refers to the rule used to constrain a value, not the constrained value itself.

this isn't a constraint, it's a likelihood ratio. we constrain the likelihood ratio by clipping it later, so in this case the constraint is lr_clip_range. please name this more clearly.

src/garage/torch/algos/vpg.py
```diff
@@ -11,6 +11,22 @@
...
+    return F.pad(tensor, (0, 0, 0, padding_length))
+
+
+def _add_padding_1d(vals, length):
+    tensor = torch.Tensor(vals)
+    padding_length = max(length - tensor.shape[-1], 0)
+    return F.pad(tensor, (0, padding_length))
+
+
+def _filter_valids(tensor, valids):
```
ryanjulian: please move this into loss_function_utils and give it a proper docstring/test

src/garage/torch/algos/loss_function_utils.py
```diff
@@ -4,7 +4,25 @@
...
+       rev(y)[t] - discount*rev(y)[t-1] = rev(x)[t]
+
+    Given the time-domain IIR filter step response, we can
+    calculate the filter response to our signal by convolving the
+    signal with the filter response function. The time-domain IIR
+    step response is calculated below as discount_filter:
+        discount_filter =
+            [1, discount, discount^2, ..., discount^N-1]
+            where the epsiode length is N.
+    """
```
ryanjulian: this still requires a full docstring.

src/garage/torch/algos/__init__.py
```diff
@@ -1,5 +1,6 @@
 """PyTorch algorithms."""
 from garage.torch.algos.ddpg import DDPG
 from garage.torch.algos.vpg import VPG
+from garage.torch.algos.ppo import PPO  # noqa: I100
```
ryanjulian: this is the correct way to override flake8 in __init__

src/garage/torch/algos/vpg.py
```diff
@@ -11,6 +11,22 @@
...
+def _add_padding_2d(vals, length):
+    tensor = torch.Tensor(vals)
+    padding_length = max(length - tensor.shape[0], 0)
+    return F.pad(tensor, (0, 0, 0, padding_length))
+
+
+def _add_padding_1d(vals, length):
+    tensor = torch.Tensor(vals)
+    padding_length = max(length - tensor.shape[-1], 0)
+    return F.pad(tensor, (0, padding_length))
```
ryanjulian: please add these to a new module called `path_utils` with full docstrings and tests

how do the 1d and 2d case differ?

src/garage/torch/algos/vpg.py
```diff
@@ -11,6 +11,22 @@
 from garage.torch.algos import loss_function_utils as loss_utils
 
 
+def _add_padding_2d(vals, length):
+    tensor = torch.Tensor(vals)
+    padding_length = max(length - tensor.shape[0], 0)
+    return F.pad(tensor, (0, 0, 0, padding_length))
+
+
+def _add_padding_1d(vals, length):
```
ryanjulian: i think these would be more clearly named as something like `pad_to_length` this makes it more clear how this is different than the default `F.pad`.

src/garage/torch/algos/vpg.py
```diff
@@ -107,34 +123,28 @@ def _check_entropy_configuration(self, entropy_method, center_adv,
...
-            self._add_padding(
-                self.policy.get_entropy(torch.Tensor(path['observations'])),
-                self.max_path_length) for path in paths
-        ])
+        self.baseline.fit(paths)
+        return average_return
 
-        valids = [len(path['actions']) for path in paths]
+    def _compute_loss(self, itr, paths, valids, obs, actions, rewards):
+        policy_entropies = self.policy.get_entropy(obs)
```
ryanjulian: this doesn't seem to obey `stop_entropy_gradient`

src/garage/torch/algos/ppo.py
```diff
@@ -0,0 +1,96 @@
...
+                 policy_lr=1e-2,
+                 lr_clip_range=1e-2,
+                 n_samples=1,
+                 discount=0.99,
+                 gae_lambda=1,
+                 center_adv=True,
+                 positive_adv=False,
+                 optimizer=None,
+                 policy_ent_coeff=0.0,
+                 use_softplus_entropy=False,
```
ryanjulian: several of these seem unused?

src/garage/torch/algos/ppo.py
```diff
@@ -0,0 +1,96 @@
...
+            dense entropy to the reward for each time step. 'regularized' adds
+            the mean entropy to the surrogate objective. See
+            https://arxiv.org/abs/1805.00909 for more details.
+    """
+
+    def __init__(self,
+                 env_spec,
+                 policy,
+                 baseline,
+                 max_path_length=500,
```
ryanjulian: this is pretty long for more environments, and will slow sampling unnecessarily.

src/garage/torch/algos/vpg.py
yonghyuc: removed

yonghyuc:
benchmark result between PyTorch PPO, TF PPO, and baseline PPO2 using 10 seeds

HalfCheetah-v2_benchmark
Hopper-v2_benchmark
InvertedPendulum-v2_benchmark
Reacher-v2_benchmark
Swimmer-v2_benchmark
Walker2d-v2_benchmark

src/garage/torch/algos/vpg.py
yonghyuc: fixed

src/garage/torch/algos/vpg.py
yonghyuc: fixed

src/garage/torch/algos/ppo.py
yonghyuc: fixed

src/garage/torch/algos/ppo.py
yonghyuc: changed the name into likelihood_ratio

ryanjulian: i don't see the new name -- did you push to origin yet?

src/garage/torch/algos/ppo.py
avnishn: @yonghyuc I agree with ryan, it seems a lot of these are not used

src/garage/torch/algos/ppo.py
```diff
@@ -0,0 +1,96 @@
...
+                         use_softplus_entropy, stop_entropy_gradient,
+                         entropy_method)
+
+        self._constraint_clip_range = lr_clip_range
+        self._old_policy_state_dict = self.policy.state_dict()
+
+    def _compute_objective(self, advantages, obs, actions, rewards):
+        # Compute constraint
+        current_policy_state_dict = self.policy.state_dict()
+        self.policy.load_state_dict(self._old_policy_state_dict)
```
avnishn: we don't need to load the old policy params every time, especially considering we will be doing multiple gradient steps from the old policy, before updating the old policy. I am not sure if this is correct. @krzentner. 

I think that it is best for us to keep around the old policy as an object, by using copy.deepcopy, and only updating the old policy after the we are at the last gradient step for the iteration.

@yonghyuc

yonghyuc: I also thought about using deepcopy method but comments says using deepcopy method could raise an error in certain PyTorch version

https://discuss.pytorch.org/t/are-there-any-recommended-methods-to-clone-a-model/483/15
https://github.com/pytorch/pytorch/issues/415#issuecomment-271226740

yonghyuc: resolved offline.

src/garage/torch/algos/ppo.py
yonghyuc: changed the value into 100

src/garage/torch/algos/vpg.py
yonghyuc: changed the code to use `_get_policy_entropy`

src/garage/torch/algos/ppo.py
yonghyuc: Currently some of them (like `use_softplus_entropy` and `entropy_method`) are used in the parent class (VPG).
Is there no case that people use the parameters in PPO? 

@ryanjulian @avnishn

src/garage/torch/algos/vpg.py
yonghyuc: Changed the name into `pad_to_last` since the method only put the padding numbers on the last

src/garage/torch/algos/vpg.py
yonghyuc: created `path_util.py` and moved the functions
made `pad_to_last` new funciton and remove `_add_padding_1d` and `_add_padding_2d`

src/garage/torch/algos/vpg.py
yonghyuc: is this funciton need to move into `loss_function_utils` not `path_utils`?
what is the main role of the two modules? (differences?)
@ryanjulian

src/garage/torch/algos/ppo.py
ryanjulian: these are used to add entropy regularization, which you haven't implemented here. It would be nice to have entropy, but you can implement it next (or add it to this PR).

Did you implement entropy in VPG already? If you did, you should implement it in PPO as well.

Don't include constructor parameters which aren't used, even if the parent constructor has them.

src/garage/torch/algos/ppo.py
ryanjulian: this doesn't appear fixed here. have you pushed your code?

src/garage/torch/algos/vpg.py
ryanjulian: path_utils is for processing raw paths from the sampler into appropriate inputs for an algorithm. it does not handle any differentiable operations.

loss_function_utils is differentiable operations typically used when computing loss functions. `filter_valids` is a differentiable operation used during computing the loss function, so it goes there.

i realize now that your padding routines are also differentiable, so they should also go into loss_function_utils.

src/garage/torch/algos/vpg.py
```diff
@@ -107,34 +123,28 @@ def _check_entropy_configuration(self, entropy_method, center_adv,
...
         """Perform one step of policy optimization."""
-        samples_data = self.process_samples(itr, paths)
+        valids, obs, actions, rewards, average_return = self.process_samples(
+            itr, paths)  # noqa: E501
 
-        self._optimize_policy(itr, paths)
-        self.baseline.fit(paths)
-
-        return samples_data['average_return']
+        loss = self._compute_loss(itr, paths, valids, obs, actions, rewards)
```
ryanjulian: this is a nit, but i think `mean()` should be here. mean is part of the loss function.

yonghyuc:
benchmark with learning rate 3e-4

HalfCheetah-v2_benchmark
Hopper-v2_benchmark
InvertedPendulum-v2_benchmark
Reacher-v2_benchmark
Walker2d-v2_benchmark


I don't know why TensorFlow version is fine even with learning rate 1e-2..
but when I set learning rate about 3e-4, both TensorFlow and PyTorch PPO works fine.

src/garage/torch/algos/loss_function_utils.py
yonghyuc: I will put full docstring in "vpg_coverage" branch

src/garage/torch/algos/ppo.py
yonghyuc: I passed the `use_softplus_entropy` and `entropy_method` to (parent) VPG class. 
Since PPO only override `_compute_objective` function which is called inside `_compute_loss` (in VPG), these paremeters would used during `_compute_loss`.
You can check [here](https://github.com/rlworkgroup/garage/blob/363ebeb326916c89a1ffa25b1f63eeafdbf0eaf7/src/garage/torch/algos/vpg.py#L135) and [here](https://github.com/rlworkgroup/garage/blob/363ebeb326916c89a1ffa25b1f63eeafdbf0eaf7/src/garage/torch/algos/vpg.py#L158)

src/garage/torch/algos/vpg.py
yonghyuc: fixed

ryanjulian:
The TensorFlow version could have better gradient estimates.

I usually attribute this to our use of a differentiable GAE, where most other implementations compute GAE outside the computation graph and therefore their gradient estimates have more noise.

ryanjulian:
Please do merge your new plotting functions!

I would prefer if we plot in this style for all future benchmarks, so please replace usages of the old plotting code with your new version (which shows confidence intervals). You can do that in this PR or a different one, if you prefer.

ryanjulian:
Why do you think `InvertedPendulum-v2` did so much worse with the different learning, but the others so much better?

ryanjulian:
It looks like codecov isn't passsing because you aren't covering several cases in the algorithm configuration options, and you didn't check the error in your helper function.

You will have to add some test cases.

See more info on the coverage report here: https://codecov.io/gh/rlworkgroup/garage/compare/09ec44b4adee0373d6cd7090f79a1e14ac56219c...363ebeb326916c89a1ffa25b1f63eeafdbf0eaf7/diff

src/garage/torch/algos/loss_function_utils.py
```diff
@@ -13,3 +31,46 @@ def compute_advantages(discount, gae_lambda, max_len, baselines, rewards):
 
     advantages = F.conv2d(deltas, filter, stride=1).squeeze()
     return advantages
+
+
+def pad_to_last(nums, total_length, axis=-1, val=0):
+    """
+    Pad val to last in nums in given axis.
+
+    length of the result in given axis should be total_length.
```
ryanjulian: please provide a complete docstring for this method and the one below it

tests/garage/torch/algos/test_ppo.py
```diff
@@ -0,0 +1,51 @@
...
+from garage.experiment import deterministic, LocalRunner
+from garage.np.baselines import LinearFeatureBaseline
+from garage.torch.algos import PPO
+from garage.torch.policies import GaussianMLPPolicy
+from tests.fixtures import snapshot_config
+
+
+class TestPPO:
+
+    @classmethod
```
ryanjulian: this should not be a class method.

if you want to do class-wide setup (shared by all test methods in the class), use `setup_class(cls)`/`teardown_class(cls)`

yonghyuc:
benchmark with 400 steps

HalfCheetah-v2_benchmark
Hopper-v2_benchmark
InvertedPendulum-v2_benchmark
Reacher-v2_benchmark
Walker2d-v2_benchmark

yonghyuc:
> I suspect that the inverted pendulum being low, and the same as the tf version in its performance, is dependent on the hyperparameters that you are using @yonghyuc . You can try to fiddle with your hyperparameters (discount and gae-lambda). Also, you can check to see if your policy for that task is learning by using sim_policy for pytorch policies. The sim_policy.py file is under examples/torch/ on the torch_sac branch.
> 
> Other than that the results look good and once you fix the codecov commit!

I agree with you and maybe baseline also affect this result.
I used LinearFeatureBaseline for both TensorFlow and PyTorch PPO.
If I use different baseline (kind of GaussianMLPBaseline), result could be better than now.

tests/garage/torch/algos/test_ppo.py
yonghyuc: delete `@classmethod` annotation

yonghyuc:
> Please do merge your new plotting functions!
> 
> I would prefer if we plot in this style for all future benchmarks, so please replace usages of the old plotting code with your new version (which shows confidence intervals). You can do that in this PR or a different one, if you prefer.

@ryanjulian 
Is it okay to do this in new PR?
Also, I am thinking about refactoring benchmark test cases because there are many duplicated codes in the benchmark test cases.
I would implement parent test case method, and make few abstract method for each algorithms.

ryanjulian:
Please don't push to your branch just to check things like CI results or formatting problems. We provide extensive local tooling for checking everything the CI checks.

Every time you push to your PR, a new build is triggered which holds up the rest of the builds for 30-45 mins.

ryanjulian:
To answer your question -- sure you can do it in the future PR. It looks like all the comments on this one aren't yet addressed.

src/garage/torch/algos/loss_function_utils.py
```diff
@@ -13,3 +31,46 @@ def compute_advantages(discount, gae_lambda, max_len, baselines, rewards):
...
+        val:
+
+    Returns:
+
+    """
+    tensor = torch.Tensor(nums)
+    axis = (axis + len(tensor.shape)) if axis < 0 else axis
+
+    if len(tensor.shape) <= axis:
+        raise IndexError(f'axis {axis} is out of range {tensor.shape}')
```
ryanjulian: f-strings are not supported in python 3.5 -- this will crash if executed.

ryanjulian:
please update the README.md to include the new algorithm.

Issue #794, opened 2019-07-15
OnPolicyVectorizedSampler should not flatten pixel observations


ahtsan:
Currently OnPolicyVectorizedSampler always flattens the observations, even for pixel observations. Spatial information in the observations will be lost.

ryanjulian:
Perhaps the solution here is for the observations to carry some type information, so we know what to do with them?

ahtsan:
Like paying a bool when constructing the TfEnv?

ryanjulian:
no. peppering your API with bools about what types are being processed is classic terrible software engineering.

I was suggesting that the akro.spaces object would know it was an image, or even better, i would just have some tag on a couple of the dims which instructs the algorithm not to flatten those dimensions.

alternatively, we could introduce an Image space.

ahtsan:
Wrap the environment with an Image space so that `OnPolicyVectorizedSampler` will not flatten the observations?

ryanjulian:
Yes, something like that.

ryanjulian:
closed by #906

PR #906, opened 2019-10-02
Replace CategoricalConvPolicy

lywong92:
- Remove all occurrences of CategoricalConvPolicy
- Rename CategoricalConvPolicyWithModel to CategoricalConvPolicy
- Create and remove integration test

Benchmark script is located in origin/benchmark_categorical_cnn_policy.

Results:
MemorizeDigits-v0_benchmark_ppo

MemorizeDigits-v0_benchmark_ppo_mean

Also tried running both versions in an atari environment PongNoFrameskip with PPO, but realized that this combination of environment and algorithm is not ideal for our testing:
image

As discussed, results from MemorizeDigits are sufficient to show that the layer implementation can be replaced with the model implementation.


examples/tf/trpo_cubecrash.py
```diff
@@ -0,0 +1,53 @@
...
+from garage.tf.envs import TfEnv
+from garage.tf.experiment import LocalTFRunner
+from garage.tf.policies import CategoricalConvPolicy
+
+
+def run_task(snapshot_config, *_):
+    """Run task."""
+    with LocalTFRunner(snapshot_config=snapshot_config) as runner:
+        env = TfEnv(normalize(gym.make('CubeCrash-v0')))
+        policy = CategoricalConvPolicy(env_spec=env.spec,
```
ahtsan: Might be better to use `CNN` for convention, so it would be `CategoricalCNNPolicy`. Please also edit other occurrence.

ryanjulian:
Can you run benchmarks using the following environments:
* MemorizeDigits-v0
* CubeCrash-v0
* CarRacing-v0
* Acrobot-v1^
* MountainCar-v0^
* CartPole-v1^
* LunarLander-v2^

^ - Using the wrappers [PixelObservationWrapper](https://github.com/openai/gym/blob/master/gym/wrappers/pixel_observation.py) and [FrameStack](https://github.com/openai/gym/blob/master/gym/wrappers/frame_stack.py) (n=4)

I omitted the Sparse and ScreenBecomesBlack variants of CubeCrash because those are meant to test off-policy algorithms and RNNs respectively.

First list found using this script:
```python
> import gym
> from gym.spaces import Box
> targets =  [s for s in gym.envs.registry.all() if s._entry_point != 'gym.envs.atari:AtariEnv' and type(s.make().observation_space) == Box and len(s.make().observation_space.shape) == 3]
> print(targets)
[EnvSpec(CubeCrashScreenBecomesBlack-v0),
 EnvSpec(CubeCrash-v0),
 EnvSpec(CarRacing-v0),
 EnvSpec(MemorizeDigits-v0),
 EnvSpec(CubeCrashSparse-v0)]
```

Second list using this additional query:
```python
> from gym.spaces import Discrete
> targets = [s for s in gym.envs.registry.all() if s._entry_point != 'gym.envs.atari:AtariEnv' and type(s.make().action_space) == Discrete and 'rgb _array' in s.make().metadata['render.modes']]
> print(targets)
[EnvSpec(MountainCar-v0),
 EnvSpec(CubeCrashScreenBecomesBlack-v0),
 EnvSpec(CartPole-v1),
 EnvSpec(CartPole-v0),
 EnvSpec(CubeCrash-v0),
 EnvSpec(Acrobot-v1),
 EnvSpec(LunarLander-v2),
 EnvSpec(MemorizeDigits-v0),
 EnvSpec(CubeCrashSparse-v0)]
```

* Along with your PR, can you add two examples (to examples/): one using one of the first set of environments, and another example with an environment using the wrappers? In your examples, you can also include the list of the other environments for which that example will work.

ahtsan:
> ```python
> Acrobot
> ```

`CarRacing` won't work since it has continuous action space?

ryanjulian:
@ahtsan gym documentation mentions that it can easily be controlled as discrete using 100% on/off actions.

krzentner:
I think even a small number of the Atari environments should be enough? It would be very surprising if we saw a large regression on some but not most of them.

ryanjulian:
@krzentner the issue is that PPO/on-policy algos can't train Atari in any reasonable number of steps, because of exploration (and this primitive is only applicable to on-policy algos).

krzentner:
> 
> 
> @krzentner the issue is that PPO/on-policy algos can't train Atari in any reasonable number of steps, because of exploration (and this primitive is only applicable to on-policy algos).

Ah, I see, my bad. The PPO paper does use Atari as a benchmark, but takes ~10M timesteps to train even the easiest envs.

The list you've given above is a good idea then (perhaps minus CarRacing, if that will take extra work).

ryanjulian:
@lywong92 did you have a chance to run the additional benchmarks?

ahtsan:
> @lywong92 did you have a chance to run the additional benchmarks?

Extra effort is needed for the environment wrapper - it can't be used right away in garage. We are working on that.

ryanjulian:
okay -- how much extra effort? if it's a lot, we can just test with the smaller set.

krzentner:
If we had a `ContinuousToDiscrete` (or `BangBangControl`) env wrapper this would be trivial. Perhaps we should add one? It might actually be useful in other environments (especially with the right clipping options).

krzentner:
After some discussion, implementing that wrapper correctly in a way that would interoperate with our policies seems difficult, so let's just skip `CarRacing` for now.

lywong92:
I ran CubeCrash-v0:

image
image

I had some trouble using the pixel wrapper in gym. I think that's solved but now I'm getting an error saying that the env can't be pickled. I wrapped the Acrobot env with PixelObservationWrapper, Grayscale, then StackFrames.

image

Do you happen to know if there's a way to resolve this?

ahtsan:
> I ran CubeCrash-v0:
> 
> image
> image
> 
> I had some trouble using the pixel wrapper in gym. I think that's solved but now I'm getting an error saying that the env can't be pickled. I wrapped the Acrobot env with PixelObservationWrapper, Grayscale, then StackFrames.
> 
> image
> 
> Do you happen to know if there's a way to resolve this?

Upodate: the `viewer` object in `gym.Env` is not pickleable. We hot fixed it by setting it to `None` before the pickling happens since it will always be created if it's None (drawback is it might get slow down a bit) - we are moving forward now.

ryanjulian:
re: the viewer -- this can be handled in GarageEnv. we already have special cases for closing the viewers (because gym doesn't do it properly)

ryanjulian:
when you post your results, can you use the confidence interval plots introduced by Yong, Utkarsh, and Anson? They make results much easier to interpret.

ahtsan:
> re: the viewer -- this can be handled in GarageEnv. we already have special cases for closing the viewers (because gym doesn't do it properly)

The issue we are having is that the viewer is not pickleable and raise error when `OnPolicyVectorizedSampler` tries to pickle it when setting up the sampling. I think `GarageEnv` handles it when the environment is closed, which is a separate issue.

ryanjulian:
@ahtsan i'm suggesting you can handle this in a similar way to how he handle closing environments.

lywong92:
I did a trial run of Acrobot-v1 with 100 iterations, and the average return never improved. I resized the image to a smaller dimension to make it run faster for the last run. I'm trying again without resizing and with more iterations.

I compared the original environment with the one after all the wrapping:
image

I'm wondering if the missing line would be a problem.

ryanjulian:
@lywong92 what you done should be enough, thought 100 iterations is a really low number. resizing should make it go faster and not really affect performance. i wouldn't expect to see signs of life until ~1000 or more (assuming the batch size is 4000). perhaps try running this with PPO rather than TRPO, which might be kinder to the CNNs. 

let's not get hung up trying to get the Pixel wrappers to work right now. the goal is to ensure that this primitive isn't any worse than before.

since it looks like CubeCrash and MemorizeDigits are the environments which work today, and they look good, can i think this is ready to merge.

can you make sure to add your benchmark script before merging?

ahtsan:
Another 10 runs for CubeCrash
CubeCrash-v0_benchmark_ppo_mean

lywong92:
> @lywong92 what you done should be enough, thought 100 iterations is a really low number. resizing should make it go faster and not really affect performance. i wouldn't expect to see signs of life until ~1000 or more (assuming the batch size is 4000). perhaps try running this with PPO rather than TRPO, which might be kinder to the CNNs.
> 
> let's not get hung up trying to get the Pixel wrappers to work right now. the goal is to ensure that this primitive isn't any worse than before.
> 
> since it looks like CubeCrash and MemorizeDigits are the environments which work today, and they look good, i think this is ready to merge.
> 
> can you make sure to add your benchmark script before merging?

Okay, will do!

lywong92:
Sorry about the lag. I rebased and am seeing a possible bug when I run examples. Currently working on it and will merge as soon as I get that figured out.

ahtsan:
This is the bug -- https://github.com/rlworkgroup/garage/issues/794

PR #916, opened 2019-10-04
Replace CategoricalGRUPolicy with model version

gitanshu:
Benchmark script at: https://github.com/rlworkgroup/garage/blob/fe08f2b97ca61838970290135b9689b8dc6ab271/tests/benchmarks/test_benchmark_ppo_categorical.py

Results: 

Assault-ramDeterministic-v4

assault1
assault2
assault3

Breakout-ramDeterministic-v4

breakout1
breakout2
breakout3

CartPole-v1

cartpole1
cartpole2
cartpole3

ChopperCommand-ramDeterministic-v4

chopper1
chopper2
chopper3

LunarLander-v2

ll1
ll2
ll3

Tutankham-ramDeterministic-v4

tutan1
tutan2
tutan3


naeioi:
Very nice result!

I am still struggling with LSTM. Can you share what change you have made into the model implementation to achieve these performance numbers? Since GRU used to suffer from unstably bad performance just like LSTM, especially in the breakout environment.

naeioi:
@ahtsan My observation is that for Breakout environment, the old policy can always converge to a best return of 1 regardless of the seed, while the new policy sometimes converges to a suboptimal solution. For example, seed 58 will give the new policy a bad performance. This also happens for LSTM.

(This figure is stolen from above)

image

gitanshu:
> Can you share what change you have made into the model implementation to achieve these performance numbers?

I didn't make any changes, just ran the benchmarks on @ahtsan's code. So he might be in a better position to answer this.

naeioi:
@ryanjulian So is it acceptable to be seed-dependent for breakout environment?

ryanjulian:
@naeioi please run your benchmark with 10 seeds on breakout and plot the results. Utkarsh and Anson can show you how to make a plot with a confidence interval. That will allow us to verify if the environment or the algorithm is the problem.

ryanjulian:
@naeioi let's focus on tracking down that bug in your PR, not this one.

You might have tracked down a systematic error in the recurrent policies, or you might have detected a feature of Breakout. In either case, it will probably apply to both GRU and LSTM (which are nearly identical in implementation), so no need to hold back both PRs.

Issue #698, opened 2019-05-29
use_gpu and use_tf apparently have no effect


ryanjulian:
```python
from garage.experiment import LocalRunner, run_experiment
from garage.np.baselines import LinearFeatureBaseline
from garage.tf.algos import TRPO
from garage.tf.envs import TfEnv
from garage.tf.policies import CategoricalMLPPolicy

def run_task(*_):
with LocalRunner() as runner:
env = TfEnv(env_name='CartPole-v1')

    policy = CategoricalMLPPolicy(
        name='policy', env_spec=env.spec, hidden_sizes=(32, 32))

    baseline = LinearFeatureBaseline(env_spec=env.spec)

    algo = TRPO(
        env_spec=env.spec,
        policy=policy,
        baseline=baseline,
        max_path_length=100,
        discount=0.99,
        max_kl_step=0.01)

    runner.setup(algo, env)
    runner.train(n_epochs=100, batch_size=4000,plot=True)
run_experiment(
run_task,
snapshot_mode='last',
seed=4,
n_parallel=4,
plot=True,
use_tf=False,
use_gpu=False

)
```
ref #697 

@xht033

ryanjulian:
@xht033

* `use_tf` is outdated and should be removed. it has no effect on whether your experiment uses tensorflow.
*`use_gpu` should be working -- how did you conclude that the GPU is being used?

xht033:
when I use nvidia-smi :
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 410.104      Driver Version: 410.104      CUDA Version: 10.0     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|===============================+======================+======================|
|   0  GeForce GTX 1060    Off  | 00000000:01:00.0  On |                  N/A |
| N/A   48C    P2    30W /  N/A |   5996MiB /  6069MiB |      5%      Default |
+-------------------------------+----------------------+----------------------+
                                                                               
+-----------------------------------------------------------------------------+
| Processes:                                                       GPU Memory |
|  GPU       PID   Type   Process name                             Usage      |
|=============================================================================|
|    0      1097      G   /usr/lib/xorg/Xorg                            18MiB |
[...]

PR #918, opened 2019-10-04
Fix use_gpu in experiment API

zequnyu:
- also removed `use_tf` as outdated

Solution refer to https://stackoverflow.com/questions/44500733/tensorflow-allocating-gpu-memory-when-using-tf-device-cpu0

### DIFF
--- a/src/garage/experiment/experiment.py
+++ b/src/garage/experiment/experiment.py
@@ -221,8 +221,7 @@ def run_experiment(method_call=None,
                    dry=False,
                    env=None,
                    variant=None,
-                   use_tf=False,
-                   use_gpu=False,
+                   force_cpu=False,
                    pre_commands=None,
                    **kwargs):
     """Serialize the method call and run the experiment using the
@@ -240,11 +239,8 @@ def run_experiment(method_call=None,
             commands without executing them.
         env (dict): Extra environment variables.
         variant (dict): If provided, should be a dictionary of parameters.
-        use_tf (bool): Used along with the Theano and GPU configuration
-            when using TensorFlow
-        use_gpu (bool): Whether the launched task is running on GPU.
-            This triggers a few configuration changes including certain
-            environment flags.
+        force_cpu (bool): Whether to set all GPU devices invisible
+            to force use CPU.
         pre_commands (str): Pre commands to run the experiment.
 
     """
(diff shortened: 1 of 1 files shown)


zequnyu:
This should address #698 as TensorFlow uses GPU by default if available. However for pytorch this is not true. I think the config for pytorch to use GPU should not be in `experiment.py`. I'm opening this for discussion. We might need a `TorchRunner` to do this?

ryanjulian:
I think that @krzentner is the best person to answer this.

src/garage/experiment/experiment.py
```diff
@@ -240,11 +239,8 @@ def run_experiment(method_call=None,
             commands without executing them.
         env (dict): Extra environment variables.
         variant (dict): If provided, should be a dictionary of parameters.
-        use_tf (bool): Used along with the Theano and GPU configuration
-            when using TensorFlow
         use_gpu (bool): Whether the launched task is running on GPU.
-            This triggers a few configuration changes including certain
-            environment flags.
+            It assumes proper configuration of CUDA to use GPU.
```
ryanjulian: it seems to me that most people with GPUs will want this to default to True, whereas people with only CPUs needs this to default to False.

AFAICT is harmless for it to default to `True` for CPU-based machines, so perhaps the default should be `True`?

Also, what if this accepts either a `bool` or a `tuple`/`list`? If it is a `True`/`False` we use all or not CPUs available. If it is a collection, we interpret that as whitelist of GPUs we are allowed to use. @krzentner I'd like to hear your thoughts.

src/garage/experiment/experiment.py
```diff
@@ -240,11 +239,8 @@ def run_experiment(method_call=None,
             commands without executing them.
         env (dict): Extra environment variables.
         variant (dict): If provided, should be a dictionary of parameters.
-        use_tf (bool): Used along with the Theano and GPU configuration
-            when using TensorFlow
         use_gpu (bool): Whether the launched task is running on GPU.
```
krzentner: Considering this flag's implementation, I actually think `use_gpu` is a pretty bad name. It doesn't ensure that the GPU is used for tensorflow or pytorch. For tensorflow, even if you set `use_gpu`, the gpu is only actually used if you have tensorflow-gpu installed. For pytorch, the gpu currently doesn't get used unless you manually move the policy to the gpu (by calling `.to(device)` or `.cuda()` on it).

Perhaps we should instead have a flag called `force_cpu`, which defaults to `False`?

ryanjulian:
@krzentner I think we should definitely delete run_experiment and do things how you describe.

The genesis of this PR is that we shouldn't have arguments in *code which exists today* which don't do anything or misleading, even if we plan to delete that code shortly. So I think if we focus on the short-term in this PR we can spend more time plotting to delete run_experiment (which doesn't have much meat left on its bones anymore, anyway).

zequnyu:
Does our CI support GPU testing? Otherwise this might need force-push.

ryanjulian:
It doesn't, but I don't see how that requires force-push.

ryanjulian:
I see why. Okay -- just ping me when it's ready to merge.

zequnyu:
@ryanjulian This is ready.

PR #919, opened 2019-10-04
Move n_epoch_cycles from Algorithm to Runner

zequnyu:
The new API allows to access it through `runner` in `train()`.

- unified name: `n_samples` to `n_epoch_cycles`


src/garage/np/algos/cem.py
```diff
@@ -106,16 +102,16 @@ def train_once(self, itr, paths):
         """
         paths = self.process_samples(itr, paths)
 
-        epoch = itr // self.n_samples
-        i_sample = itr - epoch * self.n_samples
+        epoch = itr // self.n_epoch_cycles
```
ryanjulian: uh is this what you intended in this algo? i think the original name was accurate. also, n_samples is an import hyperparam for CEM

zequnyu: Are you suggesting we should bring back it as a parameter for CEM?

ryanjulian: I am suggesting we should keep n_samples, because it is different from n_epoch_cycles.

src/garage/np/algos/cem.py
```diff
@@ -81,6 +77,8 @@ def train(self, runner):
             The average return in last epoch cycle.
 
         """
+        self.n_samples = runner.train_args.n_epoch_cycles
```
ryanjulian: so this behavior is surprising -- can you instead just make it a hyperparameter of the algorithm object? i don't think this is part of the runner interface.

src/garage/np/algos/cma_es.py
```diff
@@ -56,6 +53,8 @@ def train(self, runner):
             The average return in last epoch cycle.
 
         """
+        self.n_samples = runner.train_args.n_epoch_cycles
```
ryanjulian: please make this a hyperparameter instead.

src/garage/np/algos/cem.py
naeioi: Are you suggesting putting this to `BatchPolopt` class?

zequnyu: @ryanjulian I didn't get your point. My understanding is for CEM and CMA_ES, we set `n_samples` by defining `n_epoch_cycles` via runner interface. Would you clarify?

ryanjulian: I am proposing that we make n_samples a constructor argument and don't use `n_epoch_samples` to define `n_samples`, because I think that's confusing (they are not the same thing).

ryanjulian:
@zequnyu is this PR still alive?

zequnyu:
I’m working on another local branch to address #948, which will cover this PR. Will close this now.

PR #922, opened 2019-10-08
Replace GaussianGRUPolicy

### LINKED ISSUE #527: Refactor GaussianGRUPolicy to use garage.tf.model
* [x] GaussianGRUPolicy

nish21:
Replace GaussianGRUPolicy with model implementations. Benchmark results are attached.
Closes #527 
HalfCheetah
Hopper
InvertedDoublePendulum
InvertedPendulum
Reacher
Swimmer
Walker


ryanjulian:
Can you re-render these curves using the confidence interval method used by Utkarsh and Anson recently? It makes them much easier to interpret. 

Do you have any insight into what happened with `Swimmer` and `Walker2d`?

nish21:
Here are the confidence interval plots for the above graphs
HalfCheetah
Hopper
InvertedDoublePendulum
InvertedPendulum
Reacher
Swimmer
Walker2d


I ran Swimmer and Walker environments for 10 more seeds. This is the confidence interval plots for those
Swimmer
Walker2d

Probably the seeds in the original runs were not suitable for models. It seems like garage slightly better than models on average for walker environment, but the interval shows that they are nearly the same

ryanjulian:
LGTM

PR #924, opened 2019-10-08
Add tests and documentation for torch/VPG

yonghyuc:


src/garage/torch/algos/loss_function_utils.py
```diff
@@ -1,10 +1,25 @@
...
+
+    Args:
+        discount (float): Discount.
+        gae_lambda (float): Lambda used for generalized advantage estimation.
+        max_len (int): Maximum length of a single rollout.
+        baselines: The baseline.
+        rewards: The rewards.
+
+    Returns:
+        advantage values
```
ryanjulian: everything in a docstring requires a type, including the return value

tests/garage/torch/algos/test_vpg.py
```diff
@@ -0,0 +1,104 @@
...
+from garage.torch.algos import VPG
+from garage.torch.policies import GaussianMLPPolicy
+from tests.fixtures import snapshot_config
+
+
+class TestVPG:
+
+    def test_vpg_no_entropy(self):
+        """Test VPG with no_entropy."""
+        deterministic.set_seed(random.sample(range(100), 1)[0])
```
ryanjulian: please do not use any random data in tests, especially seeds. the whole point of setting the seed is so that the tests are deterministic.

src/garage/torch/algos/vpg.py
```diff
@@ -180,10 +180,11 @@ def _add_padding(self, tensor, length):
         return F.pad(tensor, (0, padding_length))
 
     def _get_policy_entropy(self, obs):
-        policy_entropy = self.policy.get_entropy(obs).sum()
-
         if self.stop_entropy_gradient:
-            policy_entropy.requires_grad = False
+            with torch.no_grad():
+                policy_entropy = self.policy.get_entropy(obs)
```
ryanjulian: please ensure that all class members are private (i.e. `_policy`) unless they are part of the public interface (in which case they should usually be a property, and not an attribute).

please apply this to all of the class members, not just policy.

src/garage/torch/algos/loss_function_utils.py
yonghyuc: fixed

tests/garage/torch/algos/test_vpg.py
yonghyuc: fixed

src/garage/torch/algos/vpg.py
yonghyuc: policy, baseline, discount, max_path_length, and n_samples are in the parent class (BatchPolopt). and BatchPolopt class is used in other tensorflow algorithms (ex npo)

But I can change every other class members in the VPG into private.
Is it okay to refactor BatchPolopt and every other related classes?

ryanjulian: ah -- no don't touch that one if it's going to lead you outside the torch tree.

yonghyuc: okay, just fixed variables inside the VPG class

src/garage/torch/algos/loss_function_utils.py
```diff
@@ -1,10 +1,25 @@
...
+    """
+    Calculate advantages.
+
+    Advantages are a discounted cumulative sum.
+
+    Args:
+        discount (float): Discount.
+        gae_lambda (float): Lambda used for generalized advantage estimation.
+        max_len (int): Maximum length of a single rollout.
+        baselines (torch.Tensor(float)): The baseline values.
```
ryanjulian: `torch.Tensor` is sufficient here and in the rest of this docstring.

yonghyuc: fixed

yonghyuc:
> This looks pretty good.
> 
> Please take a look at the [coverage report](https://codecov.io/gh/rlworkgroup/garage/compare/bdf371a0ae6207a1ee458d1ade00720501b3f059...4b2234ca11e83e8460ae43ca05f3d6d0c94e5b35/changes) and write tests to cover the remaining lines in red/yellow. I think that it should be quite easy to cover these lines, which is why I'm not approving yet.

add more testcases and now coverage is 98%
(After merging with pytorch branch, I can make the coverage 100% due to `predict_n` method)

tests/garage/torch/algos/test_vpg.py
```diff
@@ -0,0 +1,103 @@
...
+        with pytest.raises(ValueError, match=expected_msg):
+            VPG(**self._params)
+
+    # yapf: disable
+    def test_invalid_max_entropy_config_with_stop_entropy_gradient(self):
+        self._params['entropy_method'] = 'max'
+        self._params['center_adv'] = False
+        self._params['stop_entropy_gradient'] = False
+
+        expected_msg = 'stop_gradient should be True when entropy_method is max'  # noqa: E501
```
ryanjulian: you can match only a portion of the message to make sure it's relevant (e.g. `'stop_gradient'`), rather than the entire message. this makes less work for someone who wants to change the message a little.

tests/garage/torch/algos/test_vpg.py
```diff
@@ -0,0 +1,103 @@
...
+            'discount': 0.99,
+            'policy_lr': 1e-2
+        }
+
+    def teardown_method(self):
+        self._env.close()
+
+    def test_vpg_no_entropy(self):
+        """Test VPG with no_entropy."""
+        self._params['positive_adv'] = True
```
ryanjulian: the ValueError tests and performance tests can probably be handled easily with a [parameterized test](https://docs.pytest.org/en/latest/parametrize.html). i won't block your PR on this, but I will probably mention it next time.

for more complicated cases,  see `dict.update()` for a nice pattern for doing this:

```python
class Test:

   def setup_method(self):
        self.default_params = {
            'foo': 1.,
            'bar': 2.,
            'baz': 3.,
        }

    def test_case(self):
         self.default_params.update({
             'foo': 5.
        }
```

src/garage/torch/algos/loss_function_utils.py
```diff
@@ -1,10 +1,25 @@
...
 
 def compute_advantages(discount, gae_lambda, max_len, baselines, rewards):
-    """Calculate advantages."""
+    """
+    Calculate advantages.
+
+    Advantages are a discounted cumulative sum.
+
+    Args:
+        discount (float): Discount.
```
ryanjulian: discount factor, gamma

src/garage/torch/algos/loss_function_utils.py
```diff
@@ -1,10 +1,25 @@
...
+        gae_lambda (float): Lambda used for generalized advantage estimation.
```
ryanjulian: "Lambda, as used for Generalized Advantage Estimation (GAE)"

src/garage/torch/algos/loss_function_utils.py
```diff
@@ -1,10 +1,25 @@
...
+        max_len (int): Maximum length of a single rollout.
```
ryanjulian: around the garage codebase, this is usually called `max_path_length`

src/garage/torch/algos/loss_function_utils.py
```diff
@@ -1,10 +1,25 @@
...
+        baselines (torch.Tensor): The baseline values.
```
ryanjulian: "Baseline values, as estimated by a value function."

src/garage/torch/algos/loss_function_utils.py
```diff
@@ -1,10 +1,25 @@
-"""loss function utilities."""
+"""Differentiable utils typically used when computing loss functions."""
 import torch
 import torch.nn.functional as F
 
 
 def compute_advantages(discount, gae_lambda, max_len, baselines, rewards):
-    """Calculate advantages."""
+    """
+    Calculate advantages.
```
ryanjulian: "Calculate advantages using a baseline (value function) according to Generalized Advantage Estimation (GAE)"

src/garage/torch/algos/loss_function_utils.py
```diff
@@ -1,10 +1,25 @@
...
+
+    Args:
+        discount (float): Discount.
+        gae_lambda (float): Lambda used for generalized advantage estimation.
+        max_len (int): Maximum length of a single rollout.
+        baselines (torch.Tensor): The baseline values.
+        rewards (torch.Tensor): The rewards.
+
+    Returns:
+        advantages (torch.Tensor): advantage values
```
ryanjulian: Return values in docstrings are not named. This would be documented:

```python
Returns:
    torch.Tensor: Advantage values, as calculated using Generalized Advantage Estimation
```

src/garage/torch/algos/loss_function_utils.py
```diff
@@ -1,10 +1,25 @@
...
+    Calculate advantages.
+
+    Advantages are a discounted cumulative sum.
+
+    Args:
+        discount (float): Discount.
+        gae_lambda (float): Lambda used for generalized advantage estimation.
+        max_len (int): Maximum length of a single rollout.
+        baselines (torch.Tensor): The baseline values.
+        rewards (torch.Tensor): The rewards.
```
ryanjulian: You can omit articles, e.g. "The". This could read "Rewards" or "Time series of rewards"

src/garage/torch/algos/loss_function_utils.py
```diff
@@ -1,10 +1,25 @@
...
 import torch
 import torch.nn.functional as F
 
 
 def compute_advantages(discount, gae_lambda, max_len, baselines, rewards):
-    """Calculate advantages."""
+    """
+    Calculate advantages.
+
+    Advantages are a discounted cumulative sum.
```
ryanjulian: They are much more than this and calculating them using a GPU is fairly complicated. Please see the comprehensive documentation in the TF version for a guide of the level of detail required.

src/garage/torch/algos/loss_function_utils.py
```diff
@@ -1,10 +1,25 @@
...
+
+    Args:
```
ryanjulian: Please document comphrensivoely the expected shapes for baselines and rewards. It's very easy to provide this function with the wrong shape input tensor and get very surprising (and wrong) results.

tests/garage/torch/algos/test_vpg.py
yonghyuc: fixed!

tests/garage/torch/algos/test_vpg.py
yonghyuc: fixed

src/garage/torch/algos/loss_function_utils.py
yonghyuc: fixed

src/garage/torch/algos/loss_function_utils.py
yonghyuc: fixed

tests/garage/torch/algos/test_vpg.py
ryanjulian: i don't see the changes. did you forget to push?

src/garage/torch/algos/loss_function_utils.py
yonghyuc: fixed

src/garage/torch/algos/loss_function_utils.py
yonghyuc: fixed

src/garage/torch/algos/loss_function_utils.py
yonghyuc: fixed

src/garage/torch/algos/loss_function_utils.py
yonghyuc: fixed

src/garage/torch/algos/loss_function_utils.py
yonghyuc: fixed

src/garage/torch/algos/loss_function_utils.py
yonghyuc: @ryanjulian I add few more descriptions to `compute_advantages`
I chekced [this](https://github.com/rlworkgroup/garage/blob/09ec44b4adee0373d6cd7090f79a1e14ac56219c/src/garage/tf/misc/tensor_utils.py#L209) but I think this could be not enough.
What do you think?

src/garage/torch/algos/loss_function_utils.py
```diff
@@ -1,15 +1,61 @@
...
+        rewards:
+        [ [r_11, r_12, r_13, ... r_1n],
+          [r_21, r_22, r_23, ... r_2n],
+          ...
+          [r_m1, r_m2, r_m3, ... r_mn] ]
+
+    Args:
+        discount (float): RL discount factor (i.e. gamma).
+        gae_lambda (float): Lambda, as used for Generalized Advantage
+        Estimation (GAE).
```
ryanjulian: please indent multiline descriptions

tests/garage/torch/algos/test_vpg.py
```diff
@@ -0,0 +1,91 @@
...
+
+from garage.envs.base import GarageEnv
+from garage.experiment import deterministic
+from garage.experiment import LocalRunner
+from garage.np.baselines import LinearFeatureBaseline
+from garage.torch.algos import VPG
+from garage.torch.policies import GaussianMLPPolicy
+from tests.fixtures import snapshot_config
+
+# yapf: disable
```
ryanjulian: i think this disables yapf for everything below this line. please only disable yapf for blocks which need it.

tests/garage/torch/algos/test_vpg.py
```diff
@@ -0,0 +1,91 @@
...
+      'center_adv': True},
+        ValueError, 'center_adv'),
+    ({'entropy_method': 'max',
+      'center_adv': False,
+      'stop_entropy_gradient': False},
+        ValueError, 'entropy_method'),
+    ({'entropy_method': 'no_entropy',
+      'policy_ent_coeff': 1.0},
+        ValueError, 'policy_ent_coeff')
+]
```
ryanjulian: i think replacing this with `]  # yapf: disable` will achieve the affect you were going for.

src/garage/torch/algos/loss_function_utils.py
ryanjulian: if the comment you linked still applies to this implementation, feel free to copy and paste it.

src/garage/torch/algos/loss_function_utils.py
```diff
@@ -1,15 +1,61 @@
...
+    Calculate advantages using a baseline (value function) according to
+    Generalized Advantage Estimation (GAE)
+    The discounted cumulative sum can be computed using conv2d with filter.
+    filter:
+        [1, (discount * gae_lambda), (discount * gae_lambda) ^ 2, ...]
+        where the length is same with max_path_length.
+
+    Advantages are a discounted cumulative sum.
+
+    if paths is consisted with
```
ryanjulian: why mention paths when they are not an argument to the function?

src/garage/torch/algos/loss_function_utils.py
```diff
@@ -1,15 +1,61 @@
...
+        [ [p_11, p_12, p_13, ... p_1n],
+          [p_21, p_22, p_23, ... p_2n],
+          ...
+          [p_m1, p_m2, p_m3, ... p_mn] ]
+        (n : max_path_length, m: number of paths)
+
+    baselines and rewards are also has same shape.
```
ryanjulian: these are helpful, but you should also include a summary telling me the meaning of the elements of a numpy-style shape vector.

For instance (please check that this is correct):

```
"""python
Args:
    baselines (torch.Tensor): A 2D vector of value function estimates with
        shape (N, T), where N is the batch dimension (number of episodes) and
        T is the maximum path length experienced by the agent. If an episode
        terminates in fewer than T time steps, the remaining elements in that
        episode should be setto 0.

    rewards (torch.Tensor): A 2D vector of per-step rewards with shape (N, T),
        where N is the batch dimension (number of episodes) and T is the
        maximum path length experienced by the agent. If an episode terminates
        in fewer than T time steps, the remaining elements in that episode
        should be set to 0.
"""
```

src/garage/torch/algos/loss_function_utils.py
```diff
@@ -1,15 +1,61 @@
...
+        discount (float): RL discount factor (i.e. gamma).
+        gae_lambda (float): Lambda, as used for Generalized Advantage
+        Estimation (GAE).
+        max_path_length (int): Maximum length of a single rollout.
+        baselines (torch.Tensor): Baseline values, as estimated by a value
+        function. This should be 2D shape
+        rewards (torch.Tensor): Time series of rewards. This should be 2D
+        shape tensor
+
+    Returns:
```
ryanjulian: what is the shape of returns?

tests/garage/torch/algos/test_vpg.py
yonghyuc: add `#yapf: enable` on the bottom

tests/garage/torch/algos/test_vpg.py
yonghyuc: add `# yapf: enable` on the bottom

src/garage/torch/algos/loss_function_utils.py
yonghyuc: delete the comment

src/garage/torch/algos/loss_function_utils.py
yonghyuc: This is exactly correct explanation of this function! thank you!

src/garage/torch/algos/loss_function_utils.py
yonghyuc: add more information to the docstring

src/garage/torch/algos/loss_function_utils.py
ryanjulian: haha i wrote it, i should know ;)

PR #928, opened 2019-10-08
Support non-linearity function and module for PyTorch Module

yonghyuc:
Throw an error because of using non-linear functions in `torch.nn`.
Change into using non-linear funcitons in `torch.nn.functional` or `torch`

Previously, MultiHeadedMLPModule assumes that we are using non-linear function in torch.nn.
So we make a computational graph inside of \_\_init\_\_ and pass the type of the non-linear function to ModuleList. (`layer.add_module('non_linearity', hidden_nonlinearity)`)

However, for consistency with other modules, I changed the way we use the non-linear functions. 
(Similar with `MLPModule`, ) make linear layers only in \_\_init\_\_ and call the non-linear functions in the forward function. (`x = self._hidden_nonlinearity(x)`)

So now, we need to pass non-linear functions in the torch.nn.functional (ex. `torch.tanh` or `torch.nn.functional.relu`) instead of in the torch.nn (ex `torch.nn.Tanh`, `torch.nn.ReLU`)

### DIFF
--- a/src/garage/torch/modules/mlp_module.py
+++ b/src/garage/torch/modules/mlp_module.py
@@ -1,17 +1,16 @@
-"""
-MLP Module.
+"""MLP Module."""
 
-A Pytorch module composed only of a multi-layer perceptron (MLP), which maps
-real-valued inputs to real-valued outputs.
-"""
+from torch import nn
+from torch.nn import functional as F
 
-from torch import nn as nn
-from torch.nn import functional as F  # NOQA
+from garage.torch.modules.multi_headed_mlp_module import MultiHeadedMLPModule
 
 
-class MLPModule(nn.Module):
-    """
-    MLP Model.
+class MLPModule(MultiHeadedMLPModule):
+    """MLP Model.
+
+    A Pytorch module composed only of a multi-layer perceptron (MLP), which
+    maps real-valued inputs to real-valued outputs.
 
     Args:
         input_dim (int) : Dimension of the network input.
@@ -19,18 +18,18 @@ class MLPModule(nn.Module):
         hidden_sizes (list[int]): Output dimension of dense layer(s).
             For example, (32, 32) means this MLP consists of two
             hidden layers, each with 32 hidden units.
-        hidden_nonlinearity (callable): Activation function for intermediate
-            dense layer(s). It should return a torch.Tensor. Set it to
-            None to maintain a linear activation.
+        hidden_nonlinearity (callable or torch.nn.Module): Activation function
+            for intermediate dense layer(s). It should return a torch.Tensor.
+            Set it to None to maintain a linear activation.
         hidden_w_init (callable): Initializer function for the weight
             of intermediate dense layer(s). The function should return a
             torch.Tensor.
         hidden_b_init (callable): Initializer function for the bias
             of intermediate dense layer(s). The function should return a
             torch.Tensor.
-        output_nonlinearity (callable): Activation function for output dense
-            layer. It should return a torch.Tensor. Set it to None to
-            maintain a linear activation.
+        output_nonlinearity (callable or torch.nn.Module): Activation function
+            for output dense layer. It should return a torch.Tensor.
+            Set it to None to maintain a linear activation.
         output_w_init (callable): Initializer function for the weight
             of output dense layer(s). The function should return a
             torch.Tensor.
@@ -39,8 +38,6 @@ class MLPModule(nn.Module):
             torch.Tensor.
         layer_normalization (bool): Bool for using layer normalization or not.
 
-    Return:
-        The output torch.Tensor of the MLP
     """
 
     def __init__(self,
@@ -54,41 +51,21 @@ class MLPModule(nn.Module):
                  output_w_init=nn.init.xavier_normal_,
                  output_b_init=nn.init.zeros_,
                  layer_normalization=False):
-        super().__init__()
+        super().__init__(1, input_dim, output_dim, hidden_sizes,
+                         hidden_nonlinearity, hidden_w_init, hidden_b_init,
+                         output_nonlinearity, output_w_init, output_b_init,
+                         layer_normalization)
 
-        self._input_dim = input_dim
-        self._output_dim = output_dim
-        self._hidden_nonlinearity = hidden_nonlinearity
-        self._output_nonlinearity = output_nonlinearity
-        self._layer_normalization = layer_normalization
-        self._layers = nn.ModuleList()
+    # pylint: disable=arguments-differ
+    def forward(self, input_value):
+        """Forward method.
 
-        prev_size = input_dim
+        Args:
+            input_value (torch.Tensor): Input values with (N, *, input_dim)
+                shape.
 
-        for size in hidden_sizes:
-            layer = nn.Linear(prev_size, size)
-            hidden_w_init(layer.weight)
-            hidden_b_init(layer.bias)
-            self._layers.append(layer)
-            prev_size = size
+        Returns:
+            torch.Tensor: Output value
 
-        layer = nn.Linear(prev_size, output_dim)
-        output_w_init(layer.weight)
-        output_b_init(layer.bias)
-        self._layers.append(layer)
-
-    def forward(self, input_val):
-        """Forward method."""
-        x = input_val
-        for layer in self._layers[:-1]:
-            x = layer(x)
-            if self._hidden_nonlinearity is not None:
-                x = self._hidden_nonlinearity(x)
-            if self._layer_normalization:
-                x = nn.LayerNorm(x.shape[1])(x)
-
-        x = self._layers[-1](x)
-        if self._output_nonlinearity is not None:
-            x = self._output_nonlinearity(x)
-
-        return x
+        """
+        return super().forward(input_value)[0]
--- a/src/garage/torch/modules/multi_headed_mlp_module.py
+++ b/src/garage/torch/modules/multi_headed_mlp_module.py
@@ -1,41 +1,47 @@
 """MultiHeadedMLPModule."""
+import copy
+
+import torch
 import torch.nn as nn
 
 
 class MultiHeadedMLPModule(nn.Module):
-    """
-    MultiHeadedMLPModule Model.
+    """MultiHeadedMLPModule Model.
+
+    A PyTorch module composed only of a multi-layer perceptron (MLP) with
+    multiple parallel output layers which maps real-valued inputs to
+    real-valued outputs. The length of outputs is n_heads and shape of each
+    output element is depend on each output dimension
 
     Args:
-        n_heads (int) : Number of different output layers
-        input_dim (int) : Dimension of the network input.
-        output_dims (int, list, or tuple): Dimension of the network output.
+        n_heads (int): Number of different output layers
+        input_dim (int): Dimension of the network input.
+        output_dims (int or list or tuple): Dimension of the network output.
         hidden_sizes (list[int]): Output dimension of dense layer(s).
             For example, (32, 32) means this MLP consists of two
             hidden layers, each with 32 hidden units.
-        hidden_nonlinearity (callable): Activation function for intermediate
-            dense layer(s). It should return a torch.Tensor. Set it to
-            None to maintain a linear activation.
+        hidden_nonlinearity (callable or torch.nn.Module or list or tuple):
+            Activation function for intermediate dense layer(s).
+            It should return a torch.Tensor. Set it to None to maintain a
+            linear activation.
         hidden_w_init (callable): Initializer function for the weight
             of intermediate dense layer(s). The function should return a
             torch.Tensor.
         hidden_b_init (callable): Initializer function for the bias
             of intermediate dense layer(s). The function should return a
             torch.Tensor.
-        output_nonlinearity (callable, list, or tuple): Activation function for
-            output dense layer. It should return a torch.Tensor.
-            Set it to None to maintain a linear activation.
+        output_nonlinearities (callable or torch.nn.Module or list or tuple):
+            Activation function for output dense layer. It should return a
+            torch.Tensor. Set it to None to maintain a linear activation.
             Size of the parameter should be 1 or equal to n_head
-        output_w_init (callable, list, or tuple): Initializer function for the
-            weight of output dense layer(s). The function should return a
+        output_w_inits (callable or list or tuple): Initializer function for
+            the weight of output dense layer(s). The function should return a
             torch.Tensor. Size of the parameter should be 1 or equal to n_head
-        output_b_init (callable, list, or tuple): Initializer function for the
-            bias of output dense layer(s). The function should return a
+        output_b_inits (callable or list or tuple): Initializer function for
+            the bias of output dense layer(s). The function should return a
             torch.Tensor. Size of the parameter should be 1 or equal to n_head
         layer_normalization (bool): Bool for using layer normalization or not.
 
-    Return:
-        The list of output torch. Tensor of the MLP
     """
 
     def __init__(self,
@@ -43,7 +49,7 @@ class MultiHeadedMLPModule(nn.Module):
                  input_dim,
                  output_dims,
                  hidden_sizes,
-                 hidden_nonlinearity=nn.ReLU,
+                 hidden_nonlinearity=torch.relu,
                  hidden_w_init=nn.init.xavier_normal_,
                  hidden_b_init=nn.init.zeros_,
                  output_nonlinearities=None,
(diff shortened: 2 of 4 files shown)


tests/garage/torch/modules/test_multi_headed_mlp_module.py
```diff
@@ -78,15 +78,16 @@ def helper_make_inits(val):
...
-    module = MultiHeadedMLPModule(
-        n_heads=len(output_w_init_vals),
-        input_dim=input_dim,
-        output_dims=output_dim,
-        hidden_sizes=hidden_sizes,
-        hidden_nonlinearity=None,
-        hidden_w_init=nn.init.ones_,
-        output_nonlinearities=None,
-        output_w_inits=list(map(helper_make_inits, output_w_init_vals)))
+    module = MultiHeadedMLPModule(n_heads=len(output_w_init_vals),
```
ryanjulian: would this new version of the test fail if the old bug was still there? if not, please add a case which would have failed.

yonghyuc: So I put hidden and output non-linear function in the `test_multi_headed_mlp_module_with_different_output_dims` test case.

src/garage/torch/modules/multi_headed_mlp_module.py
```diff
@@ -61,55 +63,43 @@ def __init__(self,
...
 
     def _check_hidden_layer_parameter(self, var_name, var, n_heads):
         if isinstance(var, (list, tuple)):
             if len(var) == 1:
                 var = var * n_heads
             elif len(var) != n_heads:
-                msg = '{} should be either an integer or' \
-                      ' a collection of length n_heads ({}), but {} provided.'
-                raise ValueError(msg.format(var_name, n_heads, var))
+                raise ValueError(f'{var_name} should be matched with number '
```
ryanjulian: what was wrong with the original message?

src/garage/torch/modules/multi_headed_mlp_module.py
```diff
@@ -61,55 +63,43 @@ def __init__(self,
...
-            layer = nn.Sequential()
-            if layer_normalization:
-                layer.add_module('layer_normalization',
-                                 nn.LayerNorm(prev_size))
-
-            linear_layer = nn.Linear(prev_size, size)
-            hidden_w_init(linear_layer.weight)
-            hidden_b_init(linear_layer.bias)
-            layer.add_module('linear', linear_layer)
+        self._hidden_nonlinearity = hidden_nonlinearity
```
ryanjulian: are these also constructor args but unimplemented in the other MLPModules? It's very misleading for a constructor to have a parameter which is unused.

src/garage/torch/modules/multi_headed_mlp_module.py
yonghyuc: It wasn't wrong. I just want to make it more clear and use f-strings format in Python3

src/garage/torch/modules/multi_headed_mlp_module.py
yonghyuc: `self._hidden_nonlinearity` is used in forward function (line 112)

yonghyuc:
> I'm a little confused about this PR.
> 
> The title suggests that it's fixing a non-linearity in multi_headed_mlp_module (perhaps the default was wrong).
> 
> The content seems to rewrite how heads in this module are created, and also adds some features (layer norms, selectable output nonlinearities).
> 
> Can you add some context behind the change and explain what it does?

I updated the description above.
Basically, everything is same except how we handle the non-linear functions of hidden and output

src/garage/torch/modules/multi_headed_mlp_module.py
ryanjulian: f-strings aren't supported in Python 3.5, so we can't use them in this repo right now (we support Python 3.5+).

src/garage/torch/modules/multi_headed_mlp_module.py
```diff
@@ -61,55 +64,43 @@ def __init__(self,
...
 
-            if hidden_nonlinearity:
-                layer.add_module('non_linearity', hidden_nonlinearity)
+        self._layers = nn.ModuleList()
 
+        prev_size = input_dim
+        for size in hidden_sizes:
+            layer = nn.Linear(prev_size, size)
+            hidden_w_init(layer.weight)
+            hidden_b_init(layer.bias)
```
avnishn: @krzentner do we want to potentially setup the nn.ModuleList of each PyTorch module to contain the nonlinearities inside the module list. I think currently, they are explicitly run inside the forward function of each module, but they aren't explicitly apart of the modulelist (by using torch.functional nonlinearities instead of using nn.module nonlinearities)

src/garage/torch/modules/multi_headed_mlp_module.py
yonghyuc: fixed

src/garage/torch/modules/multi_headed_mlp_module.py
krzentner: Can we add the `torch.nn.functional` nonlinearities to the module list? That seems like a safer option, in case someone provides a `nn.Module` based nonlinearity that has parameters.

It would be pretty bad if the user can pass in a nonlinearity with parameters and we just forget to train those parameters.

tests/garage/torch/modules/test_multi_headed_mlp_module.py
```diff
@@ -128,14 +154,28 @@ def test_multi_headed_mlp_module_with_different_output_dims(
...
-            input_dim=input_dim,
-            output_dims=output_dim,
-            hidden_sizes=hidden_sizes,
-            hidden_nonlinearity=None,
-            hidden_w_init=nn.init.ones_,
-            output_nonlinearities=nonlinearity,
-            output_w_inits=list(map(helper_make_inits, w_init)),
-            output_b_inits=b_init)
+
+    expected_msg_template = 'should be either an integer or a collection of ' \
```
ryanjulian: please split lines with () per PEP8

ref: https://www.python.org/dev/peps/pep-0008/#maximum-line-length

src/garage/torch/modules/multi_headed_mlp_module.py
avnishn: To be clear: we will only be using torch.nn.nonlinearities in our modules, because we want them to be apart of the module list? I don't think one can add torch.functional non linearities to an nn.ModuleList @krzentner

src/garage/torch/modules/multi_headed_mlp_module.py
```diff
@@ -118,5 +109,15 @@ def forward(self, input_val):
         x = input_val
         for layer in self._layers:
             x = layer(x)
-
-        return [layer(x) for layer in self._output_layers]
+            if self._hidden_nonlinearity:
```
avnishn: just to be clear, what I'm trying to avoid is having to manually call the non linearities. I think they should just be defined as computations in the `nn.ModuleList`

src/garage/torch/modules/mlp_module.py
```diff
@@ -1,17 +1,17 @@
...
 
 A Pytorch module composed only of a multi-layer perceptron (MLP), which maps
 real-valued inputs to real-valued outputs.
 """
 
-from torch import nn as nn
+import copy
+
+from torch import nn
 from torch.nn import functional as F  # NOQA
```
ryanjulian: why noqa here?

src/garage/torch/modules/mlp_module.py
```diff
@@ -56,39 +54,87 @@ def __init__(self,
...
+
+        if isinstance(non_linear, type):
+            self.module = non_linear()
+        elif callable(non_linear):
+            self.module = copy.deepcopy(non_linear)
+        else:
+            raise ValueError(
+                'Non linear function {} is not supported'.format(non_linear))
+
+    # pylint: disable=arguments-differ
```
ryanjulian: why?

src/garage/torch/modules/mlp_module.py
```diff
@@ -56,39 +54,87 @@ def __init__(self,
...
+    def forward(self, input_value):
+        """Forward method.
+
+        Args:
+            input_value (torch.Tensor):  input values
+
+        Returns:
+            torch.Tensor:output value
```
ryanjulian: spacing here and in args is weird. please fix.

Argument descriptions should capitalize the first letter, just like an english sentence.

src/garage/torch/modules/mlp_module.py
```diff
@@ -56,39 +54,87 @@ def __init__(self,
...
+    def __init__(self, non_linear):
+        super().__init__()
+
+        if isinstance(non_linear, type):
+            self.module = non_linear()
+        elif callable(non_linear):
+            self.module = copy.deepcopy(non_linear)
+        else:
+            raise ValueError(
+                'Non linear function {} is not supported'.format(non_linear))
```
ryanjulian: Non-linear

src/garage/torch/modules/mlp_module.py
```diff
@@ -56,39 +54,87 @@ def __init__(self,
...
+        for layer in self._layers:
+            x = layer(x)
         return x
+
+
+class _NonLinearity(nn.Module):
+    """Wrapper class for non linear function or module.
+
+    Args:
+        non_linear (callable or type): non linear function or type to be
```
ryanjulian: Please write argument descriptions in the same style as english sentences

src/garage/torch/modules/mlp_module.py
```diff
@@ -56,39 +54,87 @@ def __init__(self,
...
-                x = self._hidden_nonlinearity(x)
-            if self._layer_normalization:
-                x = nn.LayerNorm(x.shape[1])(x)
+        if output_nonlinearity:
+            output_layer.add_module('non_linearity',
+                                    _NonLinearity(output_nonlinearity))
+        self._layers.append(output_layer)
+
+    # pylint: disable=arguments-differ
+    def forward(self, input_value):
```
ryanjulian: please document shape constraints (if any) in the parameter documentation. See the torch documentation for examples.

src/garage/torch/modules/mlp_module.py
```diff
@@ -56,39 +54,87 @@ def __init__(self,
...
+        else:
+            raise ValueError(
+                'Non linear function {} is not supported'.format(non_linear))
+
+    # pylint: disable=arguments-differ
+    def forward(self, input_value):
+        """Forward method.
+
+        Args:
+            input_value (torch.Tensor):  input values
```
ryanjulian: please document shape constraints and assumptions, if any.

src/garage/torch/modules/mlp_module.py
```diff
@@ -56,39 +54,87 @@ def __init__(self,
...
+
+        Returns:
+            torch.Tensor:output value
+
+        """
+        return self.module(input_value)
+
+    # pylint: disable=missing-return-doc, missing-return-type-doc
```
ryanjulian: why not just document it? it takes about as many characters as disabling it.

src/garage/torch/modules/mlp_module.py
```diff
@@ -1,17 +1,17 @@
...
+import copy
+
+from torch import nn
 from torch.nn import functional as F  # NOQA
 
 
 class MLPModule(nn.Module):
-    """
-    MLP Model.
+    """MLP Model.
```
ryanjulian: Please describe this in more detail.

src/garage/torch/modules/mlp_module.py
ryanjulian:
```python
        """See :class:`Object`.

        Returns:
            str: Unique representation of this object.

        """
```

src/garage/torch/modules/mlp_module.py
ryanjulian: you can remove this as soon as #1013 is submitted

src/garage/torch/modules/multi_headed_mlp_module.py
```diff
@@ -22,28 +24,32 @@ class MultiHeadedMLPModule(nn.Module):
...
+        output_b_inits (callable or list or tuple): Initializer function for
+            the bias of output dense layer(s). The function should return a
             torch.Tensor. Size of the parameter should be 1 or equal to n_head
         layer_normalization (bool): Bool for using layer normalization or not.
 
     Return:
         The list of output torch. Tensor of the MLP
+
+    Raises:
+        ValueError: n_head is equal to 1
```
ryanjulian: is this really an error? what if I am doing some hyperparameter search with several numbers of heads, and 1 is one of the cases?

ryanjulian: if you want, you can change MLPModule to:

```python
class MLPModule(MultiHeadedMLPModule):

    def __init__(self, *args, **kwargs):
        super().__init__(1, *args, **kwargs)
````

File: tests/garage/torch/modules/test_multi_headed_mlp_module.py (hunk 5 above)
Comment (yonghyuc) [reply tests/garage/torch/modules/test_multi_headed_mlp_module.py]: fixed

File: src/garage/torch/modules/mlp_module.py (hunk 9 above)
Comment (yonghyuc) [reply src/garage/torch/modules/mlp_module.py]: fixed

File: src/garage/torch/modules/mlp_module.py (hunk 10 above)
Comment (yonghyuc) [reply src/garage/torch/modules/mlp_module.py]: fixed

File: src/garage/torch/modules/mlp_module.py [hunk 17]
```diff
@@ -56,39 +54,87 @@ def __init__(self,
...
 
-        x = self._layers[-1](x)
-        if self._output_nonlinearity is not None:
-            x = self._output_nonlinearity(x)
+        Returns:
+            torch.Tensor: output value
 
+        """
+        x = input_value
+        for layer in self._layers:
```
krzentner: You should be able to replace this function with `return self._layers(input_values)`.

src/garage/torch/modules/mlp_module.py
krzentner: Perhaps we should actually globally allow not documenting `__methods__`? It's fairly obvious what `__repr__`, `__getstate__`, and `__setstate__` do.

In the mean time, yeah. Documenting looks cleaner than locally disabling the rule.

src/garage/torch/modules/multi_headed_mlp_module.py
krzentner: We should have removed the need for this loop, right?

src/garage/torch/modules/mlp_module.py
ryanjulian: before making this comment i searched (down to reading the pylint source code) and these's no current way to do this for the pylint extension which checks for arguments/returns.

i might make a PR one day, but until then we're stuck with this.

krzentner: :( I guess that's that.

ryanjulian: relevant: https://github.com/PyCQA/pylint/blob/master/pylint/extensions/docparams.py#L186

src/garage/torch/modules/mlp_module.py
yonghyuc: fixed

src/garage/torch/modules/multi_headed_mlp_module.py
yonghyuc: I think that is a good way!

src/garage/torch/modules/mlp_module.py
yonghyuc: self.layers is ModuleList, so I think we need to use for loop here.
Or we can use Sequential to compute the layers like that.

src/garage/torch/modules/mlp_module.py
yonghyuc: fixed

src/garage/torch/modules/multi_headed_mlp_module.py
yonghyuc: fixed

src/garage/torch/modules/mlp_module.py
yonghyuc: fixed

src/garage/torch/modules/mlp_module.py
```diff
@@ -1,36 +1,35 @@
...
+from garage.torch.modules.multi_headed_mlp_module import MultiHeadedMLPModule
 
 
-class MLPModule(nn.Module):
-    """
-    MLP Model.
+class MLPModule(MultiHeadedMLPModule):
+    """MLP Model.
+
+    A Pytorch module composed only of a multi-layer perceptron (MLP), which
```
ryanjulian: PyTorch

tests/garage/torch/modules/test_multi_headed_mlp_module.py
```diff
@@ -70,72 +37,129 @@
...
+    """
+    module = MultiHeadedMLPModule(n_heads=n_heads,
+                                  input_dim=input_dim,
+                                  output_dims=output_dim,
+                                  hidden_sizes=hidden_sizes,
+                                  hidden_nonlinearity=None,
+                                  hidden_w_init=nn.init.ones_,
+                                  output_nonlinearities=None,
+                                  output_w_inits=list(
+                                      map(_helper_make_inits,
```
ryanjulian: isn't this equivalent to
```python
[lambda x: nn.init.constant_(x, val) for val in output_w_init_vals]
```
?

src/garage/torch/modules/mlp_module.py
yonghyuc: fixed

tests/garage/torch/modules/test_multi_headed_mlp_module.py
yonghyuc: Actually, the result would not be the same.
I already tried your suggestion but it produced unexpected result.
You can check [this](https://stackoverflow.com/questions/6076270/lambda-function-in-list-comprehensions) 

If I want to use the list comprehension instead of the original code, I need to write
[(lambda y: (lambda x: nn.init.constant_(x, val)))(val) for val in output_w_init_vals]

ryanjulian:

ryanjulian:

PR #930, opened 2019-10-10
Remove flattening observation in samplers

ahtsan:
Flattening of observations should be decoupled to the consumer, e.g. algorithms. The assumption that the observations should be flattened is not valid in some use cases, e.g. when observations are images.

### DIFF
--- a/src/garage/tf/algos/batch_polopt.py
+++ b/src/garage/tf/algos/batch_polopt.py
@@ -50,7 +50,8 @@ class BatchPolopt(RLAlgorithm):
                  gae_lambda=1,
                  center_adv=True,
                  positive_adv=False,
-                 fixed_horizon=False):
+                 fixed_horizon=False,
+                 flatten_input=True):
         self.env_spec = env_spec
         self.policy = policy
         self.baseline = baseline
@@ -61,6 +62,7 @@ class BatchPolopt(RLAlgorithm):
         self.center_adv = center_adv
         self.positive_adv = positive_adv
         self.fixed_horizon = fixed_horizon
+        self.flatten_input = flatten_input
 
         self.episode_reward_mean = collections.deque(maxlen=100)
         if policy.vectorized:
@@ -100,7 +102,6 @@ class BatchPolopt(RLAlgorithm):
 
         """
         paths = self.process_samples(itr, paths)
-
         self.log_diagnostics(paths)
         logger.log('Optimizing policy...')
         self.optimize_policy(itr, paths)
@@ -164,10 +165,16 @@ class BatchPolopt(RLAlgorithm):
             returns.append(path['returns'])
 
         # make all paths the same length
-        obs = [path['observations'] for path in paths]
+        obs = [
+            self.env_spec.observation_space.flatten_n(path['observations'])
+            if self.flatten_input else path['observations'] for path in paths
+        ]
         obs = tensor_utils.pad_tensor_n(obs, max_path_length)
 
-        actions = [path['actions'] for path in paths]
+        actions = [
+            self.env_spec.action_space.flatten_n(path['actions'])
+            if self.flatten_input else paths['actions'] for path in paths
+        ]
         actions = tensor_utils.pad_tensor_n(actions, max_path_length)
 
         rewards = [path['rewards'] for path in paths]
--- a/src/garage/tf/algos/batch_polopt.py
+++ b/src/garage/tf/algos/batch_polopt.py
@@ -37,6 +37,9 @@ class BatchPolopt(RLAlgorithm):
             conjunction with center_adv the advantages will be
             standardized before shifting.
         fixed_horizon (bool): Whether to fix horizon.
+        flatten_input (bool): Whether to flatten input along the observation
+            dimension. If True, for example, an observation with shape (2, 4)
+            will be flattened to 8.
 
     """
 
@@ -139,6 +142,19 @@ class BatchPolopt(RLAlgorithm):
 
         max_path_length = self.max_path_length
 
+        if self.flatten_input:
+            paths = [
+                dict(
+                    observations=(self.env_spec.observation_space.flatten_n(
+                        path['observations'])),
+                    actions=(
+                        self.env_spec.action_space.flatten_n(  # noqa: E126
+                            path['actions'])),
+                    rewards=path['rewards'],
+                    env_infos=path['env_infos'],
+                    agent_infos=path['agent_infos']) for path in paths
+            ]
+
         if hasattr(self.baseline, 'predict_n'):
             all_path_baselines = self.baseline.predict_n(paths)
         else:
@@ -165,16 +181,10 @@ class BatchPolopt(RLAlgorithm):
             returns.append(path['returns'])
 
         # make all paths the same length
-        obs = [
-            self.env_spec.observation_space.flatten_n(path['observations'])
-            if self.flatten_input else path['observations'] for path in paths
-        ]
+        obs = [path['observations'] for path in paths]
         obs = tensor_utils.pad_tensor_n(obs, max_path_length)
 
-        actions = [
-            self.env_spec.action_space.flatten_n(path['actions'])
-            if self.flatten_input else paths['actions'] for path in paths
-        ]
+        actions = [path['actions'] for path in paths]
         actions = tensor_utils.pad_tensor_n(actions, max_path_length)
 
         rewards = [path['rewards'] for path in paths]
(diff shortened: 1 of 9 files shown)


ryanjulian:
Don't you need to add the flattening back into process_samples? Did you check that the on-policy algos didn't break?

ahtsan:
> Don't you need to add the flattening back into process_samples? Did you check that the on-policy algos didn't break?

All tests passed except `test_ray_batch_sampler`, since it requires one-hot representation for `Discrete` observation space. Do we need to add flattening back?

ryanjulian:
hmm is that a bug in RaySampler? What does `Discrete.sample()` return according to its API?

ryanjulian:
Please train at least one on-policy algorithm on one environment fully to make sure this is actually working. Make sure the environment has multi-dimensional observation space (otherwis `[obs] == flatten_n([obs])` and we're not testing anything :)

ahtsan:
> hmm is that a bug in RaySampler? What does `Discrete.sample()` return according to its API?

`env.sample()` returns a scalar value between 0 and n for a `env=Discrete(n)`. In garage we were calling `Discrete.flatten()` to make it one-hot.

ahtsan:
> Please train at least one on-policy algorithm on one environment fully to make sure this is actually working. Make sure the environment has multi-dimensional observation space (otherwis `[obs] == flatten_n([obs])` and we're not testing anything :)

Do you mean like `Box(m, n)`?

ahtsan:
Categorical MLP Policy 3 trials on

Assault-ramDeterminstic

CartPole
CartPole-v1_benchmark_mean

LunarLander
LunarLander-v2_benchmark_mean

ahtsan:
@ryanjulian ready for review - basically I move the flattening to `batch_poplot` and flatten the obs and actions according to a boolean flag.

src/garage/tf/misc/tensor_utils.py
```diff
@@ -48,32 +53,50 @@ def get_target_ops(variables, target_variables, tau=None):
...
     return tf.reshape(t, [-1] + list(t.shape[2:]), name=name)
 
 
 def flatten_batch_dict(d, name=None):
+    """Flatten batch dict."""
     with tf.name_scope(name, 'flatten_batch_dict', [d]):
         return {k: flatten_batch(v) for k, v in d.items()}
 
 
+def flatten_input_along_dim(t, dim, name='flatten_input_full'):
```
ryanjulian: how is this different from flatten_batch?

please include a full docstring

tests/integration_tests/test_flattening.py
```diff
@@ -0,0 +1,41 @@
...
+from garage.tf.envs import TfEnv
+from garage.tf.experiment import LocalTFRunner
+from garage.tf.policies import CategoricalMLPPolicy
+from tests.fixtures import snapshot_config, TfGraphTestCase
+from tests.fixtures.envs.wrappers import ReshapeObservation
+
+
+class TestResume(TfGraphTestCase):
+
+    def test_ppo_pendulum_flatten_input(self):
```
ryanjulian: maybe just put this along with all the other PPO tests?

src/garage/tf/algos/npo.py
```diff
@@ -214,17 +224,20 @@ def _build_inputs(self):
...
+                                            shape=[None] * 2 + list(shape),
+                                            name='policy_old_%s' % k)
                 for k, shape in policy_dist.dist_info_specs
             }
             policy_old_dist_info_vars_list = [
                 policy_old_dist_info_vars[k]
                 for k in policy_dist.dist_info_keys
             ]
 
+            if self.flatten_input:
```
ryanjulian: can you write out some comments here describing the shapes of the inputs and their flattened representation? 

how does this case contrast with the one immediately below it, which flattens the observation? why is the flattening done twice?

ryanjulian:
It will probably be simpler to merge this change if you move the flattening to process_samples, rather than trying to do it in TF.

NPO probably needs a bit of a refactor with how it deals with flattening and valid time steps. This change just makes it more confusing.

If you do it in process samples, you can just move the function call over from the sampler and not worry about new functions, tests, documentation, etc.

src/garage/tf/algos/npo.py
ryanjulian: why is this step necessary if you already flattened the input in `process_samples`?

ahtsan: In `process_samples` the samples (training data / real values) are flattened, while the flattening here is for the placeholders.

tests/integration_tests/test_flattening.py
ahtsan: sure

src/garage/tf/algos/npo.py
ryanjulian: ah. i think it would be more clear if you did this at placeholder creation-time.

src/garage/tf/misc/tensor_utils.py
ahtsan: Will do.
In short, `flatten_batch` flattens the batch dimension while keeping the sample dimension, and this function is able to flatten the observation dimension, when the given `dim` is 2 (since we have [batch_dim, time_step, observation_dim] for samples in garage).

src/garage/tf/algos/npo.py
ryanjulian: ah -- this would be much more clear if it were placed near placeholder creation and generously commented.

is there a way we can make the placeholder this dimension in the first place, rather than post-processing it?

ahtsan: I thought it would be more straight forward if we put the boolean flag in `akro.space.to_tf_placeholder()`?

ryanjulian: yes -- but i'm trying to avoid bumping the akro version number. let's see if this feature is needed in more than one algorithm first.

remember the goal here is to get models in ASAP. NPO is gonna get rewritten to use tf.distributions soon, and that might let us get rid of (at least one kind of) some of the flattening.

ahtsan: How about we do minimal changes - pass the boolean flag to algo and do the checking in sampler by accessing `self.algo.is_flatten_input`, and flatten if the flag is on? We can move the checking back to algo when we rewrite later. Agree that the goal is to get models in ASAP.

ryanjulian: that change may be minimal but it introduces a bunch of technical debt (and now you need to test a new case in the sampler).

again, why can't the algorithm just have a flag which lets it decide whether to flatten the observations?

today:
1. sampler produces flattened observations
1.  algo unconditionally flattens observations in `process_samples`
1. `optimize_policy` expects flattened observations (e.g. placeholders are flat)

tomorrow:
1. samples produces unflattened observations
1.  algo conditionally flattens observations in `process_samples`
1. `optimize_policy` checks whether to expect flattened or unflattened observations and actions.

it seems you are stuck on step (3) of "tomorrow" complexity-wise. rather than adding a whole new function to flatten your placeholder, why not infer its shape directly from the space? you know the [batch, time] dimensions already anyway.

you can file an issue in akro to add a flattening option in the tensor creation API.

```python
if self.flatten_inputs:
    obs_var = tf.compat.v1.placeholder(
        name='obs',
        dtype=tf.float32,
        shape=[None, None, self.observation_space.flat_dim])
else:
    obs_var = self.observation_space.to_tf_placeholder(name='obs', batch_dims=2)
```

ahtsan: It's a good idea, and we can avoid the new function `flatten_input_along_dim`. Will do.

ahtsan:
@ryanjulian Please see the new changes.

src/garage/tf/algos/npo.py
```diff
@@ -98,6 +99,7 @@ def __init__(self,
                  use_neg_logli_entropy=False,
                  stop_entropy_gradient=False,
                  entropy_method='no_entropy',
+                 flatten_input=True,
```
ryanjulian: you need to update the docstring

ryanjulian:
you might need to remove your changes to tensor_utils to get past codecov.

src/garage/tf/algos/batch_polopt.py
```diff
@@ -138,6 +139,18 @@ def process_samples(self, itr, paths):
 
         max_path_length = self.max_path_length
 
+        if self.flatten_input:
+            paths = list(
```
krzentner: This would definitely be clearer as a list comprehension.

ryanjulian:
See the "Note": https://www.tensorflow.org/versions/r1.14/api_docs/python/tf/keras/layers/Dense

Seems like the flattening might not even be necessary in the future.

src/garage/tf/algos/batch_polopt.py
ahtsan: @krzentner fixed

PR #931, opened 2019-10-10
Add test for ContinuousMLPQFunction

ahtsan:
Add benchmark test for ContinuousMLPQFunction, and refactor the benchmark directory as a mirror of `src/`, i.e. the benchmark test for `src/garage/tf/policies/XXX.py` can be found at `tests/benchmarks/garage/tf/policies/test_XXX.py`


ryanjulian:
Can you please split this into 2 PRs? The two changes aren't really related.

ahtsan:
> Can you please split this into 2 PRs? The two changes aren't really related.

will do

PR #938, opened 2019-10-16
OnPolicyVectorizedSampler should always flatten actions

ahtsan:
This is a bug introduced by PR #930. Regardless of observation type, sampler should always flatten the action inputs
Also add a missing constructor arguments to `TRPO`.

### DIFF
--- a/examples/tf/trpo_cubecrash.py
+++ b/examples/tf/trpo_cubecrash.py
@@ -41,7 +41,8 @@ def run_task(snapshot_config, *_):
                     baseline=baseline,
                     max_path_length=100,
                     discount=0.99,
-                    max_kl_step=0.01)
+                    max_kl_step=0.01,
+                    flatten_input=False)
 
         runner.setup(algo, env)
         runner.train(n_epochs=100, batch_size=4000)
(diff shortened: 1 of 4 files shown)


krzentner:
If this fixes a broken test, I think we should submit it. However, in the long run I don't think this is the right way to handle flattening input. The fact that the observations are flattened has basically nothing to do with the `Algorithm`, and should really be in either the environment, the policy, or a wrapper of one of those.

ryanjulian:
Can you explain why the sampler should flatten actions. 

If a (family of) algorithms always wants this, they are free to do it in `process_samples`.

ahtsan:
> Can you explain why the sampler should flatten actions.
> 
> If a (family of) algorithms always wants this, they are free to do it in `process_samples`.

Sorry, I meant the one who is responsible to flatten it and `OnPolicyVectorizedSampler` used to be the one, but now we moved it to algorithms.

ryanjulian:
It fixes a broken example, which I think is equivalent.

@krzentner I agree that the best place is probably the policy/module. I don't think doing it in `process_samples` is particularly offensive. This is an improvement over the status quo, wherein the sampler was flattening all of the observations+actions, wreaking havok for image-based environments.

@ahtsan when you submit this, can you also file an issue to investigate if this can be moved into the primitives? It makes sense but I'm not sure how that would fit with the loss function.

ryanjulian:
Please also add a simple test which would have caught this.

PR #940, opened 2019-10-17
Update gym to 0.15.3

krzentner:
Despite what the diff might indicate, this version update is actually fairly small.

I believe (but have not tested) that the code changes will also work under the old gym version.


ryanjulian:
I ran the nightly test suite locally and got a ton of `AttributeErrors`, so I think gym has once again changed the attribute API.

PR #946, opened 2019-10-19
Add multi env wrapper

utkarshjp7:
This PR does following.

- Add `MultiEnvWrapper`: A wrapper that handles multiple envs and provides observations augmented with one-hot representation of tasks..
- Add tests for `MultiEnvWrapper`.
- Add example for `MultiEnvWrapper`.

### DIFF
--- a/src/garage/envs/wrappers/__init__.py
+++ b/src/garage/envs/wrappers/__init__.py
@@ -1,5 +1,4 @@
-"""
-gym.Env wrappers.
+"""gym.Env wrappers.
 
 Used to transform an environment in a modular way.
 It is also possible to apply multiple wrappers at the same
@@ -7,6 +6,7 @@ time.
 
 Example:
     StackFrames(GrayScale(gym.make('env')))
+
 """
 from garage.envs.wrappers.atari_env import AtariEnv
 from garage.envs.wrappers.clip_reward import ClipReward
--- a/examples/tf/multi_env_ppo.py
+++ b/examples/tf/multi_env_ppo.py
@@ -1,0 +1,56 @@
+#!/usr/bin/env python3
+"""This is an example to train multiple tasks with PPO algorithm."""
+import gym
+import tensorflow as tf
+
+from garage.envs import normalize
+from garage.envs.multi_env_wrapper import MultiEnvWrapper
+from garage.experiment import run_experiment
+from garage.np.baselines import LinearFeatureBaseline
+from garage.tf.algos import PPO
+from garage.tf.envs import TfEnv
+from garage.tf.experiment import LocalTFRunner
+from garage.tf.policies import CategoricalMLPPolicy
+
+
+def run_task(snapshot_config, *_):
+    """Run task.
+
+    Args:
+        snapshot_config (garage.experiment.SnapshotConfig): The snapshot
+            configuration used by LocalRunner to create the snapshotter.
+
+        _ (object): Ignored by this function.
+
+    """
+    with LocalTFRunner(snapshot_config=snapshot_config) as runner:
+        env1 = TfEnv(normalize(gym.make('Adventure-ram-v4')))
+        env2 = TfEnv(normalize(gym.make('Alien-ram-v4')))
+        env = MultiEnvWrapper([env1, env2])
+
+        policy = CategoricalMLPPolicy(
+            env_spec=env.spec,
+            hidden_nonlinearity=tf.nn.tanh,
+        )
+
+        baseline = LinearFeatureBaseline(env_spec=env.spec)
+
+        algo = PPO(env_spec=env.spec,
+                   policy=policy,
+                   baseline=baseline,
+                   max_path_length=100,
+                   discount=0.99,
+                   gae_lambda=0.95,
+                   lr_clip_range=0.2,
+                   policy_ent_coeff=0.0,
+                   optimizer_args=dict(
+                       batch_size=32,
+                       max_epochs=10,
+                       tf_optimizer_args=dict(learning_rate=1e-3),
+                   ))
+
+        runner.setup(algo, env)
+        runner.train(n_epochs=120, batch_size=2048, plot=False)
+
+
+run_experiment(run_task, snapshot_mode='last', seed=1)
--- a/src/garage/envs/__init__.py
+++ b/src/garage/envs/__init__.py
@@ -1,3 +1,5 @@
+"""Garage wrappers for gym environments."""
+
 from garage.envs.base import GarageEnv
 from garage.envs.base import Step
 from garage.envs.env_spec import EnvSpec
--- a/src/garage/envs/multi_env_wrapper.py
+++ b/src/garage/envs/multi_env_wrapper.py
@@ -1,0 +1,196 @@
+"""A wrapper env that handles multiple tasks from different envs.
+
+Useful while training multi-task reinforcement learning algorithms.
+It provides observations augmented with one-hot representation of tasks.
+"""
+
+import random
+
+import akro
+import gym
+import numpy as np
+
+
+def round_robin_strategy(num_tasks, last_task=None):
+    """A function for sampling tasks in round robin fashion.
+
+    Args:
+        num_tasks (int): Total number of tasks.
+        last_task (int): Previously sampled task.
+
+    Returns:
+        int: task id.
+
+    """
+    if last_task is None:
+        return 0
+
+    return (last_task + 1) % num_tasks
+
+
+def uniform_random_strategy(num_tasks, _):
+    """A function for sampling tasks uniformly at random.
+
+    Args:
+        num_tasks (int): Total number of tasks.
+        _ (object): Ignored by this sampling strategy.
+
+    Returns:
+        int: task id.
+
+    """
+    return random.randint(0, num_tasks - 1)
+
+
+class MultiEnvWrapper(gym.Wrapper):
+    """A wrapper class to handle multiple gym environments.
+
+    Args:
+        envs (list(gym.Env)):
+            A list of objects implementing gym.Env.
+        sample_strategy (function(int, int)):
+            Sample strategy to be used when sampling a new task.
+
+    """
+
+    def __init__(self, envs, sample_strategy=uniform_random_strategy):
+
+        self._sample_strategy = sample_strategy
+        self._num_tasks = len(envs)
+        self._active_task_index = None
+        self._observation_space = None
+
+        super().__init__(envs[0])
+
+        self._task_envs = []
+        for env in envs:
+            if (env.observation_space.shape !=
+                    self.env.observation_space.shape):
+                raise ValueError(
+                    'Observation space of all envs should be same.')
+            if env.action_space.shape != self.env.action_space.shape:
+                raise ValueError('Action space of all envs should be same.')
+            self._task_envs.append(env)
+
+        self.env.spec.observation_space = self.observation_space
+
+    @property
+    def num_tasks(self):
+        """Total number of tasks.
+
+        Returns:
+            int: number of tasks.
+
+        """
+        return len(self._task_envs)
+
+    @property
+    def task_space(self):
+        """Task Space.
+
+        Returns:
+            akro.Box: Task space.
+
+        """
+        one_hot_ub = np.ones(self.num_tasks)
+        one_hot_lb = np.zeros(self.num_tasks)
+        return akro.Box(one_hot_lb, one_hot_ub)
+
+    @property
+    def active_task_index(self):
+        """Index of active task env.
+
+        Returns:
+            int: Index of active task.
+
+        """
+        return self._active_task_index
+
+    @property
+    def observation_space(self):
+        """Observation space.
+
+        Returns:
+            akro.Box: Observation space.
+
+        """
+        task_lb, task_ub = self.task_space.bounds
+        env_lb, env_ub = self._observation_space.bounds
+        return akro.Box(np.concatenate([task_lb, env_lb]),
+                        np.concatenate([task_ub, env_ub]))
+
+    @observation_space.setter
+    def observation_space(self, observation_space):
+        """Observation space setter.
+
+        Args:
+            observation_space (akro.Box): Observation space.
+
+        """
+        self._observation_space = observation_space
+
+    @property
+    def active_task_one_hot(self):
+        """One-hot representation of active task.
+
+        Returns:
+            numpy.ndarray: one-hot representation of active task
+
+        """
+        one_hot = np.zeros(self.task_space.shape)
+        index = self.active_task_index or 0
+        one_hot[index] = self.task_space.high[index]
+        return one_hot
+
+    def reset(self, **kwargs):
+        """Sample new task and call reset on new task env.
+
+        Args:
+            kwargs (dict): Keyword arguments to be passed to gym.Env.reset
+
+        Returns:
+            numpy.ndarray: active task one-hot representation + observation
+
+        """
+        self._active_task_index = self._sample_strategy(
+            self._num_tasks, self._active_task_index)
+        self.env = self._task_envs[self._active_task_index]
+        obs = self.env.reset(**kwargs)
+        oh_obs = self._obs_with_one_hot(obs)
+        return oh_obs
+
+    def step(self, action):
+        """gym.Env step for the active task env.
+
+        Args:
+            action (object): object to be passed in gym.Env.reset(action)
+
+        Returns:
+            object: agent's observation of the current environment
+            float: amount of reward returned after previous action
+            bool: whether the episode has ended
+            dict: contains auxiliary diagnostic information
+
+        """
+        obs, reward, done, info = self.env.step(action)
+        oh_obs = self._obs_with_one_hot(obs)
+        info['task_id'] = self._active_task_index
+        return oh_obs, reward, done, info
+
+    def close(self):
+        """Close all task envs."""
+        for env in self._task_envs:
+            env.close()
+
+    def _obs_with_one_hot(self, obs):
+        """Concatenate active task one-hot representation with observation.
+
+        Args:
+            obs (numpy.ndarray): observation
+
+        Returns:
+            numpy.ndarray: active task one-hot + observation
+
+        """
+        oh_obs = np.concatenate([self.active_task_one_hot, obs])
+        return oh_obs
--- a/tests/garage/envs/test_multi_env_wrapper.py
+++ b/tests/garage/envs/test_multi_env_wrapper.py
@@ -1,0 +1,145 @@
+"""Tests for garage.envs.multi_env_wrapper"""
+import akro
+import numpy as np
+import pytest
+
+from garage.envs.multi_env_wrapper import (MultiEnvWrapper,
+                                           round_robin_strategy,
+                                           uniform_random_strategy)
+from garage.tf.envs import TfEnv
+
+
+class TestMultiEnvWrapper:
+    """Tests for garage.envs.multi_env_wrapper"""
+
+    def test_tasks_from_same_env(self):
+        """test init with multiple tasks from same env"""
+        envs = ['CartPole-v0', 'CartPole-v0']
+        mt_env = self._init_multi_env_wrapper(envs)
+        assert mt_env.num_tasks == 2
+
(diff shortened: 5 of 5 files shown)


src/garage/envs/multi_task_env.py
```diff
@@ -0,0 +1,139 @@
...
+
+from garage.envs import GarageEnv
+
+
+class MultiTaskEnv(GarageEnv):
+    """
+    A wrapper class to handle multi-task environments.
+
+    Args:
+        task_env_cls_dict (dict(str, garage.envs.GarageEnv)):
```
ryanjulian: This design is hard to read and shouldn't be necessary with pickling. I think we should just pass a list of objects implementing `gym.Env`.

@zhanpenghe  @krzentner WDYT?

src/garage/envs/multi_task_env.py
```diff
@@ -0,0 +1,139 @@
...
+        """Observation space."""
+        return self.active_env.observation_space
+
+    @property
+    def horizon(self):
+        """Horizon of currently active task env."""
+        return self.active_env.horizon
+
+    @property
+    def valid_sample_strategies(self):
```
ryanjulian: hmm this is a strange way of doing this. why not just allow the user to inject a sampling strategy at the constructor?

src/garage/envs/multi_task_env.py
```diff
@@ -0,0 +1,139 @@
...
+        """Task's Action space."""
+        return self.active_env.action_space
+
+    @property
+    def observation_space(self):
+        """Observation space."""
+        return self.active_env.observation_space
+
+    @property
+    def horizon(self):
```
ryanjulian: AFAIK this is from a very old version of gym and can be removed.

src/garage/envs/multi_task_env.py
```diff
@@ -0,0 +1,139 @@
...
+        return self._task_envs[self.active_task]
+
+    @property
+    def num_tasks(self):
+        """Total number of tasks."""
+        return len(self._task_envs)
+
+    @property
+    def action_space(self):
+        """Task's Action space."""
```
ryanjulian: since all of your envs must match in this regard, i think you should just store this and return it.

src/garage/envs/multi_task_env.py
```diff
@@ -0,0 +1,139 @@
...
+    def active_env(self):
+        """Environment of currently active task."""
+        return self._task_envs[self.active_task]
+
+    @property
+    def num_tasks(self):
+        """Total number of tasks."""
+        return len(self._task_envs)
+
+    @property
```
ryanjulian: unfortunately i think some parts of the gym API assume this is an attribute rather than a property. i think @gitanshu knows more.

src/garage/envs/multi_task_env.py
```diff
@@ -0,0 +1,139 @@
...
+    def horizon(self):
+        """Horizon of currently active task env."""
+        return self.active_env.horizon
+
+    @property
+    def valid_sample_strategies(self):
+        """Supported sample strategies."""
+        return self._valid_sample_strategies
+
+    def reset(self, **kwargs):
```
ryanjulian: hmm don't pass `**kwargs` unless that's part of the gym API.

src/garage/envs/multi_task_env.py
```diff
@@ -0,0 +1,139 @@
...
+        """Reset the active task env."""
+        return self.active_env.reset(**kwargs)
+
+    def sample_tasks(self, n_tasks=None):
```
ryanjulian: okay -- either task sampling is internal or it is external. it doesn't make sense for this to be a public method i sampling is an internal process.

src/garage/envs/multi_task_env.py
```diff
@@ -0,0 +1,139 @@
...
+            n_tasks = self.num_tasks
+
+        if self._sample_strategy == 'all':
+            tasks = list(range(n_tasks))
+        elif self._sample_strategy == 'random':
+            tasks = np.random.randint(0, self.num_tasks, size=n_tasks).tolist()
+
+        return tasks
+
+    def set_task(self, task):
```
ryanjulian: hmm this could interact in strange ways with task sampling.

src/garage/envs/multi_task_env.py
```diff
@@ -0,0 +1,139 @@
...
+        """Set the active task."""
+        self._active_task = task % len(self._task_envs)
+
+    def step(self, action):
+        """gym.Env step for the active task env."""
+        obs, reward, done, info = self.active_env.step(action)
+        return obs, reward, done, info
+
+    def render(self, *args, **kwargs):
```
ryanjulian: please use the gym API rather than just blindly passing args/kwargs.

src/garage/envs/multi_task_env.py
```diff
@@ -0,0 +1,139 @@
...
+        """Id of currently active task."""
+        return self._active_task
+
+    @property
+    def active_env(self):
+        """Environment of currently active task."""
+        return self._task_envs[self.active_task]
+
+    @property
+    def num_tasks(self):
```
ryanjulian: in some cases this could be infinity and we should keep this in mind in the API design

src/garage/envs/one_hot_multi_task_env.py
```diff
@@ -0,0 +1,45 @@
+"""A wrapper for MultiTaskEnv."""
```
ryanjulian: i think this should go into envs/wrappers rather than just envs.

tests/garage/envs/test_multi_task_env.py
```diff
@@ -0,0 +1,95 @@
...
+            tasks[task_name] = {
+                'env': TfEnv,
+                'kwargs': {
+                    'env_name': env
+                },
+                'args': []
+            }
+
+        task_env = {k: v['env'] for k, v in tasks.items()}
+        task_kwargs = {k: v['kwargs'] for k, v in tasks.items()}
```
ryanjulian: this is my point about the args/kwargs interface being awkward, a stanza like this will appear in every launcher which uses this wrapper and it's ugly and hard to read.

we can instead make the interface "give me a list of envs" and perhaps provide some lightweight helpers for constructing those envs. we also might want a helper which allows you to define lazy-evaluated environment stubs (so we don't have to construct a bunch of environments in the local process, just so we can pickle them and re-construct them in a remote process).

src/garage/envs/multi_task_env.py
zhanpenghe: @ryanjulian I am kind of confused by what you are saying? Can you elaborate more on this?

src/garage/envs/multi_task_env.py
zhanpenghe: I agree. Also please put some default ones.

src/garage/envs/multi_task_env.py
```diff
@@ -0,0 +1,139 @@
...
+
+    @property
+    def action_space(self):
+        """Task's Action space."""
+        return self.active_env.action_space
+
+    @property
+    def observation_space(self):
+        """Observation space."""
+        return self.active_env.observation_space
```
zhanpenghe: I think there are more things we need to do here. If the environments do not have a unified observation space, we should at least zero-pad them.

src/garage/envs/multi_task_env.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def __init__(self,
+                 task_env_cls_dict,
+                 task_args,
+                 task_kwargs,
+                 sample_strategy='all'):
+
+        self._task_envs = []
+        self._active_task = 0
+        self._valid_sample_strategies = ['all', 'random']
```
zhanpenghe: I think there are more than all and random? For example, random can be uniform or under other distributions.

src/garage/envs/multi_task_env.py
zhanpenghe: Yes. I think it is important to make a different between discrete and continuous.

tests/garage/envs/test_multi_task_env.py
zhanpenghe: I agree.

src/garage/envs/multi_task_env.py
```diff
@@ -0,0 +1,139 @@
...
+
+        return tasks
+
+    def set_task(self, task):
+        """Set the active task."""
+        self._active_task = task % len(self._task_envs)
+
+    def step(self, action):
+        """gym.Env step for the active task env."""
+        obs, reward, done, info = self.active_env.step(action)
```
zhanpenghe: I think extending `info` with task information would be super helpful. For example, a task name. Task onehot can also be helpful too.

src/garage/envs/multi_task_env.py
ryanjulian: I'm arguing that either (1) this environment samples tasks itself when it needs them (in which case this method should be private or not exist), or (2) it receives already-sampled collections of environments and this method should not be part of its implementation at all.

If we need an environment sampler, then that's a different object.

It's not clear this proxy object design (where the wrapper holds a bunch of envs inside of it and switches the "active" one) is necessary under KR's new sampler API, but I can see us commiting this version while we wait for the multi-task friendly API to get merged.

src/garage/envs/multi_task_env.py
ryanjulian: I think Utkarsh already encoded here that the environments all need to have unified spaces.

I can understand wanting a feature (or wrapper) where we automatically merge and pad mismatched spaces, but I don't think we should do that automatically. It could lead to really hard to find bugs for the most common case (which is matching spaces).

src/garage/envs/multi_task_env.py
utkarshjp7: I looked into the new sampler API in more detail and I see your concern now. 
How about we let `MultiTaskEnv` sample a new task internally after `step` is called `max_path_length` times?

tests/garage/envs/test_one_hot_multi_task_env.py
```diff
@@ -0,0 +1,56 @@
+import akro
+import numpy as np
+
+from garage.envs import OneHotMultiTaskEnv
+from garage.tf.envs import TfEnv
+
+
+class TestOneHptMultiTaskEnv:
```
ryanjulian: OneHot

src/garage/envs/multi_task_env.py
ryanjulian: I think Utkarsh has already specified above that all envs must have matching action/observation spaces.

It's easy to imagine wanting a feature which automatically pads/concatenates non-matching environments, but I don't think that should be done automatically here, because it could lead to some really hard-to-find bugs for the most common use case (everything matches).

src/garage/envs/multi_task_env.py
utkarshjp7: If we have an environment sampler then we can directly feed those sampled environments to KR's new sampler API. In that case we don't really need a `MultiTaskEnv`, right?. or am I missing something here?

ryanjulian: i don't think KR's design includes an environment sampler (that is, a module which chooses environments). @krzentner

krzentner:
After offline discussion (with Ryan), it looks like we should finish this change with a design similar to what it currently has. Specifically, that means that the current `MultiTaskEnv` should become `MultiEnvWrapper` (or a similar name). Otherwise, my previous comment can basically be ignored (for now).

The rationale is this:

We have several environment that have shared action and observation spaces (and no single environments with multiple tasks, as I assume before). We would like to be able to train existing algorithms on multiple tasks (which are represented by separate environments) using existing algorithms, without modifying those algorithms. Of course, since there's a finite enumeration of environments, there will be a discrete task space, but we don't have any alternative for now.

The algorithms that will be using this wrapper will still need to know the task they're currently operating under, so ` OneHotMultiTaskEnv` should probably be merged into `MultiTaskEnv` (or `MultiEnvWrapper`).

Anyways, sorry about the confusion.

src/garage/envs/multi_task_env.py
zhanpenghe: In a finite number of goals case, implicitly switching them in reset makes sense since we can process them in preprocessor. But for meta-RL algorithms, this is making things hard since the assumption of these algos is within a meta batch, we only sample one task. 

I suggest that we do task setting explicitly in the sampler using API like `sample_tasks`+`set_task`.

src/garage/envs/wrappers/multi_env_wrapper.py
```diff
@@ -0,0 +1,151 @@
...
+
+        new_task = (self._last_task + 1) % self._num_tasks
+        self._last_task = new_task
+        return new_task
+
+
+class UniformRandomStrategy(SampleStrategy):
+    """A sample stretegy class for sampling uniformly at random."""
+
+    def sample_tasks(self):
```
zhanpenghe: This is saying sample tasks so I think it should be able to sample more than one tasks?

src/garage/envs/multi_task_env.py
ryanjulian: can you explain the assertion 

> the assumption of these [meta-RL] algos is within a meta batch, we only sample one task

meta-RL algos don't have much meaning on a single task, and I don't know of any that only sample one task in a meta-batch.

src/garage/envs/wrappers/multi_env_wrapper.py
utkarshjp7: Its a typo. It should only sample one task.

src/garage/envs/multi_task_env.py
zhanpenghe: For example, for maml, we need a batch of data that only contains one task (normally sampled from a continuous space) to compute the inner adaptation.

zhanpenghe: Of course, for one optimization step, there are normally more than one meta batches.

ryanjulian: How do you define "task" and "meta-batch"?

My definitions are:
* task: A fully-defined MDP. This could be a goal-conditioned environment and a (single) goal, or a multi-task environment and a (single) reward function
* meta-batch: A set of multiple, different MDPs (tasks). Formally, a meta-batch is only the MDPs, but we usually use it to mean a set of samples from these MDPs and sometimes also labels (or contexts, e.g. goals) for the MDPs.

Using these definitions, it seems to me that for MAML training you need meta-batches (batches with data from multiple MDPs and the MDP labels themselves), and for MAML adaptation you really just need one (non-meta) batch from some task of your choice.

I would suggest that in an implementation of MAML, if you need both of these behaviors in a single process, you might want to create two different samplers for simplicity and efficiency. Alternatively, you could keep a single sampler and reconfigure it between the training and adaptation stages.

zhanpenghe: I agree with your definition of task but not the meta-batch. Here is my definition of meta-batch: it is a batch of data that is fed into the algorithms for optimization. Each meta-batch contains data collected with a single task (an MDP) and each optimization step uses K meta-batches to optimize the models. This definition is the one that we are using in meta-world paper and originally MAML implementation/paper

utkarshjp7:
@ryanjulian @krzentner @zhanpenghe Is this good to go, or does it require any more design changes?

src/garage/envs/wrappers/multi_env_wrapper.py
krzentner: We should probably rename this to `sample_task`.

tests/garage/envs/wrappers/test_multi_env_wrapper.py
```diff
@@ -0,0 +1,125 @@
+"""Tests for garage.envs.wrappers.multi_env_wrapper"""
+import akro
+import numpy as np
+import pytest
+
+from garage.envs.wrappers import MultiEnvWrapper, RoundRobinStrategy, \
+    UniformRandomStrategy
+from garage.tf.envs import TfEnv
+
+# Disabling pylint warnings since class name and method names are self-explanatory # noqa: E501
```
ryanjulian: please remove these

src/garage/envs/wrappers/multi_env_wrapper.py
```diff
@@ -0,0 +1,147 @@
...
+
+import random
+
+import akro
+import gym
+import numpy as np
+
+
+class SampleStrategy:  # pylint: disable=too-few-public-methods
+    """Base sample stretegy class.
```
ryanjulian: strategy

src/garage/envs/wrappers/multi_env_wrapper.py
```diff
@@ -0,0 +1,147 @@
...
+"""
+
+import random
+
+import akro
+import gym
+import numpy as np
+
+
+class SampleStrategy:  # pylint: disable=too-few-public-methods
```
ryanjulian: please make this a proper ABC

ryanjulian: I propose naming this SamplingStrategy

src/garage/envs/wrappers/multi_env_wrapper.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def __init__(self, num_tasks):
+        self._num_tasks = num_tasks
+
+    def sample_task(self):
+        """Sample new task."""
+        pass
+
+
+class RoundRobinStrategy(SampleStrategy):  # noqa: E501 # pylint: disable=too-few-public-methods
```
ryanjulian: i am open to disabling this rule, but i think it did a good thing by pointing out how little logic this class holds. The intent of this rule is to prevent the proliferation of tons of single-use classes (Python isn't Java, so usually there are far fewer classes).

What if you changed SamplingStrategy from a class to a function?

```python
def round_robin_sampling(num_tasks, last_task):
    if last_task is None:
        return 0

    return (last_task + 1) % num_tasks


def uniform_random_strategy(num_tasks, _):
    return random.randint(0, num_tasks - 1)
```

if sampling gets complicated-enough to require a class, we can revisit that.

tests/garage/envs/wrappers/test_multi_env_wrapper.py
ryanjulian: i don't think they should trigger errors under the latest pre-commit configuration

src/garage/envs/wrappers/__init__.py
```diff
@@ -1,24 +1,27 @@
...
     StackFrames(GrayScale(gym.make('env')))
+
 """
 from garage.envs.wrappers.atari_env import AtariEnv
 from garage.envs.wrappers.clip_reward import ClipReward
 from garage.envs.wrappers.episodic_life import EpisodicLife
 from garage.envs.wrappers.fire_reset import FireReset
 from garage.envs.wrappers.grayscale import Grayscale
 from garage.envs.wrappers.max_and_skip import MaxAndSkip
+from garage.envs.wrappers.multi_env_wrapper import MultiEnvWrapper, \
```
ryanjulian: split with (), not \ (PEP8)

src/garage/envs/wrappers/__init__.py
```diff
@@ -1,24 +1,27 @@
...
+    RoundRobinStrategy, UniformRandomStrategy
 from garage.envs.wrappers.noop import Noop
 from garage.envs.wrappers.resize import Resize
 from garage.envs.wrappers.stack_frames import StackFrames
 
 __all__ = [
     'AtariEnv', 'ClipReward', 'EpisodicLife', 'FireReset', 'Grayscale',
-    'MaxAndSkip', 'Noop', 'Resize', 'StackFrames'
+    'MaxAndSkip', 'Noop', 'Resize', 'StackFrames', 'MultiEnvWrapper',
```
ryanjulian: can you put these into `garage.envs.multi_task` instead?

ryanjulian: hmm. on second thought just put them into garage.envs for now. we will figure out how to name multi-task/meta-RL later

src/garage/envs/wrappers/__init__.py
```diff
@@ -1,24 +1,27 @@
...
+    'UniformRandomStrategy', 'RoundRobinStrategy'
```
ryanjulian: rather than putting these into package garage.envs, just keep them in the `garage.envs.multi_env` scope for now

PR #947, opened 2019-10-21
Set fixed conda version temporarily

gitanshu:
Bug: https://github.com/conda/conda/issues/9345
Due to this bug in the latest conda installer (4.7.12), our conda builds keep breaking on CI.

This PR sets miniconda to last working version (4.7.10) until the bug gets fixed.


ryanjulian:
can you do this in 2019.02 and 2018.10 as well? i think they are broken for the same reason.

PR #953, opened 2019-10-22
Upgrade TensorFlow to 1.15

ryanjulian:


src/garage/tf/algos/ddpg.py
```diff
@@ -58,6 +60,10 @@ class DDPG(OffPolicyRLAlgorithm):
 
     """
 
+    # pylint: disable=too-many-instance-attributes
+    # pylint: disable=too-many-arguments
```
ryanjulian: please don't do whole-file disables. if you think we should globally turn off some rules, let's talk about it in the comments.

src/garage/tf/policies/continuous_mlp_policy.py
```diff
@@ -45,8 +43,12 @@ class ContinuousMLPPolicy(Policy2):
             tf.Tensor.
         input_include_goal (bool): Include goal in the observation or not.
         layer_normalization (bool): Bool for using layer normalization or not.
+
     """
 
+    # pylint: disable=too-many-arguments
+    # pylint: disable=too-many-instance-attributes
```
ryanjulian: ditto

src/garage/torch/algos/ddpg.py
```diff
@@ -45,17 +44,19 @@ class DDPG(OffPolicyRLAlgorithm):
...
             clip_return].
         max_action (float): Maximum action magnitude.
         reward_scale (float): Reward scale.
         smooth_return (bool): Whether to smooth the return for logging.
-        name (str): Name of the algorithm shown in computation graph.
 
     """
 
+    # pylint: disable=too-many-instance-attributes
+    # pylint: disable=too-many-arguments
```
ryanjulian: ditto

src/garage/tf/algos/ddpg.py
```diff
@@ -4,7 +4,7 @@
 from dowel import logger, tabular
 import numpy as np
 import tensorflow as tf
-import tensorflow.contrib as tc
+from tensorflow import contrib as tc
```
ryanjulian: does `# pylint: disable=import-error` work instead? i would rather it be there with a comment to explain the flaw for the next person then just picking a magic import string which works.

src/garage/tf/algos/ddpg.py
```diff
@@ -229,6 +235,7 @@ def __getstate__(self):
 
     def __setstate__(self, state):
         """Object.__setstate__."""
+        # pylint: disable=attribute-defined-outside-init
```
ryanjulian: we should probably keep this rule on. please only do per-line disables, not per-file or per-function (unless it's an exceptional circumstance).

this stanza appears frequently when picking. we could either do a per-line disable for these methods, or change this to

```python
    self.__dict__.clear()   # is this needed on top of update()?
    self.__dict__.update(state)
```

@krzentner WDYT?

src/garage/torch/algos/ddpg.py
```diff
@@ -136,17 +137,18 @@ def train_once(self, itr, paths):
         ])
 
         last_average_return = np.mean(self.episode_rewards)
-        for train_itr in range(self.n_train_steps):
+        for _ in range(self.n_train_steps):
             if self.replay_buffer.n_transitions_stored >= self.min_buffer_size:  # noqa: E501
                 self.evaluate = True
                 samples = self.replay_buffer.sample(self.buffer_batch_size)
-                qf_loss, y, q, policy_loss = torch_to_np(
+                # pylint: disable=unbalanced-tuple-unpacking
```
ryanjulian: the static analyzer can't figure out how long the return tuple is supposed to be.

i think this might be avoided more robustly by updating the implementation of `torch_to_np` to:

```python
return (v.numpy() for v in value_in)
```

or even
```python
return map(lambda v: v.numpy(), value_in)
```

src/garage/torch/algos/ddpg.py
```diff
@@ -136,17 +137,18 @@ def train_once(self, itr, paths):
         ])
 
         last_average_return = np.mean(self.episode_rewards)
-        for train_itr in range(self.n_train_steps):
+        for _ in range(self.n_train_steps):
             if self.replay_buffer.n_transitions_stored >= self.min_buffer_size:  # noqa: E501
```
ryanjulian: just split the line with () as PEP8 says to do.

src/garage/torch/algos/ddpg.py
```diff
@@ -181,9 +183,9 @@ def train_once(self, itr, paths):
 
         return last_average_return
 
+    # pylint: disable=arguments-differ
```
ryanjulian: uhh this is a serious error. justification for disabling it?

src/garage/torch/algos/ddpg.py
```diff
@@ -212,6 +214,7 @@ def optimize_policy(self, itr, samples):
                       0. if self.clip_pos_returns else self.clip_return)
 
         y_target = rewards + (1.0 - terminals) * self.discount * target_qvals
+        # pylint: disable=no-member
```
ryanjulian: this is a serious error. justification?

utkarshjp7:
> is there a reason torch/ddpg and ContinuousMLPPolicy are touched?
> 
> i appreciate the fixes, but you can avoid the heartache by just leaving them alone.
> 
> i think that precommit's logic for finding affected files has started to pick up commits you rebase in. i think github's behavior with rebase commits may have changed.
> 
> i have had luck with the incantation
> 
> ```shell
> git fetch origin
> git rebase origin/master
> git push -f origin tf115
> ```
> 
> if line 3 complains, i think `git push -f --no-verify origin tf115` will do the trick (and not complain in the CI or on future commits)

I added `get_regularizable_vars` in  `ContinuousMLPPolicy`  to pass the new ddpg test you added and to make it possible to train PPO with decayed policy weights. I can create a new PR for that. There is no need to touch ddpg in this PR. We can revert the ddpg changes as well.

src/garage/tf/algos/ddpg.py
utkarshjp7: Ah..you are right. I didn't intend to do it globally.

src/garage/tf/algos/ddpg.py
```diff
@@ -126,7 +129,7 @@ def __init__(self,
                                    smooth_return=smooth_return)
 
     @overrides
-    def init_opt(self):
+    def init_opt(self):  # pylint: disable=too-many-locals
```
ryanjulian: this will be handled by #959

src/garage/tf/algos/ddpg.py
```diff
@@ -229,7 +232,7 @@ def __getstate__(self):
 
     def __setstate__(self, state):
         """Object.__setstate__."""
-        self.__dict__ = state
+        self.__dict__ = state  # pylint: disable=attribute-defined-outside-init
```
ryanjulian: just use `self.__dict__.update(state)`

src/garage/tf/algos/ddpg.py
```diff
@@ -296,22 +303,21 @@ def train_once(self, itr, paths):
                 self.epoch_qs = []
 
             self.success_history.clear()
-
         return last_average_return
 
     @overrides
-    def optimize_policy(self, itr, samples_data):
+    def optimize_policy(self, itr, samples_data):  # noqa: E501 # pylint: disable=too-many-locals
```
ryanjulian: see #959

src/garage/tf/algos/ddpg.py
```diff
@@ -4,14 +4,14 @@
...
-import tensorflow.contrib as tc
+import tensorflow.contrib as tc  # pylint: disable=import-error
 
 from garage.misc.overrides import overrides
 from garage.np.algos.off_policy_rl_algorithm import OffPolicyRLAlgorithm
 from garage.tf.misc import tensor_utils
 
 
-class DDPG(OffPolicyRLAlgorithm):
+class DDPG(OffPolicyRLAlgorithm):  # noqa: E501 # pylint: disable=too-many-instance-attributes
```
ryanjulian: see #959

PR #954, opened 2019-10-23
Check for complete docstrings

ryanjulian:


ryanjulian:
Waiting to see. I think it should make things less strict on average.

PR #958, opened 2019-10-24
Enable all pylint rules

ryanjulian:
We will decide as-we-go which ones to disable


setup.py
```diff
@@ -73,34 +73,34 @@
...
 ]
 
 with open('README.md') as f:
-    readme = f.read()
+    README = f.read()
 
 # Get the package version dynamically
 with open('VERSION') as v:
-    version = v.read().strip()
+    README = v.read().strip()
```
gitanshu: This rename got mixed up
```suggestion
    VERSION = v.read().strip()
```

PR #960, opened 2019-10-25
Use separate pylint configuration for unit tests

ryanjulian:
* Makes a separate .pylintrc for unit tests
* Configures .pre-commit-config.yaml to run two different pylint jobs
* Updates pylint to the latest version (better C extension handling)
* Disables several more rules for all code

### DIFF
--- a/tests/garage/.pylintrc
+++ b/tests/garage/.pylintrc
@@ -27,6 +27,19 @@ disable =
     too-many-locals,
     # Detection seems buggy or unhelpful
     duplicate-code,
+    # Rules disabled *for unit tests only*
+    attribute-defined-outside-init,
+    differing-param-doc,
+    differing-type-doc,
+    docstring-first-line-empty,
+    missing-docstring,
+    missing-param-doc,
+    missing-return-doc,
+    missing-return-type-doc,
+    missing-type-doc,
+    no-self-use,
+    protected-access,
+    too-few-public-methods,
 
 
 [REPORTS]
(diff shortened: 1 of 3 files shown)


tests/garage/.pylintrc
```diff
@@ -0,0 +1,52 @@
+[MASTER]
+# Check docstring completeness
+load-plugins = pylint.extensions.docparams, pylint.extensions.docstyle
+
+# Unit tests have a special configuration which is checked separately
+ignore = tests/garage
```
gitanshu: Is this being ignored? Otherwise wouldn't putting this in the tests .pylintrc ignore all tests?

ryanjulian: oh good point -- in this case it's harmless because tests/garage would have to be a relative path from `pwd` -- so this will ignore anything in `tests/garage/tests/garage/*` -- I will make a follow-up PR to fix this, since the CI takes so long.

gitanshu: Sure. I think this actually wouldn't work even if there was something in `tests/garage/tests/garage/`. Pylint's documentation says this about `ignore`
```
--ignore=<file[,file...]>
    Add files or directories to the blacklist. They should be base names, not paths.
```

So even when I do `pylint --ignore=garage/np garage` or `pylint --ignore=src/garage/np garage`, I still see output for that directory. Even `pylint --ignore=garage/np garage/np` doesn't seem ti ignore anything.

Or maybe I don't understand how to use the ignore parameter.

PR #961, opened 2019-10-25
Add missing unit test to tf/DDPG

ryanjulian:
The current configuration doesn't test weight decay.

### DIFF
--- a/src/garage/tf/policies/continuous_mlp_policy.py
+++ b/src/garage/tf/policies/continuous_mlp_policy.py
@@ -11,7 +11,7 @@ from garage.tf.policies import Policy
 
 
 class ContinuousMLPPolicy(Policy):
-    """ContinuousMLPPolicy
+    """Continuous MLP Policy Network.
 
     The policy network selects action based on the state of the environment.
     It uses neural nets to fit the function of pi(s).
@@ -42,6 +42,7 @@ class ContinuousMLPPolicy(Policy):
             tf.Tensor.
         input_include_goal (bool): Include goal in the observation or not.
         layer_normalization (bool): Bool for using layer normalization or not.
+
     """
 
     def __init__(self,
@@ -105,6 +106,9 @@ class ContinuousMLPPolicy(Policy):
             obs_var (tf.Tensor): Tensor input for symbolic graph.
             name (str): Name for symbolic graph.
 
+        Returns:
+            tf.Tensor: symbolic graph of the action.
+
         """
         with tf.compat.v1.variable_scope(self._variable_scope):
             return self.model.build(obs_var, name=name)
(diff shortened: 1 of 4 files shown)


tests/garage/tf/algos/test_ddpg.py
```diff
@@ -97,3 +96,45 @@ def test_ddpg_pendulum(self):
...
+                policy_weight_decay=0.99,
+                qf_weight_decay=0.99,
+                min_buffer_size=int(5e3),
+                exploration_strategy=action_noise,
+            )
+            runner.setup(algo, env)
+            last_avg_ret = runner.train(n_epochs=10,
+                                        n_epoch_cycles=20,
+                                        batch_size=100)
+            assert last_avg_ret > 5
```
utkarshjp7: We get average return of 13 without decayed weights, while we only get 6 when using decayed weights. is this expected? @ryanjulian  @ahtsan

ryanjulian: i'm not sure. what if you make it much less decayed (e.g. 0.999)

utkarshjp7: the decay parameters are actually regularization constants. 0 means no regularization (no decay), and 1 means max regularization. Setting those parameters to 0.01 gives the desired results.

ryanjulian: ah thanks. can you make that clear in the docstring?

src/garage/tf/policies/continuous_mlp_policy.py
```diff
@@ -159,6 +179,10 @@ def clone(self, name):
 
         Args:
             name (str): Name of the newly created policy.
+
+        Returns:
+            garage.tf.policies.continuous_mlp_policy: Clone of this object
```
ryanjulian: this is a package, not a class

i think you meant `garage.tf.policies.ContinuousMLPPolicy`

utkarshjp7: Ah, yes. I fixed it.

src/garage/tf/policies/continuous_mlp_policy.py
```diff
@@ -136,19 +139,36 @@ def get_actions(self, observations):
...
         flat_obs = self.observation_space.flatten_n(observations)
         actions = self._f_prob(flat_obs)
         actions = self.action_space.unflatten_n(actions)
         return actions, dict()
 
+    def get_regularizable_vars(self):
+        """Get regularizable weight variables under the Policy scope.
+
+        Returns:
+            list(tf.Variable): List of regularizable variables.
```
ryanjulian: `list[tf.Variable]`

PR #966, opened 2019-10-28
Replace tf dense layer with keras version

ryanjulian:
tf.core.layers.dense is deprecated, so we replace it with
tf.keras.layers.Dense.


ryanjulian:
On second look this may be more complex than I first imagined, because I'm not sure keras will accept the same types for initializers and non-linearities.

PR #967, opened 2019-10-29
Sampler should not flatten observations and actions

ahtsan:
To make API consistent among all samplers, samplers should not flatten the observation nor the actions, which is done by algorithms.

### DIFF
--- a/src/garage/sampler/utils.py
+++ b/src/garage/sampler/utils.py
@@ -29,14 +29,14 @@ def rollout(env,
             distribution.
 
     Returns:
-        worker_number(int): The worker number passed into this function.
-        observations(np.array): Non-flattened array of observations.
-        actions(np.array): Non-flattened array of actions.
-        rewards(np.array): Array of rewards of shape (timesteps, 1).
-        agent_infos(dict[str, np.array]): Dictionary of stacked, non-flattened
-            `agent_info`s.
-        env_infos(dict[str, np.array]): Dictionary of stacked, non-flattened
-            `env_info`s.
+        dict[str, np.ndarray or dict]: Dictionary, with keys:
+            * observations(np.array): Non-flattened array of observations.
+            * actions(np.array): Non-flattened array of actions.
+            * rewards(np.array): Array of rewards of shape (timesteps, 1).
+            * agent_infos(dict[str, np.ndarray]): Dictionary of stacked,
+                  non-flattened `agent_info`s.
+            * env_infos(dict[str, np.ndarray]): Dictionary of stacked,
+                  non-flattened `env_info`s.
 
     """
     observations = []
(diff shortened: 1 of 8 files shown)


src/garage/sampler/utils.py
```diff
@@ -111,6 +115,6 @@ def truncate_paths(paths, max_samples):
                 truncated_last_path[k] = tensor_utils.truncate_tensor_dict(
                     v, truncated_len)
             else:
-                raise NotImplementedError
+                raise NotImplementedError()
```
ryanjulian: can you raise a ValueError instead? and include an error message which specifies the valid keys versus the ones found?

naeioi:
Not directly related to this PR, but there is also inconsistency in flattening in primitives. For example, `continuous_mlp_policy_with_model.py` flattens observation
https://github.com/rlworkgroup/garage/blob/dbec2b6058a1c7a714cef97d865997ef357bce4b/src/garage/tf/policies/continuous_mlp_policy.py#L144
while `categorical_cnn_policy.py` does not
https://github.com/rlworkgroup/garage/blob/dbec2b6058a1c7a714cef97d865997ef357bce4b/src/garage/tf/policies/categorical_cnn_policy.py#L144

To make things even worse, there is a `flatten_input` argument in `batch_polopt.py`
https://github.com/rlworkgroup/garage/blob/dbec2b6058a1c7a714cef97d865997ef357bce4b/src/garage/tf/algos/batch_polopt.py#L145

We should make all these into a single switch for flattening.

ryanjulian:
@naeioi thanks for pointing this out.

i think that's a bug in ContinuousMLPPolicy then -- for now we have decided not to flatten in the primitives.

@ahtsan what do you think?

ahtsan:
@naeioi @ryanjulian Yes I think right now flattening is duplicate in primitive and algorithm - we should probably only do that in one place (which I think we chose to be algorithm). That said, we should add flattening support in all other algorithms (except `batch_polopt`).

Regarding `categorical_cnn_policy.py`, we don't want to flatten the observations since we want to keep spatial information for CNN primitives. In that case, I think we would achieve the same goal by pass `flatten_input=False` to algorithm when using CNN primitives. Does that sound right?

ryanjulian:
@ahtsan that sounds right to me. for now let's just make it consistent and we can consider more automatic designs later.

ryanjulian:
I added some tests to try to merge this for the release, but it turns out something is still quite broken (or my tests are just wrong).

In any case, this won't make it and we will have to backport the fix I think.

@ahtsan @krzentner should be interested.

ahtsan:
> I added some tests to try to merge this for the release, but it turns out something is still quite broken (or my tests are just wrong).
> 
> In any case, this won't make it and we will have to backport the fix I think.
> 
> @ahtsan @krzentner should be interested.

I think the issue is that the dummy policy doesn't have dist_info_keys to extract agent_info. I am working on a fix.

ryanjulian:
If you look at the broken test, I think the shape of `path['observations']` is wrong for a non-flat observation space.

ahtsan:
> If you look at the broken test, I think the shape of `path['observations']` is wrong for a non-flat observation space.

Do you mean this test doesn't pass?
```python
    def test_does_not_flatten(self):
        path = utils.rollout(self.env, self.policy, max_path_length=5)
        assert path['observations'][0].shape == (4, 4)
        assert path['actions'][0].shape == (2, 2)
```

ryanjulian:
Yes

PR #968, opened 2019-10-29
Replace garage.Categorical with tfp.Distribution.OneHotCategorical

ahtsan:
This PR does the following:
- Create new classes CategoricalMLPPolicy2 and StochasticPolicy2
- Create TRPO2, NPO2 which is compatible with the new changes
- Replace garage.Categorical with tfp.Distribution.OneHotCategorical
   in CategoricalMLPPolicy2
- Remove dist_info_sym in StochasticPolicy2
- Transit test for categorical
- Unit test for CategoricalMLPPolicy2
- New example trpo_cartpole2

In CategoricalMLPPolicy2, get_action still return distribution parameters
but it's only for API compatibility and not used.

Instead of creating new placeholders when building graphs and do
dist_info_sym, this PR uses the placeholders from policies. Therefore, two
copies of policies exist in the algorithm (old policy and new policy). As using
the placeholders from policies, flattening and valids filtering are performed on
the samples data, not on the tf.Tensor.

Purpose of this PR is not to get merged, but to suggest a way to refactor the algorithms
to use tfp.Distribution.

### DIFF
--- a/src/garage/tf/algos/batch_polopt2.py
+++ b/src/garage/tf/algos/batch_polopt2.py
@@ -1,9 +1,9 @@
 """Base class for batch sampling-based policy optimization methods."""
-
 import collections
 
 from dowel import logger, tabular
 import numpy as np
+import tensorflow as tf
 
 from garage.misc import special, tensor_utils
 from garage.np.algos import RLAlgorithm
@@ -11,7 +11,7 @@ from garage.sampler import OnPolicyVectorizedSampler
 from garage.tf.samplers import BatchSampler
 
 
-class BatchPolopt(RLAlgorithm):
+class BatchPolopt2(RLAlgorithm):
     """Base class for batch sampling-based policy optimization methods.
 
     This includes various policy gradient methods like VPG, NPG, PPO, TRPO,
@@ -56,6 +56,7 @@ class BatchPolopt(RLAlgorithm):
                  flatten_input=True):
         self.env_spec = env_spec
         self.policy = policy
+        self.old_policy = policy.clone('old_policy')
         self.baseline = baseline
         self.scope = scope
         self.max_path_length = max_path_length
@@ -103,7 +104,7 @@ class BatchPolopt(RLAlgorithm):
             paths (list[dict]): A list of collected paths.
 
         Returns:
-            float: Average return.
+            float: The average return in last epoch cycle.
 
         """
         paths = self.process_samples(itr, paths)
@@ -158,8 +159,7 @@ class BatchPolopt(RLAlgorithm):
                         self.env_spec.action_space.flatten_n(  # noqa: E126
                             path['actions'])),
                     rewards=path['rewards'],
-                    env_infos=path['env_infos'],
-                    agent_infos=path['agent_infos']) for path in paths
+                    env_infos=path['env_infos']) for path in paths
             ]
         else:
             paths = [
@@ -169,8 +169,7 @@ class BatchPolopt(RLAlgorithm):
                         self.env_spec.action_space.flatten_n(  # noqa: E126
                             path['actions'])),
                     rewards=path['rewards'],
-                    env_infos=path['env_infos'],
-                    agent_infos=path['agent_infos']) for path in paths
+                    env_infos=path['env_infos']) for path in paths
             ]
 
         if hasattr(self.baseline, 'predict_n'):
@@ -213,12 +212,6 @@ class BatchPolopt(RLAlgorithm):
 
         baselines = tensor_utils.pad_tensor_n(baselines, max_path_length)
 
-        agent_infos = [path['agent_infos'] for path in paths]
-        agent_infos = tensor_utils.stack_tensor_dict_list([
-            tensor_utils.pad_tensor_dict(p, max_path_length)
-            for p in agent_infos
-        ])
-
         env_infos = [path['env_infos'] for path in paths]
         env_infos = tensor_utils.stack_tensor_dict_list([
             tensor_utils.pad_tensor_dict(p, max_path_length) for p in env_infos
@@ -233,8 +226,12 @@ class BatchPolopt(RLAlgorithm):
         undiscounted_returns = [sum(path['rewards']) for path in paths]
         self.episode_reward_mean.extend(undiscounted_returns)
 
-        ent = np.sum(self.policy.distribution.entropy(agent_infos) *
-                     valids) / np.sum(valids)
+        valid_obs = tensor_utils.flatten_batch(obs)
+        valid_obs = tensor_utils.filter_valids(valid_obs, valids)
+
+        ent = np.sum(tf.get_default_session().run(
+            self.policy.distribution.entropy(),
+            feed_dict={self.policy.model.inputs: valid_obs})) / np.sum(valids)
 
         samples_data = dict(
             observations=obs,
@@ -243,7 +240,6 @@ class BatchPolopt(RLAlgorithm):
             baselines=baselines,
             returns=returns,
             valids=valids,
-            agent_infos=agent_infos,
             env_infos=env_infos,
             paths=paths,
             average_return=np.mean(undiscounted_returns),
--- a/src/garage/tf/algos/npo2.py
+++ b/src/garage/tf/algos/npo2.py
@@ -6,7 +6,9 @@ import tensorflow as tf
 from garage.misc import special
 from garage.misc.overrides import overrides
 from garage.misc.tensor_utils import concat_tensor_list
-from garage.tf.algos.batch_polopt import BatchPolopt
+from garage.misc.tensor_utils import filter_valids
+from garage.misc.tensor_utils import flatten_batch
+from garage.tf.algos.batch_polopt2 import BatchPolopt2
 from garage.tf.misc import tensor_utils
 from garage.tf.misc.tensor_utils import center_advs
 from garage.tf.misc.tensor_utils import compute_advantages
@@ -21,7 +23,7 @@ from garage.tf.misc.tensor_utils import positive_advs
 from garage.tf.optimizers import LbfgsOptimizer
 
 
-class NPO(BatchPolopt):
+class NPO2(BatchPolopt2):
     """Natural Policy Gradient Optimization.
 
     Args:
@@ -185,6 +187,7 @@ class NPO(BatchPolopt):
         tabular.record('{}/Entropy'.format(self.policy.name), np.mean(pol_ent))
 
         self._fit_baseline(samples_data)
+        self._policy_update_ops()
 
     @overrides
     def get_itr_snapshot(self, itr):
@@ -196,20 +199,18 @@ class NPO(BatchPolopt):
         )
 
     def _build_inputs(self):
-        observation_space = self.policy.observation_space
         action_space = self.policy.action_space
 
-        policy_dist = self.policy.distribution
-
         with tf.name_scope('inputs'):
-            if self.flatten_input:
-                obs_var = tf.compat.v1.placeholder(
-                    tf.float32,
-                    shape=[None, None, observation_space.flat_dim],
-                    name='obs')
-            else:
-                obs_var = observation_space.to_tf_placeholder(name='obs',
-                                                              batch_dims=2)
+            obs_var = self.policy.model.networks['default'].input
+            old_obs_var = self.old_policy.model.networks['default'].input
+            # Ops to assign new policy weights to old policy
+            target_update_op = tensor_utils.get_target_ops(
+                self.policy.get_global_vars(),
+                self.old_policy.get_global_vars())
+            self._policy_update_ops = tf.get_default_session().make_callable(
+                target_update_op)
+
             action_var = action_space.to_tf_placeholder(name='action',
                                                         batch_dims=2)
             reward_var = tensor_utils.new_tensor(name='reward',
@@ -232,21 +233,8 @@ class NPO(BatchPolopt):
                 policy_state_info_vars[k] for k in self.policy.state_info_keys
             ]
 
-            # old policy distribution
-            policy_old_dist_info_vars = {
-                k: tf.compat.v1.placeholder(tf.float32,
-                                            shape=[None] * 2 + list(shape),
-                                            name='policy_old_%s' % k)
-                for k, shape in policy_dist.dist_info_specs
-            }
-            policy_old_dist_info_vars_list = [
-                policy_old_dist_info_vars[k]
-                for k in policy_dist.dist_info_keys
-            ]
-
             # flattened view
             with tf.name_scope('flat'):
-                obs_flat = flatten_batch_tensor(obs_var, name='obs_flat')
                 action_flat = flatten_batch_tensor(action_var,
                                                    name='action_flat')
                 reward_flat = flatten_batch_tensor(reward_var,
@@ -254,9 +242,6 @@ class NPO(BatchPolopt):
                 valid_flat = flatten_batch_tensor(valid_var, name='valid_flat')
                 policy_state_info_vars_flat = flatten_batch_tensor_dict(
                     policy_state_info_vars, name='policy_state_info_vars_flat')
-                policy_old_dist_info_vars_flat = flatten_batch_tensor_dict(
-                    policy_old_dist_info_vars,
-                    name='policy_old_dist_info_vars_flat')
 
             # valid view
             with tf.name_scope('valid'):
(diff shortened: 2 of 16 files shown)


src/garage/tf/algos/batch_polopt.py
```diff
@@ -265,9 +272,20 @@ def init_opt(self):
...
+
+        Args:
+            itr (int): Iteration.
+
+        """
         raise NotImplementedError
 
     def optimize_policy(self, itr, samples_data):
-        """Optimize the policy using the samples."""
+        """Optimize the policy using the samples.
```
ryanjulian: can you mark this abstract?

src/garage/tf/algos/batch_polopt.py
```diff
@@ -265,9 +272,20 @@ def init_opt(self):
         raise NotImplementedError
```
ryanjulian: can you mark this abstract?

src/garage/tf/algos/batch_polopt2.py
```diff
@@ -0,0 +1,287 @@
...
+
+    def init_opt(self):
+        """Initialize the optimization procedure.
+
+        If using tensorflow, this may include declaring all the variables and
+        compiling functions.
+        """
+        raise NotImplementedError
+
+    def get_itr_snapshot(self, itr):
```
ryanjulian: `@abc.abstractmethod`

src/garage/tf/algos/npo2.py
```diff
@@ -0,0 +1,575 @@
...
+                 optimizer=None,
+                 optimizer_args=None,
+                 policy_ent_coeff=0.0,
+                 use_softplus_entropy=False,
+                 use_neg_logli_entropy=False,
+                 stop_entropy_gradient=False,
+                 entropy_method='no_entropy',
+                 flatten_input=True,
+                 name='NPO'):
+        self.name = name
```
ryanjulian: _name

src/garage/tf/algos/npo2.py
```diff
@@ -0,0 +1,575 @@
...
+        tabular.record('{}/dLoss'.format(self.policy.name),
+                       loss_before - loss_after)
+        tabular.record('{}/KLBefore'.format(self.policy.name),
+                       policy_kl_before)
+        tabular.record('{}/KL'.format(self.policy.name), policy_kl)
+        pol_ent = self.f_policy_entropy(*policy_opt_input_values)
+        tabular.record('{}/Entropy'.format(self.policy.name), np.mean(pol_ent))
+
+        self._fit_baseline(samples_data)
+        self._policy_update_ops()
```
ryanjulian: can you use a verb for this method, e.g. `self._run_policy_update_ops()`?

src/garage/tf/algos/npo2.py
```diff
@@ -0,0 +1,575 @@
...
+
+    @overrides
```
ryanjulian: remove _overrides

src/garage/tf/algos/npo2.py
```diff
@@ -0,0 +1,575 @@
...
+        )
+
+    def _build_inputs(self):
+        action_space = self.policy.action_space
+
+        with tf.name_scope('inputs'):
+            obs_var = self.policy.model.networks['default'].input
+            old_obs_var = self.old_policy.model.networks['default'].input
+            # Ops to assign new policy weights to old policy
+            target_update_op = tensor_utils.get_target_ops(
```
ryanjulian: hmm can you explain using this vs `self._old_policy.parameters = self._policy.parameters`? I am genuinely curious.

src/garage/tf/algos/npo2.py
```diff
@@ -0,0 +1,575 @@
...
+                self.policy.get_global_vars(),
+                self.old_policy.get_global_vars())
+            self._policy_update_ops = tf.get_default_session().make_callable(
```
ryanjulian: this will trigger a deprecation warning. i think you need to use `tf.compat.v1.get_default_session()`

src/garage/tf/algos/npo2.py
```diff
@@ -0,0 +1,575 @@
...
+            if self.positive_adv:
+                if self.policy.recurrent:
+                    adv = positive_advs(adv, eps)
+                else:
+                    adv_valid = positive_advs(adv_valid, eps)
+
+            # Calculate loss function and KL divergence
+            with tf.name_scope('kl'):
+                if self.policy.recurrent:
+                    kl = self.old_policy.distribution.kl_divergence(
```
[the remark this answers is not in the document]
ryanjulian: checked and this looks right.

src/garage/tf/algos/npo2.py
```diff
@@ -0,0 +1,575 @@
...
+                else:
+                    kl = self.old_policy.distribution.kl_divergence(
+                        self.policy.distribution)
+                    pol_mean_kl = tf.reduce_mean(kl)
+
+            # Calculate vanilla loss
+            with tf.name_scope('vanilla_loss'):
+                if self.policy.recurrent:
+                    ll = self.policy.distribution.log_prob(
+                        i.action_var, name='log_likelihood')
```
ryanjulian: i think you can probably remove many of these `name='foo`` args, since the tfp.distributions routines have names that make sense.

src/garage/tf/algos/npo2.py
```diff
@@ -0,0 +1,575 @@
...
+
+                # Maximize E[surrogate objective] by minimizing
+                # -E_t[surrogate objective]
+                if self.policy.recurrent:
+                    loss = -tf.reduce_sum(obj) / tf.reduce_sum(i.valid_var)
+                else:
+                    loss = -tf.reduce_mean(obj)
+
+            # Diagnostic functions
+            self.f_policy_kl = tensor_utils.compile_function(
```
ryanjulian: can these be changed into `make_callable`?

src/garage/tf/algos/npo2.py
```diff
@@ -0,0 +1,575 @@
...
+                        policy_entropy = policy_entropy_valid
+
+            # This prevents entropy from becoming negative for small policy std
+            if self._use_softplus_entropy:
+                policy_entropy = tf.nn.softplus(policy_entropy)
+
+            if self._stop_entropy_gradient:
+                policy_entropy = tf.stop_gradient(policy_entropy)
+
+        self.f_policy_entropy = tensor_utils.compile_function(
```
ryanjulian: `make_callable`?

src/garage/tf/algos/reps.py
```diff
@@ -256,34 +256,36 @@ def _build_inputs(self):
...
                     action_flat,
                     valid_flat,
                     name='action_valid')    # yapf: disable
-                policy_state_info_vars_valid = filter_valids_dict(
+                policy_state_info_vars_valid = filter_valids_tensor_dict(
                     policy_state_info_vars_flat,
                     valid_flat,
                     name='policy_state_info_vars_valid')
-                policy_old_dist_info_vars_valid = filter_valids_dict(
+                policy_old_dist_info_vars_valid = filter_valids_tensor_dict(
```
ryanjulian: is `policy_old_dist_info_vars` used? AKAIK this is only used to calculate KL divergence and/or log_likelihood under the old policy.

src/garage/tf/misc/tensor_utils.py
```diff
@@ -43,38 +67,112 @@ def get_target_ops(variables, target_variables, tau=None):
...
 
 
-def filter_valids_dict(d, valid, name=None):
-    with tf.name_scope(name, 'filter_valids_dict', [d, valid]):
-        return {k: filter_valids(v, valid) for k, v in d.items()}
+def filter_valids_tensor_dict(d, valid, name=None):
+    """Filter input with valids.
+
+    Args:
+        d (dict(tf.Tensor)): Input tensor to be flattened.
```
ryanjulian: `dict[tf.Tensor]`

src/garage/tf/misc/tensor_utils.py
```diff
@@ -43,38 +67,112 @@ def get_target_ops(variables, target_variables, tau=None):
...
+        valid (tf.Tensor): Tensor with element either 0 or 1. The value of
+            i-th element being 1 indicates i-th index is valid.
+        name (str): Operation name.
+
+    Returns:
+        dict(tf.Tensor): Flattened input.
```
ryanjulian: `dict[tf.Tensor]`

src/garage/tf/algos/npo2.py
ahtsan: When assigning parameters, the `model` class matches the variables by full name - it only works when pickling the same object. Unless we change the "name checking" methodology (e.g. match the name excluding primitive itself), we still have to manually assign the vars.

ahtsan: also since policy is not yet a `model`, we need to do `self._old_policy.model.parameters = self._policy.model.parameters`.

src/garage/tf/policies/base.py
```diff
@@ -230,14 +263,38 @@ def dist_info_sym(self, obs_var, state_info_vars, name='dist_info_sym'):
...
+        """
+
+    def dist_info(self, obs, state_infos):
+        """Distribution info.
+
+        Return the distribution information about the actions.
+
+        Args:
+            obs (tf.Tensor): observation values
+            state_infos (dict): a dictionary whose values should contain
```
ryanjulian: this is missing a return value?

src/garage/tf/policies/base.py
```diff
@@ -248,4 +305,5 @@ def dist_info(self, obs, state_infos):
             state_infos (dict): a dictionary whose values should contain
                 information about the state of the policy at the time it
                 received the observation
+
```
ryanjulian: missing return value?

src/garage/tf/policies/categorical_mlp_policy2.py
```diff
@@ -0,0 +1,226 @@
+"""CategoricalMLPPolicy with model."""
```
ryanjulian: i think you can remove with "with model"

src/garage/tf/algos/reps.py
ahtsan: This is being used - I only modifiy `npo` and this is `reps`. I only renamed the helper function here.

src/garage/tf/algos/npo2.py
ryanjulian: this is in the wrong function -- the function is `_build_inputs` but here are you making a function which copies params from one policy to another.

src/garage/tf/algos/npo2.py
ryanjulian: hmm doesn't this function actually save the parameters from `policy` into `old_policy`?

to me the name suggests that it's running the policy optimization. can you rename it to something more obvious?

src/garage/tf/algos/npo2.py
ryanjulian: is there a way we can do this without using an op? can we add or change something in `model`? what are the risks of removing the full name checking?

one of the points of `model` is to avoid this thing in tensorflow where i have to make an op for things which aren't feed-forward.

we really want an API which is like `self._old_policy.model.parameters = self._policy.model.parameters`. i think it is more important to make this use case nice than the pickling one (pickling is already nasty).

src/garage/tf/algos/npo2.py
```diff
@@ -0,0 +1,575 @@
...
+                else:
+                    ll = self.policy.distribution.log_prob(
+                        i.valid.action_var, name='log_likelihood')
+
+                    vanilla = ll * adv_valid
+
+            # Calculate surrogate loss
+            with tf.name_scope('surrogate_loss'):
+                if self.policy.recurrent:
+                    lr = self.policy.distribution.prob(
```
ryanjulian: this is equivalent to 

```python
lr = tf.exp(self.policy.distribution.log_prob(i.action_var) - tf.stop_gradient(self.old_policy.distribution.log_prob(i.action_var)))
```
and we generally use that form for numerical reasons.

src/garage/tf/algos/npo2.py
```diff
@@ -0,0 +1,575 @@
...
+                        kl * i.valid_var) / tf.reduce_sum(i.valid_var)
+                else:
+                    kl = self.old_policy.distribution.kl_divergence(
+                        self.policy.distribution)
+                    pol_mean_kl = tf.reduce_mean(kl)
+
+            # Calculate vanilla loss
+            with tf.name_scope('vanilla_loss'):
+                if self.policy.recurrent:
+                    ll = self.policy.distribution.log_prob(
```
ryanjulian: i think this is a good opportunity for you to add `log_prob` and `kl_divergence` to `StochasticPolicy2` (or have StochasticPolicy inherit from `tfp.Distribution` -- not sure which is better). it will make reading the algorithms much easier.

perhaps it will be something like 

```python
class Policy(Model):
    pass

class StochasticPolicy(tfp.Distribution, Policy):
    pass


ll = self.policy.log_prob(i.action_var)
lr = tf.exp(self.policy.log_prob(i.action_var) - self.old_policy.log_prob(i.action_var))
# etc...
```
```

File: src/garage/tf/algos/npo2.py [hunk 21]
```diff
@@ -0,0 +1,575 @@
...
+            obs_var = self.policy.model.networks['default'].input
+            old_obs_var = self.old_policy.model.networks['default'].input
+            # Ops to assign new policy weights to old policy
+            target_update_op = tensor_utils.get_target_ops(
+                self.policy.get_global_vars(),
+                self.old_policy.get_global_vars())
+            self._policy_update_ops = tf.get_default_session().make_callable(
+                target_update_op)
+
+            action_var = action_space.to_tf_placeholder(name='action',
```
ryanjulian: there are 4 different ways of creating a tensor here. can we deleted 2 or 3 of them?

src/garage/tf/algos/npo2.py
```diff
@@ -0,0 +1,575 @@
...
+
+        return policy_loss_inputs, policy_opt_inputs
+
+    def _build_policy_loss(self, i):
+        policy_entropy = self._build_entropy_term(i)
+        rewards = i.reward_var
+
+        if self._maximum_entropy:
+            with tf.name_scope('augmented_rewards'):
+                rewards = i.reward_var + self.policy_ent_coeff * policy_entropy
```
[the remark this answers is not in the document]
ryanjulian: i think this should be called augmented_rewards

src/garage/tf/algos/npo2.py
```diff
@@ -0,0 +1,575 @@
...
+            if self.policy.recurrent:
+                policy_neg_log_likeli = -pol_dist.log_prob(
+                    i.action_var, name='policy_log_likeli')
+
+                if self._use_neg_logli_entropy:
+                    policy_entropy = policy_neg_log_likeli
+                else:
+                    policy_entropy = pol_dist.entropy()
+            else:
+                policy_neg_log_likeli_flat = -pol_dist.log_prob(
```
ryanjulian: this is where i don't understand how your policy could possibly have the right shape for `pol_dist.log_prob(i.flat.action_var)` because `i.flat.action_var` needs to be shape `(batch * max_path_length)` so that it can be reshaped into `(batch, max_path_length)`. If you already flattened and filtered the observations, how could your policy be outputting the right shape?

src/garage/tf/algos/npo2.py
```diff
@@ -0,0 +1,575 @@
...
+            itr=itr,
+            policy=self.policy,
+            baseline=self.baseline,
+        )
+
+    def _build_inputs(self):
+        action_space = self.policy.action_space
+
+        with tf.name_scope('inputs'):
+            obs_var = self.policy.model.networks['default'].input
```
ryanjulian: what is the shape of this placeholder?

src/garage/tf/algos/npo2.py
```diff
@@ -0,0 +1,575 @@
...
+            old_obs_var = self.old_policy.model.networks['default'].input
```
ryanjulian: now is a good time to step back and ask whether we should have models create their own placeholders or accept placeholder inputs to `build` instead.

it already seems like he `model.networks` tracking only needs to exist as long as we are still using the told `_sym` style ops on TF primitives.

src/garage/tf/algos/npo2.py
ahtsan: You are right, this will break. Looks like model should accept a placeholder instead of creating one, so we can pass a [batch, path, action_dim] placeholder and keep the original flattening / valid filtering logic. And we don't need the `model.input` as we are passing the placeholder.

ahtsan:
> eliminate a ton of the flattening/valid-filtering/`if recurrent` code by harmonizing the usage of different shapes and inputs.

`eliminate a ton of the flattening/valid-filtering/if recurrent code by harmonizing the usage of different shapes and inputs.` Do you mean like feeding a [batch, path, dim] placeholder if policy is recurrent, otherwise [batch * path, dim] so that we don't need the `i.obs_var`/`i.flat.obs_var` separation? Same applies to `action_var` so we can directly do `pol_dist.log_prob(action_var)` without the need of checking if recurrent? For valid filtering, looks like we still need that.

ryanjulian:
Feeding `[batch, path, *dim]` is an important feature of our NPO *for both recurrent and non-recurrent* because it allows us to differentiate through `compute_adv` and the augmented rewards (entropy).

The only reason we do a ton of flattening and filtering is that many of the ops on the primitives assume a `[batch, flat(*dim)]`-shaped input.

It might be much simpler, for instance, if we use [ragged tensors](https://www.tensorflow.org/guide/ragged_tensor).

ryanjulian:
@ahtsan is this PR still alive?

ahtsan:
No it's not, I am using the comments for reference though. I am breaking this PR into small PRs so I think it's fine to close this at this moment.

PR #977, opened 2019-10-30
Add tests for tf/ConjugateGradientOptimizer

utkarshjp7:
This PR does following.

- Remove `krylov.py` file. Only useful function in the file is moved to `conjugate_gradient_optimizer.py`
- Refactor and add documentation to `conjugate_gradient_optimizer.py`.
- Add tests for hessian-vector product classes, and conjugate gradient iterative solver.


src/garage/tf/optimizers/conjugate_gradient_optimizer.py
```diff
@@ -1,27 +1,109 @@
...
+class PerlmutterHvp(HessianVectorProduct):
+    """This class computes hessian-vector product using Pearlmutter's algorithm.
+
+    `Pearlmutter, Barak A. "Fast exact multiplication by the Hessian." Neural
+     computation 6.1 (1994): 147-160.`
+
+    """
+
+    def update_opt(self, f, target, inputs, reg_coeff, name=None):
+        """Update class properties.
```
ryanjulian: can you make this more specific?

src/garage/tf/optimizers/conjugate_gradient_optimizer.py
```diff
@@ -1,27 +1,109 @@
...
+            xs = tuple(self.target.flat_to_params(v, trainable=True))
+            ret = sliced_fun(self.opt_fun['f_hx_plain'], self._num_slices)(
+                inputs, xs) + self.reg_coeff * v
+            return ret
+
+        return _eval
+
+
+class PerlmutterHvp(HessianVectorProduct):
+    """This class computes hessian-vector product using Pearlmutter's algorithm.
```
ryanjulian: We generally avoid starting descriptions with "This class" -- imagine browsing docs and seeing that in front of every class.

src/garage/tf/optimizers/conjugate_gradient_optimizer.py
```diff
@@ -192,29 +305,34 @@ def __init__(self,
...
+            name=None,
+            constraint_name='constraint',
+    ):
+        """Update class properties.
+
+        Args:
+            loss (tf.Tensor): Symbolic expression for the loss function.
+            target (garage.tf.policies.Policy): A parameterized object to
+                optimize over.
+            leq_constraint (tuple(tf.Tensor, float)): A constraint provided
```
ryanjulian: collections usually use [] for type annotations (e.g. `tuple[tf.Tensor, float]`, `list[tf.Tensor]`)

tests/garage/tf/optimizers/test_conjugate_gradient_optimizer.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class TestConjugateGradientOptimizer:
+    """Test class for ConjugateGradientOptimizer and HVP classes"""
+
+    def test_cg(self):
+        """Solve Ax = b using Conjugate gradient method."""
+        a = np.random.randn(5, 5)
+        a = a.T.dot(a)  # make sure a is positive semi-definite
+        b = np.random.randn(5)
```
ryanjulian: please don't use random data in tests. you can instead use well-known sequences like pi, e, fibonacci, etc.

tests/garage/tf/optimizers/test_conjugate_gradient_optimizer.py
```diff
@@ -0,0 +1,87 @@
...
+        x = cg(a.dot, b, cg_iters=5)
+        assert np.allclose(a.dot(x), b)
+
+
+class TestPermmutterHvp:
```
ryanjulian: please use `TfGraphTestCase` if possible

src/garage/tf/optimizers/conjugate_gradient_optimizer.py
```diff
@@ -1,27 +1,109 @@
+"""Conjugate Gradient Optimizer.
+
+This optimizer uses conjugate gradient method to compute search direction,
```
ryanjulian: Please try your best not to use `This foo uses bar` -- Try an imperative statement like "Uses the Conjugate Gradient method..."

Also, for subtle code like this, if you found a good reference for the math, please include a citation with URL for it.

tests/garage/tf/optimizers/test_conjugate_gradient_optimizer.py
utkarshjp7: The cg function should solve x for any equation `Ax = b` as long as A is a positive semi definite matrix. So, it should pass regardless of the values of `A` and `b`.

ryanjulian: Sorry, I didn't meant to imply the test is wrong. I'm sure it works.

We don't use RNGs in the testing suite ever, because makes tests non-deterministic.

src/garage/tf/optimizers/conjugate_gradient_optimizer.py
```diff
@@ -1,27 +1,109 @@
...
+        """Build evaluation function.
+
+        Args:
+            inputs (tuple(numpy.ndarray)): Function f will be evaluated on
+                this inputs.
+
+        Returns:
+            function: This function can be called to get the final result.
+
+        """  # yapf: disable
```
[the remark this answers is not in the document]
ryanjulian: doesn't this disable YAPF for all lines below this pragma?

src/garage/tf/optimizers/conjugate_gradient_optimizer.py
```diff
@@ -1,27 +1,109 @@
...
+        num_slices (int): hessian-vector product function's inputs will be
+            divided into num_slices and then averaged together to improve
+            performance.
+
+    """
+
     def __init__(self, num_slices=1):
         self.target = None
         self.reg_coeff = None
         self.opt_fun = None
```
ryanjulian: please change these attributes to _private while you're here

tests/garage/tf/optimizers/test_conjugate_gradient_optimizer.py
utkarshjp7: I see, that makes sense.

src/garage/tf/optimizers/conjugate_gradient_optimizer.py
```diff
@@ -157,30 +285,15 @@ def __init__(self,
...
-         time.
-        :param debug_nan: if set to True, NanGuard will be added to the
-         compilation, and ipdb will be invoked when nan is detected
-        :param accept_violation: whether to accept the descent step if it
-         violates the line search condition after exhausting all backtracking
-         budgets
-        :return:
-        """
+                 num_slices=1):
+        Serializable.__init__()  # pylint requires to call super init
```
ryanjulian: just remove Serializable and use `__setstate__` and `__getstate__`

tests/garage/tf/optimizers/test_conjugate_gradient_optimizer.py
```diff
@@ -0,0 +1,87 @@
...
+        a = tf.Variable([0.0])
+        f = a * (x**2)
+        expected_hessian = 2 * a_val
+        vector = np.array([10.0])
+        expected_hvp = expected_hessian * vector
+        reg_coeff = 1e-5
+        hvp = FiniteDifferenceHvp()
+
+        sess = tf.compat.v1.Session()
+        sess.__enter__()
```
ryanjulian: please use TfGraphTestCase

src/garage/tf/optimizers/conjugate_gradient_optimizer.py
```diff
@@ -146,6 +251,29 @@ class ConjugateGradientOptimizer(Serializable):
...
+        cg_iters (int): The number of CG iterations used to calculate A^-1 g
+        reg_coeff (float): A small value so that A -> A + reg*I
+        subsample_factor (float): Subsampling factor to reduce samples when
+            using "conjugate gradient. Since the computation time for the
+            descent direction dominates, this can greatly reduce the overall
+            computation time.
+        backtrack_ratio (float): backtrack ratio for backtracking line search.
+        max_backtracks (int): Max number of iterations for backtrack
+            linesearch.
+        debug_nan (bool): if set to True, NanGuard will be added to the
```
ryanjulian: AFAICT this docstring is no longer true

src/garage/tf/optimizers/conjugate_gradient_optimizer.py
```diff
@@ -146,6 +251,29 @@ class ConjugateGradientOptimizer(Serializable):
...
+            compilation, and ipdb will be invoked when nan is detected
+        accept_violation (bool): whether to accept the descent step if it
+            violates the line search condition after exhausting all
+            backtracking budgets.
+        hvp_approach (garage.tf.optimizers.conjugate_gradient_optimizer.HessianVectorProduct): # noqa: E501
```
ryanjulian: i think you can use HessianVectorProduct because it's in the same file

src/garage/tf/optimizers/conjugate_gradient_optimizer.py
```diff
@@ -372,5 +525,38 @@ def optimize(self,
...
             logger.log('backtrack iters: %d' % n_iter)
-            logger.log('computing loss after')
             logger.log('optimization finished')
+
+
+def cg(f_Ax, b, cg_iters=10, residual_tol=1e-10):
+    """Use Conjugate Gradient iteration to solve Ax = b. Demmel p 312.
+
+    Args:
+        f_Ax (function): A function to compute hessian vector product.
```
ryanjulian: Hessian

tests/garage/tf/optimizers/test_conjugate_gradient_optimizer.py
```diff
@@ -0,0 +1,87 @@
...
+        b = np.random.randn(5)
+        x = cg(a.dot, b, cg_iters=5)
+        assert np.allclose(a.dot(x), b)
+
+
+class TestPermmutterHvp:
+    """Test class for PermmutterHvp"""
+
+    def test_perm_mutter_hvp(self):
+        """Test if hessian product calculations are correct"""
```
ryanjulian: Hessian

tests/garage/tf/optimizers/test_conjugate_gradient_optimizer.py
```diff
@@ -0,0 +1,87 @@
...
+        hx = hvp.build_eval(np.array([a_val]))
+        computed_hvp = hx(vector)
+        np.allclose(computed_hvp, expected_hvp)
+
+
+class TestFiniteDifferenceHvp:
+    """Test class for FiniteDifferenceHvp"""
+
+    def test_finite_difference_hvp(self):
+        """Test if hessian product calculations are correct"""
```
ryanjulian: Hessian

src/garage/tf/optimizers/conjugate_gradient_optimizer.py
```diff
@@ -1,27 +1,109 @@
...
 
 from garage.core import Serializable
-from garage.misc import krylov
 from garage.tf.misc import tensor_utils
 from garage.tf.optimizers.utils import LazyDict, sliced_fun
 
 
-class PerlmutterHvp:
+class HessianVectorProduct(abc.ABC):
+    """Base class for computing hessian-vector product.
```
ryanjulian: Hessian

src/garage/tf/optimizers/conjugate_gradient_optimizer.py
```diff
@@ -1,27 +1,109 @@
...
+
+    Args:
+        num_slices (int): hessian-vector product function's inputs will be
```
ryanjulian: Hessian

src/garage/tf/optimizers/conjugate_gradient_optimizer.py
```diff
@@ -1,27 +1,109 @@
...
         self.reg_coeff = None
         self.opt_fun = None
         self._num_slices = num_slices
 
+    @abc.abstractmethod
     def update_opt(self, f, target, inputs, reg_coeff, name=None):
+        """Update class properties.
+
+        Args:
+            f (tf.Tensor): The function whose hessian needs to be computed.
```
ryanjulian: Hessian

src/garage/tf/optimizers/conjugate_gradient_optimizer.py
```diff
@@ -1,27 +1,109 @@
...
+
+        Returns:
+            function: This function can be called to get the final result.
+
+        """  # yapf: disable
+        def _eval(v):
+            """The evaluation function.
+
+            Args:
+                v (numpy.ndarray): The vector to be multiplied with hessian.
```
ryanjulian: Hessian

src/garage/tf/optimizers/conjugate_gradient_optimizer.py
```diff
@@ -1,27 +1,109 @@
...
+
+            Returns:
+                numpy.ndarray: The product of hessian of function f and v.
```
ryanjulian: Hessian

src/garage/tf/optimizers/conjugate_gradient_optimizer.py
ryanjulian: Hessian is usually capitalized, FYI.

src/garage/tf/optimizers/conjugate_gradient_optimizer.py
```diff
@@ -1,27 +1,109 @@
...
+    `Pearlmutter, Barak A. "Fast exact multiplication by the Hessian." Neural
+     computation 6.1 (1994): 147-160.`
+
+    """
+
+    def update_opt(self, f, target, inputs, reg_coeff, name=None):
+        """Update class properties.
+
+        Args:
+            f (tf.Tensor): The function whose hessian needs to be computed.
```
ryanjulian: Hessian

src/garage/tf/optimizers/conjugate_gradient_optimizer.py
```diff
@@ -57,41 +147,60 @@ def hx_plain():
...
-    def build_eval(self, inputs):
-        def eval(x):
-            xs = tuple(self.target.flat_to_params(x, trainable=True))
-            ret = sliced_fun(self.opt_fun['f_hx_plain'], self._num_slices)(
-                inputs, xs) + self.reg_coeff * x
-            return ret
 
-        return eval
+class FiniteDifferenceHvp(HessianVectorProduct):
+    """This class computes hessian-vector product using finite difference method.
```
ryanjulian: Hessian

tests/garage/tf/optimizers/test_conjugate_gradient_optimizer.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class TestPermmutterHvp:
+    """Test class for PermmutterHvp"""
+
+    def test_perm_mutter_hvp(self):
+        """Test if hessian product calculations are correct"""
+        policy = OneParamPolicy()
+        x = policy.get_params()[0]
+        a_val = np.array([5.0])
```
ryanjulian: i would be more comfortable if you tested this on a range of cases.

at the very least, a 1x1 Hessian matrix doesn't test much because you don't even have off-diagonal elements. You would at least want a 2x2 matrix.

tests/garage/tf/optimizers/test_conjugate_gradient_optimizer.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class TestFiniteDifferenceHvp:
+    """Test class for FiniteDifferenceHvp"""
+
+    def test_finite_difference_hvp(self):
+        """Test if hessian product calculations are correct"""
+        policy = OneParamPolicy()
+        x = policy.get_params()[0]
+        a_val = np.array([5.0])
```
ryanjulian: please use at least a 2x2 matrix and perhaps a range of input/output values

src/garage/tf/optimizers/conjugate_gradient_optimizer.py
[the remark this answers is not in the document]
utkarshjp7: `yapf` adds a new line after the docstring, and then `flake8` complains about it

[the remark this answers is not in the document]
utkarshjp7: I just tested, it doesn't disable for all lines below.

src/garage/tf/regressors/bernoulli_mlp_regressor.py
```diff
@@ -153,12 +153,10 @@ def _initialize(self):
 
             self._optimizer.update_opt(loss=loss,
                                        target=self,
-                                       network_output=[y_hat],
```
utkarshjp7: Removed it since none of the `garage.tf.optimizers` uses this. It was causing some regressor tests to fail since I removed `*kwargs` from `ConjugateGradientOptimizer.update_opt`.

src/garage/tf/optimizers/conjugate_gradient_optimizer.py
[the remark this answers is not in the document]
ryanjulian: what is the flake8 error? i think it's actually the yapf knob `BLANK_LINE_BEFORE_NESTED_CLASS_OR_DEF` which seems sane to me.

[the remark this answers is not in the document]
utkarshjp7: `D202 No blank lines allowed after function docstring`

ahtsan:
Should we also test for pickling? Look like the HVP and optimizers are not pickleable / local functions are not restored, or at least we need to make sure it is properly reconstructed after unpickled.

src/garage/tf/optimizers/conjugate_gradient_optimizer.py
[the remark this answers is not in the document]
ryanjulian: Please use a `# noqa: D202` here and include a link to this issue https://github.com/PyCQA/pydocstyle/pull/395

tests/garage/tf/optimizers/test_conjugate_gradient_optimizer.py
```diff
@@ -0,0 +1,240 @@
...
+        reg_coeff = 1e-5
+        hvp = PerlmutterHvp()
+
+        self.sess.run(tf.global_variables_initializer())
+        hvp.update_hvp(f, policy, (a, ), reg_coeff)
+        hx = hvp.build_eval(np.array([a_val]))
+        computed_hvp = hx(vector)
+        np.allclose(computed_hvp, expected_hvp)
+
+    @pytest.mark.parametrize('a_val, b_val, x_val, y_val, vector', [
```
ryanjulian: is there a reason the 1x1 case can't be a case here? shouldn't the mathematical invariant still apply?

utkarshjp7:
> Should we also test for pickling? Look like the HVP and optimizers are not pickleable / local functions are not restored, or at least we need to make sure it is properly reconstructed after unpickled.

You are right. We have to call `update_opt` after unpickling to reconstruct them. I thought this issue will be solved by #982

ryanjulian:
Oh right --yes please test that it can be pickled.

tests/garage/tf/optimizers/test_conjugate_gradient_optimizer.py
utkarshjp7: The test logic depends on the number of function parameters. In 1x1 case, number of function parameters are 1, and for 2x2 its 2. It is easier to have them as separate tests.

src/garage/tf/optimizers/conjugate_gradient_optimizer.py
```diff
@@ -1,27 +1,118 @@
...
+        Returns:
+            dict: the state to be pickled for the instance.
+
+        """
+        new_dict = self.__dict__.copy()
+        del new_dict['_hvp_fun']
+        return new_dict
+
+
+class PerlmutterHvp(HessianVectorProduct):
```
nish21: This is probably pedantic. But I think there is a typo in the class name here and the tests. The three spellings in the function name, the name in the citation and the name of the test class are all different. I think It should be Pearlmutter - as in the citation.

utkarshjp7: Ah..Thanks for pointing out. Fixed it.

PR #979, opened 2019-10-31
Simpify check_precommit.sh

ryanjulian:
* Use better logic for determining relevant commit hashes
* Only check commit messages for pull requests

### DIFF
--- a/scripts/travisci/check_precommit.sh
+++ b/scripts/travisci/check_precommit.sh
@@ -1,16 +1,19 @@
 #!/usr/bin/env bash
 status=0
 
-if [[ "${TRAVIS_PULL_REQUEST}" != "false" && "${TRAVIS}" == "true" ]]; then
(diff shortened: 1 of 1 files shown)


scripts/travisci/check_precommit.sh
```diff
@@ -20,31 +23,8 @@ if [[ "${TRAVIS_PULL_REQUEST}" != "false" && "${TRAVIS}" == "true" ]]; then
...
-
-      # Print message if it fails
-      if [[ "${pass}" -ne 0 ]]; then
-        echo "Failing commit message:"
-        cat "${commit_msg}"
-      fi
-
-    done < <(git rev-list ^origin/master "${TRAVIS_BRANCH}")
-  fi
+  done < <(git log --cherry-pick --left-only --pretty="%H" \
```
krzentner: For future reference, I *think* this is the same as `git log --pretty="%H" "${ORIGIN}" --not "${SOURCE}"`

PR #982, opened 2019-10-31
Remove Serializable

nish21:
This commit removes the untested serialization magic in the codebase and replaces it with logic for pickling using Python's `pickle` module.

This PR needs to be force-pushed. Tests pass locally.


.pylintrc
```diff
@@ -27,7 +27,9 @@ disable =
     too-many-locals,
     # Detection seems buggy or unhelpful
     duplicate-code,
-
+    # Garage env wrappers access gym private variables
+    protected-access,
```
ryanjulian: please disable these individually where they occur. this is a serious violation elsewhere in the codebase so we don't want to disable it everywhere.

src/garage/envs/base.py
```diff
@@ -117,15 +123,42 @@ def step(self, action):
 
         This method is necessary to suppress a deprecated warning
         thrown by gym.Wrapper.
+
+        Args:
+            action (object): an action provided by the agent.
```
ryanjulian: Please use capitalization like regular english sentences.

src/garage/envs/dm_control/dm_control_env.py
```diff
@@ -70,11 +65,20 @@ def action_space(self):
...
-        return gym.spaces.Box(
-            low=-np.inf, high=np.inf, shape=[flat_dim], dtype=np.float32)
+        return gym.spaces.Box(low=-np.inf,
+                              high=np.inf,
+                              shape=[flat_dim],
+                              dtype=np.float32)
+
+    def __getstate__(self):
+        d = self.__dict__.copy()
+        if d['_viewer'] is not None:
```
ahtsan: This is equivalent to always setting it to `None`?

src/garage/tf/optimizers/conjugate_gradient_optimizer.py
```diff
@@ -66,8 +70,15 @@ def eval(x):
 
         return eval
 
+    def __getstate__(self):
+        """Object.__getstate__."""
+        new_dict = self.__dict__.copy()
+        del new_dict['opt_fun']
+        return new_dict
```
ahtsan: Just curious, will `opt_fun` be reconstructed when unpickled?

src/garage/tf/optimizers/conjugate_gradient_optimizer.py
```diff
@@ -374,3 +385,9 @@ def optimize(self,
             logger.log('backtrack iters: %d' % n_iter)
             logger.log('computing loss after')
             logger.log('optimization finished')
+
+    def __getstate__(self):
+        """Object.__getstate__."""
+        new_dict = self.__dict__.copy()
+        del new_dict['_opt_fun']
```
ahtsan: Same question - will `opt_fun` be reconstructed when unpickled?

ryanjulian:
Looks like you missed a real breakage among all the broken pylint rules: https://travis-ci.com/rlworkgroup/garage/jobs/251299777#L2642

src/garage/tf/optimizers/conjugate_gradient_optimizer.py
nish21: No. Unpickling a pickled optimizer object will not reconstruct `_opt_fun`. It needs to be created by calling the `update_opt` function with the parameters. But let's say a primitive uses the optimizer, pickling and unpickling a primitive calls the `update_opt` with the right parameters, and thus we have an equivalent object after de-serialization.  This behavior is consistent with the master branch (with Serializable).

src/garage/envs/base.py
```diff
@@ -36,11 +34,21 @@ class GarageEnv(gym.Wrapper, Serializable):
...
+        env (gym.Env): the env that will be wrapped
+        env_name (str): If the env_name is speficied, a gym environment
+            with that name will be created. If such an environment does not
+            exist, a `gym.error` is thrown.
+
     """
 
     def __init__(self, env=None, env_name=''):
+        # Needed for deserialization
+        self._env_name = env_name
```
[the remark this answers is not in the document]
nish21: Yes, i'll update this.

src/garage/tf/optimizers/conjugate_gradient_optimizer.py
ahtsan: Should we put some guards in the optimizers to make sure they are properly reconstructed? We are assuming user will always call `update_opt` afterwards. Or at least, we should have an integration test to make sure `update_opt` will be called afterwards.

nish21: Yes, currently we always assume the user calls `update_opt` before using other functions of the optimizer. I think it might be a good idea to raise some exception when the user does not set the parameters first.

src/garage/tf/optimizers/conjugate_gradient_optimizer.py
ahtsan: This is discussed in another thread, resolved

.pylintrc
```diff
@@ -27,7 +27,8 @@ disable =
     too-many-locals,
     # Detection seems buggy or unhelpful
     duplicate-code,
-
+    # Some wrappers don't have enough public methods
```
ryanjulian: again, please disable this within the code

src/garage/envs/base.py
```diff
@@ -1,10 +1,11 @@
 """Wrapper class that converts gym.Env into GarageEnv."""
+# pylint: disable=protected-access
```
ryanjulian: please do this at each line rather than for whole files

Issue #978, opened 2019-10-31
Update documentation for running experiments


ryanjulian:
@zhanpenghe pointed out that seems like the documentation for running experiments has drifted.

Please update https://garage.readthedocs.io/en/latest/user/experiments.html (source: https://github.com/rlworkgroup/garage/blob/master/docs/user/experiments.rst) to match reality.

zequnyu:
Closed by #983.

PR #983, opened 2019-10-31
Update experiments doc

zequnyu:
#978.

### DIFF
--- a/docs/user/experiments.rst
+++ b/docs/user/experiments.rst
@@ -10,43 +10,40 @@ We use object oriented abstractions for different components required for an exp
 
 .. code-block:: python
 
-    import gym
-
     from garage.experiment import run_experiment
     from garage.np.baselines import LinearFeatureBaseline
     from garage.tf.algos import TRPO
     from garage.tf.envs import TfEnv
+    from garage.tf.experiment import LocalTFRunner
     from garage.tf.policies import CategoricalMLPPolicy
 
 
-    def run_task(*_):
+    def run_task(snapshot_config, *_):
         """Wrap TRPO training task in the run_task function."""
-        env = TfEnv(env_name="CartPole-v1")
+        with LocalTFRunner(snapshot_config=snapshot_config) as runner:
+            env = TfEnv(env_name='CartPole-v1')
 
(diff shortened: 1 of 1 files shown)


docs/user/experiments.rst
```diff
@@ -10,43 +10,40 @@ We use object oriented abstractions for different components required for an exp
...
 
-    import gym
-
     from garage.experiment import run_experiment
     from garage.np.baselines import LinearFeatureBaseline
     from garage.tf.algos import TRPO
     from garage.tf.envs import TfEnv
-    from garage.tf.policies import CategoricalMLPPolicy
+    from garage.tf.experiment import LocalTFRunner
+    from garage.tf.policies import
```
gitanshu: Looks like the import got missed
```suggestion
    from garage.tf.policies import CategoricalMLPPolicy
```

docs/user/experiments.rst
```diff
@@ -10,43 +10,40 @@ We use object oriented abstractions for different components required for an exp
...
+
 
+    def run_task(snapshot_config, *_):
+        """Run task."""
```
gitanshu: I think the existing docstring was fine
```suggestion
        """Wrap TRPO training task in the run_task function."""
```

docs/user/experiments.rst
```diff
@@ -119,24 +116,17 @@ Additional arguments for `run_experiment`:
 Running Experiments with TensorFlow and GPU
 ===========================================
 
-To run experiments in the TensorFlow tree of garage with the GPU enabled, set the flags use_tf and use_gpu to True when calling `run_experiment`, as shown in the code below:
+Using `tensorflow-gpu`, TensorFlow has GPU enabled by default. Set the flags `force_cpu` to True to disable GPU, as shown in the code below:
```
gitanshu: Maybe a nitpick, but I suggest this to be consistent with installation docs
```suggestion
If you installed the `garage[gpu]` extra package, TensorFlow will use GPU by default. To disable GPU, set the flag `force_cpu` to True, as shown in the code below:
```

zequnyu: Thanks for suggestions! Updated.

PR #984, opened 2019-11-01
Add separate Travis job for pre-commit checks

ryanjulian:
This PR adds a separate CI job for pre-commit checks. This makes it
easier to distinguish at a glance pre-commit failures from test
failures. It also allows us to disable this job for non-PR builds.

The downside of this PR is that it adds an additionao build job and
~5 minutes of overhead to the PR build process, because it must pull
the Docker cache. For builds which update the Docker container, this
will add an additional ~10 minutes to rebuild the container. This can
be addressed in a later PR by adding a separate build stage to build
and cache the container.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,4 +1,5 @@
 fail_fast: false # set to true to have pre-commit stop running hooks after the first failure.
+default_stages: [commit, push]
 
 repos:
 -   repo: local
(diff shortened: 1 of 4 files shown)


nish21:
Just curious, could this be combined with the verify conda job?

ryanjulian:
No, because we need to verify pipenv/conda on both branches and pull requests, but only check pre-commit on pull requests.

PR #986, opened 2019-11-01
Remove garage.misc.special and garage.envs.util

ryanjulian:
* Removes dead code in garage.misc.special, moves remaining functions
  to garage.misc.tensor_utils
* Refactor garage.envs.NormalizedEnv to use new gym helper functions
* Refactor a few algorithms for the newly-moved functions


src/garage/envs/normalized_env.py
```diff
@@ -25,7 +24,7 @@ def __init__(
         self._flatten_obs = flatten_obs
 
         self._obs_alpha = obs_alpha
-        flat_obs_dim = flat_dim(env.observation_space)
+        flat_obs_dim = gym.spaces.utils.flatdim(env.observation_space)
```
ahtsan: should be `flat_dim`?

PR #987, opened 2019-11-01
Wrap examples in tests to run on CI

gitanshu:
Doesn't include the test for sim_policy.py. Will do that separately.

Update: Also exludes tf/ppo_memorize_digits, tf/dqn_pong.py and tf/trpo_cubecrash.py for now, since they take too long to run on CI even with 1 epoch. Still figuring out decent enough parameters to run them on CI, but want to get this merged first.


zequnyu:
I’m not sure whether it is good to add parser to /examples, because it will be strange for users coming to use garage. Do you have better ideas? @ryanjulian @krzentner

ryanjulian:
I think that it is fine as long as the examples run with no arguments (i.e. they all have good defaults and the parsing is optional)

examples/step_env.py
```diff
@@ -12,7 +21,12 @@
 # Step randomly until interrupted
 try:
     print('Press Ctrl-C to stop...')
+    steps = 0
     while True:
+        if args.n_max_steps:
```
ryanjulian: hmm just set a default for n_steps, it will make the logic simpler.

examples/step_dm_control_env.py
```diff
@@ -15,7 +24,12 @@
 # Step randomly until interrupted
 try:
     print('Press Ctrl-C to stop...')
+    steps = 0
     while True:
+        if args.n_max_steps:
```
ryanjulian: just set a default

tests/integration_tests/test_examples.py
```diff
@@ -1,39 +1,84 @@
...
+import pytest
+
+from garage.envs import GridWorldEnv
+from garage.np.policies import ScriptedPolicy
 from garage.tf.envs import TfEnv
-from garage.tf.policies import CategoricalMLPPolicy
 
+examples_root_dir = 'examples/'
+blacklist = [
+    os.path.join(examples_root_dir, 'resume_training.py'),
```
ryanjulian: you might find that using pathlib makes some of this logic simpler to write

tests/integration_tests/test_examples.py
```diff
@@ -1,39 +1,84 @@
...
-        policy = CategoricalMLPPolicy(name='policy',
-                                      env_spec=env.spec,
-                                      hidden_sizes=(32, 32))
+@pytest.mark.parametrize('filepath', enumerate_examples())
+def test_examples(filepath):
+    args = [
+        filepath,
+    ]
+    if filepath in {
+            os.path.join(examples_root_dir, 'step_env.py'),
```
ryanjulian: this is kind of complicated. why not just make a different test method for these?

examples/step_env.py
gitanshu: I was trying to keep the behavior same if example is run without arguments. If I set a default number, it will stop after that number of steps. I could change the argument to '--run_once' with no value though. Did you mean something different?

krzentner:
> I’m not sure whether it is good to add parser to /examples, because it will be strange for users coming to use garage. Do you have better ideas? @ryanjulian @krzentner

I think you're right. It might sound like a somewhat sketchy idea, but how about adding a special, just-for-testing, argument parsing to `run_experiment`?

ryanjulian:
I don't think adding argparse to the examples is super offensive, but a special flag for run_experiment is also fine. I'll add that we hope to delete run_experiment in its current form soon, so I'm not sure how long this fix would last.

krzentner:
After some discussion offline, a better method might be to add an environment variable (e.g. `GARAGE_EXAMPLE_TEST_MODE`), look for that variable in `LocalRunner`, and set that environment variable in the subprocesses used to test the experiments. This should be a fairly permanent solution.

ryanjulian:
@krzentner would you be amenable to using [click](https://click.palletsprojects.com/en/7.x/) to add the option rather than `argparse`? It's already a dependency and adds a lot less noise to the example file. It also doesn't require environment variable magic.

ryanjulian:
here's what it would look like

```python
#!/usr/bin/env python3
"""
This is an example to train a task with Cross Entropy Method.

Here it runs CartPole-v1 environment with 100 epoches.

Results:
    AverageReturn: 100
    RiseTime: epoch 8
"""
import click

from garage.np.algos import CEM
from garage.sampler import OnPolicyVectorizedSampler
from garage.tf.experiment import LocalTFRunner


@click.command()
@click.option('--n_epochs', default=100, help='Number of optimization epochs to run')
def run_task(snapshot_config, n_epochs, *_):
    """Train CEM with Cartpole-v1 environment."""
    with LocalTFRunner(snapshot_config=snapshot_config) as runner:
        env = TfEnv(env_name='CartPole-v1')

        policy = CategoricalMLPPolicy(name='policy',
                                      env_spec=env.spec,
                                      hidden_sizes=(32, 32))

        n_samples = 20

        algo = CEM(env_spec=env.spec,
                   policy=policy,
                   baseline=baseline,
                   best_frac=0.05,
                   max_path_length=100,
                   n_samples=n_samples)

        runner.setup(algo, env, sampler_cls=OnPolicyVectorizedSampler)
        # NOTE: make sure that n_epoch_cycles == n_samples !
        runner.train(n_epochs=args.n_epochs,
                     batch_size=1000,
                     n_epoch_cycles=n_samples)


run_experiment(run_task, snapshot_mode='last', seed=1)
```

krzentner:
Yeah, they definitely looks like a reasonable solution.

gitanshu:
It seems like click wouldn't work with the run_experiment api, since by annotating run_task with @click.command(), the method_call argument passed to run_experiment changes from a function to a click.core.Command object

examples/np/cma_es_cartpole.py
```diff
@@ -16,6 +18,20 @@
...
 
+parser = argparse.ArgumentParser()
+parser.add_argument('--n_epochs',
+                    type=int,
+                    default=100,
+                    help='Number of epochs to run')
+args = parser.parse_args()
+
+parser = argparse.ArgumentParser()
+parser.add_argument('--n_epochs',
```
ryanjulian: this is repeated. is that intentional?

examples/step_env.py
ryanjulian: i meant it's fine to change the original behavior a little to make it easier to test.

examples/np/cma_es_cartpole.py
gitanshu: Ah. Good catch. Not intentional. Somehow got added twice during a regex replace op

ryanjulian:
Hmm, let's not get too bogged down in this.

At the expense of some beauty, this works:
```python
#!/usr/bin/env python3
"""
This is an example to train a task with VPG algorithm.

Here it runs CartPole-v1 environment with 100 iterations.

Results:
    AverageReturn: 100
    RiseTime: itr 13
"""
import click

from garage.tf.experiment import LocalTFRunner


@click.command()
@click.option('--n_epochs', default=100,
    help='Number of optimization epochs to run')
def main(n_epochs):

        """Run task."""
        with LocalTFRunner(snapshot_config=snapshot_config) as runner:
            env = TfEnv(env_name='CartPole-v1')

            policy = CategoricalMLPPolicy(name='policy',
                                          env_spec=env.spec,
                                          hidden_sizes=(32, 32))

            algo = TRPO(env_spec=env.spec,
                        policy=policy,
                        baseline=baseline,
                        max_path_length=100,
                        discount=0.99,
                        max_kl_step=0.01)

            runner.setup(algo, env)
            runner.train(n_epochs=n_epochs, batch_size=4000)


    run_experiment(
        run_task,
        snapshot_mode='last',
        seed=1,
    )

if __name__ == '__main__':
    main()
```

@krzentner WDYT?

krzentner:
> @krzentner WDYT?

I'm fine with this being a temporary solution, but I don't really see the point. I still think the environment variable is the preferable long-term solution (and easier, in that it doesn't actually require modifying most examples). I think this is especially important because of how `run_experiment` interferes from using click in the obvious way (on `run_task`). It would be kinda bad to have a discouraged method of passing experiment options in *every* one of our examples.

But really, the environment variable method is very easy to implement. It requires fewer new lines of code than the proposed change to a single example, and means that we don't need to modify most of the examples. I've put the required change in a [commit here](https://github.com/rlworkgroup/garage/commit/be231468496b6bcfac97225a89616e75f0aa1dbb).

ryanjulian:
Okay I'll relent. Just use the environment variable. Will this work for all the examples?

krzentner:
It doesn't work for `sim_policy.py` or the examples that don't use the runner (i.e. `step_env.py`, and `step_dm_control_env.py`). I just tested that it works for all of the other examples.

tests/integration_tests/test_examples.py
```diff
@@ -1,39 +1,84 @@
...
+examples_root_dir = 'examples/'
+blacklist = [
+    os.path.join(examples_root_dir, 'resume_training.py'),
+]
+sim_policy_path = os.path.join(examples_root_dir, 'sim_policy.py')
+
+
+def enumerate_examples():
+    examples = []
+    for (dirpath, _, filenames) in os.walk(examples_root_dir):
```
ryanjulian: `glob` will make your life a hell of a lot easier here

tests/integration_tests/test_examples.py
```diff
@@ -1,39 +1,84 @@
...
+from unittest.mock import Mock
+
+import pytest
+
+from garage.envs import GridWorldEnv
+from garage.np.policies import ScriptedPolicy
 from garage.tf.envs import TfEnv
-from garage.tf.policies import CategoricalMLPPolicy
 
+examples_root_dir = 'examples/'
```
ryanjulian: i don't think this is really necessary. if someone moves the examples around in the repo they will need to update this string regardless. in any case, it should be ALL_CAPS

tests/integration_tests/test_examples.py
ryanjulian: pathlib would make this less ugly. or just use absolute paths.

tests/integration_tests/test_examples.py
```diff
@@ -1,39 +1,84 @@
...
+blacklist = [
```
ryanjulian: BLACKLIST

tests/integration_tests/test_examples.py
```diff
@@ -1,39 +1,84 @@
...
+    os.path.join(examples_root_dir, 'resume_training.py'),
+]
+sim_policy_path = os.path.join(examples_root_dir, 'sim_policy.py')
```
ryanjulian: SIM_POLICY_PATH

tests/integration_tests/test_examples.py
```diff
@@ -1,39 +1,84 @@
 """
```
ryanjulian: i think you could make this file a lot simpler if you defined glob strings for the different types of examples and wrote a test case for each glob string.

conditionals inside test cases is a bad code sell. if a file needs to be handled differently, then it's a different test case.

tests/integration_tests/test_examples.py
```diff
@@ -1,39 +1,48 @@
...
+
+
+def enumerate_algo_examples():
+    """Return a list of paths for all algo examples
+
+    Returns:
+        List[str]: list of path strings
+
+    """
+    return map(str, EXAMPLES_ROOT_DIR.glob('*/**/*.py'))
```
ryanjulian: hmm why not `glob('**.py')`? + a blacklist -- then you don't need to refactor if the examples folder gets reorganized.

gitanshu: That's what I did earlier. Maybe I misunderstood this comment: https://github.com/rlworkgroup/garage/pull/987#discussion_r342887200

tests/integration_tests/test_examples.py
```diff
@@ -1,39 +1,48 @@
...
-                                      hidden_sizes=(32, 32))
+    Args:
+        filepath (str): path string of example
 
-        baseline = LinearFeatureBaseline(env_spec=env.spec)
+    """
+    env = os.environ.copy()
+    env['GARAGE_EXAMPLE_TEST_N_EPOCHS'] = '1'
+    assert subprocess.run([filepath, '--n_epochs', '1'], check=True,
+                          env=env).returncode == 0
```
ryanjulian: doesn't `check=True` have the same affect as this assert? AFAICT you will never fail this assert because `subprocess.run` with raise a CalledProcessError if the process fails.

Have you verified that this breaks when an example is broken?

tests/integration_tests/test_examples.py
ryanjulian: i like the updated version, i'm just suggesting:
```python
NON_ALGORITHM_EXAMPLES = [
    EXAMPLES_ROOT_DIR / 'foo.py',
    EXAMPLES_ROOT_DIR / 'bar.py',
]

def enumerate_algo_examples():
    all_examples = EXAMPLES_ROOT_DIR.glob('**/*.py')
    return [e for e in all_examples if e not in NON_ALGORITHM_EXAMPLES]
```

AFAICT your current test would not catch `examples/algo_example.py` because it assumes that algo examples are always inside a subdir

tests/integration_tests/test_examples.py
```diff
@@ -1,39 +1,48 @@
...
-This is an integration test to make sure scripts from examples/
-work when running `python examples/xx/xxx.py`.
-"""
-from garage.experiment import LocalRunner, run_experiment
-from garage.np.baselines import LinearFeatureBaseline
-from garage.tf.algos import VPG
-from garage.tf.envs import TfEnv
-from garage.tf.policies import CategoricalMLPPolicy
+import os
+from pathlib import Path
```
ryanjulian: import pathlib

tests/integration_tests/test_examples.py
gitanshu: One of the examples I removed in another PR had an ImportError and the test for that. I tried that on an example now and pytest shows a failure, although it's not an AssertionError, but a CalledProcessError. So it never reached the assertion. Without check=True, it works fine and throws AssertionError. 

But another thing I found while checking this is that for something like ValueError, the test still passes, and if it's a NameError, the test just hangs. If I do sys.exit(1), I assumed it would explicitly set the return code but the assertion still passed. These things are with or without check=True :confused:

tests/integration_tests/test_examples.py
gitanshu: Calling it a blacklist doesn't seem appropriate if some of the files on that list have their own tests. So I am thinking a function local "exclude" list instead

gitanshu: Yes, this seems better. I'll change to this.

tests/integration_tests/test_examples.py
ryanjulian: The experiment wrapper is swallowing exceptions, and it doesn't return return code of the experiment process:
https://github.com/rlworkgroup/garage/blob/master/src/garage/experiment/experiment.py#L261

I think you should change 264-265 to just `raise` which re-raises the exception after printing it. I also think you need to either raise an exception if `subprocess.call` doesn't return 0 or you can do:
```
# experiment.py
return subprocess.call(...)

# launcher
return_code = run_experiment(...)
sys.exit(return_code)
```

I think raising an exception is more natural.

Seems like everything would be fixed if you changed experiment.py to use `subprocess.run(..., check=True)` and stop swallowing the exceptions.

gitanshu: Setting timeout parameter on subprocess.run could be one possible solution to this but might be flaky

gitanshu: Ah okay. I'll try that.

gitanshu: That worked for ValueErrors, NameError still hangs and doesn't seem to be reaching the except block in experiment.py:263

ryanjulian: is the NameError raised in the launcher file or in the algorithm itself?

tests/integration_tests/test_examples.py
```diff
@@ -1,39 +1,48 @@
+"""This is an integration test to make sure scripts from examples/
```
ryanjulian: can you mark these all with `@pytest.mark.no_cover`? we don't really want to be considering a line covered just because it appears in an example.

tests/integration_tests/test_examples.py
gitanshu: In the launcher. From this line: https://github.com/rlworkgroup/garage/blob/d4c2d8711c205d247d2106dc02b976f79171b319/src/garage/experiment/experiment_wrapper.py#L165, control goes to run_task, error is raised there, and control doesn't return.

ryanjulian: hmm rather than going down this rabbit hole (which is ultimately caused by bad code we're trying to delete), perhaps just install this and set a timeout for the test: https://pypi.org/project/pytest-timeout/

src/garage/experiment/experiment.py
```diff
@@ -258,11 +258,14 @@ def run_experiment(method_call=None,
         try:
             if env is None:
                 env = dict()
-            subprocess.call(command, shell=True, env=dict(os.environ, **env))
+            subprocess.run(command,
+                           shell=True,
+                           env=dict(os.environ, **env),
+                           check=True)
         except Exception as e:
+            import sys
```
krzentner: Why `import sys`?

tests/integration_tests/test_examples.py
krzentner: Please add a comment along the lines of
> Don't use check=True, since that causes test to hang on failure.

src/garage/experiment/experiment.py
gitanshu: Oops! Missed deleting this when I was testing if calling sys.exit() here returns the error code in test.

krzentner: I'm surprised the precommit didn't catch this.

tests/integration_tests/test_examples.py
gitanshu: This check=True was not the cause of hangs, this one caused the test to exit with an Error before the assertion was evaluated. Adding that in comment. The hang was because check was False in experiment.py, and I guess it makes sense there by looking at it since it's in a try except, so we definitely want it to throw error on failure. I can add a comment there too if you think it's helpful.

krzentner: Ah, I see. I don't really think we need a comment then, but feel free to use your best judgement.

tests/integration_tests/test_examples.py
```diff
@@ -1,39 +1,63 @@
...
+    env = os.environ.copy()
+    env['GARAGE_EXAMPLE_TEST_N_EPOCHS'] = '1'
+    # Don't use check=True, since that causes subprocess to throw an error
+    # in case of failure before the assertion is evaluated
+    assert subprocess.run([filepath, '--n_epochs', '1'], check=False,
+                          env=env).returncode == 0
+
+
+@pytest.mark.no_cover
+@pytest.mark.timeout(10)
```
ryanjulian: you might need to adjust these because travis instances can be slow

Makefile
```diff
@@ -26,7 +26,7 @@ ci-job-precommit: docs
 
 ci-job-normal:
 	pytest -n $$(nproc) --cov=garage -v -m \
-	    'not nightly and not huge and not flaky and not large'
+	    'not nightly and not huge and not flaky and not large' --durations=0
```
ryanjulian: don't forget to update `make test`

gitanshu: Done

ryanjulian:
@gitanshu let me know if/when this needs to be force-pushed

gitanshu:
@ryanjulian this is ready for force-push

ryanjulian:

PR #990, opened 2019-11-02
Remove layers lib based example

gitanshu:


ryanjulian:
Do you mind fixing up #989 while you're at it?

Issue #911, opened 2019-10-02
Merge benchmark scripts for ContinuousMLPBaseline


ryanjulian:
Please merge the benchmark script used for https://github.com/rlworkgroup/garage/pull/898 into tests/benchmarks (omitting, of course, the now-deleted primitive).

ryanjulian:
fixed by #996

PR #996, opened 2019-11-05
Add benchmark script for ContinuousMLPBaseline

avnishn:
This PR is meant to address #911


avnishn:
@naeioi anyone can use any policy that they would like, this benchmark script only serves as an example of how one might go about benchmarking a baseline, specifically, the continuous mlp baseline.

tests/benchmarks/garage/tf/baselines/test_benchmark_continuous_mlp_baseline.py
```diff
@@ -0,0 +1,155 @@
...
+
+algo_params = {
+    'n_envs':
+    8,
+    'n_epochs':
+    20,
+    'n_rollout_steps':
+    2048,
+    'discount':
+    0.99,
```
ryanjulian: this formatting is super strange. did yapf do this?

Issue #865, opened 2019-09-11
Eliminate unnecessary worker startup


krzentner:
There's a bunch of places where `stateful_pool`'s workers get initialized. They should actually only get initialized when `BatchSampler` (or `ParallelVecEnvExecutor`) need them.

naeioi:
To be more specific, `stateful_pool` is only used in parallel samplers, so the `singleton_pool` created whenever `stateful_pool` is imported is not necessary. The instance should be managed by the sampler.

Also, I think the name of `BatchSampler` is misleading. The name suggests it works by sampling batches serially while it is actually a parallel sampler.

zequnyu:
Closed by #1006.

PR #1006, opened 2019-11-06
Eliminate unnecessary worker startup

zequnyu:
#865.

Before this PR:
1. `n_parallel` is passed as a parameter to experiment API to set `parallel_sampler`
2. `LocalRunner` set `max_cpus`, which is the same as above

This PR addresses:

1. remove `n_parallel` setup from `experiment.py`
2. for tf algorithms, `LocalTFRunner` sets `max_cpus` to initialize `parallel_sampler`. (this is due to tf session not being fork safe, therefore `parallel_sampler` must be initialized before setting sessions)
3. for all other algorithms, remove `max_cpus` from `LocalRuner` and move it to `batch_sampler`, so that it is only initialized when in need

Also addresses:
- relocate and rename a test for `tf_batch_sampler`

In order to pass pylint, also does:
- remove `BaseSampler` which is unnecessary, use `Sampler` as base class directly
- update docs, set initialization inside `__init()__`..., for `LocalRunner`, `RaySampler`, etc.

### DIFF
--- a/src/garage/experiment/experiment_wrapper.py
+++ b/src/garage/experiment/experiment_wrapper.py
@@ -20,7 +20,6 @@ import psutil
 
 from garage.experiment import deterministic, SnapshotConfig
 import garage.plotter
-from garage.sampler import parallel_sampler
 import garage.tf.plotter
 
 
@@ -43,12 +42,6 @@ def run_experiment(argv):
 
     default_exp_name = 'experiment_%s_%s' % (timestamp, rand_id)
     parser = argparse.ArgumentParser()
-    parser.add_argument(
-        '--n_parallel',
-        type=int,
-        default=1,
-        help=('Number of parallel workers to perform rollouts. '
-              "0 => don't start any workers"))
     parser.add_argument('--exp_name',
                         type=str,
                         default=default_exp_name,
@@ -125,11 +118,6 @@ def run_experiment(argv):
     if args.seed is not None:
         deterministic.set_seed(args.seed)
 
-    if args.n_parallel > 0:
-        parallel_sampler.initialize(n_parallel=args.n_parallel)
-        if args.seed is not None:
-            parallel_sampler.set_seed(args.seed)
-
     if not args.plot:
         garage.plotter.Plotter.disable()
         garage.tf.plotter.Plotter.disable()
@@ -171,8 +159,6 @@ def run_experiment(argv):
     except BaseException:
         children = garage.plotter.Plotter.get_plotters()
         children += garage.tf.plotter.Plotter.get_plotters()
-        if args.n_parallel > 0:
-            children += [parallel_sampler]
         child_proc_shutdown(children)
         raise
 
--- a/src/garage/experiment/local_runner.py
+++ b/src/garage/experiment/local_runner.py
@@ -7,6 +7,7 @@ from dowel import logger, tabular
 
 from garage.experiment.deterministic import get_seed, set_seed
 from garage.experiment.snapshotter import Snapshotter
+from garage.sampler import parallel_sampler
 
 
 class ExperimentStats:
@@ -114,10 +115,12 @@ class LocalRunner:
                                         snapshot_config.snapshot_mode,
                                         snapshot_config.snapshot_gap)
 
-        if max_cpus > 1:
-            # pylint: disable=import-outside-toplevel
-            from garage.sampler import singleton_pool
-            singleton_pool.initialize(max_cpus)
+        parallel_sampler.initialize(max_cpus)
+
+        seed = get_seed()
+        if seed is not None:
+            parallel_sampler.set_seed(seed)
+
         self._has_setup = False
         self._plot = False
 
--- a/src/garage/sampler/base.py
+++ b/src/garage/sampler/base.py
@@ -1,35 +1,39 @@
+"""Base class of Sampler."""
 import abc
 
 
 class Sampler(abc.ABC):
+    """Sampler interface."""
+
     @abc.abstractmethod
     def start_worker(self):
-        """Initialize the sampler,
+        """Initialize the sampler.
 
         e.g. launching parallel workers if necessary.
+
         """
-        pass
 
     @abc.abstractmethod
-    def obtain_samples(self, itr):
+    def obtain_samples(self, itr, batch_size, whole_paths):
         """Collect samples for the given iteration number.
 
         Args:
-            itr (int): Iteration number.
+            itr (int): Number of iteration.
+            batch_size (int): Number of environment steps in one batch.
+            whole_paths (bool): Whether to use whole path or truncated.
 
         Returns:
             list[dict]: A list of paths.
 
         """
-        pass
 
     @abc.abstractmethod
     def shutdown_worker(self):
         """Terminate workers if necessary."""
-        pass
 
 
 class BaseSampler(Sampler):
+    # pylint: disable=abstract-method
     """Base class for sampler.
 
     Args:
--- a/src/garage/sampler/batch_sampler.py
+++ b/src/garage/sampler/batch_sampler.py
@@ -1,5 +1,4 @@
 """Class with batch-based sampling."""
-
 from garage.sampler import parallel_sampler
 from garage.sampler.base import BaseSampler
 from garage.sampler.utils import truncate_paths
@@ -14,9 +13,6 @@ class BatchSampler(BaseSampler):
 
     """
 
-    def __init__(self, algo, env):
-        super().__init__(algo, env)
-
     def start_worker(self):
         """Start workers."""
         parallel_sampler.populate_task(self.env,
@@ -28,7 +24,17 @@ class BatchSampler(BaseSampler):
         parallel_sampler.terminate_task(scope=self.algo.scope)
 
     def obtain_samples(self, itr, batch_size=None, whole_paths=True):
-        """Obtain samples."""
+        """Obtain samples.
+
+        Args:
+            itr (int): Number of iteration.
+            batch_size (int): Number of environment steps in one batch.
+            whole_paths (bool): Whether to use whole path or truncated.
+
+        Returns:
+            list[dict]: A list of paths.
+
+        """
         if not batch_size:
             batch_size = self.algo.max_path_length
 
--- a/src/garage/sampler/is_sampler.py
+++ b/src/garage/sampler/is_sampler.py
@@ -1,3 +1,4 @@
+"""Importance sampling sampler."""
 import copy
 from math import exp
 from math import log
@@ -10,14 +11,16 @@ from garage.sampler.utils import truncate_paths
 
 
 class ISSampler(BatchSampler):
-    """
+    """Importance sampling sampler.
+
     Sampler which alternates between live sampling iterations using
     BatchSampler and importance sampling iterations.
 
     Args:
         algo (garage.np.algos.RLAlgorithm): An algorithm instance.
         env (garage.envs.GarageEnv): An environement instance.
-        n_backtrack (str/int): Number of past policies to update from
+        n_backtrack (int): Number of past policies to update from.
+            If None, it uses all past policies.
         n_is_pretrain (int): Number of importance sampling iterations to
             perform in beginning of training
         init_is (bool): Set initial iteration (after pretrain) an
@@ -34,7 +37,7 @@ class ISSampler(BatchSampler):
     def __init__(self,
                  algo,
                  env,
-                 n_backtrack='all',
+                 n_backtrack=None,
                  n_is_pretrain=0,
                  init_is=0,
                  skip_is_itrs=False,
@@ -58,27 +61,51 @@ class ISSampler(BatchSampler):
 
     @property
     def history(self):
-        """
+        """list: History of policies.
+
         History of policies that have interacted with the environment and the
-        data from interaction episode(s)
+        data from interaction episode(s).
+
         """
         return self._hist
 
     def add_history(self, policy_distribution, paths):
-        """
-        Store policy distribution and paths in history
+        """Store policy distribution and paths in history.
+
+        Args:
+            policy_distribution (garage.tf.distributions.Distribution): Policy distribution. # noqa: E501
+            paths (list): Paths.
+
         """
         self._hist.append((policy_distribution, paths))
 
-    def get_history_list(self, n_past='all'):
+    def get_history_list(self, n_past=None):
+        """Get list of (distribution, data) tuples from history.
+
+        Args:
+            n_past (int): Number of past policies to update from.
+                If None, it uses all past policies.
+
+        Returns:
+            list: A list of paths.
+
         """
-        Get list of (distribution, data) tuples from history
-        """
-        if n_past == 'all':
+        if n_past is None:
             return self._hist
         return self._hist[-min(n_past, len(self._hist)):]
 
     def obtain_samples(self, itr, batch_size=None, whole_paths=True):
+        """Collect samples for the given iteration number.
+
+        Args:
+            itr (int): Number of iteration.
+            batch_size (int): Number of environment steps in one batch.
+            whole_paths (bool): Whether to use whole path or truncated.
+
+        Returns:
+            list[dict]: A list of paths.
+
+        """
         # Importance sampling for first self.n_is_pretrain iterations
         if itr < self.n_is_pretrain:
             paths = self._obtain_is_samples(itr, batch_size, whole_paths)
@@ -95,7 +122,18 @@ class ISSampler(BatchSampler):
         self._is_itr = (self._is_itr + 1) % 2
         return paths
 
-    def _obtain_is_samples(self, itr, batch_size=None, whole_paths=True):
+    def _obtain_is_samples(self, _itr, batch_size=None, whole_paths=True):
+        """Collect IS samples for the given iteration number.
+
+        Args:
+            _itr (int): Number of iteration.
+            batch_size (int): Number of batch size.
+            whole_paths (bool): Whether to use whole path or truncated.
+
+        Returns:
+            list: A list of paths.
+
+        """
         if batch_size is None:
             batch_size = self.algo.max_path_length
 
@@ -126,6 +164,21 @@ class ISSampler(BatchSampler):
                                  hist_variance_penalty=0.0,
                                  max_is_ratio=10,
                                  ess_threshold=0):
+        """Return sample of IS weighted paths.
+
+        Args:
+            policy (object): The policy.
+            hist_policy_distribution (list): Histogram policy distribution.
+            max_samples (int): Max number of samples.
+            paths (list): Paths.
+            hist_variance_penalty (float): Histogram variance penalty.
+            max_is_ratio (float): Maximum of IS ratio.
+            ess_threshold (float): Effective sample size estimate.
+
+        Returns:
+            list: A list of paths.
+
+        """
         if not paths:
             return []
 
--- a/src/garage/sampler/ray_sampler.py
+++ b/src/garage/sampler/ray_sampler.py
@@ -2,9 +2,9 @@
 
 Uses a data parallel design.
 Included is a sampler that deploys sampler workers.
-
 The sampler workers must implement some type of set agent parameters
-function, and a rollout function
+function, and a rollout function.
+
 """
 from collections import defaultdict
 import pickle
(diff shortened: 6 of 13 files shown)


src/garage/experiment/experiment_wrapper.py
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
+# pylint: skip-file
```
ryanjulian: which errors does this suffer? if this is going to be in the codebase, we need to fix it.

src/garage/experiment/local_runner.py
```diff
@@ -48,20 +48,27 @@ class LocalRunner:
 
     """
 
-    def __init__(self, snapshot_config, max_cpus=1):
+    def __init__(self, snapshot_config):
         self._snapshotter = Snapshotter(snapshot_config.snapshot_dir,
                                         snapshot_config.snapshot_mode,
                                         snapshot_config.snapshot_gap)
+        self.algo = None
```
ryanjulian: please make these _private

ryanjulian: and all others below

src/garage/experiment/local_runner.py
```diff
@@ -117,7 +123,7 @@ def obtain_samples(self, itr, batch_size=None):
                 This is a hint that the sampler may or may not respect.
 
         Returns:
-            One batch of samples.
+            list: One batch of samples.
```
ryanjulian: what's inside the list?

src/garage/experiment/local_runner.py
```diff
@@ -129,10 +135,13 @@ def save(self, epoch, paths=None):
         """Save snapshot of current batch.
 
         Args:
-            itr(int): Index of iteration (epoch).
-            paths(dict): Batch of samples after preprocessed. If None,
+            epoch (int): Index of iteration (epoch).
+            paths (dict): Batch of samples after preprocessed. If None,
```
ryanjulian: what's inside the dict?

src/garage/experiment/local_runner.py
```diff
@@ -165,7 +174,7 @@ def restore(self, from_dir, from_epoch='last'):
                 Not applicable when snapshot_mode='last'.
 
         Returns:
-            A SimpleNamespace for train()'s arguments.
+            SimpleNamespace: A SimpleNamespace for train()'s arguments.
```
ryanjulian: this type doesn't tell the user anything about the interface, which means it should probably be different.

in python this kind of type is often a namedtuple:
```python
import collections

TrainArgs = collections.namedtuple('TrainArgs', [
    'n_epochs',
    'n_epoch_cycles',
    'batch_size',
    'plot',
    'store_paths',
    'pause_for_plot',
    'start_epoch'])
```

For some reason when this was initially implemented we used `SimpleNamespace` instead of `namedtuple`. I'm not sure why. If you still *really* need `SimpleNamespace` then you can at least create a type by doing this:
```python
import types

class TrainArgs(types.SimpleNamespace):
    pass
```

Unfortunately this gives the user no information on what is in a `TrainArgs` but at least it has its own type.

src/garage/experiment/local_runner.py
```diff
@@ -234,7 +243,10 @@ def train(self,
             pause_for_plot(bool): Pause for plot.
 
         Returns:
-            The average return in last epoch cycle.
+            float: The average return in last epoch cycle.
+
+        Raises:
+            Exception: if self.has_setup is False.
```
ryanjulian: Please create a new exception class for this. You can call it LocalRunnerException.

```python
class LocalRunnerException(Exception):
    pass
```

src/garage/experiment/local_runner.py
```diff
@@ -129,10 +135,13 @@ def save(self, epoch, paths=None):
...
 
         Args:
-            itr(int): Index of iteration (epoch).
-            paths(dict): Batch of samples after preprocessed. If None,
+            epoch (int): Index of iteration (epoch).
+            paths (dict): Batch of samples after preprocessed. If None,
                 no paths will be logged to the snapshot.
 
+        Raises:
+            Exception: if self.has_setup is False.
```
ryanjulian: please create a new `LocalRunnerException` class for this (see comment below)

src/garage/experiment/local_runner.py
```diff
@@ -309,8 +321,19 @@ def resume(self,
...
+            plot (bool): weather to plot
+            store_paths (bool): whether to store paths
+            pause_for_plot(bool): Pause for plot.
+
         Returns:
-            The average return in last epoch cycle.
+            float: The average return in last epoch cycle.
+
+        Raises:
+            Exception: if restore() is not called before resume()
```
ryanjulian: Please use a special exception class.

src/garage/sampler/base.py
```diff
@@ -1,43 +1,31 @@
...
         """
-        pass
 
     @abc.abstractmethod
     def shutdown_worker(self):
         """Terminate workers if necessary."""
-        pass
-
-
-class BaseSampler(Sampler):
```
ryanjulian: why delete this? it was created as part of a normal pattern: defining an interface (`Sampler`) distinct from its most common implementation `BaseSampler`, even of the base implementation is obvious.

src/garage/sampler/batch_sampler.py
```diff
@@ -1,21 +1,30 @@
...
-    def __init__(self, algo, env):
-        super().__init__(algo, env)
+    def __init__(self, algo, env, max_cpus=1):
+        parallel_sampler.initialize(max_cpus)
+
+        seed = deterministic.get_seed()
+        if seed is not None:
+            parallel_sampler.set_seed(seed)
+
+        self.algo = algo
```
ryanjulian: please make members _private while you're here.

src/garage/sampler/is_sampler.py
```diff
@@ -1,23 +1,25 @@
+"""ISSampler."""
```
ryanjulian: Please don't just use the class name as the module comment. This should describe what's inside the module.

src/garage/sampler/is_sampler.py
```diff
@@ -1,23 +1,25 @@
...
 
-from garage.sampler import BatchSampler
+from garage.sampler import Sampler
 from garage.sampler.utils import truncate_paths
 
 
-class ISSampler(BatchSampler):
-    """
+class ISSampler(Sampler):
+    """ISSampler.
```
ryanjulian: Please don't use the class name as the first line of the comment. I'm surprised this passes the linter.

See http://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings for a guide on how to write good docstrings.

src/garage/sampler/is_sampler.py
```diff
@@ -1,23 +1,25 @@
...
+
     Sampler which alternates between live sampling iterations using
     BatchSampler and importance sampling iterations.
 
     Args:
         algo (garage.np.algos.RLAlgorithm): An algorithm instance.
         env (garage.envs.GarageEnv): An environement instance.
-        n_backtrack (str/int): Number of past policies to update from
+        n_backtrack (str): Number of past policies to update from
```
ryanjulian: i'm confused by this arg. `n_backtrack` makes it sound like a number -- how is it a string? The argument comment should at least explain this.

src/garage/sampler/is_sampler.py
```diff
@@ -54,31 +56,58 @@ def __init__(self,
         self._hist = []
         self._is_itr = init_is
 
-        super().__init__(algo, env)
+        self.algo = algo
```
ryanjulian: please make members _private while you're here.

src/garage/sampler/is_sampler.py
```diff
@@ -54,31 +56,58 @@ def __init__(self,
...
+        self.env = env
 
     @property
     def history(self):
-        """
+        """History of policies.
```
ryanjulian: Properties should be documented like attributes, not functions. See https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html for an example.

src/garage/sampler/is_sampler.py
```diff
@@ -54,31 +56,58 @@ def __init__(self,
...
         """
         return self._hist
 
     def add_history(self, policy_distribution, paths):
-        """
-        Store policy distribution and paths in history
+        """Store policy distribution and paths in history.
+
+        Args:
+            policy_distribution (object): Policy distribution.
```
ryanjulian: is it really just an object? what type usually is here?

src/garage/sampler/is_sampler.py
```diff
@@ -54,31 +56,58 @@ def __init__(self,
...
         """
         self._hist.append((policy_distribution, paths))
 
     def get_history_list(self, n_past='all'):
-        """
-        Get list of (distribution, data) tuples from history
+        """Get list of (distribution, data) tuples from history.
+
+        Args:
+            n_past (str): Number of past policies to update from
```
ryanjulian: does this also accept ints? if so it needs to be documented. Perhaps you can change the function to accept only ints and use `None` for all? That will make the documentation easier to write and the interface clearer.

src/garage/sampler/is_sampler.py
```diff
@@ -126,6 +166,21 @@ def _sample_isweighted_paths(self,
                                  hist_variance_penalty=0.0,
                                  max_is_ratio=10,
                                  ess_threshold=0):
+        """Return sample of IS weighted paths.
+
+        Args:
+            policy (object): the policy
```
ryanjulian: please make these arguments more descriptive.

src/garage/sampler/is_sampler.py
```diff
@@ -180,3 +235,9 @@ def _sample_isweighted_paths(self,
                 return []
 
         return samples
+
+    def start_worker(self):
```
ryanjulian: these methods could go in `BaseSampler`, allowing you to not define these stubs everywhere.

src/garage/sampler/ray_sampler.py
```diff
@@ -16,17 +16,20 @@
...
-        - algo: A garage algo object
-        - env: A gym/akro env object
-        - should_render(bool): should the sampler render the trajectories
-        - sampler_worker_cls: If none, uses the default SamplerWorker
+        algo (garage.np.algos.RLAlgorithm): A garage algo object.
+        env (garage.envs.GarageEnv): A gym/akro env object.
+        seed (int): Seed.
+        should_render (bool): Should the sampler render the trajectories.
+        num_processors (int): Number of processors.
+        sampler_worker_cls (Sampler): If none, uses the default SamplerWorker
```
ryanjulian: None

src/garage/sampler/ray_sampler.py
```diff
@@ -135,7 +148,11 @@ def _process_trajectory(self, result):
         Converts that trajectory to garage friendly format.
 
         Args:
-            - result: ray object id of ready to be collected trajectory.
+            result (int): ray object id of ready to be collected trajectory.
```
ryanjulian: though it's appreciated, you should be able to skip docstrings on _private methods

src/garage/sampler/ray_sampler.py
```diff
@@ -187,7 +205,11 @@ def set_agent(self, flattened_params):
         """Set the agent params.
 
         Args:
-            - flattened_params(): model parameters in numpy format
+            flattened_params(numpy.array): model parameters in numpy format
```
ryanjulian: the type is actually `numpy.ndarray`

src/garage/tf/experiment/local_tf_runner.py
```diff
@@ -91,7 +101,14 @@ def __enter__(self):
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
-        """Leave session."""
+        """Leave session.
```
ryanjulian: did pylint force you to document this?

src/garage/tf/samplers/batch_sampler.py
```diff
@@ -3,27 +3,45 @@
...
+    Args:
+        algo (garage.np.algos.RLAlgorithm): The algorithm.
+        env (gym.Env): The environment.
+        n_envs (int): Number of environments.
+
+    """
 
     def __init__(self, algo, env, n_envs):
-        super(BatchSampler, self).__init__(algo, env)
+        self.algo = algo
```
ryanjulian: please make these members _private while you're here

ryanjulian:
This look mostly good but I'm confused about the removal of `max_cpus` from LocalRunner -- the point of that argument was to allow someone to set how many CPUs are allowed to be used from the launcher file. Now they have to set it using sampler_args.

You can imagine in the future we could add `use_gpus=[1, 2]` and etc.

This argument isn't useful just for BatchSampler, for instance we should probably be creating the TF session as a multi-CPU session as [here](https://github.com/rlworkgroup/garage/pull/715/files#diff-b814f86389a2258aa2d6561cb212c3a2R184) but I don't think we currently are.

src/garage/tf/experiment/local_tf_runner.py
zequnyu: Yes...

src/garage/sampler/ray_sampler.py
zequnyu: Pylint is forcing this as well

zequnyu:
Based on the comments, I bring `max_cups` back to `LocalRunner`. However I'm a bit confused. If `LocalRunner` owns `max_cups`, we should probably delete it from `sampler`? @ryanjulian

ryanjulian:
@zequnyu is this PR still alive?

zequnyu:
Yes, I think I've addressed all your comments, if not missed. Please take a look.

ryanjulian:
@krzentner you are the expert here. PTAL

ryanjulian:
@Mergifyio rebase

src/garage/sampler/batch_sampler.py
```diff
@@ -28,7 +36,18 @@ def shutdown_worker(self):
         parallel_sampler.terminate_task(scope=self.algo.scope)
 
     def obtain_samples(self, itr, batch_size=None, whole_paths=True):
-        """Obtain samples."""
+        # pylint: disable=arguments-differ
+        """Obtain samples.
+
+        Args:
+            itr (int): number of iteration
```
naeioi: Capitalize. End with a period.

src/garage/sampler/is_sampler.py
```diff
@@ -58,27 +61,52 @@ def __init__(self,
...
-        if n_past == 'all':
+        if n_past is None:
             return self._hist
         return self._hist[-min(n_past, len(self._hist)):]
 
     def obtain_samples(self, itr, batch_size=None, whole_paths=True):
+        """Collect samples for the given iteration number.
+
+        Args:
+            itr (int): number of iteration
```
naeioi: Capitalized. End with a period.

src/garage/sampler/is_sampler.py
```diff
@@ -95,7 +123,18 @@ def obtain_samples(self, itr, batch_size=None, whole_paths=True):
         self._is_itr = (self._is_itr + 1) % 2
         return paths
 
-    def _obtain_is_samples(self, itr, batch_size=None, whole_paths=True):
+    def _obtain_is_samples(self, _itr, batch_size=None, whole_paths=True):
+        """Collect IS samples for the given iteration number.
+
+        Args:
+            _itr (int): number of iteration
```
naeioi: Same as above.

src/garage/experiment/local_runner.py
```diff
@@ -117,10 +116,12 @@ def __init__(self, snapshot_config, max_cpus=1):
                                         snapshot_config.snapshot_mode,
                                         snapshot_config.snapshot_gap)
 
-        if max_cpus > 1:
-            # pylint: disable=import-outside-toplevel
-            from garage.sampler import singleton_pool
-            singleton_pool.initialize(max_cpus)
+        parallel_sampler.initialize(max_cpus)
```
naeioi: Shouldn't this be placed in the TF Runner?

src/garage/sampler/batch_sampler.py
```diff
@@ -11,10 +12,17 @@ class BatchSampler(BaseSampler):
     Args:
         algo (garage.np.algos.RLAlgorithm): The algorithm.
         env (gym.Env): The environment.
+        max_cpus (int): The maximum number of parallel sampler workers.
 
     """
 
-    def __init__(self, algo, env):
+    def __init__(self, algo, env, max_cpus=1):
```
naeioi: How do you pass `max_cpus` from runner to sampler?

src/garage/experiment/local_runner.py
krzentner: Is this actually necessary? Can't we afford to start the sampler in `setup`, once we actually know the sampler?

ryanjulian:
I will wait for @krzentner and @naeioi to approve, since I haven't followed along well-enough to judge the PR.

naeioi:
I would like to know your opinion on the idea of `is_parallel` tag on samples. @zequnyu . It avoids duplicate initialization code, by testing this `is_parallel` and initialize a sampler only in the runner.

src/garage/experiment/local_runner.py
zequnyu: We can't set it in `setup()` at least for all tf algorithms due to tf sessions.

zequnyu:
We can set `max_cpus` in `setup()` or `sampler.__init__()` for those who don't have tf sessions. In this way, it will be:

For all tf algorithms:
- initialize `max_cpus` in `LocalTFRunner.__init__()`
- no need to initialize in `tf_batch_sampler`

For all non-tf:
- initialize `max_cpus` in `batch_sampler`
- no need to initialize in `LocalRunner.__init__()`

This is exactly the code I submitted at the first time of this PR, which will result in some duplicated code. I'm not sure whether I understand @naeioi suggestion (introducing `is_parallel` sign to sampler) correctly. I feel this is over complicated than it should be.

After some thinking I feel the best way for now is to put `parallel_sampler.initialize(max_cpus)` into `LocalRunner`. In this way:

For all algorithms:
- initialize `max_cpus` in `LocalRunner.__init__()`

The only assumption for this is we are actually assuming when user customize the `max_cpus` value, they mean to use `batch_sampler`.

In this case, all this PR does is essentially remove `n_parallels` setup from the experiment API, which is definitely unnecessary.

src/garage/tf/samplers/batch_sampler.py
```diff
@@ -9,21 +9,38 @@
 
 
 def worker_init_tf(g):
-    """Initialize the tf.Session on a worker."""
+    """Initialize the tf.Session on a worker.
+
+    Args:
+        g (LocalTFRunner): The Runner.
```
ryanjulian: is `g` actually LocalTFRunner? I thought it's just some global state object.

src/garage/sampler/batch_sampler.py
```diff
@@ -28,7 +24,18 @@ def shutdown_worker(self):
         parallel_sampler.terminate_task(scope=self.algo.scope)
 
     def obtain_samples(self, itr, batch_size=None, whole_paths=True):
-        """Obtain samples."""
+        # pylint: disable=arguments-differ
+        """Obtain samples.
+
+        Args:
+            itr (int): Number of iteration.
+            batch_size (int): Number of batch size.
```
naeioi: batch_size (int): Number of environment steps in one batch.

src/garage/sampler/is_sampler.py
naeioi: This should be `Importance sampling sampler`.

src/garage/sampler/is_sampler.py
naeioi: Can you fix this?

src/garage/sampler/batch_sampler.py
```diff
@@ -28,7 +24,18 @@ def shutdown_worker(self):
         parallel_sampler.terminate_task(scope=self.algo.scope)
 
     def obtain_samples(self, itr, batch_size=None, whole_paths=True):
-        """Obtain samples."""
+        # pylint: disable=arguments-differ
```
ryanjulian: why do the args differ from the interface?

tests/garage/tf/samplers/test_tf_batch_sampler.py
```diff
@@ -0,0 +1,56 @@
...
+from garage.tf.samplers import BatchSampler
+from tests.fixtures import snapshot_config
+
+
+class TestTFSampler:
+    # Note:
+    #   test_batch_sampler should pass if tested independently
+    #   from other tests, but cannot be tested on CI.
+    #
+    #   This is because nose2 runs all tests in a single process,
```
ryanjulian: we don't use nose anymore

Issue #799, opened 2019-07-16
IndexError when running the her_ddpg_fetchreach.py


13331151:
(garage) liemzuvon@liemzuvon-MS-7B23:~/code/garage$ ./examples/tf/her_ddpg_fetchreach.py 
WARNING: Logging before flag parsing goes to stderr.
W0716 21:13:03.309802 140498767222592 deprecation_wrapper.py:119] From /home/liemzuvon/anaconda3/envs/garage/lib/python3.6/site-packages/garage-2019.6.0.dev0-py3.6.egg/garage/tf/algos/ddpg.py:78: The name tf.train.AdamOptimizer is deprecated. Please use tf.compat.v1.train.AdamOptimizer instead.

W0716 21:13:03.642822 140498767222592 deprecation.py:506] From /home/liemzuvon/anaconda3/envs/garage/lib/python3.6/site-packages/tensorflow/python/ops/init_ops.py:1251: calling VarianceScaling.__init__ (from tensorflow.python.ops.init_ops) with dtype is deprecated and will be removed in a future version.
Instructions for updating:
Call initializer instance with the dtype argument instead of passing it to the constructor
[...]

PR #1009, opened 2019-11-06
Fix examples in tf branch

ahtsan:
This PR performs a few fixes so that all of the examples in tf branch work (except her_ddpg_fetchreach.py which is being handled as mentioned in #799).

### DIFF
--- a/src/garage/envs/base.py
+++ b/src/garage/envs/base.py
@@ -1,6 +1,7 @@
 """Wrapper class that converts gym.Env into GarageEnv."""
 
 import collections
+import copy
 
 import akro
 import gym
@@ -141,19 +142,26 @@ class GarageEnv(gym.Wrapper):
         return self.env.step(action)
 
     def __getstate__(self):
-        """See `Object.__getstate__."""
-        self.close()
-        return self.__dict__
+        """See `Object.__getstate__.
 
-    def __setstate__(self, state):
-        """See `Object.__setstate__.
-
-        Args:
-            state (dict): Unpickled state of this object.
+        Returns:
+            dict: The instance’s dictionary to be pickled.
 
         """
-        obj = type(self)(state['_env'], state['_env_name'])
-        self.__dict__.update(obj.__dict__)
+        # the viewer object is not pickleable
+        # we first make a copy of the viewer
+        env = self.env
+        # detect gym.Wrapper
+        while issubclass(env.__class__, gym.Wrapper):
+            env = env.unwrapped
+        _viewer = env.viewer
+        # remove the viewer and make a copy of the state
+        env.viewer = None
+        state = copy.deepcopy(self.__dict__)
+        # assign the viewer back to self.__dict__
+        env.viewer = _viewer
+        # the returned state doesn't have the viewer
+        return state
 
 
 def Step(observation, reward, done, **kwargs):  # noqa: N802
--- a/src/garage/envs/base.py
+++ b/src/garage/envs/base.py
@@ -151,8 +151,8 @@ class GarageEnv(gym.Wrapper):
         # the viewer object is not pickleable
         # we first make a copy of the viewer
         env = self.env
-        # detect gym.Wrapper
-        while issubclass(env.__class__, gym.Wrapper):
+        # get the inner env if it is a gym.Wrapper
+        if issubclass(env.__class__, gym.Wrapper):
             env = env.unwrapped
         if 'viewer' in env.__dict__:
             _viewer = env.viewer
--- a/src/garage/envs/base.py
+++ b/src/garage/envs/base.py
@@ -154,14 +154,16 @@ class GarageEnv(gym.Wrapper):
         # detect gym.Wrapper
         while issubclass(env.__class__, gym.Wrapper):
             env = env.unwrapped
-        _viewer = env.viewer
-        # remove the viewer and make a copy of the state
-        env.viewer = None
-        state = copy.deepcopy(self.__dict__)
-        # assign the viewer back to self.__dict__
-        env.viewer = _viewer
-        # the returned state doesn't have the viewer
-        return state
+        if 'viewer' in env.__dict__:
+            _viewer = env.viewer
+            # remove the viewer and make a copy of the state
+            env.viewer = None
+            state = copy.deepcopy(self.__dict__)
+            # assign the viewer back to self.__dict__
+            env.viewer = _viewer
+            # the returned state doesn't have the viewer
+            return state
+        return self.__dict__
 
 
 def Step(observation, reward, done, **kwargs):  # noqa: N802
--- a/src/garage/envs/base.py
+++ b/src/garage/envs/base.py
@@ -140,6 +140,11 @@ class GarageEnv(gym.Wrapper):
         """
         return self.env.step(action)
 
+    def __getstate__(self):
+        """See `Object.__getstate__."""
+        self.close()
+        return self.__dict__
+
     def __setstate__(self, state):
         """See `Object.__setstate__.
 
--- a/src/garage/misc/tensor_utils.py
+++ b/src/garage/misc/tensor_utils.py
@@ -138,19 +138,6 @@ def pad_tensor_dict(tensor_dict, max_len, mode='zero'):
     return ret
 
 
-def stack_tensor_list(tensor_list):
-    """Stack tensor list.
-
-    Args:
-        tensor_list (list[numpy.ndarray]): List of tensors.
-
-    Returns:
-        numpy.ndarray: Stacked list of tensor.
-
-    """
-    return np.array(tensor_list)
-
-
 def stack_tensor_dict_list(tensor_dict_list):
     """Stack a list of dictionaries of {tensors or dictionary of tensors}.
 
@@ -171,22 +158,9 @@ def stack_tensor_dict_list(tensor_dict_list):
         if isinstance(example, dict):
             v = stack_tensor_dict_list(dict_list)
         else:
-            v = stack_tensor_list(dict_list)
+            v = np.array(dict_list)
         ret[k] = v
     return ret
-
-
-def concat_tensor_list(tensor_list):
-    """Concatenate list of tensor.
-
-    Args:
-        tensor_list (list[numpy.ndarray]): List of tensors.
-
-    Returns:
-        numpy.ndarray: list of tensor.
-
-    """
-    return np.concatenate(tensor_list, axis=0)
 
 
 def concat_tensor_dict_list(tensor_dict_list):
@@ -209,7 +183,7 @@ def concat_tensor_dict_list(tensor_dict_list):
         if isinstance(example, dict):
             v = concat_tensor_dict_list(dict_list)
         else:
-            v = concat_tensor_list(dict_list)
+            v = np.concatenate(dict_list, axis=0)
         ret[k] = v
     return ret
 
--- a/src/garage/misc/tensor_utils.py
+++ b/src/garage/misc/tensor_utils.py
@@ -1,35 +1,77 @@
+"""Utiliy functions for tensors."""
 import gym.spaces
 import numpy as np
 import scipy.signal
 
 
 def discount_cumsum(x, discount):
-    # See https://docs.scipy.org/doc/scipy/reference/tutorial/signal.html#difference-equation-filtering  # noqa: E501
-    # Here, we have y[t] - discount*y[t+1] = x[t]
-    # or rev(y)[t] - discount*rev(y)[t-1] = rev(x)[t]
+    """Discounted cumulative sum.
+
+    See https://docs.scipy.org/doc/scipy/reference/tutorial/signal.html#difference-equation-filtering  # noqa: E501
+    Here, we have y[t] - discount*y[t+1] = x[t]
+    or rev(y)[t] - discount*rev(y)[t-1] = rev(x)[t]
+
+    Args:
+        x (np.ndarrary): Input.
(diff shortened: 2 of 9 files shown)


src/garage/tf/q_functions/continuous_mlp_q_function.py
```diff
@@ -138,6 +155,11 @@ def clone(self, name):
 
         Args:
             name (str): Name of the newly created q-function.
+
+        Returns:
+            object: A new instance of ContinuousMLPQFunction with same
```
ryanjulian: shouldn't this be ContinuousMLPQFunction

src/garage/misc/tensor_utils.py
```diff
@@ -1,35 +1,77 @@
...
     if np.isclose(vary, 0):
         if np.var(ypred) > 0:
             return 0
-        else:
-            return 1
+        return 1
     return 1 - np.var(y - ypred) / (vary + 1e-8)
 
 
 def flatten_tensors(tensors):
```
ryanjulian: doesn't this flatten a collection of tensors?

src/garage/misc/tensor_utils.py
```diff
@@ -1,35 +1,77 @@
...
+
+    """
     if tensors:
         return np.concatenate([np.reshape(x, [-1]) for x in tensors])
-    else:
-        return np.asarray([])
+    return np.asarray([])
 
 
 def unflatten_tensors(flattened, tensor_shapes):
```
ryanjulian: doesn't this unflatten a collection of tensors?

src/garage/misc/tensor_utils.py
```diff
@@ -64,90 +138,94 @@ def pad_tensor_dict(tensor_dict, max_len, mode='zero'):
...
+            v = stack_tensor_dict_list(dict_list)
         else:
-            v = concat_tensor_list_subsample([x[k] for x in tensor_dict_list],
-                                             f)
+            v = stack_tensor_list(dict_list)
         ret[k] = v
     return ret
 
 
 def concat_tensor_list(tensor_list):
```
ryanjulian: is this equivalent to np.stack?

src/garage/misc/tensor_utils.py
```diff
@@ -163,10 +241,32 @@ def split_tensor_dict_list(tensor_dict):
...
 
 def truncate_tensor_list(tensor_list, truncated_len):
+    """Truncate list of tensor.
+
+    Args:
+        tensor_list (list[numpy.ndarray]): List of tensors.
+        truncated_len (int): Length to truncate.
+
+    Returns:
+        numpy.ndarray: list of tensor.
```
ryanjulian: doesn't this return a list, trivially?

src/garage/misc/tensor_utils.py
```diff
@@ -163,10 +241,32 @@ def split_tensor_dict_list(tensor_dict):
...
+
+    """
     return tensor_list[:truncated_len]
```
ryanjulian: why wouldn't i just splice my list just like this?

ahtsan: Yes, that is. For most of the functions, I didn't think about refactoring/replacing with the equivalent np function since this PR mainly addressed the errors when running the examples.

ahtsan:
Ok I will address the comments. Parts for fixing samples:
- `garage.Env` call `self.close()` when pickling since the viewer is not pickleable. If experiment is run with `plot=True` it will break.
- `stack_tensor_dict_list` and `concat_tensor_dict_list` assumes `env_info` from all parallel envs contain the same key but it's not true.
- `ContinuousMLPQFunction` doesn't pass `self._input_include_goal` when it's cloned.

src/garage/envs/base.py
```diff
@@ -140,6 +140,11 @@ def step(self, action):
         """
         return self.env.step(action)
 
+    def __getstate__(self):
+        """See `Object.__getstate__."""
+        self.close()
```
krzentner: I'm not convinced this is the right thing to do. Pickling an object shouldn't modify its state.

Also, most environments already know how to pickle themselves, so it's probably best to just rely on that (and not define `__getstate__` and `__setstate__` for this class).

Is there some environment in particular that needs this?

ahtsan: after a second thought, actually it meant to be closing its viewer

ryanjulian: Hmm yeah why do we need to close the environment to pickle it? The original env should be able to live on.

ahtsan: This is meant to be closing the viewer

ryanjulian: It's not valid to call *any* method on `gym.Env` after calling `gym.Env.close()`. That means your environment is inoperable after pickling.

Also, why should the viewer close when I pickle it? What if I was using that viewer?

ahtsan: To make it clear, instead of calling `self.close()`, we call `self._close_viewer_window`
(https://github.com/rlworkgroup/garage/blob/7b6d7a528f623d8e6c985cf7ee4d39c4b73fdebd/src/garage/envs/base.py#L70).

If a viewer object is not available when needed, it will create one, e.g. https://github.com/openai/gym/blob/master/gym/envs/box2d/car_racing.py#L345

ryanjulian: we can't close the viewer while we're pickling. you can make a copy of `self.__dict__` and delete the viewer reference from that copy, but you can't close the viewer.

imagine i'm running a training loop like the one below. the viewer window would flicker on the screen.
```python
env = gym.make('Foo-v1')
env.render()  # opens viewer window
algo = FooAlgo()

while training:
    paths = obtain_samples(policy, env, batch_size)
    algo.train_once(paths)
    save_snapshot(env, policy)  # closes viewer window
    rollout(env, policy, max_path_length=100, render=True)  # opens viewer window
```

ahtsan: emm ok. It makes sense. I will remove it from `self.__dict__`.

ahtsan: turns out we can't remove the viewer object, gym.Env assumes it's always there. If we remove that, we will get
```
    if self.viewer is None:
AttributeError: 'CartPoleEnv' object has no attribute 'viewer'
```

ryanjulian: can you remove it from a copy of self.__dict__?

ahtsan: At the same time we pickle and unpickle the envs and use it
https://github.com/rlworkgroup/garage/blob/master/src/garage/sampler/on_policy_vectorized_sampler.py#L29
so we can't really remove the viewer attribute, but at the same time we can't have it being a `viewer` object.

ahtsan: How about
```python
    def __getstate__(self):
        """See `Object.__getstate__."""
        _state = self.__dict__.copy()
        _state['env'].__dict__['env'].__dict__['viewer'] = None
        return _state

        """See `Object.__setstate__.

        Args:
            state (dict): Unpickled state of this object.

        """
        obj = type(self)(state['_env'], state['_env_name'])
        self.__dict__.update(obj.__dict__)
```

krzentner: You need to do something like this:
```python
state = self.__dict__.copy()
if 'viewer' in state:
    state['viewer'] = None
return state
```

ahtsan: yes, just the viewer object is deep down, inside the nested env

krzentner: You probably need something like this:
```python
    def __getstate__(self):
        """See `Object.__getstate__."""
        state = self.__dict__.copy()
        state['env'] = copy.copy(state['env'])
        if 'viewer' in state['env'].__dict__:
            state['env'].__dict__ = None
        return state

        """See `Object.__setstate__.

        Args:
            state (dict): Unpickled state of this object.

        """
        self.__dict__ = state
```

ahtsan: why do we need to copy `state['env']`? If we already have a copy of `self.__dict__`, shouldn't we just assign the viewer to `None`? like
```python
    def __getstate__(self):
        """See `Object.__getstate__."""
        state = self.__dict__.copy()
        # state['env'] = copy.copy(state['env'])  
        if 'viewer' in state['env'].__dict__['env'].__dict__:
            state['env'].__dict__['env'].__dict__['viewer'] = None
        return state
```
And we don't even need `__setstate__` anymore

krzentner: The copy of `self.__dict__` is shallow, so `state['env']` still refers to the original environment. If you modify that, you're modifying the original environment.

ahtsan: How about we change the first line to
state = self.__dict__.deepcopy()

ahtsan: Turns out we can't deep copy the dict if a viewer instance is in there. Here's my solution
```python
    def __getstate__(self):
        # we first make a copy of the viewer
        _viewer = self.__dict__['env'].__dict__['env'].__dict__['viewer']
        # remove the viewer and make a copy of the state
        self.__dict__['env'].__dict__['env'].__dict__['viewer'] = None
        state = copy.deepcopy(self.__dict__)
        # assign the viewer back to self.__dict__
        self.__dict__['env'].__dict__['env'].__dict__['viewer'] = _viewer
        # the returned state doesn't have the viewer
        return state
```

ryanjulian: i don't think this *exact* solution can work in general, because we have no idea how many wrappers are around `env` there are.

You could detect a `gym.Wrapper` and use `Wrapper.unwrapped` though.

Which gym envs have this problem? is it only the MuJoCo ones?

ryanjulian: also, do you have any insight into why this passes https://github.com/rlworkgroup/garage/blob/master/tests/garage/tf/envs/test_base.py?

krzentner: Yeah, to begin with all environments should be pickleable without need for a wrapper.

If this continues to be a problem, perhaps we should use constructor objects after all. Alternatively, we could monkey patch the viewer type to be "pickleable" by throwing away its state.

ahtsan: > https://github.com/rlworkgroup/garage/blob/master/tests/garage/tf/envs/test_base.py

I got segmentation fault when running the test.

src/garage/envs/base.py
```diff
@@ -154,14 +154,16 @@ def __getstate__(self):
         # detect gym.Wrapper
         while issubclass(env.__class__, gym.Wrapper):
```
ryanjulian: AFAICT this statement should only ever execute once, because `env.unwrapped` is recursive: https://github.com/openai/gym/blob/master/gym/core.py#L253

ahtsan:
@krzentner It's ready for review

src/garage/sampler/on_policy_vectorized_sampler.py
```diff
@@ -34,14 +43,33 @@ def start_worker(self):
 
         self.vec_env = VecEnvExecutor(
             envs=envs, max_path_length=self.algo.max_path_length)
-        self.env_spec = self.env.spec
 
     def shutdown_worker(self):
         """Shutdown workers."""
         self.vec_env.close()
 
+    # pylint: disable=too-many-statements, import-outside-toplevel
```
ryanjulian: please move `import time` to the top of the file. there's no need for it to be a late import.

ryanjulian:
since you have my review, you can get another review from anyone on the team to merge (not just KR).

src/garage/sampler/on_policy_vectorized_sampler.py
```diff
@@ -122,6 +149,5 @@ def obtain_samples(self, itr, batch_size=None, whole_paths=True):
 
         if whole_paths:
```
zequnyu: Could be one line to make it consistent with other samplers:
`return paths if whole_paths else truncate_paths(paths, batch_size)`

src/garage/envs/base.py
```diff
@@ -140,15 +141,29 @@ def step(self, action):
         """
         return self.env.step(action)
 
-    def __setstate__(self, state):
```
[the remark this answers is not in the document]
ahtsan: In `__getstate__`, we didn't remove the viewer object for `self` (original env object), rather we make a copy and return that copy as the state. After unpickling, the viewer object in the newly constructed env is gone, there is no way we can populate it back unless we construct a new one (because we remove that in the state). However, the viewer in `self` is still there.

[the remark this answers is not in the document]
ahtsan: @nish21 Does that make sense?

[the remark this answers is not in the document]
nish21: But could you verify the behavior with current __setstate__? I think the current implementation of setstate restores the viewer as it's creating a new object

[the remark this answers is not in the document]
nish21: I think what I did was, while pickling, I don't pickle the viewer object -- remove viewer from the copy of dict. But during setstate, i restore the viewer and updating the params with the pickled object params.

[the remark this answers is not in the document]
ahtsan: How did you restore the viewer object? Where is the viewer object kept and how is it passed to the newly unpickled env?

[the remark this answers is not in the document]
ryanjulian: to be clear, we probably don't want to restore viewer state on environments while unpickling. The most common pickling case for an environment is to transfer it from the main training to a worker process. We don't want a worker process spending cycles on rendering.

ryanjulian:
@ahtsan when you are done with this, mark it with label `ready-to-merge` and it should get auto-merged.

ryanjulian:

PR #1010, opened 2019-11-06
Add MT-TRPO example

### LINKED ISSUE #872: Implement MT-TRPO

utkarshjp7:
This PR fixes #872

### DIFF
--- a/examples/tf/multi_env_trpo.py
+++ b/examples/tf/multi_env_trpo.py
@@ -1,0 +1,46 @@
+#!/usr/bin/env python3
+"""This is an example to train multiple tasks with TRPO algorithm."""
+from garage.envs import normalize
+from garage.envs import PointEnv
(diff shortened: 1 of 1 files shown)


examples/tf/multi_env_trpo.py
```diff
@@ -0,0 +1,51 @@
...
+
+    Args:
+        snapshot_config (garage.experiment.SnapshotConfig): The snapshot
+            configuration used by LocalRunner to create the snapshotter.
+
+        _ (object): Ignored by this function.
+
+    """
+    with LocalTFRunner(snapshot_config=snapshot_config) as runner:
+        env1 = TfEnv(normalize(gym.make('Adventure-ram-v4')))
```
ryanjulian: maybe you could use PointEnv for these in the future, since this is unlikely to actually train to any solution...

PR #1017, opened 2019-11-09
Add stats of total env steps

naeioi:
This PR adds a `TotalEnvSteps` to log. 

Changelist:
- Add a `ExperimentStats` class to LocalRunner. This class records statistics of an experiment. Currently there are `total_itr`, `total_env_steps`, `total_epoch` and `last_path`.
- Add `TrainArgs` and `SetupArgs` to LocalRunner. They keep arguments passed to `train()` and `setup()`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,7 +12,7 @@ repos:
     -   id: check-pylint-normal
         name: Check pylint (except unit tests)
         types: [file, python]
-        exclude: ^tests/garage/.|tests/fixtures/.*$  # exclude unit tests
+        exclude: ^tests/garage/.*$  # exclude unit tests
         require_serial: true  # pylint does its own parallelism
         language: system
         entry: pylint
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,7 +12,7 @@ repos:
     -   id: check-pylint-normal
         name: Check pylint (except unit tests)
         types: [file, python]
-        exclude: ^tests/garage/.*$  # exclude unit tests
+        exclude: ^tests/garage/.|tests/fixtures/.*$  # exclude unit tests
         require_serial: true  # pylint does its own parallelism
         language: system
         entry: pylint
--- a/.pylintrc
+++ b/.pylintrc
@@ -26,7 +26,7 @@ disable =
     too-many-arguments,
     too-many-locals,
     # Detection seems buggy or unhelpful
-    duplicate-code,
+    duplicate-code
 
 [REPORTS]
 msg-template = {path}:{line:3d},{column}: {msg} ({symbol})
--- a/scripts/garage
+++ b/scripts/garage
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 """Script for resuming training from command line."""
+# pylint: disable=no-name-in-module
 import click
 
 from garage.experiment import run_experiment
@@ -8,23 +9,26 @@ from garage.tf.experiment import LocalTFRunner
 
 @click.group()
 def cli():  # noqa: D103
-    pass
+    """The main command group."""
 
 
 @cli.command()
 @click.argument('from_dir')
 @click.option(
-    '--from_epoch',
-    default='last',
-    help='Index of epoch to restore from. Can be "first", "last" or a number. '
-    'Not applicable when snapshot_mode="last"')
-@click.option(
     '--log_dir',
     default=None,
-    help='Path to save the log snapshot. If not specified, will be the same '
+    help='Log path for resumed experiment. If not specified, will be the same '
     'as from_dir.')
+# pylint: disable=bad-docstring-quotes
+@click.option('--from_epoch',
+              default='last',
+              help='When there are multiple snapshots, '
+              'specify the index of epoch to restore from. '
+              'Can be "first", "last" or a number. '
+              'Not applicable when snapshot_mode="last"')
 def resume(from_dir, from_epoch, log_dir):
-    """Resume from command line."""
+    # pylint: disable=missing-param-doc, missing-type-doc
+    """Resume from experiment saved in FROM_DIR."""
     if log_dir is None:
         log_dir = from_dir
 
--- a/src/garage/experiment/local_runner.py
+++ b/src/garage/experiment/local_runner.py
@@ -1,11 +1,75 @@
 """Provides algorithms with access to most of garage's features."""
 import copy
 import time
-import types
 
 from dowel import logger, tabular
 
+from garage.experiment.deterministic import get_seed, set_seed
 from garage.experiment.snapshotter import Snapshotter
+
+
+class ExperimentStats:
+    # pylint: disable=too-few-public-methods
+    """Statistics of a experiment.
+
+    Args:
+        total_epoch (int): Total epoches.
+        total_itr (int): Total Iterations.
+        total_env_steps (int): Total environment steps collected.
+        last_path (list[dict]): Last sampled paths.
+
+    """
+
+    def __init__(self, total_epoch, total_itr, total_env_steps, last_path):
+        self.total_epoch = total_epoch
+        self.total_itr = total_itr
+        self.total_env_steps = total_env_steps
+        self.last_path = last_path
+
+
+class SetupArgs:
+    # pylint: disable=too-few-public-methods
+    """Arguments to setup a runner.
+
+    Args:
+        sampler_cls (garage.sampler.Sampler): A sampler class.
+        sampler_args (dict): Arguments to be passed to sampler constructor.
+        seed (int): Random seed.
+
+    """
+
+    def __init__(self, sampler_cls, sampler_args, seed):
+        self.sampler_cls = sampler_cls
+        self.sampler_args = sampler_args
+        self.seed = seed
+
+
+class TrainArgs:
+    # pylint: disable=too-few-public-methods
+    """Arguments to call train() or resume().
+
+    Args:
+        n_epochs (int): Number of epochs.
+        n_epoch_cycles (int): Number of batches of samples in each epoch.
+            This is only useful for off-policy algorithm.
+            For on-policy algorithm this value should always be 1.
+        batch_size (int): Number of environment steps in one batch.
+        plot (bool): Visualize policy by doing rollout after each epoch.
+        store_paths (bool): Save paths in snapshot.
+        pause_for_plot (bool): Pause for plot.
+        start_epoch (int): The starting epoch. Used for resume().
+
+    """
+
+    def __init__(self, n_epochs, n_epoch_cycles, batch_size, plot, store_paths,
+                 pause_for_plot, start_epoch):
+        self.n_epochs = n_epochs
+        self.n_epoch_cycles = n_epoch_cycles
+        self.batch_size = batch_size
+        self.plot = plot
+        self.store_paths = store_paths
+        self.pause_for_plot = pause_for_plot
+        self.start_epoch = start_epoch
 
 
 class LocalRunner:
@@ -54,13 +118,29 @@ class LocalRunner:
                                         snapshot_config.snapshot_gap)
 
         if max_cpus > 1:
+            # pylint: disable=import-outside-toplevel
             from garage.sampler import singleton_pool
             singleton_pool.initialize(max_cpus)
-        self.has_setup = False
-        self.plot = False
+        self._has_setup = False
+        self._plot = False
 
         self._setup_args = None
-        self.train_args = None
+        self._train_args = None
+        self._stats = ExperimentStats(total_itr=0,
+                                      total_env_steps=0,
+                                      total_epoch=0,
+                                      last_path=None)
+
+        self._algo = None
+        self._env = None
+        self._policy = None
+        self._sampler = None
+        self._plotter = None
+
+        self._start_time = None
+        self._itr_start_time = None
+        self.step_itr = None
+        self.step_path = None
 
     def setup(self, algo, env, sampler_cls=None, sampler_args=None):
         """Set up runner for algorithm and environment.
@@ -79,77 +159,84 @@ class LocalRunner:
             sampler_args (dict): Arguments to be passed to sampler constructor.
 
         """
-        self.algo = algo
-        self.env = env
-        self.policy = self.algo.policy
+        self._algo = algo
+        self._env = env
+        self._policy = self._algo.policy
 
         if sampler_args is None:
             sampler_args = {}
         if sampler_cls is None:
             sampler_cls = algo.sampler_cls
-        self.sampler = sampler_cls(algo, env, **sampler_args)
+        self._sampler = sampler_cls(algo, env, **sampler_args)
 
-        self.has_setup = True
+        self._has_setup = True
 
-        self._setup_args = types.SimpleNamespace(sampler_cls=sampler_cls,
-                                                 sampler_args=sampler_args)
+        self._setup_args = SetupArgs(sampler_cls=sampler_cls,
+                                     sampler_args=sampler_args,
+                                     seed=get_seed())
 
     def _start_worker(self):
         """Start Plotter and Sampler workers."""
-        self.sampler.start_worker()
-        if self.plot:
+        self._sampler.start_worker()
+        if self._plot:
+            # pylint: disable=import-outside-toplevel
             from garage.tf.plotter import Plotter
-            self.plotter = Plotter(self.env, self.policy)
-            self.plotter.start()
+            self._plotter = Plotter(self._env, self._policy)
+            self._plotter.start()
 
     def _shutdown_worker(self):
         """Shutdown Plotter and Sampler workers."""
-        self.sampler.shutdown_worker()
-        if self.plot:
-            self.plotter.close()
+        self._sampler.shutdown_worker()
+        if self._plot:
+            self._plotter.close()
 
     def obtain_samples(self, itr, batch_size=None):
         """Obtain one batch of samples.
 
         Args:
-            itr(int): Index of iteration (epoch).
-            batch_size(int): Number of steps in batch.
+            itr (int): Index of iteration (epoch).
+            batch_size (int): Number of steps in batch.
                 This is a hint that the sampler may or may not respect.
 
         Returns:
-            One batch of samples.
+            list[dict]: One batch of samples.
 
         """
-        if self.train_args.n_epoch_cycles == 1:
+        if self._train_args.n_epoch_cycles == 1:
             logger.log('Obtaining samples...')
-        return self.sampler.obtain_samples(
-            itr, (batch_size or self.train_args.batch_size))
 
-    def save(self, epoch, paths=None):
+        paths = self._sampler.obtain_samples(
+            itr, (batch_size or self._train_args.batch_size))
+
+        self._stats.total_env_steps += sum([len(p['rewards']) for p in paths])
+
+        return paths
+
+    def save(self, epoch):
         """Save snapshot of current batch.
 
         Args:
-            itr(int): Index of iteration (epoch).
-            paths(dict): Batch of samples after preprocessed. If None,
-                no paths will be logged to the snapshot.
+            epoch (int): Epoch.
+
+        Raises:
+            NotSetupError: if save() is called before the runner is set up.
 
         """
-        if not self.has_setup:
-            raise Exception('Use setup() to setup runner before saving.')
+        if not self._has_setup:
+            raise NotSetupError('Use setup() to setup runner before saving.')
 
         logger.log('Saving snapshot...')
 
         params = dict()
         # Save arguments
         params['setup_args'] = self._setup_args
-        params['train_args'] = self.train_args
+        params['train_args'] = self._train_args
+        params['stats'] = self._stats
 
         # Save states
-        params['env'] = self.env
-        params['algo'] = self.algo
-        if paths:
-            params['paths'] = paths
-        params['last_epoch'] = epoch
+        params['env'] = self._env
+        params['algo'] = self._algo
+
         self._snapshotter.save_itr_params(epoch, params)
 
         logger.log('Saved')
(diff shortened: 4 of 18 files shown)


.pylintrc
```diff
@@ -21,12 +21,16 @@ disable =
...
+    # Algorithms and neural networks generally have complex logic
     too-many-instance-attributes,
     too-many-arguments,
     too-many-locals,
+    too-many-branches,
+    too-many-statements,
     # Detection seems buggy or unhelpful
     duplicate-code,
+    # Helper functions without a self use are common in base class
+    no-self-use
```
ryanjulian: i'm not convinced about this one. Perhaps this is a sign they should be moved into the module as helper functions. W

scripts/garage
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
+# pylint: skip-file
```
ryanjulian: why? it can be linted like anything else.

src/garage/experiment/local_runner.py
```diff
@@ -1,10 +1,12 @@
 """Provides algorithms with access to most of garage's features."""
+# pylint: disable=attribute-defined-outside-init, import-outside-toplevel
```
ryanjulian: please only use `# pylint: disable=` on a per-line basis. Per-file ignores mask future problems.

src/garage/experiment/local_runner.py
```diff
@@ -60,7 +62,11 @@ def __init__(self, snapshot_config, max_cpus=1):
         self.plot = False
 
         self._setup_args = None
-        self.train_args = None
+        self._train_args = None
+        self._stats = types.SimpleNamespace(total_itr=0,
```
ryanjulian: is there a reason to keep using `SimplerNamespace` over `namedtuple`? I think `SimpleNamespace` messes with type inference.

src/garage/experiment/local_runner.py
```diff
@@ -117,21 +124,28 @@ def obtain_samples(self, itr, batch_size=None):
...
+        if self._train_args.n_epoch_cycles == 1:
             logger.log('Obtaining samples...')
-        return self.sampler.obtain_samples(
-            itr, (batch_size or self.train_args.batch_size))
 
-    def save(self, epoch, paths=None):
+        paths = self.sampler.obtain_samples(
+            itr, (batch_size or self._train_args.batch_size))
+
+        for path in paths:
```
ryanjulian:
```python
self._total_env_steps += sum([p['rewards'] for path in path])
```

src/garage/experiment/local_runner.py
```diff
@@ -60,7 +62,11 @@ def __init__(self, snapshot_config, max_cpus=1):
         self.plot = False
```
ryanjulian: while you're here, can you make all of these attrs `_private`? you can use @property if they need to be part of the API.

src/garage/experiment/local_runner.py
```diff
@@ -165,39 +178,47 @@ def restore(self, from_dir, from_epoch='last'):
...
+        n_epoch_cycles = self._train_args.n_epoch_cycles
+        batch_size = self._train_args.batch_size
+        store_paths = self._train_args.store_paths
+        pause_for_plot = self._train_args.pause_for_plot
 
         fmt = '{:<20} {:<15}'
         logger.log('Restore from snapshot saved in %s' %
                    self._snapshotter.snapshot_dir)
-        logger.log(fmt.format('Train Args', 'Value'))
+        logger.log(fmt.format('-- Train Args --', '-- Value --'))
```
ryanjulian: hmm maybe we can add a `dowel` feature which recognizes a format string and implements this API...

src/garage/experiment/local_runner.py
ryanjulian: ...but perhaps the sampler should just return how many steps it collected?

src/garage/experiment/snapshotter.py
```diff
@@ -41,21 +41,46 @@ def __init__(self,
 
     @property
     def snapshot_dir(self, ):
```
ryanjulian: why all the trailing commas?

src/garage/experiment/snapshotter.py
```diff
@@ -41,21 +41,46 @@ def __init__(self,
...
-        """Save the parameters if at the right iteration."""
+        """Save the parameters if at the right iteration.
+
+        Args:
+            itr (int): Number of iterations. Used as the index of snapshot.
+            params (obj): Content of snapshot to be saved.
+
+        Raises:
+            ValueError: If snapshot_mode is not one of "all", "last" and "gap".
+            OSError: If fail to remove the old snapshot file.
```
ryanjulian: for this `OSError` and the `Exception` below, I think it would be simpler if you just defined `class SnapshotterException(Exception): pass` and then raised those.

src/garage/experiment/local_runner.py
```diff
@@ -117,21 +124,28 @@ def obtain_samples(self, itr, batch_size=None):
...
         """Save snapshot of current batch.
 
         Args:
-            itr(int): Index of iteration (epoch).
-            paths(dict): Batch of samples after preprocessed. If None,
-                no paths will be logged to the snapshot.
+            epoch(int): Epoch.
+
+        Raises:
+            Exception: if save() is called before the runner is set up.
```
ryanjulian: please define a more specific exception class for this module. Using `Exception` means that clients have to `catch(Exception):` which is an antipattern.

src/garage/experiment/local_runner.py
```diff
@@ -309,23 +336,36 @@ def resume(self,
...
-        self.train_args.n_epoch_cycles = (n_epoch_cycles
-                                          or self.train_args.n_epoch_cycles)
+        self._train_args.n_epochs = n_epochs or self._train_args.n_epochs
+        self._train_args.batch_size = batch_size or self._train_args.batch_size
+        self._train_args.n_epoch_cycles = (n_epoch_cycles
+                                           or self._train_args.n_epoch_cycles)
 
         if plot is not None:
-            self.train_args.plot = plot
+            self._train_args.plot = plot
```
ryanjulian: `self._train_args.plot = plot or self._train_args.plot`

src/garage/sampler/off_policy_vectorized_sampler.py
```diff
@@ -60,6 +64,7 @@ def shutdown_worker(self):
         self.vec_env.close()
 
     def obtain_samples(self, itr, batch_size):
+        # pylint: disable=arguments-differ
```
ryanjulian: why?

src/garage/np/algos/off_policy_rl_algorithm.py
```diff
@@ -79,26 +82,34 @@ def train(self, runner):
...
+
+        Args:
+            paths (list[dict]): A list of collected paths.
+
+        """
         self.policy.log_diagnostics(paths)
         self.qf.log_diagnostics(paths)
 
     def process_samples(self, itr, paths):
+        # pylint: disable=unused-argument
```
ryanjulian: a better pattern i have seen for this is to do `del unused_argument` which helps the static analyzer see that the arg is now out of scope and avoids the error. all in one line of code.

src/garage/tf/algos/batch_polopt.py
```diff
@@ -264,10 +275,14 @@ def init_opt(self):
         """
         raise NotImplementedError
 
-    def get_itr_snapshot(self, itr):
-        """Get all the data that should be saved in this snapshot iteration."""
-        raise NotImplementedError
-
     def optimize_policy(self, itr, samples_data):
-        """Optimize the policy using the samples."""
+        # pylint: disable=unused-argument
```
ryanjulian: try 
```python
    del itr
    del samples_data
```

src/garage/experiment/local_runner.py
ryanjulian: wait, in general you should use `tabular` for this kind of formatting. is there a reason you didn't?

src/garage/experiment/local_runner.py
ryanjulian: ...and why put these into their own type anyway? can they just be fields or a dict?

i see why you want to group them, and why not to use `namedtuple` (because it's immutable). perhaps you could just define real classes for these though. it's not much boilerplate and it would make it easier to see/validate exactly what you expect to be in these variables.

i think as they get bigger, these types are starting to hide some complexity/assumptions.

src/garage/experiment/local_runner.py
```diff
@@ -165,39 +178,47 @@ def restore(self, from_dir, from_epoch='last'):
                 Not applicable when snapshot_mode='last'.
 
         Returns:
-            A SimpleNamespace for train()'s arguments.
+            types.SimpleNamespace: A SimpleNamespace for train()'s arguments.
```
ryanjulian: i think it is starting to become time to define `class TrainArgs:` --- this return type doesn't tell me anything about what to expect. with `TrainArgs` I can just look at the type.

src/garage/experiment/local_runner.py
```diff
@@ -165,39 +178,47 @@ def restore(self, from_dir, from_epoch='last'):
...
+        logger.log(fmt.format('-- Train Args --', '-- Value --'))
         logger.log(fmt.format('n_epochs', n_epochs))
         logger.log(fmt.format('last_epoch', last_epoch))
         logger.log(fmt.format('n_epoch_cycles', n_epoch_cycles))
         logger.log(fmt.format('batch_size', batch_size))
         logger.log(fmt.format('store_paths', store_paths))
         logger.log(fmt.format('pause_for_plot', pause_for_plot))
+        logger.log(fmt.format('-- Stats --', '-- Value --'))
+        logger.log(fmt.format('last_itr', last_itr))
+        logger.log(fmt.format('total_env_steps', total_env_steps))
```
ryanjulian: this is only useful to us if it's logged to `tabular`, because `tabular` logs are sent to TensorBoardOutput (where we can use this key for the x-axis). 

there is a separate issue to unify `tabular` and `log` in dowel (by adding either logger.logkv(key, value) or logger.log(value, key=None)` options, but i think it can be out-of-scope for this PR

src/garage/experiment/local_runner.py
```diff
@@ -233,21 +254,24 @@ def train(self,
             store_paths(bool): Save paths in snapshot.
             pause_for_plot(bool): Pause for plot.
 
+        Raises:
+            Exception: If train() is called before setup().
```
ryanjulian: definitely define an Exception type for this module

src/garage/experiment/local_runner.py
```diff
@@ -274,26 +298,29 @@ def step_epochs(self):
...
+        for epoch in range(self._train_args.start_epoch,
+                           self._train_args.n_epochs):
+            self._itr_start_time = time.time()
+            with logger.prefix('epoch #%d | ' % epoch):
+                yield epoch
+                save_path = (self.step_path
+                             if self._train_args.store_paths else None)
+
+                self._stats.last_path = save_path
+                self._stats.total_epoch = epoch
```
ryanjulian: i think that you could make this significantly simpler by defining a `Stats` or `Metrics` class.

For instance, you could define some quantities as counters and only support incrementing them. You could also log the entire class `tabular.log(self._stats)`.

.pylintrc
naeioi: I will do so.

`no-self-use` is added mainly because `BaseAlgo.train()` calls `process_samples()` and some children algorithm may want to overwrite it. If `process_samples()` is moved out of `BaseAlgo` class, overwritting is not possible.

scripts/garage
naeioi: pylint seems to be buggy because of the name of this file (`garage`). But anyway I will replace `skip-file` with specific disables.

src/garage/experiment/local_runner.py
naeioi: They are immutable in terms of fields. The values are actually mutable. For example, `stats` gets updated every epoch. 

Originally I used dict for `setup_args` and `train_args`, but you suggested `SimpleNamespace` for better visibility. (in #633 , [comment](https://github.com/rlworkgroup/garage/pull/633#discussion_r277916964))

I think both `nametuple` and `SimpleNamespace` are better than dict because they prescribe fields. Where do you think I should place these types? My first idea is to put them out of `LocalRunner` but still keep them in this file.

src/garage/experiment/local_runner.py
naeioi: Good idea. So the idea is to change the signature of `obtain_samples`, right? I think I tried this but somehow gave up halfway... I will try it again.

src/garage/experiment/local_runner.py
naeioi: tabular will log these data to all the sinks (console, csv and tensorboard), but I only want to log them once at the start of the experiment to console. Otherwise, they will spoil the table fields.

src/garage/experiment/snapshotter.py
naeioi: It should not be there. I will remove it.

src/garage/experiment/snapshotter.py
naeioi: Should I put this exception in `snapshotter.py`, or in a seperate file?

src/garage/sampler/off_policy_vectorized_sampler.py
naeioi: This signature conflicts with the parent class. I will try fix it.

src/garage/experiment/local_runner.py
naeioi: Let's make it in another PR. This will involve the problem if existing stats should be included in this new class.

src/garage/experiment/local_runner.py
ryanjulian: Yes I agree. It hasn't worked out -- what do you think?

I think that namedtuple, dict, and SimpleNamespace don't fit the bill. Since they are becoming key types for the Runner, I am suggesting we make them (simple) real classes.

src/garage/experiment/local_runner.py
ryanjulian: it was just an idea. it's a fairly common for APIs which "collect a bunch of stuff" to return "that stuff and how much of it", though that's an old pattern from C.

@krzentner is working on the sampler API so I'm sure she'll want to weigh in on this.

src/garage/experiment/local_runner.py
ryanjulian: okay, i see. i didn't realize this was under `restore`

we should probably provide a way in dowel to only print a table once, but perhaps that's for another day.

do you log `total_env_steps` to tabular anywhere?

src/garage/experiment/snapshotter.py
ryanjulian: you usually put the exception for a file within that file.

src/garage/sampler/off_policy_vectorized_sampler.py
```diff
@@ -37,6 +36,9 @@ def __init__(self, algo, env, n_envs=None, no_reset=True):
         self.n_envs = n_envs
         self.no_reset = no_reset
 
+        self.vec_env = None
```
ryanjulian: if you are going to visit this class, can you make its members _private?

src/garage/sampler/on_policy_vectorized_sampler.py
```diff
@@ -15,22 +16,35 @@
...
+
+    """
 
     def __init__(self, algo, env, n_envs=None):
         if n_envs is None:
             n_envs = singleton_pool.n_parallel * 4
         super().__init__(algo, env)
         self.n_envs = n_envs
 
+        self.vec_env = None
```
ryanjulian: should these go in the base class?

src/garage/sampler/on_policy_vectorized_sampler.py
```diff
@@ -15,22 +16,35 @@
 
 
 class OnPolicyVectorizedSampler(BatchSampler):
```
ryanjulian: if you are going to visit this class, can you make its members _private?

src/garage/sampler/on_policy_vectorized_sampler.py
```diff
@@ -41,7 +55,22 @@ def shutdown_worker(self):
...
+                If whole_paths=False, paths will be truncated to contain
+                exactly batch_size steps.
+            whole_paths (bool): Return full paths if True. Otherwise truncate
+                paths to contain exactly batch_size steps.
+
+        Returns:
+            list[dict]: Collected paths.
+
+        """
         logger.log('Obtaining samples for iteration %d...' % itr)
```
ryanjulian: can you use .format() here while you're in the neighborhood?

src/garage/sampler/on_policy_vectorized_sampler.py
```diff
@@ -122,6 +150,6 @@ def obtain_samples(self, itr, batch_size=None, whole_paths=True):
 
         if whole_paths:
             return paths
-        else:
-            paths_truncated = truncate_paths(paths, batch_size)
-            return paths_truncated
+
+        paths_truncated = truncate_paths(paths, batch_size)
```
ryanjulian: why not `return truncate_paths(...)`?

src/garage/experiment/local_runner.py
naeioi: It's logged in algorithms, before call to `train_once()`.

src/garage/experiment/local_runner.py
naeioi: So far these args and stats are just key-value pairs, with a fixed set of fields. No validation, no transformation, so I think namedtuple is sufficient.

src/garage/experiment/local_runner.py
naeioi: I found that changing signature of `sampler.obtain_samples()` is quite intrusive. Because there are many samplers, and I need to use duplicate code to patch each of them.

Instead, I can put it in `runner.obtain_samples()`, which is basically a wrapper of `sampler.obtain_samples()` that meant to do these auxiliary tasks. This is much clearer to me.

src/garage/sampler/on_policy_vectorized_sampler.py
naeioi: I see there is some duplicate code in off- and on- policy samplers. Moving them to their base class, `BatchSampler` is a good idea. But `ISSampler` also inherits `BatchSampler`, and it does not have vec_env`. I don't feel like to make this change in this PR.

scripts/garage
```diff
@@ -8,23 +9,27 @@ from garage.tf.experiment import LocalTFRunner
...
+# pylint: disable=bad-docstring-quotes
+@click.option('--from_epoch',
+              default='last',
+              help='When there are multiple snapshots, '
+              'specify the index of epoch to restore from. '
+              'Can be "first", "last" or a number. '
+              'Not applicable when snapshot_mode="last"')
 def resume(from_dir, from_epoch, log_dir):
-    """Resume from command line."""
+    # pylint: disable=missing-param-doc, missing-type-doc
```
ryanjulian: why not just document them?

scripts/garage
```diff
@@ -8,23 +9,27 @@ from garage.tf.experiment import LocalTFRunner
...
-    default='last',
-    help='Index of epoch to restore from. Can be "first", "last" or a number. '
-    'Not applicable when snapshot_mode="last"')
 @click.option(
     '--log_dir',
     default=None,
-    help='Path to save the log snapshot. If not specified, will be the same '
+    help='Log path for resumed experiment. If not specified, will be the same '
     'as from_dir.')
+# pylint: disable=bad-docstring-quotes
```
ryanjulian: i'm confused why this even applies -- i don't even see a docstring here?

src/garage/experiment/local_runner.py
```diff
@@ -54,13 +55,30 @@ def __init__(self, snapshot_config, max_cpus=1):
...
-        self.train_args = None
+        self._train_args = None
+        self._stats = types.SimpleNamespace(total_itr=0,
+                                            total_env_steps=0,
+                                            total_epoch=0,
+                                            last_path=None)
+
+        self.algo = None
+        self.env = None
+        self.policy = None
```
ryanjulian: can you make these private? if they need to be part of the API, you can add properties for them.

src/garage/experiment/local_runner.py
ryanjulian: why not log it here? are you trying to avoid having Runner log to tabular?

src/garage/experiment/snapshotter.py
```diff
@@ -40,22 +40,46 @@ def __init__(self,
...
     def save_itr_params(self, itr, params):
-        """Save the parameters if at the right iteration."""
+        """Save the parameters if at the right iteration.
+
+        Args:
+            itr (int): Number of iterations. Used as the index of snapshot.
+            params (obj): Content of snapshot to be saved.
+
+        Raises:
+            ValueError: If snapshot_mode is not one of "all", "last" and "gap".
```
ryanjulian: or

src/garage/tf/algos/batch_polopt.py
```diff
@@ -264,10 +276,20 @@ def init_opt(self):
         """
         raise NotImplementedError
 
-    def get_itr_snapshot(self, itr):
-        """Get all the data that should be saved in this snapshot iteration."""
-        raise NotImplementedError
-
     def optimize_policy(self, itr, samples_data):
-        """Optimize the policy using the samples."""
+        # pylint: disable=no-self-use
```
ryanjulian: i think all your pylint problems here will go away i you just declare this method abstract (which it is, anyway).

src/garage/experiment/local_runner.py
ryanjulian: alright, please give namedtuple a try

scripts/garage
```diff
@@ -8,23 +9,27 @@ from garage.tf.experiment import LocalTFRunner
 
 @click.group()
 def cli():  # noqa: D103
+    # pylint: disable=missing-function-docstring
```
ryanjulian: why not just add a one-line docstring? It's as long as this pragma to disable it lol

src/garage/experiment/local_runner.py
naeioi: Oh, I totally missed that. Yes, they need to be mutable. Okay, I will make them into classes.

src/garage/experiment/local_runner.py
naeioi: Make sense. My original idea to follow the pattern that most logs are done in algorithms. I will try to do so. If I can't make it then let's leave it to future.

scripts/garage
naeioi: This docstring will be printed out to console when running `./garage`

scripts/garage
naeioi: pylint thinks `help='...'` is a docstring.

scripts/garage
naeioi: And most likely shorter!

scripts/garage
naeioi: Because this docstring will be printed out to console when running `./garage`

src/garage/experiment/local_runner.py
naeioi: I am blocked by off-policy algorithms when trying to move this log to the runner. 

Basically, an algorithm is in control of the log flow. For off-policy algorithms, they usually have some warm-up epochs, when they don't do any logging. This information is invisible to the runner. 

If `TotalEnvSteps` is logged in the runner, it will be the only thing logged at the first several epochs. And that will confuses both `CsvOuput` and `TensorboardOutput`.

naeioi: So I will just keep it logged in algorithms.

scripts/garage
ryanjulian: ...okay not sure what to say about that lol.

src/garage/experiment/local_runner.py
ryanjulian: okay thanks for trying. I think the right place to do this is in the sampler, to keep all algorithms honest.

Can you file an issue for that and assign it to KR?

naeioi: Will do.

.pre-commit-config.yaml
```diff
@@ -12,7 +12,7 @@ repos:
     -   id: check-pylint-normal
         name: Check pylint (except unit tests)
         types: [file, python]
-        exclude: ^tests/garage/.*$  # exclude unit tests
+        exclude: ^tests/garage/.|tests/fixtures/.*$  # exclude unit tests
```
ryanjulian: actually my intention here was that test fixtures should also be documented. please remove this part.

src/garage/experiment/local_runner.py
```diff
@@ -54,13 +118,29 @@ def __init__(self, snapshot_config, max_cpus=1):
...
+
+        self._algo = None
+        self._env = None
+        self._policy = None
+        self._sampler = None
+        self._plotter = None
+
+        self._start_time = None
+        self._itr_start_time = None
+        self.step_itr = None
```
ryanjulian: please make these _private

src/garage/experiment/local_runner.py
```diff
@@ -165,52 +252,60 @@ def restore(self, from_dir, from_epoch='last'):
                 Not applicable when snapshot_mode='last'.
 
         Returns:
-            A SimpleNamespace for train()'s arguments.
+            types.SimpleNamespace: A SimpleNamespace for train()'s arguments.
```
ryanjulian: i think you forgot to update this

src/garage/np/algos/batch_polopt.py
```diff
@@ -145,5 +146,15 @@ def process_samples(self, itr, paths):
         return samples_data
 
     def get_itr_snapshot(self, itr, samples_data):
-        """Return data saved in the snapshot for this iteration."""
+        # pylint: disable=unused-argument, no-self-use
```
ryanjulian: just
```
del itr
del samples_data
```

src/garage/experiment/local_runner.py
naeioi: This is a variable that gets updated in algorithms.

PR #1018, opened 2019-11-12
Add Pytorch TRPO

utkarshjp7:
Implemented Trust Region Policy Optimization in PyTorch.

Benchmarks are currently running and should be finished by tomorrow. I opened this PR to get some feedback since Initial results and tests looked good.

### DIFF
--- a/src/garage/torch/algos/trpo.py
+++ b/src/garage/torch/algos/trpo.py
@@ -1,11 +1,12 @@
-"""Proximal Policy Optimization (PPO)."""
+"""Trust Region Policy Optimization."""
 import torch
 
 from garage.torch.algos import VPG
+from garage.torch.optimizers import ConjugateGradientOptimizer
 
 
-class PPO(VPG):
-    """Proximal Policy Optimization (PPO).
+class TRPO(VPG):
+    """Trust Region Policy Optimization (TRPO).
 
     Args:
         env_spec (garage.envs.EnvSpec): Environment specification.
@@ -13,12 +14,11 @@ class PPO(VPG):
         baseline (garage.np.baselines.Baseline): The baseline.
         max_path_length (int): Maximum length of a single rollout.
         policy_lr (float): Learning rate for training policy network.
-        lr_clip_range (float): The limit on the likelihood ratio between
-            policies.
-        n_samples (int): Number of train_once calls per epoch.
+        num_train_per_epoch (int): Number of train_once calls per epoch.
         discount (float): Discount.
         gae_lambda (float): Lambda used for generalized advantage
             estimation.
+        max_kl (float): Maximum KL divergence between old and new policies.
         center_adv (bool): Whether to rescale the advantages
             so that they have mean 0 and standard deviation 1.
         positive_adv (bool): Whether to shift the advantages
@@ -46,12 +46,12 @@ class PPO(VPG):
                  env_spec,
                  policy,
                  baseline,
-                 max_path_length=500,
+                 max_path_length=100,
                  policy_lr=3e-4,
-                 lr_clip_range=2e-1,
-                 n_samples=1,
+                 num_train_per_epoch=1,
                  discount=0.99,
-                 gae_lambda=0.97,
+                 gae_lambda=0.98,
+                 max_kl=0.01,
                  center_adv=True,
                  positive_adv=False,
                  optimizer=None,
@@ -60,16 +60,20 @@ class PPO(VPG):
                  use_softplus_entropy=False,
                  stop_entropy_gradient=False,
                  entropy_method='no_entropy'):
+        if optimizer is None:
+            optimizer = ConjugateGradientOptimizer
+            optimizer_args = {'max_constraint_value': max_kl}
+
         super().__init__(env_spec, policy, baseline, max_path_length,
-                         policy_lr, n_samples, discount, gae_lambda,
+                         policy_lr, num_train_per_epoch, discount, gae_lambda,
                          center_adv, positive_adv, optimizer, optimizer_args,
                          policy_ent_coeff, use_softplus_entropy,
                          stop_entropy_gradient, entropy_method)
 
-        self._lr_clip_range = lr_clip_range
+        self._kl = None
 
     def _compute_objective(self, advantages, valids, obs, actions, rewards):
-        """Compute objective using surrogate value and clipped surrogate value.
+        """Compute the surrogate objective.
 
         Args:
             advantages (torch.Tensor): Expected rewards over the actions
@@ -82,22 +86,19 @@ class PPO(VPG):
             torch.Tensor: Calculated objective values
 
         """
-        # Compute constraint
         with torch.no_grad():
             old_ll = self._old_policy.log_likelihood(obs, actions)
-        new_ll = self.policy.log_likelihood(obs, actions)
 
+        new_ll = self.policy.log_likelihood(obs, actions)
         likelihood_ratio = (new_ll - old_ll).exp()
 
         # Calculate surrogate
         surrogate = likelihood_ratio * advantages
 
-        # Clipping the constraint
-        likelihood_ratio_clip = torch.clamp(likelihood_ratio,
-                                            min=1 - self._lr_clip_range,
-                                            max=1 + self._lr_clip_range)
+        return surrogate
 
-        # Calculate surrotate clip
-        surrogate_clip = likelihood_ratio_clip * advantages
-
-        return torch.min(surrogate, surrogate_clip)
+    def _optimize(self, itr, paths, valids, obs, actions, rewards):
+        self._optimizer.step(
+            f_loss=lambda: self._compute_loss(itr, paths, valids, obs, actions,
+                                              rewards),
+            f_constraint=lambda: self._compute_kl_constraint(obs))
--- a/src/garage/torch/algos/vpg.py
+++ b/src/garage/torch/algos/vpg.py
@@ -1,5 +1,6 @@
 """Vanilla Policy Gradient (REINFORCE)."""
 import collections
+import copy
 
 from dowel import tabular
 import numpy as np
@@ -9,6 +10,7 @@ import torch.nn.functional as F
 from garage.misc import tensor_utils
 from garage.np.algos import BatchPolopt
 from garage.torch.algos import loss_function_utils
+from garage.torch.utils import flatten_batch
 
 
 class VPG(BatchPolopt):
@@ -34,6 +36,7 @@ class VPG(BatchPolopt):
             standardized before shifting.
         optimizer (object): The optimizer of the algorithm. Should be the
             optimizers in torch.optim.
+        optimizer_args (dict): Arguments required to initialize the optimizer.
         policy_ent_coeff (float): The coefficient of the policy entropy.
             Setting it to zero would mean no entropy regularization.
         use_softplus_entropy (bool): Whether to estimate the softmax
@@ -61,6 +64,7 @@ class VPG(BatchPolopt):
             center_adv=True,
             positive_adv=False,
             optimizer=None,
+            optimizer_args=None,
             policy_ent_coeff=0.0,
             use_softplus_entropy=False,
             stop_entropy_gradient=False,
@@ -83,15 +87,19 @@ class VPG(BatchPolopt):
                                           stop_entropy_gradient,
                                           policy_ent_coeff)
         self._episode_reward_mean = collections.deque(maxlen=100)
-        self._optimizer = optimizer(policy.parameters(),
-                                    lr=policy_lr,
-                                    eps=1e-5)
+
+        if optimizer_args is None:
+            optimizer_args = {'lr': policy_lr, 'eps': 1e-5}
+
+        self._optimizer = optimizer(policy.parameters(), **optimizer_args)
 
         super().__init__(policy=policy,
                          baseline=baseline,
                          discount=discount,
                          max_path_length=max_path_length,
                          n_samples=n_samples)
+
+        self._old_policy = copy.deepcopy(self.policy)
 
     @staticmethod
     def _check_entropy_configuration(entropy_method, center_adv,
@@ -124,15 +132,27 @@ class VPG(BatchPolopt):
 
         """
         valids, obs, actions, rewards = self.process_samples(itr, paths)
-        average_return = self._log(itr, paths)
 
         loss = self._compute_loss(itr, paths, valids, obs, actions, rewards)
 
+        # Memorize the policy state_dict
+        self._old_policy.load_state_dict(self.policy.state_dict())
+
         self._optimizer.zero_grad()
-        # using a negative because optimizers use gradient descent,
-        # whilst we want gradient ascent.
-        (-loss).backward()
-        self._optimizer.step()
+        loss.backward()
+
+        kl_before = self._compute_kl_constraint(obs).detach()
+        self._optimize(itr, paths, valids, obs, actions, rewards)
+
+        with torch.no_grad():
+            loss_after = self._compute_loss(itr, paths, valids, obs, actions,
+                                            rewards)
+            kl = self._compute_kl_constraint(obs)
+            policy_entropy = self._compute_policy_entropy(obs)
+            average_return = self._log(itr, paths, loss.item(),
+                                       loss_after.item(), kl_before.item(),
+                                       kl.item(),
+                                       policy_entropy.mean().item())
 
         self.baseline.fit(paths)
         return average_return
@@ -188,7 +208,31 @@ class VPG(BatchPolopt):
             objective += self._policy_ent_coeff * policy_entropies
 
         valid_objectives = loss_function_utils.filter_valids(objective, valids)
-        return torch.cat(valid_objectives).mean()
+        return -torch.cat(valid_objectives).mean()
+
+    def _compute_kl_constraint(self, obs):
+        """Compute KL divergence.
+
+        Compute the KL divergence between the old policy distribution and
+        current policy distribution.
+
+        Args:
+            obs (torch.Tensor): Observation from the environment.
+
+        Returns:
+            torch.Tensor: Calculated mean KL divergence.
+
+        """
+        flat_obs = flatten_batch(obs)
+        with torch.no_grad():
+            old_dist = self._old_policy.forward(flat_obs)
+
+        new_dist = self.policy.forward(flat_obs)
+
+        kl_constraint = torch.distributions.kl.kl_divergence(
+            old_dist, new_dist)
+
+        return kl_constraint.mean()
 
     def _compute_policy_entropy(self, obs):
         """Compute entropy value of probability distribution.
@@ -247,6 +291,9 @@ class VPG(BatchPolopt):
             return torch.Tensor(self.baseline.predict_n(path))
         return torch.Tensor(self.baseline.predict(path))
 
+    def _optimize(self, itr, paths, valids, obs, actions, rewards):  # pylint: disable=unused-argument  # noqa: E501
+        self._optimizer.step()
+
     def process_samples(self, itr, paths):
         """Process sample data based on the collected paths.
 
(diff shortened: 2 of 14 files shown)


src/garage/torch/algos/trpo.py
```diff
@@ -0,0 +1,174 @@
...
+        surrogate = likelihood_ratio * advantages
+
+        return surrogate
+
+    def _optimize(self, itr, paths, valids, obs, actions, rewards):
+        closure = self._build_closure(itr, paths, valids, obs, actions,
+                                      rewards)
+        self._optimizer.step(closure)
+
+    def _build_closure(self, itr, paths, valids, obs, actions, rewards):
```
ryanjulian: is there a reason you did this instead of just using lambdas?

src/garage/torch/algos/vpg.py
```diff
@@ -188,7 +198,7 @@ def _compute_loss(self, itr, paths, valids, obs, actions, rewards):
             objective += self._policy_ent_coeff * policy_entropies
 
         valid_objectives = loss_function_utils.filter_valids(objective, valids)
-        return torch.cat(valid_objectives).mean()
+        return -1 * torch.cat(valid_objectives).mean()
```
ryanjulian: why not just `-torch.cat(valid_objectives).mean()`?

src/garage/torch/optimizers/conjugate_gradient_optimizer.py
```diff
@@ -0,0 +1,203 @@
...
+        self._backtrack_ratio = backtrack_ratio
+        self._accept_violation = accept_violation
+        self._max_constraint_value = max_constraint_value
+        self._hvp_reg_coeff = hvp_reg_coeff
+
+    def step(self, closure):
+        """Take an optimization step.
+
+        Args:
+            closure (tuple[function]): Functions to compute loss and
```
ryanjulian: why not separate arguments?

src/garage/torch/algos/trpo.py
```diff
@@ -0,0 +1,174 @@
...
+            the mean entropy to the surrogate objective. See
+            https://arxiv.org/abs/1805.00909 for more details.
+
+    """
+
+    def __init__(self,
+                 env_spec,
+                 policy,
+                 baseline,
+                 max_path_length=500,
```
ryanjulian: this is really high for most environments, which often are <100

src/garage/torch/algos/trpo.py
```diff
@@ -0,0 +1,174 @@
...
+                 policy_lr=3e-4,
+                 n_samples=1,
```
ryanjulian: n_samples? what does this do?

ryanjulian:
Code is looking pretty great, how are the benchmarks?

src/garage/experiment/deterministic.py
```diff
@@ -15,13 +15,16 @@ def set_seed(seed):
 
     """
     seed %= 4294967294
-    global seed_
+    global seed_  # pylint: disable=global-statement
```
ryanjulian: you can disable this function-wide here. the reason is fairly obvious.

src/garage/torch/optimizers/conjugate_gradient_optimizer.py
utkarshjp7: This is overriding `torch.optim.Optimizer` 's step function, which only takes one parameter.

ryanjulian: ah i see.

ryanjulian: it seems clear to me that torch intends `closure` to be a single function. If you want to pseudo-implement their API, I think it would be more analogous to change the amend the signature to accept a second closure (which computes the constraint).

ryanjulian: @krzentner

src/garage/torch/optimizers/conjugate_gradient_optimizer.py
```diff
@@ -0,0 +1,203 @@
...
+"""
+import numpy as np
+import torch
+from torch.optim import Optimizer
+
+from garage.misc.tensor_utils import unflatten_tensors
+from garage.torch.utils import update_tensor_list_from_flat_tensor
+
+
+def build_hessian_vector_product(func, params, reg_coeff=1e-5):
```
ryanjulian: does this need to be public?

src/garage/torch/optimizers/conjugate_gradient_optimizer.py
```diff
@@ -0,0 +1,203 @@
...
+    """Computes Hessian-vector product using Pearlmutter's algorithm.
+
+    `Pearlmutter, Barak A. "Fast exact multiplication by the Hessian." Neural
+    computation 6.1 (1994): 147-160.`
+
+    Args:
+        func (function): A function that returns a torch.Tensor. Hessian of
```
ryanjulian: Callable is the type for this.

utkarshjp7:
I am not sure what happened with HalfCheetah. Any thoughts?

/home/utkarsh/resl/garage/tests/benchmarks/garage/tf/algos/data/local/benchmarks/trpo/2019-11-11-16-09-33-414849/Hopper-v2_benchmark.png
/home/utkarsh/resl/garage/tests/benchmarks/garage/tf/algos/data/local/benchmarks/trpo/2019-11-11-16-09-33-414849/Reacher-v2_benchmark.png
/home/utkarsh/resl/garage/tests/benchmarks/garage/tf/algos/data/local/benchmarks/trpo/2019-11-11-16-09-33-414849/Swimmer-v2_benchmark.png
/home/utkarsh/resl/garage/tests/benchmarks/garage/tf/algos/data/local/benchmarks/trpo/2019-11-11-16-09-33-414849/Walker2d-v2_benchmark.png
/home/utkarsh/resl/garage/tests/benchmarks/garage/tf/algos/data/local/benchmarks/trpo/2019-11-11-16-09-33-414849/HalfCheetah-v2_benchmark.png

ryanjulian:
Hmm, I'm also not sure what happened with `HalfCheetah` -- but it's actually consistent with the rest of your results, which are slightly lower than TF in all cases. `HalfCheetah` is just harder than the rest, so the difference looks bigger.

It seems like the variance of the PyTorch plots is much lower. Did you ensure that the initial standard deviation of the policies are the same in both benchmarks? What about other hyperparameters?

There's a reason for every metric which the TF TRPO implementation plots, so make sure your version is plotting every one of those too. How do the Policy/MeanKL plots compare? Policy/Entropy? Baseline/ExplainedVariance? Check all these before resorting to more drastic debugging.

avnishn:
I'd go ahead and take a look at the plots for mean kl, mean mu, mean std. Those will probably be the most helpful in debugging the decreased performance on cheetah @utkarshjp7

src/garage/torch/algos/ppo.py
```diff
@@ -83,17 +80,13 @@ def _compute_objective(self, advantages, valids, obs, actions, rewards):
...
-        # pylint: disable=unused-argument
         # Compute constraint
         with torch.no_grad():
             old_ll = self._old_policy.log_likelihood(obs, actions)
         new_ll = self.policy.log_likelihood(obs, actions)
 
         likelihood_ratio = (new_ll - old_ll).exp()
 
-        # Memorize the policy state_dict
-        self._old_policy.load_state_dict(self.policy.state_dict())
```
avnishn: Why have you made changes to PPO, was there some issue?

utkarshjp7: `TRPO` and `PPO` both needs to use self._old_policy, so I moved the common code `vpg.py`

utkarshjp7:
@ryanjulian The hyperparameters are all same for both versions. I added logging to PyTorch TRPO and started benchmarks with 5 trials. I will post the results once its finished.

ryanjulian:
Great! Note that you can upload your results to https://tensorboard.dev for easy reviewing.

ryanjulian:
Please prioritize this PR, since it is blocking your MAML implementation.

utkarshjp7:
The latest benchmark data can be found here -> 
https://tensorboard.dev/experiment/OP0fqpjNQzawLEuLDslKjA

It is same as the plots I posted before, but with the additional logging. The main difference I observed is that lower bound of mean policy KL in tensorflow is 6.5e-3 while in PyTorch its 0.0.

ryanjulian:
In future runs I recommend you make the name of the PyTorch policy also "GaussianMLPPolicy" so that these plots overlap. They're hard to analyze apart.

Also, LinearFeatureBaseline/ExplainedVariance is a really essential debugging stat for on-policy RL so make sure to add that to the torch version. Perhaps we should resurrect the `log_diagnostics` interface or similar to make these things consistent.

@krzentner WDYT?

ryanjulian:
Some disorganized thoughts on these plots:

looking at `Reacher-v2/trial_1_seed_18/garage` vs `Reacher-v2/trial_1_seed_18/garage_pytorch`:
* LossAfter for PyTorch is *never* positive which is a bit strange. Loss magnitudes and signs don't have much meaning in policy gradients, but it's weird to see it clamped at 0. This suggests to me that either the optimizer or loss function is truncating, scaling, or normalizing something unexpectedly.
* MeanKL magnitudes are as-expected, but the PyTorch version has many steps with 0 MeanKL -- does this mean the line search is failing and it aborts the optimization for those steps? Do you produce some sort of warning when this happens?
* dLoss in PyTorch is normal but again has many itrs where it is 0. This adds more weight to my suspicion above that many optimization epochs are failing.
* All of the above seems to apply to HalfCheetah experiments

Here's some plots to chew on:
Screen Shot 2019-11-13 at 11 45 36 AM

Notice that KL, LossAfter, and dLoss are *all* 0 for itrs 486 and 487. This suggests to me that your optimizer is silently failing to create an update. Perhaps it can't calculate an update if the the loss is > 0?

src/garage/torch/algos/trpo.py
utkarshjp7: https://lintlyci.github.io/Flake8Rules/rules/E731.html

src/garage/experiment/deterministic.py
```diff
@@ -15,13 +15,17 @@ def set_seed(seed):
...
     global seed_
     seed_ = seed
     random.seed(seed)
     np.random.seed(seed)
     if 'tensorflow' in sys.modules:
-        import tensorflow as tf
+        import tensorflow as tf  # pylint: disable= import-outside-toplevel
         tf.compat.v1.set_random_seed(seed)
+    if 'torch' in sys.modules:
+        import torch  # pylint: disable= import-outside-toplevel
```
ryanjulian: the [torch docs](https://pytorch.org/docs/stable/notes/randomness.html) mention a couple more steps necessary to make reproducibility work.

they also warn that using them can have a performance impact on GPU, so if we are going to set these options we should probably make it non-default and also throw up a warning banner about it to the console.

src/garage/torch/algos/trpo.py
ryanjulian: please remove it if it's unused.

src/garage/torch/algos/trpo.py
ryanjulian: i'd prefer you just `# noqa: E731` them than add 50 lines of boilerplate. also, the rule forbids assigning them, not passing them as function arguments. seems to me you could:

```python
def _optimizer(...):
    self._optimizer.step(
        compute_loss=lambda: self._compute_loss(itr, paths, valids, obs, actions, rewards)
        compute_kl=lambda: self._compute_kl_constraint(obs)
    )
```

ryanjulian: up to you. i think either is a better option than what you had to do here.

src/garage/torch/algos/vpg.py
```diff
@@ -61,6 +64,7 @@ def __init__(
             center_adv=True,
             positive_adv=False,
             optimizer=None,
+            optimizer_args=None,
```
ryanjulian: i'm really not a fan of this pattern of passing dicts of args for constructors into other constructors. can we somehow construct-and-pass the optimizer instead? or just flatten these args into the parent constructor?

@krzentner your thoughts?

src/garage/torch/algos/vpg.py
```diff
@@ -308,5 +361,12 @@ def _log(self, itr, paths):
         tabular.record('StdReturn', np.std(undiscounted_returns))
         tabular.record('MaxReturn', np.max(undiscounted_returns))
         tabular.record('MinReturn', np.min(undiscounted_returns))
+        tabular.record('{0}/LossBefore'.format(self.policy.name), loss_before)
```
ryanjulian: you can use `tabular.prefix` for this: https://github.com/rlworkgroup/dowel/blob/master/src/dowel/tabular_input.py

you might also be interested in `tabular.record_misc_stat`

src/garage/torch/policies/base.py
```diff
@@ -3,37 +3,88 @@
 
 
 class Policy(abc.ABC):
-    """
-    Policy base class without Parameterzied.
+    """Policy base class without Parameterzied.
```
ryanjulian: i think "without Parameterized" is pretty outdated here.

src/garage/torch/policies/base.py
```diff
@@ -3,37 +3,88 @@
...
+
+        Args:
+            observation (torch.Tensor): Observation from the environment.
+
+        Returns:
+            tuple:
+                * torch.Tensor: Predicted action.
+                * dict:
+                    * list[float]: Mean of the distribution
+                    * list[float]: Standard deviation of logarithmic values of
```
ryanjulian: it is actually the log of the stddev, not the stddev of the log

src/garage/torch/policies/base.py
```diff
@@ -3,37 +3,88 @@
...
+
+        Args:
+            observations (torch.Tensor): Observations from the environment.
+
+        Returns:
+            tuple:
+                * torch.Tensor: Predicted actions.
+                * dict:
+                    * list[float]: Mean of the distribution
+                    * list[float]: Standard deviation of logarithmic values of
```
ryanjulian: log(std) not std(log)

src/garage/torch/policies/deterministic_mlp_policy.py
```diff
@@ -11,47 +10,86 @@
...
+
+        Args:
+            observation (torch.Tensor): Observation from the environment.
+
+        Returns:
+            tuple:
+                * torch.Tensor: Predicted action.
+                * dict:
+                    * list[float]: Mean of the distribution
+                    * list[float]: Standard deviation of logarithmic values of
```
ryanjulian: log(std)

src/garage/torch/policies/deterministic_mlp_policy.py
```diff
@@ -11,47 +10,86 @@
...
+
+        Args:
+            observations (torch.Tensor): Observations from the environment.
+
+        Returns:
+            tuple:
+                * torch.Tensor: Predicted actions.
+                * dict:
+                    * list[float]: Mean of the distribution
+                    * list[float]: Standard deviation of logarithmic values of
```
ryanjulian: log(std)

src/garage/torch/policies/gaussian_mlp_policy.py
```diff
@@ -13,15 +13,17 @@ class GaussianMLPPolicy(Policy, GaussianMLPModule):
...
         env_spec (garage.envs.env_spec.EnvSpec): Environment specification.
-        kwargs : Additional keyword arguments passed to the GaussianMLPModule.
+        kwargs (dict): Additional keyword arguments passed to the
+            GaussianMLPModule.
+        name (str): Name of policy.
 
     """
 
-    def __init__(self, env_spec, **kwargs):
+    def __init__(self, env_spec, name='GaussianMLPPolicy', **kwargs):
```
ryanjulian: please don't swallow `**kwargs` in the constructor. (I know you didn't write this, but let's remove it)

tests/garage/torch/optimizers/test_torch_conjugate_gradient_optimizer.py
```diff
@@ -0,0 +1,114 @@
...
+import pytest
+import torch
+
+from garage.torch.optimizers.conjugate_gradient_optimizer import (
+    build_hessian_vector_product, conjugate_gradient)
+
+# pylint: disable=not-callable  #https://github.com/pytorch/pytorch/issues/24807  # noqa: E501
+
+
+class TestConjugateGradientOptimizer:
```
ryanjulian: note that pytest is perfectly happy running test functions outside of classes, as long as they are named `test_*`

src/garage/experiment/deterministic.py
utkarshjp7: I will create a separate PR for this if this is not a high priority.

src/garage/torch/policies/gaussian_mlp_policy.py
utkarshjp7: If we remove it then how can one change the non-linearility of layer or initialization of weights?, or the point hear is that no one should be changing that through `GaussianMLPPolicy`?

utkarshjp7: I see what you mean. We should explicitly mention each keyword argument that this class supports?

src/garage/torch/algos/trpo.py
utkarshjp7: It means "Number of train_once calls per epoch.". It gets passed down to its parent, and all the way up to `garage.np.Batchpoplot`, and used inside `train` function. I agree that the name is confusing and should be changed to something like `num_train_per_epoch`.

Whether we should remove this completely or not is a question for a separate PR I think.

utkarshjp7:
Benchmark results look much better after fixing a bug in optimizer. These are averaged over 5 trials.

HalfCheetah-v2_benchmark
Hopper-v2_benchmark
Reacher-v2_benchmark
Swimmer-v2_benchmark
Walker2d-v2_benchmark

ryanjulian:
Looks like the torch version is 2x slower than the TF version?
Screen Shot 2019-11-14 at 8 32 31 AM

Any insight into why that is? Perhaps we haven't enabled CPU parallelism in torch? We don't need to fix it for this PR but it would be nice to know what's going on.

src/garage/torch/policies/gaussian_mlp_policy.py
ryanjulian: yes. always.

src/garage/experiment/deterministic.py
ryanjulian: lgtm

ryanjulian:
I'm still worried that PyTorch my be systematically just a *little bit* worse.

What do these look like with 10 trials? And what's the average performance gap for each at itr 999?

utkarshjp7:
Difference in means at itr 999 averaged over 10 trials. (tensorflow - pytorch)

HalfCheetah-v2: -17.11101
Hopper-v2: 13.40361
Reacher-v2: 0.71291
Swimmer-v2: 1.15873
Walker2d-v2: 1.80943

HalfCheetah-v2_benchmark
Hopper-v2_benchmark
Reacher-v2_benchmark
Swimmer-v2_benchmark
Walker2d-v2_benchmark

ryanjulian:
Okay, this is as close to equal as algos get. I think the performance is ready.

Did you have a chance to look at the runtime?

utkarshjp7:
I ran python profiler, and it seems the overhead in PyTorch is computing gradients (the `backward` function call). These stats are for running PyTorch TRPO for 10 epochs. 

```
   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      230    8.429    0.037    8.429    0.037 {method 'run_backward' of 'torch._C._EngineBase' objects}
     3208    3.015    0.001    3.015    0.001 {built-in method cholesky}
     6416    0.832    0.000    0.832    0.000 {built-in method tanh}
    52225    0.661    0.000    0.661    0.000 {method 'step' of 'mujoco_py.cymj.MjSim' objects}
     8808    0.500    0.000    0.500    0.000 {built-in method addmm}
      292    0.485    0.002    1.697    0.006 inspect.py:714(getmodule)
      452    0.453    0.001    0.453    0.001 {built-in method triangular_solve}
  198/179    0.389    0.002    0.462    0.003 {built-in method _imp.create_dynamic}
     3866    0.314    0.000    0.314    0.000 {built-in method marshal.loads}
```

ryanjulian:
I think that this is a general issue in PyTorch which we can investigate later. You can try some of the suggestions here: https://github.com/pytorch/pytorch/issues/975

Is your CPU usage near 100% (on all cores) during the optimization phase? Anyway, please don't block commit on this one since it looks like it's framework-wide.

PR #1019, opened 2019-11-13
SAC in PyTorch

avnishn:
_(I use the term loss here, but in my code, loss is referred to as an **objective**)_

The current sac has an issue with its policy loss. **I'll upload the graphs later today, to better explain this through visualization**, but at the moment, the policy loss shares a similar shape of what it should compared to SOTA implementations such as TF-agents, **but** the policy loss decreases at a much slower rate than the TF-agents implementation of sac. 

This slow decreasing policy loss causes our policy to learn slower over a million timesteps, and therefore our implementation is not as good as other state of the art implementations. 

This could be because of an incorrect policy(actor) loss, or an incorrect q-function (critic) loss. 

Currently I'm writing tests for checking if the policy loss is correct. I have verified that the temperature loss is computed correctly. but not the policy loss or q-function loss.


examples/torch/sac_half_cheetah.py
```diff
@@ -0,0 +1,74 @@
...
+    """Set up environment and algorithm and run the task."""
+    runner = LocalRunner(snapshot_config)
+    env = GarageEnv(normalize(gym.make('HalfCheetah-v2')))
+
+    policy = TanhGaussianMLPPolicy(env_spec=env.spec,
+                               hidden_sizes=[256, 256],
+                               hidden_nonlinearity=nn.ReLU,
+                               output_nonlinearity=None)
+
+    qf1 = ContinuousMLPQFunction(env_spec=env.spec,
```
ryanjulian: if `qf1`/`qf2`, by definition in the algorithm, need to be the same network, then i suggest they be copied by the algo.

src/garage/torch/algos/sac.py
```diff
@@ -0,0 +1,264 @@
...
+                self.log_alpha = torch.tensor([self.initial_log_entropy], dtype=torch.float, requires_grad=True)
+                self.alpha_optimizer = optimizer([self.log_alpha], lr=self.policy_lr)
+        else:
+            self.alpha = alpha
+
+        self.episode_rewards = []
+        self.success_history = []
+
+    # 0) update policy using updated min q function
+    # 1) compute targets from Q functions
```
ryanjulian: these would be more helpful in-line

src/garage/torch/policies/gaussian_mlp_policy.py
```diff
@@ -36,6 +36,8 @@ def forward(self, inputs):
     def get_action(self, observation):
         """Get a single action given an observation."""
         with torch.no_grad():
+            if(type(observation) != torch.Tensor):
```
ryanjulian: the type of the observation is either torch.tensor or it's np.ndarray, it's not both.

examples/torch/sac_half_cheetah.py
```diff
@@ -0,0 +1,74 @@
...
+                                hidden_sizes=[256, 256],
+                                hidden_nonlinearity=F.relu)
+
+    qf2 = ContinuousMLPQFunction(env_spec=env.spec,
+                                hidden_sizes=[256, 256],
+                                hidden_nonlinearity=F.relu)
+
+    target_qf1 = ContinuousMLPQFunction(env_spec=env.spec,
```
ryanjulian: same story -- if these must be the same model class then you should `copy.deepcopy()` them internally.

examples/torch/sim_policy.py
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env python3
```
ryanjulian: why is this different from the non-torch sim_policy?

src/garage/sampler/ray_sampler.py
```diff
@@ -241,4 +241,4 @@ def rollout(self):
             np.array(actions),\
             np.array(rewards),\
             dict(agent_infos),\
-            dict(env_infos)
+            dict(env_infos)
```
ryanjulian: did you touch this file on accident?

src/garage/torch/algos/sac.py
```diff
@@ -0,0 +1,258 @@
+"""This modules creates a sac model in PyTorch."""
```
ryanjulian: Can you please make this more descriptive? I don't think it's a model.

You should probably include a citation for the paper you based this on, and the codebase if necessary.

src/garage/torch/algos/sac.py
```diff
@@ -0,0 +1,258 @@
...
+import copy
+
+from dowel import logger, tabular
+import numpy as np
+import torch
+import torch.nn.functional as F
+
+from garage.np.algos.off_policy_rl_algorithm import OffPolicyRLAlgorithm
```
ryanjulian: why not `from garage.np.algos import OffPolicyRLAlgorithm`?

src/garage/torch/algos/sac.py
```diff
@@ -0,0 +1,258 @@
...
+                 min_buffer_size=int(1e4),
+                 target_update_tau=5e-3,
+                 policy_lr=3e-4,
+                 qf_lr=3e-4,
+                 optimizer=torch.optim.Adam,
+                 clip_pos_returns=False,
+                 clip_return=np.inf,
+                 max_action=None,
+                 smooth_return=True,
+                 input_include_goal=False):
```
ryanjulian: i think we are trying to remove this -- is it required of OffPolicyRLAlgorithm?

src/garage/torch/algos/sac.py
```diff
@@ -0,0 +1,258 @@
...
+
+        self.policy = policy
+        self.qf1 = qf1
+        self.qf2 = qf2
+        self.replay_buffer = replay_buffer
```
ryanjulian: please make all members _private

src/garage/torch/algos/sac.py
```diff
@@ -0,0 +1,258 @@
...
+                         max_path_length=max_path_length,
+                         buffer_batch_size=buffer_batch_size,
+                         min_buffer_size=min_buffer_size,
+                         replay_buffer=replay_buffer,
+                         use_target=True,
+                         discount=discount,
+                         smooth_return=smooth_return)
+
+        # use 2 target q networks
+        self.target_qf1 = copy.deepcopy(self.qf1)
```
ryanjulian: i thought you accept these as arguments???

src/garage/torch/algos/sac.py
```diff
@@ -0,0 +1,258 @@
...
+        self.target_qf2 = copy.deepcopy(self.qf2)
+        # self.target_qf1 = target_qf1
+        # self.target_qf2 = target_qf2
+        self.policy_optimizer = optimizer(self.policy.parameters(),
+                                          lr=self.policy_lr)
+        self.qf1_optimizer = optimizer(self.qf1.parameters(), lr=self.qf_lr)
+        self.qf2_optimizer = optimizer(self.qf2.parameters(), lr=self.qf_lr)
+        # automatic entropy coefficient tuning
+        self.use_automatic_entropy_tuning = use_automatic_entropy_tuning
+        if self.use_automatic_entropy_tuning and not alpha:
```
ryanjulian: please document this very carefully in the class docstring. if there are mutually-incompatible sets of arguments, please validate them and raise a ValueError if you get such a set.

src/garage/torch/algos/sac.py
```diff
@@ -0,0 +1,258 @@
...
+                self.target_entropy = target_entropy
+            else:
+                self.target_entropy = -np.prod(
+                    self.env_spec.action_space.shape).item()
+            self.log_alpha = torch.tensor([self.initial_log_entropy], dtype=torch.float, requires_grad=True)
+            self.alpha_optimizer = optimizer([self.log_alpha], lr=self.policy_lr)
+        else:
+            self.alpha = alpha
+
+        self.episode_rewards = []
```
ryanjulian: again, please make all members _private

src/garage/torch/algos/sac.py
```diff
@@ -0,0 +1,258 @@
...
+
+    # 0) update policy using updated min q function
+    # 1) compute targets from Q functions
+    # 2) update Q functions using optimizer
+    # 3) query Q functons, take min of those functions
+    def train_once(self, itr, paths):
+        """
+        """
+        paths = self.process_samples(itr, paths)
+        self.episode_rewards.extend([
```
ryanjulian: i think logging these stats is more appropriate in process_samples?

src/garage/torch/algos/sac.py
```diff
@@ -0,0 +1,258 @@
...
+        obs = samples["observation"]
+        actions = samples["action"]
+        rewards = samples["reward"]
+        terminals = samples["terminal"]
+        next_obs = samples["next_observation"]
+
+        action_dists = self.policy(torch.Tensor(obs))
+        if ((itr % (self.gradient_steps - 1) == 0) and itr != 0):
+            entropy = action_dists.entropy()
+            tabular.record("policy/mu_entropy", torch.mean(entropy).detach().numpy().item())
```
ryanjulian: The tabular paths here should use prefixes that correspond with model or algorithm names (whichever is most specific).

src/garage/torch/algos/sac.py
```diff
@@ -0,0 +1,258 @@
...
+            alpha = self.log_alpha.detach().exp().item()
+        else:
+            alpha = self.alpha
+        min_q_new_actions = torch.min(self.qf1(torch.Tensor(obs), torch.Tensor(new_actions)), 
+                            self.qf2(torch.Tensor(obs), torch.Tensor(new_actions)))
+        policy_objective = ((alpha * log_pi) - min_q_new_actions.flatten()).mean()
+        tabular.record("policy_loss", policy_objective.item())
+
+        '''
+        QF Loss
```
ryanjulian: these block comments are a sure sign that this function should be split up into private subroutines.

src/garage/torch/algos/sac.py
```diff
@@ -0,0 +1,258 @@
...
+        """
+
+        obs = samples["observation"]
+        actions = samples["action"]
+        rewards = samples["reward"]
+        terminals = samples["terminal"]
+        next_obs = samples["next_observation"]
+
+        action_dists = self.policy(torch.Tensor(obs))
+        if ((itr % (self.gradient_steps - 1) == 0) and itr != 0):
```
ryanjulian: you should probably put this block in its own subroutine

src/garage/torch/algos/sac.py
```diff
@@ -0,0 +1,258 @@
...
+
+        self.qf2_optimizer.zero_grad()
+        qf2_loss.backward()
+        self.qf2_optimizer.step()
+
+        self.policy_optimizer.zero_grad()
+        policy_objective.backward()
+        self.policy_optimizer.step()
+
+    def temperature_objective(self, log_pi):
```
ryanjulian: seems like most of these should be _private

src/garage/torch/algos/sac.py
```diff
@@ -0,0 +1,258 @@
...
+        """
+        implemented inside optimize_policy
+        """
+        alpha_loss = 0
+        if self.use_automatic_entropy_tuning:
+            alpha_loss = (-(self.log_alpha)* (log_pi.detach() + self.target_entropy)).mean()
+        return alpha_loss
+
+    def actor_objective(self, obs, log_pi, new_actions):
```
ryanjulian: _private

src/garage/torch/algos/sac.py
```diff
@@ -0,0 +1,258 @@
...
+        alpha = self.log_alpha.detach().exp()
+        min_q_new_actions = torch.min(self.qf1(torch.Tensor(obs), torch.Tensor(new_actions)), 
+                            self.qf2(torch.Tensor(obs), torch.Tensor(new_actions)))
+        policy_objective = ((alpha * log_pi) - min_q_new_actions.flatten()).mean()
+        return policy_objective
+
+    def critic_objective(self,samples):
```
ryanjulian: _private

src/garage/torch/algos/sac.py
```diff
@@ -0,0 +1,258 @@
...
+        '''
+        QF Loss
+        '''
+        obs = samples["observation"]
```
ryanjulian: didn't you already do this unpacking somewhere else? why not do it once and then call these functions using the unpacked values?

src/garage/torch/distributions/tanh_normal_distribution.py
```diff
@@ -0,0 +1,84 @@
...
+                                    "must also define a custom .expand() method.".
+                                    format(self.__class__.__name__, cls.__name__))
+    return self.__new__(type(self)) if _instance is None else _instance
+
+
+class TanhNormal(TransformedDistribution):
+    """ A Multivariate Gaussian Distribution with tanh output.
+
+    Args: 
+        loc (Tensor): mean of the distribution
```
ryanjulian: please use the full package path for types, in this case `torch.Tensor`

src/garage/torch/distributions/tanh_normal_distribution.py
```diff
@@ -0,0 +1,84 @@
...
+from torch.distributions import MultivariateNormal
+from torch.distributions import Transform
+from torch.distributions.transformed_distribution import TransformedDistribution
+import torch.nn.functional as F
+
+class TanhTransform(Transform):
+    """
+    Tanh transformation for torch distributions.
+    """
+    domain = constraints.real
```
ryanjulian: unless they are part of the interface, please make these _private

src/garage/torch/modules/__init__.py
```diff
@@ -1,12 +1,14 @@
 """Pytorch modules."""
 
-from garage.torch.modules.gaussian_mlp_module import \
-    GaussianMLPIndependentStdModule, GaussianMLPModule, \
-    GaussianMLPTwoHeadedModule
+from garage.torch.modules.gaussian_mlp_module import (GaussianMLPBaseModule,
```
ryanjulian: just add a separate line for each

src/garage/torch/modules/__init__.py
```diff
@@ -1,12 +1,14 @@
...
+    GaussianMLPIndependentStdModule, GaussianMLPModule, 
+    GaussianMLPTwoHeadedModule)
 from garage.torch.modules.mlp_module import MLPModule
 from garage.torch.modules.multi_headed_mlp_module import MultiHeadedMLPModule
+from garage.torch.modules.tanh_gaussian_mlp_module import TanhGaussianMLPTwoHeadedModule
```
ryanjulian: didn't yong refactor this into a MultiHeaded variant (instead of TwoHeaded)?

src/garage/torch/modules/tanh_gaussian_mlp_module.py
```diff
@@ -0,0 +1,208 @@
...
+from torch import nn
+
+from garage.torch.distributions import TanhNormal
+from garage.torch.modules.mlp_module import MLPModule
+from garage.torch.modules.multi_headed_mlp_module import MultiHeadedMLPModule
+
+
+class TanhGaussianMLPBaseModule(nn.Module):
+    """
+    GaussianMLPModel.
```
ryanjulian: Please make these comments more descriptive. The name of the class is not a sufficient summary line.

src/garage/torch/modules/tanh_gaussian_mlp_module.py
```diff
@@ -0,0 +1,208 @@
+"""GaussianMLPModule."""
```
ryanjulian: Please make this summary more descriptive.

src/garage/torch/modules/tanh_gaussian_mlp_module.py
```diff
@@ -0,0 +1,208 @@
...
+        self._hidden_nonlinearity = hidden_nonlinearity
+        self._hidden_w_init = hidden_w_init
+        self._hidden_b_init = hidden_b_init
+        self._output_nonlinearity = output_nonlinearity
+        self._output_w_init = output_w_init
+        self._output_b_init = output_b_init
+        self._layer_normalization = layer_normalization
+
+        if self._std_parameterization not in ('exp', 'softplus'):
+            raise NotImplementedError
```
ryanjulian: please raise a ValueError with a useful error message.

src/garage/torch/modules/tanh_gaussian_mlp_module.py
ryanjulian: Is there no way to implement this as a subclass or wrapper on GaussianMLPModule(s)?

src/garage/torch/policies/tanh_gaussian_mlp_policy.py
```diff
@@ -0,0 +1,72 @@
...
+class TanhGaussianMLPPolicy(Policy, TanhGaussianMLPTwoHeadedModule):
+    """
+    GaussianMLPPolicy.
+
+    A policy that contains a MLP to make prediction based on a gaussian
+    distribution.
+
+    Args:
+        env_spec (garage.envs.env_spec.EnvSpec): Environment specification.
+        module : GaussianMLPModule to make prediction based on a gaussian
```
ryanjulian: this requires a type

tests/garage/torch/algos/test_sac.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def action(self, unused_observation):
+        action = torch.Tensor([self._action], dtype=torch.float32)
+        return action
+
+    def parameters(self):
+        return torch.zeros(5)
+
+
+class DummyCriticNet():
```
ryanjulian: no need for trailing ()

tests/garage/torch/algos/test_sac.py
```diff
@@ -0,0 +1,146 @@
...
+    def __call__(self, observation, actions):
+        # Biggest state is best state.
+        value = torch.max(observation, dim=-1).values
+        # Biggest action is best action.
+        q_value = torch.max(actions, axis=-1).values
+        ret = value + q_value
+        return ret
+
+
+class TestSacLosses():
```
ryanjulian: seems like there's a lot of repetitive tests here -- maybe you can to define a `setup_method` instead?

note that pytest is perfectly happy to test ordinary functions rather than unittest-style classes. see: https://docs.pytest.org/en/latest/getting-started.html and https://docs.pytest.org/en/latest/fixture.html#fixtures

also: there's no need for a trailing ()

tests/garage/torch/algos/test_sac.py
```diff
@@ -0,0 +1,146 @@
...
+                  qf1=DummyCriticNet(),
+                  qf2=DummyCriticNet(),
+                  target_qf1=DummyCriticNet(),
+                  target_qf2=DummyCriticNet(),
+                  use_automatic_entropy_tuning=True,
+                  replay_buffer=None,
+                  discount=0.9,
+                  buffer_batch_size=2,
+                  target_entropy=3.0,
+                  #initial_log_entropy=4.0,
```
ryanjulian: plz remove comments

tests/garage/torch/algos/test_sac.py
```diff
@@ -0,0 +1,146 @@
...
+        rewards = torch.FloatTensor([10, 20])
+        terminals = [[0.], [0.]]
+        next_observations = torch.FloatTensor([[5, 6], [7, 8]])
+        samples = {"observation" : observations, "action" : actions, "reward" : rewards,
+                    "terminal" : terminals, "next_observation" : next_observations}
+        td_targets = [7.3, 19.1]
+        pred_td_targets = [7., 10.]
+
+        # Expected critic loss has factor of 2, for the two TD3 critics.
+        expected_loss = 2 * F.mse_loss(
```
ryanjulian: can you use an actual number here? what happens if torch breaks the MSE API?

ryanjulian:
Don't forget to update `README.md` with the new algorithm!

src/garage/torch/policies/gaussian_mlp_policy.py
```diff
@@ -55,6 +55,8 @@ def get_action(self, observation):
 
         """
         with torch.no_grad():
```
utkarshjp7: I think `torch.no_grad()` is not necessary here.

src/garage/torch/modules/tanh_gaussian_mlp_module.py
```diff
@@ -0,0 +1,208 @@
...
+    def _get_mean_and_log_std(self, inputs):
+        pass
+
+    def forward(self, inputs):
+        """Forward method."""
+        mean, log_std_uncentered = self._get_mean_and_log_std(inputs)
+
+        if self._min_std_param or self._max_std_param:
+            log_std_uncentered = log_std_uncentered.clamp(
+                min=self._to_scalar_if_not_none(self._min_std_param),
```
utkarshjp7: Is `None` ever expected here? Do you get error if the value is `None` ? If `None` is not expected then you should throw `ValueError` when its `None`.

examples/torch/sac_half_cheetah.py
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env python3
+"""
+This is an example to train a task with DDPG algorithm written in PyTorch.
```
lywong92: SAC*

examples/torch/sac_pendulum.py
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env python3
+"""
+This is an example to train a task with DDPG algorithm written in PyTorch.
```
lywong92: SAC*

examples/torch/sac_half_cheetah.py
```diff
@@ -0,0 +1,74 @@
...
+
+Here it creates a gym environment InvertedDoublePendulum. And uses a DDPG with
```
lywong92: HalfCheetah*

examples/torch/sac_point_mass.py
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env python3
+"""
+This is an example to train a task with DDPG algorithm written in PyTorch.
```
lywong92: SAC*

examples/torch/sac_point_mass.py
```diff
@@ -0,0 +1,65 @@
...
+
+Here it creates a gym environment InvertedDoublePendulum. And uses a DDPG with
```
lywong92: HalfCheetah*

src/garage/torch/algos/sac.py
```diff
@@ -0,0 +1,258 @@
...
+    def optimize_policy(self, itr, samples):
+        """ Optimize the policy based on the policy objective from the sac paper.
+
+        Args:
+            itr (int) - current training iteration
+            samples() - samples recovered from the replay buffer
+        Returns:
+            None
+        """
+
```
lywong92: You can use np_to_torch from garage.torch.utils on samples here to convert them to torch tensors, so you have have to keep wrapping everything in torch.Tensors later on.

src/garage/torch/policies/tanh_gaussian_mlp_policy.py
```diff
@@ -0,0 +1,72 @@
+"""TanhGaussianMLPPolicy."""
```
lywong92: Can you add tests for this?

src/garage/torch/modules/tanh_gaussian_mlp_module.py
lywong92: Can you add tests for this module?

src/garage/torch/policies/tanh_gaussian_mlp_policy.py
```diff
@@ -0,0 +1,72 @@
...
+    :return:
+    """
+
+    def __init__(self, env_spec, **kwargs):
+        self._obs_dim = env_spec.observation_space.flat_dim
+        self._action_dim = env_spec.action_space.flat_dim
+
+        Policy.__init__(self, env_spec)
+        TanhGaussianMLPTwoHeadedModule.__init__(self,
+                                   input_dim=self._obs_dim,
```
lywong92: Can we make input_dim (maybe and output_dim) configurable parameters? PEARL uses tanh_gaussian_mlp_policy and its input_dim needs to be obs_dim + some number denoting latent dimension. I think this also adds more flexibility in general for input and output dimensions.

examples/torch/sac_half_cheetah.py
krzentner: I'm not sure if that actually works though, since these networks need to have different initializations.

Also, AFAICT, there's technically no requirement that these be the same architecture?

examples/torch/sac_half_cheetah.py
krzentner: I'm not convinced they need to be? Also, they definitely need to be initialized differently, which is tricky to do without assuming an initialization strategy in the algorithm or extending the `torch.nn.Module` API.

src/garage/sampler/off_policy_vectorized_sampler.py
```diff
@@ -59,7 +59,7 @@ def shutdown_worker(self):
         """Terminate workers if necessary."""
         self.vec_env.close()
 
-    def obtain_samples(self, itr, batch_size):
+    def obtain_samples(self, itr, batch_size, add_transitions_to_buffer=True):
```
krzentner: So if I set this to False, the transitions don't get added to the replay buffer? Why is that useful?

src/garage/torch/algos/sac.py
krzentner: No, there are four total Q functions in SAC. There's the two Q functions, each of which has a copy with low-pass filtered parameters called the "target" Q function.

src/garage/torch/algos/sac.py
```diff
@@ -0,0 +1,258 @@
...
+        self.qf2_optimizer = optimizer(self.qf2.parameters(), lr=self.qf_lr)
+        # automatic entropy coefficient tuning
+        self.use_automatic_entropy_tuning = use_automatic_entropy_tuning
+        if self.use_automatic_entropy_tuning and not alpha:
+            if target_entropy:
+                self.target_entropy = target_entropy
+            else:
+                self.target_entropy = -np.prod(
+                    self.env_spec.action_space.shape).item()
+            self.log_alpha = torch.tensor([self.initial_log_entropy], dtype=torch.float, requires_grad=True)
```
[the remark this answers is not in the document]
krzentner: Probably in case someone sets `initial_log_entropy` to `0` (an int). It's a good idea to do this, since otherwise the user will get a cryptic error in the middle of SAC. Maybe add a comment why though?

src/garage/torch/algos/sac.py
```diff
@@ -0,0 +1,322 @@
...
+        """
+        last_return = None
+
+        for epoch in runner.step_epochs():
+            if self.replay_buffer.n_transitions_stored < self.min_buffer_size:
+                batch_size = self.min_buffer_size
+            else:
+                batch_size = None
+            for cycle in range(self.n_epoch_cycles):
+                # eval_steps = runner.obtain_samples(-1, 30, add_transitions_to_buffer=False)
```
krzentner: Definitely remove commented code before committing it. If you need it later, put it in another branch.

src/garage/torch/algos/sac.py
```diff
@@ -0,0 +1,322 @@
...
+        elif ((gradient_step % (self.gradient_steps - 1) == 0) and gradient_step != 0):
+            tabular.record("qf_loss/{}".format("qf1_loss"), float(qf1_loss))
+            tabular.record("qf_loss/{}".format("qf2_loss"), float(qf2_loss))
+        self.qf1_optimizer.zero_grad()
+        qf1_loss.backward()
+        # for name, param in self.qf1.named_parameters():
+        #     if param.requires_grad and (param.grad is not None):
+        #         tabular.record("qf1_grads/{}".format(name), float(param.grad.norm()))
+        #         tabular.record("qf2_grads_max/{}".format(name), float(param.grad.max()))
+        #         tabular.record("qf2_grads_min/{}".format(name), float(param.grad.min()))
```
krzentner: By the way, it's often helpful to actually look at the maximum absolute gradient (i.e. `param.grad.abs().max()`), since the actual sign of the gradient is kinda arbitrary.

Also, you should put this commented code in another branch.

krzentner:
@avnishn Are you still planning to use this pull request, or should I only pay attention to the "Torch SAC" one?

avnishn:
We can close this PR. Thanks KR

PR #1023, opened 2019-11-15
Modify BatchPolopt to return flattened sample paths

ahtsan:
This is the first PR for a series of 3 PRs, aiming to replace garage distribution with tfp.Distribution.

Currently we are generating dense samples ([batch, time_step, *dims]) with a "valids" array in the base class BatchPolopt. The valids array indicates whether an index is valid in a path, i.e. valids[i][j] is 1 if the j-th step in i-path is valid, otherwise 0.
When we construct loss in most of other on-policy algorithms, we feed the dense samples directly for recurrent policies, or have to perform reshaping and filtering for non-recurrent policies.

By changing to feed flattened sample paths and valid, we instead feed the flattened samples directly for non-recurrent policies, or reshape to dense from the flattened samples with valids.

This PR modifies BatchPolpot to return flattened sample paths, [B * (T), *dims], where T is the variable-length of each batch.

### DIFF
--- a/src/garage/sampler/on_policy_vectorized_sampler.py
+++ b/src/garage/sampler/on_policy_vectorized_sampler.py
@@ -24,6 +24,19 @@ class OnPolicyVectorizedSampler(BatchSampler):
         n_envs (int): Number of environment instances to setup.
             This parameter has effect on sampling performance.
 
+    In obtain_samples(), N paths are generated. Each path is a dictionary,
+    with keys and values as following:
+    - observations : numpy.ndarray with shape [Batch, *obs_dims]
+    - actions      : numpy.ndarray with shape [Batch, *act_dims]
+    - rewards      : numpy.ndarray with shape [Batch, ]
+    - env_infos    : A dictionary with each key representing one environment
+        info, value being a numpy.ndarray with shape [Batch, ?]. One example
+        is "ale.lives" for atari environments.
+    - agent_infos  : A dictionary with each key representing one agent info,
+        value being a numpy.ndarray with shape [Batch, ?]. One example is
+        "prev_action", which is used for recurrent policy as previous action
+        input, merged with the observation input as the state input.
+
     """
 
     def __init__(self, algo, env, n_envs=None):
--- a/src/garage/sampler/on_policy_vectorized_sampler.py
+++ b/src/garage/sampler/on_policy_vectorized_sampler.py
@@ -76,11 +76,13 @@ class OnPolicyVectorizedSampler(BatchSampler):
                 * rewards: numpy.ndarray with shape [Batch, ]
                 * env_infos: A dictionary with each key representing one
                   environment info, value being a numpy.ndarray with shape
-                  [Batch, ?]. One example is "ale.lives" for atari environments.
-                * agent_infos: A dictionary with each key representing one agent
-                  info, value being a numpy.ndarray with shape [Batch, ?]. One example
-                  is "prev_action", which is used for recurrent policy as previous
-                  action input, merged with the observation input as the state input.
+                  [Batch, ?]. One example is "ale.lives" for atari
+                  environments.
+                * agent_infos: A dictionary with each key representing one
+                  agent info, value being a numpy.ndarray with shape
+                  [Batch, ?]. One example is "prev_action", which is used
+                  for recurrent policy as previous action input, merged with
+                  the observation input as the state input.
 
         """
         logger.log('Obtaining samples for iteration %d...' % itr)
--- a/src/garage/sampler/on_policy_vectorized_sampler.py
+++ b/src/garage/sampler/on_policy_vectorized_sampler.py
@@ -24,19 +24,6 @@ class OnPolicyVectorizedSampler(BatchSampler):
         n_envs (int): Number of environment instances to setup.
             This parameter has effect on sampling performance.
 
-    In obtain_samples(), N paths are generated. Each path is a dictionary,
-    with keys and values as following:
-    - observations : numpy.ndarray with shape [Batch, *obs_dims]
-    - actions      : numpy.ndarray with shape [Batch, *act_dims]
-    - rewards      : numpy.ndarray with shape [Batch, ]
-    - env_infos    : A dictionary with each key representing one environment
-        info, value being a numpy.ndarray with shape [Batch, ?]. One example
-        is "ale.lives" for atari environments.
-    - agent_infos  : A dictionary with each key representing one agent info,
-        value being a numpy.ndarray with shape [Batch, ?]. One example is
-        "prev_action", which is used for recurrent policy as previous action
-        input, merged with the observation input as the state input.
-
     """
 
     def __init__(self, algo, env, n_envs=None):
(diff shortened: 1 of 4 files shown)


src/garage/tf/algos/batch_polopt2.py
```diff
@@ -0,0 +1,281 @@
...
+                 baseline,
+                 scope=None,
+                 max_path_length=500,
+                 discount=0.99,
+                 gae_lambda=1,
+                 center_adv=True,
+                 positive_adv=False,
+                 fixed_horizon=False,
+                 flatten_input=True):
+        self.env_spec = env_spec
```
ryanjulian: Please change these to _private

src/garage/tf/algos/batch_polopt2.py
```diff
@@ -0,0 +1,281 @@
...
+    def process_samples(self, itr, paths):
+        """Return processed sample data based on the collected paths.
+
+        Args:
+            itr (int): Iteration number.
+            paths (list[dict]): A list of collected paths.
+
+        Returns:
+            dict: Processed sample data, with key
+                * observations  : (numpy.ndarray), shape [B * (T), *obs_dims]
```
ryanjulian: I appreciate the extra effort on formatting, but unfortunately I think the documentation generator doesn't used fixed-width, so these won't align when they're rendered.

I think you should stick to the style used for args.

If you want to present something as fixed width (e.g. the shapes), you can use `backticks` just like in markdown

src/garage/tf/algos/batch_polopt2.py
```diff
@@ -0,0 +1,281 @@
...
+        for key in self.policy.state_info_keys:
+            agent_infos[key] = np.concatenate(
+                [infos[key] for infos in agent_infos_path])
+
+        env_infos = [path['env_infos'] for path in paths]
+        env_infos = tensor_utils.stack_tensor_dict_list([
+            tensor_utils.pad_tensor_dict(p, max_path_length) for p in env_infos
+        ])
+
+        valids = [np.ones_like(path['returns']) for path in paths]
```
ryanjulian: Shouldn't this be a numpy array? This is a list

ryanjulian: Same for lengths

src/garage/tf/algos/batch_polopt2.py
```diff
@@ -0,0 +1,281 @@
...
+        tabular.record('Extras/EpisodeRewardMean',
+                       np.mean(self.episode_reward_mean))
+        tabular.record('NumTrajs', len(paths))
+        tabular.record('StdReturn', np.std(undiscounted_returns))
+        tabular.record('MaxReturn', np.max(undiscounted_returns))
+        tabular.record('MinReturn', np.min(undiscounted_returns))
+
+        return samples_data
+
+    def init_opt(self):
```
ryanjulian: Can you make the following 3 methods abstract? And also this class an ABC?

src/garage/tf/algos/batch_polopt2.py
```diff
@@ -0,0 +1,281 @@
...
+            ]
+
+        if hasattr(self.baseline, 'predict_n'):
+            all_path_baselines = self.baseline.predict_n(paths)
+        else:
+            all_path_baselines = [
+                self.baseline.predict(path) for path in paths
+            ]
+
+        for idx, path in enumerate(paths):
```
ryanjulian: Is there a reason this and the for loop below can't be merged?

src/garage/tf/algos/batch_polopt2.py
ahtsan: valids is not returned so I guess it’s fine and yes, lengths should be a numpy array

src/garage/tf/algos/batch_polopt2.py
ahtsan: They can be merged, will do

src/garage/tf/algos/batch_polopt2.py
ahtsan: Will do

src/garage/tf/algos/batch_polopt2.py
ahtsan: Constructor argument `fixed_horizon` is not being used. Is it supposed to allow us to return either dense or flattened samples?

src/garage/tf/algos/batch_polopt2.py
ahtsan: It inherits `RLAlgorithm` which is already an ABC

src/garage/tf/algos/batch_polopt2.py
ahtsan: How about if I put it in the docstring body instead of under returns?

src/garage/tf/algos/batch_polopt2.py
ryanjulian: an ABC can inherit from an ABC

src/garage/tf/algos/batch_polopt2.py
ryanjulian: sure. that doesn't change what i said about spacing, though.

src/garage/sampler/on_policy_vectorized_sampler.py
```diff
@@ -23,6 +23,19 @@ class OnPolicyVectorizedSampler(BatchSampler):
         env (gym.Env): A gym/akro env object
         n_envs (int): Number of parallel environments used for sampling.
 
+    In obtain_samples(), N paths are generated. Each path is a dictionary,
+    with keys and values as following:
+    - observations : numpy.ndarray with shape [Batch, *obs_dims]
```
ryanjulian: again, these will render poorly on readthedocs.io. please just don't exploit fixed-width in docstrings.

src/garage/tf/algos/batch_polopt2.py
ryanjulian: are you sure we shouldn't return valids? why not, if we have it?

ahtsan: We will be passing `lengths`, which serves the same purpose as `valids`. But sure, we can also return `valids` as another format of `lengths`. For example, we have `valids` to be 
[1, 1, 1, 1, 0, 0, 0, 0, 0, 0] and we will have `lengths` to be [4].

ryanjulian: i think it might be convenient in the future for some users, and converting from `lengths` is a pain. Can you make sure it's a `numpy.ndarray`?

We can reconsider this when you and KR looks at the policy+sampler APIs. For now I think it's best to include it.

ahtsan: Sounds good

src/garage/tf/algos/batch_polopt2.py
```diff
@@ -0,0 +1,316 @@
...
+                  the j-th sample in i-th path is valid, otherwise 0.
+                - agent_infos: (dict), see
+                  OnPolicyVectorizedSampler.obtain_samples()
+                - env_infos: (dict), see
+                  OnPolicyVectorizedSampler.obtain_samples()
+                - paths: (list[dict]) The original path with observation or
+                  action flattened
+                - average_return: (numpy.float64)
+
+            where B = batch size, (T) = variable-length of each trajectory,
```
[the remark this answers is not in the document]
ahtsan: Right, the idea of `(T)` actually comes from ragged tensor
https://github.com/tensorflow/tensorflow/blob/r2.0/tensorflow/python/ops/ragged/ragged_tensor.py#L85
without the context it's really confusing. Thanks for pointing that out.

PR #1029, opened 2019-11-17
Fix pickling in GarageEnv

ahtsan:
This fixes the nightly test failure
https://travis-ci.com/rlworkgroup/garage/jobs/257220015

### DIFF
--- a/src/garage/envs/base.py
+++ b/src/garage/envs/base.py
@@ -165,6 +165,15 @@ class GarageEnv(gym.Wrapper):
             return state
         return self.__dict__
 
+    def __setstate__(self, state):
+        """See `Object.__setstate__.
+
+        Args:
+            state (dict): Unpickled state of this object.
+
+        """
+        self.__init__(state['_env'], state['_env_name'])
+
 
 def Step(observation, reward, done, **kwargs):  # noqa: N802
     """Create a namedtuple from the results of environment.step(action).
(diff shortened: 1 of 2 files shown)


src/garage/envs/base.py
```diff
@@ -165,6 +165,16 @@ def __getstate__(self):
...
 
+    def __setstate__(self, state):
+        """See `Object.__setstate__.
+
+        Args:
+            state (dict): Unpickled state of this object.
+
+        """
+        obj = type(self)(state['_env'], state['_env_name'])
+        self.__dict__.update(obj.__dict__)
```
utkarshjp7: can't you just do `self.__init__(state['_env'], state['_env_name'])`? I think that is easier to read.

ryanjulian:
I would like @krzentner to weigh in on this, since she is much more familiar with the Python object lifecycle.

ryanjulian:
Can you add a unit test which would have caught this?

src/garage/envs/base.py
krzentner: Yeah, if `__init__` works, it's better to just call that, in case the environment constructs a circular reference (e.g. `gym.CarRacing`).

ryanjulian:

src/garage/envs/base.py
ahtsan: Fixed

ahtsan:

PR #1030, opened 2019-11-17
Remove optimizer_args from torch algos

utkarshjp7:
This PR removes `optimizer_args` pattern from torch algorithm constructors. ~All torch algorithms now accept a constructed optimizer instead of an optimizer class.~

All torch algorithms now accept a optimizer type or a tuple of optimizer type and optimizer args. The algorithm internally constructs an optimizer using a new utility function `_make_optimizer`.

### DIFF
--- a/examples/torch/ddpg_pendulum.py
+++ b/examples/torch/ddpg_pendulum.py
@@ -1,6 +1,5 @@
 #!/usr/bin/env python3
-"""
-This is an example to train a task with DDPG algorithm written in PyTorch.
+"""This is an example to train a task with DDPG algorithm written in PyTorch.
 
 Here it creates a gym environment InvertedDoublePendulum. And uses a DDPG with
 1M steps.
@@ -21,7 +20,15 @@ from garage.torch.q_functions import ContinuousMLPQFunction
 
 
 def run_task(snapshot_config, *_):
-    """Set up environment and algorithm and run the task."""
+    """Set up environment and algorithm and run the task.
+
+    Args:
+        snapshot_config (garage.experiment.SnapshotConfig): The snapshot
+            configuration used by LocalRunner to create the snapshotter.
+            If None, it will create one with default settings.
+        _ : Unused parameters
+
+    """
     runner = LocalRunner(snapshot_config)
     env = GarageEnv(normalize(gym.make('InvertedDoublePendulum-v2')))
 
@@ -40,6 +47,8 @@ def run_task(snapshot_config, *_):
                                        size_in_transitions=int(1e6),
                                        time_horizon=100)
 
+    policy_optimizer = (torch.optim.Adagrad, {'lr': 1e-4, 'lr_decay': 0.99})
+
     ddpg = DDPG(env_spec=env.spec,
                 policy=policy,
                 qf=qf,
@@ -48,10 +57,9 @@ def run_task(snapshot_config, *_):
                 min_buffer_size=int(1e4),
                 exploration_strategy=action_noise,
                 target_update_tau=1e-2,
-                policy_lr=1e-4,
-                qf_lr=1e-3,
                 discount=0.9,
-                optimizer=torch.optim.Adam)
+                policy_optimizer=policy_optimizer,
+                qf_optimizer=torch.optim.Adam)
 
     runner.setup(algo=ddpg, env=env)
 
--- a/src/garage/torch/algos/__init__.py
+++ b/src/garage/torch/algos/__init__.py
@@ -1,4 +1,9 @@
 """PyTorch algorithms."""
+from garage.torch.algos._utils import _Default  # noqa: F401
+from garage.torch.algos._utils import compute_advantages  # noqa: F401
+from garage.torch.algos._utils import filter_valids  # noqa: F401
+from garage.torch.algos._utils import make_optimizer  # noqa: F401
+from garage.torch.algos._utils import pad_to_last  # noqa: F401
 from garage.torch.algos.ddpg import DDPG
 from garage.torch.algos.vpg import VPG
 from garage.torch.algos.ppo import PPO  # noqa: I100
--- a/src/garage/torch/algos/ddpg.py
+++ b/src/garage/torch/algos/ddpg.py
@@ -7,6 +7,7 @@ import numpy as np
 import torch
 
 from garage.np.algos.off_policy_rl_algorithm import OffPolicyRLAlgorithm
+from garage.torch.algos import _Default, make_optimizer
 from garage.torch.utils import np_to_torch, torch_to_np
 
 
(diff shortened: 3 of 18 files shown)


src/garage/torch/algos/ddpg.py
```diff
@@ -116,12 +110,25 @@ def __init__(self,
 
         self.target_policy = copy.deepcopy(self.policy)
         self.target_qf = copy.deepcopy(self.qf)
-        self.policy_optimizer = optimizer(self.policy.parameters(),
-                                          lr=self.policy_lr)
-        self.qf_optimizer = optimizer(self.qf.parameters(), lr=self.qf_lr)
+        if policy_optimizer is None:
+            policy_optimizer = torch.optim.Adam(self.policy.parameters(),
```
ryanjulian: Can you add a launcher which actually uses a different optimizer? How do they know they should create the optimizer with self.policy.parameters()?

I think it might be better to pass the type in the constructor but have a better way of passing the optimizer args.

@krzentner WDYT?

src/garage/torch/algos/ddpg.py
```diff
@@ -136,11 +143,12 @@ def train_once(self, itr, paths):
...
         last_average_return = np.mean(self.episode_rewards)
-        for train_itr in range(self.n_train_steps):
-            if self.replay_buffer.n_transitions_stored >= self.min_buffer_size:  # noqa: E501
+        for _ in range(self.n_train_steps):
+            if (self.replay_buffer.n_transitions_stored >=
+                    self.min_buffer_size):
                 self.evaluate = True
                 samples = self.replay_buffer.sample(self.buffer_batch_size)
-                qf_loss, y, q, policy_loss = torch_to_np(
+                qf_loss, y, q, policy_loss = torch_to_np(  # pylint: disable=unbalanced-tuple-unpacking  # noqa: E501
```
ryanjulian: i think if you change the implementation of `torch_to_np`/`np_to_torch` to tuple comprehensions (rather than the naive for/append loops) then pylint might be able to infer these lengths. 

if that doesn't fix it, let me know and i'll disable the rule.

examples/torch/ddpg_pendulum.py
```diff
@@ -21,7 +20,15 @@
 
 
 def run_task(snapshot_config, *_):
-    """Set up environment and algorithm and run the task."""
+    """Set up environment and algorithm and run the task.
```
ryanjulian: can you provide an example which uses a different optimizer? That will actually allow us to compare the API before and after.

utkarshjp7: This launcher now uses different optimizer. You can also check out the benchmark files of `vpg`, `trpo` and `ppo` to find examples of new API.

src/garage/torch/algos/ddpg.py
utkarshjp7: I added example of new API in ddpg launcher. 

People can know the `optimizer` type is `torch.optim.Optimizer` from algo documentation, and then they can refer to the torch documentation for further details.

src/garage/torch/algos/ddpg.py
utkarshjp7: changing to tuple comprehensions fixed the issue

ryanjulian: awesome!

krzentner:
Ryan and I discussed this for some time. We eventually came to the following argument:
  - In most of garage, we expose the important arguments in easy-to-reach places (e.g. `hidden_layers` arguments to policies).
  - We should also do that in algorithms for consistency, and because we feel this way of providing ease-of-use is best for garage's users.
  - For most users, the most important parameter is the learning rate. In particular, the learning rate is a more important parameter than the algorithm itself.
  - Therefore the learning rate parameters (e.g. `lr`, `max_kl`, `policy_lr`, etc.) should be exposed to the users as arguments to the algorithm, so that it can be specified without specifying the optimizer itself.
  - There should still be a way of passing a custom optimizer for algorithms where doing so makes sense (i.e. gradient methods that aren't TRPO).

After thinking about what that specifically looks like, we think that the following design would be best:
 - For algorithms where the optimizer cannot reasonably be changed from the default (TRPO), all hyper-parameters of the optimizer should be exposed as arguments on the algorithm. It should not be possible to specify the optimizer as an argument. In other words, we should remove the `optimizer` argument from TRPO, and keep / add the `max_kl` argument.
 - For algorithms where it is reasonable to set the optimizer to most torch optimizers, there should be an `optimizer` argument which specifies the type of the optimizer. There should also be a parameter to set that optimizer's important hyperparameters (e.g. `lr`, `policy_lr`, `q_lr`, etc.).
 - For algorithms where the optimizer can be specified, it should also be possible to pass a tuple of (optimizer type, optimizer hyper parameters) as the `optimizer`. If a  tuple is passed, the second parameter must be a dictionary, and the other arguments specifying hyper-parameters for that algorithm (e.g. `policy_lr`) should not be passed. If any of them are passed, the algorithm should raise `ValueError`.

By the way, this probably also involves writing a utility function, which we think should probably be in something like this:
```python
# In src/garage/torch/algos/trpo.py:

class TRPO:
    def __init__(self, ..., max_kl=0.2):
        ...
    ...

# In src/garage/torch/algos/_utils.py

class _Default:
    pass

def _make_optimizer(optimizer_type, module, **kwargs):
    if isinstance(optimizer_type, tuple):
        opt_type, opt_args = optimizer_type
        for name, arg in kwargs.items():
            if not isinstance(arg, Default):
                raise ValueError('Should not specify {} and explicit optimizer args'.format(name))
        return opt_type(module.parameters(), **opt_args)
    else:
        return policy_optimizer(module.parameters(), **kwargs)

# In src/garage/torch/algos/__init__.py
from garage.torch.algos._utils import _Default, _make_optimizer
...
__all__ = ... # Should not include _Default, etc.

# In src/garage/torch/algos/ddpg.py
from garage.torch.algos import _Default, _make_optimizer

class DDPG:
    def __init__(self, policy_optimizer=torch.optim.Adam, policy_lr=_Default(1e-3), ...):
        ...
        self.policy_opt = _make_optimizer(policy_optimizer, self.policy, lr=policy_lr)
        ...
    ...
```

ryanjulian:
@utkarshjp7 when you do this, can you also file an issue to make this consistent in the TF and NP trees?

src/garage/torch/algos/trpo.py
```diff
@@ -46,27 +42,24 @@ def __init__(self,
                  env_spec,
                  policy,
                  baseline,
+                 optimizer=None,
                  max_path_length=100,
-                 policy_lr=3e-4,
                  num_train_per_epoch=1,
                  discount=0.99,
                  gae_lambda=0.98,
-                 max_kl=0.01,
```
ryanjulian: can you make this `max_kl_step` to match the tf branch? and make sure the rest of the hyperparam names match the TF branch?

if you feel strongly, you can change them in tf instead.

src/garage/torch/algos/__init__.py
```diff
@@ -1,4 +1,7 @@
 """PyTorch algorithms."""
+from garage.torch.algos._utils import (  # noqa: F401
```
ryanjulian: might as well just put them on their own lines each.

it seems to me that compute_advantages, filter_valids, and pad_to_last can call me public...

ryanjulian: actually @krzentner should be expose `make_optimizer` as well?

src/garage/torch/algos/ddpg.py
```diff
@@ -69,21 +75,20 @@ def __init__(self,
...
+                 policy_lr=_Default(1e-4),
+                 qf_lr=_Default(1e-3),
                  clip_pos_returns=False,
                  clip_return=np.inf,
                  max_action=None,
                  reward_scale=1.,
                  smooth_return=True):
         action_bound = env_spec.action_space.high
         self.tau = target_update_tau
-        self.policy_lr = policy_lr
```
ryanjulian: while you are here, please make all these members _private

src/garage/torch/algos/__init__.py
utkarshjp7: All these function are only used by `garage.torch.algo` module, and were written for a specific use case. I think we should keep them private until we have a use for them elsewhere.

src/garage/torch/algos/ddpg.py
utkarshjp7: Done.

src/garage/torch/algos/trpo.py
utkarshjp7: I renamed the argument in torch branch.

src/garage/torch/algos/__init__.py
```diff
@@ -1,4 +1,9 @@
 """PyTorch algorithms."""
+from garage.torch.algos._utils import _compute_advantages  # noqa: F401
+from garage.torch.algos._utils import _Default  # noqa: F401
+from garage.torch.algos._utils import _filter_valids  # noqa: F401
```
ryanjulian: i think most of these can be public utilities

src/garage/torch/algos/ddpg.py
```diff
@@ -22,37 +23,42 @@ class DDPG(OffPolicyRLAlgorithm):
...
-        qf_lr (float): Learning rate for training Q-value network.
         discount(float): Discount factor for the cumulative return.
         policy_weight_decay (float): L2 weight decay factor for parameters
             of the policy network.
         qf_weight_decay (float): L2 weight decay factor for parameters
             of the q value network.
-        optimizer (torch.optimizer): Optimizer for training policy network
-            and Q-vaule network.
-        criterion (torch.nn.modules.loss): Loss function for Q-value newtork.
+        policy_optimizer (Union[type, tuple[type, dict]]): Type of optimizer
```
ryanjulian: put code samples in
```
`backticks`
```

src/garage/torch/algos/ppo.py
```diff
@@ -11,11 +11,15 @@ class PPO(VPG):
         env_spec (garage.envs.EnvSpec): Environment specification.
         policy (garage.torch.policies.base.Policy): Policy.
         baseline (garage.np.baselines.Baseline): The baseline.
+        optimizer (Union[type, tuple[type, dict]]): Type of optimizer.
+            This can be an optimizer type such as 'torch.optim.Adam' or a
```
ryanjulian: code samples in backticks

PR #1031, opened 2019-11-18
Fix torch packaging and use package in the CI

ryanjulian:
Issue #1024 was not detected by the CI, because we were accidentally
testing repository code instead of the package. This change makes all
parts of the CI build, then install, the same version of the package
shipped to PyPI.

* Also adds __init__.py to fix the packaging.


krzentner:
Yeah, the fix part of this definitely needs to be backported.

ryanjulian:

ryanjulian:

PR #1044, opened 2019-11-19
Make docs deps match package deps

ryanjulian:


docs/requirements.txt
```diff
@@ -13,8 +12,8 @@ pyprind
 python-dateutil
 scikit-image
 scipy
-tensorflow<1.13,>=1.12.0
-tensorflow-probability<0.6.0,>=0.5.0
+# tensorflow<1.16,>=1.15.0
```
gitanshu: Why was this commented? And I guess not providing this would mean we need to add it to mock imports in sphinx?

ryanjulian: I noticed that installing TF makes rtfd.io run out of memory, but now many of the TF and torch modules aren't documented in the auto-generated docs. Not sure what to do about this one.

krzentner: Surely we can't be the only one with this problem?

krzentner: I've opened an issue to address this in the future, so let's merge this change.

PR #1052, opened 2019-11-19
Refactor local_runner train method

yonghyuc:
Refactor local_runner to make the runner simpler and easy to understand

### DIFF
--- a/src/garage/experiment/local_runner.py
+++ b/src/garage/experiment/local_runner.py
@@ -349,7 +349,19 @@ class LocalRunner:
 
         self._plot = plot
 
-        return self._algo.train(self)
+        # For some algorithms like CMA_ES or CEM
+        self._algo.prepare_train()
+
+        last_return = None
+        for _ in self.step_epochs():
+            for _ in range(self._algo.n_samples):
+                self.step_path = self.obtain_samples(self.step_itr)
+                tabular.record('TotalEnvSteps', self.total_env_steps)
+                last_return = self._algo.train_once(self.step_itr,
+                                                    self.step_path)
+                self.step_itr += 1
+
+        return last_return
 
     def step_epochs(self):
         """Step through each epoch.


yonghyuc:
I think we can move `train` method from algorithms to runner.
By doing this, we can make the overall structure simpler

prev : runner -> algo -> runner (sampling)
this: runner -> sampler, algo

What do you guys think?

naeioi:
We have discussed this in https://github.com/rlworkgroup/garage/issues/704. We should also ask @zequnyu to comment on this. The current structure is the result of a tradeoff between more control on learning loop in algorithms, and separation of concerns to other components from algorithms.

krzentner:
I think that it's actually much better the way things are now (with the `Algorithm` owning the loop), for the following reasons:
 - Soon, the runner won't be handling sampling, so it won't be runner -> algorithm -> runner anyways, and this change would result in more complex data-flow / control flow.
 - This makes the algorithm API bigger, instead of smaller.
 - Meta-learning algorithms will almost certainly want a more complicated loop structure than this.
 - This prevents us from simplifying CMA-ES and CEM (see #736).

Is the main goal here to hide the `step_path` and `step_itr` runner fields? I do think it would be good if we changed the API to remove those fields while keeping the same functionality.

In particular, it's kinda messy that we assume there's one set of paths to save each epoch, so we should probably have a `maybe_save_paths()` method on runner, instead of using `step_path`. We should also be able to handle `step_itr` automatically in `step_epochs()`.

yonghyuc:
> I think that it's actually much better the way things are now (with the `Algorithm` owning the loop), for the following reasons:
> 
>     * Soon, the runner won't be handling sampling, so it won't be runner -> algorithm -> runner anyways, and this change would result in more complex data-flow / control flow.
> 
>     * This makes the algorithm API bigger, instead of smaller.
> 
>     * Meta-learning algorithms will almost certainly want a more complicated loop structure than this.
> 
>     * This prevents us from simplifying CMA-ES and CEM (see #736).
> 
> 
> Is the main goal here to hide the `step_path` and `step_itr` runner fields? I do think it would be good if we changed the API to remove those fields while keeping the same functionality.
> 
> In particular, it's kinda messy that we assume there's one set of paths to save each epoch, so we should probably have a `maybe_save_paths()` method on runner, instead of using `step_path`. We should also be able to handle `step_itr` automatically in `step_epochs()`.

I thought this change could make algorithms simpler (maybe only for batch_polopt?)
But since, you said that runner won't be handling sampling and Meta-learning algorithms need other way to be trained, I think this issue and PR is not valid. I would close this PR and issue.

zequnyu:
I think @krzentner and @naeioi have provided sufficient reasons for not coming back to our original design, which is pretty much the same as you proposed here.

> In particular, it's kinda messy that we assume there's one set of paths to save each epoch, so we should probably have a `maybe_save_paths()` method on runner, instead of using `step_path`. We should also be able to handle `step_itr` automatically in `step_epochs()`.

We could do a quick refactoring to make `step_itr` automatic in runner (by just moving it to `step_epochs()`). But is there any better solution than introducing `maybe_save_paths()`?

@krzentner do you plan to move sampling to the algorithms?

krzentner:
I do have a plan to "move sampling to algorithms" (by making algorithms use the sampler interface directly). I'm not sure if there is a better solution than `maybe_save_paths()`. We (probably) want to keep the `save_paths` parameter in the runner, but that means that we have to somehow tell the runner what the paths are. Alternatively, if we have a proper `TrajectoryBatch` type, we could somehow tweak how it's pickled, but I don't know if that would really be cleaner.

PR #1053, opened 2019-11-20
Improvements for PyTorch TRPO

utkarshjp7:
This PR adds two minor improvements to PyTorch TRPO.

1. To compute the hessian-vector product, we need to compute the gradients of constraint function only once each optimization `step`, hence the gradient calculations can be moved from inside `eval` to `build_hessian_vector_product` function. This gives 15% of runtime improvement. Check the benchmark below.
2. Very very rarely, if the gradients of loss function are too large, then `step_dir` becomes `nan`. When this happens in `tf` version, we discard the current optimization step and just set the parameters back to previous value. We do the same thing in PyTorch, however we get an error before we reach to the code that sets the parameters back to previous value. The error is caused when we create `MultivariateNormal` object (to compute the loss) and the co-variance matrix contains `nan`. I fixed this by replacing `nan` with 0. in `step_dir`.

Before this PR, PyTorch runtime is around 10m
before

After this PR, PyTorch tuntime is around 8.5m
Screenshot 2019-11-19 18:51:02

This is benchmark after the Fix averaged over 10 trials to prove the performance of algo doesn't have any change.
HalfCheetah-v2_benchmark

### DIFF
--- a/src/garage/torch/optimizers/conjugate_gradient_optimizer.py
+++ b/src/garage/torch/optimizers/conjugate_gradient_optimizer.py
@@ -30,6 +30,8 @@ def _build_hessian_vector_product(func, params, reg_coeff=1e-5):
 
     """
     param_shapes = [p.shape or torch.Size([1]) for p in params]
+    f = func()
+    f_grads = torch.autograd.grad(f, params, create_graph=True)
 
     def _eval(vector):
         """The evaluation function.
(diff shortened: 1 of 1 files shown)


src/garage/torch/optimizers/conjugate_gradient_optimizer.py
```diff
@@ -160,6 +161,9 @@ def step(self, f_loss, f_constraint):  # pylint: disable=arguments-differ
         # Compute step direction
         step_dir = _conjugate_gradient(f_Ax, flat_loss_grads, self._cg_iters)
 
+        # Replace nan with 0.
+        step_dir[step_dir.ne(step_dir)] = 0.
```
nish21: I think you could use `torch.isnan` here which is slightly more readable.

Issue #1016, opened 2019-11-09
Remove `get_itr_snapshot` from algorithms


naeioi:
`get_itr_snapshot()` in algorithms is a legacy from rllab to do snapshot. We now enforce an algorithm to be pickable, so this interface should be deprecated and should be removed from the codebase.

naeioi:
Closed by #1054

PR #1054, opened 2019-11-20
Remove get_itr_snapshot()

naeioi:
This PR addresses #1016 . It removes all `get_itr_snapshot()` from algorithms. This interface was a legacy from rllab to do snapshotting and is deprecated because we now require an algorithm to be pickable/unpickable by rewriting `__setstate__()` and `__getstate__()`.

Some outdated files in fixtures are not touched. They are used by the sigint test and have diverged from the other parts of the codebase for long. I think we should make an investigation to see if zombie processes are still a problem. If not then just remove this test.

### DIFF
--- a/src/garage/tf/algos/dqn.py
+++ b/src/garage/tf/algos/dqn.py
@@ -1,3 +1,4 @@
+"""Deep Q-Learning Network algorithm."""
 from dowel import tabular
 import numpy as np
 import tensorflow as tf
@@ -21,8 +22,9 @@ class DQN(OffPolicyRLAlgorithm):
         policy (garage.tf.policies.base.Policy): Policy.
         qf (object): The q value network.
         replay_buffer (garage.replay_buffer.ReplayBuffer): Replay buffer.
-        exploration_strategy (garage.np.exploration_strategies.
-            ExplorationStrategy): Exploration strategy.
+        exploration_strategy
+            (garage.np.exploration_strategies.ExplorationStrategy):
+            Exploration strategy.
         n_epoch_cycles (int): Epoch cycles.
         min_buffer_size (int): The minimum buffer size for replay buffer.
         buffer_batch_size (int): Batch size for replay buffer.
@@ -96,8 +98,7 @@ class DQN(OffPolicyRLAlgorithm):
                                   smooth_return=smooth_return)
 
     def init_opt(self):
-        """
-        Initialize the networks and Ops.
+        """Initialize the networks and Ops.
 
         Assume discrete space for dqn, so action dimension
         will always be action_space.n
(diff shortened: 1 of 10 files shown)


src/garage/tf/algos/dqn.py
```diff
@@ -209,12 +219,20 @@ def train_once(self, itr, paths):
...
+
+        Args:
+            itr (int): Iterations.
+            samples_data (list): Processed batch data.
+
+        Returns:
+            numpy.float64: Loss of policy.
+
+        """
+        del itr
```
zequnyu: This seems strange to me. If this is not needed inside the function, shouldn't it not be passed, or at least marked as `_`?

naeioi: This is to fight with the recent change on stricter pylint checks. Pylint detects and rejects unused variables, and this `del` line will suppress this check.
Since `train_once()` is a virtial method, remove `itr` will require removing it from the parent class, so is not an option. 
If marked with `_`, pylint will complain about the name does not match with that in the parent class.

src/garage/tf/algos/npo.py
```diff
@@ -155,10 +161,19 @@ def init_opt(self):
...
     def optimize_policy(self, itr, samples_data):
-        """Optimize policy."""
+        """Optimize policy.
+
+        Args:
+            itr (int): Iteration number.
+            samples_data (dict): Processed sample data.
+                See process_samples() for details.
+
+        Raises:
```
ryanjulian: i don't see the raise statement anywhere?

src/garage/tf/algos/reps.py
```diff
@@ -15,7 +15,8 @@
 from garage.tf.optimizers import LbfgsOptimizer
 
 
-class REPS(BatchPolopt):
+# pylint: disable=differing-param-doc, differing-type-doc
```
ryanjulian: why?

src/garage/tf/algos/reps.py
```diff
@@ -15,7 +15,8 @@
...
+class REPS(BatchPolopt):  # noqa: D416
```
ryanjulian: what is D416?

src/garage/tf/algos/reps.py
```diff
@@ -68,13 +71,23 @@ def __init__(self,
...
-                 dual_optimizer_args=dict(maxiter=50),
+                 dual_optimizer_args=None,
                  name='REPS'):
+        optimizer_args = optimizer_args or dict(max_opt_itr=50)
+        dual_optimizer_args = dual_optimizer_args or dict(maxiter=50)
+
         self.name = name
         self._name_scope = tf.name_scope(self.name)
 
+        self.feat_diff = None
```
ryanjulian: can you make all of these attrs _private while you are visiting?

tests/fixtures/algos/dummy_tf_algo.py
```diff
@@ -1,17 +1,28 @@
...
     def optimize_policy(self, itr, samples_data):
-        pass
+        """Optimize the policy using the samples.
+
+        Args:
+            itr (int): Iteration number.
+            samples_data (dict): Processed sample data.
+                See process_samples() for details.
+
+        Raises:
```
ryanjulian: i don't see a `raise` statement anywhere?

src/garage/tf/algos/reps.py
naeioi: This is because pylint cannot recognize the reference format. I tried to use this format, 
```
References:
  ...
```
But the complaints did not go away.

src/garage/tf/algos/reps.py
naeioi: This is also because flake8 cannot recognize the refences section in docstring.

src/garage/tf/algos/reps.py
naeioi: Fixed. Also did so for NPO.py

src/garage/tf/algos/npo.py
naeioi: My fault. Fixed.

PR #1058, opened 2019-11-20
Add TrajectoryBatch data type

ryanjulian:
Recent experience has shown that many bugs and API ambiguities could
be avoided if some of the abstract data types passed around by
components of garage were well-defined. This PR attempts to define one
such data type, TrajectoryBatch, which represents trajectory data in
the form frequently used by on-policy RL algorithms.

An advantage of this approach is that documenting the inputs and
outputs of various parts of the data processing pipeline becomes much
simpler. In the future, we can also add validation checks and other
helpers to these classes to make data type creation and conversion
more predictable and safer.

### DIFF
--- a/src/garage/_dtypes.py
+++ b/src/garage/_dtypes.py
@@ -1,0 +1,169 @@
+"""Data types for agent-based learning."""
+import collections
+
+import numpy as np
+
+
+class TrajectoryBatch(
+        collections.namedtuple('TrajectoryBatch', [
+            'env_spec',
+            'observations',
+            'actions',
+            'rewards',
+            'terminals',
+            'env_infos',
+            'agent_infos',
+            'lengths',
+        ])):
+    # pylint: disable=missing-return-doc, missing-return-type-doc, missing-param-doc, missing-type-doc  # noqa: E501
+    r"""A tuple representing a batch of whole trajectories.
+
+    A :class:`TrajectoryBatch` represents a batch of whole trajectories
+    produced when one or more agents interacts with one or more environments.
+
+    +-----------------------+-------------------------------------------------+
+    | Symbol                | Description                                     |
+    +=======================+=================================================+
+    | :math:`N`             | Trajectory index dimension                      |
+    +-----------------------+-------------------------------------------------+
+    | :math:`[T]`           | Variable-length time dimension of each          |
+    |                       | trajectory                                      |
+    +-----------------------+-------------------------------------------------+
+    | :math:`S^*`           | Single-step shape of a time-series tensor       |
+    +-----------------------+-------------------------------------------------+
+    | :math:`N \bullet [T]` | A dimension computed by flattening a            |
+    |                       | variable-length time dimension :math:`[T]` into |
+    |                       | a single batch dimension with length            |
+    |                       | :math:`sum_{i \in N} [T]_i`                     |
+    +-----------------------+-------------------------------------------------+
+
+    Attributes:
+        env_spec (garage.envs.EnvSpec): Specification for the environment from
+            which this data was sampled.
+        observations (numpy.ndarray): A numpy array of shape
+            :math:`(N \bullet [T], O^*)` containing the (possibly
+            multi-dimensional) observations for all time steps in this batch.
+            These must conform to :obj:`env_spec.observation_space`.
+        actions (numpy.ndarray): A  numpy array of shape
+            :math:`(N \bullet [T], A^*)` containing the (possibly
+            multi-dimensional) actions for all time steps in this batch. These
+            must conform to :obj:`env_spec.action_space`.
+        rewards (numpy.ndarray): A numpy array of shape
+            :math:`(N \bullet [T])` containing the rewards for all time steps
+            in this batch.
+        terminals (numpy.ndarray): A boolean numpy array of shape
+            :math:`(N \bullet [T])` containing the termination signals for all
+            time steps in this batch.
+        env_infos (dict): A dict of numpy arrays arbitrary environment state
+            information. Each value of this dict should be a numpy array of
+            shape :math:`(N \bullet [T])` or :math:`(N \bullet [T], S^*)`.
+        agent_infos (numpy.ndarray): A dict of numpy arrays arbitrary agent
+            state information. Each value of this dict should be a numpy array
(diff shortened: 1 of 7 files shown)


src/garage/_dtypes.py
```diff
@@ -0,0 +1,51 @@
...
+        `N * (T)` represents flattening the `N` and `(T)` dimensions into a
+        single fixed-length batch dimension, which can be imagined as arranging
+        all `N` trajectories along a single time axis, from left to right.
+
+    Attributes:
+        observations (numpy.ndarray): A numpy array of shape `[N * (T), *dims]`
+            containing the (possibly multi-dimensional) observations for all
+            time steps in this batch.
+        actions (numpy.ndarray): A  numpy array of shape `[N * (T), *dims]`
+            containg the (possibly multi-dimensional) actions for all time
```
ahtsan: containing

src/garage/_dtypes.py
```diff
@@ -0,0 +1,51 @@
...
+        terminals (numpy.ndarray): A boolean numpy array of shape `[N * (T)]`
+            containing the termination signals for all time steps in this batch
+        env_infos (numpy.record): A structured numpy array containing
+            arbitrary environment state information, stored as numpy arrays of
+            shape `[N * (T)]` or `[N * (T), *dims]`.
+        agent_infos (numpy.record): A structured numpy array containing
+            arbitrary agent state information, stored as numpy arrays of shape
+            `[N * (T)]` or `[N * (T), *dims]`. For example, this may contain
+            the hidden states from an RNN policy.
+        lengths (numpy.ndarray): A uint numpy array of shape `[N * (T)]`
```
ahtsan: I think `lengths` has shape `[N]` since each entry represents the length of each trajectory.

ryanjulian: good point, thanks

ryanjulian:
@ahtsan the purpose of an abstract data type is to be universal, so adding subclasses is usually a bad idea.

what are some examples of the data the algorithms add?

ryanjulian:
@ahtsan note that this is tuple type, so its immutable anyway. there's no way for an algorithm to modify it (but it could create another one).

I think you're referring to BatchPolopt which does significant processing on trajectories. Perhaps BatchPolopt can define a different data type like this one. Then you could view `process_samples` as a routine which transforms a `TrajectoryBatch` into a `BatchPoloptBatch` (for instance).

ahtsan:
Yes, I agree we should have another `BatchPoloptBatch`

ryanjulian:
@krzentner I'd like your opinion before this is merged

@ahtsan does this match with what you're implementing for the distributions refactor?

ahtsan:
@ryanjulian My changes are mostly in the `BatchPolopt2` so I guess that would be under `BatchPoloptBatch`

ryanjulian:
@ahtsan how does BatchPolopt differ from this data structure?

ryanjulian:
@avnishn No, it's TBD what we will use for single transitions and batches of non-trajectory transitions.

You could imagine they will be types which are similar to this.

avnishn:
Sounds good to me ryan

krzentner:
I think this is generally the right idea. My main complaint is that it is immediately going to be used to hold both trajectories with flattened observations, and trajectories with unflattened observations. Without some clear indication of which it contains, we're in much the same mess we are when we started. Worse, we won't catch these cases until we test an environment with e.g. unflattened observations.

I think that we should also store the env spec in this type, and check consistency with the env spec (by calling `.contains(array[0])` on each array and matching space) when it's constructed. We should probably also create a way of getting the data in flattened form (which is easy to do with the env spec). This also prevents the problem of "I have observations, but I don't know how to flatten them," since trajectories will always come with an env spec.

ryanjulian:
great point -- i will update it

is `EnvSpec` enough? what about the specs for `env_infos` and `agent_infos`?

ahtsan:
I think we need `max_path_length` when we unflatten observations into dense format (we basically reconstruct each trajectory by `lengths` and pad it to `max_path_length`).

For `agent_infos`, currently we have `policy.state_info_spec` to store the keys in `agent_info`.
We should actually have another `env_info_spec` to store keys in `env_info`, but not sure who should be the owner, maybe the algorithm? as a policy should not be environment-dependent. But I think we still don't have an example of how `env_info` will be used in codebase.

ryanjulian:
if there an element of the system which doesn't know `max_path_length` when it needs to process this data type? will you be able to remove `max_path_length` from algorithm args (and make it just a sampler arg) if we use this dtype?

`policy.state_info_spec` should be changed to `policy.agent_info_spec` (or `agent_info` should be renamed to `policy_info`), and it should contain a dict of `akro.spaces` objects attached to the keys.

`env_info_spec` is unfortunately a gray area, singe gym doesn't even have such an interface. perhaps we can have GarageEnvs provide `env_info_spec`. that way, env_infos will still have all the data from env_infos, but we can only automatically process data mentioned in `env_info_spec`.

ahtsan:
> if there an element of the system which doesn't know `max_path_length` when it needs to process this data type?

This dtype is for the samples generated by on-policy sampler so the consumer would mostly be on-policy algorithm. Currently we pass `max_path_length` to `BatchPolopt` but it's not enforced by any interface.

> will you be able to remove `max_path_length` from algorithm args (and make it just a sampler arg) if we use this dtype?

Yes I think we can remove `max_path_length` in algorithm args. And we include `max_path_length` in this dtype?

> 
> `policy.state_info_spec` should be changed to `policy.agent_info_spec` (or `agent_info` should be renamed to `policy_info`), and it should contain a dict of `akro.spaces` objects attached to the keys.

Yes, it's very confusing now. That's a good idea.

> 
> `env_info_spec` is unfortunately a gray area, singe gym doesn't even have such an interface. perhaps we can have GarageEnvs provide `env_info_spec`. that way, env_infos will still have all the data from env_infos, but we can only automatically process data mentioned in `env_info_spec`.

Sounds good. We should put that in `EnvSpec` though, since algorithm only has access to `EnvSpec` but not the `env` itself.

ryanjulian:
Do your newer versions of NPO use `max_path_length` in `init_opt`? If it does, then it needs it (because it won't have any data during graph building).

ahtsan:
I forgot to check NPO and yes it needs it. Let's stick with what it is now.

PR #1059, opened 2019-11-21
New primitives for PEARL

lywong92:
This PR contains implementation of a recurrent encoder and an inference network needed in PEARL SAC. I'll keep updating this as I progress through the rest of the implementation. I just put everything under the algo folder for now. Please let me know if I should place these elsewhere.


ryanjulian:
Let's get these submitted and then you can follow up with the algorithm. That will help us avoid a mega-PR.

src/garage/torch/algos/recurrent_encoder.py
```diff
@@ -0,0 +1,79 @@
+# pylint: disable=attribute-defined-outside-init
+"""A recurrent network with LSTM for encoding context of RL tasks."""
+
+import torch
+from torch import nn as nn
+
+from garage.torch.modules import MLPModule
+
+
+class RecurrentEncoder(MLPModule):
```
ryanjulian: you can put this in a new module called `garage.torch.embeddings`

src/garage/torch/modules/mlp_module.py
```diff
@@ -56,6 +56,12 @@ def __init__(self,
                          output_nonlinearity, output_w_init, output_b_init,
                          layer_normalization)
 
+        self.input_dim = input_dim
```
ryanjulian: can you make these _private?

src/garage/torch/algos/recurrent_encoder.py
```diff
@@ -0,0 +1,79 @@
...
+    """This recurrent network encodes context of RL tasks.
+
+    Context is stored in the terms of observation, action, and reward, and this
+    network uses an MLP module followed by an LSTM model for encoding.
+
+    Args:
```
ryanjulian: hmm. this can cause maintainability problems. can you list the args/kwargs explicitly?

src/garage/torch/algos/pearl_inference_network.py
```diff
@@ -0,0 +1,224 @@
...
+    variable Z. This inference network estimates the posterior probability of
+    z given past transitions. It uses context information stored in the
+    encoder to infer the probabilistic value of z and samples from a policy
+    conditioned on z.
+
+    Args:
+        latent_dim (int): Latent context variable dimension.
+        context_encoder (garage.torch.algo.RecurrentEncoder): Recurrent
+            context encoder.
+        policy (garage.torch.policies): Policy used to train the network.
```
ryanjulian: `garage.torch.policies.Policy`

src/garage/torch/algos/pearl_inference_network.py
```diff
@@ -0,0 +1,224 @@
...
+            use_information_bottlebeck (bool): True if latent context is not
+                deterministic.
+            use_next_obs_in_context (bool): True if next observation is used in
+                distinguishing tasks.
+
+    """
+
+    def __init__(self, latent_dim, context_encoder, policy, **kwargs):
+        super().__init__()
+        self.latent_dim = latent_dim
```
ryanjulian: can you please make all attrs _private?

src/garage/torch/algos/pearl_inference_network.py
```diff
@@ -0,0 +1,224 @@
...
+            recurrent (bool): True if encoder is recurrent; false if it is
+                permutation-invariant.
+            use_information_bottlebeck (bool): True if latent context is not
+                deterministic.
+            use_next_obs_in_context (bool): True if next observation is used in
+                distinguishing tasks.
+
+    """
+
+    def __init__(self, latent_dim, context_encoder, policy, **kwargs):
```
ryanjulian: hmm, there is no need here for **kwargs if your class has no keyword args.

src/garage/torch/modules/mlp_module.py
lywong92: I need to be able access output_dim outside. Should I just add a getter function?

src/garage/torch/algos/recurrent_encoder.py
```diff
@@ -0,0 +1,79 @@
...
+        # hn = last hidden state with gradients
+        out, (hn,
+              _) = self.lstm(out,
+                             (self.hidden, torch.zeros(self.hidden.size())))
+        self.hidden = hn
+        # take the last hidden state to predict z
+        out = out[:, -1, :]
+
+        # output layer
+        output = self._output_layers[-1](out)
```
katerakelly: why only use the last output layer?

src/garage/torch/algos/pearl_inference_network.py
```diff
@@ -0,0 +1,224 @@
...
+"""
+
+import torch
+from torch import nn as nn
+import torch.nn.functional as F
+
+from garage.torch.utils import np_to_torch
+
+
+class PEARLInferenceNetwork(nn.Module):
```
katerakelly: Usually the term inference network refers to the network that parameterizes the variational approximation to the posterior. I think it's weird to call something that contains the policy an inference network.

katerakelly:
I agree with Ryan's comment about re-using modules for algs other than PEARL if you're planning on adding other meta-RL algs as well. Maybe this is helpful: https://github.com/tristandeleu/pytorch-meta

src/garage/torch/algos/__init__.py
```diff
@@ -1,7 +1,11 @@
 """PyTorch algorithms."""
 from garage.torch.algos.ddpg import DDPG
-from garage.torch.algos.vpg import VPG
+from garage.torch.algos.pearl_inference_network import PEARLInferenceNetwork
```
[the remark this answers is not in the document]
lywong92: We haven't decided the structure for meta algorithms and their primitives yet, so I placed them with our torch algos for now. I'm gonna move them.

src/garage/torch/algos/pearl_inference_network.py
lywong92: How does inference agent sound? Or do you have any suggestions?

src/garage/torch/algos/pearl_inference_network.py
```diff
@@ -0,0 +1,224 @@
...
+    """A network for infering a posterior over the latent context Z.
+
+    In PEARL, policies are conditioned on current state and a latent context
```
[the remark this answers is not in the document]
lywong92: I will include both and make a note that the paper refers to it as context.

src/garage/torch/algos/pearl_inference_network.py
lywong92: @ryanjulian Do you mean that I don't need **kwargs because I don't need it for the super init call?

ryanjulian: when you add `**kwargs` to the end of a Python function, it will capture any keyword arguments which that function is ever called with, even ones you never expected.

This seems convenient at first (often for writing libraries), but for general-purpose Python code it easily creates hard-to-find bugs.
* Using **kwargs turns off the check for whether you passed a kwarg which wasn't expected.
* Passing **kwargs to a super constructor carries these unexpected arguments all the way down the call chain. This means that if you need to refactor something (e.g. by removing or adding an argument), now you're not sure what is actually getting passed down the call chain.
* When you collect kwargs using **kwargs and pass them to a super constructor, you're hiding which parts of the super constructor you actually use. If you keep your call to the super constructor totally-explicit, it's easy to see the relationship between `super` and the child class
* Swallowing kwargs removes your actual arguments from the call signature. When you document this using the docstring, users will only see `def foo(bar, baz, **kwargs)` -- then they will have to explore the superclass (or 2 or 3) to find the actual arugments. It's better just to copy them and the documentation, even if that seems repetitive, because the user wants to see the documentation for the component they are trying to use as fast as possible, regardless of the internal structure of the library.

src/garage/torch/algos/pearl_inference_network.py
```diff
@@ -0,0 +1,224 @@
...
+        params = self.context_encoder(context)
+        params = params.view(context.size(0), -1,
+                             self.context_encoder.output_dim)
+        # given context, compute mean and variance of q(z|c)
+        if self.use_ib:
+            mu = params[..., :self.latent_dim]
+            sigma_squared = F.softplus(params[..., self.latent_dim:])
+            z_params = []
+            # compute mu, sigma of product of gaussians
+            for m, s in zip(torch.unbind(mu), torch.unbind(sigma_squared)):
```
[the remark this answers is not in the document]
lywong92: Noted!

src/garage/torch/algos/pearl_inference_network.py
```diff
@@ -0,0 +1,224 @@
...
+
+        Args:
+            obs (torch.Tensor): Observation values.
+
+        Returns:
+            torch.Tensor: Output action values.
+
+        """
+        z = self.z
+        obs = torch.unsqueeze(obs, 0)
```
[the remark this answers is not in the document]
lywong92: In our policies, we usually have get_action and get_actions, and we add an extra dimension in get_action.

src/garage/torch/algos/recurrent_encoder.py
```diff
@@ -0,0 +1,79 @@
...
+            out = layer(out)
+            if self._hidden_nonlinearity is not None:
+                out = self._hidden_nonlinearity(out)
+        out = out.view(task, seq, -1)
+
+        # add LSTM before output layer
+        # step through the entire sequence of LSTM all at once
+        # out = all hidden states in the sequence
+        # hn = last hidden state with gradients
+        out, (hn,
```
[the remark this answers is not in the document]
lywong92: Flake8 keeps changing it to this formatting.

src/garage/torch/algos/recurrent_encoder.py
lywong92: Our MLP implementation inherits from an MLP class with multiple parallel output layers, and an instance of MLP only has 1 output layer.

src/garage/torch/embeddings/recurrent_encoder.py
```diff
@@ -0,0 +1,111 @@
+# pylint: disable=attribute-defined-outside-init
+"""A recurrent network with LSTM for encoding context of RL tasks."""
+
+import torch
+from torch import nn
+
+from garage.torch.modules import MLPModule
+
+
+class RecurrentEncoder(MLPModule):
```
krzentner: It feels pretty weird that this inherits from `MLPModule`, since it's not obvious that the MLP layers will be used to encode the input, then the output layer will be used to encode the final output layer. I think this would probably be cleaner by just making MLPModule a field?

It's not really a must-fix issue though.

src/garage/torch/policies/context_conditioned_policy.py
```diff
@@ -0,0 +1,219 @@
+# pylint: disable=attribute-defined-outside-init
```
krzentner: As far as the code is concerned, I think this file looks good. I haven't checked the math though, since I'm not familiar with PEARL.

lywong92:
@ryanjulian Could you take a look at this again?

src/garage/torch/embeddings/recurrent_encoder.py
ryanjulian: yeah this is pretty strange. what do you use from MLPModule that you wouldn't get from module?

perhaps you actually just want your recurrent encoder to have-a MLPModule (rather than be-a MLPModule)

src/garage/torch/embeddings/recurrent_encoder.py
```diff
@@ -0,0 +1,110 @@
...
+                torch.Tensor.
+            layer_normalization (bool): Bool for using layer normalization or
+                not.
+
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._hidden_dim = self._hidden_sizes[-1]
+        # hidden dimension should be (task, 1, feat)
```
ryanjulian: hmm if this is essential, considering checking and raising an exception or asserting.

use an assert if the rest of your code should have already assured this

use an exception if incorrect user input could cause this to happen.

src/garage/torch/policies/context_conditioned_policy.py
```diff
@@ -0,0 +1,271 @@
...
+        z = self.z
+        obs = tu.from_numpy(obs[None])
+        obs_in = torch.cat([obs, z], dim=1)
+        action, info = self._policy.get_action(obs_in)
+        action = np.squeeze(action, axis=0)
+        info['mean'] = np.squeeze(info['mean'], axis=0)
+        return action, info
+
+    def compute_kl_div(self):
+        """Compute KL( q(z|c) || p(z) ).
```
ryanjulian: you can use a math environment to show these very nicely in LaTeX.

see https://github.com/rlworkgroup/garage/blob/master/src/garage/_dtypes.py for an example

src/garage/torch/policies/context_conditioned_policy.py
```diff
@@ -0,0 +1,271 @@
...
+        """Return context.
+
+        Returns:
+            torch.Tensor: Context values.
+
+        """
+        return self._context
+
+
+def _product_of_gaussians(mus, sigmas_squared):
```
ryanjulian: hmm perhaps this can go into garage.torch.utils?

src/garage/torch/utils.py
```diff
@@ -130,3 +130,43 @@ def global_device():
     # pylint: disable=global-statement
     global _DEVICE
     return _DEVICE
+
+
+def from_numpy(*args):
```
ryanjulian: please coordinate with Avnish -- i think these appear in one of his PRs as well.

tests/garage/torch/modules/test_mlp_module.py
```diff
@@ -138,4 +138,5 @@ def test_mlp_with_learnable_non_linear_function(self):
         output.sum().backward()
 
         for tt in module.parameters():
-            assert torch.all(torch.ne(tt.grad, 0))
+            if tt.grad is not None:
```
ryanjulian: can you explain why this branch is needed?

conditionals in test cases usually indicated something fishy is going on

tests/garage/torch/policies/test_context_conditioned_policy.py
```diff
@@ -0,0 +1,115 @@
...
+class TestContextConditionedPolicy:
+    """Test for ContextConditionedPolicy."""
+    # yapf: disable
+    @pytest.mark.parametrize('latent_dim, hidden_sizes, updates', [
+        (5, (3, 3, 3), 5),
+        (7, (5, 7, 5), 7),
+        (10, (9, 4, 2), 10),
+    ])
+    # yapf: enable
+    def test_module(self, latent_dim, hidden_sizes, updates):
```
ryanjulian: can you split this test into a setup phase followed by cases for each of your major assertions? that will make it easier to understand and maintain.

src/garage/torch/policies/context_conditioned_policy.py
```diff
@@ -0,0 +1,271 @@
...
+    # pylint: disable=arguments-differ
+    def forward(self, obs, context):
+        """Given observations and context, get actions and probs from policy.
+
+        Args:
+            obs (torch.Tensor): Observation values.
+            context (torch.Tensor): Context values.
+
+        Returns:
+            torch.Tensor: Output action values.
```
ryanjulian: this doesn't seem to match the actual returns...

tests/garage/torch/modules/test_mlp_module.py
lywong92: There's supposed to be no changes in this file. I forgot to delete it.

src/garage/torch/policies/context_conditioned_policy.py
lywong92: Done.

src/garage/torch/embeddings/mlp_encoder.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class MLPEncoder(MLPModule):
+    """This MLP network encodes context of RL tasks.
+
+    Context is stored in the terms of observation, action, and reward, and this
+    network uses an MLP module for encoding it.
+
+    Args:
+        *args: MLPModule arguments.
```
ryanjulian: please just reproduce the arguments here

src/garage/torch/embeddings/mlp_encoder.py
```diff
@@ -0,0 +1,49 @@
...
+                torch.Tensor.
+            output_b_init (callable): Initializer function for the bias
+                of output dense layer(s). The function should return a
+                torch.Tensor.
+            layer_normalization (bool): Bool for using layer normalization or
+                not.
+
+    """
+
+    def reset(self, num_tasks=1):
```
ryanjulian: hmm perhaps you want to define a ContextEncoder ABC? it might make this class make a lot more sense.

src/garage/torch/policies/context_conditioned_policy.py
```diff
@@ -0,0 +1,268 @@
+# pylint: disable=attribute-defined-outside-init, protected-access
```
ryanjulian: can you explain why each of these is necessary?

src/garage/torch/embeddings/context_encoder.py
```diff
@@ -0,0 +1,15 @@
+# pylint: disable=too-few-public-methods
+"""Base context encoder class."""
+import abc
+
+
+class ContextEncoder(abc.ABC):
+    """Base class of context encoders for training meta-RL algorithms."""
+
+    def reset(self, num_tasks=1):
```
ryanjulian: Do you want to mark this as abstract? That will force implementers to implement it.

src/garage/torch/policies/context_conditioned_policy.py
```diff
@@ -0,0 +1,268 @@
...
+    information stored in the encoder to infer the probabilistic value of z and
+    samples from a policy conditioned on z.
+
+    Args:
+        latent_dim (int): Latent context variable dimension.
+        context_encoder (garage.torch.embeddings): Recurrent or
+            permutation-invariant context encoder.
+        policy (garage.torch.policies.Policy): Policy used to train the
+            network.
+        use_ib (bool): True if latent context is not deterministic; false
```
ryanjulian: please make this longer -- `use_information_bottleneck`. it's more typing, but makes the API more discoverable.

src/garage/torch/policies/context_conditioned_policy.py
```diff
@@ -0,0 +1,268 @@
...
+        # over latent task belief maintained in the policy; z is a sample from
+        # this distribution that the policy is conditioned on
+        self.register_buffer('z', torch.zeros(1, latent_dim))
+        self.register_buffer('z_means', torch.zeros(1, latent_dim))
+        self.register_buffer('z_vars', torch.zeros(1, latent_dim))
+
+        self.reset_belief()
+
+    def reset_belief(self, num_tasks=1):
+        """Reset q(z|c) to the prior and sample a new z from the prior.
```
ryanjulian: you can use math environments to make these equations render as LaTeX. see `src/garage/_dtypes.py` for an example.

src/garage/torch/policies/context_conditioned_policy.py
```diff
@@ -0,0 +1,268 @@
...
+                    torch.unbind(self.z_means), torch.unbind(self.z_vars))
+            ]
+            z = [d.rsample() for d in posteriors]
+            self.z = torch.stack(z)
+        else:
+            self.z = self.z_means
+
+    def detach_z(self):
+        """Disable backprop through z."""
+        self.z = self.z.detach()
```
ryanjulian: hmm this seems like it might be  dangerous bit of statefulness. is this reversible?

@krzentner PTAL

src/garage/torch/policies/context_conditioned_policy.py
```diff
@@ -0,0 +1,268 @@
...
+
+    def update_context(self, inputs):
+        """Append single transition to the current context.
+
+        Args:
+            inputs (dict): Dictionary of transition information in np arrays .
+
+        """
+        o, a, r, no, _, _ = inputs
```
ryanjulian: Is it possible for you to use https://github.com/rlworkgroup/garage/blob/master/src/garage/_dtypes.py#L410 for this? If not, can you accept a concrete set of args rather than an opaque dict?

src/garage/torch/policies/context_conditioned_policy.py
```diff
@@ -0,0 +1,268 @@
...
+        if self._context is None:
+            self._context = data
+        else:
+            self._context = torch.cat([self._context, data], dim=1)
+
+    def infer_posterior(self, context):
+        """Compute q(z|c) as a function of input context and sample new z.
+
+        Args:
+            context (torch.Tensor): Context values.
```
ryanjulian: can you tell us something about the shape of context? See some of Anson and Avnish's recent PRs for an example.

src/garage/torch/policies/context_conditioned_policy.py
```diff
@@ -0,0 +1,268 @@
...
+        else:
+            self.z_means = torch.mean(params, dim=1)
+        self.sample_from_belief()
+
+    # pylint: disable=arguments-differ
+    def forward(self, obs, context):
+        """Given observations and context, get actions and probs from policy.
+
+        Args:
+            obs (torch.Tensor): Observation values.
```
ryanjulian: can you tell us something about the expected shapes of these args.

src/garage/torch/policies/context_conditioned_policy.py
```diff
@@ -0,0 +1,268 @@
...
+            context (torch.Tensor): Context values.
+
+        Returns:
+            tuple:
```
ryanjulian: which of these is used most commonly?

src/garage/torch/policies/context_conditioned_policy.py
```diff
@@ -0,0 +1,268 @@
...
+        mean = dist.mean.to('cpu').detach().numpy()
+        log_std = (dist.variance**.5).log().to('cpu').detach().numpy()
+
+        return (actions, mean, log_std, log_pi, pre_tanh), task_z
+
+    def get_action(self, obs):
+        """Sample action from the policy, conditioned on the task embedding.
+
+        Args:
+            obs (torch.Tensor): Observation values.
```
ryanjulian: can you tell us something about the expcted shapes of args and returns?

src/garage/torch/policies/context_conditioned_policy.py
```diff
@@ -0,0 +1,268 @@
...
+            for post in posteriors
+        ]
+        kl_div_sum = torch.sum(torch.stack(kl_divs))
+        return kl_div_sum
+
+    def reset(self, dones=None):
+        """Reset the environment.
+
+        Args:
+            dones (numpy.ndarray): Reset values.
```
ryanjulian: can you tell us something about the expected shape of dones?

src/garage/torch/policies/context_conditioned_policy.py
```diff
@@ -0,0 +1,268 @@
...
+
+        """
+        return [self._context_encoder, self._policy]
+
+    @property
+    def context(self):
+        """Return context.
+
+        Returns:
+            torch.Tensor: Context values.
```
ryanjulian: can you tell us something about the expected shape of this tensor?

src/garage/torch/utils.py
```diff
@@ -130,3 +130,19 @@ def global_device():
...
     global _DEVICE
     return _DEVICE
+
+
+def product_of_gaussians(mus, sigmas_squared):
+    """Compute mu, sigma of product of gaussians.
+
+    Args:
+        mus (torch.Tensor): Means.
+        sigmas_squared (torch.Tensor): Variances.
```
ryanjulian: please include shape information

src/garage/torch/policies/context_conditioned_policy.py
ryanjulian: also, please only use these disables at the callsite, rather than for the entire file

src/garage/torch/policies/context_conditioned_policy.py
```diff
@@ -0,0 +1,283 @@
...
+            context (torch.Tensor): Context values, with shape
+                :math:`(X, N, C)`. X is the number of tasks. C is the
+                combined size of observation, action, reward, and next
+                observation if next observation is used in context. Otherwise,
+                C is the combined size of observation, action, and reward.
+
+        """
+        params = self._context_encoder(context)
+        # pylint: disable=protected-access
+        params = params.view(context.size(0), -1,
```
ryanjulian: why can't this be a property instead (and not need protected access)?

src/garage/torch/policies/context_conditioned_policy.py
```diff
@@ -0,0 +1,283 @@
...
+        else:
+            self._context = torch.cat([self._context, data], dim=1)
+
+    # pylint: disable=attribute-defined-outside-init
+    def infer_posterior(self, context):
+        r"""Compute :math:`q(z \| c)` as a function of input context and sample new z.
+
+        Args:
+            context (torch.Tensor): Context values, with shape
+                :math:`(X, N, C)`. X is the number of tasks. C is the
```
ryanjulian: what is N?

src/garage/torch/policies/context_conditioned_policy.py
```diff
@@ -0,0 +1,283 @@
...
+            data = torch.cat([o, a, r, no], dim=2)
+        else:
+            data = torch.cat([o, a, r], dim=2)
+
+        if self._context is None:
+            self._context = data
+        else:
+            self._context = torch.cat([self._context, data], dim=1)
+
+    # pylint: disable=attribute-defined-outside-init
```
ryanjulian: please correct this by declaring this attribute in the constructor (either it has a sane default value, or your other methods will need to check that it has been initialized).

src/garage/torch/policies/context_conditioned_policy.py
```diff
@@ -0,0 +1,283 @@
...
+
+        Returns:
+            tuple:
+                * torch.Tensor: Predicted action values.
+                * np.ndarray: Mean of distribution.
+                * np.ndarray: Log std of distribution.
+                * torch.Tensor: Log likelihood of distribution.
+                * torch.Tensor: Sampled values from distribution before
+                    applying tanh transformation.
+            torch.Tensor: z values, with shape :math:`(N, latent_dim)`.
```
ryanjulian: please use only symbols, not text, in math environments.

are the z values the most frequently used output here? if so, it would be most pythonic for z to be the first output, and the others to be packed into the second tuple, e.g.

```python
z, _ = foo.forward()
```

In fact, how many of these outputs are actually used? If one of them isn't just remove it from the returns.

src/garage/torch/policies/context_conditioned_policy.py
```diff
@@ -0,0 +1,283 @@
...
+                    torch.unbind(self.z_means), torch.unbind(self.z_vars))
+            ]
+            z = [d.rsample() for d in posteriors]
+            self.z = torch.stack(z)
+        else:
+            self.z = self.z_means
+
+    def detach_z(self):
+        """Disable backprop through z."""
+        # pylint: disable=attribute-defined-outside-init
```
ryanjulian: please correct this.

src/garage/torch/policies/context_conditioned_policy.py
```diff
@@ -0,0 +1,283 @@
...
+                torch.unbind(self.z_means), torch.unbind(self.z_vars))
+        ]
+        kl_divs = [
+            torch.distributions.kl.kl_divergence(post, prior)
+            for post in posteriors
+        ]
+        kl_div_sum = torch.sum(torch.stack(kl_divs))
+        return kl_div_sum
+
+    def log_diagnostics(self, eval_statistics):
```
ryanjulian: this style of log diagnostics API isn't used in garage -- instead we log directly from log diagnostics.

Your log should have a prefix with the name of your primitive, i.e. the info is logged to `Name/ZMeanEval`, etc.

The default name should be the class name, but you should provide users the option (constructor argument) to name the primitive something different (self._name).

src/garage/torch/policies/context_conditioned_policy.py
```diff
@@ -0,0 +1,283 @@
...
+
+    @property
+    def networks(self):
+        """Return context_encoder and policy.
+
+        Returns:
+            list: Encoder and policy networks.
+
+        """
+        return [self._context_encoder, self._policy]
```
ryanjulian: what is this used for?

src/garage/torch/policies/context_conditioned_policy.py
```diff
@@ -0,0 +1,270 @@
...
+        self.z_means = mu
+        self.z_vars = var
+        # sample a new z from the prior
+        self.sample_from_belief()
+        # reset the context collected so far
+        self._context = None
+        # reset any hidden state in the encoder network (relevant for RNN)
+        self._context_encoder.reset(num_tasks)
+
+    # pylint: disable=attribute-defined-outside-init
```
ryanjulian: please also correct this one

src/garage/torch/policies/context_conditioned_policy.py
```diff
@@ -0,0 +1,270 @@
...
+            self.z = torch.stack(z)
+        else:
+            self.z = self.z_means
+
+    def detach_z(self):
+        """Disable backprop through z."""
+        # pylint: disable=attribute-defined-outside-init
+        self.z = self.z.detach()
+
+    # pylint: disable=attribute-defined-outside-init
```
ryanjulian: please correct this

src/garage/torch/policies/context_conditioned_policy.py
ryanjulian: basically, i'm asking if this is really the best return signature.

if most people just want `z`, then i suggest your signature look like

```python
"""
    Returns:
        torch.Tensor: z values....
        tuple: A bunch of the other stuff
"""

# This would give you a callsite like this
z, _ = Encoder.forward(stuff)
```
This pattern puts the most common output as the first, and then if you really need something you can unpack what we throw away in `_`.

Another question I have is, "Do you have examples of your code using every element of the return tuple?" If not, I suggest you just remove them. Big APIs are hard to keep promises for. If you use one part of a symmetric pair, such as mean but not log_std, it's fine to also return log_std. But I'm skeptical that all of these values are used.

PR #1065, opened 2019-11-22
Add instructions for documenting tensor shapes

ryanjulian:
Updates CONTRIBUTING.md with guidelines for documenting tensor shapes.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -168,12 +168,58 @@ These are garage-specific rules which are not part of the aforementioned style g
     import dowel.tabular as tabular
     ```
 
+### Other languages
+Non-Python files (including XML, HTML, CSS, JS, and Shell Scripts) should follow the [Google Style Guide](https://github.com/google/styleguide) for that language
+
+YAML files should use 2 spaces for indentation.
+
+### Whitespace (all languages)
+* Use Unix-style line endings
+* Trim trailing whitespace from all lines
+* All files should end in a single newline
+
 ## Documentation
 Python files should provide docstrings for all public methods which follow [PEP257](https://www.python.org/dev/peps/pep-0257/) docstring conventions and [Google](http://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings) docstring formatting. A good docstring example can be found [here](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html).
 
-Additional standards:
+### Docstrings
 * Docstrings for `__init__` should be included in the class docstring as suggested in the [Google example](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html).
 * Docstrings should provide full type information for all arguments, return values, exceptions, etc. according to the Google format
+* When documenting fields which are numpy arrays or other tensor types (and collections thereof), please carefully document the expected input shape of the field. See below for shape conventions.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -168,12 +168,58 @@ These are garage-specific rules which are not part of the aforementioned style g
...
+Always use the Sphinx `:math:` directive to render your shapes properly.
+
+| Symbol    | Description                                                                                             |
+|-----------|---------------------------------------------------------------------------------------------------------|
+| `(...)`   | Tensor shapes are enclosed in parentheses, e.g a batch of `(N, S^*)` samples                            |
+| `N`       | Batch dimension (e.g. trajectories or samples)                                                          |
+| `T`       | Time dimension                                                                                          |
+| `S^*`     | Variadic parts of a tensor shape, which will be broadcast or ignored                                    |
+| `[.]`     | Variable-length dimensions are enclosed in square brackets, e.g. `[K]` if `K` is the dimension variable |
+| `\bullet` | Flattening operator, e.g. `(N \bullet T, K)` has a flattened first dimension of length `N * T`          |
```
naeioi: I think it will be helpful to point out that then length of `N \bullet [T]` will be `sum(len(T_i))`

CONTRIBUTING.md
```diff
@@ -168,12 +168,58 @@ These are garage-specific rules which are not part of the aforementioned style g
...
+    Args:
+        paths (list[numpy.ndarray]): A list of :math:`N` tensors to combine
+            into a single batch of tensors, with elements of shape
+            :math:`([T], S^*)`
+            
+    Returns:
+        numpy.ndarray: Time-flattened version of `paths`, with shape
+            :math:`(N \bullet [T], S^*)`
+    """
+```
```
ahtsan: Should we also list out the keys and values / only the keys if paths is a dictionary? I think that's usually the case in garage. If so, I believe it would be helpful to illustrate how to render that nicely, e.g.

- Keys and values:
```
...
    Args:
        paths (dict[list]): A dictionary with values of shape :math:`(N \bullet [T], S^*)`.

    Note:
        The returned paths contain the following keys and values are of shape:
        * `observation`: :math:`(N \bullet [T], observation_dim^*)`
        * `action`: :math:`(N \bullet [T], action_dim^*)`
        * `reward`: :math: `(N \bullet [T])`
...
```

Only keys:
```
...
    Args:
        paths (dict[list]): A dictionary with values of shape :math:`(N \bullet [T], S^*)`.

    Note:
        The returned paths contain the keys: `observation`, `action`, `reward`...
...
```

ryanjulian: Yes. Feel free to amend this branch :)

ahtsan: Which one do you think it's better? (I think it's always better to be more detail, but sometimes it might just be an overkill)

ryanjulian: the first one, assuming it renders nicely in Sphinx

you should just use
```python
        * `observation`: :math:`(N \bullet [T], S^*)`
        * `action`: :math:`(N \bullet [T], S^*)`
```

the words won't render nicely in Sphinx, plus it's obvious from the context what `S^*` should be.

ryanjulian: I will let you send the PR to update this @ahtsan

PR #1069, opened 2019-11-23
Set TotalEnvSteps as the default Tensorboard x-axis

naeioi:
This PR sets `TotalEnvSteps` as the default tensorboard scalar log x-axis. This PR depends on https://github.com/rlworkgroup/dowel/pull/38.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,7 @@ REQUIRED = [
     'click',
     'cloudpickle',
     'cma==2.7.0',
-    'dowel==0.0.2',
+    'dowel==0.0.3',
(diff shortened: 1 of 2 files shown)


ryanjulian:
We need to merge the dowel changes first, release a dowel version, then bump the version in garage.

naeioi:
This PR will be ready when dowel 0.0.3 is tagged.

ryanjulian:
Let me know when this is ready and I will force-push it.

naeioi:
@ryanjulian If we don't care about the codedev, then this should be ready. 
The uncovered code was the serializer in the experiment wrapper.

PR #1079, opened 2019-11-27
Add tests for long running examples

gitanshu:
These tests take batch_size or buffer_size as arg using click. Command.main(standalone_mode=False) is used here to disable the implicit sys.exit() at end of click command processing, since that causes experiment subprocess to exit.

Ref: https://click.palletsprojects.com/en/7.x/exceptions/

### DIFF
--- a/examples/tf/dqn_pong.py
+++ b/examples/tf/dqn_pong.py
@@ -1,9 +1,9 @@
 #!/usr/bin/env python3
-"""
-This is an example to train a task with DQN algorithm in pixel environment.
+"""This is an example to train a task with DQN algorithm in pixel environment.
 
 Here it creates a gym environment Pong, and trains a DQN with 1M steps.
 """
+import click
 import gym
 
 from garage.envs.wrappers.clip_reward import ClipReward
@@ -24,8 +24,18 @@ from garage.tf.policies import DiscreteQfDerivedPolicy
 from garage.tf.q_functions import DiscreteCNNQFunction
 
 
-def run_task(snapshot_config, *_):
-    """Run task."""
+def run_task(snapshot_config, variant_data, *_):
+    """Run task.
+
+    Args:
+        snapshot_config (garage.experiment.SnapshotConfig): The snapshot
+            configuration used by LocalRunner to create the snapshotter.
+
+        variant_data (dict): Custom arguments for the task.
+
+        *_ (object): Ignored by this function.
+
+    """
     with LocalTFRunner(snapshot_config=snapshot_config) as runner:
         n_epochs = 100
         n_epoch_cycles = 20
@@ -45,9 +55,10 @@ def run_task(snapshot_config, *_):
 
         env = TfEnv(env)
 
-        replay_buffer = SimpleReplayBuffer(env_spec=env.spec,
-                                           size_in_transitions=int(5e4),
-                                           time_horizon=1)
+        replay_buffer = SimpleReplayBuffer(
+            env_spec=env.spec,
+            size_in_transitions=variant_data['buffer_size'],
+            time_horizon=1)
 
         qf = DiscreteCNNQFunction(env_spec=env.spec,
                                   filter_dims=(8, 4, 3),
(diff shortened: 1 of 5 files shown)


gitanshu:
Haven't added pytest timeout yet for these. Waiting to see how long they take to run on CI.

examples/tf/dqn_pong.py
```diff
@@ -24,8 +24,18 @@
 from garage.tf.q_functions import DiscreteCNNQFunction
 
 
-def run_task(snapshot_config, *_):
-    """Run task."""
+def run_task(snapshot_config, variant_data, *_):
```
ryanjulian: `*_` _is_ variant_data

examples/tf/ppo_memorize_digits.py
```diff
@@ -15,8 +15,18 @@
 from garage.tf.policies import CategoricalCNNPolicy
 
 
-def run_task(snapshot_config, *_):
-    """Run task."""
+def run_task(snapshot_config, variant_data, *_):
```
ryanjulian: `*_` is variant_data

examples/tf/trpo_cubecrash.py
```diff
@@ -15,8 +15,18 @@
 from garage.tf.policies import CategoricalCNNPolicy
 
 
-def run_task(snapshot_config, *_):
-    """Run task."""
+def run_task(snapshot_config, variant_data, *_):
```
ryanjulian: `*_` is variant_data

examples/tf/dqn_pong.py
gitanshu: Acc to this:
https://github.com/rlworkgroup/garage/blob/02076981cd07b2aeee44e57bb6d19b0195426a9a/src/garage/experiment/experiment_wrapper.py#L165-L166
Wouldn't `*_` be `[variant_data, args.resume_from_dir, args.resume_from_epoch]`? Should I use `_[0]['buffer_size']` instead?

ryanjulian: I stand corrected.

ryanjulian: note that `*_` should appear as that in the docstring (not just `_`)

ryanjulian:
Looks like you need to bump the time for ppo_memorize_digits

Can you enable the timing summary for the long-running test job too? It would be nice to see so we can balance the test jobs in the future.

tests/integration_tests/test_examples.py
```diff
@@ -54,6 +54,50 @@ def test_algo_examples(filepath):
...
+    """Test tf/ppo_memorize_digits.py with reduced batch size for reduced memory
+    consumption.
+    """
+    env = os.environ.copy()
+    env['GARAGE_EXAMPLE_TEST_N_EPOCHS'] = '1'
+    assert subprocess.run([
+        str(EXAMPLES_ROOT_DIR / 'tf/ppo_memorize_digits.py'), '--batch_size',
+        '4'
+    ],
+                          check=False,
```
ryanjulian: weird spacing

ryanjulian:
@gitanshu this is blocking `v2019.10.1` -- do you think it can be merged soon?

tests/integration_tests/test_examples.py
gitanshu: yapf wasn't having it, so I extracted popenargs into a variable

gitanshu:
@avnishn this is only for examples which take too much time running even 1 epoch on CI, so if a new example falls under the same category, it'd need a similar solution. This might or might not improve when the new experiment API lands. If this turns out to be case for many new examples, we can review the design.

gitanshu:
@ryanjulian I'll get this merged today

examples/tf/dqn_pong.py
gitanshu: Not sure I get it. Are you saying `[variant_data, args.resume_from_dir, args.resume_from_epoch]` collectively represent variant_data?

I changed it from `_` to `*_` in docstring.

ryanjulian:
@mergifyio rebase

ryanjulian:

PR #1083, opened 2019-12-03
Make matplotlib a dev dependency

ryanjulian:
* Makes matplotlib a dev dependency, since it's quite heavy and not
  actually used in any runtime code.
* Removes usage in OUStrategy and updates the file with docstrings


ryanjulian:
@mergifyio rebase

PR #1092, opened 2019-12-08
Cache parameters shape and values without tag

ahtsan:
This PR removes the use of `**tags` in caching parameters shape and values.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -16,7 +16,7 @@ extend-ignore =
     F841  # Unused variables are checked by pylint
 
 [tool:pytest]
-addopts = -n auto -rfE -s --strict-markers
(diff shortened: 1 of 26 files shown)


setup.cfg
```diff
@@ -16,7 +16,7 @@ extend-ignore =
     F841  # Unused variables are checked by pylint
 
 [tool:pytest]
-addopts = -n auto -rfE -s --strict-markers
+adopts = -n auto -rfE -s --strict-markers
```
ryanjulian: pretty sure this should be `addopts`: http://doc.pytest.org/en/latest/customize.html#adding-default-options

ahtsan: ok

Issue #1095, opened 2019-12-11
Don't scale rewards in off_policy_vectorized_sampler


avnishn:
Currently the off policy sampler, in the process of adding samples to the replay buffer, adds the rewards scaled by the reward factor. I believe that this is an implementation detail that should be added inside of the algorithm for reduced ambiguity.

https://github.com/rlworkgroup/garage/blob/169c822573d3899911590988450e2466767f2451/src/garage/sampler/off_policy_vectorized_sampler.py#L141

PR #1096, opened 2019-12-11
Remove reward scaling from OffPolicySampler

avnishn:
Currently the when adding rewards to transitions in the replay buffer,
we add the reward scaled by the reward factor. This is something that should
be handled in the algorithm, not in the sampler. Furthermore it was incredibly
ambiguous that this was the case.

reference #1095


avnishn:
You're right, reward scaling happens only in off policy algorithms, correct? I'll go ahead and correspondingly make the fixes in the algorithms.

ryanjulian:
I don't know. It happens in whichever algorithms were using this code before.

avnishn:
closed by #1097

PR #1097, opened 2019-12-12
Refactor reward scale into OffPolicyRLAlgorithm

krzentner:
Note that TD3 was actually applying reward scaling twice.


src/garage/np/algos/off_policy_rl_algorithm.py
```diff
@@ -91,6 +91,7 @@ def train(self, runner):
         for _ in runner.step_epochs():
             for cycle in range(self.steps_per_epoch):
                 runner.step_path = runner.obtain_samples(runner.step_itr)
+                runner.step_path['rewards'] *= self.reward_scale
```
ryanjulian: will this affect the rewards reported during evaluation?

krzentner: I believe the current plan is to call `OffPolicyVectorizedSampler` (or just rollout) to evaluate the off policy algorithms, so this should not.

PR #1099, opened 2019-12-16
TEMP: Disable xdist to see CI errors

ryanjulian:


krzentner:
BTW, if this is about all of the tests segfaulting, that happens on my workstation too.

ryanjulian:
This was to ensure I updated the MuJoCo key properly, since pytest-xdist swallows some errors (and it could only be tested in the CI). Closing now.

Issue #1049, opened 2019-11-19
Replace run_experiment with a decorator


ryanjulian:
@krzentner will furnish more details on the design here

krzentner:
We have a large number of requirements we would like our experiment interface to have:
 - It should naturally capture what we think of as an "experiment" - it should have a name, parameters, and manage the life cycle of related objects. In python, the natural way of specifying a "thing with parameters" is a function. Functions are also fairly natural ways of specifying life cycles, and have names.
 - It should be very natural to invoke an experiment with unique parameters.
 - The parameters of an experiment should be saved, and invoking the same experiment with different parameters should result in different log directories.
 - It should be possible to run the experiment from the current process, since sub-processes add significant complexity to development and debugging.
 - The contents of a log directory should be predictable (in other words, all experiment directories should have their sub-files in the same locations).
 - It should be very easy to add command-line parameters, and they should be properly handled as experiment parameters (i.e. saved in `variant.json`).
 - The implementation should not be particular about how command-line argument handling is performed.
 - The implementation should also not be particular about how experiments are distributed across computers.

Essentially, these constraints imply that we want experiments to act in the following ways:
 - They should be callable with their parameters, like normal functions.
 - They should create unique log directories based on those parameters.
 - They should also have access to critical experiment information (at least what's currently contained in the `SnapshotConfig`).

The most natural way to fulfill these requirements is as a decorator which turns a function into an `Experiment` (or perhaps `ExperimentTemplate`).

It would be used something like this:
```python
from garage.experiment import experiment

...

@experiment
def car_trpo(ctxt, max_kl_step=0.01):
    runner = LocalTFRunner(ctxt)
    ...
    runner.train(...)

for max_kl_step in [0.001, 0.01, 0.1]:
    # Produces three experiment directories, car_trpo_1, car_trpo_2, and car_trpo_3
    car_goal_mt_trpo(max_kl_step)
```

Or:
```python
import clize  # Add command line arguments
from garage.experiment import experiment

...

# Produces a new experiment directory each time it's invoked.
@clize.run
@experiment(prefix='remote', dry=True)
def car_trpo(ctxt, max_kl_step=0.01):
    runner = LocalTFRunner(ctxt)
    ...
    runner.train(...)
```

ryanjulian:
this looks great

[...]

Issue #1034, opened 2019-11-19
LocalRunner opens *many* plotter windows


ryanjulian:
Steps to reproduce:
* Open `examples/tf/trpo_cartpole.py`
* Set `plot=True` on both `runner.train` and `run_experiment`
* Run the example
* Watch as a new plotter window open after each episode. The inactive windows also appear to be orphan processes.

zequnyu:
I'll look into it.

ryanjulian:
can you also look at #354 while you're at it?

PR #1100, opened 2019-12-17
Wrap experiment

krzentner:
This implements a minimal version of the interface described in #1049.
It has a few limitations and doesn't implement every feature we might
need, but it's already probably better than run_experiment.

I believe this implementation can be cleanly extended to provide
everything we want from the interface.

The most important difference (besides all the new features) is that it
doesn't call garage.deterministic.set_seed. Since this is a parameter
that should be logged in variant.json, it should actually be passed into
the experiment function. Perhaps the runner should call this function
instead?

Since I want comments, I'm pushing this up before updating the examples.

### DIFF
--- a/src/garage/__init__.py
+++ b/src/garage/__init__.py
@@ -2,5 +2,6 @@
 from garage._dtypes import TimeStep
 from garage._dtypes import TrajectoryBatch
 from garage._functions import log_performance
+from garage.experiment.experiment import wrap_experiment
 
-__all__ = ['TimeStep', 'TrajectoryBatch', 'log_performance']
+__all__ = ['wrap_experiment', 'TimeStep', 'TrajectoryBatch', 'log_performance']
--- a/src/garage/experiment/experiment.py
+++ b/src/garage/experiment/experiment.py
@@ -1,158 +1,21 @@
+"""Tools for running experiments with Garage."""
 # flake8: noqa
 import base64
-import collections
 import datetime
-import inspect
+import enum
+import functools
+import json
 import os
 import os.path as osp
+import pathlib
 import pickle
 import re
 import subprocess
 
 import cloudpickle
+import dowel
+from dowel import logger
 import dateutil.tz
-import numpy as np
-
-
-class AttrDict(dict):
-
-    def __init__(self, *args, **kwargs):
-        super(AttrDict, self).__init__(*args, **kwargs)
-        self.__dict__ = self
-
-
-class VariantDict(AttrDict):
-
-    def __init__(self, d, hidden_keys):
-        super(VariantDict, self).__init__(d)
-        self._hidden_keys = hidden_keys
-
-    def dump(self):
-        return {k: v for k, v in self.items() if k not in self._hidden_keys}
-
-
-class VariantGenerator:
-    """
-    Usage:
-
-    | vg = VariantGenerator()
-    | vg.add("param1", [1, 2, 3])
-    | vg.add("param2", ['x', 'y'])
-    | vg.variants() => # all combinations of [1,2,3] x ['x','y']
-
-    Supports noncyclic dependency among parameters:
-    | vg = VariantGenerator()
-    | vg.add("param1", [1, 2, 3])
-    | vg.add("param2", lambda param1: [param1+1, param1+2])
-    | vg.variants() => # ..
-    """
-
-    def __init__(self):
-        self._variants = []
-        self._populate_variants()
-        self._hidden_keys = []
-        for k, vs, cfg in self._variants:
-            if cfg.get('hide', False):
-                self._hidden_keys.append(k)
-
-    def add(self, key, vals, **kwargs):
-        self._variants.append((key, vals, kwargs))
-
-    def _populate_variants(self):
-        methods = inspect.getmembers(
-            self.__class__,
-            predicate=lambda x: inspect.isfunction(x) or inspect.ismethod(x))
-        methods = [
-            x[1].__get__(self, self.__class__) for x in methods
-            if getattr(x[1], '__is_variant', False)
-        ]
-        for m in methods:
-            self.add(m.__name__, m, **getattr(m, '__variant_config', dict()))
-
-    def variants(self, randomized=False):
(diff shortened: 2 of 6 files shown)


src/garage/experiment/experiment.py
```diff
@@ -318,3 +201,279 @@ def to_local_command(params,
...
+        Args:
+            o (object): Object to encode.
+
+        Returns:
+            str: Object encoded in JSON.
+
+        """
+
+        # Why is this method hidden? What does that mean?
+        # pylint: disable=method-hidden
```
ryanjulian: i think it means that this method is overriding some symbol which is not its same kind (maybe a classmethod or property?)

src/garage/tf/plotter/plotter.py
```diff
@@ -47,9 +63,12 @@ def __init__(self,
...
         # Needed in order to draw glfw window on the main thread
-        if ('Darwin' in platform.platform()):
-            self.rollout(
-                env, policy, max_path_length=np.inf, animated=True, speedup=5)
+        if 'Darwin' in platform.platform():
+            self.rollout(env,
+                         policy,
+                         max_path_length=np.inf,
+                         animated=True,
+                         speedup=5)
```
zequnyu: Does this solve #1034?

src/garage/__init__.py
```diff
@@ -1,4 +1,5 @@
 """Garage Base."""
 from garage._dtypes import TrajectoryBatch
+from garage.experiment.experiment import wrap_experiment
 
-__all__ = ['TrajectoryBatch']
+__all__ = ['wrap_experiment', 'TrajectoryBatch']
```
ryanjulian: hmm we should talk about if we want this in the `garage` package or the `garage.experiment` package -- it seems confusing to export it in both.

krzentner: I generally think we should put it in the `garage` package. I'm not completely certain what the function should be called though.

src/garage/experiment/experiment.py
krzentner: Maybe. I didn't actually modify this code, I just moved it from `experiment_wrapper.py`.

krzentner: It's completely possible that this code doesn't actually work. Certainly no one has written the corresponding parsing code.

tests/garage/experiment/test_experiment.py
```diff
@@ -82,3 +50,45 @@ def test_experiment_with_variant():
     assert len(folder_content_diff) == 1
     exp_folder_name = folder_content_diff.pop()
     assert exp_folder_name.startswith('test_prefix')
+
+
+# Pylint gets confused by @wrap_experiment adding an argument.
+# pylint: disable=no-value-for-parameter
```
krzentner: @ryanjulian This is actually a design point we need to discuss. `@wrap_experiment` injects an argument, which pylint has no clue how to handle.

krzentner: For now (in the examples), I've given the context a default value of None.

krzentner:
I've updated a few of the examples to use `wrap_experiment`. I definitely think we need to do something about the RNG seed, but it can wait for another PR.

tests/garage/experiment/test_experiment.py
ryanjulian: so pylint will complain for every usage of wrap_experiment?

src/garage/experiment/experiment.py
```diff
@@ -1,158 +1,21 @@
+"""Tools for running experiments with Garage."""
```
ryanjulian: this file contains so much nonsense which we're trying to delete. can you put your new stuff into a different module so it's clear what's deprecated and what's not?

i'm even fine with you renaming this `experiment_deprecated` or whatever

src/garage/experiment/experiment.py
```diff
@@ -318,3 +201,279 @@ def to_local_command(params,
...
+            # Apply ourselves as a decorator
+            self.function = args[0]
+            functools.update_wrapper(self, self.function)
+            return self
+        else:
+            ctxt = self._make_context(*args, **kwargs)
+            return self.function(ctxt, *args, **kwargs)
+
+
+def wrap_experiment(function=None,
```
ryanjulian: in the spirit of most python decorators, shall we just call this `@experiment`?

```python
from garage.experiment import experiment

@experiment
def my_cool_exp(ctxt):
   # cool stuff here

@experiment
def my_cool_exp2(ctxt):
    # even cooler stuff here
```

src/garage/tf/plotter/plotter.py
```diff
@@ -1,3 +1,4 @@
+"""Renders rollouts of the policy as it trains."""
```
ryanjulian: i have a mental TODO to move this and garage.plotter into garare.experiment, but does not need to be done here.

src/garage/experiment/experiment.py
krzentner: I would like to, but moving the existing `garage.experiment` module would take significant work right now.

tests/garage/experiment/test_experiment.py
krzentner: Yes. It's quite annoying. I'm not quite sure what the best solution is.

src/garage/experiment/experiment.py
krzentner: That would be good to do at some point soon. Especially once people have gotten used to `wrap_experiment`.

tests/garage/experiment/test_experiment.py
krzentner: For now, I've made the `ctxt` parameter "optional", which is sufficient to make pylint happy. I appear to have deleted all of the example updates when I rebased this PR, so I'll push up changes to examples in a new PR.

krzentner:
> can this PR include at least one usage example for examples/?

Oh, I actually didn't see your new comments and submitted this. I had a changed that updated a bunch of the examples, but I somehow dropped it from the branch when I rebased. I'll push up a new PR with the change, once I verify the examples still work.

PR #1101, opened 2019-12-17
MAML Implementation in PyTorch

utkarshjp7:
Here is the benchmark against ProMP implementation for `HalfCheetahVel` env.

HalfCheetah-v2_benchmark

TODO:
- [ ] Add code for meta-testing.
- [ ] Run benchmarks with metaworld tasks.
- [ ] Run benchmarks with HalfCheetahDir env and ant envs. These two envs can be found here https://github.com/tristandeleu/pytorch-maml-rl/tree/master/maml_rl/envs/mujoco
- [ ] Make CI pass.


src/garage/_dtypes.py
```diff
@@ -167,3 +167,14 @@ def __new__(cls, env_spec, observations, actions, rewards, terminals,
         return super().__new__(TrajectoryBatch, env_spec, observations,
                                actions, rewards, terminals, env_infos,
                                agent_infos, lengths)
+
+class TorchTrajectoryBatch(
```
ryanjulian: wait, how does this differ other than having no validation?

src/garage/torch/algos/maml.py
```diff
@@ -0,0 +1,383 @@
...
+from garage.torch.algos import _Default, make_optimizer, VPG
+from garage.torch.optimizers import DiffSGD
+from garage.torch.utils import update_module_params
+
+
+class MAML:
+    """Model-Agnostic Meta-Learning (MAML).
+
+    Args:
+        env (garage.envs.base): A gym environment.
```
ryanjulian: garage.envs.base is a package. I think you want GarageEnv

src/garage/torch/algos/maml.py
```diff
@@ -0,0 +1,383 @@
...
+        policy (garage.torch.policies.base.Policy): Policy.
+        baseline (garage.np.baselines.Baseline): The baseline.
+        meta_optimizer (Union[type, tuple[type, dict]]): Type of optimizer.
+            This can be an optimizer type such as `torch.optim.Adam` or a
+            tuple of type and dictionary, where dictionary contains arguments
+            to initialize the optimizer e.g. `(torch.optim.Adam, {'lr' = 1e-3})`  # noqa: E501
+        meta_batch_size (int): Number of tasks.
```
ryanjulian: this requires more explanation. it's not the number of tasks, it's the number of tasks sampled per batch. it's possible for the number of tasks to be greater or less than this number.

src/garage/torch/algos/maml.py
```diff
@@ -0,0 +1,383 @@
...
+        inner_lr (double): Adaptation learning rate.
+        num_grad_updates (int): Number of adaptation gradient steps.
+        inner_algo (garage.torch.algos.vpg): The inner algorithm used for
```
ryanjulian: vpg is a package, VPG is an algorithm.

src/garage/torch/algos/maml.py
```diff
@@ -0,0 +1,383 @@
...
+                 inner_algo=None):
+        if inner_algo is None:
+            inner_algo = VPG(env.spec, policy, baseline)
+
+        if policy.vectorized:
+            self.sampler_cls = OnPolicyVectorizedSampler
+        else:
+            self.sampler_cls = BatchSampler
+
+        self.policy = policy
```
ryanjulian: please make all of these private

src/garage/torch/algos/maml.py
```diff
@@ -0,0 +1,383 @@
...
+                [meta_batch_size * (num_grad_updates + 1)]
+            all_params (list[dict]): A list of named parameter dictionaries.
+                Each dictionary contains key value pair of names (str) and
+                parameters (torch.Tensor).
+
+        Returns:
+            float: Average return.
+
+        """
+        kl_before = self._compute_kl_constraint(itr,
```
ryanjulian: does this still work if i use PPO instead?

src/garage/torch/algos/maml.py
```diff
@@ -0,0 +1,383 @@
...
+            self._inner_algo.compute_policy_entropy(samples.observations)
+            for samples in task_samples
+        ]
+        return torch.stack(entropies).mean()
+
+    def _set_task(self, runner, task):
+        for env in runner._sampler._vec_env.envs:
+            env.set_task(task)
+
+    @property
```
ryanjulian: hmm there's not much reason to have a private property. just use the attribute directly.

src/garage/torch/algos/vpg.py
```diff
@@ -157,28 +157,26 @@ def train_once(self, itr, paths):
...
-            obs (torch.Tensor): Observation from the environment.
-            actions (torch.Tensor): Predicted action.
-            rewards (torch.Tensor): Feedback from the environment.
+            batch_samples (garage.TorchTrajectoryBatch): Samples data.
 
         Returns:
             torch.Tensor: Calculated mean value of loss
 
         """
         # pylint: disable=unused-argument
```
ryanjulian: please just `del` the unused argument instead

src/garage/torch/modules/gaussian_mlp_module.py
```diff
@@ -144,8 +145,7 @@ def forward(self, inputs):
         else:
             std = log_std_uncentered.exp().exp().add(1.).log()
 
-        cov = (std**2).diag_embed()
-        dist = MultivariateNormal(mean, cov)
+        dist = Independent(Normal(mean, std), 1)
```
ryanjulian: can you make this fix its own PR instead?

src/garage/torch/optimizers/differentiable_sgd.py
```diff
@@ -0,0 +1,58 @@
+"""Differentiable Stochastic Gradient Descent Optimizer.
```
ryanjulian: can you submit this+tests as a separate PR?

zhanpenghe:
To get a better sense of the performance, can you show more information of the training?
For example, discounted returned before and after one/several steps adaptation. 

I will take a closer look on the codes later.

src/garage/torch/algos/maml.py
```diff
@@ -0,0 +1,383 @@
...
+            paths (list[dict]): A list of collected paths
+
+        Returns:
+            garage.TorchTrajectoryBatch: Processed samples data.
+
+        """
+        for path in paths:
+            path['returns'] = tensor_utils.discount_cumsum(
+                path['rewards'], self._inner_algo.discount)
+        self.baseline.fit(paths)
```
ryanjulian: fitting here seems dangerous to me. you're modifying state in a method which claims to only process samples. based on this signature, i should be able to call this method repeatedly with no ill effects (not the case).

i think you should move fitting the baseline to the loop above this one.

src/garage/_dtypes.py
ryanjulian: differ from TrajectoryBatch, that is

ryanjulian: is it to add baselines?

The only point in adding a new type is to define an interface between two modules. `TrajectoryBatch` defines the data interface between samplers and algorithms. `baselines` is internal to policy gradient algorithms.

I'm not sure what structure makes sense here, but you should first try just passing baselines as its own argument to the functions you need. If that doesn't work, I suggest you define a new dtype, perhaps in torch/VPG.

ryanjulian: also, whatever you do with this, please break it into a separate PR

src/garage/envs/__init__.py
```diff
@@ -6,6 +6,7 @@
 from garage.envs.grid_world_env import GridWorldEnv
 from garage.envs.normalized_env import normalize
 from garage.envs.point_env import PointEnv
+from garage.envs.half_cheetah_vel import HalfCheetahVelEnv
```
ryanjulian: can you break this env into a separate PR?

src/garage/torch/algos/vpg.py
```diff
@@ -7,6 +7,7 @@
 import torch
 import torch.nn.functional as F
 
+from garage import TorchTrajectoryBatch
```
ryanjulian: please make any updates to this algo their own PR. it will make QA and code review immensely easier.

src/garage/torch/utils.py
```diff
@@ -44,3 +44,20 @@ def flatten_batch(tensor):
 
     """
     return tensor.reshape((-1, ) + tensor.shape[2:])
+
+
+def update_module_params(module, new_params):
```
ryanjulian: @krzentner can you comment on the torchi-ness of this approach? who uses it?

if this is never intended to be used by a package user, please make sure it's an _private function.

src/garage/torch/algos/maml.py
ryanjulian: perhaps you can define an ABC for the algorithms MAML can support? I think that's better than just asserting that all VPG child classes can use MAML. VPG is really written as an algorithm, not an interface.

src/garage/torch/algos/maml.py
```diff
@@ -0,0 +1,383 @@
...
+    def __init__(self,
+                 env,
+                 policy,
+                 baseline,
+                 meta_optimizer=torch.optim.Adam,
+                 meta_batch_size=40,
+                 inner_lr=0.1,
+                 num_grad_updates=1,
+                 inner_algo=None):
+        if inner_algo is None:
```
ryanjulian: this is a dangerous default to me. most people don't want to run MAML with VPG. just make this a required argument.

src/garage/torch/algos/maml.py
```diff
@@ -0,0 +1,383 @@
...
+            runner.step_itr += 1
+
+        return last_return
+
+    def train_once(self, itr, all_samples, all_params):
+        """Train the algorithm once.
+
+        Args:
+            itr (int): Iteration number.
+            all_samples (list[list[garage.TorchTrajectoryBatch]]): Two
```
ryanjulian: i think seeing `list[list[BatchDataType]]` is a sign that your batch data type is not very useful. what are you trying to encode here? does it actually need a new type? why is there a list of lists, could it be flat instead?

src/garage/torch/algos/maml.py
```diff
@@ -0,0 +1,383 @@
...
+            garage.TorchTrajectoryBatch: Processed samples data.
+
+        """
+        for path in paths:
+            path['returns'] = tensor_utils.discount_cumsum(
+                path['rewards'], self._inner_algo.discount)
+        self.baseline.fit(paths)
+        return self._inner_algo.process_samples(itr, paths)
+
+    def _log(self, itr, all_samples, loss_before, loss_after, kl_before, kl,
```
ryanjulian: i think this needlessly obfuscates logging, which is most useful in-line with code. logs are comments which happen to produce output.

i think you can split this into (1) a few well-place logger calls throughout the code and (2) one or more calls to a uniform evaluation function in RLAlgorithm (consult Zequn for info on (2)). If Zequn is not done by the time you commit this, you can include the evaluation part here, but he will have to refactor.

src/garage/torch/algos/vpg.py
```diff
@@ -208,7 +206,7 @@ def _compute_loss(self, itr, paths, valids, obs, actions, rewards):
         valid_objectives = filter_valids(objective, valids)
         return -torch.cat(valid_objectives).mean()
 
-    def _compute_kl_constraint(self, obs):
+    def compute_kl_constraint(self, obs):
```
ryanjulian: this name is not accurate, because not all policy gradient algorithms have a KL constraint.

src/garage/torch/algos/vpg.py
```diff
@@ -232,7 +230,7 @@ def _compute_kl_constraint(self, obs):
 
         return kl_constraint.mean()
 
-    def _compute_policy_entropy(self, obs):
+    def compute_policy_entropy(self, obs):
```
ryanjulian: will calling this method at an arbitrary time cause the gradients to be wrong if self._stop_entropy_gradient is false? if so, you can't make this method public.

src/garage/torch/algos/vpg.py
```diff
@@ -289,7 +287,7 @@ def _get_baselines(self, path):
             return torch.Tensor(self.baseline.predict_n(path))
         return torch.Tensor(self.baseline.predict(path))
 
-    def _optimize(self, itr, paths, valids, obs, actions, rewards):  # pylint: disable=unused-argument  # noqa: E501
+    def _optimize(self, itr, batch_samples):  # pylint: disable=unused-argument  # noqa: E501
```
ryanjulian: just delete the unused arugment

src/garage/torch/utils.py
krzentner: People almost never do this. Except for the weird case were you need your parameters to be in the middle of the tape, you can just use `module.load_state_dict()`.

krzentner: Having said that, this function does basically do the same thing as `load_state_dict()` (with the minor different that it doesn't erase all of the parameter `grad` / `grad_fn`), so perhaps we should consider it fairly torchy?

src/garage/_dtypes.py
krzentner: I think the idea is that this contains `torch.FloatTensor` instead of `np.ndarray`. It seems like a reasonable thing to have, but probably belongs in `garage.torch`.

src/garage/torch/algos/maml.py
```diff
@@ -0,0 +1,383 @@
...
+    """
+
+    def __init__(self,
+                 env,
+                 policy,
+                 baseline,
+                 meta_optimizer=torch.optim.Adam,
+                 meta_batch_size=40,
+                 inner_lr=0.1,
+                 num_grad_updates=1,
```
zhanpenghe: `num_of_inner_steps`?

src/garage/torch/algos/maml.py
```diff
@@ -0,0 +1,383 @@
...
+            self._set_task(runner, task)
+
+            for j in range(self._num_grad_updates + 1):
+                paths = runner.obtain_samples(runner.step_itr)
+                batch_samples = self._process_samples(runner.step_itr, paths)
+                all_samples[i].append(batch_samples)
+
+                # Skip last iteration
+                if j != self._num_grad_updates:
+                    self._adapt(runner.step_itr, batch_samples, set_grad=False)
```
zhanpenghe: Please correct me if I am wrong:
since `set_grad` is set to be `False`, these gradient is actually never applied to the policy during the sampling process? So in this case, each sampling step is using the same policy.

src/garage/envs/__init__.py
krzentner: I'm working on a PR to add this (and `HalfCheetahDirEnv`).

naeioi:
I tested with HalfCheetahDir for one trial. Both ProMP and garage use the HalfCheetahDir in garage. While garage does learn, it learns much slower than ProMP. 

image
Also notice that garage starts with a large negative reward, while ProMP starts around 0 reward. I am investigating on this.

naeioi:
Update: ProMP scales the expected policy output range to [-10,10]. This results in a small initial average return. If I add this to garage, the initial return agrees with ProMP. See https://github.com/jonasrothfuss/ProMP/blob/93ae339e23dfc6e1133f9538f2c7cc0ccee89d19/meta_policy_search/envs/normalized_env.py#L113.

But this is not enough. ProMP usually takes off after 10 batches on HalfCheetahDir. garage's learning curve is much flatter. I suspect this has something to do with the optimizer.

ryanjulian:
Could it be a difference with the baseline? Did you see Anson's thread about when ProMP updates the baseline vs garage (before optimization vs after)

naeioi:
@ryanjulian A quick glimpse shows me that the baseline is already fitted before prediction. But I need a more careful inspection on this.

naeioi:
Update: Results using tanh in garage as the hidden activation function in MLP. (ReLU, the default, was used in the last experiment). 

image

ProMP also uses tanh but not ReLU. 

The last experiment shows that garage learns but reaches a plateau of performance (return stops growing at around 50). I guess this is because after feeding MLP with some amount of data (1e7 steps shown by the last figure)
- a significant number of neurons with negative output will die out because of ReLU's property
- the rest of the network is not expressive enough to fit the data (underfitting)
- tanh does not have this 'gradient dying' issue

The original MAML implementation uses ReLU, but it also uses a larger network, so it helps to mitigate underfitting. ([100, 100] hidden size. Larger than [64, 64] used in garage and ProMP)

ryanjulian:
This is excellent work! Yes typically RL uses tanh for these reasons. This performance is getting close but still a little less than ProMP. You probably should check the seed sensitivity by running a several and checking the confidence interval. 

One thing which strikes me is that the garage version is higher-variance. This is usually a result of high learning rate or a baseline which is not as good. It is worth it to take a close look at the baselines in the two versions and see if there are discrepancies.

naeioi:
Update: MAML in garage now has a performance matching ProMP. In this result, both garage and ProMP fit baseline before prediction. I will test fitting baseline after prediction tommorrow. 

image

ryanjulian:
Awesome work! I look forward to the results. Based on Anson's experiments, the difference before fitting the baseline before vs after may go away as you collect more trajectories per task.

naeioi:
Update: Result of fit-baseline-after-prediction. Both ProMP and garage suffer from a severe performance downgrade.

image

In the above experiment, I use an independent baseline for each environment. If a single baseline is used, the performance is much better than independent-baseline, but still worse than fit-first. Anson's experiments with fit-later in RL^2 use a shared baseline, and the results (worse than fit-first) agree with MAML. 

image

Note: 40 paths are sampled for each environment, which is pretty large, but I don't see it helps with the performance.

src/garage/torch/algos/maml.py
naeioi: This is part of TRPO optimizer. PPO should work, but no literature has ever showed a result of PPO.

src/garage/torch/algos/maml.py
naeioi: I think there is a misunderstanding here. The `inner_algo` is used only to compute the inner loss (inner KL is also computed, but only for logging purpose), so VPG and PPO should be the only two valid algorithms. The variant usually goes into meta optimization (e.g. the TRPO in ProMP-TRPO refers to the meta optimization).

PR #1103, opened 2019-12-20
Add a special Docker build stage to nightlies

ryanjulian:
Adds a special stage to nightly builds, which recreates the CI docker
container from scratch, rather than building it from the cache. It then
pushes the from-scratch container to Docker Hub for future build
jobs. This helps avoid missing dependency-related breakages due to using
very old cached versions of dependencies.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,8 +16,18 @@ env:
 
 jobs:
   include:
+    - stage: "Build"
+      if: type = cron AND branch = master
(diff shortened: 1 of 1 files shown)


ryanjulian:
This is only active for cron jobs, but you can see a sample build here: https://travis-ci.com/rlworkgroup/garage/builds/141848819

ryanjulian:
@Mergifyio rebase

PR #1104, opened 2019-12-22
Make policy stateless

ahtsan:
This PR makes recurrent policies stateless by decoupling agent info and policies.
Currently recurrent policies store the agent info, e.g. previous action and hidden state
which can be a problem if it's modified unexpectedly.

Proposed API: Call `policy.get_initial_state()` to retrieve the initial state and during sampling,
we pass the `agent_info` as an argument to both `policy.get_action()` and `policy.reset()`.
Consumer of policies will be responsible to modify the state given to policies, in this case
update the previous action in `agent_infos` before passing to policies, similar to update
`obses` with next observation.

Purpose of the first commit is only to demonstrate the API. When API is finalized, all policies will be applied with the same change.

An example is also added to demonstrate the policy is learning. Below is the average score for `GaussianGRUPolicy` in `InvertedDoublePendulum`.
Screenshot 2019-12-21 at 5 35 48 PM

### DIFF
--- a/examples/tf/trpo_pendulum_recurrent.py
+++ b/examples/tf/trpo_pendulum_recurrent.py
@@ -1,0 +1,56 @@
+#!/usr/bin/env python3
+"""This is an example to train a task with TRPO algorithm.
+
+It uses an LSTM-based recurrent policy.
+
+Here it runs CartPole-v1 environment with 100 iterations.
+
+Results:
+    AverageReturn: 100
+    RiseTime: itr 13
+"""
+from garage.experiment import run_experiment
+from garage.np.baselines import LinearFeatureBaseline
+from garage.tf.algos import TRPO
+from garage.tf.envs import TfEnv
+from garage.tf.experiment import LocalTFRunner
+from garage.tf.optimizers import ConjugateGradientOptimizer
+from garage.tf.optimizers import FiniteDifferenceHvp
+from garage.tf.policies import GaussianGRUPolicy
+
+
+def run_task(snapshot_config, *_):
+    """Defines the main experiment routine.
+
+    Args:
+        snapshot_config (garage.experiment.SnapshotConfig): Configuration
+            values for snapshotting.
+        *_ (object): Hyperparameters (unused).
+
+    """
+    with LocalTFRunner(snapshot_config=snapshot_config) as runner:
+        env = TfEnv(env_name='InvertedDoublePendulum-v2')
+
+        policy = GaussianGRUPolicy(name='policy', env_spec=env.spec)
+
+        baseline = LinearFeatureBaseline(env_spec=env.spec)
+
+        algo = TRPO(env_spec=env.spec,
+                    policy=policy,
+                    baseline=baseline,
+                    max_path_length=100,
+                    discount=0.99,
+                    max_kl_step=0.01,
+                    optimizer=ConjugateGradientOptimizer,
+                    optimizer_args=dict(hvp_approach=FiniteDifferenceHvp(
+                        base_eps=1e-5)))
+
+        runner.setup(algo, env)
+        runner.train(n_epochs=100, batch_size=1024)
+
+
+run_experiment(
+    run_task,
+    snapshot_mode='last',
+    seed=1,
+)
--- a/src/garage/np/policies/base.py
+++ b/src/garage/np/policies/base.py
@@ -8,24 +8,57 @@ class Policy(abc.ABC):
     Args:
         env_spec (garage.envs.env_spec.EnvSpec): Environment specification.
 
+    Note:
+        `policy_info` is passed as an argument in some APIs in Policy. This
+        is intended to allow policies to pass state to themselves between
+        calls. Users should consider its contents to be private, and should
+        never assume this variable contains anything in particular, or even
+        that its contents is consistent between calls to the same Policy
+        object. In most use cases for non-recurrent policies, `policy_info`
+        will be an empty dictionary while for recurrent-policies, it may
+        store its previous hidden/cell state as the policy state.
+
+        User should never mutate the contents of `policy_info` outside Policy,
+        and no policy should ever require the user inspect or mutate
+        `policy_info` as part of its interface.
+
+        Implementers of Policy may occasionally find it useful to store some
+        debugging information in `policy_info`, but any information
+        implementers would like to be consistently logged by users should be
+        exposed explicitly through the log_diagnostics interface.
+
     """
 
     def __init__(self, env_spec):
         self._env_spec = env_spec
 
     @abc.abstractmethod
-    def get_action(self, observation):
+    def get_action(self, observation, policy_info=None):
         """Get action sampled from the policy.
 
         Args:
             observation (np.ndarray): Observation from the environment.
+            policy_info (dict): Info for the policy.
 
         Returns:
-            (np.ndarray): Action sampled from the policy.
+            np.ndarray: Action sampled from the policy.
 
         """
 
-    def reset(self, dones=None):
+    @abc.abstractmethod
+    def get_actions(self, observations, policy_infos=None):
+        """Get actions sampled from the policy.
+
+        Args:
+            observations (list[np.ndarray]): Observations from the environment.
+            policy_infos (dict): Infos for the policy.
+
+        Returns:
+            np.ndarray: Actions sampled from the policy.
+
+        """
+
+    def reset(self, policy_infos, dones=None):
         """Reset the policy.
 
         If dones is None, it will be by default np.array([True]) which implies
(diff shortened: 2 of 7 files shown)


examples/tf/trpo_pendulum_recurrent.py
```diff
@@ -0,0 +1,56 @@
...
+
+        baseline = LinearFeatureBaseline(env_spec=env.spec)
+
+        algo = TRPO(env_spec=env.spec,
+                    policy=policy,
+                    baseline=baseline,
+                    max_path_length=100,
+                    discount=0.99,
+                    max_kl_step=0.01,
+                    optimizer=ConjugateGradientOptimizer,
```
ryanjulian: this is a nit, but TRPO examples should use ConjugateGradientOptimizer (and these optimizer_args) by default

src/garage/sampler/on_policy_vectorized_sampler.py
```diff
@@ -103,27 +103,31 @@ def obtain_samples(self, itr, batch_size=None, whole_paths=True):
 
         policy = self.algo.policy
 
+        agent_infos = policy.get_initial_state(dones)
```
ryanjulian: refresh my memory on why `dones` is a necessary part of this API?

src/garage/sampler/on_policy_vectorized_sampler.py
```diff
@@ -103,27 +103,31 @@ def obtain_samples(self, itr, batch_size=None, whole_paths=True):
...
         while n_samples < batch_size:
             t = time.time()
-            policy.reset(dones)
+            agent_infos = policy.reset(agent_infos, dones)
 
-            actions, agent_infos = policy.get_actions(obses)
+            actions, agent_infos = policy.get_actions(obses, agent_infos)
```
ryanjulian: please always call keyword arguments as keywords, i.e. `policy.get_actions(obses, agent_infos=agent_infos)` -- this helps avoid really hard to find refactoring bugs.

src/garage/sampler/on_policy_vectorized_sampler.py
```diff
@@ -153,6 +157,9 @@ def obtain_samples(self, itr, batch_size=None, whole_paths=True):
 
             process_time += time.time() - t
             pbar.inc(len(obses))
+
+            if policy.state_include_action:
```
ryanjulian: this is a terrible interface. only `Policy` should be mutating `agent_infos`, i.e. to the user of this API, `agent_info` should be a completely opaque data blob.

src/garage/tf/policies/base.py
```diff
@@ -23,41 +23,68 @@ def __init__(self, name, env_spec):
         self._cached_param_shapes = None
 
     @abc.abstractmethod
-    def get_action(self, observation):
+    def get_action(self, observation, agent_infos):
```
ryanjulian: i think this should be a kwarg with default `None`. recurrent policies can raise a `ValueError` if it's not specified, but there's no reason to force everyone to pass something for it.

src/garage/tf/policies/base.py
```diff
@@ -23,41 +23,68 @@ def __init__(self, name, env_spec):
...
+            agent_infos (dict): Infos for previous action and hidden states.
 
         Returns:
             (np.ndarray): Action sampled from the policy.
 
         """
 
     @abc.abstractmethod
-    def get_actions(self, observations):
+    def get_actions(self, observations, agent_infos):
```
ryanjulian: please make this a kwarg.

src/garage/tf/policies/base.py
ryanjulian: we should take this opportunity to make sure `agent_infos` is the right name for this stuff.

i suggest you rename it to `policy_info`(s) because we don't have anything in `garage` named `Agent` and probably won't in the future. if someone comes across `policy_info` in a debugger they have a good chance of guessing it came from a `Policy` but if they come across `agent_info` they won't find anything named `agent` which is unnecessarily obtuse.

please leave a note in the policy interface which makes it clear the purpose of `policy_info` and what people should and should not do with it. basically it will be something like


Note:
`policy_info` is intended to allow policies to pass state to themselves between calls to `get_actions`. Users should consider its contents to be private, and should never assume this variable contains anything in particular, or even that its contents is consistent between calls to the same `Policy` object. User should *never* mutate the contents of `policy_info` and no `Policy` should ever require the user inspect or mutate `policy_info` as part of its interface. Implementers of `Policy` may occasionally find it useful to store some debugging information in `policy_info`, but any information implementers would like to be consistently logged by users should be exposed explicitly through the `get_diagnostics` interface.

src/garage/tf/policies/base.py
```diff
@@ -23,41 +23,68 @@ def __init__(self, name, env_spec):
...
             dones (numpy.ndarray): Bool that indicates terminal state(s).
 
         """
 
+    def get_initial_state(self, dones=None):
+        """Initial state.
+
+        Note:
+            If `dones` is None, it will be by default `np.array([True])` which
+            implies the policy will not be "vectorized", i.e. number of
```
ryanjulian: hmm i think that instead `get_action` should also be stateless with respect to the vectorization. it seems here that the shape of `dones` implies that `get_actions` will be called only with shape `(P)` in the future, but I think this is an artificial constraint.

it's pretty trivial to require vectorized users to tile `initial_state` into a shape which matches `observations`.

a consistent way to implement this to add a function `get_initial_states` to this interface and consider that the default way it will be used. the relationship between `get_initial_state` and `get_initial_states` is the same as the relationship between `get_action` and `get_actions`: `get_initial_states` is fully-vectorized and returns the initial state tensor given some prototype array to determine the batch size, `get_initial_state` can always be implemented as something like `self.get_initial_states(batch_size)[0]`

src/garage/sampler/on_policy_vectorized_sampler.py
ahtsan: so that policy can obtain the number of parallel environments. We can actually explicitly pass that rather than dones.

src/garage/sampler/on_policy_vectorized_sampler.py
ahtsan: originally I am passing it as `next_action` in`agent_infos` so policy can use that as `prev_action` in next iteration. Would that be better?

src/garage/tf/policies/base.py
```diff
@@ -23,41 +23,68 @@ def __init__(self, name, env_spec):
...
+            parallel environments for training data sampling = 1.
+
+        Args:
+            dones (list[bool]): Terminal signals with shape :math:`(P)`, where
+                P is the number of parallel environments for training data
+                sampling.
+
+        Returns:
+            dict: Initial state. For non-recurrent policies, it will be an
```
ryanjulian: this API just returns an initial `policy_info` object, correct? In that case, you can include detailed information on the contents of `policy_info` in notes in the individual classes, but don't include it here. we can't document every primitive which implements this interface in the base class.

src/garage/sampler/on_policy_vectorized_sampler.py
ryanjulian: see my other comments in this PR. i don't think `dones` is the right way to do this.

src/garage/sampler/on_policy_vectorized_sampler.py
ryanjulian: yes. the policy can put whatever it wants in `agent_infos` and if it needs information from the past, that's where you put it.

src/garage/tf/policies/base.py
ahtsan: `...given some prototype array to determine the batch size...`
How about directly passing the batch size? like
```python
    def get_initial_state(self):
        return self.get_initial_states(batch_size=1)

    def get_initial_states(self, batch_size):
        initial_hidden = self.model.networks['default'].init_hidden.eval()
        prev_actions = np.zeros((batch_size, self.action_space.flat_dim))
        prev_hiddens = np.full((batch_size, self._hidden_dim), initial_hidden)

        return dict(prev_action=prev_actions, prev_hidden=prev_hiddens)
```

ryanjulian: that seems reasonable to me. are you sure about the implementation of `get_initial_state`? don't you need `self.get_initial_states(batch_size=1)[0]` or `np.squeeze(self.get_initial_states(batch_size=1)`?

src/garage/sampler/on_policy_vectorized_sampler.py
ahtsan: fixed

src/garage/tf/policies/base.py
ahtsan: fixed

src/garage/tf/policies/base.py
ahtsan: fixed

src/garage/tf/policies/base.py
ahtsan: fixed

src/garage/tf/policies/base.py
ahtsan: I think I need to do `{k: v[0] for k, v in state.items()}`

src/garage/tf/policies/base.py
ahtsan: fixed

ahtsan:
I moved the interface to `garage.np.Policy` and have `garage.tf.Policy` inherits `garage.np.Policy`. Should `garage.torch.Policy` also inherit `garage.np.Policy`? Otherwise the interface is not being enforced by the base class.

src/garage/np/policies/base.py
```diff
@@ -57,6 +121,28 @@ def recurrent(self):
         """
         return False
 
+    @property
+    def vectorized(self):
```
krzentner: I thought we were getting rid of this method and making everything vectorized?

PR #1106, opened 2019-12-24
Needs review Sample vs Sample batch

avnishn:
This isn't a complete pr (needs some further error checking and unit tests). 

For off-policy algorithms, we need to support different return dtypes. 

I propose 2 types.

Either we have a dtype `SampleBatch` which is a tuple containing the necessary values for off policy, and each value of the tuple is a dict or nd array where each element of the dict or ndarray belongs to a specific transition

OR

We have a dtype `Sample` which is a tuple and each element of the tuple is some info about the transition (single observation, single action, single reward, single agent_env info) and then we can have the user return a sample or a list of samples from the sampler. 

Which design works better in the context of garage?
I think both work, but the second is better because it ends up being more versatile. Furthermore its less ambiguous if all of the components of a single sample are grouped together, as opposed to the first design where each element of the tuple is an array of all of some element (e.g an array of all the rewards) and each index of that array corresponds to an individual sample.

### DIFF
--- a/src/garage/_dtypes.py
+++ b/src/garage/_dtypes.py
@@ -167,3 +167,199 @@ class TrajectoryBatch(
         return super().__new__(TrajectoryBatch, env_spec, observations,
                                actions, rewards, terminals, env_infos,
                                agent_infos, lengths)
+
+class SamplesBatch(
+        collections.namedtuple('SamplesBatch', [
+            'env_spec',
+            'observations',
+            'actions',
+            'rewards',
+            'next_observations',
+            'terminals',
+            'env_infos',
+            'agent_infos',
+        ])):
+    # pylint: disable=missing-return-doc, missing-return-type-doc, missing-param-doc, missing-type-doc  # noqa: E501
+    r"""A tuple representing a batch of whole trajectories.
+
+    A :class:`SamplesBatch` represents a batch of samples
+    produced when one or more agents interacts with one or more environments.
+
+    Attributes:
+        env_spec (garage.envs.EnvSpec): Specification for the environment from
+            which this data was sampled.
+        observations (numpy.ndarray): A numpy array of shape
+            :math:`(N \bullet [T], O^*)` containing the (possibly
+            multi-dimensional) observations for all time steps in this batch.
+            These must conform to :obj:`env_spec.observation_space`.
+        actions (numpy.ndarray): A  numpy array of shape
+            :math:`(N \bullet [T], A^*)` containing the (possibly
+            multi-dimensional) actions for all time steps in this batch. These
+            must conform to :obj:`env_spec.action_space`.
+        rewards (numpy.ndarray): A numpy array of shape
+            :math:`(N \bullet [T])` containing the rewards for all time steps
+            in this batch.
+        terminals (numpy.ndarray): A boolean numpy array of shape
+            :math:`(N \bullet [T])` containing the termination signals for all
+            time steps in this batch.
+        env_infos (dict): A dict of numpy arrays arbitrary environment state
+            information. Each value of this dict should be a numpy array of
+            shape :math:`(N \bullet [T])` or :math:`(N \bullet [T], S^*)`.
+        agent_infos (numpy.ndarray): A dict of numpy arrays arbitrary agent
+            state information. Each value of this dict should be a numpy array
+            of shape :math:`(N \bullet [T])` or :math:`(N \bullet [T], S^*)`.
+            For example, this may contain the hidden states from an RNN policy.
+
+    Raises:
+        ValueError: If any of the above attributes do not conform to their
+            prescribed types and shapes.
+
+    """
+    __slots__ = ()
+
+    def __new__(cls, env_spec, observations, actions, rewards, next_observations,
+                 terminals, env_infos, agent_infos):  # noqa: D102
+        # pylint: disable=too-many-branches
+
+        first_observation = observations[0]
+        first_next_observation =  next_observation[0]
+        first_action = actions[0]
+
+        # observations
+        if not env_spec.observation_space.contains(first_observation):
+            raise ValueError(
+                'observations must conform to observation_space {}, but got '
+                'data with shape {} instead.'.format(
+                    env_spec.observation_space, first_observation))
+
+        if not env_spec.observation_space.contains(first_next_observation):
+            raise ValueError(
+                'next_observations must conform to observation_space {}, but got '
+                'data with shape {} instead.'.format(
+                    env_spec.observation_space, first_next_observation))
+
+        # actions
+        # TODO: Shape check can be removed for gym>=0.15.4 # pylint: disable=fixme # noqa: E501
+        if not (env_spec.action_space.contains(first_action)
+                and first_action.shape == env_spec.action_space.shape):
+            raise ValueError(
+                'actions must conform to action_space {}, but got data with '
+                'shape {} instead.'.format(env_spec.action_space,
+                                           first_action))
+
+        if terminals.dtype != np.bool:
+            raise ValueError(
+                'Terminals tensor must be dtype np.bool, but got tensor '
+                'of dtype {} instead.'.format(terminals.dtype))
+
+        # env_infos
+        for key, val in env_infos.items():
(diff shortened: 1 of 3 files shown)


src/garage/_dtypes.py
```diff
@@ -167,3 +167,199 @@ def __new__(cls, env_spec, observations, actions, rewards, terminals,
...
+            'observation',
+            'action',
+            'reward',
+            'next_observation',
+            'terminal',
+            'env_info',
+            'agent_info',
+        ])):
+    # pylint: disable=missing-return-doc, missing-return-type-doc, missing-param-doc, missing-type-doc  # noqa: E501
+    r"""A tuple representing a batch of whole trajectories.
```
ahtsan: it's only representing a single sample?

src/garage/_dtypes.py
```diff
@@ -167,3 +167,199 @@ def __new__(cls, env_spec, observations, actions, rewards, terminals,
...
+
+    A :class:`Sample` represents single sample sampled from the environment.
+
+    Attributes:
+        env_spec (garage.envs.EnvSpec): Specification for the environment from
+            which this data was sampled.
+        observation (float): Observation at the current time step in the
```
ahtsan: it's float only if dimension of observation space is 1?

src/garage/_dtypes.py
```diff
@@ -167,3 +167,199 @@ def __new__(cls, env_spec, observations, actions, rewards, terminals,
...
+            environment.
+        action (float): Action sampled from the policy given the observation
```
ahtsan: it's float only if dimension of action space is 1?

avnishn: ah you're right

src/garage/_dtypes.py
```diff
@@ -167,3 +167,199 @@ def __new__(cls, env_spec, observations, actions, rewards, terminals,
...
+            'observations',
+            'actions',
+            'rewards',
+            'next_observations',
+            'terminals',
+            'env_infos',
+            'agent_infos',
+        ])):
+    # pylint: disable=missing-return-doc, missing-return-type-doc, missing-param-doc, missing-type-doc  # noqa: E501
+    r"""A tuple representing a batch of whole trajectories.
```
krzentner: I think this is incorrect?

src/garage/_dtypes.py
```diff
@@ -167,3 +167,199 @@ def __new__(cls, env_spec, observations, actions, rewards, terminals,
...
+                    env_spec.observation_space, first_observation))
+
+        if not env_spec.observation_space.contains(first_next_observation):
+            raise ValueError(
+                'next_observations must conform to observation_space {}, but got '
+                'data with shape {} instead.'.format(
+                    env_spec.observation_space, first_next_observation))
+
+        # actions
+        # TODO: Shape check can be removed for gym>=0.15.4 # pylint: disable=fixme # noqa: E501
```
krzentner: We have merged the gym update, so this check can be removed.

src/garage/_dtypes.py
```diff
@@ -167,3 +167,199 @@ def __new__(cls, env_spec, observations, actions, rewards, terminals,
...
+        if not env_spec.observation_space.contains(next_observation):
+            raise ValueError(
+                'next_observation must conform to observation_space {}, but got '
+                'data with shape {} instead.'.format(
+                    env_spec.observation_space, next_observation))
+
+        # action
+        # TODO: Shape check can be removed for gym>=0.15.4 # pylint: disable=fixme # noqa: E501
+        if not (env_spec.action_space.contains(action)
+                and action.shape == env_spec.action_space.shape):
```
naeioi: I see this is copied from the old code, but the shape test is actually redundant.

src/garage/_dtypes.py
```diff
@@ -167,3 +167,199 @@ def __new__(cls, env_spec, observations, actions, rewards, terminals,
...
+
+    Raises:
+        ValueError: If any of the above attributes do not conform to their
+            prescribed types and shapes.
+
+    """
+    __slots__ = ()
+
+    def __new__(cls, env_spec, observations, actions, rewards, next_observations,
+                 terminals, env_infos, agent_infos):  # noqa: D102
```
naeioi: Is it that the only differences between `SamplesBatch` and `TrajectoryBatch` are 
- `SamplesBatch` incorporates `next_observations`
- But does not incorporate `lengths`?

src/garage/_dtypes.py
```diff
@@ -167,3 +167,199 @@ def __new__(cls, env_spec, observations, actions, rewards, terminals,
...
+            'env_infos',
+            'agent_infos',
+        ])):
+    # pylint: disable=missing-return-doc, missing-return-type-doc, missing-param-doc, missing-type-doc  # noqa: E501
+    r"""A tuple representing a batch of whole trajectories.
+
+    A :class:`SamplesBatch` represents a batch of samples
+    produced when one or more agents interacts with one or more environments.
+
+    Attributes:
```
naeioi: This docstring needs update to reflect the code

src/garage/_dtypes.py
ryanjulian: the samples in a sample batch don't necessarily have any meaningful time relationship.

lywong92:
I also think `Sample` is better for our use case and easier to understand. This is looking good; just need to address the existing comments.

src/garage/_dtypes.py
```diff
@@ -167,3 +167,199 @@ def __new__(cls, env_spec, observations, actions, rewards, terminals,
...
+            prescribed types and shapes.
+
+    """
+    __slots__ = ()
+
+    def __new__(cls, env_spec, observation, action, reward, next_observation,
+                terminal, env_info, agent_info):  # noqa: D102
+        # pylint: disable=too-many-branches
+
+        if type(observation) != float:
```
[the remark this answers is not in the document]
ryanjulian: note that even the `np.ndarray` doesn't need to have element type `np.float` -- it's perfectly valid for observations to be `np.int32` or even `np.bool`

src/garage/_dtypes.py
```diff
@@ -167,3 +167,199 @@ def __new__(cls, env_spec, observations, actions, rewards, terminals,
...
+
+    Attributes:
+        env_spec (garage.envs.EnvSpec): Specification for the environment from
+            which this data was sampled.
+        observation (float): Observation at the current time step in the
+            environment.
+        action (float): Action sampled from the policy given the observation
+            at the current time step.
+        reward (float): The reward for taking the action at the current state.
+        terminals (boolean): A boolean representing the termination signal for
```
ryanjulian: does this data structure need to represent the difference between a true terminal and a timeout terminal? are those treated differently in off-policy algorithms? does it even need to know if a sample is a timeout terminal?

src/garage/_dtypes.py
```diff
@@ -167,3 +167,199 @@ def __new__(cls, env_spec, observations, actions, rewards, terminals,
...
+                'shape {} instead.'.format(env_spec.action_space,
+                                           action))
+
+        # rewards
+        if type(reward) != float:
+            raise ValueError(
+                'Rewards must be type {}, but got type {} '
+                'instead.'.format(float, type(reward)))
+
+        if type(terminal) != bool:
```
ryanjulian: we should probably stick to using `np.` types for everything in this dtype.

src/garage/_dtypes.py
```diff
@@ -167,3 +167,199 @@ def __new__(cls, env_spec, observations, actions, rewards, terminals,
         return super().__new__(TrajectoryBatch, env_spec, observations,
```
ryanjulian: you need to add your new types to `garage/__init__.py` so that they can be imported...

src/garage/_dtypes.py
```diff
@@ -167,3 +167,199 @@ def __new__(cls, env_spec, observations, actions, rewards, terminals,
...
                                actions, rewards, terminals, env_infos,
                                agent_infos, lengths)
+
+class SamplesBatch(
```
ryanjulian: i think you should focus on defining a good `Sample` dtype and then consider later if defining a collection dtype is even necessary.

PR #1109, opened 2019-12-26
Convert RaySampler to use TrajectoryBatch

krzentner:
Since I'm refactoring samplers to the new API anyways, I may as well use TrajectoryBatch. Since the other PR has been fully reviewed, I'm putting this change in this PR instead.

### DIFF
--- a/src/garage/_dtypes.py
+++ b/src/garage/_dtypes.py
@@ -185,6 +185,40 @@ class TrajectoryBatch(
                                last_observations, actions, rewards, terminals,
                                env_infos, agent_infos, lengths)
 
+    @classmethod
+    def concatenate(cls, *batches):
+        """Create a TrajectoryBatch by concatenating TrajectoryBatches.
+
+        Args:
+            batches (list[TrajectoryBatch]): Batches to concatenate.
+
+        Returns:
+            TrajectoryBatch: The concatenation of the batches.
+
+        """
+        if __debug__:
+            for b in batches:
+                assert (set(b.env_infos.keys()) == set(
+                    batches[0].env_infos.keys()))
+                assert (set(b.agent_infos.keys()) == set(
+                    batches[0].agent_infos.keys()))
+        env_infos = {
+            k: np.concatenate([b.env_infos[k] for b in batches])
+            for k in batches[0].env_infos.keys()
+        }
+        agent_infos = {
+            k: np.concatenate([b.agent_infos[k] for b in batches])
+            for k in batches[0].agent_infos.keys()
+        }
+        return cls(
+            batches[0].env_spec,
+            np.concatenate([batch.observations for batch in batches]),
+            np.concatenate([batch.last_observations for batch in batches]),
+            np.concatenate([batch.actions for batch in batches]),
+            np.concatenate([batch.rewards for batch in batches]),
+            np.concatenate([batch.terminals for batch in batches]), env_infos,
+            agent_infos, np.concatenate([batch.lengths for batch in batches]))
+
 
 class TimeStep(
         collections.namedtuple('TimeStep', [
--- a/src/garage/sampler/ray_sampler.py
+++ b/src/garage/sampler/ray_sampler.py
@@ -9,10 +9,10 @@ function, and a rollout function.
 from collections import defaultdict
 import pickle
 
-import numpy as np
 import psutil
 import ray
 
+from garage import TrajectoryBatch
 from garage.misc.prog_bar_counter import ProgBarCounter
 from garage.sampler.sampler import Sampler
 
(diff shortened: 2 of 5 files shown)


src/garage/_dtypes.py
```diff
@@ -167,3 +167,36 @@ def __new__(cls, env_spec, observations, actions, rewards, terminals,
...
+        """Create a TrajectoryBatch by concatenating TrajectoryBatches.
+
+        Args:
+            batches (list[TrajectoryBatch]): Batches to concatenate.
+
+        Returns:
+            TrajectoryBatch: The concatenation of the batches.
+
+        """
+        if __debug__:
```
avnishn: what is the purpose of this statement?

src/garage/sampler/ray_sampler.py
```diff
@@ -9,95 +9,125 @@
...
-                 seed,
-                 should_render=False,
+                 worker_factory,
+                 agents,
+                 envs,
                  num_processors=None,
                  sampler_worker_cls=None):
-        super().__init__(algo, env)
+        # pylint: disable=super-init-not-called
         self._sampler_worker = ray.remote(SamplerWorker if sampler_worker_cls
```
avnishn: This statement should come after checking to see if ray is initialized 

``` 
if not ray.is_initialized():
           ray.init(log_to_driver=False)
```

src/garage/sampler/ray_sampler.py
```diff
@@ -9,95 +9,125 @@
...
         self._num_workers = (num_processors if num_processors else
                              psutil.cpu_count(logical=False))
         self._all_workers = defaultdict(None)
-        self._idle_worker_ids = list(range(self._num_workers))
-        self._active_worker_ids = []
+        self._workers_started = False
+        self.start_worker()
+
+    @classmethod
+    def from_worker_factory(cls, worker_factory, agents, envs):
```
avnishn: I don't quite understand why we need an additional function for constructing the sampler other than the constructor of the sampler itself. @krzentner can you please share your thoughts on this design choice

src/garage/sampler/worker.py
```diff
@@ -0,0 +1,277 @@
...
+        self._rewards = []
+        self._terminals = []
+        self._lengths = []
+        self._agent_infos = defaultdict(list)
+        self._env_infos = defaultdict(list)
+        self._path_length = 0
+        self._last_obs = self.env.reset()
+        self.agent.reset()
+
+    def step_rollout(self):
```
avnishn: should we be using the gym termination signal for keeping track of how environment termination occurs. Ironically enough, sac implementation in spinning up keeps track of environment termination this way as well, even though open ai created gym, and the environment wrapper that solves this problem.

src/garage/sampler/ray_sampler.py
ryanjulian: if you have this feedback, please put it in the PR which introduces this API, not this PR (which merely refactors to use TrajectoryBatch).

src/garage/sampler/worker.py
ryanjulian: I think we should fix this universally throughout garage (probably in `GarageEnv`) rather than implementing a fix in every sampler.

src/garage/_dtypes.py
ryanjulian: it checks that all elements have the same keys. that might be slow, so it's hidden behind a debug flag.

src/garage/sampler/ray_sampler.py
```diff
@@ -9,95 +9,125 @@
...
     def start_worker(self):
         """Initialize a new ray worker."""
-        # Pickle the environment once, instead of once per worker.
-        env_pkl = pickle.dumps(self.env)
-        # We need to pickle the agent so that we can e.g. set up the TF Session
+        if self._workers_started:
+            return
+        self._workers_started = True
+        # We need to pickle the agent so that we can e.g. set up the TF.Session
         # in the worker *before* unpicling it.
```
ryanjulian: nit: unpickle

src/garage/sampler/sampler.py
```diff
@@ -0,0 +1,117 @@
...
+    def start_worker(self):
+        """Initialize the sampler.
+
+        i.e. launching parallel workers if necessary.
+
+        This method is deprecated, please launch workers in construct instead.
+        """
+
+    @abc.abstractmethod
+    def obtain_samples(self, itr, num_samples, agent_update, env_update=None):
```
ryanjulian: any reason this interface doesn't use TrajectoryBatch?

src/garage/_dtypes.py
krzentner: It's worth mentioning that by default in python `__debug__ == True`.

src/garage/sampler/ray_sampler.py
krzentner: Basically the rationale is: `__init__` in interfaces is unusual and current implementations of `__init__` are all different.

src/garage/sampler/ray_sampler.py
krzentner: Thanks!

src/garage/sampler/sampler.py
krzentner: Oh whoops. I missed the docstring. It does now.

PR #1114, opened 2019-12-31
Add support for Sample and SamplesBatch Dtypes

avnishn:
- Garage missing support for Dtypes for OffPolicyRLA's
- Sample dtype is all the info collected from an env at
  a single timestep, and an action that an agent would
  take at that time step.
- SamplesBatch is an tuple of samples.

### DIFF
--- a/src/garage/_dtypes.py
+++ b/src/garage/_dtypes.py
@@ -105,9 +105,7 @@ class TrajectoryBatch(
                                                     observations.shape[0]))
 
         # actions
-        # TODO: Shape check can be removed for gym>=0.15.4 # pylint: disable=fixme # noqa: E501
-        if not (env_spec.action_space.contains(first_action)
-                and first_action.shape == env_spec.action_space.shape):
+        if not env_spec.action_space.contains(first_action):
             raise ValueError(
                 'actions must conform to action_space {}, but got data with '
                 'shape {} instead.'.format(env_spec.action_space,
@@ -128,12 +126,12 @@ class TrajectoryBatch(
         # terminals
         if terminals.shape != (inferred_batch_size, ):
             raise ValueError(
-                'Terminals tensor must have shape {}, but got shape {} '
+                'terminals tensor must have shape {}, but got shape {} '
                 'instead.'.format(inferred_batch_size, terminals.shape))
 
         if terminals.dtype != np.bool:
             raise ValueError(
-                'Terminals tensor must be dtype np.bool, but got tensor '
+                'terminals tensor must be dtype np.bool, but got tensor '
                 'of dtype {} instead.'.format(terminals.dtype))
 
         # env_infos
@@ -167,3 +165,90 @@ class TrajectoryBatch(
         return super().__new__(TrajectoryBatch, env_spec, observations,
                                actions, rewards, terminals, env_infos,
                                agent_infos, lengths)
+
+
+class TimeStep(
+        collections.namedtuple('TimeStep', [
+            'env_spec',
+            'observation',
+            'action',
+            'reward',
+            'next_observation',
+            'terminal',
+            'env_info',
+            'agent_info',
+        ])):
+    # pylint: disable=missing-return-doc, missing-return-type-doc, missing-param-doc, missing-type-doc  # noqa: E501
+    r"""A tuple representing a single TimeStep.
+
+    A :class:`TimeStep` represents a single sample when an agent interacts with
+        an environment.
+
+    Attributes:
+        env_spec (garage.envs.EnvSpec): Specification for the environment from
+            which this data was sampled.
+        observation (numpy.ndarray): A numpy array of shape :math:`(O^*)`
+            containing the observation for the this time step in the
+            environment. These must conform to
+            :obj:`env_spec.observation_space`.
+        action (numpy.ndarray): A numpy array of shape :math:`(A^*)`
+            containing the action for the this time step. These must conform
+            to :obj:`env_spec.action_space`.
+        reward (float): A float representing the reward for taking the action
+            given the observation, at the this time step.
+        terminals (bool): The termination signal for the this time step.
+        env_info (dict): A dict arbitrary environment state information.
+        agent_info (numpy.ndarray): A dict of arbitrary agent
+            state information. For example, this may contain the hidden states
+            from an RNN policy.
+
+
+    Raises:
+        ValueError: If any of the above attributes do not conform to their
+            prescribed types and shapes.
+
+    """
+
+    def __new__(cls, env_spec, observation, action, reward, next_observation,
+                terminal, env_info, agent_info):  # noqa: D102
+
+        # observation
+        if not env_spec.observation_space.contains(observation):
+            raise ValueError(
+                'observation must conform to observation_space {}, but got '
+                'data with shape {} instead.'.format(
+                    env_spec.observation_space, observation))
+
+        if not env_spec.observation_space.contains(next_observation):
+            raise ValueError(
+                'next_observation must conform to observation_space {},'
+                ' but got data with shape {} instead.'.format(
+                    env_spec.observation_space, next_observation))
+
+        # action
+        if not env_spec.action_space.contains(action):
+            raise ValueError(
+                'action must conform to action_space {}, but got data with '
+                'shape {} instead.'.format(env_spec.action_space, action))
+
+        if not isinstance(agent_info, dict):
+            raise ValueError('agent_info must be type {}, but got type {} '
+                             'instead.'.format(dict, type(agent_info)))
+
+        if not isinstance(env_info, dict):
+            raise ValueError('env_info must be type {}, but got type {} '
+                             'instead.'.format(dict, type(env_info)))
+
+        # rewards
+        if not isinstance(reward, float):
+            raise ValueError('reward must be type {}, but got type {} '
+                             'instead.'.format(float, type(reward)))
+
+        if not isinstance(terminal, bool):
+            raise ValueError(
+                'terminal must be dtype bool, but got dtype {} instead.'.
+                format(type(terminal)))
+
+        return super().__new__(TimeStep, env_spec, observation, action, reward,
+                               next_observation, terminal, env_info,
+                               agent_info)
--- a/tests/garage/test_dtypes.py
+++ b/tests/garage/test_dtypes.py
@@ -2,6 +2,7 @@ import gym.spaces
 import numpy as np
 import pytest
 
+from garage import TimeStep
 from garage import TrajectoryBatch
 from garage.envs import EnvSpec
 
@@ -47,7 +48,7 @@ def traj_data():
     }
 
 
-def test_new(traj_data):
+def test_new_traj(traj_data):
     t = TrajectoryBatch(**traj_data)
     assert t.env_spec is traj_data['env_spec']
     assert t.observations is traj_data['observations']
@@ -59,7 +60,7 @@ def test_new(traj_data):
     assert t.lengths is traj_data['lengths']
 
 
-def test_lengths_shape_mismatch(traj_data):
+def test_lengths_shape_mismatch_traj(traj_data):
     with pytest.raises(ValueError,
                        match='Lengths tensor must be a tensor of shape'):
         traj_data['lengths'] = traj_data['lengths'].reshape((4, -1))
@@ -67,7 +68,7 @@ def test_lengths_shape_mismatch(traj_data):
         del t
 
 
-def test_lengths_dtype_mismatch(traj_data):
+def test_lengths_dtype_mismatch_traj(traj_data):
     with pytest.raises(ValueError,
                        match='Lengths tensor must have an integer dtype'):
         traj_data['lengths'] = traj_data['lengths'].astype(np.float32)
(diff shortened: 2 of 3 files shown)


src/garage/_dtypes.py
```diff
@@ -167,3 +166,128 @@ def __new__(cls, env_spec, observations, actions, rewards, terminals,
...
+            raise ValueError(
+                'Terminal must be dtype bool, but got dtype {} instead.'.
+                format(type(terminal)))
+
+        return super().__new__(Sample, env_spec, observation, action, reward,
+                               next_observation, terminal, env_info,
+                               agent_info)
+
+
+class SamplesBatch(collections.namedtuple('SamplesBatch', [
```
ryanjulian: i don't see what purpose this type serves -- it's basically `Tuple[List]` which means it's not immutable at all (the inner list is still mutable). If you wanted an immutable sequence of samples you would just use a tuple, but why bother when the elements are immutable?

Please remove this for now and just stick with regular collections of samples. If an interface wants to specify that you can't reorder the samples, it can do that using a tuple. If it doesn't care about order, it can use a list (i.e. regular old python).

src/garage/_dtypes.py
```diff
@@ -167,3 +166,128 @@ def __new__(cls, env_spec, observations, actions, rewards, terminals,
...
+            raise ValueError(
+                'action must conform to action_space {}, but got data with '
+                'shape {} instead.'.format(env_spec.action_space, action))
+
+        if not isinstance(agent_info, dict):
+            raise ValueError('Agent_info must be type {}, but got type {} '
+                             'instead.'.format(dict, type(agent_info)))
+
+        if not isinstance(env_info, dict):
+            raise ValueError('Env_info must be type {}, but got type {} '
```
ryanjulian: `env_info`

src/garage/_dtypes.py
```diff
@@ -167,3 +166,128 @@ def __new__(cls, env_spec, observations, actions, rewards, terminals,
...
+                ' but got data with shape {} instead.'.format(
+                    env_spec.observation_space, next_observation))
+
+        # action
+        if not env_spec.action_space.contains(action):
+            raise ValueError(
+                'action must conform to action_space {}, but got data with '
+                'shape {} instead.'.format(env_spec.action_space, action))
+
+        if not isinstance(agent_info, dict):
```
ryanjulian: `agent_info`

src/garage/_dtypes.py
```diff
@@ -167,3 +166,128 @@ def __new__(cls, env_spec, observations, actions, rewards, terminals,
...
+            :obj:`env_spec.observation_space`.
+            actions (numpy.ndarray): A numpy array of shape
+            :math:`(A^*)` containing the action for the current time step.
+            These must conform to :obj:`env_spec.action_space`.
+        reward (float): A float representing the reward for taking
+            the action given the observation, at the current time step.
+        terminals (bool): The termination signal for the current time step.
+        env_info (dict): A dict arbitrary environment state
+            information.
+        agent_infos (numpy.ndarray): A dict of arbitrary agent
```
ryanjulian: `agent_info`? No need  for this to be plural, since there's only one.

src/garage/_dtypes.py
```diff
@@ -167,3 +166,128 @@ def __new__(cls, env_spec, observations, actions, rewards, terminals,
...
+            'env_info',
+            'agent_info',
+        ])):
+    # pylint: disable=missing-return-doc, missing-return-type-doc, missing-param-doc, missing-type-doc  # noqa: E501
+    r"""A tuple representing a single sample.
+
+    A :class:`Sample` represents a single sample when an agent interacts with
+        an environment.
+
+    Attributes:
```
ryanjulian: rather than calling them "current time steps" I suggest you say "this time step", because if you ar manipulating instances of this dtype there is no notion of "current" vs "other"

src/garage/_dtypes.py
```diff
@@ -167,3 +166,128 @@ def __new__(cls, env_spec, observations, actions, rewards, terminals,
...
+        env_spec (garage.envs.EnvSpec): Specification for the environment from
+            which this data was sampled.
+        observation (numpy.ndarray): A numpy array of shape
+            :math:`(O^*)` containing the observation for the current time step
```
ryanjulian: can you ensure these are consistently wrapped to the 78-char line?

src/garage/_dtypes.py
```diff
@@ -167,3 +166,128 @@ def __new__(cls, env_spec, observations, actions, rewards, terminals,
...
+            in the environment. These must conform to
+            :obj:`env_spec.observation_space`.
+            actions (numpy.ndarray): A numpy array of shape
```
ryanjulian: bad indent? also, actions plural?

src/garage/_dtypes.py
```diff
@@ -167,3 +166,128 @@ def __new__(cls, env_spec, observations, actions, rewards, terminals,
         return super().__new__(TrajectoryBatch, env_spec, observations,
                                actions, rewards, terminals, env_infos,
                                agent_infos, lengths)
+
+
+class Sample(
```
ryanjulian: i don't think the name of this type should be `Sample` because:

1. The name is already overloaded. For instance, when you say sample, do you mean the verb "to sample"? Or do you mean sampling from a distribution? or sampling from an environment? It's important that the codebase be easily grep-able. Data types should clearly be nouns, not verbs.
1. We don't use the term sample already anyway, no need to add ambiguity
1. We already use the term `TotalEnvSteps` to count these in algorithms. Whatever we choose here should be consistent with that name.

Whatever we use should also be consistent with the idea of a trajectory -- in principle you should be able to transform a bunch of time-ordered Samples into a TrajectoryBatch.

Candidate names:
* Step -- already used by gym
* TimeStep (used by tf_agents)
* EnvStep
* EnvSample

Of these I think that `TimeStep` and `EnvStep` make the most sense. `EnvStep` is kind of redundant (where else do steps come from?) so I suggest you name this `TimeStep` and also refactor usages of `TotalEnvSteps` to read `TotalTimeSteps`.

tests/garage/test_dtypes.py
```diff
@@ -2,9 +2,13 @@
 import numpy as np
 import pytest
 
+from garage import Sample
+from garage import SamplesBatch
 from garage import TrajectoryBatch
 from garage.envs import EnvSpec
 
+# ===========================TrajectoryBatch-Test=========================== #
```
ryanjulian: please don't leave useless block comments like this

ryanjulian: if you want to group tests, put them in different files or use a class within the same file.

src/garage/_dtypes.py
avnishn: I agree. I added had added this in reference to a comment you made on the previous PR. I will remove this dtype for now.

src/garage/_dtypes.py
avnishn: resolved

src/garage/_dtypes.py
avnishn: resolved

src/garage/_dtypes.py
avnishn: resolved, thanks

src/garage/_dtypes.py
avnishn: These are all wrapped to 79 chars.

src/garage/_dtypes.py
avnishn: resolved

tests/garage/test_dtypes.py
avnishn: removed this and similar block comments

src/garage/_dtypes.py
ryanjulian: i was referring to indenting things like `:math:`(O^*)` too early, creating lots of whitespace on the right.

src/garage/_dtypes.py
```diff
@@ -167,3 +166,91 @@ def __new__(cls, env_spec, observations, actions, rewards, terminals,
...
+
+    """
+
+    def __new__(cls, env_spec, observation, action, reward, next_observation,
+                terminal, env_info, agent_info):  # noqa: D102
+
+        # observations
+        if not env_spec.observation_space.contains(observation):
+            raise ValueError(
+                'observations must conform to observation_space {}, but got '
```
ryanjulian: please update all of these messages with the correct english (i.e. singular vs plural)

src/garage/_dtypes.py
```diff
@@ -167,3 +166,91 @@ def __new__(cls, env_spec, observations, actions, rewards, terminals,
...
+            raise ValueError('agent_info must be type {}, but got type {} '
+                             'instead.'.format(dict, type(agent_info)))
+
+        if not isinstance(env_info, dict):
+            raise ValueError('env_info must be type {}, but got type {} '
+                             'instead.'.format(dict, type(env_info)))
+
+        # rewards
+        if not isinstance(reward, float):
+            raise ValueError('Rewards must be type {}, but got type {} '
```
ryanjulian: reward? make sure the terms used in error messages are exactly the symbol names from the code

src/garage/_dtypes.py
```diff
@@ -167,3 +166,91 @@ def __new__(cls, env_spec, observations, actions, rewards, terminals,
...
+                             'instead.'.format(float, type(reward)))
+
+        if not isinstance(terminal, bool):
+            raise ValueError(
+                'Terminal must be dtype bool, but got dtype {} instead.'.
```
ryanjulian: terminal

src/garage/_dtypes.py
```diff
@@ -167,3 +166,91 @@ def __new__(cls, env_spec, observations, actions, rewards, terminals,
...
+        observation (numpy.ndarray): A numpy array of shape
+            :math:`(O^*)` containing the observation for the current time step
+            in the environment. These must conform to
+            :obj:`env_spec.observation_space`.
+        actions (numpy.ndarray): A numpy array of shape
+            :math:`(A^*)` containing the action for the current time step.
+            These must conform to :obj:`env_spec.action_space`.
+        reward (float): A float representing the reward for taking
+            the action given the observation, at the current time step.
+        terminals (bool): The termination signal for the current time step.
```
ryanjulian: terminal

src/garage/_dtypes.py
```diff
@@ -167,3 +166,91 @@ def __new__(cls, env_spec, observations, actions, rewards, terminals,
...
+        an environment.
+
+    Attributes:
+        env_spec (garage.envs.EnvSpec): Specification for the environment from
+            which this data was sampled.
+        observation (numpy.ndarray): A numpy array of shape
+            :math:`(O^*)` containing the observation for the current time step
+            in the environment. These must conform to
+            :obj:`env_spec.observation_space`.
+        actions (numpy.ndarray): A numpy array of shape
```
ryanjulian: action, not actions

this comment is also indented too early.

src/garage/_dtypes.py
ryanjulian: please split on the `.` operator

src/garage/_dtypes.py
```diff
@@ -167,3 +166,91 @@ def __new__(cls, env_spec, observations, actions, rewards, terminals,
         return super().__new__(TrajectoryBatch, env_spec, observations,
                                actions, rewards, terminals, env_infos,
                                agent_infos, lengths)
+
+
+class TimeStep(
```
ryanjulian: when manually indenting code and comments, please use the classic terminal style of justified wrapping -- that is, making maximum use of horizontal space before wrapping a line.

src/garage/_dtypes.py
```diff
@@ -105,7 +105,6 @@ def __new__(cls, env_spec, observations, actions, rewards, terminals,
                                                     observations.shape[0]))
 
         # actions
-        # TODO: Shape check can be removed for gym>=0.15.4 # pylint: disable=fixme # noqa: E501
```
ryanjulian: you removed the TODO comment but not the shape check it refers to! please fix.

src/garage/_dtypes.py
```diff
@@ -167,3 +166,91 @@ def __new__(cls, env_spec, observations, actions, rewards, terminals,
...
+            'reward',
+            'next_observation',
+            'terminal',
+            'env_info',
+            'agent_info',
+        ])):
+    # pylint: disable=missing-return-doc, missing-return-type-doc, missing-param-doc, missing-type-doc  # noqa: E501
+    r"""A tuple representing a single sample.
+
+    A :class:`Sample` represents a single sample when an agent interacts with
```
ryanjulian: comment does not match the code

src/garage/_dtypes.py
ryanjulian: i just realized here that validating every sample could have a big performance impact -- perhaps you can profile running `rollout` using this data structure versus a regular tuple?

avnishn: I'll run a profile. I'm writing a sampler to use with it right now that I'll include in this PR.

src/garage/_dtypes.py
avnishn: resolved

src/garage/_dtypes.py
avnishn: resolved

src/garage/_dtypes.py
avnishn: resolved

src/garage/_dtypes.py
avnishn: resolved

src/garage/_dtypes.py
avnishn: resolved

src/garage/_dtypes.py
avnishn: resolved

src/garage/_dtypes.py
avnishn: resolved

avnishn:
@Mergifyio rebase

Issue #1091, opened 2019-12-07
Add a performance evaluation interface to the RLAlgorithm base class


ryanjulian:
Right now, performance of each algorithm is calculated within that algorithm. This means each algorithm may use different logger keys, math implementations, or metrics entirely to report its performance, and makes comparing results challenging.

I chatted with @krzentner and believe we can solve this in a fairly simple way, with a 2-part process

1. Define a performance evaluation API in the RLAlgorithm base class. This calculates and logs the performance of *any* algorithm using a batch of samples from the algorithm.
2. Each algorithm defines for itself how to generate the evaluation batch, and calls the evaluation API with the relevant frequency (e.g. once per epoch). For on-policy algorithms, the evaluation batch can just be the most recent batch of samples. Off-policy algorithms might need to do something more complicated, such as asking the replay buffer for the most recent `n` complete paths, or running some rollouts themselves (some on-policy algorithms add noise to their actions during sampling, but not during evaluation). Some algorithms might need to do something even more different (even having a whole other sampler for evaluation), but *how* they get the evaluation samples doesn't matter as long as they pass the batch to the API.

```python

class RLAlgorithm:
    
    @classmethod
    def evaluate_performance(evaluation_batch):
    """Evaluate the performance of the algorithm using a batch of trajectories.

    Args:
        evaluation_batch (TrajectoryBatch): A :obj:`TrajectoryBatch` of evaluation
            trajectories, representing the best current performance of the
            algorithm.
    """

Most of this code already exists in multiple places, e.g.
https://github.com/rlworkgroup/garage/blob/master/src/garage/tf/algos/batch_polopt.py#L257
https://github.com/rlworkgroup/garage/blob/master/src/garage/tf/algos/ddpg.py#L288
https://github.com/rlworkgroup/garage/blob/master/src/garage/tf/algos/dqn.py#L215
https://github.com/rlworkgroup/garage/blob/master/src/garage/np/algos/batch_polopt.py#L134
https://github.com/rlworkgroup/garage/blob/master/src/garage/np/algos/off_policy_rl_algorithm.py#L132

refs:
https://github.com/rlworkgroup/garage/issues/1084
https://github.com/rlworkgroup/garage/issues/1077
https://github.com/rlworkgroup/garage/issues/1015
https://github.com/rlworkgroup/garage/blob/master/src/garage/_dtypes.py#L7

avnishn:
[...]

Issue #1021, opened 2019-11-15
Runner should log total environment steps (and other vital stats)


ryanjulian:
This will make it easier to compute algorithm performance consistently.

zequnyu:
Is this what you are looking for?
https://github.com/rlworkgroup/garage/blob/169c822573d3899911590988450e2466767f2451/src/garage/tf/algos/batch_polopt.py#L94

ryanjulian:
Yes. I think this is basically a duplicate of the evaluation PRs you are working on.

zequnyu:
Closed by #1017.

Issue #1015, opened 2019-11-07
Log success rate from all algorithms (or samplers)


ryanjulian:
We usually focus on AverageDiscountedReturn, but many RL environments use an arbitrary shaped reward (which means ADR doesn't help much) and are better measured using success rate.

Success rate is simple to calculate. If I have a batch of `paths`, then the success rate is just:
```python
success_rate = np.mean([np.any(p['dones']) for p in paths])
```
i.e. it's just the portion of paths which actually experienced a termination condition before being the end of the episode.

naeioi:
Success rate is useful to reason about the process of experiment, I wonder if it is also useful in training policy. 

Is this issue a priority? Currently, only DDPG logs success rate. I see this more of a problem of code consistency.

ryanjulian:
Fairly high-priority, since it's needed to run metaworld benchmarks. I think this can be covered by @zequnyu 's work on #1091 . Perhaps if you want to help tackle it you can help him finish that issue? It might require multiple PRs.

naeioi:
@gitanshu recently reached out to me on this issue. Do you feel like working on this?

ryanjulian:
@krzentner @zequnyu is this fixed? for *all* algorithms?

zequnyu:
This is covered by #1116 and #1142.

PR #1116, opened 2020-01-05
Add evaluation interface

zequnyu:
This PR covers part of #1091.

What it has:
-  add an `evaluate_performance()` method which can be used by all algorithms to report performance
- add an evaluation sampler for off-policy algos to collect rollouts, created by LocalRunner when in need
- clean up some code in cem and cma_es

Next steps:
- refactor `process_samples()`, maybe into a pure function, so that variables such as `ent`
https://github.com/rlworkgroup/garage/blob/c55683cd383ec00f008e5871fece772e96ee93e2/src/garage/np/algos/batch_polopt.py#L140 can be added to `evaluate_performance(). Currently it's hard to separate it.
- add log success rate ( #1015). Add `dones` to `OnPolicyVectorizedSampler`'s return paths?
- use `TrajectoryBatch`? This definitely needs refactoring `process_samples()`
- move total environment steps https://github.com/rlworkgroup/garage/blob/c55683cd383ec00f008e5871fece772e96ee93e2/src/garage/np/algos/batch_polopt.py#L73 into `evaluate_performance()`? (#1021) Maybe this will make the code more complicated.

Opening this draft for feedback.

### DIFF
--- a/src/garage/experiment/local_runner.py
+++ b/src/garage/experiment/local_runner.py
@@ -1,6 +1,7 @@
 """Provides algorithms with access to most of garage's features."""
 import copy
 import os
+import pickle
 import time
 
 from dowel import logger, tabular
@@ -431,6 +432,15 @@ class LocalRunner:
             self._train_args.pause_for_plot = pause_for_plot
 
         return self._algo.train(self)
+
+    def get_env_copy(self):
+        """Get a copy of the environment.
+
+        Returns:
+            garage.envs.GarageEnv: An environement instance.
+
+        """
+        return pickle.loads(pickle.dumps(self._env))
 
     @property
     def total_env_steps(self):
--- a/src/garage/np/algos/base.py
+++ b/src/garage/np/algos/base.py
@@ -1,6 +1,11 @@
 """Interface of RLAlgorithm."""
 import abc
 
+from dowel import tabular
+import numpy as np
+
+from garage.misc import tensor_utils as np_tensor_utils
+
 
 class RLAlgorithm(abc.ABC):
     """Base class for all the algorithms.
@@ -25,4 +30,56 @@ class RLAlgorithm(abc.ABC):
             The average return in last epoch cycle or None.
 
         """
-        pass
+
+    def evaluate_performance(self, itr, batch):
+        # pylint: disable=no-self-use
+        r"""Evaluate the performance of the algorithm.
+
+        Args:
+            itr (int): Iteration number.
+            batch (dict): Evaluation trajectories, representing
+                the best current performance of the algorithm, with keys:
+                * env_spec (garage.envs.EnvSpec): Specification for the
+                environment from which this data was sampled.
+                * observations (numpy.ndarray): A numpy array containing the
+                    observations for all time steps in this batch.
+                * actions (numpy.ndarray): A  numpy array containing the
+                    actions for all time steps in this batch.
+                * rewards (numpy.ndarray): A numpy array containing the
+                    rewards for all time steps in this batch.
+                * terminals (numpy.ndarray): A boolean numpy array
+                    containing the termination signals for all time steps
+                    in this batch.
+                * env_infos (dict): A dict of numpy arrays arbitrary
+                    environment state information.
+                * agent_infos (numpy.ndarray): A dict of numpy arrays
+                    arbitrary agent state information.
+                * lengths (numpy.ndarray): An integer numpy array
+                    containing the length of each trajectory in this batch.
+                * discount (float): Discount value, from algorithm's property.
+
+        Returns:
+            numpy.ndarray: Undiscounted returns.
+
+        """
+        returns = []
+        for reward in batch['rewards']:
+            rtn = np_tensor_utils.discount_cumsum(reward, batch['discount'])
+            returns.append(rtn)
+
+        average_discounted_return = np.mean([rtn[0] for rtn in returns])
+
+        undiscounted_returns = [sum(reward) for reward in batch['rewards']]
+
+        tabular.record('Iteration', itr)
+        tabular.record('Evaluation/NumTrajs', len(returns))
+
+        tabular.record('Evaluation/AverageDiscountedReturn',
+                       average_discounted_return)
+        tabular.record('Evaluation/AverageReturn',
+                       np.mean(undiscounted_returns))
+        tabular.record('Evaluation/StdReturn', np.std(undiscounted_returns))
+        tabular.record('Evaluation/MaxReturn', np.max(undiscounted_returns))
+        tabular.record('Evaluation/MinReturn', np.min(undiscounted_returns))
+
+        return undiscounted_returns
--- a/src/garage/np/algos/batch_polopt.py
+++ b/src/garage/np/algos/batch_polopt.py
@@ -111,36 +111,55 @@ class BatchPolopt(RLAlgorithm):
                 path['rewards'], self.discount)
             returns.append(path['returns'])
 
+        obs = [path['observations'] for path in paths]
+        obs = tensor_utils.pad_tensor_n(obs, max_path_length)
+
+        actions = [path['actions'] for path in paths]
+        actions = tensor_utils.pad_tensor_n(actions, max_path_length)
+
+        rewards = [path['rewards'] for path in paths]
+        rewards = tensor_utils.pad_tensor_n(rewards, max_path_length)
+
         agent_infos = [path['agent_infos'] for path in paths]
         agent_infos = tensor_utils.stack_tensor_dict_list([
             tensor_utils.pad_tensor_dict(p, max_path_length)
             for p in agent_infos
         ])
 
+        env_infos = [path['env_infos'] for path in paths]
+        env_infos = tensor_utils.stack_tensor_dict_list([
+            tensor_utils.pad_tensor_dict(p, max_path_length) for p in env_infos
+        ])
+
+        terminals = [path['dones'] for path in paths]
+
         valids = [np.ones_like(path['returns']) for path in paths]
         valids = tensor_utils.pad_tensor_n(valids, max_path_length)
 
-        average_discounted_return = (np.mean(
-            [path['returns'][0] for path in paths]))
-
-        undiscounted_returns = [sum(path['rewards']) for path in paths]
-        self.episode_reward_mean.extend(undiscounted_returns)
+        lengths = np.asarray([v.sum() for v in valids])
 
         ent = np.sum(self.policy.distribution.entropy(agent_infos) *
                      valids) / np.sum(valids)
 
-        samples_data = dict(average_return=np.mean(undiscounted_returns))
+        undiscounted_returns = self.evaluate_performance(
+            itr,
+            dict(env_spec=None,
+                 observations=obs,
+                 actions=actions,
+                 rewards=rewards,
+                 terminals=terminals,
+                 env_infos=env_infos,
+                 agent_infos=agent_infos,
+                 lengths=lengths,
+                 discount=self.discount))
 
-        tabular.record('Iteration', itr)
-        tabular.record('AverageDiscountedReturn', average_discounted_return)
-        tabular.record('AverageReturn', np.mean(undiscounted_returns))
-        tabular.record('Extras/EpisodeRewardMean',
-                       np.mean(self.episode_reward_mean))
-        tabular.record('NumTrajs', len(paths))
+        self.episode_reward_mean.extend(undiscounted_returns)
+
         tabular.record('Entropy', ent)
         tabular.record('Perplexity', np.exp(ent))
-        tabular.record('StdReturn', np.std(undiscounted_returns))
-        tabular.record('MaxReturn', np.max(undiscounted_returns))
-        tabular.record('MinReturn', np.min(undiscounted_returns))
+        tabular.record('Extras/EpisodeRewardMean',
+                       np.mean(self.episode_reward_mean))
+
+        samples_data = dict(average_return=np.mean(undiscounted_returns))
 
         return samples_data
--- a/src/garage/np/algos/cma_es.py
+++ b/src/garage/np/algos/cma_es.py
@@ -37,12 +37,22 @@ class CMAES(BatchPolopt):
         super().__init__(policy, baseline, discount, max_path_length,
                          n_samples)
         self.env_spec = env_spec
-        self.policy = policy
 
         self.sigma0 = sigma0
 
+        self._es = None
+        self._all_params = None
+        self._cur_params = None
+        self._all_returns = None
+
     def _sample_params(self):
-        return self.es.ask()
+        """Return sample parameters.
+
+        Returns:
+            np.ndarray: A numpy array of parameter values.
+
+        """
+        return self._es.ask()
 
     def train(self, runner):
         """Initialize variables and start training.
@@ -53,16 +63,16 @@ class CMAES(BatchPolopt):
                 such as snapshotting and sampler control.
 
         Returns:
-            The average return in last epoch cycle.
+            float: The average return in last epoch cycle.
 
         """
         init_mean = self.policy.get_param_values()
-        self.es = cma.CMAEvolutionStrategy(init_mean, self.sigma0,
-                                           {'popsize': self.n_samples})
-        self.all_params = self._sample_params()
-        self.cur_params = self.all_params[0]
-        self.policy.set_param_values(self.cur_params)
-        self.all_returns = []
+        self._es = cma.CMAEvolutionStrategy(init_mean, self.sigma0,
+                                            {'popsize': self.n_samples})
+        self._all_params = self._sample_params()
+        self._cur_params = self._all_params[0]
+        self.policy.set_param_values(self._cur_params)
+        self._all_returns = []
 
         return super().train(runner)
 
@@ -72,6 +82,9 @@ class CMAES(BatchPolopt):
         Args:
             itr (int): Iteration number.
             paths (list[dict]): A list of collected paths.
+
+        Returns:
+            float: The average return in last epoch cycle.
 
         """
         paths = self.process_samples(itr, paths)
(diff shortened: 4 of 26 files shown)


src/garage/np/algos/batch_polopt.py
```diff
@@ -133,14 +136,13 @@ def process_samples(self, itr, paths):
 
         tabular.record('Iteration', itr)
```
ryanjulian: I think Itr, AverageDiscountedReturn, EpisodeRewardMean, and Num(Evaluation)Trajs should all go in evaluate_performance

src/garage/experiment/local_runner.py
```diff
@@ -169,6 +171,12 @@ def setup(self, algo, env, sampler_cls=None, sampler_args=None):
             sampler_cls = algo.sampler_cls
         self._sampler = sampler_cls(algo, env, **sampler_args)
 
+        # pylint: disable=import-outside-toplevel
```
ryanjulian: this is ugly. i think that the algorithm should do evaluation sampling instead

@krzentner

src/garage/np/algos/cma_es.py
```diff
@@ -34,14 +34,23 @@ def __init__(self,
...
                  sigma0=1.):
-        super().__init__(policy, baseline, discount, max_path_length,
+        super().__init__(env_spec, policy, baseline, discount, max_path_length,
                          n_samples)
-        self.env_spec = env_spec
-        self.policy = policy
 
         self.sigma0 = sigma0
 
+        self.es = None
```
ryanjulian: please make these all _private while you are here

src/garage/np/algos/off_policy_rl_algorithm.py
```diff
@@ -94,6 +96,10 @@ def train(self, runner):
                 last_return = self.train_once(runner.step_itr,
                                               runner.step_path)
                 if cycle == 0 and self.evaluate:
```
ryanjulian: hmm is `self.evaluate` a constructor parameter? I think we should be evaluating all the time, not much reason to turn if off ever

src/garage/np/algos/off_policy_rl_algorithm.py
```diff
@@ -72,6 +72,8 @@ def __init__(self,
 
         self.sampler_cls = OffPolicyVectorizedSampler
 
+        self.evaluation_batch = None
```
ryanjulian: any good reason this needs to be stored instead of just fetched and sent to evaluate_performance in a single block?

src/garage/sampler/evaluation_sampler.py
```diff
@@ -0,0 +1,61 @@
+"""Sampler for evaluating off-policy algorithms."""
+
+from garage.sampler.base import BaseSampler
+from garage.sampler.utils import rollout
+
+
+class EvaluationSampler(BaseSampler):
```
ryanjulian: maybe this is just SimpleSampler

tests/garage/sampler/test_evaluation_sampler.py
```diff
@@ -0,0 +1,51 @@
...
+from garage.sampler.evaluation_sampler import EvaluationSampler
+from garage.tf.algos import DDPG
+from garage.tf.envs import TfEnv
+from garage.tf.experiment import LocalTFRunner
+from garage.tf.policies import ContinuousMLPPolicy
+from garage.tf.q_functions import ContinuousMLPQFunction
+from tests.fixtures import snapshot_config, TfGraphTestCase
+
+
+class TestEvaluationSampler(TfGraphTestCase):
```
ryanjulian: hmm please make this a unit test, not an integration test.

src/garage/sampler/evaluation_sampler.py
naeioi: I doubt an `EvaluationSampler` external to algorithms can cover all situations. Some challenging examples have been explained in #1091 . So instead of an additional sampler, I propose to just expose `rollout()` to the algorithm. This function should be the barebone to do anything with sampling.

ryanjulian: This is why I proposed doing evaluation sampling in the algorithm. It can use whichever class it wants, including none.

I think Avnish's PR has a class called SimpleSampler which is basically equivalent to this one.

src/garage/np/algos/base.py
```diff
@@ -25,4 +28,24 @@ def train(self, runner):
             The average return in last epoch cycle or None.
 
         """
-        pass
+
+    def evaluate_performance(self, batch):
```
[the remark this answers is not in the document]
ryanjulian: evaluating performance for any algorithm, not just on-policy, requires a batch of full trajectories. on-policy algorithms get that for free. off-policy algorithms need to sample them separately.

`evaluate_performance` *is* a fancy version of `log_diagnostics`, but what's important is that it's shared for all algorithms (nobody should override it, except maybe meta-RL).

having a shared function is the simplest way for all algorithms to use the same evaluation statistics. if we need something fancier, we can always change it later.

src/garage/sampler/evaluation_sampler.py
zequnyu: It's just by our current design, algorithms don't know `env`. Is it a good practice to use `env` passed by `runner.train()`?

zequnyu: So I added the stats, then tried to use `rollout` in algorithms directly, instead of using a `Sampler`. It seems better than before.

src/garage/experiment/local_runner.py
```diff
@@ -132,7 +132,7 @@ def __init__(self, snapshot_config, max_cpus=1):
                                       last_path=None)
 
         self._algo = None
-        self._env = None
+        self.env = None
```
ryanjulian: _env

src/garage/np/algos/base.py
```diff
@@ -25,4 +28,34 @@ def train(self, runner):
...
+                * episode_reward_mean (float): Average episode reward.
+                * average_discounted_return (float): Average discounted return.
+                * undiscounted_returns (numpy.ndarray): A numpy array of shape
+                    :math:`(N \bullet [T])` containing the undiscounted returns
+                    for all time steps in this batch.
+
+        """
+        tabular.record('Evaluation/Iteration', batch['itr'])
+        tabular.record('Evaluation/NumTrajs', batch['num_trajs'])
+        tabular.record('Evaluation/EpisodeRewardMean',
```
ryanjulian: i think you need to update several scripts in test/benchmarks with this new name

ryanjulian: this should be calculated here, from the batch itself

src/garage/np/algos/base.py
```diff
@@ -25,4 +28,34 @@ def train(self, runner):
...
+                       batch['episode_reward_mean'])
+        tabular.record('Evaluation/AverageDiscountedReturn',
```
ryanjulian: this should be calculated here

src/garage/experiment/local_runner.py
zequnyu: It is used in off-policy algorithms base class, or is there other way we can have the `env` inside algorithms?

src/garage/np/algos/base.py
```diff
@@ -25,4 +28,34 @@ def train(self, runner):
...
+                * itr (int): Iteration number.
+                * num_trajs (int): Number of trajectories.
+                * episode_reward_mean (float): Average episode reward.
+                * average_discounted_return (float): Average discounted return.
+                * undiscounted_returns (numpy.ndarray): A numpy array of shape
+                    :math:`(N \bullet [T])` containing the undiscounted returns
+                    for all time steps in this batch.
+
+        """
+        tabular.record('Evaluation/Iteration', batch['itr'])
```
ryanjulian: I think this can just be `Iteration`

src/garage/np/algos/base.py
```diff
@@ -25,4 +28,34 @@ def train(self, runner):
...
 
         """
-        pass
+
+    def evaluate_performance(self, batch):
+        # pylint: disable=no-self-use
+        r"""Evaluate the performance of the algorithm.
+
+        Args:
+            batch (dict): Evaluation trajectories, representing
```
ryanjulian: this is the wrong dtype. this interface should accept TrajectoryBatch (or the equivalent list, if you don't want to adopt it yet).

src/garage/np/algos/base.py
```diff
@@ -25,4 +28,34 @@ def train(self, runner):
...
+                * num_trajs (int): Number of trajectories.
+                * episode_reward_mean (float): Average episode reward.
+                * average_discounted_return (float): Average discounted return.
+                * undiscounted_returns (numpy.ndarray): A numpy array of shape
+                    :math:`(N \bullet [T])` containing the undiscounted returns
+                    for all time steps in this batch.
+
+        """
+        tabular.record('Evaluation/Iteration', batch['itr'])
+        tabular.record('Evaluation/NumTrajs', batch['num_trajs'])
```
ryanjulian: you should calculate this from the batch itself, then it can't be wrong.

src/garage/np/algos/batch_polopt.py
```diff
@@ -30,7 +31,9 @@ class BatchPolopt(RLAlgorithm):
 
     """
 
-    def __init__(self, policy, baseline, discount, max_path_length, n_samples):
+    def __init__(self, env_spec, policy, baseline, discount, max_path_length,
+                 n_samples):
+        self.env_spec = env_spec
```
ryanjulian: where is this used?

zequnyu: Every algorithms have `env_spec` so I think it should be in the base class?

ryanjulian: can you explain why you added this? i don't see where it is used.

src/garage/np/algos/off_policy_rl_algorithm.py
```diff
@@ -169,3 +181,48 @@ def train_once(self, itr, paths):
...
+
+            paths.append(path)
+
+        average_discounted_return = (np.mean(
+            [path['returns'][0] for path in paths]))
+
+        undiscounted_returns = [sum(path['rewards']) for path in paths]
+        self.episode_reward_mean.extend(undiscounted_returns)
+
+        return dict(itr=itr,
```
ryanjulian: the intention of this interface is for you to provide `evaluate_performance` with the raw data and it calculates all these quantities.

src/garage/np/algos/base.py
ryanjulian: you will need to update several parts of the codebase which assume this is named `Itr` -- i think it's Runner or Experiment

ryanjulian: you can ask @naeioi

src/garage/np/algos/batch_polopt.py
zequnyu: Seems to me every algorithms have env_spec so I think it should be in the base class?

src/garage/np/algos/base.py
zequnyu: @naeioi Did you get the point here?

src/garage/experiment/local_runner.py
ryanjulian: oh ok. i think you can also use `runner.env`? if you are going to use the env, you should probably make a copy.

src/garage/np/algos/batch_polopt.py
ryanjulian: oh, are the algorithms actually using `env_spec`? if so, then yes add it to the base class. if it's just there and not used, then remove it. if only one or two algorithms need it, probably don't include it.

src/garage/np/algos/base.py
naeioi: `Iteration` is not used in anywhere anymore, but other metrics are used in benchmarks. See for example this file https://github.com/rlworkgroup/garage/blob/c55683cd383ec00f008e5871fece772e96ee93e2/tests/benchmarks/garage/tf/algos/test_benchmark_ddpg.py#L117-L118

Benchmarks should be the only place these metrics are accessed by their logged named.

zequnyu: Isn't this just label name string? What changes do we need to do?

naeioi: It has to match the name when logged using `tabular.record(name, value)`.

naeioi: I was wrong about `Iteration` not being used anywhere. Actually, there are many references in benchmark files. e.g. in https://github.com/rlworkgroup/garage/blob/c55683cd383ec00f008e5871fece772e96ee93e2/tests/benchmarks/garage/tf/policies/test_benchmark_categorical_cnn_policy.py#L72

They need to be renamed to `Evaluation/Iteration` if the iteration is logged under this name.

ryanjulian: @naeioi do you use `Iteration` to log alternative x-axis?

naeioi: I don't. These are old benchmarks. I use 'TotalEnvSteps' now.

src/garage/np/algos/off_policy_rl_algorithm.py
```diff
@@ -169,3 +178,63 @@ def train_once(self, itr, paths):
 
         """
         raise NotImplementedError
+
+    def _obtain_evaluation_samples(self, batch_size=1000):
```
ryanjulian: the evaluation trajectories must be whole (have `done=True`) or reach `max_path_length`, so I don't think `batch_size` is a useful parameter here. Plus, you use `batch_size` == `max_path_length`, which I think is confusing.

I think the signature of this function should be `_obtain_evaluation_samples(self, num_trajs=100)`. It collects `num_trajs` evaluation trajectories which are either terminated or length `self.max_path_length`. I think `num_trajs = 100` is a sane default. 10 trajectories is kind of low. If 100 is too slow, I recommend going with 30.

src/garage/tf/algos/batch_polopt.py
```diff
@@ -232,40 +236,41 @@ def process_samples(self, itr, paths):
...
         )
 
-        tabular.record('Iteration', itr)
-        tabular.record('AverageDiscountedReturn', average_discounted_return)
-        tabular.record('AverageReturn', np.mean(undiscounted_returns))
-        tabular.record('Extras/EpisodeRewardMean',
-                       np.mean(self.episode_reward_mean))
-        tabular.record('NumTrajs', len(paths))
-        tabular.record('Entropy', ent)
-        tabular.record('Perplexity', np.exp(ent))
```
ryanjulian: it's fine to still log `Entropy` and `Perplexity` here -- did you move these somewhere else?

src/garage/np/algos/base.py
```diff
@@ -25,4 +30,60 @@ def train(self, runner):
...
+                    containing the termination signals for all time steps
+                    in this batch.
+                * env_infos (dict): A dict of numpy arrays arbitrary
+                    environment state information.
+                * agent_infos (numpy.ndarray): A dict of numpy arrays
+                    arbitrary agent state information.
+                * lengths (numpy.ndarray): An integer numpy array
+                    containing the length of each trajectory in this batch.
+                * discount (float): Discount value.
+                * episode_reward_mean (collections.deque): Average episode
```
ryanjulian: this should be calculated from the `rewards` data

src/garage/np/algos/base.py
```diff
@@ -25,4 +30,60 @@ def train(self, runner):
...
+                * terminals (numpy.ndarray): A boolean numpy array
+                    containing the termination signals for all time steps
+                    in this batch.
+                * env_infos (dict): A dict of numpy arrays arbitrary
+                    environment state information.
+                * agent_infos (numpy.ndarray): A dict of numpy arrays
+                    arbitrary agent state information.
+                * lengths (numpy.ndarray): An integer numpy array
+                    containing the length of each trajectory in this batch.
+                * discount (float): Discount value.
```
ryanjulian: this should be a property of `RLAlgorithm`, i.e. `self.discount_factor`

src/garage/np/algos/off_policy_rl_algorithm.py
```diff
@@ -169,3 +178,63 @@ def train_once(self, itr, paths):
...
+        r"""Sample the policy for 10 trajectories and return average values.
+
+        Args:
+            batch_size (int): Number of environment steps in one batch.
+
+        Returns:
+            dict: Evaluation trajectories, representing
+                the best current performance of the algorithm, with keys:
```
ryanjulian: hmm this is wrapped strangely -- isn't there room on the previous line for `the base current performance`?

src/garage/tf/algos/batch_polopt.py
zequnyu: It's right above (L257).

src/garage/np/algos/base.py
zequnyu: It is using self.discount from algorithms. But since base.py is an interface so I passed it.

src/garage/np/algos/base.py
zequnyu: I'm trying to avoid using any self variable (in this case `self.episode_reward_mean`) in the interface. So I passed this from the batch.

ryanjulian: I see, you don't want to change how EpisodeRewardMean is calculated.

In that case, just move logging it back into the algorithm and log it under `Extras/EpisodeRewardMean` as before. This metric is only used for comparing to some openai/baselines results, anyway.

src/garage/experiment/local_runner.py
ryanjulian: Can you ask the sampler or runner for a copy of the environment? The algorithm has access to both of those.

src/garage/np/algos/base.py
ryanjulian: Can you add this as a constructor argument + instance variable + read-only property of RLAlgorithm?

src/garage/np/algos/batch_polopt.py
ryanjulian: How many algorithms do not use env_spec? How many do? When in doubt don't require it. We can improve other parts of the RLAlgorithm interface later.

src/garage/np/algos/base.py
ryanjulian: Okay, leave it for now. Let's focus on getting this merged.

zequnyu:
To summarize, this PR does:

- add an evaluate_performance() method which can be used by all algorithms to report performance
- add a method for OffPolicyRLAalgorithm to obtain whole trajectories
- add dones signals to OnPolicyVectorizedSampler and rollout function
- fix return type of DiscreteQfDerivedPolicy

src/garage/np/algos/batch_polopt.py
zequnyu: For current implementation, all `torch` and `np` algorithms require it. But none seems to actually use it.

src/garage/experiment/local_runner.py
zequnyu: A copy of `gym` environment? From stackoverflow it seems impossible.

ryanjulian: new_env = `pickle.loads(pickle.dumps(env))`

src/garage/np/algos/batch_polopt.py
ryanjulian: TF algorithms use it

ryanjulian: i don't think you should make it part of the base class in this PR, since it's only needed for TF

zequnyu: ok I'll roll it back

PR #1117, opened 2020-01-05
Torch SAC

avnishn:
This PR is open so that contributors can look at the changes made for adding a Torch version of SAC. 

The most recent change which has caused these results: 
https://tensorboard.dev/experiment/RYtOW3uVQjKa0BXKV1AVcA/#scalars
https://tensorboard.dev/experiment/jX46pXb8SvGZzPKMyzKtEA/#scalars
https://tensorboard.dev/experiment/oyU36yz0SVqrSKZbFX58aw/#scalars

The results, while positive, are incredibly positive (on the order of 1e5 average returns, which is highly unlikely). 

The change was disabling the `done` termination signal from gym (refer to changes in `src/garage/envs/base.py`). 

This change caused 2 noticeable differences:
1) The average returns are now 2 orders of magnitude higher than they were before
2) the average returns curve has changed so that the curve no longer plateaus.

I can think of 2 reasons why this has happened.
1) Our off policy vectorized sampler has issues. I think these may be issues with resetting the sampler after every single call to obtain samples. (I have set the `no_reset` flag to false so that it operates in a way that is supposed to reset the envs of the sampler after each call to `obtain_samples`)
2) I am incorrectly ignoring the gym termination signal. (`refer to garage/envs/base.py`)

A little bit of background about the gym termination signal.

- OpenAi gym environments are designed to set the done signal of environment to True every time step is called, after a certain number of certain number of environment steps. This is bad because it changes the MDP of the environment
- For gym environments that are time, limit terminated, when they are terminated, they include an entry in the `env_infos` called `Timelimit.truncated`. This entry is equal to the `not` of the termination signal caused by the environment being in a terminal state (aka the actual/True done signal).
- The change I have made to `garage/envs/base.py` is to set the done signal to the actual/true done signal whenever the environment has been timelimit terminated.


ryanjulian:
Great work! This will be much easier to review if you split it into several PRs.

I think you can at least split it into:
1. Correct the gym termination signal
2. Correctly implement bootstrapping in off-policy algorithms
3. Add primitives for SAC (e.g. TanhNormal distribution and policy)
4. SimpleSampler
5. The SAC replay buffer (why does it need its own)
6. The algorithm itself.

ryanjulian:
As for the very high return, this is probably a measurement issue. You can work on tracking that down while you send in the other PRs.

Perhaps you are doing something like taking the sum of returns at each time step whereas the other implementations use the return at the last time step? That would explain 2 order of magnitude.

src/garage/envs/base.py
```diff
@@ -148,7 +148,10 @@ def step(self, action):
                 debugging, and sometimes learning)
 
         """
-        return self.env.step(action)
+        observation, reward, done, info = self.env.step(action)
+        if 'TimeLimit.truncated' in info:
```
ryanjulian: you should include a comment explaining why we're doing this

src/garage/replay_buffer/sac_replay_buffer.py
```diff
@@ -0,0 +1,51 @@
+import numpy as np
+
+
+class SACReplayBuffer:
```
ryanjulian: how does this differ from any of our other replay buffers?

src/garage/sampler/simple_sampler.py
```diff
@@ -0,0 +1,105 @@
...
+from garage.sampler.sampler import Sampler
+import numpy as np
+import time 
+class SimpleSampler(Sampler):
+    """Generic sampler used for off policy sampling
+
+    Args:
+        agent(Policy): Agent used to select actions.
+        env(gym.Env): Environment to perform actions in.
+    Notes:
```
ryanjulian: its seems like this is redundant with KR's new sampler interface?

src/garage/torch/algos/sac.py
```diff
@@ -0,0 +1,304 @@
...
+import numpy as np
+import torch
+import torch.nn.functional as F
+
+from garage.np.algos.off_policy_rl_algorithm import OffPolicyRLAlgorithm
+from garage.torch.utils import np_to_torch, torch_to_np
+from collections import deque
+
+class SAC(OffPolicyRLAlgorithm):
+    """ A SAC Model in Torch.
```
ryanjulian: please include a citation to the authoritative paper on this algorithm

src/garage/envs/base.py
```diff
@@ -148,7 +148,10 @@ def step(self, action):
...
+                done = not info['TimeLimit.truncated']
```
krzentner: I don't think this is actually correct? If `done` was `True`, but `info['TimeLimit.truncated']` is `False`, `done` should still be `True`. Although this might work if the key is only ever set when truncation occurs. I'd rather not rely on that, so just use something like this:
```python
if info.get('TimeLimit.truncated', False):
    done = False
```

src/garage/replay_buffer/sac_replay_buffer.py
krzentner: I the goal here just to store one sample at a time? Can you just add a method to `PathBuffer` instead that inserts samples outside of a trajectory?

src/garage/sampler/simple_sampler.py
krzentner: I think it's fine to leave this for now. I haven't yet refactored everything else around the sampler interface, and when I do I'll refactor this one to be a minimal implementation of the interface (which I want to have anyways).

src/garage/torch/algos/sac.py
```diff
@@ -0,0 +1,304 @@
...
+        qf1_loss, qf2_loss = self.critic_objective(samples)
+        if(self.gradient_steps == 1):
+            tabular.record("qf_loss/{}".format("qf1_loss"), float(qf1_loss))
+            tabular.record("qf_loss/{}".format("qf2_loss"), float(qf2_loss))
+        elif ((gradient_step % (self.gradient_steps - 1) == 0) and gradient_step != 0):
+            tabular.record("qf_loss/{}".format("qf1_loss"), float(qf1_loss))
+            tabular.record("qf_loss/{}".format("qf2_loss"), float(qf2_loss))
+        self.qf1_optimizer.zero_grad()
+        qf1_loss.backward()
+        # for name, param in self.qf1.named_parameters():
```
krzentner: I'd rather this plotting be controlled by a debug argument or something instead of commented out.

src/garage/torch/distributions/tanh_normal_2.py
```diff
@@ -0,0 +1,70 @@
...
+        if return_pre_tanh_value:
+            action_infos = namedtuple("action_infos", ["pre_tanh_action", "action"])
+            return action_infos(z, torch.tanh(z))
+        else:
+            return torch.tanh(z)
+    @property
+    def loc(self):
+        return self.normal.loc
+
+    # @property
```
krzentner: If there's no reason to avoid this method, uncomment it. Otherwise, delete it and leave a comment explaining why it's not defined.

src/garage/torch/distributions/tanh_normal_distribution.py
```diff
@@ -0,0 +1,88 @@
+import numpy as np
```
krzentner: Wait, so are the contents of this file actually used? I guess they're correct anyways, so we might reasonably want to include them.

tests/garage/sampler/test_simple_sampler.py
```diff
@@ -0,0 +1,66 @@
...
+        self.policy = ScriptedPolicy(
+            scripted_actions=[2, 2, 1, 0, 3, 1, 1, 1, 2, 2, 1, 1, 1, 2, 2, 1])
+        self.algo = Mock(env_spec=self.env.spec,
+                         policy=self.policy,
+                         max_path_length=16)
+
+    def teardown_method(self):
+        self.env.close()
+
+    def test_ray_batch_sampler(self):
```
krzentner: Please rename.

src/garage/torch/policies/tanh_gaussian_mlp_policy_2.py
```diff
@@ -0,0 +1,72 @@
+"""TanhGaussianMLPPolicy."""
+import torch
+
+from garage.torch.modules import TanhGaussianMLPTwoHeadedModule2
+from garage.torch.policies import Policy
+
+
+class TanhGaussianMLPPolicy2(Policy, TanhGaussianMLPTwoHeadedModule2):
+    """
+    GaussianMLPPolicy.
```
krzentner: Please describe in a little more detail what makes this policy unique.

src/garage/torch/policies/tanh_gaussian_mlp_policy.py
```diff
@@ -0,0 +1,72 @@
+"""TanhGaussianMLPPolicy."""
+import torch
+
+from garage.torch.modules import TanhGaussianMLPTwoHeadedModule
+from garage.torch.policies import Policy
+
+
+class TanhGaussianMLPPolicy(Policy, TanhGaussianMLPTwoHeadedModule):
+    """
+    GaussianMLPPolicy.
```
krzentner: Please describe in a little more detail what makes this policy unique.

src/garage/torch/algos/sac.py
```diff
@@ -0,0 +1,304 @@
...
+            samples() - samples recovered from the replay buffer
+        Returns:
+            None
+        """
+
+        obs = samples["observation"]
+
+        # ================update q functions===============#
+        qf1_loss, qf2_loss = self.critic_objective(samples)
+        if(self.gradient_steps == 1):
```
krzentner: Can you refactor out this `if ... elif` logic into a helper method (instead of writing all of the logging code twice)?

src/garage/torch/algos/sac.py
```diff
@@ -0,0 +1,304 @@
...
+            tabular.record("qf_loss/{}".format("qf1_loss"), float(qf1_loss))
```
krzentner: There's no need to format if the value is right there, just write out the full string.

src/garage/replay_buffer/sac_replay_buffer.py
avnishn: This slipped into the PR because I was testing primitives to see if they were broken. I can remove this.

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,305 @@
...
+        parameter broadcasting in `__init__.py`, when an instance is first
+        created.
+
+        Args:
+            batch_shape (torch.Size): the desired expanded size.
+            _instance(instance): new instance provided by subclasses that
+                need to override `.expand`.
+
+        Returns:
+            Instance: New distribution instance with batch dimensions expanded
```
ryanjulian: Instance is not a type. this returns TanhNormal

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,305 @@
...
+
+        """
+        clip_up = (x > upper).float()
+        clip_low = (x < lower).float()
+        with torch.no_grad():
+            clip = ((upper - x) * clip_up + (lower - x) * clip_low)
+        return x + clip
+
+    def __str__(self):
+        """Returns the Name of the class.
```
ryanjulian: that's not actually what str does. https://rszalski.github.io/magicmethods/

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,305 @@
...
+
+    def __repr__(self):
+        """Returns the parameterization of the distribution.
+
+        Returns:
+            str: The parameterization of the distribution and underlying
+                distribution.
+
+        """
+        return ('TanhNormal Dist: Mean {}, stdev {}\n \
```
ryanjulian: probably better to keep this in a single line, and to do something succinct like

`TanhNormal<Normal(loc={}, scale={})>`

if torch.distributions already has conventions for `__str__` and `__repr__`, follow how they do it.

ryanjulian:
any progress on splitting this change? it is basically unmergable this big.

PR #1118, opened 2020-01-07
Disable Gym Time limits by default

avnishn:
A little bit of background about the gym termination signal.

OpenAi gym environments are designed to set the done signal
of environment to True every time step is called, after a
certain number of certain number of environment steps.
This is bad because it changes the MDP of the environment
For gym environments that are time, limit terminated, when
they are terminated, they include an entry in the env_infos
called Timelimit.truncated. This entry is equal to the not
of the termination signal caused by the environment being
in a terminal state (aka the actual/True done signal).
The change I have made to garage/envs/base.py is to set
the done signal to the actual/true done signal whenever
the environment has been timelimit terminated.

This PR sets the done signal to be True only if caused by an
environment termination, and not a time limit termination. 
The time limit termination signal will be saved inside env_infos as 
'TimeLimitTerminated'.

### DIFF
--- a/src/garage/envs/base.py
+++ b/src/garage/envs/base.py
@@ -148,7 +148,19 @@ class GarageEnv(gym.Wrapper):
                 debugging, and sometimes learning)
 
         """
-        return self.env.step(action)
+        observation, reward, done, info = self.env.step(action)
+        # gym envs that are wrapped in TimeLimit wrapper modify
+        # the done/termination signal to be true whenever a time
+        # limit expiration occurs. The following statement sets
+        # the done signal to be True only if caused by an
+        # environment termination, and not a time limit
+        # termination. The time limit termination signal
+        # will be saved inside env_infos as
+        # 'GarageEnv.TimeLimitTerminated'
+        if 'TimeLimit.truncated' in info:
+            info['GarageEnv.TimeLimitTerminated'] = done  # done = True always
+            done = not info['TimeLimit.truncated']
+        return observation, reward, done, info
 
     def __getstate__(self):
         """See `Object.__getstate__.
--- a/tests/garage/envs/test_garage_env.py
+++ b/tests/garage/envs/test_garage_env.py
@@ -20,3 +20,12 @@ class TestGarageEnv:
         assert garage_env.env.viewer is not None
         garage_env.close()
         assert garage_env.env.viewer is None
+
+    def test_time_limit_env(self):
+        garage_env = GarageEnv(env_name='Pendulum-v0')
+        garage_env.reset()
+        for _ in range(200):
+            _, _, done, info = garage_env.step(
+                garage_env.spec.action_space.sample())
+        assert not done and info['TimeLimit.truncated']
+        assert info['GarageEnv.TimeLimitTerminated']


naeioi:
Let me clarify:
- A time-limit wrapper returns done=True when a pre-specified time threshod is reached. 
- So when such a wrapped env returns done=True, it maybe because of either of
  1. Steps exceeded time-limit.
  2. The underlying env terminated. 
- This PR overrides behavior i., by returning done=False in such case, and sets env_info.TimeLimitTerminated=True.

Can you provide a use case of this patch?

src/garage/envs/base.py
```diff
@@ -148,7 +148,19 @@ def step(self, action):
...
+        # the done/termination signal to be true whenever a time
+        # limit expiration occurs. The following statement sets
+        # the done signal to be True only if caused by an
+        # environment termination, and not a time limit
+        # termination. The time limit termination signal
+        # will be saved inside env_infos as
+        # 'TimeLimitTerminated'
+        if 'TimeLimit.truncated' in info:
+            info['TimeLimitTerminated'] = done
+            done = not info['TimeLimit.truncated']
```
krzentner: If `TimeLimit.truncated` is ever set to `False`, then this would be wrong. Please change the logic so it's correct in that case.

avnishn:
@naeioi 

So what you said is good up until your last point. The PR overides behavior by overiding the time limit termination behavior. A case could be encountered where there is a time limit interrupt and there is a terminal environment state. In that case done = true will still be returned, because of the terminal environment state. 

A use case of this patch would be in off policy algorithms where the Q functions are learned. The target computed used to update the Q function relies on knowing when states are terminal or not. If a state is mistaken as terminal due to the done signal being True because of a time limit termination, then this messes up the learning of the Q function. Did that answer your question?

src/garage/envs/base.py
avnishn: I don't think this is the case. Looking at the code of the time limit wrapper, it seems that TimeLimit.truncated equals the Not of the original done signal caused by any environment termination. Is there something I'm missing here @krzentner ?

avnishn: I don't think this is the case. Looking at the code of the time limit wrapper, it seems that TimeLimit.truncated equals the Not of the original done signal caused by any environment termination. Is there something I'm missing here @krzentner ?

src/garage/envs/base.py
```diff
@@ -148,7 +148,19 @@ def step(self, action):
                 debugging, and sometimes learning)
 
         """
-        return self.env.step(action)
+        observation, reward, done, info = self.env.step(action)
+        # gym envs that are wrapped in timelimit wrapper modify
```
ryanjulian: TimeLimit

src/garage/envs/base.py
```diff
@@ -148,7 +148,19 @@ def step(self, action):
...
+        # the done/termination signal to be true whenever a time
+        # limit expiration occurs. The following statement sets
+        # the done signal to be True only if caused by an
+        # environment termination, and not a time limit
+        # termination. The time limit termination signal
+        # will be saved inside env_infos as
+        # 'TimeLimitTerminated'
```
ryanjulian: can you make this `GarageEnv.Terminated`? It will help downstream users figure out where the key came from.

src/garage/envs/base.py
ryanjulian: can you provide a link?

src/garage/envs/base.py
ryanjulian: perhaps the more useful signal would be `GarageEnv.terminal_corrected`

src/garage/envs/base.py
ryanjulian: For those following along at home: https://github.com/openai/gym/blob/master/gym/wrappers/time_limit.py#L19

I agree that the logic is weird if an environment reached both termination and the time limit in the same step, but due to the implementation of `TimeLimit`, `not info[TimeLimit.truncated]` is the only way to recover the original `done` signal.

avnishn: Oh yes, thanks, that is the link. It is the reference to the misunderstanding about `info[TimeLimit.truncated] ` @krzentner

src/garage/envs/base.py
avnishn: `GarageEnv.terminal_corrected` implies we are passing the corrected terminal value in the dict. Aren't we already doing this with the Done value?

I will change the signal name from `TimeLimitTerminated` to `GarageEnv.Terminated`

avnishn:
I found the place where tf.agents and softlearning are handling it:

tf agents: https://github.com/tensorflow/agents/blob/96e7b31d71023217ffe9e7febff2ef2c5db88513/tf_agents/environments/suite_gym.py#L18

soft learning:
https://github.com/rail-berkeley/softlearning/blob/46374df0294b9b5f6dbe65b9471ec491a82b6944/softlearning/environments/adapters/gym_adapter.py#L82

ryanjulian:
@Mergifyio rebase

src/garage/envs/base.py
ryanjulian: `TimeLimit.truncated` is already in `info` and can tell the user what the original value was. I wanted to provide a signal in `info` that done has been modified.

ryanjulian: if you simply want to store the original value (as here) then i suggest you put it at `GarageEnv.TimeLimit.done` and also add `GarageEnv.done_corrected = True`

avnishn:
@Mergifyio rebase

src/garage/envs/base.py
krzentner: Okay, I guess that makes sense (even if it's weird).

ahtsan: Want to clarify -- does `GarageEnv.TimeLimitTerminated` tell us it reached time limit or not and now `done` will only be true if it's from env termination?

PR #1124, opened 2020-01-08
New worker for PEARL

lywong92:
Adding a new worker for PEARL sampling because it needs to be able to store context and resample belief in the policy when obtaining samples.


src/garage/replay_buffer/meta_replay_buffer.py
```diff
@@ -0,0 +1,145 @@
...
+            path (dict): Dictionary containing path information.
+
+        """
+        for _, (obs, action, reward, next_obs, terminal) in enumerate(
+                zip(path['observations'], path['actions'], path['rewards'],
+                    path['next_observations'], path['terminals'])):
+            self.add_sample(obs, action, reward, terminal, next_obs)
+        self.terminate_episode()
+
+    def size(self):
```
ryanjulian: is this the size in samples or paths?

src/garage/replay_buffer/meta_replay_buffer.py
```diff
@@ -0,0 +1,145 @@
...
+
+        Returns:
+            int: Current size of buffer.
+
+        """
+        return self._size
+
+    def _advance(self):
+        """Increment size of buffer."""
+        self._top = (self._top + 1) % self._max_replay_buffer_size
```
ryanjulian: i think you should document in the class docstring that this is a FIFO replay buffer

src/garage/replay_buffer/meta_replay_buffer.py
```diff
@@ -0,0 +1,145 @@
...
+        if self._size < self._max_replay_buffer_size:
+            self._size += 1
+
+    def terminate_episode(self):
```
ryanjulian: hmm...what is the purpose of this API?

@krzentner

src/garage/replay_buffer/meta_replay_buffer.py
```diff
@@ -0,0 +1,145 @@
...
+        self._cur_episode_start = self._top
+
+    def clear(self):
+        """Clear buffer."""
+        self._top = 0
+        self._size = 0
+        self._episode_starts = []
+        self._cur_episode_start = 0
+
+    def sample_data(self, indices):
```
ryanjulian: this seems to retrieve data (`indices` are known) rather than sample it). can we call this `get`?

src/garage/replay_buffer/meta_replay_buffer.py
```diff
@@ -0,0 +1,145 @@
...
+        """
+        return dict(
+            observations=self._observations[indices],
+            actions=self._actions[indices],
+            rewards=self._rewards[indices],
+            terminals=self._terminals[indices],
+            next_observations=self._next_obs[indices],
+        )
+
+    def random_batch(self, batch_size):
```
ryanjulian: this is what is conventionally called `sample`

src/garage/replay_buffer/meta_replay_buffer.py
```diff
@@ -0,0 +1,145 @@
...
+            batch_size (int): Size of random batch.
+
+        Returns:
+            dict: Dictionary containing random batch.
+
+        """
+        indices = np.random.randint(0, self._size, batch_size)
+        return self.sample_data(indices)
+
+    def random_sequence(self, batch_size):
```
ryanjulian: it seems that this samples trajectories? is there some constraint on these snippets? do they need to be full trajectories? some minimum size?

src/garage/replay_buffer/meta_replay_buffer.py
```diff
@@ -0,0 +1,145 @@
+"""A replay buffer memory for meta RL."""
```
ryanjulian: @krzentner please take a look here to make sure this is consistent with your API for samples vs paths

src/garage/replay_buffer/multi_task_replay_buffer.py
```diff
@@ -0,0 +1,145 @@
...
+    Args:
+        max_replay_buffer_size (int): Maximum buffer size.
+        env (garage.envs.GarageEnv): Meta environment.
+        tasks (list): A list of task indices.
+
+    """
+
+    def __init__(self, max_replay_buffer_size, env, tasks):
+
+        self.env = env
```
ryanjulian: why does this need env?

src/garage/replay_buffer/multi_task_replay_buffer.py
```diff
@@ -0,0 +1,145 @@
+"""A replay buffer memory for meta-RL."""
```
ryanjulian: @krzentner PTAL ASAP

src/garage/sampler/in_place_sampler.py
```diff
@@ -0,0 +1,73 @@
+# pylint: disable=unnecessary-pass
```
ryanjulian: does something bad happen if you remove these?

src/garage/replay_buffer/meta_replay_buffer.py
krzentner: So, the design of this replay buffer is fine, but I don't understand why it's necessary at all.

What does this buffer implement that `PathBuffer` doesn't?
If it's something small (like adding individual samples), then I can just add it to `PathBuffer`.
If it's something really big I somehow can't see in this code, it really needs to be called out in the class docstring.

src/garage/replay_buffer/multi_task_replay_buffer.py
```diff
@@ -0,0 +1,145 @@
...
+    def clear_buffer(self, task):
+        """Clear a task buffer by its index.
+
+        Args:
+            task (int): Task index.
+
+        """
+        self.task_buffers[task].clear()
+
+    def random_batch(self, task, batch_size, sequence=False):
```
krzentner: What's the rationale behind this method? If seems confusing that it wraps both `random_sequence` and `random_batch` but is called `random_sequence`.

src/garage/replay_buffer/multi_task_replay_buffer.py
krzentner: I don't fully understand the role of this class in the context of PEARL yet. It seems to be fine to have a replay buffer that keeps each tasks' samples in a separate buffer, but I don't quite get why it's useful. This seems like a reasonable way to implement it.

src/garage/replay_buffer/meta_replay_buffer.py
lywong92: It does mostly the same thing as PathBuffer but with a different structure of how data is stored. It has the `random_sequence` method which samples trajectories, and it's called when a recurrent encoder is used in PEARL. I took this from the oyster repo.

src/garage/replay_buffer/meta_replay_buffer.py
lywong92: samples

src/garage/replay_buffer/meta_replay_buffer.py
lywong92: It marks the beginning of every path because the buffer stores the samples individually.

src/garage/replay_buffer/meta_replay_buffer.py
lywong92: I'm not sure. This is the sampling method called when a recurrent encoder is used in PEARL, but I don't see any examples in the repo where they use a recurrent encoder, so I didn't look into this.

src/garage/replay_buffer/multi_task_replay_buffer.py
lywong92: The algorithm keeps a buffer for each task, and I think this looks cleaner. I just followed the implementation on oyster, but I suppose I can keep a list of buffers in the algo without this buffer class.

src/garage/replay_buffer/multi_task_replay_buffer.py
lywong92: If a recurrent encoder is used in PEARL, then `sequence` is True, and this samples trajectories instead of random transitions.

src/garage/replay_buffer/multi_task_replay_buffer.py
ryanjulian: this looks generally fine to me, but how does it work for meta-RL on continuous spaces? i.e. HalfCheetahRandomVel has "infinite" tasks, because you sample a new meta-batch of environments (velocities) on every meta-batch

src/garage/replay_buffer/multi_task_replay_buffer.py
ryanjulian: i think that this should be a different method. methods should not change their return types based on arguments.

src/garage/replay_buffer/meta_replay_buffer.py
ryanjulian: @krzentner please make the call here

src/garage/replay_buffer/meta_replay_buffer.py
ryanjulian: hmm please note this in the docstring.

src/garage/replay_buffer/meta_replay_buffer.py
ryanjulian: @krzentner please make a call here

src/garage/torch/policies/deterministic_mlp_policy.py
```diff
@@ -63,6 +64,7 @@ def get_action(self, observation):
                         distribution
 
         """
+        observation = torch.Tensor(observation)
```
ryanjulian: i think it's fair for this primitive to assume that its input is a `torch.Tensor`. it even says so in the docstring.

is there a reason you can't do this upstream of this primitive?

ryanjulian:
Assuming everything passes muster, I'm OK merging this now if KR is okay with it. Clearly, unifying sampling for meta-RL is an area we need to work on.

src/garage/sampler/pearl_sampler.py
```diff
@@ -0,0 +1,73 @@
...
+        """
+        assert max_samples < np.inf or max_trajs < np.inf
+        paths = []
+        n_steps_total = 0
+        n_trajs = 0
+
+        while n_steps_total < max_samples and n_trajs < max_trajs:
+            path = rollout(self.env,
+                           self.policy,
+                           max_path_length=self.max_path_length,
```
avnishn: Do we ever want max path length to be infinite? This would make it such that there's never a reset in some environments that definitely need a reset after a certain number of steps. For example, standard mujoco environments should have max path length set to 1000.

src/garage/sampler/pearl_sampler.py
```diff
@@ -0,0 +1,73 @@
...
+                       max_samples=np.inf,
+                       max_trajs=np.inf,
+                       accum_context=True,
+                       resample=1):
+        """Obtain samples in the environment up to max_samples or max_trajs.
+
+        Args:
+            max_samples (int): Maximum number of samples.
+            max_trajs (int):  Maximum number of trajectories.
+            accum_context (bool): Whether or not to accumulate the collected
```
avnishn: I'm unsure of what is meant by this. Is this terminology specific to PEARL

src/garage/sampler/pearl_sampler.py
```diff
@@ -0,0 +1,73 @@
...
+
+        while n_steps_total < max_samples and n_trajs < max_trajs:
+            path = rollout(self.env,
+                           self.policy,
+                           max_path_length=self.max_path_length,
+                           accum_context=accum_context)
+
+            # save the latent context that generated this trajectory
+            if accum_context:
+                path['context'] = self.policy.z.detach().cpu().numpy()
```
avnishn: does `self.policy.z` ever update during the duration of the sampling loop? If not, can we store this value instead of querying ever time.

src/garage/sampler/utils.py
```diff
@@ -13,7 +13,8 @@ def rollout(env,
             max_path_length=np.inf,
             animated=False,
             speedup=1,
-            deterministic=False):
+            deterministic=False,
+            accum_context=False):
```
avnishn: I think we should create a separate rollout function that has accum_context, and keep it inside the file `pearl_sampler.py` if accumulation context is only used in the context of PEARL

src/garage/sampler/utils.py
```diff
@@ -75,10 +83,16 @@ def rollout(env,
             timestep = 0.05
             time.sleep(timestep / speedup)
 
+    observations = np.array(observations)
+    next_observations = np.vstack(
```
avnishn: Does modifying the return still allow you to pass all of the tests? If so, then this is fine. if not, create a separate rollout function specifically for pearl.

src/garage/torch/policies/deterministic_mlp_policy.py
avnishn: +1

src/garage/sampler/utils.py
ryanjulian: i agree. @krzentner would have a better idea of what to do.

src/garage/sampler/pearl_sampler.py
ryanjulian: task context is the information given to you which allows you to identify the task.

src/garage/sampler/pearl_sampler.py
ryanjulian: i'm not sure 

> 1. Based on comments and code similarities, not sure if we need a new meta-rl replay buffer. Perhaps a refactor using an existing replay buffer can come later if we decide.
> 2. What is the reasoning behind a multi task replay buffer for PEARL? I saw your comment:
> 
> > The algorithm keeps a buffer for each task, and I think this looks cleaner. I just followed the implementation on oyster, but I suppose I can keep a list of buffers in the algo without this buffer class.
> 
> I'm with you on keeping a list of buffers as opposed to a separate MT replay buffer
> 
> 1. Can you create your own rollout function separate from the sampler/utils `rollout` function.
>    It seems that you have modified the return types of the `rollout function` and added some other features. A question I find myself asking is "Will the added functionality be used by algorithms besides PEARL?"

Let's keep the major changes to a minimum here. We can make better changes once everything is merged.

Since many of the ways of improving this code involve unifying it with other algorithms, it is better to merge high-quality working code than to insist that it be high-quality, working, and unified.

avnishn: Kk

avnishn: Kk

src/garage/sampler/__init__.py
```diff
@@ -16,8 +17,19 @@
...
+    'RaySampler',
+    'SamplerWorker',
+    'ParallelVecEnvExecutor',
+    'VecEnvExecutor',
+    'OffPolicyVectorizedSampler',
+    'OnPolicyVectorizedSampler',
+    'WorkerFactory',
+    'Worker',
+    'DefaultWorker',
+    'PEARLSampler',
```
ryanjulian: can this be achieved by adding a PEARLWorker class @krzentner ?

Anson does a similar pattern for RL2: He implements an RL2 worker which lives in `garage.tf.algos.rl2` rather than the central sampler package.

src/garage/sampler/pearl_sampler.py
```diff
@@ -0,0 +1,156 @@
+# pylint: disable=unnecessary-pass
```
ryanjulian: you can deal with this by just omitting `pass` from your empty methods.

src/garage/sampler/__init__.py
krzentner: It can definitely be achieved with that.
Alternatively, just put z in `agent_infos`. It's there for a reason.

src/garage/sampler/__init__.py
```diff
@@ -16,8 +17,19 @@
...
+    'RaySampler',
+    'SamplerWorker',
+    'ParallelVecEnvExecutor',
+    'VecEnvExecutor',
+    'OffPolicyVectorizedSampler',
+    'OnPolicyVectorizedSampler',
+    'WorkerFactory',
+    'Worker',
+    'DefaultWorker',
+    'PEARLWorker',
```
ryanjulian: please put this class in `src/garage/torch/algos/pearl.py` -- since it is only used by the PEARL algorithm.

Its import path will be `from garage.torch.algos.pearl import PEARLWorker`

ryanjulian:
Please address my comments before submit. @krzentner should also probably take a look.

src/garage/torch/algos/pearl.py
```diff
@@ -0,0 +1,99 @@
...
+        Returns:
+            garage.TrajectoryBatch: The collected trajectory.
+
+        """
+        self.start_rollout()
+        while not self.step_rollout():
+            pass
+        self._agent_infos['context'] = [self.agent.z.detach().cpu().numpy()
+                                        ] * self._max_path_length
+        self.agent.sample_from_belief()
```
krzentner: It might make more sense to put this at the start, since agent updates happen in between calls to this method.

Issue #1125, opened 2020-01-09
Refactor self.evaluate to class property


zequnyu:
https://github.com/rlworkgroup/garage/pull/1116#discussion_r363103106

https://github.com/rlworkgroup/garage/pull/1126#discussion_r365560995

zequnyu:
Closed by #1139.

PR #1126, opened 2020-01-10
Remove self.evaluate from algorithms

zequnyu:
#1125.

Also fix a minor bug (epoch should be int).

### DIFF
--- a/src/garage/np/algos/off_policy_rl_algorithm.py
+++ b/src/garage/np/algos/off_policy_rl_algorithm.py
@@ -65,7 +65,6 @@ class OffPolicyRLAlgorithm(RLAlgorithm):
         self.min_buffer_size = min_buffer_size
         self.rollout_batch_size = rollout_batch_size
         self.reward_scale = reward_scale
-        self.evaluate = False
         self.input_include_goal = input_include_goal
         self.smooth_return = smooth_return
         self.max_path_length = max_path_length
@@ -94,7 +93,7 @@ class OffPolicyRLAlgorithm(RLAlgorithm):
                 runner.step_path = runner.obtain_samples(runner.step_itr)
                 last_return = self.train_once(runner.step_itr,
                                               runner.step_path)
-                if cycle == 0 and self.evaluate:
+                if cycle == 0:
                     self.evaluate_performance(
                         runner.step_itr,
                         self._obtain_evaluation_samples(runner.get_env_copy()))
--- a/src/garage/tf/algos/ddpg.py
+++ b/src/garage/tf/algos/ddpg.py
@@ -109,6 +109,8 @@ class DDPG(OffPolicyRLAlgorithm):
         self.target_policy = policy.clone('target_policy')
         self.target_qf = qf.clone('target_qf')
 
+        self._first_min_buffer_size_done = False
+
         super(DDPG, self).__init__(env_spec=env_spec,
                                    policy=policy,
                                    qf=qf,
@@ -271,7 +273,7 @@ class DDPG(OffPolicyRLAlgorithm):
         for _ in range(self.n_train_steps):
             if (self.replay_buffer.n_transitions_stored >=
                     self.min_buffer_size):
-                self.evaluate = True
+                self._first_min_buffer_size_done = True
                 qf_loss, y_s, qval, policy_loss = self.optimize_policy(
                     itr, paths)
 
@@ -283,10 +285,8 @@ class DDPG(OffPolicyRLAlgorithm):
         if itr % self.steps_per_epoch == 0:
             logger.log('Training finished')
 
-            if self.evaluate:
+            if self._first_min_buffer_size_done:
                 tabular.record('Epoch', epoch)
-                tabular.record('AverageReturn', np.mean(self.episode_rewards))
-                tabular.record('StdReturn', np.std(self.episode_rewards))
                 tabular.record('Policy/AveragePolicyLoss',
                                np.mean(self.episode_policy_losses))
                 tabular.record('QFunction/AverageQFunctionLoss',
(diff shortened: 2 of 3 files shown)


src/garage/tf/algos/ddpg.py
```diff
@@ -283,25 +282,24 @@ def train_once(self, itr, paths):
...
-                               np.mean(np.abs(self.epoch_qs)))
-                tabular.record('QFunction/AverageY', np.mean(self.epoch_ys))
-                tabular.record('QFunction/MaxY', np.max(self.epoch_ys))
-                tabular.record('QFunction/AverageAbsY',
-                               np.mean(np.abs(self.epoch_ys)))
-                if self.input_include_goal:
-                    tabular.record('AverageSuccessRate',
-                                   np.mean(self.success_history))
+            tabular.record('Epoch', epoch)
+            tabular.record('AverageReturn', np.mean(self.episode_rewards))
```
ryanjulian: shouldn't AverageReturn and StdReturn come from `RLAlgorithm.evaluate_performance`?

src/garage/tf/algos/dqn.py
```diff
@@ -191,32 +191,28 @@ def train_once(self, itr, paths):
...
-                tabular.record('Epoch', epoch)
-                tabular.record('AverageReturn', np.mean(self.episode_rewards))
-                tabular.record('StdReturn', np.std(self.episode_rewards))
-                tabular.record('Episode100RewardMean', mean100ep_rewards)
-                tabular.record('{}/Episode100LossMean'.format(self.qf.name),
-                               mean100ep_qf_loss)
+            mean100ep_rewards = round(np.mean(self.episode_rewards[-100:]), 1)
+            mean100ep_qf_loss = np.mean(self.episode_qf_losses[-100:])
+            tabular.record('Epoch', epoch)
+            tabular.record('AverageReturn', np.mean(self.episode_rewards))
```
ryanjulian: shouldn't AverageReturn and StdReturn come from the new interface?

zequnyu: Actually I need some clarification. It seems like these numbers are wrongly calculated and we should use the independently collected trajectories to calculate?

ryanjulian: Yes, exactly.

src/garage/tf/algos/dqn.py
```diff
@@ -191,32 +191,28 @@ def train_once(self, itr, paths):
...
                     self.min_buffer_size):
-                self.evaluate = True
                 qf_loss = self.optimize_policy(epoch, None)
                 self.episode_qf_losses.append(qf_loss)
 
-        if self.evaluate:
-            if itr % self.target_network_update_freq == 0:
-                self._qf_update_ops()
+        if itr % self.target_network_update_freq == 0:
+            self._qf_update_ops()
```
ahtsan: A flag is needed so qf is not updated during the first `min_buffer_size` steps. Previously it's guarded by `self.evaluate`.

src/garage/np/algos/off_policy_rl_algorithm.py
```diff
@@ -94,7 +93,7 @@ def train(self, runner):
                 runner.step_path = runner.obtain_samples(runner.step_itr)
                 last_return = self.train_once(runner.step_itr,
                                               runner.step_path)
-                if cycle == 0 and self.evaluate:
```
zequnyu: @ahtsan Is this the correct way to use `self.evaluate`?

ahtsan: I think it's fine without `self.evaluate` here because it only evaluates performance. The cases that we should put a guard are those modifying the policy/Q function, like the one I commented in `dqn.py` where we shouldn't update qf during first `min_buffer_size` steps.

src/garage/tf/algos/ddpg.py
```diff
@@ -271,7 +271,6 @@ def train_once(self, itr, paths):
         for _ in range(self.n_train_steps):
             if (self.replay_buffer.n_transitions_stored >=
                     self.min_buffer_size):
-                self.evaluate = True
```
ahtsan: You might want to keep this flag here

src/garage/tf/algos/dqn.py
```diff
@@ -191,32 +191,28 @@ def train_once(self, itr, paths):
...
         paths = self.process_samples(itr, paths)
-        epoch = itr / self.steps_per_epoch
+        epoch = itr // self.steps_per_epoch
 
         self.episode_rewards.extend(paths['undiscounted_returns'])
         last_average_return = np.mean(self.episode_rewards)
         for _ in range(self.n_train_steps):
             if (self.replay_buffer.n_transitions_stored >=
                     self.min_buffer_size):
-                self.evaluate = True
```
ahtsan: and this one too

src/garage/tf/algos/dqn.py
```diff
@@ -191,32 +191,28 @@ def train_once(self, itr, paths):
...
                 qf_loss = self.optimize_policy(epoch, None)
                 self.episode_qf_losses.append(qf_loss)
 
-        if self.evaluate:
```
ahtsan: and this

src/garage/tf/algos/ddpg.py
ahtsan: maybe you should change the name to avoid confusion, like `self.start_train` or sth like that

src/garage/tf/algos/dqn.py
ahtsan: FYI: ddpg uses soft-update so it updates QFunction every iterations after `min_buffer_size.
https://github.com/rlworkgroup/garage/blob/779f87fc5bdab93b45cdd4a4d0d5e2b6949e6cda/src/garage/tf/algos/ddpg.py#L272
However, dqn uses hard-update it only updates once, hence an extra guard is required.

zequnyu:
I'm going to close it and cover this part in a new PR which updates all algorithms usage to `evaluate_performance`.

PR #1127, opened 2020-01-10
Sampler for RL2

ahtsan:


src/garage/sampler/rl2_sampler.py
```diff
@@ -0,0 +1,195 @@
...
+
+class RL2Sampler(BatchSampler):
+    """BatchSampler which uses VecEnvExecutor to run multiple environments.
+
+    This sampler is for RL^2. See https://arxiv.org/pdf/1611.02779.pdf.
+
+    Args:
+        algo (garage.np.algos.RLAlgorithm): An algorithm instance.
+        env (garage.envs.GarageEnv): An environement instance.
+        meta_batch_size (int): Number of meta batch for sampling.
```
naeioi: Should it be `Number of environments in one meta batch`?

src/garage/sampler/rl2_sampler.py
```diff
@@ -0,0 +1,195 @@
...
+            'Number of vectorized '
+            'environments should be a multiple of meta_batch_size.')
+        envs_per_worker = n_envs // self._meta_batch_size
+
+        tasks = self.env.sample_tasks(self._meta_batch_size)
+        vec_envs = []
+        for task in tasks:
+            vec_env = pickle.loads(pickle.dumps(self.env))
+            vec_env.set_task(task)
+            vec_envs.extend([vec_env] * envs_per_worker)
```
naeioi: `[vec_env] * envs_per_worker` This will make a list of references to vec_env, not actually copies. Is it intended?

src/garage/sampler/rl2_sampler.py
```diff
@@ -0,0 +1,195 @@
...
+                  agent info, value being a numpy.ndarray with shape
+                  :math:`[N, S^*]`. One example is "prev_action", which is used
+                  for recurrent policy as previous action input, merged with
+                  the observation input as the state input.
+
+        """
+        logger.log('Obtaining samples for iteration %d...' % itr)
+
+        if batch_size is None:
+            batch_size = (self._episode_per_task * self.algo.max_path_length *
```
naeioi: Shall we force `batch_size` to be equal to `self._episode_per_task * self.algo.max_path_length * self._n_envs` in this context?

src/garage/sampler/rl2_sampler.py
```diff
@@ -0,0 +1,195 @@
...
+                        dones=[],
+                        env_infos=[],
+                        agent_infos=[],
+                    )
+                running_paths[idx]['observations'].append(observation)
+                running_paths[idx]['actions'].append(action)
+                running_paths[idx]['rewards'].append(reward)
+                running_paths[idx]['dones'].append(done)
+                running_paths[idx]['env_infos'].append(env_info)
+                running_paths[idx]['agent_infos'].append(agent_info)
```
naeioi: I think you mentioned in Slack that additional data (something like `task_id`, IIRC) is needed so that an algorithm can split paths from different tasks. Since you are not adding it here, do you find another workaround?

src/garage/sampler/rl2_sampler.py
ahtsan: I think it is the number of meta batch. What you suggested sounds more like the number of actual vectorized environments instances used for sampling, for each meta batch.

src/garage/sampler/rl2_sampler.py
ahtsan: Good catch, it's indeed a careless bug. I wasn't aware of it since I have been using `envs_per_work=1`. Thanks for pointing out.

src/garage/sampler/rl2_sampler.py
ahtsan: Yes, that's a good idea. How we should enforce it is another issue tho, since batch_size is a required argument. If we make it optional / none, we can set it to `self._episode_per_task * self.algo.max_path_length * self._n_envs`

src/garage/sampler/rl2_sampler.py
ahtsan: Yes. In order to split the paths, I construct an array by `np.cumsum([len(path['rewards']) for path in paths])` so I can split the paths by `np.split` using this array.

src/garage/sampler/rl2_sampler.py
```diff
@@ -0,0 +1,195 @@
...
+        meta_batch_size (int): Number of meta batch for sampling.
+        episode_per_task (int): Number of episode for each task.
+        n_envs (int): Number of environment instances to setup.
+            This parameter has effect on sampling performance.
+
+    """
+
+    def __init__(self,
+                 algo,
+                 env,
```
ryanjulian: envs? shouldn't this be multiple environments?

src/garage/sampler/rl2_sampler.py
```diff
@@ -0,0 +1,195 @@
...
+        self._episode_per_task = episode_per_task
+        self._vec_env = None
+        self._env_spec = self.env.spec
+
+    def start_worker(self):
+        """Start workers."""
+        n_envs = self._n_envs
+        assert n_envs >= self._meta_batch_size, (
+            'Number of vectorized environments '
+            'should be at least meta_batch_size.')
```
ryanjulian: why? if it's less than meta-batch size can't i just run them in series?

src/garage/sampler/rl2_sampler.py
```diff
@@ -0,0 +1,195 @@
...
+        assert n_envs % self._meta_batch_size == 0, (
+            'Number of vectorized '
```
ryanjulian: again, this seems like an artificial requirement.

src/garage/sampler/rl2_sampler.py
```diff
@@ -0,0 +1,195 @@
...
+        n_envs (int): Number of environment instances to setup.
+            This parameter has effect on sampling performance.
+
+    """
+
+    def __init__(self,
+                 algo,
+                 env,
+                 meta_batch_size,
+                 episode_per_task,
```
ryanjulian: why can't i decide this when i call `obtain_samples`? i think you are putting logic in the sample which should instead be in the algorithm or runner

src/garage/sampler/rl2_sampler.py
```diff
@@ -0,0 +1,195 @@
...
+        episode_per_task (int): Number of episode for each task.
+        n_envs (int): Number of environment instances to setup.
+            This parameter has effect on sampling performance.
+
+    """
+
+    def __init__(self,
+                 algo,
+                 env,
+                 meta_batch_size,
```
ryanjulian: why can't i decide this when i call `obtain_samples`?

ahtsan: It's required to indicate how many parallel environments should be created

src/garage/sampler/rl2_sampler.py
ahtsan: Are we only using one environment and set it with different tasks, or different environments?

src/garage/sampler/rl2_sampler.py
ahtsan: How should we handle the allocation of environments over tasks If we don't enforce it with some constraints? e.g. meta_batch_size = 5 but there are 3 parallel envs.

src/garage/sampler/rl2_sampler.py
ryanjulian: different environments. see metaworld ML10 and ML45.

src/garage/sampler/rl2_sampler.py
ryanjulian: i agree this is challenging within this implementation, but does it have to be this limiting?

what would you do if meta_batch_size was 100 but you can only have 10 CPUs for workers?

ahtsan: 10 envs at a time, for 10 times. How about limiting `n_workers % meta_batch_size == 0` if `n_workers < meta_batch_size`, and `meta_batch_size % n_workers == 0`  if `meta_batch_size > n_workers`?

ahtsan:
@ryanjulian @naeioi Please take another took

src/garage/sampler/rl2_sampler.py
naeioi: What about just `assert` it

src/garage/sampler/rl2_sampler.py
naeioi: Ah. Smart solution.

src/garage/sampler/rl2_sampler.py
naeioi: Then I think it should be `Size of meta batch`.

src/garage/sampler/rl2_sampler.py
naeioi: I see that both ProMP and original MAML implementation has a single-task interface. Can this functionality - train on different environments - supported through an env wrapper?

src/garage/sampler/rl2_sampler.py
ahtsan: Fixed

src/garage/sampler/rl2_sampler.py
```diff
@@ -0,0 +1,224 @@
...
+        """Setup workers.
+
+        Args:
+            env_indices (List[Int]): Indices of environments to be assigned
+                to workers for sampling.
+
+        """
+        vec_envs = []
+        for env_ind in env_indices:
+            vec_env = pickle.loads(pickle.dumps(self.envs[env_ind]))
```
krzentner: It would probably be cleaner to use `copy.deepcopy`.

src/garage/sampler/rl2_sampler.py
```diff
@@ -0,0 +1,224 @@
...
+            vec_envs.extend([vec_env for env in range(self._envs_per_worker)])
```
krzentner: Shouldn't you copy here, not above?

src/garage/sampler/rl2_sampler.py
```diff
@@ -0,0 +1,224 @@
...
+        seed0 = deterministic.get_seed()
+        if seed0 is not None:
+            for (i, e) in enumerate(vec_envs):
+                e.seed(seed0 + i)
+
+        self._vec_env = VecEnvExecutor(
+            envs=vec_envs, max_path_length=self.algo.max_path_length)
+
+    # pylint: disable=too-many-statements
+    def obtain_samples(self, itr, batch_size=None, whole_paths=True):
```
krzentner: I've read through this function three times, and I still don't understand why it's different from the other samplers.

Please write much more documentation about what's going on or split this function up.

src/garage/sampler/rl2_sampler.py
```diff
@@ -0,0 +1,224 @@
...
+        vec_envs = []
+        for env_ind in env_indices:
+            vec_env = pickle.loads(pickle.dumps(self.envs[env_ind]))
+            vec_envs.extend([vec_env for env in range(self._envs_per_worker)])
+        seed0 = deterministic.get_seed()
+        if seed0 is not None:
+            for (i, e) in enumerate(vec_envs):
+                e.seed(seed0 + i)
+
+        self._vec_env = VecEnvExecutor(
```
krzentner: Wait, so do you create new environments every worker every time `obtain_samples` is called without ever closing any of them? This seems very likely to lead to significant memory / resource leaks.

src/garage/sampler/rl2_sampler.py
ahtsan: Good point. I will add more docstring. In summary, RL^2 needs to concatenate all paths for each environment/task, so we need to keep track of the source of each pat. Here we will have `paths[idx]` containing all paths for the `idx`-th environment/task.

src/garage/sampler/rl2_sampler.py
ahtsan: Good catch

src/garage/sampler/rl2_sampler.py
ahtsan: I think so, will fix

src/garage/sampler/rl2_sampler.py
```diff
@@ -0,0 +1,224 @@
...
+        self.algo = algo
+        self.envs = env
+
+        self._n_envs = n_envs
+        self._meta_batch_size = meta_batch_size
+        self._vec_env = None
+        self._envs_per_worker = None
+        self._vec_envs_indices = None
+
+    def start_worker(self):
```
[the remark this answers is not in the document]
ahtsan: This is somehow complying to the runner API as `start_worker()` seems like a method dedicated for setting the vec_env. For now, should we move the content to `__init__` and just pass in `start_worker()`?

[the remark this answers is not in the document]
krzentner: Yes. `start_worker()` is deprecated.

src/garage/sampler/rl2_sampler.py
ryanjulian: Multiple environments is the normal case, single-task environments is the special case. You can use the wrapper for now to get your code merged, but @krzentner designed the sampler to specifically support this.

PR #1128, opened 2020-01-11
Prepare VPG and others for MAML

naeioi:
This PR is part of the breakdown of the original MAML PR #1101 . It prepares existing codebase for MAML.

### DIFF
--- a/src/garage/torch/algos/vpg.py
+++ b/src/garage/torch/algos/vpg.py
@@ -131,54 +131,56 @@ class VPG(BatchPolopt):
                 * average_return: (float)
 
         """
-        valids, obs, actions, rewards = self.process_samples(itr, paths)
+        obs, actions, rewards, valids, baselines = self.process_samples(
+            itr, paths)
 
-        loss = self._compute_loss(itr, paths, valids, obs, actions, rewards)
+        loss = self._compute_loss(itr, obs, actions, rewards, valids,
+                                  baselines)
 
-        # Memorize the policy state_dict
         self._old_policy.load_state_dict(self.policy.state_dict())
 
         self._optimizer.zero_grad()
         loss.backward()
 
         kl_before = self._compute_kl_constraint(obs).detach()
-        self._optimize(itr, paths, valids, obs, actions, rewards)
+        self._optimize(itr, obs, actions, rewards, valids, baselines)
 
         with torch.no_grad():
-            loss_after = self._compute_loss(itr, paths, valids, obs, actions,
-                                            rewards)
+            loss_after = self._compute_loss(itr, obs, actions, rewards, valids,
+                                            baselines)
             kl = self._compute_kl_constraint(obs)
             policy_entropy = self._compute_policy_entropy(obs)
-            average_return = self._log(itr, paths, loss.item(),
-                                       loss_after.item(), kl_before.item(),
-                                       kl.item(),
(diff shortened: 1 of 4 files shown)


src/garage/torch/algos/vpg.py
```diff
@@ -323,26 +332,37 @@ def process_samples(self, itr, paths):
...
+            pad_to_last(self._get_baselines(path),
+                        total_length=self.max_path_length) for path in paths
+        ])
 
-        return valids, obs, actions, rewards
+        return obs, actions, rewards, valids, baselines
 
-    def _log(self, itr, paths, loss_before, loss_after, kl_before, kl,
-             policy_entropy):
+    def evaluate_performance(self, itr, batch):
```
ryanjulian: please call `RLAlgorithm.evaluate_performance` instead.

src/garage/torch/algos/vpg.py
```diff
@@ -131,54 +131,56 @@ def train_once(self, itr, paths):
...
+                itr, {
+                    'paths': paths,
+                    'loss_before': loss.item(),
+                    'loss_after': loss_after.item(),
+                    'kl_before': kl_before.item(),
+                    'kl': kl.item(),
+                    'policy_entropy': policy_entropy.mean().item()
+                })
 
         self.baseline.fit(paths)
```
ahtsan: Shall we fit baseline with the augmented reward? Are you using the original reward or augmented reward here?

ryanjulian: I recommend you fit with augmented reward -- you want the baseline to be as accurate as possible. This only works if the baseline input has enough information to predict the augmented reward, though.

src/garage/torch/algos/vpg.py
zequnyu: Our intention for `evaluate_performance()` is not to override the function, but to call it with the batch.

naeioi: How should I incorporate additional metrics, then? Metrics like `LossBefore, KL, Entropy` are not included in `np/algos/base.py: evaluate_performance()`.

src/garage/torch/algos/vpg.py
naeioi: Here, the original reward is fed to the baseline. This is inherited from the old implementation of torch/VPG. If we were to fix this I think we better make it in a separate PR.

src/garage/torch/algos/vpg.py
ryanjulian: those can be logged within the algorithm, but the standardized evaluation metrics should always be logged via `RLAlgorithm.evaluate_performance`. This makes sure they are all calculated the same across algorithms.

naeioi: Are you suggesting that I override `evaluate_performance()` and call `RLAlgorithm.evaluate_performance()` in it, plus additional logs.

ahtsan: maybe like calling `evaluate_performance()` like
https://github.com/rlworkgroup/garage/blob/master/src/garage/tf/algos/batch_polopt.py#L244
and log your own metrics like
https://github.com/rlworkgroup/garage/blob/master/src/garage/tf/algos/batch_polopt.py#L259

naeioi: Thanks for point this out.

PR #1133, opened 2020-01-14
Local sampler

krzentner:
The change introduces a simpler sampler using the new interface.

Note that this PR is stacked on the required changes, so it looks a fair bit more complicated.

### DIFF
--- a/src/garage/sampler/local_sampler.py
+++ b/src/garage/sampler/local_sampler.py
@@ -1,0 +1,106 @@
+"""Sampler that runs workers in the main process."""
+
+from garage import TrajectoryBatch
+from garage.sampler.sampler import Sampler
+
+
+class LocalSampler(Sampler):
+    """Sampler that runs workers in the main process.
+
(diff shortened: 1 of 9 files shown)


src/garage/sampler/local_sampler.py
```diff
@@ -0,0 +1,104 @@
+"""Sampler that runs workers in the main process."""
+
+from garage import TrajectoryBatch
+from garage.sampler.sampler import Sampler
+
+
+class LocalSampler(Sampler):
+    """Sampler that runs workers in the main process.
```
avnishn: Is the purpose of saying Local sampler to distinguish from a sampler that uses some parallelization strategy? If so, maybe we can include a little blurb about that?

krzentner:
> Do you (@krzentner) think the default worker should be returning trajectories? What about batches of `TimeSteps`?
I definitely think that `DefaultWorker` should return `TrajectoryBatch`s from `collect_rollout()`.

PR #1134, opened 2020-01-14
Make Ray TF Actually Work Again

krzentner:
This PR contains all of the changes to actually run the ray tensorflow example again.

Of note:
 - Replaces the Ray TF Sampler with LocalTFRunner injecting a TFWorkerWrapper into the sampler.
 - Handles the sampler returning TrajectoryBatch.
 - Fixes shutdown of workers.


ryanjulian:
Looks like this is held up mainly on tests for `src/garage/tf/worker.py`. 

Feel free to force push this if a near-future change will have tests which use it anyway.

src/garage/_dtypes.py
```diff
@@ -124,7 +125,14 @@ def __new__(cls, env_spec, observations, last_observations, actions,
                     len(lengths), last_observations.shape[0]))
 
         # actions
-        if not env_spec.action_space.contains(first_action):
```
avnishn: why do you want to remove this check?

PR #1136, opened 2020-01-16
Add MAML

naeioi:
This the second part of the breakdown of the original MAML PR #1101 . It adds MAML to garage. Tests and benchmarks are not included.


src/garage/torch/algos/maml.py
```diff
@@ -0,0 +1,475 @@
...
+
+        Returns:
+            tuple: Tuple of (all_samples, all_params).
+                all_samples (list[MAMLTrajectoryBatch]): A list of size
+                    [meta_batch_size * (num_grad_updates + 1)]
+                all_params (list[dict]): A list of named parameter
+                    dictionaries.
+
+        """
+        tasks = self._env.sample_tasks(self._meta_batch_size)
```
ahtsan: so here you pass in a single environment and set task when needed, while another way is to pass multiple environments at the first place. Does that matter? I'm now doing the second approach for RL^2 -- should we keep it consistent and if so, which one is better? @ryanjulian @krzentner

src/garage/torch/algos/maml.py
```diff
@@ -0,0 +1,475 @@
...
+                 inner_algo=None):
+        if inner_algo is None:
+            inner_algo = VPG(env.spec, policy, baseline)
+
+        if policy.vectorized:
+            self.sampler_cls = OnPolicyVectorizedSampler
+        else:
+            self.sampler_cls = BatchSampler
+
+        self.policy = policy
```
ryanjulian: please use `_private` for class members.

src/garage/torch/algos/maml.py
ryanjulian: please use an API which assumes each task is an independent `gym.Env` object.

src/garage/torch/algos/maml.py
```diff
@@ -0,0 +1,475 @@
...
+        baseline (garage.np.baselines.Baseline): The baseline.
+        meta_optimizer (Union[torch.optim.Optimizer, tuple]):
+            Type of optimizer.
+            This can be an optimizer type such as `torch.optim.Adam` or a tuple
+            of type and dictionary, where dictionary contains arguments to
+            initialize the optimizer e.g. `(torch.optim.Adam, {'lr' = 1e-3})`.
+        meta_batch_size (int): Number of tasks sampled per batch.
+        inner_lr (double): Adaptation learning rate.
+        num_grad_updates (int): Number of adaptation gradient steps.
+        inner_algo (garage.torch.algos.VPG): The inner algorithm used for
```
ryanjulian: this is really an inner loss function right? rather than reaching into an inner algorithm's private fields, can you make a module for the loss functions instead (and share it between algorithms)?

src/garage/torch/algos/maml.py
```diff
@@ -0,0 +1,475 @@
...
+        if policy.vectorized:
+            self.sampler_cls = OnPolicyVectorizedSampler
+        else:
+            self.sampler_cls = BatchSampler
+
+        self.policy = policy
+        self.max_path_length = inner_algo.max_path_length
+        self._env = env
+        self._baselines = [
+            copy.deepcopy(baseline) for _ in range(meta_batch_size)
```
ryanjulian: what if `meta_batch_size` != `number of tasks`? does this require that the index of each task in the meta-batch be consistent?

src/garage/torch/algos/maml.py
```diff
@@ -0,0 +1,475 @@
...
+        inner_algo (garage.torch.algos.VPG): The inner algorithm used for
+            computing loss.
+
+    """
+
+    def __init__(self,
+                 env,
+                 policy,
+                 baseline,
+                 meta_optimizer=torch.optim.Adam,
```
ryanjulian: what if i'm doing MAML-TRPO?

src/garage/torch/algos/maml.py
naeioi: Because runner has a reference to `algo.policy`, I will need to fix all algorithms. The same goes to `sampler_cls`.

ahtsan: why not add a property?
```
@property
def policy(self):
    return self._policy
```

naeioi: I can do this if this is acceptable.

src/garage/torch/algos/maml.py
naeioi: _re:_ inner_algo is only used for its loss function.
Yes and no. The loss function of inner_algo decides how a policy is adapted to a specific task, so yes. But MAML also reuses parameters and helper functions from inner_algo, e.g. `max_path_length` and `compute_kl_constrait()`, so no. Note that, these params and helpers are not necessarily bound to inner_algo - We can also move them to MAML, although doing so will make some duplicate code.

_re:_ Modularize loss function.
This is an interesting idea. The only two available options for on-policy algorithms in garage should be the loss function of VPG and PPO, right?

src/garage/torch/algos/maml.py
naeioi: I assume that `meta_batch_size` is defined as the number of tasks. Can you explain what is the difference between the two?

src/garage/torch/algos/maml.py
naeioi: Use `ConjugateGradientOptimizer` instead.

naeioi: I just notice that `torch.optim.Adam.step()` accepts only one callable to compute the loss, while in `_meta_optimize()` two callable are passed to `optimizer.step()` - This to comply to  `ConjugateGradientOptimizer` that needs an additional callable to compute KL. I will fix this.

src/garage/torch/algos/maml.py
naeioi: I am still not sure where should I get tasks. Should I expect sample_tasks() to be called outside of MAML? Or just leave the current implementation here and wait for KR's meta-task sampler.

src/garage/torch/algos/maml.py
naeioi: Thanks for the detailed explanation. I plan to implement MAML-TRPO, MAML-PPO and MAML-VPG by subclassing base MAML. It will be similar to how we implement tf/PPO on tf/NPO, by specifying `inner_algo` and `optimizer` in the subclass.

src/garage/torch/algos/maml.py
naeioi: Thanks for the explanation. 

Prior arts uses one baseline for all `tasks[i][j]`. This is OK because the baseline is not recurrent and fit-before-predict. From previous experiments, I see no performance impact if I apply this to garage. 

Use `meta_batch_size` baselines is to support fit-after-predict. In such a case, using only one baseline implies that we fit on `tasks[i][j]` and predict on `tasks[i][j+1]`, which is wrong.

ryanjulian: I see. Did you decide to use fit-after-predict or fit-before-predict by default?

ryanjulian: I wasn't aware supporting both would require such a big architectural change.

If all of the prior art uses fit-before-predict, then I suggest you do the same and eliminate some complexity.

I think the understand the nature of the problem better now, and fit-before-predict is a little bit of a hack to make on-policy meta-RL work. I won't go into details here but feel free to ask me in person.

naeioi: OK. I will use fit-before-predict because its give SOTA performance.

ryanjulian: LGTM

ahtsan: In the case of tf branch, we need n baselines since the sample paths are processed in batch, so we can use the i-th baseline for the i-th task, Unlike ProMP where it loops through the paths and fit/predict one by one.

ryanjulian: @ahtsan I'm not sure I follow, since the `ith` task isn't necessarily the same task between batches. This would reveal the task ID.

ahtsan: I meant the `i-th` environment, i.e. the `i-th` element of `self._vec_env.envs` in the sampler.

src/garage/torch/algos/maml.py
```diff
@@ -0,0 +1,479 @@
...
+            tabular.record('LossAfter', loss_after)
+            tabular.record('dLoss', loss_before - loss_after)
+            tabular.record('KLBefore', kl_before)
+            tabular.record('KLAfter', kl)
+            tabular.record('Entropy', policy_entropy)
+
+        return average_return
+
+
+class MAMLTrajectoryBatch(
```
ryanjulian: it's okay to define and use this internally, but please make sure it's not exported to the API. 

Is it really worth it to define an entire new ADT just for a single function call interface? Why not just pass `TrajectoryBatch` and `baseline` as separate arguments?

Why not add a `TorchTrajectoryBatch` to `garage.torch` which has a `from_numpy` constructor (which takes `TrajectoryBatch`)?

src/garage/torch/algos/maml.py
ryanjulian: Do you assume the the `i`-th element of `self._vec_env.envs` corresponds to the same task between batches?

i.e. imagine there are only 5 tasks, with IDs 0-4. Do you assume that that if task 0 appears at `self._vec_env.envs[0]` on iteration 1 that it will also appear at `self._vec_env.envs[0]` on iteration 2?

ahtsan: Yes if we only set `self._vec_env.envs[0]` to task 0 at the first place and we did not set it to something else after, since we did not update ``self._vec_env.envs[0]`.
No if we set it to something else for each iteration, and `self._vec_env.envs[0]` will be what it is set to be.

Basically what I did is to align `baselines[i]` with `self._vec_env.envs[i]`, regardless of what the `i-th` task is, so that I can fit and predict for `self._vec_env.envs[i]` using `baselines[i]` in any iteration, without knowing what the `i-th` task is.

ryanjulian: this sounds fine to me

a slight optimization for RL2 might be to fit a single baseline `global_baseline` with all data at the end of each iteration after policy optimization (`global_baseline.fit(all_paths)`). then, at the beginning of the next iteration, `[baselines[i].parameters = global_baseline.parameters for i in range(meta_batch_size)]`. this allows all of your baselines to share parameters between updates. this might allow your baseline to learn a little faster, especially at the beginning.

src/garage/torch/algos/maml.py
naeioi: Added MAML-VPG, MAML-TRPO and MAML-PPO.

src/garage/torch/algos/maml/maml_base.py
```diff
@@ -0,0 +1,487 @@
...
+
+from garage.misc import tensor_utils
+from garage.sampler import OnPolicyVectorizedSampler
+from garage.tf.samplers import BatchSampler
+from garage.torch.algos import _Default, make_optimizer
+from garage.torch.optimizers import ConjugateGradientOptimizer, DiffSGD
+from garage.torch.utils import update_module_params
+
+
+class MAMLBase:
```
ryanjulian: hmm probably just call this maml.py / MAML

src/garage/torch/algos/maml/maml_base.py
```diff
@@ -0,0 +1,487 @@
...
+            kl (float): KL divergence after optimization step.
+            policy_entropy (float): Policy entropy.
+
+        Returns:
+            float: The average return in last epoch cycle.
+
+        """
+        tabular.record('Iteration', itr)
+
+        for i in range(self._num_grad_updates + 1):
```
ryanjulian: Is there a way to just use `RLAlgorithm.evaluate_performance`?

src/garage/torch/algos/maml/maml_base.py
```diff
@@ -0,0 +1,487 @@
...
+            runner.step_itr += 1
+
+        return last_return
+
+    def train_once(self, itr, all_samples, all_params):
+        """Train the algorithm once.
+
+        Args:
+            itr (int): Iteration number.
+            all_samples (list[list[MAMLTrajectoryBatch]]): A two
```
ryanjulian: this could use just `TrajectoryBatch` if you passed baselines separately.

src/garage/torch/algos/maml.py
naeioi: I tried and found it difficult to add a `TorchTrajectoryBatch`. For example, I was not sure how to handle `last_observations`.  I think it will be much easier after we use samplers (e.g. `LocalSampler`) that return `TrajectoryBatch` in the inner algorithm (i.e. VPG).

src/garage/torch/algos/maml/maml_base.py
```diff
@@ -0,0 +1,487 @@
...
+
+        """
+        tasks = self._env.sample_tasks(self._meta_batch_size)
+        all_samples = [[] for _ in range(len(tasks))]
+        all_params = []
+        theta = dict(self._policy.named_parameters())
+
+        for i, task in enumerate(tasks):
+            self._set_task(runner, task)
+            self._set_baseline(i)
```
ryanjulian: i thought you said that all tasks share a baseline? why do you need multiple baselines?

src/garage/torch/algos/maml/maml_base.py
naeioi: I don't think so because it's the post-adaptation performance that matters, and the performance number is averaged over tasks.

src/garage/torch/algos/maml/maml_base.py
```diff
@@ -0,0 +1,487 @@
...
+            discounted_returns = [
+                tensor_utils.discount_cumsum(path_rewards,
+                                             self._inner_algo.discount)[0]
+                for path_rewards in all_rewards
+            ]
+            undiscounted_returns = np.sum(all_rewards, axis=-1)
+            average_return = np.mean(undiscounted_returns)
+            self._episode_reward_mean[i].extend(undiscounted_returns)
+
+            with tabular.prefix('Update_{0}/'.format(i)):
```
ryanjulian: reporting for multiple steps can be considered debugging information, but I think the algorithm should also report the statistics for the final step to generic prefixes (`Evaluation/Train` and `Evaluation/Test`?)

src/garage/torch/algos/maml/maml_base.py
ryanjulian: or `MetaRLAlgorithm.evaluate_train_performance` and `MetaRLAlgorithm.evaluate_test_performance`?

src/garage/torch/algos/maml/maml_base.py
```diff
@@ -0,0 +1,487 @@
...
+            tabular.record('LossAfter', loss_after)
+            tabular.record('dLoss', loss_before - loss_after)
+            tabular.record('KLBefore', kl_before)
+            tabular.record('KLAfter', kl)
+            tabular.record('Entropy', policy_entropy)
+
+        return average_return
+
+
+class MAMLTrajectoryBatch(
```
ryanjulian: see my previous comments on this -- is there a way we can just use `TrajectoryBatch` and just pass baselines separately? ADTs are not very helpful if there is one for each algorithm.

src/garage/torch/algos/maml/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""Collection of MAML algorithms."""
```
ryanjulian: i'm not convinced this package needs to exist. can you keep it flat for now? (i.e. put these in .algos)

src/garage/torch/algos/maml/maml_base.py
```diff
@@ -0,0 +1,487 @@
...
+        self._cur_baseline = self._baselines[0]
+        self._num_grad_updates = num_grad_updates
+        self._meta_batch_size = meta_batch_size
+        self._inner_algo = inner_algo
+        self._inner_optimizer = DiffSGD(self._policy, lr=inner_lr)
+        self._meta_optimizer = make_optimizer(meta_optimizer,
+                                              policy,
+                                              lr=_Default(1e-3),
+                                              eps=_Default(1e-5))
+        self._episode_reward_mean = [
```
ryanjulian: why do we need this?

src/garage/torch/algos/maml/maml_base.py
```diff
@@ -0,0 +1,487 @@
...
+                 num_grad_updates=1):
+        if policy.vectorized:
+            self.sampler_cls = OnPolicyVectorizedSampler
+        else:
+            self.sampler_cls = BatchSampler
+
+        self.max_path_length = inner_algo.max_path_length
+        self._policy = policy
+        self._env = env
+        self._baselines = [
```
ryanjulian: i thought you said you use a single baseline for all tasks?

src/garage/torch/algos/maml/maml_base.py
```diff
@@ -0,0 +1,487 @@
...
+                    dictionaries.
+
+        """
+        tasks = self._env.sample_tasks(self._meta_batch_size)
+        all_samples = [[] for _ in range(len(tasks))]
+        all_params = []
+        theta = dict(self._policy.named_parameters())
+
+        for i, task in enumerate(tasks):
+            self._set_task(runner, task)
```
ryanjulian: this is ugly -- can you do this in the sampler instead?

src/garage/torch/algos/maml/maml_base.py
```diff
@@ -0,0 +1,487 @@
...
+                kl = self._inner_algo._compute_kl_constraint(
+                    task_samples[-1].observations)
+            kls.append(kl)
+
+            update_module_params(self._policy, theta)
+            update_module_params(self._old_policy, old_theta)
+
+        return torch.stack(kls).mean()
+
+    def _compute_loss(self, itr, batch_samples):
```
ryanjulian: seeing that this has a single-line implementation, i think it hides information more than making things clearer.

src/garage/torch/algos/maml/maml_ppo.py
```diff
@@ -0,0 +1,85 @@
...
+
+    """
+
+    def __init__(self,
+                 env,
+                 policy,
+                 baseline,
+                 inner_lr=_Default(3e-4),
+                 lr_clip_range=2e-1,
+                 max_path_length=500,
```
ryanjulian: this is quite long. most MuJoCo environments use something like 100. is this what you found in ProMP?

src/garage/torch/optimizers/__init__.py
```diff
@@ -1,5 +1,6 @@
 """PyTorch optimizers."""
 from garage.torch.optimizers.conjugate_gradient_optimizer import (
     ConjugateGradientOptimizer)
+from garage.torch.optimizers.differentiable_sgd import DiffSGD
```
ryanjulian: please make the inner package name match the class (`DiffSGD` --> `diff_sgd`)

src/garage/torch/algos/maml/maml.py
```diff
@@ -0,0 +1,478 @@
...
+        self._baseline = baseline
+        self._num_grad_updates = num_grad_updates
+        self._meta_batch_size = meta_batch_size
+        self._inner_algo = inner_algo
+        self._inner_optimizer = DiffSGD(self._policy, lr=inner_lr)
+        self._meta_optimizer = make_optimizer(meta_optimizer,
+                                              policy,
+                                              lr=_Default(1e-3),
+                                              eps=_Default(1e-5))
+        self._episode_reward_mean = [
```
ryanjulian: why is  this needed? AFAIK it's only in a couple off-policy algorithms to compare to openai/baselines implementations

src/garage/torch/algos/maml/maml.py
```diff
@@ -0,0 +1,478 @@
...
+            runner.step_itr += 1
+
+        return last_return
+
+    def train_once(self, itr, all_samples, all_params):
+        """Train the algorithm once.
+
+        Args:
+            itr (int): Iteration number.
+            all_samples (list[list[MAMLTrajectoryBatch]]): A two
```
ryanjulian: this could just be TrajectoryBatch if you pass `baselines` separately.

src/garage/torch/algos/maml/maml.py
```diff
@@ -0,0 +1,478 @@
...
+                Each dictionary contains key value pair of names (str) and
+                parameters (torch.Tensor).
+
+        Returns:
+            float: Average return.
+
+        """
+        old_theta = dict(self._policy.named_parameters())
+
+        kl_before = self._compute_kl_constraint(itr,
```
ryanjulian: how expensive is computing these KLs and losses before/after? perhaps we can try to capture them in the optimizer or turn it off if it's expensive?

src/garage/torch/algos/maml/maml.py
```diff
@@ -0,0 +1,478 @@
...
+        for path in paths:
+            path['returns'] = tensor_utils.discount_cumsum(
+                path['rewards'], self._inner_algo.discount)
+
+        self._baseline.fit(paths)
+        obs, actions, rewards, valids, baselines \
+            = self._inner_algo.process_samples(itr, paths)
+        return MAMLTrajectoryBatch(obs, actions, rewards, valids, baselines)
+
+    def evaluate_performance(self, itr, all_samples, loss_before, loss_after,
```
ryanjulian: it is essential that the performance evaluation part of the algorithms is shared. 

i would prefer if this called `RLAlgorithm.evaluate_performance` instead. Alternatively, it may be necessary to introduce `MetaRLAlgorithm` with `evaluate_train_performance` and `evaluate_test_performance`

src/garage/torch/algos/maml/maml_base.py
naeioi: I don't know. This is in `np/batch_polopt` and `tf/batch_polopt`. I can remove this.

src/garage/torch/algos/maml/maml.py
```diff
@@ -0,0 +1,478 @@
...
+            discounted_returns = [
+                tensor_utils.discount_cumsum(path_rewards,
+                                             self._inner_algo.discount)[0]
+                for path_rewards in all_rewards
+            ]
+            undiscounted_returns = np.sum(all_rewards, axis=-1)
+            average_return = np.mean(undiscounted_returns)
+            self._episode_reward_mean[i].extend(undiscounted_returns)
+
+            with tabular.prefix('Update_{0}/'.format(i)):
```
ryanjulian: it is fine to have per-step logging for debugging purposes, but each meta-RL algorithm should report a single authoritative performance number for meta-train and meta-test performance.

src/garage/torch/algos/maml/maml.py
```diff
@@ -0,0 +1,478 @@
...
+
+        """
+        # pylint: disable=protected-access
+        entropies = [
+            self._inner_algo._compute_policy_entropy(samples.observations)
+            for samples in task_samples
+        ]
+        return torch.stack(entropies).mean()
+
+    def _set_task(self, runner, task):
```
ryanjulian: this is really ugly. can you not handle this in the sampler?

@krzentner PTAL

src/garage/torch/algos/maml/maml.py
```diff
@@ -0,0 +1,478 @@
...
+            tabular.record('LossAfter', loss_after)
+            tabular.record('dLoss', loss_before - loss_after)
+            tabular.record('KLBefore', kl_before)
+            tabular.record('KLAfter', kl)
+            tabular.record('Entropy', policy_entropy)
+
+        return average_return
+
+
+class MAMLTrajectoryBatch(
```
ryanjulian: i still don't see the advantage of this dtype over TrajectoryBatch. is there a reason you can't use TrajectoryBatch?

src/garage/torch/algos/maml/maml.py
```diff
@@ -0,0 +1,478 @@
...
+                kl = self._inner_algo._compute_kl_constraint(
+                    task_samples[-1].observations)
+            kls.append(kl)
+
+            update_module_params(self._policy, theta)
+            update_module_params(self._old_policy, old_theta)
+
+        return torch.stack(kls).mean()
+
+    def _compute_loss(self, itr, batch_samples):
```
ryanjulian: IMO this obfuscates more than it makes clear. the implementation is 1 line, and within MAML you would assume that "loss" refers to the MAML loss, i.e. the outer loss. this computers the inner loss. 

i think it's really important that, when i encounter this line, i know you are calling the inner algorithm. you can achieve that my renaming this method to `_compute_inner_algo_loss` or just inlining it.

src/garage/torch/algos/maml/maml.py
```diff
@@ -0,0 +1,478 @@
...
+            task_samples (list[MAMLTrajectoryBatch]): Samples data for
+                one task.
+
+        Returns:
+            torch.Tensor: Computed entropy value.
+
+        """
+        # pylint: disable=protected-access
+        entropies = [
+            self._inner_algo._compute_policy_entropy(samples.observations)
```
ryanjulian: why not stack the samples first, then compute the entropies vectorized? shouldn't that be faster?

src/garage/torch/algos/maml/maml.py
```diff
@@ -0,0 +1,478 @@
...
+                    dictionaries.
+
+        """
+        tasks = self._env.sample_tasks(self._meta_batch_size)
+        all_samples = [[] for _ in range(len(tasks))]
+        all_params = []
+        theta = dict(self._policy.named_parameters())
+
+        for i, task in enumerate(tasks):
+            self._set_task(runner, task)
```
ryanjulian: this is really ugly and relies on some assumptions about the sampler

@krzentner PTAL

ryanjulian: is there a reason you can't do one round of sampling for *all tasks* per inner update (i.e. per `j`)?

another way of asking that is: "is there a way to put the `j` loop on the outside and the `i` loop on the inside here?

src/garage/torch/algos/maml/maml.py
ryanjulian: is this equivalent to

```python
return self._inner_algo._compute_loss(itr, *batch_samples)
```
?

if so, it seems like it would be best to just inline it.

src/garage/torch/algos/maml.py
ryanjulian: perhaps, for now, you can subclass runner so that it calls `sample_task/set_task` for you after every itr.

@krzentner PTAL

src/garage/torch/algos/maml.py
ryanjulian: why is `last_observations` different from any other numpy array?

i think on-policy algorithms generally discard `last_observations` anyway, right?

src/garage/torch/algos/maml/maml_ppo.py
naeioi: No. This is the default number of `torch/PPO.py` in garage.

src/garage/torch/algos/maml/maml.py
naeioi: I removed it.

src/garage/torch/algos/maml/maml.py
naeioi: I can investigate more time on refactoring to use `TrajectoryBatch` in the algorithm. I just feel that since we are not using samplers that return `TrajectoryBatch`, we are not solving the issue from its root cause.

src/garage/torch/algos/maml/maml_ppo.py
ryanjulian: i suggest you set it to match typical values from ProMP

ryanjulian:
if you get the test coverage up to spec, i'm okay with you merging this as-is and coming back to the evaluation/sampling/TrajectoryBatch changes later.

if you could follow-up with a PR which creates a meta-RL global evaluate_performance, i think it would benefit everyone quickly.

in that case, please still address the smaller comments about `_compute_entropy`, `_compute_loss`, the default `max_path_length`.

src/garage/torch/algos/maml.py
ryanjulian: you can merge this, but please leave an issue to unify it with TrajectoryBatch later

naeioi:
@Mergifyio rebase

src/garage/torch/algos/maml_trpo.py
```diff
@@ -0,0 +1,91 @@
...
+                 gae_lambda=1,
+                 center_adv=True,
+                 positive_adv=False,
+                 policy_ent_coeff=0.0,
+                 use_softplus_entropy=False,
+                 stop_entropy_gradient=False,
+                 entropy_method='no_entropy',
+                 meta_batch_size=40,
+                 num_grad_updates=1):
+        inner_algo = VPG(env.spec,
```
krzentner: Shouldn't this be TRPO? Did I miss something?

src/garage/torch/algos/maml.py
```diff
@@ -0,0 +1,456 @@
...
+
+        for _ in runner.step_epochs():
+            all_samples, all_params = self._obtain_samples(runner)
+            tabular.record('TotalEnvSteps', runner.total_env_steps)
+            last_return = self.train_once(runner, all_samples, all_params)
+            runner.step_itr += 1
+
+        return last_return
+
+    def train_once(self, itr, all_samples, all_params):
```
krzentner: You should probably make this method private.

src/garage/torch/algos/maml.py
```diff
@@ -0,0 +1,456 @@
...
+
+        """
+        obs = torch.stack([samples.observations for samples in task_samples])
+        # pylint: disable=protected-access
+        entropies = self._inner_algo._compute_policy_entropy(obs)
+        return entropies.mean()
+
+    def _set_task(self, runner, task):
+        # pylint: disable=protected-access, no-self-use
+        for env in runner._sampler._vec_env.envs:
```
krzentner: Yeah, that looks pretty bad. Can you just use `LocalSampler` and environment updates?

src/garage/torch/algos/maml.py
krzentner: I think in this case it will just be easy to use the new sampler. The performance might be slightly worse using `LocalSampler`, but there should be any code changes required.

src/garage/torch/algos/maml.py
```diff
@@ -0,0 +1,456 @@
...
+
+        losses = []
+        for task_samples, task_params in zip(all_samples, all_params):
+            for i in range(self._num_grad_updates):
+                self._adapt(itr, task_samples[i], set_grad=set_grad)
+
+            update_module_params(self._old_policy, task_params)
+            with torch.set_grad_enabled(set_grad):
+                # pylint: disable=protected-access
+                loss = self._inner_algo._compute_loss(itr, *task_samples[-1])
```
krzentner: It would be nice to change the inner algorithms to make it clear this is part of the public API.

src/garage/torch/algos/maml_trpo.py
naeioi: Because MAML-TRPO uses VPG's loss function and TRPO style meta-policy optimizer. Ryan has a discussion at [here](https://github.com/rlworkgroup/garage/pull/1136#discussion_r367729281)

src/garage/torch/algos/maml.py
naeioi: Yes! I want to explore this solution since `LocalSampler` gives `TrajectoryBatch` back.

src/garage/torch/algos/maml.py
naeioi: @ryanjulian What do you think?

ryanjulian: yes -- we can focus on this later

PR #1137, opened 2020-01-17
RL^2 algorithm

ahtsan:
This PR adds the RL^2 algorithm and some new helper functions.

### DIFF
--- a/src/garage/misc/tensor_utils.py
+++ b/src/garage/misc/tensor_utils.py
@@ -163,6 +163,40 @@ def stack_tensor_dict_list(tensor_dict_list):
     return ret
 
 
+def stack_and_pad_tensor_n(paths, key, max_len):
+    """Stack and pad array of tensors.
+
+    Args:
+        paths (dict[numpy.ndarray]): Tensors to be stacked and padded.
+        key (str): Key of the values in the paths.
+        max_len (int): Maximum length.
+
+    Returns:
+        numpy.ndarray: Stacked and padded tensor.
+
+    """
+    ret = np.asarray([path[key] for path in paths])
+    ret = pad_tensor_n(ret, max_len)
+    return ret
+
+
+def stack_and_pad_tensor_dict(paths, key, max_len):
+    """Stack and pad array of tensors.
+
+    Args:
+        paths (dict[numpy.ndarray]): Tensors to be stacked and padded.
+        key (str): Key of the values in the paths.
+        max_len (int): Maximum length.
+
+    Returns:
+        numpy.ndarray: Stacked and padded tensor.
+
+    """
+    ret = [path[key] for path in paths]
+    ret = stack_tensor_dict_list([pad_tensor_dict(p, max_len) for p in ret])
+    return ret
+
+
 def concat_tensor_dict_list(tensor_dict_list):
     """Concatenate dictionary of list of tensor.
 
--- a/src/garage/misc/tensor_utils.py
+++ b/src/garage/misc/tensor_utils.py
@@ -164,36 +164,29 @@ def stack_tensor_dict_list(tensor_dict_list):
 
 
 def stack_and_pad_tensor_n(paths, key, max_len):
-    """Stack and pad array of tensors.
+    """Stack and pad array of list of tensors.
+
+    Input paths are a list of N dicts, each with values of shape
+    :math:`(D, S^*)`. This function stack and pad the values with the input
+    key with max_len, so output will be shape :math:`(N, D, S^*)`.
 
     Args:
-        paths (dict[numpy.ndarray]): Tensors to be stacked and padded.
-        key (str): Key of the values in the paths.
-        max_len (int): Maximum length.
+        paths (list[dict]): List of dict to be stacked and padded.
+            Value of each dict will be shape of :math:`(D, S^*)`.
+        key (str): Key of the values in the paths to be stacked and padded.
+        max_len (int): Maximum length for padding.
 
     Returns:
-        numpy.ndarray: Stacked and padded tensor.
-
-    """
-    ret = np.asarray([path[key] for path in paths])
-    ret = pad_tensor_n(ret, max_len)
-    return ret
-
-
-def stack_and_pad_tensor_dict(paths, key, max_len):
-    """Stack and pad array of tensors.
-
-    Args:
-        paths (dict[numpy.ndarray]): Tensors to be stacked and padded.
-        key (str): Key of the values in the paths.
-        max_len (int): Maximum length.
-
-    Returns:
-        numpy.ndarray: Stacked and padded tensor.
+        numpy.ndarray: Stacked and padded tensor. Shape: :math:`(N, D, S^*)`
+            where K is the len of input paths.
 
     """
     ret = [path[key] for path in paths]
-    ret = stack_tensor_dict_list([pad_tensor_dict(p, max_len) for p in ret])
+    if isinstance(ret[0], dict):
+        ret = stack_tensor_dict_list(
+            [pad_tensor_dict(p, max_len) for p in ret])
+    else:
+        ret = pad_tensor_n(ret, max_len)
     return ret
 
 
--- a/src/garage/np/algos/base.py
+++ b/src/garage/np/algos/base.py
@@ -71,6 +71,12 @@ class RLAlgorithm(abc.ABC):
 
         undiscounted_returns = [sum(reward) for reward in batch['rewards']]
 
+        if 'success' in batch['env_infos']:
+            success_rate = sum(
+                [path['env_infos']['success'][-1]
+                 for path in batch]) * 1.0 / len(batch)
+            tabular.record('SuccessRate', success_rate)
+
         tabular.record('Iteration', itr)
         tabular.record('Evaluation/NumTrajs', len(returns))
 
--- a/src/garage/tf/algos/rl2.py
+++ b/src/garage/tf/algos/rl2.py
@@ -273,6 +273,7 @@ class RL2(RLAlgorithm):
         dones = np_tensor_utils.stack_and_pad_tensor_n(paths, 'dones', max_len)
         returns = np_tensor_utils.stack_and_pad_tensor_n(
             paths, 'returns', max_len)
+
         agent_infos = np_tensor_utils.stack_and_pad_tensor_n(
             paths, 'agent_infos', max_len)
         env_infos = np_tensor_utils.stack_and_pad_tensor_n(
--- a/src/garage/tf/algos/rl2.py
+++ b/src/garage/tf/algos/rl2.py
@@ -3,81 +3,96 @@
 Reference: https://arxiv.org/pdf/1611.02779.pdf.
 """
 import collections
+import copy
 
-from dowel import logger, tabular
+from dowel import tabular
 import numpy as np
 
-import garage
 from garage.misc import tensor_utils as np_tensor_utils
-from garage.np.algos import RLAlgorithm
+from garage.tf.algos import BatchPolopt
 
 
-class RL2(RLAlgorithm):
+class RL2(BatchPolopt):
+    # pylint: disable=abstract-method
     """RL^2 .
 
     Args:
-        inner_algo (garage.tf.algos.BatchPolopt): Inner algorithm. RL^2 only
-            works with garage.tf.algos.BatchPolopt.
-        max_path_length (int): Maximum length for trajectories with respect
-            to RL^2. Notice that it is differen from the maximum path length
-            for the inner algorithm.
-
-    Raises:
-        ValueError: If the inner algorithm is not of type
-            garage.tf.algos.BatchPolopt.
+        env_spec (garage.envs.EnvSpec): Environment specification.
+        policy (garage.tf.policies.base.Policy): Policy.
+        baseline (garage.tf.baselines.Baseline): The baseline.
+        scope (str): Scope for identifying the algorithm.
+            Must be specified if running multiple algorithms
+            simultaneously, each using different environments
+            and policies.
+        max_path_length (int): Maximum length of a single rollout.
+        episode_per_task (int): Number of episode to be sampled per task.
+        discount (float): Discount.
+        gae_lambda (float): Lambda used for generalized advantage
+            estimation.
+        center_adv (bool): Whether to rescale the advantages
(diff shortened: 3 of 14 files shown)


src/garage/misc/tensor_utils.py
```diff
@@ -163,6 +163,40 @@ def stack_tensor_dict_list(tensor_dict_list):
...
+def stack_and_pad_tensor_n(paths, key, max_len):
+    """Stack and pad array of tensors.
+
+    Args:
+        paths (dict[numpy.ndarray]): Tensors to be stacked and padded.
+        key (str): Key of the values in the paths.
+        max_len (int): Maximum length.
+
+    Returns:
+        numpy.ndarray: Stacked and padded tensor.
```
ryanjulian: Can you tell us something about the input/output shapes?

src/garage/misc/tensor_utils.py
```diff
@@ -163,6 +163,40 @@ def stack_tensor_dict_list(tensor_dict_list):
...
+    ret = np.asarray([path[key] for path in paths])
+    ret = pad_tensor_n(ret, max_len)
+    return ret
+
+
+def stack_and_pad_tensor_dict(paths, key, max_len):
+    """Stack and pad array of tensors.
+
+    Args:
+        paths (dict[numpy.ndarray]): Tensors to be stacked and padded.
```
ryanjulian: Can you tell us something about the input/output shapes?

src/garage/tf/algos/rl2.py
```diff
@@ -0,0 +1,316 @@
...
+    def train(self, runner):
+        """Obtain samplers and start actual training for each epoch.
+
+        Args:
+            runner (LocalRunner): LocalRunner is passed to give algorithm
+                the access to runner.step_epochs(), which provides services
+                such as snapshotting and sampler control.
+
+        Returns:
+            float: The average return in last epoch cycle.
```
ryanjulian: AFAICT we don't use "epoch cycles" anymore

src/garage/tf/algos/rl2.py
```diff
@@ -0,0 +1,316 @@
...
+        Args:
+            itr (int): Iteration number.
+            paths (list[dict]): A list of collected paths.
+
+        Returns:
+            numpy.float64: Average return.
+
+        """
+        paths = self.process_samples(itr, paths)
+        self._inner_algo.log_diagnostics(paths)
```
ryanjulian: is this guaranteed to exist?

src/garage/tf/algos/rl2.py
```diff
@@ -0,0 +1,316 @@
...
+        logger.log('Optimizing policy...')
+        self._inner_algo.optimize_policy(itr, paths)
+        return paths['average_return']
+
+    def process_samples(self, itr, paths):
```
ryanjulian: can this be private?

src/garage/tf/algos/rl2.py
```diff
@@ -0,0 +1,316 @@
...
+                 returns=_returns,
+                 env_infos=_env_infos,
+                 agent_infos=_agent_infos,
+                 lengths=_lengths,
+                 discount=self._discount,
+                 episode_reward_mean=self._inner_algo.episode_reward_mean))
+
+        self._inner_algo.episode_reward_mean.extend(undiscounted_returns)
+
+        # Log success rate for MetaWorld
```
ryanjulian: can you make an issue on metaworld to utilize termination conditions for this instead?

src/garage/tf/algos/rl2.py
```diff
@@ -0,0 +1,316 @@
...
+                 rewards=_rewards,
+                 terminals=_terminals,
+                 returns=_returns,
+                 env_infos=_env_infos,
+                 agent_infos=_agent_infos,
+                 lengths=_lengths,
+                 discount=self._discount,
+                 episode_reward_mean=self._inner_algo.episode_reward_mean))
+
+        self._inner_algo.episode_reward_mean.extend(undiscounted_returns)
```
ryanjulian: is `episode_reward_mean` guaranteed to exist?

src/garage/tf/algos/rl2.py
```diff
@@ -0,0 +1,316 @@
...
+
+        # Log success rate for MetaWorld
+        if 'success' in all_paths[0]['env_infos']:
+            success_rate = sum(
```
ryanjulian: i think it makes more sense to log `SuccessRate` as `0-1`

src/garage/tf/algos/rl2.py
```diff
@@ -0,0 +1,316 @@
...
+                [path['env_infos']['success'][-1]
+                 for path in all_paths]) * 100.0 / len(all_paths)
+            tabular.record('SuccessRate', success_rate)
+
+        tabular.record('Entropy', ent)
+        tabular.record('Perplexity', np.exp(ent))
+        tabular.record('Extras/EpisodeRewardMean',
+                       np.mean(self._inner_algo.episode_reward_mean))
```
ryanjulian: is this guaranteed to exist?

src/garage/tf/algos/rl2.py
```diff
@@ -0,0 +1,316 @@
...
+    Args:
+        inner_algo (garage.np.algos.RLAlgorithm): Inner algorithm.
+        max_path_length (int): Maximum length for trajectories with respect
+            to RL^2. Notice that it is different from the maximum path length
+            for the inner algorithm.
+
+    """
+
+    def __init__(self, inner_algo, max_path_length):
+        self._inner_algo = inner_algo
```
ryanjulian: what are valid values for `inner_algo`? What is its interface? If we don't have a proper interface for this, you can at least add some checks to make sure the input is sane.

src/garage/tf/algos/rl2.py
```diff
@@ -0,0 +1,316 @@
...
+            paths (list[dict]): A list of collected paths.
+
+        Returns:
+            numpy.float64: Average return.
+
+        """
+        paths = self.process_samples(itr, paths)
+        self._inner_algo.log_diagnostics(paths)
+        logger.log('Optimizing policy...')
+        self._inner_algo.optimize_policy(itr, paths)
```
ryanjulian: is this guaranteed to exist?

src/garage/tf/algos/rl2.py
```diff
@@ -0,0 +1,316 @@
...
+            _lengths, _env_infos, _agent_infos) = \
+            self._stack_paths(
+                max_len=self._max_path_length,
+                paths=all_paths
+            )
+
+        ent = np.sum(self._policy.distribution.entropy(agent_infos) *
+                     valids) / np.sum(valids)
+
+        # performance is evaluated per individual paths
```
ryanjulian: do you mean across all paths?

src/garage/tf/algos/rl2.py
ryanjulian: also, can you add this to `RLAlgorithm.evaluate_performance` instead?

src/garage/tf/algos/rl2.py
ahtsan: Looks like `garage.tf.algos.BatchPolopt` has the proper interface for it. Does it make sense to enforce `inner_algo` to be this type? or better to define a `TfRLAlgorithm`, have `garage.tf.algos.BatchPolopt` inherited from it and enforce `inner_algo` to be `TfRLAlgorithm`?
Basically we want the algo to have `optimize_policy` (After a second check, `log_diagnostics` seems to be deprecated and unnecessary).

ryanjulian: We don't have a true interface for "works with RL2," and defining one is difficult, so I suggest 2 solutions to this problem (they are not mutually exclusive):

1) `inner_algo` is a string argument with possible values `vpg`, `trpo`, and `ppo` (did I miss any?). Your constructor configures RL2 to use the private interfaces accordingly.
2) `RL2` is not a public class users instantiate -- instead they usually instantiate classes `RL2VPG`, `RL2PPO`, and `RL2TRPO` which implement the respective variants.

ahtsan:
> Please include at least 1 example script

It would require the rl2 sampler and the new `NPO` to come in place, since without them rl2 won't run but at the same time we are trying to split them into separate PRs.

src/garage/tf/algos/rl2.py
ahtsan: Yes will fix the wording

src/garage/tf/algos/rl2.py
ahtsan: In TF branch, all `vpg`, `trpo` and `ppo` inherit from `BatchPolopt` which has the only required method `optimize_policy` as its abstract method. In this case, I can enforce `inner_algo` to be a `garage.tf.algos.BatchPolopt`?

src/garage/tf/algos/rl2.py
ahtsan: What exactly does "utilize termination conditions for this" do?

src/garage/tf/algos/rl2.py
ryanjulian: this is fine with me. i'll refrain from making more comments until i see the example.

src/garage/tf/algos/rl2.py
ryanjulian: it means that `env_info['success']` and `done` should be the same thing.

ryanjulian:
Can you add the example now that the RL2 sampler is in place? I'd rather get this merged and iterate then have it sit in PR hell for longer.

src/garage/tf/algos/rl2.py
```diff
@@ -0,0 +1,316 @@
...
+
+    def __init__(self, inner_algo, max_path_length):
+        self._inner_algo = inner_algo
+        self._max_path_length = max_path_length
+        self._env_spec = inner_algo.env_spec
+        self._flatten_input = inner_algo.flatten_input
+        self._policy = inner_algo.policy
+        self._discount = inner_algo.discount
+
+    def train(self, runner):
```
krzentner: AFAICT, this method is identical to `garage.tf.BatchPolopt.train`. I guess it would be weird to inherit from that though, and it would also be weird to call just one method from it.

src/garage/tf/algos/rl2.py
krzentner: I think changing metaworld is probably the wrong approach given our timeline. At least, that's what I did for `log_performance` (which should obviate the need for this code).

ahtsan:
@ryanjulian @krzentner I changed my code -- now RL2 inherits from BatchPolopt and we will have RL2NPO and RL2PPO, which both share functions in `RL2`.

src/garage/np/algos/base.py
```diff
@@ -71,6 +71,12 @@ def evaluate_performance(self, itr, batch):
 
         undiscounted_returns = [sum(reward) for reward in batch['rewards']]
 
+        if 'success' in batch['env_infos']:
```
krzentner: I've already done this in my PR, which should be merged soon.

src/garage/tf/algos/rl2_npo.py
```diff
@@ -0,0 +1,733 @@
...
+            valid_var=valid_var,
+            length_var=length_var,
+            policy_state_info_vars_list=policy_state_info_vars_list,
+            policy_old_dist_info_vars_list=policy_old_dist_info_vars_list,
+        )
+
+        return policy_loss_inputs, policy_opt_inputs
+
+    # pylint: disable=too-many-branches, too-many-statements
+    def _build_policy_loss(self, i):
```
krzentner: This is way too much copied code. There are definitely better ways to implement RL2 without copying all of this.

src/garage/tf/misc/tensor_utils.py
```diff
@@ -272,12 +278,132 @@ def positive_advs(advs, eps, name=None):
...
+                                     axis=2)
+        returns = tf.nn.conv1d(rewards_pad,
+                               return_filter,
+                               stride=1,
+                               padding='VALID')
+    return returns
+
+
+def discounted_returns_meta_learn(discount, max_len, episode_per_task,
+                                  rewards):
```
krzentner: I really think it shouldn't be necessary to implement this function. Isn't it basically the same as the function above?

src/garage/tf/misc/tensor_utils.py
```diff
@@ -272,12 +278,132 @@ def positive_advs(advs, eps, name=None):
...
+    with tf.name_scope(name, 'center_adv_local', [advs, axes, eps, max_len]):
+        mean, variance = tf.nn.moments(advs, axes=axes)
+        mean = tf.tile(tf.expand_dims(mean, axis=axes), [1, max_len])
+        variance = tf.tile(tf.expand_dims(variance, axis=axes), [1, max_len])
+        advs = (advs - mean) / (tf.sqrt(variance) + eps
+                                )  # epsilon to avoid dividing by zero
+    return advs, mean
+
+
+def compute_advantages_meta_learn(discount,
```
krzentner: Is this function also a copy? I doesn't look new.

src/garage/tf/algos/rl2_npo.py
krzentner: After discussing this and reviewing the code carefully, I think Ryan and I understand a lot better what's going on here.
We also re-read part of the original RL^2 paper. We discovered a few things:
 - Originally, RL^2 was implemented with all tasks using the same RNN / GRU baseline. ProMP doesn't do this.
 - According to the paper, advantages are computed across the "trial" a.k.a. all paths in the rollout. This means that you shouldn't need `compute_advantages_meta_learn` or `discounted_returns_meta_learn`, since you can just call the original methods on the concatenated paths.

As for the implementation, we think you can avoid copying by doing the following:
 - Modify the NPO implementation by refactoring out all of the parts you need to change. For example, you'll probably need to create an empty `fit_baseline_early()` method in NPO. Then, create a subclass of NPO (maybe called `_NPO_RL2` which is used as an `_inner_algo` in your `RL2` class. This subclass can over-ride all of the unique behavior (such as implementing the `fit_baseline_early()` method), without needing to copy any code.

Sorry if my prior advice on inheriting from `BatchPolopt` lead you to doing it this way. Hopefully this plan makes sense?
Let me know if you have any questions, or find yourself copying code. It should be possible to avoid. Thanks.

ahtsan:
I will submit another PR so it can be merged easier.

PR #1139, opened 2020-01-19
Rename self.evaluate for off-policy algos

zequnyu:
- also removes duplicate performance recording after `evaluate_performance()` being introduced.

### DIFF
--- a/src/garage/np/algos/off_policy_rl_algorithm.py
+++ b/src/garage/np/algos/off_policy_rl_algorithm.py
@@ -66,7 +66,6 @@ class OffPolicyRLAlgorithm(RLAlgorithm):
         self.min_buffer_size = min_buffer_size
         self.rollout_batch_size = rollout_batch_size
         self.reward_scale = reward_scale
-        self.evaluate = False
         self.input_include_goal = input_include_goal
         self.smooth_return = smooth_return
         self.max_path_length = max_path_length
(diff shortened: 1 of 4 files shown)


src/garage/np/algos/off_policy_rl_algorithm.py
```diff
@@ -65,7 +65,7 @@ def __init__(self,
         self.min_buffer_size = min_buffer_size
         self.rollout_batch_size = rollout_batch_size
         self.reward_scale = reward_scale
-        self.evaluate = False
+        self._first_buffer_size_steps_done = False
```
ryanjulian: This tracks whether the replay buffer is at least `min_buffer_size` right? So why not just test that directly?

`self.replay_buffer.n_transitions_stored >= self.min_buffer_size`

Perhaps the original author was concerned about having such a long conditional everywhere? If so, you can add a helper property:

```python
@property
def _buffer_prefilled(self):
    return self.replay_buffer.n_transitions_stored >= self.min_buffer_size
```

ryanjulian: This tracks whether the replay buffer is at least `min_buffer_size` right? So why not just test that directly?

`self.replay_buffer.n_transitions_stored >= self.min_buffer_size`

Perhaps the original author was concerned about having such a long conditional everywhere? If so, you can add a helper property:

```python
@property
def _buffer_prefilled(self):
    return self.replay_buffer.n_transitions_stored >= self.min_buffer_size
```

ryanjulian:
@krzentner is this PR still relevant?

PR #1140, opened 2020-01-21
Tanh Distribution for PyTorch Policies

avnishn:
Addition of tanh distribution used for policies
in SAC and PEARL.

### DIFF
--- a/src/garage/torch/distributions/__init__.py
+++ b/src/garage/torch/distributions/__init__.py
@@ -1,0 +1,4 @@
+"""PyTorch Custom Distributions."""
+from garage.torch.distributions.tanh_normal import TanhNormal
+
+__all__ = ['TanhNormal']
--- a/src/garage/torch/distributions/tanh_normal.py
+++ b/src/garage/torch/distributions/tanh_normal.py
@@ -1,0 +1,263 @@
+"""A Gaussian distribution with tanh transformation."""
+import torch
+from torch.distributions import Normal
+from torch.distributions.independent import Independent
+
+
+class TanhNormal(torch.distributions.Distribution):
+    r"""A distribution induced by applying a tanh transformation to a Gaussian random variable.
+
+    Algorithms like SAC and Pearl use this transformed distribution.
+    It can be thought of as a distribution of X where
+        :math: `Y ~ \mathcal{N}(\mu, \sigma)`
+        :math: `X = tanh(Y)`
+
+    Args:
+        loc (torch.Tensor): The mean of this distribution.
+        scale (torch.Tensor): The stdev of this distribution.
+
+    """ # noqa: 501
+
+    def __init__(self, loc, scale):
+        self._normal = Independent(Normal(loc, scale), 1)
+        super().__init__()
+
+    def log_prob(self, value, pre_tanh_value=None, epsilon=1e-6):
+        """The log likelihood of a sample on the this Tanh Distribution.
+
+        Args:
+            value (torch.Tensor): The sample whose loglikelihood is being
+                computed.
+            pre_tanh_value (torch.Tensor): The value prior to having the tanh
+                function applied to it but after it has been sampled from the
+                normal distribution.
+            epsilon (float): Regularization constant. Making this value larger
+                makes the computation more stable but less precise.
+
+        Note:
+              when pre_tanh_value is None, an estimate is made of what the
+              value is. This leads to a worse estimation of the log_prob.
+              If the value being used is collected from functions like
+              `sample` and `rsample`, one can instead use functions like
+              `sample_return_pre_tanh_value` or
+              `rsample_return_pre_tanh_value`
+
+
+        Returns:
+            torch.Tensor: The log likelihood of value on the distribution.
+
+        """
+        # pylint: disable=arguments-differ
+        if pre_tanh_value is None:
+            pre_tanh_value = torch.log((1 + value) / (1 - value)) / 2
+        norm_lp = self._normal.log_prob(pre_tanh_value)
+        ret = (norm_lp - torch.sum(
+            torch.log(self._clip_but_pass_gradient((1. - value**2)) + epsilon),
+            axis=-1))
+        return ret
+
+    def sample(self, sample_shape=torch.Size()):
+        """Return a sample, sampled from this TanhNormal Distribution.
+
+        Args:
+            sample_shape (list): Shape of the returned value.
+
+        Note:
+            Gradients `do not` pass through this operation.
+
+        Returns:
+            torch.Tensor: Sample from this TanhNormal distribution.
+
+        """
+        with torch.no_grad():
+            return self.rsample(sample_shape=sample_shape)
+
+    def rsample(self, sample_shape=torch.Size()):
+        """Return a sample, sampled from this TanhNormal Distribution.
+
+        Args:
+            sample_shape (list): Shape of the returned value.
+
+        Note:
+            Gradients pass through this operation.
+
+        Returns:
+            torch.Tensor: Sample from this TanhNormal distribution.
+
+        """
+        z = self._normal.rsample(sample_shape)
+        return torch.tanh(z)
+
+    def rsample_with_pre_tanh_value(self, sample_shape=torch.Size()):
+        """Return a sample, sampled from this TanhNormal distribution.
+
+        Returns the sampled value before the tanh transform is applied and the
+        sampled value with the tanh transform applied to it.
+
+        Args:
+            sample_shape (list): shape of the return.
+
+        Note:
+            Gradients pass through this operation.
+
+        Returns:
+            torch.Tensor: Samples from this distribution.
+            torch.Tensor: Samples from the underlying
+                :obj:`torch.distributions.Normal` distribution, prior to being
+                transformed with `tanh`.
+
+        """
+        z = self._normal.rsample(sample_shape)
+        return z, torch.tanh(z)
+
+    def cdf(self, value):
+        """Returns the CDF at the value.
+
+        Returns the cumulative density/mass function evaluated at
+        `value` on the underlying normal distribution.
+
+        Args:
+            value (torch.Tensor): The element where the cdf is being evaluated
+                at.
+
+        Returns:
+            torch.Tensor: the result of the cdf being computed.
+
+        """
+        return self._normal.cdf(value)
+
+    def icdf(self, value):
+        """Returns the icdf function evaluated at `value`.
+
+        Returns the icdf function evaluated at `value` on the underlying
+        normal distribution.
+
+        Args:
+            value (torch.Tensor): The element where the cdf is being evaluated
+                at.
+
+        Returns:
+            torch.Tensor: the result of the cdf being computed.
+
+        """
+        return self._normal.icdf(value)
+
+    @classmethod
+    def _from_distribution(cls, new_normal):
+        """Construct a new TanhNormal distribution from a normal distribution.
+
+        Args:
+            new_normal (Independent(Normal)): underlying normal dist for
+                the new TanhNormal distribution.
+
+        Returns:
+            TanhNormal: A new distribution whose underlying normal dist
+                is new_normal.
+
+        """
+        # pylint: disable=protected-access
+        new = cls(torch.zeros(1), torch.zeros(1))
+        new._normal = new_normal
+        return new
+
+    def expand(self, batch_shape, _instance=None):
+        """Returns a new TanhNormal distribution.
+
+        (or populates an existing instance provided by a derived class) with
+        batch dimensions expanded to `batch_shape`. This method calls
+        :class:`~torch.Tensor.expand` on the distribution's parameters. As
+        such, this does not allocate new memory for the expanded distribution
+        instance. Additionally, this does not repeat any args checking or
+        parameter broadcasting in `__init__.py`, when an instance is first
+        created.
+
+        Args:
+            batch_shape (torch.Size): the desired expanded size.
+            _instance(instance): new instance provided by subclasses that
+                need to override `.expand`.
+
+        Returns:
+            Instance: New distribution instance with batch dimensions expanded
+            to `batch_size`.
+
+        """
+        new_normal = self._normal.expand(batch_shape, _instance)
+        new = self._from_distribution(new_normal)
+        return new
+
+    def enumerate_support(self, expand=True):
+        """Returns tensor containing all values supported by a discrete dist.
+
+        The result will enumerate over dimension 0, so the shape
+        of the result will be `(cardinality,) + batch_shape + event_shape`
+        (where `event_shape = ()` for univariate distributions).
+
+        Note that this enumerates over all batched tensors in lock-step
+        `[[0, 0], [1, 1], ...]`. With `expand=False`, enumeration happens
+        along dim 0, but with the remaining batch dimensions being
+        singleton dimensions, `[[0], [1], ..`.
+
+        To iterate over the full Cartesian product use
+        `itertools.product(m.enumerate_support())`.
+
+        Args:
+            expand (bool): whether to expand the support over the
+                batch dims to match the distribution's `batch_shape`.
+
+        Note:
+            Calls the enumerate_support function of the underlying normal
+            distribution.
+
+        Returns:
+            torch.Tensor: Tensor iterating over dimension 0.
+
+        """
+        return self._normal.enumerate_support(expand)
+
+    @property
+    def mean(self):
+        """torch.Tensor: mean of the distribution."""
+        return torch.tanh(self._normal.mean)
+
+    @property
+    def variance(self):
+        """torch.Tensor: variance of the underlying normal distribution."""
+        return self._normal.variance
+
+    def entropy(self):
+        """Returns entropy of the underlying normal distribution.
+
+        Returns:
+            torch.Tensor: entropy of the underlying normal distribution.
+
+        """
+        return self._normal.entropy()
+
+    @staticmethod
+    def _clip_but_pass_gradient(x, lower=0., upper=1.):
+        """Clipping function that allows for gradients to flow through.
+
+        Args:
+            x (torch.Tensor): value to be clipped
+            lower (float): lower bound of clipping
+            upper (float): upper bound of clipping
+
+        Returns:
+            torch.Tensor: x clipped between lower and upper.
+
+        """
+        clip_up = (x > upper).float()
+        clip_low = (x < lower).float()
+        with torch.no_grad():
+            clip = ((upper - x) * clip_up + (lower - x) * clip_low)
+        return x + clip
+
+    def __repr__(self):
+        """Returns the parameterization of the distribution.
+
+        Returns:
+            str: The parameterization of the distribution and underlying
+                distribution.
+
+        """
+        return self.__class__.__name__
--- a/tests/garage/torch/distributions/test_tanh_normal_dist.py
+++ b/tests/garage/torch/distributions/test_tanh_normal_dist.py
@@ -1,0 +1,73 @@
+"""Tests for the tanh transformed normal distribution."""
+import torch
+
+from garage.torch.distributions import TanhNormal
+
+
+class TestBenchmarkTanhNormalDistribution:
+    """Tests for the tanh normal distribution."""
+
+    def test_new_tanh_normal(self):
+        """Tests the tanh_normal constructor."""
+        mean = torch.ones(1)
+        std = torch.ones(1)
+        dist = TanhNormal(mean, std)
+        del dist
+
+    def test_tanh_normal_bounds(self):
+        """Test to make sure the tanh_normal dist obeys the bounds (-1,1)."""
+        mean = torch.ones(1) * 100
+        std = torch.ones(1) * 100
+        dist = TanhNormal(mean, std)
+        assert dist.mean <= 1.
+        del dist
+        mean = torch.ones(1) * -100
+        std = torch.ones(1) * 100
+        dist = TanhNormal(mean, std)
+        assert dist.mean >= -1.
+
+    def test_tanh_normal_rsample(self):
+        """Test the bounds of the tanh_normal rsample function."""
+        mean = torch.zeros(1)
+        std = torch.ones(1)
+        dist = TanhNormal(mean, std)
+        sample = dist.rsample()
+        pre_tanh_action, action = dist.rsample_with_pre_tanh_value()
+        assert (pre_tanh_action.tanh() == action).all()
+        assert -1 <= action <= 1.
+        assert -1 <= sample <= 1.
+        del dist
+
+    def test_tanh_normal_log_prob(self):
+        """Verify the correctnes of the tanh_normal log likelihood function."""
+        mean = torch.zeros(1)
+        std = torch.ones(1)
+        dist = TanhNormal(mean, std)
+        pre_tanh_action = torch.Tensor([[2.0960]])
+        action = pre_tanh_action.tanh()
+        log_prob = dist.log_prob(action, pre_tanh_action)
+        log_prob_approx = dist.log_prob(action)
+        assert torch.allclose(log_prob, torch.Tensor([-0.2798519]))
+        assert torch.allclose(log_prob_approx, torch.Tensor([-0.2798519]))
+        del dist
+
+    def test_tanh_normal_expand(self):
+        """Test for expand function.
+
+        Checks whether expand returns a distribution that has potentially a
+        different batch size from the already existing distribution.
+
+        """
+        mean = torch.zeros(1)
+        std = torch.ones(1)
+        dist = TanhNormal(mean, std)
+        new_dist = dist.expand((2, ))
+        sample = new_dist.sample()
+        assert sample.shape == torch.Size((2, 1))
+
+    def test_tanh_normal_repr(self):
+        """Test that the repr function outputs the class name."""
+        mean = torch.zeros(1)
+        std = torch.ones(1)
+        dist = TanhNormal(mean, std)
+        assert repr(dist) == 'TanhNormal'


src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,117 @@
...
+        Args:
+            value(torch.tensor): The sample whose loglikelihood is being
+                                 computed.
+            pre_tanh_value(torch.tensor): The value prior to having the tanh
+                                          function applied to it but after it
+                                          has been sampled from the normal
+                                          distribution.
+
+        Returns:
+            torch.tensor with log likelihood.
```
lywong92: The documentation style should be:
```
Returns:
    torch.tensor: Log likelihood.
```

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,117 @@
...
+            named tuple with fields "pre_tanh_action" and "action"
+
+        """
+        action_infos = namedtuple('action_infos',
+                                  ['pre_tanh_action', 'action'])
+        with torch.no_grad():
+            z = self._normal.sample()
+            if return_pre_tanh_value:
+                return action_infos(z, torch.tanh(z))
+            return action_infos(None, torch.tanh(z))
```
lywong92: IMO, it looks cleaner with one return statement.

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,117 @@
...
+            with torch.no_grad():
+                clip = ((upper - x) * clip_up + (lower - x) * clip_low)
+            return x + clip
+
+        if pre_tanh_value is None:
+            pre_tanh_value = torch.log((1 + value) / (1 - value)) / 2
+        ret = self._normal.log_prob(pre_tanh_value) - torch.sum(torch.log(
+            _clip_but_pass_gradient((1. - value**2), lower=0., upper=1.) +
+            self._epsilon),
+                                                                axis=-1)
```
lywong92: Formatting seems a little off?

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,117 @@
...
+        if return_pre_tanh_value:
+            return action_infos(z, torch.tanh(z))
+        return action_infos(None, torch.tanh(z))
+
+    @property
+    def loc(self):  # noqa D102
+        return self._normal.loc
+
+    @property
+    def mean(self):  # noqa D102
```
krzentner: Might want to note in a docstring that this value is technically not correct (but much easier to compute than the true value, and typically quite close).

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,117 @@
...
+        return torch.tanh(self._normal.mean)
+
+    @property
+    def variance(self):  # noqa D102
```
krzentner: Ditto.

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,151 @@
+"""A Gaussian distribution with tanh transformation."""
+from collections import namedtuple
+
+import torch
+from torch.distributions import Distribution, MultivariateNormal
+
+
+class TanhNormal(Distribution):
+    # pylint: disable=abstract-method, arguments-differ
```
ryanjulian: which method did you not implement? why?

where do arguments differ?

please only disable something like arugments-differ directly at the violation line, not for the whole class

avnishn: 1) I don't implement some functions that are properties of a distribution. They aren't necessary for implementation or debugging so I have found.

2) I disabled arguments differ for log_prob and for r_sample, because I need the option for different types of returns.

avnishn: One such function I don't implement is a function that calculates a cdf on the distribution.

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,154 @@
...
+
+            """
+            clip_up = (x > upper).float()
+            clip_low = (x < lower).float()
+            with torch.no_grad():
+                clip = ((upper - x) * clip_up + (lower - x) * clip_low)
+            return x + clip
+
+        if pre_tanh_value is None:
+            pre_tanh_value = torch.log((1 + value) / (1 - value)) / 2
```
naeioi: I think we need a guard/clipping for the case when abs(x) >= 1.

src/garage/torch/distributions/tanh_normal.py
avnishn: I've fixed 2) for now

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,154 @@
...
+from torch.distributions import Distribution, MultivariateNormal
+
+
+class TanhNormal(Distribution):
+    # pylint: disable=abstract-method
+    """A gaussian distribution that has a tanh transformation applied to it.
+
+    Algorithms like SAC and Pearl use this transformed distribution.
+    It can be thought of as a distribution of X where
+        Y ~ N(mean, cov)
```
ryanjulian: the `~` is usually meant to me "is distributed as" which is not what you mean here.

what you are saying is
```
.. :math:
   Y ~ \mathcal{N}(\mu, \sigma)
   X = tanh(Y)
```

also, this is an excellent opportunity to use the `:math:` environment in sphinx so that your docstring gets turned into rendered latex.

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,154 @@
...
+        X ~ tanh(Y)
+
+    Args:
+        mean(torch.tensor): The mean of this distribution.
+        cov(torch.tensor): The covariance of this distribution.
```
ryanjulian: are you sure this is covariance? doesn't `torch.distributions.Normal` use stddev?

also, `torch.distributions.Normal` uses the named `loc` and `scale` -- please follow that API so that users are not surprised.

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,154 @@
...
+        epsilon(float): Coefficient of stability used when calculating
```
ryanjulian: This is a regularization coefficient. Should it be larger or smaller to make the calculation more stable?

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,154 @@
...
+        Args:
+            return_pre_tanh_value(bool): When true, returns the sampled value
+                                         before the Tanh transform is applied.
+
+        Returns:
+            action_infos: named tuple with fields
+                          "pre_tanh_action" and "action"
+
+        """
+        action_infos = namedtuple('action_infos',
```
naeioi: A nested namedtuple is not picklable. I suggest using a dict or put this namedtuple to the top level of this file.

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,154 @@
...
+
+    """
+
+    def __init__(self, mean, cov, epsilon=1e-6):
+        self._normal = MultivariateNormal(mean, cov)
+        self._epsilon = epsilon
+        super().__init__()
+
+    def log_prob(self, value, pre_tanh_value=None):
+        # pylint: disable=arguments-differ
```
ryanjulian: why do args differ? please justify every non-obvious pylint disable with a nearby comment explaining why

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,154 @@
...
+        Args:
+            value(torch.tensor): The sample whose loglikelihood is being
+                                 computed.
+            pre_tanh_value(torch.tensor): The value prior to having the tanh
+                                          function applied to it but after it
+                                          has been sampled from the normal
+                                          distribution.
+
+        Notes:
+            - pre_tanh_value should usually not be None, but as the option
```
ryanjulian: if you don't have any usages where this is the case, please remove the option. it makes the API simpler and the test coverage easier.

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,154 @@
...
+            available just in case a value is not/cannot be passed.
+
+        Returns:
+            torch.tensor: log likelihood.
```
ryanjulian: Please use a full english phrase

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,154 @@
...
+        super().__init__()
+
+    def log_prob(self, value, pre_tanh_value=None):
+        # pylint: disable=arguments-differ
+        """The log likelihood of a sample on the this Tanh Distribution.
+
+        Args:
+            value(torch.tensor): The sample whose loglikelihood is being
+                                 computed.
+            pre_tanh_value(torch.tensor): The value prior to having the tanh
```
ryanjulian: our docstrings are indented just like code, 4 spaces for hanging indents (not aligned with the `:`)

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,154 @@
...
+                                          function applied to it but after it
+                                          has been sampled from the normal
```
ryanjulian: Normal (caps)

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,154 @@
...
+
+    def log_prob(self, value, pre_tanh_value=None):
+        # pylint: disable=arguments-differ
+        """The log likelihood of a sample on the this Tanh Distribution.
+
+        Args:
+            value(torch.tensor): The sample whose loglikelihood is being
+                                 computed.
+            pre_tanh_value(torch.tensor): The value prior to having the tanh
+                                          function applied to it but after it
```
ryanjulian: what is "it"?

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,154 @@
...
+    # pylint: disable=abstract-method
+    """A gaussian distribution that has a tanh transformation applied to it.
+
+    Algorithms like SAC and Pearl use this transformed distribution.
+    It can be thought of as a distribution of X where
+        Y ~ N(mean, cov)
+        X ~ tanh(Y)
+
+    Args:
+        mean(torch.tensor): The mean of this distribution.
```
ryanjulian: there should be spaces between all of these argument names and their types
`mean (torch.Tensor)`

src/garage/torch/distributions/tanh_normal.py
ryanjulian: torch.Tensor

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,154 @@
+"""A Gaussian distribution with tanh transformation."""
+from collections import namedtuple
+
+import torch
+from torch.distributions import Distribution, MultivariateNormal
```
ryanjulian: please don't import naked symbols from third-party libraries. only import whole packages. see https://github.com/rlworkgroup/garage/blob/master/CONTRIBUTING.md#garage-specific-python-style

```python
import torch.distributions

class TanhNormal(torch.distributions.Distribution):
```

src/garage/torch/distributions/tanh_normal.py
ryanjulian: torch.Tensor

src/garage/torch/distributions/tanh_normal.py
ryanjulian: also, notes don't need bullets if there's only one of them

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,154 @@
...
+        Returns:
+            torch.tensor: log likelihood.
+
+        """
+
+        def _clip_but_pass_gradient(x, lower=0., upper=1.):
+            """Clipping function that allows for gradients to flow through.
+
+            Args:
+                x(torch.tensor): value to be clipped
```
ryanjulian: torch.Tensor

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,154 @@
...
+        Notes:
+            - pre_tanh_value should usually not be None, but as the option
+            available just in case a value is not/cannot be passed.
+
+        Returns:
+            torch.tensor: log likelihood.
+
+        """
+
+        def _clip_but_pass_gradient(x, lower=0., upper=1.):
```
ryanjulian: FYI _private methods don't need docstrings (but are welcome)

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,154 @@
...
+                    _clip_but_pass_gradient((1. - value**2))  # noqa: E128 E501
+                    + self._epsilon),
+                axis=-1))
+        return ret
+
+    def sample(self, return_pre_tanh_value=False):
+        # pylint: disable=arguments-differ
+        """Return a sample, sampled from this Tanh Normal Distribution.
+
+        Gradients should *not* pass through this operation.
```
ryanjulian: isn't this true in torch.Distributions, because rsample is the differentiable path?

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,154 @@
...
+
+        Args:
+            return_pre_tanh_value(bool): When true, returns the sampled value
+                                         before the Tanh transform is applied.
+
+        Returns:
+            action_infos: named tuple with fields
```
ryanjulian: this is strange -- doesn't the parent API return just a value? in that case the anagous thing to do is to just use multiple returns (i.e. value, pre_tanh_value = foo.sample())

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,154 @@
...
+            pre_tanh_value = torch.log((1 + value) / (1 - value)) / 2
+        ret = (
+            self._normal.log_prob(pre_tanh_value) - torch.sum(
+                torch.log(
+                    _clip_but_pass_gradient((1. - value**2))  # noqa: E128 E501
+                    + self._epsilon),
+                axis=-1))
+        return ret
+
+    def sample(self, return_pre_tanh_value=False):
```
ryanjulian: doesn't the parent API take a `sample_shape` argument? you can't just declare you're implementing `Distribution` and then not do that.

ryanjulian: to keep parity with the `Distribution` API, you can implement `sample` without the new argument/return value and then add some other method e.g. `sample_and_transform` with the semantics you want.

ryanjulian: sorry but no. the return behavior of a method should never be affected by its arguments.

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,154 @@
...
+        action_infos = namedtuple('action_infos',
+                                  ['pre_tanh_action', 'action'])
+        with torch.no_grad():
+            z = self._normal.sample()
+            if return_pre_tanh_value:
+                return action_infos(z, torch.tanh(z))
+            return action_infos(None, torch.tanh(z))
+
+    def rsample(self, return_pre_tanh_value=False):
+        # pylint: disable=arguments-differ
```
ryanjulian: please faithfully implement the interface. see my comments on sample.

src/garage/torch/distributions/tanh_normal.py
ryanjulian: is there a reason this can't be a private method? inline methods are very rare in Python

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,154 @@
...
+        mean(torch.tensor): The mean of this distribution.
+        cov(torch.tensor): The covariance of this distribution.
+        epsilon(float): Coefficient of stability used when calculating
+                        the log likehilhood of a sample on a distribution.
+
+    """
+
+    def __init__(self, mean, cov, epsilon=1e-6):
+        self._normal = MultivariateNormal(mean, cov)
+        self._epsilon = epsilon
```
ryanjulian: AFAICT epsilon is only used by `log_prob`, so why not make it an optional parameter of that function rather than the class?

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,154 @@
...
+        z = self._normal.rsample()
+        if return_pre_tanh_value:
+            return action_infos(z, torch.tanh(z))
+        return action_infos(None, torch.tanh(z))
+
+    @property
+    def mean(self):
+        """Returns mean of the distribution.
+
+        Note: This is the mean of the underlying normal distribution
```
ryanjulian: i think you are supposed to put the note body on a new line?

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,154 @@
...
+        action_infos = namedtuple('action_infos',
+                                  ['pre_tanh_action', 'action'])
+        z = self._normal.rsample()
+        if return_pre_tanh_value:
+            return action_infos(z, torch.tanh(z))
+        return action_infos(None, torch.tanh(z))
+
+    @property
+    def mean(self):
+        """Returns mean of the distribution.
```
ryanjulian: properties are documented like attributes. see https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,154 @@
...
+
+        Returns:
+            torch.tensor: mean of the distribution
+
+        """
+        return torch.tanh(self._normal.mean)
+
+    @property
+    def variance(self):
+        """Returns variance of the underlying normal distribution.
```
ryanjulian: documented like an attribute

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,154 @@
...
+
+        Returns:
+            torch.tensor: variance of the underlying normal distribution.
+
+        """
+        return self._normal.variance
+
+    def entropy(self):
```
ryanjulian: documented like an attribute

src/garage/torch/distributions/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""PyTorch Custom Distributions."""
+from garage.torch.distributions.tanh_normal import TanhNormal
```
ryanjulian: i think there should be a newline before `__all__`?

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,154 @@
+"""A Gaussian distribution with tanh transformation."""
+from collections import namedtuple
+
+import torch
+from torch.distributions import Distribution, MultivariateNormal
+
+
+class TanhNormal(Distribution):
+    # pylint: disable=abstract-method
```
ryanjulian: which method did you skip?

src/garage/torch/distributions/tanh_normal.py
ryanjulian: use Normal, not MultivariateNormal unless your covariance matrix is non-diagonal

src/garage/torch/distributions/tanh_normal.py
ryanjulian: you can also use the same fix i suggested below (new API for this functionality)

avnishn: resolved

ryanjulian:
To raise your coverage you just need to mock out `self._normal`

Please try to get this merged ASAP

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,290 @@
...
+        Y ~ N(mean, cov)
+        X ~ tanh(Y)
+
+    Args:
+        mean (torch.tensor): The mean of this distribution.
+        std (torch.tensor): The stdev of this distribution.
+
+    """
+
+    def __init__(self, mean, std):
```
ryanjulian: `Normal` uses `loc, scale` -- please stick to that convention

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,290 @@
...
+        with torch.no_grad():
+            clip = ((upper - x) * clip_up + (lower - x) * clip_low)
+        return x + clip
+
+    def __str__(self):
+        """Returns the Name of the class.
+
+        Returns:
+            str: The class name.
+
```
ryanjulian: is this consistent with what PyTorch does?

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,290 @@
...
+        """
+        return 'TanhNormal'
+
+    def __repr__(self):
+        """Returns the parameterization of the distribution.
+
+        Returns:
+            str: The parameterization of the distribution and underlying
+                distribution.
```
ryanjulian: is this consistent with what PyTorch does?

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,290 @@
...
+        # pylint: disable=arguments-differ
+        if pre_tanh_value is None:
+            pre_tanh_value = torch.log((1 + value) / (1 - value)) / 2
+        norm_lp = self._normal.log_prob(pre_tanh_value)
+        ret = (norm_lp - torch.sum(
+            torch.log(self._clip_but_pass_gradient((1. - value**2)) + epsilon),
+            axis=-1))
+        return ret
+
+    def sample_return_pre_tanh_value(self, sample_shape=torch.Size()):
```
ryanjulian: perhaps `sample_unwrapped` or `sample_with_pretanh_value`?

i looked in your SAC PR and couldn't find any usage of this function. What is it for?

Would it be sufficient to provide `sample_unwrapped` (which samples the wrapped distribution) and `sample` (which provides the wrapped samples)? does anyone use both values at once in the differentiable path?

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,290 @@
...
+        Returns the sampled value before the Tanh transform is applied and the
+        sampled value with the Tanh transform applied to it.
+
+        Args:
+            sample_shape (list): Shape of the returned value.
+
+        Note: Gradients `do not` pass through this operation.
+
+        Returns:
+            action_infos: named tuple with fields
```
ryanjulian: returning a namedtuple from an ordinary method with multiple return values is not very pythonic. this just adds overhead. just use multiple returns.

returning a tuple in Python is equivalent to multiple returns, so you should document this as-such

```python
"""
    Returns:
        torch.Tensor: Samples from this distribution.
        torch.Tensor: Samples from the underlying :obj:`torch.distributions.Normal` distribution, prior to being transformed with `tanh`.

File: src/garage/torch/distributions/tanh_normal.py [hunk 38]
```diff
@@ -0,0 +1,290 @@
+"""A Gaussian distribution with tanh transformation."""
+from collections import namedtuple
+
+import torch
+from torch.distributions import Normal
+from torch.distributions.independent import Independent
+
+
+class TanhNormal(torch.distributions.Distribution):
+    """A gaussian distribution that has a tanh transformation applied to it.
```
ryanjulian: Gaussian (caps)

A distribution induced by applying a tanh transformation to a Gaussian random variable.

avnishn: resolved.

src/garage/torch/distributions/tanh_normal.py
avnishn: I'll change the names but keep the function. There is a chance that the function is needed later on (not in SAC however)

ryanjulian: if the function has no known use or user then it should be deleted.

src/garage/torch/distributions/tanh_normal.py
avnishn: It isn't but also, they have no specific rules that I can find.

avnishn: This is really only useful for debugging and logging.

avnishn:
> To raise your coverage you just need to mock out `self._normal`
> 
> Please try to get this merged ASAP

@ryanjulian I'm unclear how mocking out self._normal will raise my test coverage.
Looking at the codecov report, it seems that where I am lacking is in having tests for 
attributes and the __str__ and __repr__ functions

ryanjulian:
how else would you test the attributes other than by mocking `self._normal`??

src/garage/torch/distributions/tanh_normal.py
ryanjulian: i'm 100% aware of what repr does.

what i meant was "make the output of this look exactly like it came from torch.distributions`

src/garage/torch/distributions/tanh_normal.py
avnishn: I've removed it.

src/garage/torch/distributions/tanh_normal.py
avnishn: ok.

src/garage/torch/distributions/tanh_normal.py
avnishn: distributions doesn't implement this function. I've gone ahead and removed it.

src/garage/torch/distributions/tanh_normal.py
ryanjulian: What was the original use for it? When did you decide to remove it? (previously i thought it was essential to SAC working).

avnishn: no this function isn't necessary to making sac work. `rsample_with_pretanh_value` is necessary to SAC working.

avnishn:
@Mergifyio rebase

src/garage/torch/distributions/tanh_normal.py
ryanjulian: oh, in that case i take back what i said. yes you want the API to be symmetric so you should have both methods

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,262 @@
+"""A Gaussian distribution with tanh transformation."""
+import torch
+from torch.distributions import Normal
+from torch.distributions.independent import Independent
+
+
+class TanhNormal(torch.distributions.Distribution):
+    r"""A distribution induced by applying a tanh transformation to a Gaussian random variable.
+
+    Algorithms like SAC and Pearl use this transformed distribution.
```
ryanjulian: PEARL

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,262 @@
...
+        Args:
+            value (torch.Tensor): The sample whose loglikelihood is being
+                computed.
+            pre_tanh_value (torch.Tensor): The value prior to having the tanh
+                function applied to it but after it has been sampled from the
+                normal distribution.
+            epsilon (float): Stabilization coefficient.
+
+        Note:
+              when pre_tanh_value is None, an estimate is made of what the
```
ryanjulian: This needs to be a properly capitalized and punctuated english paragraph

src/garage/torch/distributions/tanh_normal.py
ryanjulian: wait. is this not a property?

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,262 @@
...
+              value is. This leads to a worse estimation of the log_prob.
+              If the value being used is collected from functions like
+              `sample` and `rsample`, one can instead use functions like
+              `sample_return_pre_tanh_value` or
```
ryanjulian: didn't you remove this?

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,262 @@
...
+
+        Returns:
+            torch.Tensor: Sample from this Tanh Normal distribution.
+
+        """
+        with torch.no_grad():
+            return self.rsample(sample_shape=sample_shape)
+
+    def rsample(self, sample_shape=torch.Size()):
+        """Return a sample, sampled from this Tanh Normal Distribution.
```
ryanjulian: why is distribution capitalized? isn't this a TanhNormal distribution?

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,262 @@
...
+            torch.Tensor: x clipped between lower and upper.
+
+        """
+        clip_up = (x > upper).float()
+        clip_low = (x < lower).float()
+        with torch.no_grad():
+            clip = ((upper - x) * clip_up + (lower - x) * clip_low)
+        return x + clip
+
+    def __repr__(self):
```
ryanjulian: repr doesn't return a parameterization...

Just call it a string representation

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,262 @@
...
+    Algorithms like SAC and Pearl use this transformed distribution.
+    It can be thought of as a distribution of X where
+        :math: `Y ~ \mathcal{N}(\mu, \sigma)`
+        :math: `X = tanh(Y)`
+
+    Args:
+        loc (torch.Tensor): The mean of this distribution.
+        scale (torch.Tensor): The stdev of this distribution.
+
+    """ # noqa: 501
```
ryanjulian: wait, why?

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,262 @@
+"""A Gaussian distribution with tanh transformation."""
+import torch
+from torch.distributions import Normal
+from torch.distributions.independent import Independent
+
+
+class TanhNormal(torch.distributions.Distribution):
+    r"""A distribution induced by applying a tanh transformation to a Gaussian random variable.
```
ryanjulian: this is supposed to fit on one line. the solution is not to also exceed the line length :P

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,262 @@
...
+        scale (torch.Tensor): The stdev of this distribution.
+
+    """ # noqa: 501
+
+    def __init__(self, loc, scale):
+        self._normal = Independent(Normal(loc, scale), 1)
+        super().__init__()
+
+    def log_prob(self, value, pre_tanh_value=None, epsilon=1e-6):
+        """The log likelihood of a sample on the this Tanh Distribution.
```
ryanjulian: is it a Tanh Distribution or a TanhNormal distribution?

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,262 @@
...
+
+        Args:
+            value (torch.Tensor): The sample whose loglikelihood is being
+                computed.
+            pre_tanh_value (torch.Tensor): The value prior to having the tanh
+                function applied to it but after it has been sampled from the
+                normal distribution.
```
ryanjulian: Normal

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,262 @@
...
+            epsilon (float): Stabilization coefficient.
```
ryanjulian: Regularization constant. Making this value larger (smaller?) makes the computation more stable but less precise.

src/garage/torch/distributions/tanh_normal.py
```diff
@@ -0,0 +1,262 @@
...
+        if pre_tanh_value is None:
+            pre_tanh_value = torch.log((1 + value) / (1 - value)) / 2
+        norm_lp = self._normal.log_prob(pre_tanh_value)
+        ret = (norm_lp - torch.sum(
+            torch.log(self._clip_but_pass_gradient((1. - value**2)) + epsilon),
+            axis=-1))
+        return ret
+
+    def sample(self, sample_shape=torch.Size()):
+        """Return a sample, sampled from this Tanh Normal Distribution.
```
ryanjulian: TanhNormal distribution

ryanjulian:
@mergifyio rebase

src/garage/torch/distributions/tanh_normal.py
avnishn: But there isn't a way to express this in a shorter way

src/garage/torch/distributions/tanh_normal.py
avnishn: larger

PR #1142, opened 2020-01-21
Base class and utilities for Meta-RL Algorithms

krzentner:
I have not yet tested the code in this change, although I will soon (and include them in this PR).

The main goal here is to be efficient with environment instances while meeting the needs of all Meta-RL algorithms. I also add some basic functionality needed for using the new sampler API with PyTorch.


src/garage/experiment/local_runner.py
```diff
@@ -497,6 +504,17 @@ def total_env_steps(self):
         """
         return self._stats.total_env_steps
 
+    @total_env_steps.setter
```
ryanjulian: this is this needed?

src/garage/experiment/meta_evaluator.py
```diff
@@ -0,0 +1,102 @@
...
+from garage.misc.tensor_utils import discount_cumsum
+from garage.sampler import LocalSampler
+
+
+class MetaEvaluator:
+    """Evaluates Meta-RL algorithms on test environments.
+
+    Args:
+        runner (garage.experiment.LocalRunner): Runner.
+        test_task_sampler (garage.experiment.TaskSampler): Sampler for test
```
ryanjulian: can this go into something like `garage.meta` or `garage.mtrl`?

src/garage/experiment/meta_evaluator.py
```diff
@@ -0,0 +1,102 @@
...
+            tasks. To demonstrate the effectiveness of a meta-learning method,
+            these should be different from the training tasks.
+        max_path_length (int): Maximum path length used for evaluation
+            trajectories.
+        n_test_tasks (int or None): Number of test tasks to sample each time
+            evaluation is performed. Note that tasks are sampled "without
+            replacement". If None, is set to `test_task_sampler.n_tasks`.
+        prefix (str): Prefix to use when logging. Defaults to MetaTest. For
+            example, this results in logging the key 'MetaTest/SuccessRate'.
+            If not set to `MetaTest`, it should probably be set to `MetaTrain`.
```
ryanjulian: can you make trivial subclasses `MetaTestEvaluator` and `MetaTrainEvaluator`?

src/garage/experiment/meta_evaluator.py
```diff
@@ -0,0 +1,102 @@
...
+                 test_task_sampler,
+                 max_path_length,
+                 n_test_tasks=None,
+                 prefix='MetaTest'):
+        self._test_task_sampler = test_task_sampler
+        if n_test_tasks is None:
+            n_test_tasks = test_task_sampler.n_tasks
+        self._n_test_tasks = n_test_tasks
+        self._test_sampler = runner.make_sampler(
+            LocalSampler, n_workers=1, max_path_length=max_path_length)
```
ryanjulian: should the number of workers be configurable? what if this is expensive?

src/garage/experiment/meta_evaluator.py
```diff
@@ -0,0 +1,102 @@
...
+            traj = self._test_sampler.obtain_samples(self._eval_itr, 1, policy,
+                                                     env_up)
+            adapted_policy = algo.adapt(traj)
+            adapted_traj = self._test_sampler.obtain_samples(
+                self._eval_itr, 1, adapted_policy)
+            adapted_trajectories.append(adapted_traj)
+        self.log_evaluation(adapted_trajectories,
+                            getattr(algo, 'discount', 1.0))
+
+    def log_evaluation(self, trajectories, discount):
```
ryanjulian: is there a way to share some or all of this logic with `RLAlgorithm.evaluate_performance`?

can this be called `evaluate_performance`?

how about
```python
def log_evaluation(self, trajectories, discount):

    with tabular.prefix(self._prefix):
        RLAlgorithm.evaluate_performance(trajectories, discount)
```

src/garage/experiment/task_sampler.py
```diff
@@ -0,0 +1,227 @@
...
+                least once in this batch. Ignored for continuous task spaces.
+
+        Returns:
+            list[EnvUpdate]: Batch of sampled environment updates, which, when
+                invoked on environments, will configure them with new tasks.
+                See :py:class:`~EnvUpdate` for more information.
+
+        """
+
+    @property
```
ryanjulian: document properties like attributes

ryanjulian: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

src/garage/np/algos/__init__.py
```diff
@@ -3,9 +3,11 @@
 from garage.np.algos.batch_polopt import BatchPolopt
 from garage.np.algos.cem import CEM
 from garage.np.algos.cma_es import CMAES
+from garage.np.algos.meta_algorithm import MetaRLAlgorithm
```
ryanjulian: meta_rl_algorithm

src/garage/np/algos/meta_algorithm.py
```diff
@@ -0,0 +1,35 @@
+"""Interface of Meta-RL ALgorithms."""
+import abc
+
+from garage.np.algos import RLAlgorithm
+
+
+class MetaRLAlgorithm(RLAlgorithm):
```
ryanjulian: can this also host the evaluation calculation code, as in `RLAlgorithm`? these two classes should at least be consistent.

src/garage/np/algos/meta_algorithm.py
```diff
@@ -0,0 +1,35 @@
...
+    """Base class for Meta-RL Algorithms."""
+
+    @abc.abstractmethod
+    def new_exploration_policy(self):
```
ryanjulian: i'm not sure about this name because it implies (to me) that a `__new__` is being called somewhere. can it be called `get_exploration_policy` instead?

if getting the exploration policy never implies non-trivial work, you could make this a property (`exploration_policy`). however i'm not sure that's necessarily the case for all algorithms.

src/garage/np/algos/meta_algorithm.py
```diff
@@ -0,0 +1,35 @@
...
+        """
+
+    @abc.abstractmethod
+    def adapt_policy(self, exploration_policy, trajectory):
+        """Produce a policy adapted for a task.
+
+        Args:
+            exploration_policy (garage.Policy): The policy used to explore the
+                environment. Must not be used after this method is called.
+            trajectory (garage.TrajectoryBatch): Trajectory to adapt to,
```
ryanjulian: can this be `trajectories`?

src/garage/sampler/env_update.py
```diff
@@ -0,0 +1,134 @@
...
+        return self._env
+
+    def __getstate__(self):
+        """Get the pickle state.
+
+        Returns:
+            dict: The pickled state.
+
+        """
+        warnings.warn('ExistingEnvUpdate is generally not the most efficient '
```
ryanjulian: maybe test this and see how annoying it is?

src/garage/sampler/worker.py
```diff
@@ -141,13 +143,18 @@ def update_agent(self, agent_update):
         """Update an agent, assuming it implements garage.Policy.
 
         Args:
-            agent_update(Dict[str, np.array]): Parameters to agent, which
-                should have been generated by calling
-                `policy.get_param_values`. Note that other implementations of
-                `Worker` may take different types for this parameter.
+            agent_update (np.ndarray or dict or garage.Policy): If a
```
ryanjulian: can we make this just like `env_update`? basically just a callable that's applied to the policy.

src/garage/torch/policies/deterministic_mlp_policy.py
```diff
@@ -64,7 +65,7 @@ def get_action(self, observation):
 
         """
         with torch.no_grad():
-            x = self.forward(observation.unsqueeze(0))
+            x = self.forward(torch.Tensor(observation).unsqueeze(0))
```
ryanjulian: shouldn't this accept numpy array, not torch.Tensor

ryanjulian: also, shouldn't it return ndarray?

ryanjulian: same for get_actions

src/garage/torch/policies/gaussian_mlp_policy.py
```diff
@@ -115,7 +116,7 @@ def get_action(self, observation):
 
         """
         with torch.no_grad():
-            observation = observation.unsqueeze(0)
+            observation = torch.Tensor(observation).unsqueeze(0)
```
ryanjulian: same typing issues here

src/garage/experiment/meta_evaluator.py
krzentner: I don't really think that's necessary? Just pass the parameter.

src/garage/experiment/meta_evaluator.py
krzentner: If it's too expensive (which seems fairly unlikely), I can add another worker and parallel sampling or something in a later PR.

krzentner: BTW, adding more workers here (with `LocalSampler`), wouldn't actually make anything faster.

src/garage/experiment/meta_evaluator.py
krzentner: Yep. Heck of a refactor though. Done.

src/garage/experiment/task_sampler.py
```diff
@@ -0,0 +1,227 @@
...
+    else:
+        blocks = []
+        for _ in range(math.ceil(n_to_sample / n_available_tasks)):
+            s = np.arange(n_available_tasks)
+            np.random.shuffle(s)
+            blocks.append(s)
+        return np.concatenate(blocks)[:n_to_sample]
+
+
+class TaskSampler:
```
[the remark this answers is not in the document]
krzentner: Yes

src/garage/experiment/task_sampler.py
krzentner: Neat, this actually encourages me to use properties more.

src/garage/np/algos/meta_algorithm.py
krzentner: I'm not convinced that's actually a great pattern, mostly because then all of the test data needs to get passed into the algorithm, where it could potentially get mixed up with training data. Let's try it being separate for now.

src/garage/np/algos/meta_algorithm.py
krzentner: Yeah, good catch.

src/garage/torch/policies/deterministic_mlp_policy.py
krzentner: Done.

src/garage/torch/policies/gaussian_mlp_policy.py
krzentner: Done.

krzentner:
The code coverage looks pretty bad because I marked the tests that use metaworld as huge. Most of the functionality is tested, aside from the `MetaEvaluator` itself, since I don't have any implementations to test against. I think this is about ready to merge (although refactors have made it massively balloon in size).

ryanjulian:
Regarding the tests -- can you boost coverage by making a multi-task version of something trivial like PointEnv? Or a mock?

setup.py
```diff
@@ -39,6 +39,7 @@
 EXTRAS['mujoco'] = [
     'mujoco-py<2.1,>=2.0',
     'gym[all]' + GYM_VERSION,
+    'metaworld @ git+https://github.com/rlworkgroup/metaworld.git@dfdbc7cf495678ee96b360d1e6e199acc141b36c',  # noqa: E501
```
ryanjulian: hmm as the maintainer of this package i'm way of putting this in a garage release. the package isn't very stable or even really packaged very much. at best i'd put this in a different extra like `benchmarks` or something.

src/garage/_functions.py
```diff
@@ -0,0 +1,49 @@
+"""Functions exposed directly in the garage namespace."""
+from dowel import tabular
+import numpy as np
+
+from garage.misc.tensor_utils import discount_cumsum
+
+
+def log_performance(itr, batch, discount, prefix='Evaluation'):
```
ryanjulian: should we keep propagating `prefix=` args, or expect people to use `tabular.prefix`?

src/garage/experiment/meta_evaluator.py
ryanjulian: The only reason I asked was to help keep logging consistent for meta-RL algos

src/garage/experiment/task_sampler.py
```diff
@@ -0,0 +1,224 @@
...
+            list[EnvUpdate]: Batch of sampled environment updates, which, when
+                invoked on environments, will configure them with new tasks.
+                See :py:class:`~EnvUpdate` for more information.
+
+        """
+
+    @property
+    def n_tasks(self):
+        """int or None: The number of tasks if known and finite."""
+        # pylint: disable=redundant-returns-doc
```
ryanjulian: does this exception still apply?

src/garage/np/algos/meta_rl_algorithm.py
```diff
@@ -0,0 +1,35 @@
...
+
+    @abc.abstractmethod
+    def get_exploration_policy(self):
+        """Return a policy used before adaptation to a specific task.
+
+        Each time it is retrieved, this policy should only be evaluated in one
+        task.
+
+        Returns:
+            garage.Policy: The "un-adapted" policy.
```
ryanjulian: Hmm i think this could be more specific. You can say that it's the policy used to obtain samples which are later used for meta-RL adaptation. It's up to the algorithm to decide what this policy will be (for instance, it could be a random walk).

src/garage/np/algos/meta_rl_algorithm.py
```diff
@@ -0,0 +1,35 @@
...
+
+        """
+
+    @abc.abstractmethod
+    def adapt_policy(self, exploration_policy, trajectories):
+        """Produce a policy adapted for a task.
+
+        Args:
+            exploration_policy (garage.Policy): The policy used to explore the
```
ryanjulian: the exploration policy and pre-adaptation policy don't need to be the same. 

what shall it not be used for?
what are the chances people actually follow this rule?
are you saying this method will mutate the policy?
in the case of a naive meta-RL based continual learner (e.g. RL2), does this even apply?

src/garage/np/algos/meta_rl_algorithm.py
```diff
@@ -0,0 +1,35 @@
...
+        Each time it is retrieved, this policy should only be evaluated in one
+        task.
+
+        Returns:
+            garage.Policy: The "un-adapted" policy.
+
+        """
+
+    @abc.abstractmethod
+    def adapt_policy(self, exploration_policy, trajectories):
```
ryanjulian: perhaps this should be called `exploration_trajectories` or `new_task_trajectories`

src/garage/np/algos/meta_rl_algorithm.py
```diff
@@ -0,0 +1,35 @@
...
+        """Produce a policy adapted for a task.
+
+        Args:
+            exploration_policy (garage.Policy): The policy used to explore the
+                environment. Must not be used after this method is called.
+            trajectories (garage.TrajectoryBatch): Trajectories to adapt to,
+                generated by exploration_policy exploring the environment.
+
+        Returns:
+            garage.Policy: The adapted policy.
```
ryanjulian: perhaps be more explicit: "The policy adapted to the new task data represented in exploration_trajectories"

src/garage/np/policies/fixed_policy.py
```diff
@@ -0,0 +1,105 @@
+"""Policy that performs a fixed sequence of actions."""
+import numpy as np
+
+from garage.misc.tensor_utils import stack_tensor_dict_list
+from garage.np.policies.base import Policy
+
+
+class FixedPolicy(Policy):
```
ryanjulian: other frameworks call this `ScriptedPolicy` so I vote we adopt that name.

src/garage/sampler/worker.py
ryanjulian: can you leave an issue about making a type for parameters so that this interface can be less messy?

src/garage/np/policies/fixed_policy.py
krzentner: We already have `ScriptedPolicy`, but it does something totally different. We can consider renaming these sometime in the future.

src/garage/np/algos/meta_rl_algorithm.py
krzentner: The goal here is to express that ownership of the policy returned by `get_exploration_policy()` is returned to the meta algorithm. This makes it as simple as possible for the implementer of the meta algorithm.

setup.py
krzentner: After some offline discussion, we decided it's fine to include it in `dev`, so that it can be used in tests for the time being.

src/garage/_functions.py
krzentner: Honestly, I do think propagating `prefix=` args is kinda an anti-pattern, although it is more concise than using `tabular.prefix` (at least in the cases so far). It's convenient for now and it's certainly better than what preceded it (manually implementing tabular prefixes repeatedly), so I say we keep it for the time being.

src/garage/np/algos/meta_rl_algorithm.py
ryanjulian: okay i didn't actually get that from your docstring. this seems subtle enough that maybe it needs to be handeld by a note and maybe a short example or two.

PR #1153, opened 2020-01-25
Meta RL sampler improvements

krzentner:
This PR contains a few improvements that I think are essential for easily using the new sampler architecture for meta (and multi-task) RL. They are:
 - Allow passing multiple environments / `EnvUpdate`s to sampler init methods. This means that, if the set of tasks is known ahead of time, the environments can potentially just be set once, when the sampler is initialized.
 - Allow sampling exactly `n` trajectories per worker. This makes gather, e.g. three trajectories per environment easy. See the `RaySampler` or `LocalSampler` tests for examples.
 - Using lambda's inside `EnvUpdate`s for `RaySampler` (this was already possible in `LocalSampler`). This ensures that e.g. `SetTaskSampler` or `ConstructEnvsSampler` can easily be used with environment wrappers with `RaySampler`.

I believe that @ahtsan might need some of these changes for his more efficient RL^2 sampling.

### DIFF
--- a/src/garage/sampler/local_sampler.py
+++ b/src/garage/sampler/local_sampler.py
@@ -36,8 +36,8 @@ class LocalSampler(Sampler):
             worker_factory(i) for i in range(worker_factory.n_workers)
         ]
         for worker, agent, env in zip(self._workers, self._agents, self._envs):
-            worker.agent = agent
-            worker.env = env
+            worker.update_agent(agent)
+            worker.update_env(env)
 
     @classmethod
     def from_worker_factory(cls, worker_factory, agents, envs):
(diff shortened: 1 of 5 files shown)


src/garage/sampler/local_sampler.py
```diff
@@ -100,6 +116,38 @@ def obtain_samples(self, itr, num_samples, agent_update, env_update=None):
                 if completed_samples > num_samples:
```
ahtsan: should it be `completed_samples >= num_samples`? Otherwise, there will always be one extra more iteration.

krzentner: Oh. Yeah, that's a bug. It's not a bug in this PR though, so I'll fix it in another PR.

PR #1156, opened 2020-01-26
Fix an off-by-one error in LocalSampler

krzentner:
Before, LocalSampler always gathered at least one more sample than
requested. This generally isn't a problem, but made sampling slightly
less predictable with LocalSampler.


krzentner:
@Mergifyio rebase

ryanjulian:
this won't merge automatically without another approving review (also it needs to pass the CI)

PR #1157, opened 2020-01-27
Add TaskOnehotWrapper

krzentner:
A multi-task wrapper which is efficient to use with multi-task aware
samplers. Instead of wrapping multiple environments in a single wrapper,
each `TaskOnehotWrapper` wraps a single environment. The wrapper also
provides a class function for constructing wrappers around a list of
environments.

### DIFF
--- a/src/garage/envs/task_onehot_wrapper.py
+++ b/src/garage/envs/task_onehot_wrapper.py
@@ -1,0 +1,168 @@
+"""Wrapper for appending one-hot task encodings to individual task envs.
+
+See `~TaskOnehotWrapper.wrap_env_list` for the main way of using this module.
+
+"""
+import akro
+import gym
+import numpy as np
+
+from garage.envs import EnvSpec
+
+
+class TaskOnehotWrapper(gym.Wrapper):
+    """Append a one-hot task representation to an environment.
+
+    See TaskOnehotWrapper.wrap_env_list for the recommended way of creating
+    this class.
+
+    Args:
+        env (gym.Env): The environment to wrap.
+        task_index (int): The index of this task among the tasks.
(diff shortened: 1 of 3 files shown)


src/garage/envs/task_onehot_wrapper.py
```diff
@@ -0,0 +1,166 @@
...
+            numpy.ndarray: active task one-hot representation + observation
+
+        """
+        return self._obs_with_one_hot(self.env.reset(**kwargs))
+
+    def step(self, action):
+        """gym.Env step for the active task env.
+
+        Args:
+            action (object): object to be passed in env.step(action)
```
ryanjulian: isn't this known to be a np.ndarray?

src/garage/envs/task_onehot_wrapper.py
```diff
@@ -0,0 +1,166 @@
+"""Wrapper for appending one-hot task encodings to individual task envs.
+
+See `~TaskOnehotWrapper.wrap_env_list` for the main way of using this module.
```
ryanjulian: `MultiTaskOneHotEnv`?

krzentner: Well, each env will only have a single task. I'm trying to keep the name saliently different from `MultiEnvWrapper`. Probably a better question is if I should drop `Wrapper` from the name and move it to `garage.envs.wrappers`.

src/garage/envs/task_onehot_wrapper.py
krzentner: It's almost always a `np.ndarray`. Theoretically, it can just be an integer for discrete action space environments.

PR #1159, opened 2020-01-28
Fix computation of SuccessRate in log_performance

krzentner:
I used `getattr` to get the `'success'` field to compute `SuccessRate` in `log_performance`.

However, `env_info` is a dictionary, so I should have used `get()` instead, so this results in `SuccessRate` always being 0.

If you were running a benchmark on success rate before that used this code, you need to restart / re-run it with this change.

### DIFF
--- a/src/garage/_functions.py
+++ b/src/garage/_functions.py
@@ -26,10 +26,8 @@ def log_performance(itr, batch, discount, prefix='Evaluation'):
         returns.append(discount_cumsum(trajectory.rewards, discount))
         undiscounted_returns.append(sum(trajectory.rewards))
         completion.append(float(trajectory.terminals.any()))
-        success.append(
-            float(
-                getattr(trajectory.env_infos, 'success',
-                        np.asarray([])).any()))
+        if 'success' in trajectory.env_infos:
+            success.append(trajectory.env_infos['success'].any())
 
     average_discounted_return = np.mean([rtn[0] for rtn in returns])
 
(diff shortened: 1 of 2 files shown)


src/garage/_functions.py
```diff
@@ -43,7 +41,6 @@ def log_performance(itr, batch, discount, prefix='Evaluation'):
         tabular.record('MaxReturn', np.max(undiscounted_returns))
         tabular.record('MinReturn', np.min(undiscounted_returns))
         tabular.record('CompletionRate', np.mean(completion))
-        if np.mean(success) > 0:
-            tabular.record('SuccessRate', np.mean(success))
+        tabular.record('SuccessRate', np.mean(success))
```
naeioi: This will log `SuccessRate` even when there is no `success` field in `env_infos`. I suggest using a flag to control logging it.

avnishn: The way that KR has written the array `success` , so that it is empty whenever there is no field called "success" in env_infos. I think this is fine because therefore, when env_infos is logged, then the success rate will be 0. 

Also consider this, what if environment only includes success in the env_infos dictionary after a certain number of timesteps taken. On our end, we wouldn't be able to log this if `success rate` wasn't a metric, because dowel isn't set up to add keys to the logger after the first training iteration.

krzentner: I think Keren is right that this will cause spam as-is. I looked through all the Meta-World environments, and they all seem to consistently output this key. Since we only need one environment to output the key in the first batch, I think this change is fine. However, I'm wondering if we should make log warnings (such as new keys) into errors in the runner in the future.

PR #1160, opened 2020-02-02
Make wrap_experiment archive the launcher repo

krzentner:
Also, save metadata.json and fix the wrapping logic so that click and
clize work.

### DIFF
--- a/src/garage/experiment/experiment.py
+++ b/src/garage/experiment/experiment.py
@@ -12,15 +12,18 @@ import pathlib
 import pickle
 import re
 import subprocess
+import warnings
 
 import cloudpickle
 import dowel
 from dowel import logger
 import dateutil.tz
+import __main__ as main
 
 exp_count = 0
 now = datetime.datetime.now(dateutil.tz.tzlocal())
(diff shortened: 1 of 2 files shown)


krzentner:
Right now, this might be useful to @naeioi and @yonghyuc (or anyone else using `wrap_experiment`), since it automatically creates `metadata.json`.

krzentner:
> Does this archive only tracked files? What if someone has a dirty repo?

It archives tracked and untracked files, but not ignored files (those matched by `.gitignore`).

ryanjulian:
Perfect -- and it finds the nearest git repo in my parent directory tree?

krzentner:
> Perfect -- and it finds the nearest git repo in my parent directory tree?

Yeah. Basically, it uses `__main__.__file__` to find the launcher script, then `git rev-parse --show-toplevel` from the same directory as that script to find the "nearest" git repo of the launcher, and archives all of the non-ignored files.

krzentner:
Unfortunately, the `git archive` command cannot archive files that haven't been added to the repo yet, so it uses `git ls-files` and `tar` manually. I'll add a comment clarifying that and some tests.

PR #1164, opened 2020-02-12
Wrap experiment fix sig

krzentner:
This PR is on top of the `wrap_experiment` archive repo change, so it's a bit hard to read.

Basically the goal with this change is to make the external signature of experiment templates actually match the arguments they take. This makes integration with tools that read those signatures (like clize or ipython tab completion) work much better.

### DIFF
--- a/src/garage/experiment/experiment.py
+++ b/src/garage/experiment/experiment.py
@@ -5,6 +5,7 @@ import datetime
 import enum
 import functools
 import gc
+import inspect
 import json
 import os
 import os.path as osp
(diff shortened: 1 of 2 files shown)


avnishn:
Can you explain your original blurb in a different way @krzentner? I'm not sure if I quite understand why the new feature, and it would be helpful to have an example :)

avnishn:
@mergifyio rebase

src/garage/experiment/experiment.py
```diff
@@ -234,6 +235,43 @@ def _make_sequential_log_dir(log_dir):
...
+    """
+    func_sig = inspect.signature(function)
+    new_params = []
+    is_first_param = True
+    for param in func_sig.parameters.values():
+        if is_first_param:
+            # Don't output it to the experiment params, since it will contain
+            # the context.
+            if param.name != 'ctxt':
+                raise ValueError('Experiment functions should have a first '
```
naeioi: Also need to check if `function` has an empty parameter list.

src/garage/experiment/experiment.py
```diff
@@ -234,6 +235,43 @@ def _make_sequential_log_dir(log_dir):
...
+                                 'parameter named ctxt')
+            is_first_param = False
+        else:
+            new_params.append(
+                inspect.Parameter(name=param.name,
+                                  kind=inspect.Parameter.KEYWORD_ONLY,
+                                  default=param.default,
+                                  annotation=param.annotation))
+    return inspect.Signature(new_params,
```
naeioi: What about this
```suggestion
    return inspect.Signature(func_sig.parameters[1:],
```
plus name check on the first parameter? This will saves the extra `new_params` list.

naeioi:
@Mergifyio rebase

src/garage/experiment/experiment.py
krzentner: They also need to be made keyword only, so this isn't sufficient.

src/garage/experiment/experiment.py
krzentner: Oh! Good catch.

ryanjulian:
@Mergifyio rebase

ryanjulian:
@krzentner is this ready to be merged? does it need a force-push?

krzentner:
Yeah, the test coverage was only 84%, because of some line not counting for some reason (even though it must be run).

PR #1165, opened 2020-02-13
Fix PointEnv

CatherineSue:
PointEnv didn't update its point location, which makes the environment
not step and always return constant reward.

### DIFF
--- a/src/garage/envs/point_env.py
+++ b/src/garage/envs/point_env.py
@@ -27,12 +27,10 @@ class PointEnv(gym.Env):
             goal=np.array((1., 1.), dtype=np.float32),
             done_bonus=0.,
             never_done=False,
-            action_scale=1.,
     ):
         self._goal = np.array(goal, dtype=np.float32)
         self._done_bonus = done_bonus
         self._never_done = never_done
-        self._action_scale = action_scale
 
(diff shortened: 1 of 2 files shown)


src/garage/envs/point_env.py
```diff
@@ -27,10 +27,12 @@ def __init__(
             goal=np.array((1., 1.), dtype=np.float32),
             done_bonus=0.,
             never_done=False,
+            action_scale=1.,
```
ryanjulian: please don't scale actions inside environments. it's up to the algorithm to do this.

ryanjulian:
@ahtsan feel free to merge this when you're happy with it

src/garage/envs/point_env.py
```diff
@@ -79,6 +79,7 @@ def step(self, action):
         a = action.copy()  # NOTE: we MUST copy the action before modifying it
         a = np.clip(a, self.action_space.low, self.action_space.high)
 
+        self._point = np.clip(self._point + a, -5, 5)
```
ryanjulian, line 82: perhaps define `ARENA_SIZE = 5` at the top of the file to make for easier maintenance.

naeioi:
Add some tests. Please take a second look.

PR #1166, opened 2020-02-13
Add MultitaskLinearFeatureBaseline

CatherineSue:


src/garage/np/baselines/multi_task_linear_feature_baseline.py
```diff
@@ -0,0 +1,17 @@
+"""Linear Feature Baseline for Multitasks."""
+import numpy as np
+
+from garage.np.baselines import LinearFeatureBaseline
+
+
+class MultiTaskLinearFeatureBaseline(LinearFeatureBaseline):
```
ryanjulian: it seems like this could be shortened by modifying LinearFeatureBaseline to take a list of feature keys from paths.

krzentner: That sounds pretty complex, given that it's not exactly obvious what procession to do for each key. I think the current design is fine, since it's fairly terse and very flexible.

naeioi: This class is specific to TE but its name does not reflect this - `path['tasks']` and `path['latents']` are available in only TE. My idea is to remove `_feature()` from baselines, and let algorithms decide what to feed to a baseline

ryanjulian: thats is cool with me as long as you don't get stuck refactoring for a week :)

naeioi:
I will add MultitaskLinearFeatureBaseline inside TE, but not as a piece that is available to other parts of garage

PR #1167, opened 2020-02-13
Add sampler features for embeddings

CatherineSue:
* Add a tuple for trajs batch returned from embedding sampler
* Add a task_embedding_worker to control rollouts of embedding
* Change local sampler and local_tf_runner API to let user pass a
  customized worker through experiment launcher.

### DIFF
--- a/src/garage/envs/multi_env_wrapper.py
+++ b/src/garage/envs/multi_env_wrapper.py
@@ -70,7 +70,7 @@ class MultiEnvWrapper(gym.Wrapper):
     def __init__(self,
                  envs,
                  sample_strategy=uniform_random_strategy,
-                 mode='vanilla'):
+                 mode='add-onehot'):
         assert mode in ['vanilla', 'add-onehot', 'del-onehot']
 
         self._sample_strategy = sample_strategy
@@ -91,6 +91,8 @@ class MultiEnvWrapper(gym.Wrapper):
             if env.action_space.shape != self.env.action_space.shape:
                 raise ValueError('Action space of all envs should be same.')
             self._task_envs.append(env)
+
+        self.env.spec.observation_space = self.observation_space
 
     @property
     def num_tasks(self):
@@ -207,6 +209,7 @@ class MultiEnvWrapper(gym.Wrapper):
         elif self._mode == 'del-onehot':
             obs = obs[:-self._num_tasks]
         info['task_id'] = self._active_task_index
+        info['task_name'] = self.active_task_name
         return obs, reward, done, info
 
     def close(self):
--- a/src/garage/envs/multi_env_wrapper.py
+++ b/src/garage/envs/multi_env_wrapper.py
@@ -45,29 +45,39 @@ def uniform_random_strategy(num_tasks, _):
 
 
 class MultiEnvWrapper(gym.Wrapper):
-    """A wrapper class to handle multiple gym environments.
+    """A wrapper class to handle multiple environments.
+
+    This wrapper adds an integer 'task_id' to env_info every timestep.
 
     Args:
         envs (list(gym.Env)):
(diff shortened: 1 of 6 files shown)


src/garage/_dtypes.py
```diff
@@ -492,3 +492,225 @@ def __new__(cls, env_spec, observation, action, reward, next_observation,
         return super().__new__(TimeStep, env_spec, observation, action, reward,
                                next_observation, terminal, env_info,
                                agent_info)
+
+
+class TaskEmbeddingTrajectoryBatch(
```
ryanjulian: this is definitely the wrong way to do this. task embeddings should just stuff/unpack this stuff from env_infos

krzentner: Yeah, I agree. It's preferable not to create a bunch of new `TrajectoryBatch` "subtypes."

src/garage/sampler/task_embedding_worker.py
```diff
@@ -0,0 +1,204 @@
...
+import gym
+import numpy as np
+
+from garage import TaskEmbeddingTrajectoryBatch
+from garage.experiment import deterministic
+from garage.sampler.worker import Worker
+from garage.sampler.env_update import EnvUpdate
+
+
+class TaskEmbeddingWorker(Worker):
```
krzentner: It's probably best to inherit from `DefaultWorker` here, instead of copying the bodies of the `worker_init`, `update_agent`, and `update_env` methods.

src/garage/_dtypes.py
krzentner: An alternative compromise would be to add an optional field to `TrajectoryBatch` called `worker_infos` or something, to avoid potentially conflicting with existing `env_infos`.

naeioi:
This PR is ready for review. Note that this is built upon and waiting on https://github.com/rlworkgroup/garage/pull/1169, so I set the rebase target to `mt_policy`.

src/garage/sampler/__init__.py
```diff
@@ -15,21 +15,12 @@
...
-    'OffPolicyVectorizedSampler',
-    'OnPolicyVectorizedSampler',
-    'WorkerFactory',
-    'Worker',
-    'DefaultWorker',
+    'BatchSampler', 'Sampler', 'ISSampler', 'singleton_pool', 'LocalSampler',
+    'RaySampler', 'SamplerWorker', 'MultiprocessingSampler',
+    'ParallelVecEnvExecutor', 'VecEnvExecutor', 'OffPolicyVectorizedSampler',
+    'OnPolicyVectorizedSampler', 'WorkerFactory', 'Worker', 'DefaultWorker',
+    'TaskEmbeddingWorker'
```
ryanjulian, line 25: please implement functionality-specific workers within the smallest package context which uses them. in this case, it looks like you should implement this worker inside `garage/tf/algos/te_ppo.py` or wherever you are putting the task embeddings implementation.

ryanjulian:
@naeioi is this ready to merge?

naeioi:
@avnishn take a look at MultiEnvWrapper. I make some changes so that it can handle with both MT* and Task Embedding.

src/garage/envs/multi_env_wrapper.py
```diff
@@ -45,29 +45,39 @@ def uniform_random_strategy(num_tasks, _):
...
+              observation, and it excludes that.
+              Use case: metaworld environments with Task Embedding.
+
     """
 
     def __init__(self,
                  envs,
                  sample_strategy=uniform_random_strategy,
-                 metaworld_mt=False):
+                 mode='vanilla'):
```
krzentner, line 73: Can we change this to `'add-onehot'` by default?

PR #1168, opened 2020-02-13
Add TensorFlow embedding models

CatherineSue:
* Add Embedding APIs and GaussianMLPEmbedding
* Add sample_sym to DiagonalGaussian to enable get_action_sym in the
  future embeddings or policies.


src/garage/tf/embeddings/base.py
```diff
@@ -0,0 +1,249 @@
...
+        """State info specifcation.
+
+        Returns:
+            List[str]: keys and shapes for the information related to the
+                policy's state when taking an action.
+
+        """
+        return list()
+
+    def terminate(self):
```
ryanjulian: i don't think this API is necessary

src/garage/tf/embeddings/base.py
```diff
@@ -0,0 +1,249 @@
...
+    def recurrent(self):
+        """Whether the embedding uses recurrent network or not.
+
+        Returns:
+            bool: Indicating if the policy is recurrent.
+
+        """
+        return False
+
+    def log_diagnostics(self):
```
ryanjulian: i think you can remove this

src/garage/tf/embeddings/base.py
```diff
@@ -0,0 +1,249 @@
...
+    @property
+    def embedding_spec(self):
+        return self._embedding_spec
+
+    @property
+    def recurrent(self):
+        """Whether the embedding uses recurrent network or not.
+
+        Returns:
+            bool: Indicating if the policy is recurrent.
```
ryanjulian: this is an embedding, not a policy

src/garage/tf/embeddings/base.py
```diff
@@ -0,0 +1,249 @@
...
+            (np.ndarray): Latents sampled from the policy.
+
+        """
+
+    def reset(self):
+        """Reset the embedding."""
+        pass
+
+    @property
+    def vectorized(self):
```
ryanjulian: you can remove this -- all embeddings are assumed to be vectorized

naeioi:
@ryanjulian @ahtsan Please take a second look. I added a bunch of tests.

src/garage/tf/embeddings/__init__.py
```diff
@@ -0,0 +1,8 @@
+"""Embeddings."""
+from garage.tf.embeddings.base import Embedding, EmbeddingSpec
```
ryanjulian, line 2: please don't use a `base` package -- instead put the classes in their own files

ryanjulian: also -- please follow the style for this module introduced in `garage.torch.embeddings`

src/garage/tf/embeddings/base.py
```diff
@@ -0,0 +1,275 @@
+"""Base class for embedding networks in TensorFlow."""
+import abc
+
+import tensorflow as tf
+
+from garage.misc.tensor_utils import flatten_tensors, unflatten_tensors
+
+
+class EmbeddingSpec:
```
ryanjulian, line 9: i think we already have one of these in _dtypes.py

src/garage/tf/embeddings/base.py
```diff
@@ -0,0 +1,275 @@
...
+    """
+
+    # pylint: disable=too-few-public-methods
+
+    def __init__(self, input_space, latent_space):
+        self.input_space = input_space
+        self.latent_space = latent_space
+
+
+class Embedding(abc.ABC):
```
ryanjulian, line 25: can this be a garage-wide interface, shared with `garage.torch`?

ryanjulian, line 25: i think that this should probably be called Encoder

A Decoder maps from a low-dimensional space to a high-dim one.

An embedding might be implemented using an encoder-decoder pair, just an encoder, or just a decoder.

src/garage/tf/embeddings/utils.py
```diff
@@ -0,0 +1,23 @@
+"""Utilities for embedding."""
```
ryanjulian, line 1: this should probably be an akro features instead? does akro have this?

src/garage/tf/embeddings/base.py
ahtsan: I was thinking about the same thing -- put all the common functions inside `garage.BaseModel`? Ultimately all primitives will be a BaseModel / inherits BaseModel.

ahtsan: But that's probably hard to be done in this PR

src/garage/tf/embeddings/utils.py
naeioi: I removed this. I will add this back in a PR when it is actually needed.

src/garage/__init__.py
```diff
@@ -1,5 +1,5 @@
 """Garage Base."""
-from garage._dtypes import InOutSpec
+from garage._dtypes import ModuleInOutSpec
```
ryanjulian, line 2: why the name change?

src/garage/tf/models/base.py
```diff
@@ -441,3 +445,216 @@ def __setstate__(self, state):
         """
         self.__dict__.update(state)
         self._networks = {}
+
+
+class Module(abc.ABC):
```
ryanjulian, line 450: thank you for doing this -- gets us one step closer to moving to `tf.Module`

src/garage/__init__.py
ryanjulian: i would prefer if you kept the old name -- nothing says this type can only be used by Modules

ryanjulian:
Don't work too hard on the API right now -- focus on getting it merged (no need to remove your work if it passes coverage+CI, though).

This looks like an excellent improvement in TF, but there's a lot of work already planned there (e.g. moving garage-wide interfaces outside of `garage.np`, trying to replace `tf.Model` with `tf.Module`, unifying the primitive APIs like Policy, etc.). No need to get ahead of yourself!

src/garage/tf/models/base.py
ahtsan: What's the most distinct difference between a model and a module?

naeioi: @ahtsan Module is built on top of Model. Policy, Regressor, Baseline are all Modules.

ryanjulian: As I state in my comment below, it's clear there will be significant refactoring for all of these classes in the near future, so I'd like to avoid a lengthy design discussion now.

@naeioi 's PR make a significant improvement (collecting a lot of repetitive code) towards unifying the TF primitive APIs, so I'm happy to merge it right now, even if it's not necessarily globally-optimal.

If you want a better and more careful design for the TF primitive tree, the best thing you can do is get TE merged, so we can cut the release, so we can start working on the API unification.

See https://github.com/rlworkgroup/garage/milestone/14 for what I plan on having us work on as soon as the release is cut. Notice any themes?

naeioi: Cool. Once I have the CI ready I will merge this.

ahtsan: @ryanjulian Sounds good.
@naeioi Great work!

naeioi:
@Mergifyio rebase

PR #1170, opened 2020-02-13
Fix observation_space error in MultiEnvWrapper

CatherineSue:
MultiEnvWrapper didn't reset its env.spec.observation_space in reset(),
this caused an error of imcompatible shape in sampler. Fixed it.

### DIFF
--- a/src/garage/envs/multi_env_wrapper.py
+++ b/src/garage/envs/multi_env_wrapper.py
@@ -157,6 +157,7 @@ class MultiEnvWrapper(gym.Wrapper):
         self.env = self._task_envs[self._active_task_index]
         obs = self.env.reset(**kwargs)
         oh_obs = self._obs_with_one_hot(obs)
+        self.env.spec.observation_space = self.observation_space
         return oh_obs
 
     def step(self, action):


src/garage/envs/multi_env_wrapper.py
```diff
@@ -157,6 +157,7 @@ def reset(self, **kwargs):
         self.env = self._task_envs[self._active_task_index]
         obs = self.env.reset(**kwargs)
         oh_obs = self._obs_with_one_hot(obs)
+        self.env.spec.observation_space = self.observation_space
```
ryanjulian: this should be unchanging because all environments have the same observation and action spaces. did you find an example where this is not the case?

ryanjulian: it looks like you should instead makes the checks on L66 assert that everything is totally equal, rather than just having the same shape.

naeioi:
I am closing this PR because MultiEnvWrapper is not compatible with Task Embedding. MultiEnvWrapper uses an augmented observation space (obs + one-hot task id), which is implicit to the algorithm, but TE requires an explitcit task id to train the policy

PR #1171, opened 2020-02-13
Add task embedding algorithms

CatherineSue:
* Add npo, trpo, and ppo task embedding algorithms
* Add an experiment launcher using ppo task embedding and point
  environment
* Add util functions for inference calculation in task embedding

### DIFF
--- a/examples/tf/te_ppo_metaworld_ml1_push.py
+++ b/examples/tf/te_ppo_metaworld_ml1_push.py
@@ -1,13 +1,11 @@
 #!/usr/bin/env python3
 """This is an example to train Task Embedding PPO with PointEnv."""
 # pylint: disable=no-value-for-parameter
-import akro
 import click
 from metaworld.benchmarks import ML1
-import numpy as np
 import tensorflow as tf
 
-from garage import InOutSpec, wrap_experiment
+from garage import wrap_experiment
 from garage.envs import GarageEnv, normalize
 from garage.envs.multi_env_wrapper import MultiEnvWrapper
 from garage.experiment.deterministic import set_seed
@@ -41,7 +39,7 @@ def te_ppo_ml1_push(ctxt, seed, n_epochs, batch_size_per_task):
     envs = [GarageEnv(normalize(ML1.get_train_tasks('push-v1')))]
     env = MultiEnvWrapper(envs, mode='del-onehot')
 
-    latent_length = 4
+    latent_length = 2
     inference_window = 6
     batch_size = batch_size_per_task
     policy_ent_coeff = 2e-2
@@ -57,31 +55,8 @@ def te_ppo_ml1_push(ctxt, seed, n_epochs, batch_size_per_task):
 
     with LocalTFRunner(snapshot_config=ctxt) as runner:
 
-        latent_lb = np.zeros(latent_length, )
-        latent_ub = np.ones(latent_length, )
-        latent_space = akro.Box(latent_lb, latent_ub)
-
-        # trajectory space is (TRAJ_ENC_WINDOW, act_obs) where act_obs is a
-        # stacked vector of flattened actions and observations
-        obs_lb, obs_ub = env.observation_space.bounds
-        obs_lb_flat = env.observation_space.flatten(obs_lb)
-        obs_ub_flat = env.observation_space.flatten(obs_ub)
-        traj_lb = np.stack([obs_lb_flat] * inference_window)
-        traj_ub = np.stack([obs_ub_flat] * inference_window)
-        traj_space = akro.Box(traj_lb, traj_ub)
-
-        task_embed_spec = InOutSpec(env.task_space, latent_space)
-        traj_embed_spec = InOutSpec(traj_space, latent_space)
-
-        inference = GaussianMLPEncoder(
-            name='inference',
-            embedding_spec=traj_embed_spec,
-            hidden_sizes=[20, 10],
-            std_share_network=True,
-            init_std=2.0,
-            output_nonlinearity=tf.nn.tanh,
-            min_std=embedding_min_std,
-        )
+        task_embed_spec = TEPPO.get_encoder_spec(env.task_space,
+                                                 latent_dim=latent_length)
 
         task_encoder = GaussianMLPEncoder(
             name='embedding',
@@ -90,6 +65,21 @@ def te_ppo_ml1_push(ctxt, seed, n_epochs, batch_size_per_task):
             std_share_network=True,
             init_std=embedding_init_std,
             max_std=embedding_max_std,
+            output_nonlinearity=tf.nn.tanh,
+            min_std=embedding_min_std,
+        )
+
+        traj_embed_spec = TEPPO.get_infer_spec(
+            env.spec,
+            latent_dim=latent_length,
+            inference_window_size=inference_window)
+
+        inference = GaussianMLPEncoder(
+            name='inference',
+            embedding_spec=traj_embed_spec,
+            hidden_sizes=[20, 10],
+            std_share_network=True,
+            init_std=2.0,
             output_nonlinearity=tf.nn.tanh,
             min_std=embedding_min_std,
         )
--- a/examples/tf/te_ppo_metaworld_ml1_push.py
+++ b/examples/tf/te_ppo_metaworld_ml1_push.py
@@ -46,8 +46,8 @@ def te_ppo_ml1_push(ctxt, seed, n_epochs, batch_size_per_task):
     encoder_ent_coeff = 2.e-4
     inference_ce_coeff = 5e-2
     max_path_length = 100
-    embedding_init_std = 0.01
-    embedding_max_std = 0.02
+    embedding_init_std = 0.1
+    embedding_max_std = 0.2
     embedding_min_std = 1e-6
     policy_init_std = 1.0
     policy_max_std = None
--- a/examples/tf/te_ppo_metaworld_ml1_push.py
+++ b/examples/tf/te_ppo_metaworld_ml1_push.py
@@ -8,13 +8,13 @@ import tensorflow as tf
 from garage import wrap_experiment
 from garage.envs import GarageEnv, normalize
 from garage.envs.multi_env_wrapper import MultiEnvWrapper
+from garage.experiment import LocalTFRunner
 from garage.experiment.deterministic import set_seed
 from garage.np.baselines import LinearMultiFeatureBaseline
 from garage.sampler import LocalSampler
 from garage.tf.algos import TEPPO
 from garage.tf.algos.te import TaskEmbeddingWorker
 from garage.tf.embeddings import GaussianMLPEncoder
-from garage.tf.experiment import LocalTFRunner
 from garage.tf.policies import GaussianMLPTaskEmbeddingPolicy
 
 
--- a/examples/tf/te_ppo_metaworld_ml1_push.py
+++ b/examples/tf/te_ppo_metaworld_ml1_push.py
@@ -43,7 +43,7 @@ def te_ppo_ml1_push(ctxt, seed, n_epochs, batch_size_per_task):
     inference_window = 6
     batch_size = batch_size_per_task
     policy_ent_coeff = 2e-2
-    encoder_ent_coeff = 2.e-4
+    encoder_ent_coeff = 2e-4
     inference_ce_coeff = 5e-2
     max_path_length = 100
     embedding_init_std = 0.1
@@ -61,7 +61,7 @@ def te_ppo_ml1_push(ctxt, seed, n_epochs, batch_size_per_task):
         task_encoder = GaussianMLPEncoder(
             name='embedding',
             embedding_spec=task_embed_spec,
-            hidden_sizes=[20, 20],
+            hidden_sizes=(20, 20),
             std_share_network=True,
             init_std=embedding_init_std,
             max_std=embedding_max_std,
@@ -77,7 +77,7 @@ def te_ppo_ml1_push(ctxt, seed, n_epochs, batch_size_per_task):
         inference = GaussianMLPEncoder(
             name='inference',
             embedding_spec=traj_embed_spec,
-            hidden_sizes=[20, 10],
+            hidden_sizes=(20, 10),
             std_share_network=True,
             init_std=2.0,
             output_nonlinearity=tf.nn.tanh,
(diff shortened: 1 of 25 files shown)


src/garage/tf/algos/__init__.py
```diff
@@ -25,4 +28,7 @@
     'TNPG',
     'TRPO',
     'VPG',
+    'NPOTaskEmbedding',
```
ryanjulian: this should be TEPPO TETRPO, TENPO, etc.following the same pattern as RL2PPO, MAMLTRPO, etc.

examples/tf/ppo_multi_env_embed.py
```diff
@@ -0,0 +1,189 @@
...
+                     worker_class=TaskEmbeddingWorker)
+        runner.train(n_epochs=600, batch_size=v.batch_size, plot=False)
+
+
+config = dict(
+    tasks=TASKS,
+    latent_length=1,
+    inference_window=2,
+    batch_size=1024 * len(TASKS),
+    policy_ent_coeff=2e-2,  # 2e-2
```
ryanjulian: please remove these random comments with hyperparams

src/garage/tf/algos/npo_task_embedding.py
```diff
@@ -0,0 +1,1349 @@
...
+from garage.tf.misc.tensor_utils import pad_tensor_n
+from garage.tf.misc.tensor_utils import pad_tensor_dict
+from garage.tf.misc.tensor_utils import positive_advs
+from garage.tf.misc.tensor_utils import stack_tensor_dict_list
+from garage.tf.misc.tensor_utils import stack_tensor_list
+from garage.tf.optimizers import LbfgsOptimizer
+from garage.tf.policies import StochasticMultitaskPolicy
+
+
+class NPOTaskEmbedding(BatchPolopt):
```
krzentner, line 32: Having a complete copy of NPO is a serious issue. These duplicates never stay in sync, and result in a big maintenance burden.

Please see RL2's changes for alternate ways of implementing this.

examples/tf/ppo_multi_env_embed.py
```diff
@@ -0,0 +1,189 @@
...
+            'never_done': False,
+            'action_scale': 0.1,
+            'done_bonus': 0.0,
+        }
+    }
+    for i, g in enumerate(goals)
+}
+
+
+def run_task(snapshot_config, v, *_):
```
ryanjulian, line 50: please use the new wrap_experiment API

ryanjulian:
Don't forget to update the README!

naeioi:
This PR is ready for review. 

Benchmark result on PointEnv (4 tasks). https://tensorboard.dev/experiment/DMCG87X9T2KJfWxcQLFV0Q

Benchmark result on MT10. https://tensorboard.dev/experiment/eW2g24URT16pLQn3X1aSrQ

examples/tf/te_ppo_metaworld_ml1_push.py
```diff
@@ -0,0 +1,144 @@
...
+    with LocalTFRunner(snapshot_config=ctxt) as runner:
+
+        latent_lb = np.zeros(latent_length, )
+        latent_ub = np.ones(latent_length, )
+        latent_space = akro.Box(latent_lb, latent_ub)
+
+        # trajectory space is (TRAJ_ENC_WINDOW, act_obs) where act_obs is a
+        # stacked vector of flattened actions and observations
+        obs_lb, obs_ub = env.observation_space.bounds
+        obs_lb_flat = env.observation_space.flatten(obs_lb)
```
ryanjulian, line 67: can you add a helper classmethod to TEPPO to make this less painful? See PEARL and RL2 for examples.

src/garage/misc/tensor_utils.py
```diff
@@ -318,3 +318,73 @@ def slice_nested_dict(dict_or_array, start, stop):
...
+        NotImplementedError: If step_size is not 1.
+        ValueError: If window size is larger than the input tensor.
+
+    """
+    if window > t.shape[0]:
+        raise ValueError('`window` must be <= `t.shape[0]`')
+    if window == t.shape[0]:
+        return np.stack([t] * window)
+
+    if step_size != 1:
```
ryanjulian, line 367: hmm probably just remove this param lol

src/garage/tf/embeddings/encoder.py
```diff
@@ -8,6 +8,52 @@
 class Encoder(BaseEncoder, Module):
     """Base class for encoders in TensorFlow."""
 
+    def forward_n(self, input_values):
```
ryanjulian, line 11: forward is always vectorized, right?

ryanjulian:
Can you put your TensorBoard.dev links here?

src/garage/tf/embeddings/encoder.py
naeioi: The non-vectorized version is used in `policy.get_latent()`, and `get_latent()` is used in sampler where a latent is sampled once per trajectory.

I make it this way also to keep consistent with the policy interface - it has `get_action()` as well as `get_actions()`.

naeioi:
@ryanjulian The tensorboard links for PointEnv and MT10 is posted in an earlier comment.

ryanjulian:
This looks like a working implementation, I have just one remaining concern. 
The tensorboard.dev links you shared show the encoder at 0 entropy:
Screen Shot 2020-06-09 at 4 42 02 PM

Were the encoders initialized with 0 entropy? Is embedding_ent_coeff non-zero? Or is this a logging error?

src/garage/tf/embeddings/encoder.py
ryanjulian: by your own analogy, shouldn't the name of these be get_latent() and get_latents()? Forward has a specific meaning in torch. In garage.tf.Module, the approximate equivalent is `build`

naeioi: `get_latent()` is in the policy while `forward()` is in the encoder. The first one makes a call to the second one.  `build()` assumes symbolic input. But in this case, both take actual values as input.

Maybe it is more appropriate to rename `forward()` to `get_latent()`? (The same name as it is in policies)

ryanjulian: I think get_latent is better

naeioi:
@ryanjulian re: entropy

This is not a logging error. I checked and found that low encoder entropy is due to the fact that I set the std of the encoder to a very low value (min: 1e-6, max: 0.02, init: 0.01). With this setting, the encoder is almost deterministic. I can get a learning TE only in this way.

naeioi:
I fixed NaNs using an early NPO clipping. Previously, clipping happens after `tf.exp()`, but a large update might overflow `tf.exp()`.

https://github.com/rlworkgroup/garage/blob/8210413a0e9011c463658e283e80b12458082c01/src/garage/tf/algos/te_npo.py#L692-L694

This is more of a band-aid solution than a real fix, because TE does not have the NaNs issue without tfp. The non-tfp code is archived in `embedding_algo_bak` branch.

Benchmark results: [PointEnv](https://tensorboard.dev/experiment/4t4jH1mLSqiG9YMRYQqiGg/#scalars), [MT10](https://tensorboard.dev/experiment/uN1ioY6eTESLCQ4WYD3E1g/#scalars), [MT50](https://tensorboard.dev/experiment/wok7vpZnTnysQMjHYWkWRg/#scalars). As of the time I write comment, these experiment are still running.

These results are not as good as previously posted, because they use a larger encoder variance. (0.1 magnitudes instead of 0.01 magnitudes).

examples/tf/te_ppo_metaworld_ml1_push.py
```diff
@@ -0,0 +1,132 @@
...
+
+        traj_embed_spec = TEPPO.get_infer_spec(
+            env.spec,
+            latent_dim=latent_length,
+            inference_window_size=inference_window)
+
+        inference = GaussianMLPEncoder(
+            name='inference',
+            embedding_spec=traj_embed_spec,
+            hidden_sizes=[20, 10],
```
ryanjulian, line 80: to set a good example, can you always use tuples for hidden-sizes?

examples/tf/te_ppo_metaworld_ml1_push.py
```diff
@@ -0,0 +1,132 @@
...
+        batch_size_per_task (int): Batch size of samples for each task.
+
+    """
+    set_seed(seed)
+    envs = [GarageEnv(normalize(ML1.get_train_tasks('push-v1')))]
+    env = MultiEnvWrapper(envs, mode='del-onehot')
+
+    latent_length = 2
+    inference_window = 6
+    batch_size = batch_size_per_task
```
ryanjulian, line 44: this seems inconsistent because you have 50 tasks but are only using 1 `batch_size_per_task` worth of samples?

examples/tf/te_ppo_metaworld_ml1_push.py
```diff
@@ -0,0 +1,132 @@
...
+    policy_ent_coeff = 2e-2
+    encoder_ent_coeff = 2.e-4
```
ryanjulian, line 46: no need for the trailing . -- xe-y statements are always floats

src/garage/tf/algos/te_npo.py
```diff
@@ -0,0 +1,1191 @@
...
+
+        logger.log('Fitting baseline...')
+        self._baseline.fit(paths)
+
+        self._old_policy.model.parameters = self.policy.model.parameters
+        self._old_policy.encoder.model.parameters = (
+            self.policy.encoder.model.parameters)
+        self._old_inference.model.parameters = self._inference.model.parameters
+
+    def paths_to_tensors(self, itr, paths):
```
ryanjulian, line 259: is there no way you can use `garage.np.paths_to_tensors` and then just add some extra info?

src/garage/tf/algos/te_npo.py
```diff
@@ -0,0 +1,1191 @@
...
+                in garage.tf.optimizers.
+            optimizer_args (dict): The arguments of the optimizer.
+
+        Returns:
+            obj: Policy optimizer. Should be one of the optimizers
+                in garage.tf.optimizers.
+            dict: The arguments of the optimizer.
+
+        """
+        # pylint: disable=no-self-use
```
ryanjulian, line 460: why not just make it a function in this module?

src/garage/tf/algos/te_npo.py
```diff
@@ -0,0 +1,1191 @@
...
+        if optimizer is None:
+            optimizer = LbfgsOptimizer
+        if optimizer_args is None:
+            optimizer_args = dict()
+        return optimizer, optimizer_args
+
+    def _build_inference_optimizer(self, optimizer, optimizer_args):
```
ryanjulian, line 467: same question -- can't this be a function?

also, this function seems superfluous -- can't you just use build_optimizer twice?

src/garage/tf/algos/te_npo.py
```diff
@@ -0,0 +1,1191 @@
...
+        """Build up optimizer for inference.
+
+        Args:
+            optimizer (obj): Policy optimizer. Should be one of the optimizers
+                in garage.tf.optimizers.
+            optimizer_args (dict): The arguments of the optimizer.
+
+        Returns:
+            obj: Policy optimizer. Should be one of the optimizers
```
ryanjulian, line 476: obj is not a type :)

src/garage/tf/algos/te_npo.py
```diff
@@ -0,0 +1,1191 @@
...
+
+        Args:
+            i (namedtuple): Collection of variables to compute encoder KL
+                divergence.
+
+        Returns:
+            tf.Tensor: Encoder KL divergence.
+
+        """
+        del i
```
ryanjulian, line 815: just remove the i argument?

src/garage/tf/algos/te_npo.py
```diff
@@ -0,0 +1,1191 @@
...
+
+        Returns:
+            list(np.ndarray): Flatten inference optimization input values.
+
+        """
+        infer_state_info_list = [
+            samples_data['trajectory_infos'][k]
+            for k in self._inference.state_info_keys
+        ]
+        # pylint: disable=unexpected-keyword-arg
```
ryanjulian, line 921: can you explain this?

src/garage/tf/algos/te_npo.py
```diff
@@ -0,0 +1,1191 @@
...
+
+    def visualize_distribution(self):
+        """Visualize encoder distribution."""
+        num_tasks = self.policy.task_space.flat_dim
+        all_tasks = np.eye(num_tasks, num_tasks)
+        _, latent_infos = self.policy.encoder.get_latents(all_tasks)
+
+        for task in range(num_tasks):
+            for i in range(self.policy.latent_space.flat_dim):
+                # pylint: disable=protected-access
```
ryanjulian, line 1033: i don't see any protected access?

src/garage/tf/algos/te_npo.py
naeioi: optimizers don't yet have a base class.

src/garage/tf/algos/te_npo.py
naeioi: Pylint passes on my local machine, but it failed on CI. I removed this line and see how it goes.

examples/tf/te_ppo_metaworld_ml1_push.py
naeioi: I think the 50 internal goals are obscure to the algorithm. In case of ML1, the algorithm should just see one task, so it uses only 1 `batch_size_per_task`.

src/garage/tf/algos/te_npo.py
ryanjulian: Can you use the type Object?

src/garage/tf/embeddings/gaussian_mlp_encoder.py
```diff
@@ -118,11 +137,15 @@ def __init__(self,
     def build(self, embedding_input, name=None):
         """Build encoder.
 
+        After build, self.distribution is a Gaussian distribution conitioned
```
ahtsan, line 140: Would it still be possible to build this encoder with a different input, and retrieve the resulting outputs, e.g. distribution? It seems like enforcing build to assign an attribute is not what we want. Instead we may build in constructor and assign these attributes, and build should return the outputs.

naeioi: I agree. `build()` better has no side-effect.

PR #1172, opened 2020-02-14
Add option for NPO to fit baseline first

ahtsan:
Refactor NPO to support either fit baseline first or fit baseline after. Also, expose functions for calculating advantages and returns for NPO variants.

### DIFF
--- a/examples/tf/ppo_pendulum.py
+++ b/examples/tf/ppo_pendulum.py
@@ -71,7 +71,7 @@ def tf_ppo_pendulum(ctxt=None, seed=1):
             entropy_method='max',
             policy_ent_coeff=0.02,
             center_adv=False,
-            fit_baseline_first=True,
+            fit_baseline_order='after',
         )
 
         runner.setup(algo, env)
--- a/examples/tf/ppo_pendulum.py
+++ b/examples/tf/ppo_pendulum.py
@@ -71,6 +71,7 @@ def tf_ppo_pendulum(ctxt=None, seed=1):
             entropy_method='max',
             policy_ent_coeff=0.02,
             center_adv=False,
+            fit_baseline_first=True,
         )
 
         runner.setup(algo, env)
--- a/src/garage/misc/tensor_utils.py
+++ b/src/garage/misc/tensor_utils.py
@@ -23,7 +23,7 @@ def discount_cumsum(x, discount):
                                 axis=0)[::-1]
 
 
-def explained_variance_1d(ypred, y):
+def explained_variance_1d(ypred, y, valids):
     """Explained variation for 1D inputs.
 
     It is the proportion of the variance in one variable that is explained or
@@ -32,11 +32,14 @@ def explained_variance_1d(ypred, y):
     Args:
         ypred (np.ndarray): Sample data from the first variable.
         y (np.ndarray): Sample data from the second variable.
+        valids (np.ndarray): Array indicating valid indices.
 
     Returns:
         float: The explained variance.
 
     """
+    ypred = ypred[valids.astype(np.bool)]
+    y = y[valids.astype(np.bool)]
     assert y.ndim == 1 and ypred.ndim == 1
     vary = np.var(y)
     if np.isclose(vary, 0):
--- a/src/garage/tf/algos/npo.py
+++ b/src/garage/tf/algos/npo.py
@@ -8,7 +8,6 @@ from garage.tf.algos.batch_polopt import BatchPolopt
 from garage.tf.misc.tensor_utils import center_advs
 from garage.tf.misc.tensor_utils import compile_function
 from garage.tf.misc.tensor_utils import compute_advantages
-from garage.tf.misc.tensor_utils import concat_tensor_list
 from garage.tf.misc.tensor_utils import discounted_returns
 from garage.tf.misc.tensor_utils import filter_valids
 from garage.tf.misc.tensor_utils import filter_valids_dict
@@ -68,8 +67,9 @@ class NPO(BatchPolopt):
         flatten_input (bool): Whether to flatten input along the observation
             dimension. If True, for example, an observation with shape (2, 4)
             will be flattened to 8.
-        fit_baseline_first (bool): Whether to fit baseline first or after. If
-            true, baseline will be fitted before making prediction.
+        fit_baseline_order (str): Whether to fit baseline before or after,
+            either 'before' or 'after'. If 'before', baseline will be fitted
+            before making prediction. Default: 'after'.
         name (str): The name of the algorithm.
 
     Note:
@@ -107,7 +107,7 @@ class NPO(BatchPolopt):
                  stop_entropy_gradient=False,
                  entropy_method='no_entropy',
                  flatten_input=True,
-                 fit_baseline_first=False,
+                 fit_baseline_order='after',
                  name='NPO'):
         self._name = name
         self._name_scope = tf.name_scope(self._name)
@@ -115,7 +115,11 @@ class NPO(BatchPolopt):
         self._use_neg_logli_entropy = use_neg_logli_entropy
         self._stop_entropy_gradient = stop_entropy_gradient
         self._pg_loss = pg_loss
-        self._will_fit_baseline_first = fit_baseline_first
+        self._fit_baseline_order = fit_baseline_order
+
+        if fit_baseline_order not in ['before', 'after']:
+            raise ValueError('Invalid fit_baseline_order')
+
         if optimizer is None:
             if optimizer_args is None:
                 optimizer_args = dict()
@@ -174,8 +178,8 @@ class NPO(BatchPolopt):
                 See process_samples() for details.
 
         """
-        if self._will_fit_baseline_first:
-            self._fit_baseline_first(samples_data)
+        if self._fit_baseline_order == 'before':
+            self._fit_baseline_before(samples_data)
 
         policy_opt_input_values = self._policy_opt_input_values(samples_data)
         # Train policy network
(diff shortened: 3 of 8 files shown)


src/garage/tf/algos/npo.py
```diff
@@ -105,13 +107,15 @@ def __init__(self,
...
                  flatten_input=True,
+                 fit_baseline_first=False,
                  name='NPO'):
         self._name = name
         self._name_scope = tf.name_scope(self._name)
         self._use_softplus_entropy = use_softplus_entropy
         self._use_neg_logli_entropy = use_neg_logli_entropy
         self._stop_entropy_gradient = stop_entropy_gradient
         self._pg_loss = pg_loss
+        self._will_fit_baseline_first = fit_baseline_first
```
ryanjulian: please keep member variables the same as constructor args, so they're easy to track

examples/tf/ppo_pendulum.py
```diff
@@ -71,6 +71,7 @@ def tf_ppo_pendulum(ctxt=None, seed=1):
             entropy_method='max',
             policy_ent_coeff=0.02,
             center_adv=False,
+            fit_baseline_first=True,
```
ryanjulian: this is hard to understand if i don't know the internal of this algorithm.

maybe you can make it a string arg named 'fit_baselin' instead, with options `'before'` and `'after'`

later we might make an in-graph baseline (that is, the baseline and policy are in the same model), in which case we could introduce the option `'joint'` or `'during'` or something.

src/garage/tf/algos/npo.py
```diff
@@ -607,6 +639,39 @@ def _fit_baseline(self, samples_data):
         else:
             self.baseline.fit(paths)
 
+    def _get_advantages(self, baseline_var, reward_var, name):
```
ryanjulian: what is the purpose of these two helper functions? they just call single other functions.

src/garage/tf/algos/npo.py
```diff
@@ -562,7 +594,7 @@ def _build_entropy_term(self, i):
 
         return policy_entropy
 
-    def _fit_baseline(self, samples_data):
+    def _fit_baseline_after(self, samples_data):
```
ryanjulian: please find a way to structure this so that the baseline fitting routine is shared between the cases.

src/garage/tf/algos/npo.py
```diff
@@ -68,6 +68,8 @@ class NPO(BatchPolopt):
         flatten_input (bool): Whether to flatten input along the observation
             dimension. If True, for example, an observation with shape (2, 4)
             will be flattened to 8.
+        fit_baseline_first (bool): Whether to fit baseline first or after. If
```
ryanjulian: don't forget to update the documentation of child classes

src/garage/tf/algos/npo.py
ahtsan: The "calculate returns in separate rollouts" version of NPO will need to override this method.

src/garage/tf/algos/npo.py
ahtsan: Do you mean to make this NOT optional in all the super classes?

ahtsan: Do you mean doing `fit_baseline_order=fit_baseline_order` in constructor of child classes of `NPO`, like `PPO`?

examples/tf/ppo_pendulum.py
```diff
@@ -71,6 +71,7 @@ def tf_ppo_pendulum(ctxt=None, seed=1):
             entropy_method='max',
             policy_ent_coeff=0.02,
             center_adv=False,
+            fit_baseline_order='after',
```
ryanjulian: please call this `fit_baseline`

src/garage/misc/tensor_utils.py
```diff
@@ -32,11 +32,14 @@ def explained_variance_1d(ypred, y):
     Args:
         ypred (np.ndarray): Sample data from the first variable.
         y (np.ndarray): Sample data from the second variable.
+        valids (np.ndarray): Array indicating valid indices.
```
ryanjulian: can you provide us with some information on the expected shapes and/or types of this arrays?

examples/tf/ppo_pendulum.py
ahtsan: there's already a function called `_fit_baseline`

src/garage/tf/algos/npo.py
```diff
@@ -68,6 +67,9 @@ class NPO(BatchPolopt):
         flatten_input (bool): Whether to flatten input along the observation
             dimension. If True, for example, an observation with shape (2, 4)
             will be flattened to 8.
+        fit_baseline_order (str): Whether to fit baseline before or after,
```
ryanjulian: this needs much better documentation. it assumes the user knows all about the internals of NPO.

Please make the point of reference the optimization step, not the prediction. In the normal case, we fit the baseline after optimizing the policy. In the strange case, we fit the baseline after optimizing the policy.

src/garage/tf/algos/npo.py
```diff
@@ -68,6 +67,9 @@ class NPO(BatchPolopt):
...
+            either 'before' or 'after'. If 'before', baseline will be fitted
+            before making prediction. Default: 'after'.
```
ryanjulian: Please always use complete english phrases. i.e. "Defaults to `'after'`".

src/garage/tf/algos/npo.py
```diff
@@ -105,13 +107,19 @@ def __init__(self,
...
         self._name = name
         self._name_scope = tf.name_scope(self._name)
         self._use_softplus_entropy = use_softplus_entropy
         self._use_neg_logli_entropy = use_neg_logli_entropy
         self._stop_entropy_gradient = stop_entropy_gradient
         self._pg_loss = pg_loss
+        self._fit_baseline_order = fit_baseline_order
+
+        if fit_baseline_order not in ['before', 'after']:
+            raise ValueError('Invalid fit_baseline_order')
```
ryanjulian: please provide more information to the user. What value did she specify, and what are the valid choices?

src/garage/tf/algos/npo.py
```diff
@@ -192,6 +203,87 @@ def optimize_policy(self, itr, samples_data):
...
+        Args:
+            samples_data (dict): Processed sample data.
+                See process_samples() for details.
+
+        """
+        self._fit_baseline(samples_data)
+
+        # update baseline prediction for optimization
+        paths = samples_data['paths']
+        baselines = [self.baseline.predict(path) for path in paths]
```
ryanjulian: doesn't baseline have a predict_n function which does this vectorized?

ryanjulian: this is extremely strange and hard to follow. your method name says you are fitting the baseline, but then you also run the prediction step here. that's impossible to debug.

it doesn't seem like you even need different versions of `fit_baseline_before` and `fit_baseline_after` -- they both do the same thing. 

you should represent this in the logic of optimizer_policy instead. hiding it behind these methods is very confusing.

src/garage/tf/algos/npo.py
ryanjulian: this version of NPO calculates returns in separate rollouts....

calculating returns across rollouts is unique to RL2 -- whatever is overriding should be in there. it makes no sense to make a special version of NPO which will only be used by RL2.

src/garage/tf/algos/ppo.py
```diff
@@ -77,6 +81,7 @@ def __init__(self,
                  stop_entropy_gradient=False,
                  entropy_method='no_entropy',
                  flatten_input=True,
+                 fit_baseline_order='after',
```
ryanjulian: what inheritance hierarchy do you plan for RL2/RL2PPO/RL2TRPO?

it is dangerous for this to be a constructor option in PPO when there is no reason for anyone to use it except for in RL2.

ahtsan: I plan to have RL2PPO inherit RL2NPO, which inherits NPO by overriding `fit_baseline_before`. The conflicting part here is that if it isn't for RL2, NPO won't need the separation of `fit_baseline_before` and `fit_baseline_after`, so if you look at NPO alone separately, the refactor is weird and unnecessary.

How about we enforce PPO to have `fit_baseline_order='after'` while provides both options for RL2PPO/RL2TRPO?

examples/tf/ppo_pendulum.py
ryanjulian: then rename it. the public API is more important.

src/garage/tf/algos/npo.py
ahtsan: Do you think we should create a new class RL2NPO without inheriting NPO?

ryanjulian: if you are asking me if you should copy/paste, no.

if you can define a class RL2 which uses composition instead of inheritance for NPO, and then inherit RL2PPO and RL2TRPO from that, then yes.

take a look at Keren's MAML for what this looks like.

ahtsan: I can't use most of the code from NPO if I do composition. NPO builds the policy loss in a way that is not applicable to RL2. In order to implement "calculate returns with respect to each separate rollouts in a meta batch",  I think I can only reuse most of the code from NPO only if I inherit it and override some of the functions, especially the way it computes returns and advantages.

ahtsan: @krzentner What do you think?

krzentner: I think it's fine to have these hook point methods. Admittedly, it would be better to do something similar to MAML, but that's more difficult to do in this case while still allowing us to support variations like concatenated paths, since the related logic is spread throughout the algorithm.

src/garage/tf/algos/ppo.py
ryanjulian: see my comments above. it makes no sense to have methods `fit_baseline_before` and `fit_baseline_after` -- there is only one `fit_baseline` and what you want to control is when it's called.

I think your inheritance tree should be RL2NPO <-- NPO, RL2PPO <-- RL2NPO, RL2TRPO <-- RL2NPO. That way you can specify `fit_baseline='after'` for TRPO and PPO and allow it to be an option for RL2PPO and RL2TRPO.

ahtsan:
I will submit another PR so it can be merged easier.

PR #1173, opened 2020-02-16
Add clear method to PathBuffer

lywong92:
PEARL uses PathBuffer, and it needs a clear method when buffer is used to store context.

### DIFF
--- a/src/garage/replay_buffer/path_buffer.py
+++ b/src/garage/replay_buffer/path_buffer.py
@@ -178,7 +178,7 @@ class PathBuffer:
             seg_b (list): List of indices of the second segment.
 
         Returns:
-            Bool: True iff the input ranges overlap at at least one index.
+            bool: True iff the input ranges overlap at at least one index.
 
         """
         # Empty segments never overlap.
--- a/src/garage/replay_buffer/path_buffer.py
+++ b/src/garage/replay_buffer/path_buffer.py
@@ -40,8 +40,8 @@ class PathBuffer:
             path_array = path.get(key, None)
             if path_array is None:
                 raise ValueError('Key {} missing from path.'.format(key))
-            elif (len(path_array.shape) != 2
-                  or path_array.shape[1] != buf_arr.shape[1]):
+            if (len(path_array.shape) != 2
+                    or path_array.shape[1] != buf_arr.shape[1]):
                 raise ValueError('Array {} has wrong shape.'.format(key))
         path_len = self._get_path_length(path)
         first_seg, second_seg = self._next_path_segments(path_len)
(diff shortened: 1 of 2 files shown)


src/garage/replay_buffer/path_buffer.py
```diff
@@ -99,14 +119,43 @@ def _next_path_segments(self, n_indices):
...
 
     def _get_or_allocate_key(self, key, array):
+        """Get or allocate key in the buffer.
+
+        Args:
+            key (str): Key in buffer.
+            array (numpy.ndarray): Array corresponding to key.
+
+        Returns:
+            buf_arr: A numpy.ndarray corresponding to key in the buffer.
```
ryanjulian: this should be a type, not a variable name

src/garage/replay_buffer/path_buffer.py
```diff
@@ -9,8 +9,10 @@ class PathBuffer:
 
     This buffer only stores valid steps, and doesn't require paths to
     have a maximum length.
+
     Args:
-        capacity_in_steps (int): total memory allocated for the buffer
+        capacity_in_transitions (int): Total memory allocated for the buffer.
```
ryanjulian: This is a really long arg name. I suggest we stick with steps since it is also used here https://github.com/rlworkgroup/garage/blob/master/src/garage/_dtypes.py#L410

src/garage/replay_buffer/path_buffer.py
```diff
@@ -124,8 +173,12 @@ def _get_path_length(path):
     def _segments_overlap(seg_a, seg_b):
         """Compute if two segments overlap.
 
+        Args:
+            seg_a (list): List of indices of the first segment.
+            seg_b (list): List of indices of the second segment.
+
         Returns:
-            True iff the input ranges overlap at at least one index.
+            Bool: True iff the input ranges overlap at at least one index.
```
ryanjulian: i think it's bool (lowercase)

src/garage/replay_buffer/path_buffer.py
lywong92: What about `_transitions_stored` and `sample_transitions`? Are you suggesting to also change these to steps?

src/garage/replay_buffer/path_buffer.py
```diff
@@ -124,8 +173,12 @@ def _get_path_length(path):
     def _segments_overlap(seg_a, seg_b):
         """Compute if two segments overlap.
 
+        Args:
+            seg_a (list): List of indices of the first segment.
+            seg_b (list): List of indices of the second segment.
```
krzentner: The types of these arguments should be `range`

src/garage/replay_buffer/path_buffer.py
krzentner: So Linda's just making the docstring consistent with the code I wrote in this case.
This argument is long. Probably the whole `PathBuffer` should use the term step(s), instead of transition(s), but I think that's out of scope for this PR.

PR #1175, opened 2020-02-17
Env wrapper for RL2

ahtsan:


src/garage/envs/rl2_env.py
```diff
@@ -0,0 +1,98 @@
...
+
+        Returns:
+            gym.spaces.Box: Augmented observation space.
+
+        """
+        obs_flat_dim = np.prod(env.observation_space.shape)
+        action_flat_dim = np.prod(env.action_space.shape)
+        if self._max_obs_dim is not None:
+            obs_flat_dim = self._max_obs_dim
+        return gym.spaces.Box(low=-np.inf,
```
krzentner: Why not just call `akro.Box` here?

src/garage/envs/rl2_env.py
```diff
@@ -0,0 +1,98 @@
...
+    reward and terminal signal to form new observation.
+
+    Also, different tasks could have different observation dimension.
+    An example is in ML45 from MetaWorld (reference:
+    https://arxiv.org/pdf/1910.10897.pdf). This wrapper pads the
+    observation to the maximum observation dimension with zeros.
+
+    Args:
+        env (gym.Env): An env that will be wrapped.
+        max_obs_dim (int): Maximum observation dimension in the environments
```
krzentner: This is a needed feature to actually use ML45, but I think it probably belongs in a separate wrapper. We can address that in a later PR after submitting this.

src/garage/envs/rl2_env.py
ahtsan: Updated

src/garage/envs/__init__.py
```diff
@@ -8,9 +8,10 @@
 from garage.envs.half_cheetah_vel_env import HalfCheetahVelEnv
 from garage.envs.normalized_env import normalize
 from garage.envs.point_env import PointEnv
+from garage.envs.rl2_env import RL2Env
```
ryanjulian: can this be part of the `garage.tf.algos.rl2` module instead?

is there any way that RL2 can automatically wrap the environment with this wrapper?

ahtsan: Yes I will put this inside rl2 class and make RL2 automatically wrap the environment with this wrapper

ahtsan: Actually, the env never get passed to RL2 -- It only lives in the runner so RL2 can't wrap it.

ryanjulian: ah. this will have to wait for the refactor where the algorithm owns the sampler. 

can you please file an issue about this?

PR #1176, opened 2020-02-18
Tanh Gaussian policy for torch

avnishn:
This PR is for the Tanh Gaussian Policy thats used in (mt)SAC.

It is a working product, but can use some refactor. I will bring the refactors in a later PR.

### DIFF
--- a/src/garage/torch/modules/gaussian_mlp_module.py
+++ b/src/garage/torch/modules/gaussian_mlp_module.py
@@ -64,6 +64,9 @@ class GaussianMLPBaseModule(nn.Module):
                exponential transformation.
             - softplus: the std will be computed as log(1+exp(x)).
         layer_normalization (bool): Bool for using layer normalization or not.
+        normal_distribution_cls (torch.distribution): normal distribution class
+            to be constructed and returned by a call to forward. By default, is
+            `torch.distributions.Normal`.
 
     """
 
@@ -88,7 +91,8 @@ class GaussianMLPBaseModule(nn.Module):
                  std_output_nonlinearity=None,
                  std_output_w_init=nn.init.xavier_uniform_,
                  std_parameterization='exp',
-                 layer_normalization=False):
+                 layer_normalization=False,
+                 normal_distribution_cls=Normal):
         super().__init__()
 
         self._input_dim = input_dim
@@ -111,6 +115,7 @@ class GaussianMLPBaseModule(nn.Module):
         self._output_w_init = output_w_init
         self._output_b_init = output_b_init
         self._layer_normalization = layer_normalization
+        self._norm_dist_class = normal_distribution_cls
 
         if self._std_parameterization not in ('exp', 'softplus'):
             raise NotImplementedError
@@ -120,12 +125,30 @@ class GaussianMLPBaseModule(nn.Module):
             self._init_std = torch.nn.Parameter(init_std_param)
         else:
             self._init_std = init_std_param
+            self.register_buffer('init_std', self._init_std)
 
         self._min_std_param = self._max_std_param = None
         if min_std is not None:
             self._min_std_param = torch.Tensor([min_std]).log()
+            self.register_buffer('min_std_param', self._min_std_param)
         if max_std is not None:
             self._max_std_param = torch.Tensor([max_std]).log()
+            self.register_buffer('max_std_param', self._max_std_param)
+
+    def to(self, *args, **kwargs):
+        """Move the module to the specified device.
+
+        Args:
+            *args: args to pytorch to function.
+            **kwargs: keyword args to pytorch to function.
+
+        """
+        super().to(*args, **kwargs)
+        buffers = dict(self.named_buffers())
+        if not isinstance(self._init_std, torch.nn.Parameter):
+            self._init_std = buffers['init_std']
+        self._min_std_param = buffers['min_std_param']
+        self._max_std_param = buffers['max_std_param']
 
     @abc.abstractmethod
     def _get_mean_and_log_std(self, *inputs):
@@ -152,8 +175,10 @@ class GaussianMLPBaseModule(nn.Module):
             std = log_std_uncentered.exp()
         else:
             std = log_std_uncentered.exp().exp().add(1.).log()
-
-        dist = Independent(Normal(mean, std), 1)
+        dist = self._norm_dist_class(mean, std)
+        # Makes it so that a sample from the distribution is treated as a
+        # single sample and not dist.batch_shape samples.
+        dist = Independent(dist, len(dist.batch_shape))
 
         return dist
 
@@ -211,6 +236,9 @@ class GaussianMLPModule(GaussianMLPBaseModule):
                exponential transformation
             - softplus: the std will be computed as log(1+exp(x))
         layer_normalization (bool): Bool for using layer normalization or not.
+        normal_distribution_cls (torch.distribution): normal distribution class
+            to be constructed and returned by a call to forward. By default, is
+            `torch.distributions.Normal`.
 
     """
 
@@ -229,7 +257,8 @@ class GaussianMLPModule(GaussianMLPBaseModule):
                  min_std=1e-6,
                  max_std=None,
                  std_parameterization='exp',
-                 layer_normalization=False):
+                 layer_normalization=False,
+                 normal_distribution_cls=Normal):
         super(GaussianMLPModule,
               self).__init__(input_dim=input_dim,
                              output_dim=output_dim,
@@ -245,7 +274,8 @@ class GaussianMLPModule(GaussianMLPBaseModule):
                              min_std=min_std,
                              max_std=max_std,
                              std_parameterization=std_parameterization,
-                             layer_normalization=layer_normalization)
+                             layer_normalization=layer_normalization,
+                             normal_distribution_cls=normal_distribution_cls)
 
         self._mean_module = MLPModule(
             input_dim=self._input_dim,
@@ -266,9 +296,8 @@ class GaussianMLPModule(GaussianMLPBaseModule):
             *inputs: Input to the module.
 
         Returns:
-            tuple:
-                * mean (torch.Tensor): The mean of Gaussian distribution.
-                * std (torch.Tensor): The variance of Gaussian distribution.
+            torch.Tensor: The mean of Gaussian distribution.
+            torch.Tensor: The variance of Gaussian distribution.
 
         """
         assert len(inputs) == 1
@@ -334,6 +363,9 @@ class GaussianMLPIndependentStdModule(GaussianMLPBaseModule):
                exponential transformation
             - softplus: the std will be computed as log(1+exp(x))
         layer_normalization (bool): Bool for using layer normalization or not.
+        normal_distribution_cls (torch.distribution): normal distribution class
+            to be constructed and returned by a call to forward. By default, is
+            `torch.distributions.Normal`.
 
     """
 
@@ -358,7 +390,8 @@ class GaussianMLPIndependentStdModule(GaussianMLPBaseModule):
                  std_output_nonlinearity=None,
                  std_output_w_init=nn.init.xavier_uniform_,
                  std_parameterization='exp',
-                 layer_normalization=False):
+                 layer_normalization=False,
+                 normal_distribution_cls=Normal):
         super(GaussianMLPIndependentStdModule,
               self).__init__(input_dim=input_dim,
                              output_dim=output_dim,
@@ -380,7 +413,8 @@ class GaussianMLPIndependentStdModule(GaussianMLPBaseModule):
                              std_output_nonlinearity=std_output_nonlinearity,
                              std_output_w_init=std_output_w_init,
                              std_parameterization=std_parameterization,
-                             layer_normalization=layer_normalization)
+                             layer_normalization=layer_normalization,
+                             normal_distribution_cls=normal_distribution_cls)
 
         self._mean_module = MLPModule(
             input_dim=self._input_dim,
@@ -425,9 +459,8 @@ class GaussianMLPIndependentStdModule(GaussianMLPBaseModule):
             *inputs: Input to the module.
 
         Returns:
-            tuple:
-                * mean (torch.Tensor): The mean of Gaussian distribution.
-                * std (torch.Tensor): The variance of Gaussian distribution.
+            torch.Tensor: The mean of Gaussian distribution.
+            torch.Tensor: The variance of Gaussian distribution.
 
         """
         return self._mean_module(*inputs), self._log_std_module(*inputs)
@@ -473,6 +506,9 @@ class GaussianMLPTwoHeadedModule(GaussianMLPBaseModule):
                exponential transformation
             - softplus: the std will be computed as log(1+exp(x))
         layer_normalization (bool): Bool for using layer normalization or not.
+        normal_distribution_cls (torch.distribution): normal distribution class
+            to be constructed and returned by a call to forward. By default, is
+            `torch.distributions.Normal`.
 
     """
 
@@ -491,7 +527,8 @@ class GaussianMLPTwoHeadedModule(GaussianMLPBaseModule):
                  min_std=1e-6,
                  max_std=None,
                  std_parameterization='exp',
-                 layer_normalization=False):
+                 layer_normalization=False,
+                 normal_distribution_cls=Normal):
         super(GaussianMLPTwoHeadedModule,
               self).__init__(input_dim=input_dim,
                              output_dim=output_dim,
@@ -507,7 +544,8 @@ class GaussianMLPTwoHeadedModule(GaussianMLPBaseModule):
                              min_std=min_std,
                              max_std=max_std,
                              std_parameterization=std_parameterization,
-                             layer_normalization=layer_normalization)
+                             layer_normalization=layer_normalization,
+                             normal_distribution_cls=normal_distribution_cls)
 
         self._shared_mean_log_std_network = MultiHeadedMLPModule(
             n_heads=2,
@@ -532,9 +570,8 @@ class GaussianMLPTwoHeadedModule(GaussianMLPBaseModule):
             *inputs: Input to the module.
 
         Returns:
-            tuple:
-                * mean (torch.Tensor): The mean of Gaussian distribution.
-                * std (torch.Tensor): The variance of Gaussian distribution.
+            torch.Tensor: The mean of Gaussian distribution.
+            torch.Tensor: The variance of Gaussian distribution.
 
         """
         return self._shared_mean_log_std_network(*inputs)
--- a/src/garage/torch/policies/tanh_gaussian_mlp_policy.py
+++ b/src/garage/torch/policies/tanh_gaussian_mlp_policy.py
@@ -1,17 +1,19 @@
-"""GaussianMLPPolicy."""
+"""TanhGaussianMLPPolicy."""
+import numpy as np
 import torch
 from torch import nn
 
-from garage.torch.modules import GaussianMLPModule
+from garage.torch.distributions import TanhNormal
+from garage.torch.modules import GaussianMLPTwoHeadedModule
 from garage.torch.policies import Policy
 import garage.torch.utils as tu
 
 
-class GaussianMLPPolicy(Policy, GaussianMLPModule):
-    """MLP whose outputs are fed into a Normal distribution..
+class TanhGaussianMLPPolicy(Policy, GaussianMLPTwoHeadedModule):
+    """Multiheaded MLP whose outputs are fed into a TanhNormal distribution.
 
     A policy that contains a MLP to make prediction based on a gaussian
-    distribution.
+    distribution with a tanh transformation.
 
     Args:
         env_spec (garage.envs.env_spec.EnvSpec): Environment specification.
@@ -36,57 +38,57 @@ class GaussianMLPPolicy(Policy, GaussianMLPModule):
         output_b_init (callable): Initializer function for the bias
             of output dense layer(s). The function should return a
             torch.Tensor.
-        learn_std (bool): Is std trainable.
         init_std (float): Initial value for std.
             (plain value - not log or exponentiated).
-        min_std (float): Minimum value for std.
-        max_std (float): Maximum value for std.
+        min_std (float): If not None, the std is at least the value of min_std,
+            to avoid numerical issues (plain value - not log or exponentiated).
+        max_std (float): If not None, the std is at most the value of max_std,
+            to avoid numerical issues (plain value - not log or exponentiated).
         std_parameterization (str): How the std should be parametrized. There
             are two options:
             - exp: the logarithm of the std will be stored, and applied a
                exponential transformation
             - softplus: the std will be computed as log(1+exp(x))
         layer_normalization (bool): Bool for using layer normalization or not.
-        name (str): Name of policy.
 
     """
 
     def __init__(self,
                  env_spec,
                  hidden_sizes=(32, 32),
-                 hidden_nonlinearity=torch.tanh,
+                 hidden_nonlinearity=nn.ReLU,
                  hidden_w_init=nn.init.xavier_uniform_,
                  hidden_b_init=nn.init.zeros_,
                  output_nonlinearity=None,
                  output_w_init=nn.init.xavier_uniform_,
                  output_b_init=nn.init.zeros_,
-                 learn_std=True,
                  init_std=1.0,
-                 min_std=1e-6,
-                 max_std=None,
+                 min_std=np.exp(-20.),
+                 max_std=np.exp(2.),
                  std_parameterization='exp',
-                 layer_normalization=False,
-                 name='GaussianMLPPolicy'):
+                 layer_normalization=False):
+
         self._obs_dim = env_spec.observation_space.flat_dim
         self._action_dim = env_spec.action_space.flat_dim
 
-        Policy.__init__(self, env_spec, name)
-        GaussianMLPModule.__init__(self,
-                                   input_dim=self._obs_dim,
-                                   output_dim=self._action_dim,
-                                   hidden_sizes=hidden_sizes,
-                                   hidden_nonlinearity=hidden_nonlinearity,
-                                   hidden_w_init=hidden_w_init,
-                                   hidden_b_init=hidden_b_init,
-                                   output_nonlinearity=output_nonlinearity,
-                                   output_w_init=output_w_init,
-                                   output_b_init=output_b_init,
-                                   learn_std=learn_std,
-                                   init_std=init_std,
-                                   min_std=min_std,
-                                   max_std=max_std,
-                                   std_parameterization=std_parameterization,
-                                   layer_normalization=layer_normalization)
+        Policy.__init__(self, env_spec, name='TanhGaussianPolicy')
+        GaussianMLPTwoHeadedModule.__init__(
+            self,
+            input_dim=self._obs_dim,
+            output_dim=self._action_dim,
+            hidden_sizes=hidden_sizes,
+            hidden_nonlinearity=hidden_nonlinearity,
+            hidden_w_init=hidden_w_init,
+            hidden_b_init=hidden_b_init,
+            output_nonlinearity=output_nonlinearity,
+            output_w_init=output_w_init,
+            output_b_init=output_b_init,
+            init_std=init_std,
+            min_std=min_std,
+            max_std=max_std,
+            std_parameterization=std_parameterization,
+            layer_normalization=layer_normalization,
+            normal_distribution_cls=TanhNormal)
 
     def get_action(self, observation):
         r"""Get a single action given an observation.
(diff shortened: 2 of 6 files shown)


src/garage/torch/modules/gaussian_mlp_module.py
```diff
@@ -120,12 +120,30 @@ def __init__(self,
...
+        """Move the module to the specified device.
+
+        Args:
+            *args: args to pytorch to function.
+            **kwargs: keyword args to pytorch to function.
+
+        """
+        super().to(*args, **kwargs)
+        buffers = dict(self.named_buffers())
+        if not isinstance(self._init_std, torch.nn.Parameter):
```
krzentner: It would probably be a good idea to add a comment mentioning that this is necessary since `.to`, when called on Tensors / Parameters directly, doesn't operate in place.

src/garage/torch/policies/tanh_gaussian_mlp_policy.py
```diff
@@ -0,0 +1,184 @@
...
+                        the distribution
+
+        """
+        with torch.no_grad():
+            if not isinstance(observation, torch.Tensor):
+                observation = tu.from_numpy(observation)
+            observation = torch.Tensor(observation).unsqueeze(0)
+            dist = self.forward(observation)
+            ret_mean = dist.mean.squeeze(0).cpu().numpy()
+            ret_log_std = (dist.variance**.5).log().squeeze(0).cpu().numpy()
```
krzentner: Generally, if you want a square root, you should just call `.sqrt()`.

I thought it would be less efficient to use `**0.5`, but I was surprised to find that pytorch actually will [turn `pow(n, 0.5)` back into sqrt on its own](https://github.com/pytorch/pytorch/blob/7a3ed36309f48cb833f1690991c7b0f59da6ce11/aten/src/ATen/native/cuda/PowKernel.cu#L122). It's still more readable to use `.sqrt()` though.

src/garage/torch/policies/tanh_gaussian_mlp_policy.py
```diff
@@ -0,0 +1,184 @@
...
+                    * list[float]: Standard deviation of logarithmic values of
+                        the distribution
+
+        """
+        with torch.no_grad():
+            if not isinstance(observations, torch.Tensor):
+                observations = tu.from_numpy(observations)
+            dist = self.forward(torch.Tensor(observations))
+            ret_mean = dist.mean.cpu().numpy()
+            ret_log_std = (dist.variance**.5).log().cpu().numpy()
```
krzentner: Use `.sqrt()` here please.

tests/garage/torch/policies/test_tanh_gaussian_mlp_policy.py
```diff
@@ -0,0 +1,131 @@
...
+                                       hidden_sizes=(1, ),
+                                       init_std=init_std,
+                                       hidden_nonlinearity=None,
+                                       std_parameterization='exp',
+                                       hidden_w_init=nn.init.ones_,
+                                       output_w_init=nn.init.ones_)
+        assert policy.vectorized
+
+    def test_to(self):
+        """Test Tanh Gaussian Policy can be moved to cpu."""
```
krzentner: Is this really the best we can do? Isn't it already on the cpu? Would this test actually catch a bug in `.to()`?

It would probably be better to make a test that starts `if torch.cuda.is_available():`, and then makes sure that `.get_action` still works after moving the policy to `cuda:0`.

src/garage/torch/policies/tanh_gaussian_mlp_policy.py
```diff
@@ -0,0 +1,184 @@
...
+                * dict:
+                    * list[float]: Mean of the distribution
+                    * list[float]: Standard deviation of logarithmic values of
+                        the distribution
+
+        """
+        with torch.no_grad():
+            if not isinstance(observation, torch.Tensor):
+                observation = tu.from_numpy(observation)
+            observation = torch.Tensor(observation).unsqueeze(0)
```
krzentner: Why call `torch.Tensor` again?

src/garage/torch/policies/tanh_gaussian_mlp_policy.py
```diff
@@ -0,0 +1,184 @@
...
+                * dict:
+                    * list[float]: Mean of the distribution
+                    * list[float]: Standard deviation of logarithmic values of
+                        the distribution
+
+        """
+        with torch.no_grad():
+            if not isinstance(observations, torch.Tensor):
+                observations = tu.from_numpy(observations)
+            dist = self.forward(torch.Tensor(observations))
```
krzentner: Why call `torch.Tensor` again?

avnishn:
https://codecov.io/gh/rlworkgroup/garage/commit/a541f5cb997ce17c3e972579be34d79f54a5713b

This is the code cov report. Correct me if I'm wrong but I don't think anything is wrong here @krzentner

src/garage/torch/modules/tanh_gaussian_mlp_module.py
```diff
@@ -0,0 +1,242 @@
...
+            layer_normalization=self._layer_normalization)
+
+    def _get_mean_and_log_std(self, *inputs):
+        """Get mean and std of Tanh Gaussian distribution given inputs.
+
+        Args:
+            *inputs: Input to the module.
+
+        Returns:
+            tuple:
```
ryanjulian: you don't need to do `tuple:` -- the docstring style specifically just says to list one per line

return values don't have names.

basically, like this
```
torch.Tensor: Mean of Tanh Gaussian distribution.
torch.Tensor: Variance of Tanh Gaussian distribution.
```

is that the variance, std, or log std?

src/garage/torch/modules/tanh_gaussian_mlp_module.py
```diff
@@ -0,0 +1,242 @@
...
+                * mean (torch.Tensor): The mean of Tanh Gaussian distribution.
+                * std (torch.Tensor): The variance of Tanh Gaussian
+                    distribution.
+
+        """
+        return self._shared_mean_log_std_network(inputs[0])
```
ryanjulian: why does this take *inputs and then only use inputs[0]? what's at inputs[0]?

src/garage/torch/modules/tanh_gaussian_mlp_module.py
```diff
@@ -0,0 +1,242 @@
...
+                         std_hidden_sizes=std_hidden_sizes,
+                         std_hidden_nonlinearity=std_hidden_nonlinearity,
+                         std_hidden_w_init=std_hidden_w_init,
+                         std_hidden_b_init=std_hidden_b_init,
+                         std_output_nonlinearity=torch.tanh,
+                         std_output_w_init=std_output_w_init,
+                         std_parameterization=std_parameterization,
+                         layer_normalization=layer_normalization)
+
+    def forward(self, *inputs):
```
ryanjulian: why would you take *inputs and then only use inputs[0]? just name the single input you're going to take.

src/garage/torch/modules/tanh_gaussian_mlp_module.py
```diff
@@ -0,0 +1,242 @@
...
+        """Forward method.
+
+        Args:
+            *inputs(torch.Tensor): Input to the module.
```
ryanjulian: we need to know something about the expected shape of the input

src/garage/torch/modules/tanh_gaussian_mlp_module.py
```diff
@@ -0,0 +1,242 @@
...
+
+        Returns:
+            torch.Tensor: Module output.
```
ryanjulian: what can you tell me about the expected shape of the output?

src/garage/torch/modules/tanh_gaussian_mlp_module.py
```diff
@@ -0,0 +1,242 @@
...
+    @abc.abstractmethod
+    def _get_mean_and_log_std(self, *inputs):
+        pass
+
+
+class TanhGaussianMLPTwoHeadedModule(TanhGaussianMLPBaseModule):
+    """TanhGaussianMLPModule which has an output head for the mean and stdev.
+
+    Args:
+        input_dim (int): Input dimension of the model.
```
ryanjulian: can't i infer this from the input?

src/garage/torch/modules/tanh_gaussian_mlp_module.py
```diff
@@ -0,0 +1,242 @@
...
+            are two options:
+            - exp: the logarithm of the std will be stored, and applied a
+               exponential transformation
+            - softplus: the std will be computed as log(1+exp(x))
+        layer_normalization (bool): Bool for using layer normalization or not.
+
+    """
+
+    def __init__(self,
+                 input_dim,
```
ryanjulian: can't you infer this from the input?

src/garage/torch/policies/tanh_gaussian_mlp_policy.py
```diff
@@ -0,0 +1,184 @@
...
+            layer_normalization=layer_normalization)
+
+    def get_action(self, observation):
+        """Get a single action given an observation.
+
+        Args:
+            observation (np.ndarray): Observation from the environment.
+
+        Returns:
+            tuple:
```
ryanjulian: again returning a tuple is equivalent to multiple return values. please just use that style when documenting.

src/garage/torch/policies/tanh_gaussian_mlp_policy.py
```diff
@@ -0,0 +1,184 @@
...
+                * np.ndarray: Predicted action.
```
ryanjulian: what can you tell me about the expected shape of these inputs and outputs?

src/garage/torch/policies/tanh_gaussian_mlp_policy.py
```diff
@@ -0,0 +1,184 @@
...
+                * dict:
+                    * list[float]: Mean of the distribution
```
ryanjulian: why not return a tuple? the number of heads is fixed.

src/garage/torch/policies/tanh_gaussian_mlp_policy.py
```diff
@@ -0,0 +1,184 @@
...
+                    dict(mean=ret_mean, log_std=ret_log_std))
+
+    def get_actions(self, observations):
+        """Get actions given observations.
+
+        Args:
+            observations (np.ndarray): Observations from the environment.
+
+        Returns:
+            tuple:
```
ryanjulian: use multiple return style

src/garage/torch/policies/tanh_gaussian_mlp_policy.py
```diff
@@ -0,0 +1,184 @@
...
+            ret_mean = dist.mean.squeeze(0).cpu().numpy()
+            ret_log_std = (dist.variance**.5).log().squeeze(0).cpu().numpy()
+            return (dist.rsample().squeeze(0).cpu().numpy(),
+                    dict(mean=ret_mean, log_std=ret_log_std))
+
+    def get_actions(self, observations):
+        """Get actions given observations.
+
+        Args:
+            observations (np.ndarray): Observations from the environment.
```
ryanjulian: please specify shapes

src/garage/torch/policies/tanh_gaussian_mlp_policy.py
```diff
@@ -0,0 +1,184 @@
...
+            dist = self.forward(torch.Tensor(observations))
+            ret_mean = dist.mean.cpu().numpy()
+            ret_log_std = (dist.variance**.5).log().cpu().numpy()
+            return (dist.rsample().cpu().numpy(),
+                    dict(mean=ret_mean, log_std=ret_log_std))
+
+    def log_likelihood(self, observation, action):
+        """Compute log likelihood given observations and action.
+
+        Args:
```
ryanjulian: please specify shapes for all tensor inputs/outputs

src/garage/torch/policies/tanh_gaussian_mlp_policy.py
```diff
@@ -0,0 +1,184 @@
...
+        Returns:
+            torch.Tensor: Calculated log likelihood value of the action given
+                observation
+
+        """
+        dist = self.forward(observation)
+        return dist.log_prob(action)
+
+    def entropy(self, observation):
+        """Get entropy given observations.
```
ryanjulian: please specify shapes for all tensor inputs/outputs

src/garage/torch/policies/tanh_gaussian_mlp_policy.py
```diff
@@ -0,0 +1,184 @@
...
+
+        """
+        dist = self.forward(observation)
+        return dist.entropy()
+
+    def reset(self, dones=None):
+        """Reset the environment.
+
+        Args:
+            dones (numpy.ndarray): Reset values
```
ryanjulian: please specify shapes.

also, what is `dones`?? isn't that something an environment produces?

src/garage/torch/policies/tanh_gaussian_mlp_policy.py
```diff
@@ -0,0 +1,184 @@
...
+
+        """
+
+    @property
+    def vectorized(self):
```
ryanjulian: i thought this was deprecated? don't we assume all policies are vectorized.

src/garage/torch/utils.py
```diff
@@ -76,3 +78,21 @@ def update(m, name, param):
                 update(named_modules[module_name], param_name, new_param)
         else:
             update(module, name, new_param)
+
+
+def from_numpy(array):
```
ryanjulian: is this really clearer than the torch equivalent?

src/garage/torch/modules/tanh_gaussian_mlp_module.py
```diff
@@ -0,0 +1,242 @@
+"""GaussianMLPModule."""
```
ryanjulian: documentation is incorrect. also, this should be a complete sentence or phase.

src/garage/torch/modules/tanh_gaussian_mlp_module.py
```diff
@@ -0,0 +1,242 @@
...
+import torch
+from torch import nn
+
+from garage.torch.distributions import TanhNormal
+from garage.torch.modules.gaussian_mlp_module import GaussianMLPBaseModule
+from garage.torch.modules.multi_headed_mlp_module import MultiHeadedMLPModule
+
+
+class TanhGaussianMLPBaseModule(GaussianMLPBaseModule):
+    """TanhGaussianMLPModel.
```
ryanjulian: please actually describe what this thing is and what it does.

src/garage/torch/modules/tanh_gaussian_mlp_module.py
```diff
@@ -0,0 +1,242 @@
...
+        dist = TanhNormal(mean, std)
+        return dist
+
+    @abc.abstractmethod
+    def _get_mean_and_log_std(self, *inputs):
+        pass
+
+
+class TanhGaussianMLPTwoHeadedModule(TanhGaussianMLPBaseModule):
+    """TanhGaussianMLPModule which has an output head for the mean and stdev.
```
ryanjulian: please describe what this thing is and why it exists

src/garage/torch/policies/tanh_gaussian_mlp_policy.py
```diff
@@ -0,0 +1,184 @@
...
+import torch
+from torch import nn
+
+from garage.torch.modules import TanhGaussianMLPTwoHeadedModule
+from garage.torch.policies import Policy
+import garage.torch.utils as tu
+
+
+class TanhGaussianMLPPolicy(Policy, TanhGaussianMLPTwoHeadedModule):
+    """TanhGaussianMLPPolicy.
```
ryanjulian: please don't abuse the one-liner rule like this. your summary line should be an actual summary.

ryanjulian:
what is RLas?

avnishn:
@ryanjulian RLas = reinforcement learning algorithms

src/garage/torch/utils.py
avnishn: In parallel I'm working on a pr for torch utils. It adds the necessary torch utility functions and their proper documentation. The reason that we've added functions like this is because there is no way of defining a global gpu mode in torch, so we have this class, and wrapper functions that automatically move tensors over to gpu if a user has specified `torch_utils.set_gpu(True)`

ryanjulian: does this need be in this PR, or could it appear in that PR or a separate one?

src/garage/torch/modules/tanh_gaussian_mlp_module.py
avnishn: Sorry, I'm not sure what you mean by this. This is the constructor for a pytorch nn.module that is the base for the `TanhGaussianPolicy`. Wouldn't one need to construct a policy/module first before passing it some input(such as an observation)

src/garage/torch/utils.py
ryanjulian: i.e. would removing it from this PR break something, or merely remove a feature?

avnishn: Sorry ignore what I said above, it didn't answer your question. You're right it is not clearer than the pytorch doc. In my upcoming torch utils pr, I've updated docstrings to use their torch equivalents.

tests/garage/torch/policies/test_tanh_gaussian_mlp_policy.py
avnishn: I initially was unsure of how to go about this because we don't have gpu's on travis. I'll go ahead and implement your idea.

src/garage/torch/modules/tanh_gaussian_mlp_module.py
ryanjulian: i was basically asking why this needs to be a member variable if it's only used in `forward` where it can be inferred from the shape of `input`, but it appears to be a result of the weird inheritance hierarchy of torch.modules more than anything.

src/garage/torch/modules/tanh_gaussian_mlp_module.py
```diff
@@ -0,0 +1,242 @@
...
+import numpy as np
+import torch
+from torch import nn
+
+from garage.torch.distributions import TanhNormal
+from garage.torch.modules.gaussian_mlp_module import GaussianMLPBaseModule
+from garage.torch.modules.multi_headed_mlp_module import MultiHeadedMLPModule
+
+
+class TanhGaussianMLPBaseModule(GaussianMLPBaseModule):
```
ryanjulian: why inherit when you can have-a GaussianMLPBaseModule?

src/garage/torch/policies/tanh_gaussian_mlp_policy.py
```diff
@@ -0,0 +1,184 @@
...
+import numpy as np
+import torch
+from torch import nn
+
+from garage.torch.modules import TanhGaussianMLPTwoHeadedModule
+from garage.torch.policies import Policy
+import garage.torch.utils as tu
+
+
+class TanhGaussianMLPPolicy(Policy, TanhGaussianMLPTwoHeadedModule):
```
ryanjulian: @krzentner is this the inheritance hierarchy you want?

src/garage/torch/utils.py
ryanjulian: sorry, i was not asking that. i was asking, "is making this helper function really clearer than just using its one-line implementation everywhere?"

src/garage/torch/policies/tanh_gaussian_mlp_policy.py
avnishn: This terminology is in the torch gaussian policy.

src/garage/torch/utils.py
avnishn: I see your point. I think that its correct to not make functions for one liners. Can I make this fix in a subsequent pr?

ryanjulian: file an issue and assign yourself. put it in the 2020.02 milestone.

src/garage/torch/policies/tanh_gaussian_mlp_policy.py
ryanjulian: that didn't answer my question. @krzentner is vectorized needed anymore?

src/garage/torch/utils.py
krzentner: Wait, this function automatically handles setting the right device. It's definitely helpful for decreasing the number of pain points when moving an algorithm to GPU.

I don't want to see hundreds of "one liners" using `garage.torch.utils._DEVICE` everywhere.

src/garage/torch/policies/gaussian_mlp_policy.py
```diff
@@ -88,58 +89,70 @@ def __init__(self,
                                    layer_normalization=layer_normalization)
 
     def get_action(self, observation):
-        """Get a single action given an observation.
+        r"""Get a single action given an observation.
```
krzentner: Why is the `r` needed?

src/garage/torch/policies/tanh_gaussian_mlp_policy.py
krzentner: Sadly, yes, vectorized is needed, since `garage.np.algos.BatchPolopt` checks it.

src/garage/torch/policies/tanh_gaussian_mlp_policy.py
krzentner: Oh yeah. This is perhaps the most confusing parameter of the vectorized policy interface. It's called `dones` everywhere else, so I'm not really sure we should change it here. Having said that, it's only passed in a few places, and never by keyword.

It's certainly not a blocking detail.

src/garage/torch/policies/gaussian_mlp_policy.py
ryanjulian: it's required to use ` i think

src/garage/torch/modules/gaussian_mlp_module.py
```diff
@@ -64,6 +64,9 @@ class GaussianMLPBaseModule(nn.Module):
                exponential transformation.
             - softplus: the std will be computed as log(1+exp(x)).
         layer_normalization (bool): Bool for using layer normalization or not.
+        normal_distribution_cls (torch.distribution): normal distribution class
```
ryanjulian: is this used anywhere? what is it for?

src/garage/torch/modules/gaussian_mlp_module.py
```diff
@@ -211,6 +236,9 @@ class GaussianMLPModule(GaussianMLPBaseModule):
                exponential transformation
             - softplus: the std will be computed as log(1+exp(x))
         layer_normalization (bool): Bool for using layer normalization or not.
+        normal_distribution_cls (torch.distribution): normal distribution class
```
ryanjulian: does this type exist? torch.distribution sounds like a package.

src/garage/torch/modules/gaussian_mlp_module.py
```diff
@@ -266,9 +296,8 @@ def _get_mean_and_log_std(self, *inputs):
             *inputs: Input to the module.
 
         Returns:
-            tuple:
-                * mean (torch.Tensor): The mean of Gaussian distribution.
-                * std (torch.Tensor): The variance of Gaussian distribution.
+            torch.Tensor: The mean of Gaussian distribution.
+            torch.Tensor: The variance of Gaussian distribution.
```
ryanjulian: wait does it return the std, the log_std, or the variance? these distinctions matter.

src/garage/torch/modules/gaussian_mlp_module.py
```diff
@@ -425,9 +459,8 @@ def _get_mean_and_log_std(self, *inputs):
...
```
ryanjulian: same question as above

src/garage/torch/modules/gaussian_mlp_module.py
```diff
@@ -532,9 +570,8 @@ def _get_mean_and_log_std(self, *inputs):
...
```
ryanjulian: same question

src/garage/torch/policies/gaussian_mlp_policy.py
```diff
@@ -88,58 +89,70 @@ def __init__(self,
...
         Returns:
             tuple:
                 * np.ndarray: Predicted actions.
+                    :math:`batch_dim \bullet env_spec.action_space`.
                 * dict:
-                    * list[float]: Mean of the distribution
-                    * list[float]: Standard deviation of logarithmic values of
-                        the distribution
+                    * np.ndarray[float]: Mean of the distribution.
+                    * np.ndarray[float]: Standard deviation of logarithmic
```
ryanjulian: uhh this sounds wrong. it's it the log of the std? you can also use equations instead.

src/garage/torch/policies/gaussian_mlp_policy.py
```diff
@@ -88,58 +89,70 @@ def __init__(self,
...
             observation (torch.Tensor): Observation from the environment.
-            action (torch.Tensor): Predicted action.
+                Shape is :math:`env_spec.observation_space`.
+            action (torch.Tensor): Predicted action. Shape is
+                :math:`env_spec.action_space`.
 
         Returns:
             torch.Tensor: Calculated log likelihood value of the action given
-                observation
+                observation. Shape is :math:`1`.
```
ryanjulian: no need to mention shape if it is a scalar. Just say it's a scalar.

src/garage/torch/policies/tanh_gaussian_mlp_policy.py
```diff
@@ -0,0 +1,196 @@
...
+            observation (np.ndarray): Observation from the environment.
+                Shape is :math:`env_spec.observation_space`.
+
+        Returns:
+            tuple:
+                * np.ndarray: Predicted action. Shape is
+                    :math:`env_spec.action_space`.
+                * dict:
+                    * np.ndarray[float]: Mean of the distribution.
+                    * np.ndarray[float]: Standard deviation of logarithmic
```
ryanjulian: log of the std

src/garage/torch/policies/tanh_gaussian_mlp_policy.py
```diff
@@ -0,0 +1,196 @@
...
+            ret_log_std = (dist.variance.sqrt()).log().squeeze(0).cpu().numpy()
+            return (dist.rsample().squeeze(0).cpu().numpy(),
+                    dict(mean=ret_mean, log_std=ret_log_std))
+
+    def get_actions(self, observations):
+        r"""Get actions given observations.
+
+        Args:
+            observations (np.ndarray): Observations from the environment.
+                Shape is :math:`batch_dim \bullet env_spec.observation_space`.
```
ryanjulian: wait -- really? this seems really weird. isn't the shape just `(batch_dim, S^*)`

also, please use the symbols defined in the documentation for _dtypes.py rather than inventing your own or using words like `batch_dim`

src/garage/torch/policies/tanh_gaussian_mlp_policy.py
```diff
@@ -0,0 +1,196 @@
...
+
+        Returns:
+            tuple:
+                * np.ndarray: Predicted actions. Shape is
+                    :math:`batch_dim \bullet env_spec.action_space`.
```
ryanjulian: i don't think this conforms to \bullet as defined in _dtypes.py

src/garage/torch/policies/tanh_gaussian_mlp_policy.py
```diff
@@ -0,0 +1,196 @@
...
+            return (dist.rsample().cpu().numpy(),
+                    dict(mean=ret_mean, log_std=ret_log_std))
+
+    def log_likelihood(self, observation, action):
+        r"""Compute log likelihood given observations and action.
+
+        Args:
+            observation (torch.Tensor): Observation from the environment.
+                Shape is :math:`env_spec.observation_space`.
+            action (torch.Tensor): Predicted action. Shape is
```
ryanjulian: wait, isn't this a batched operation? is it only defined on a single action?

src/garage/torch/policies/tanh_gaussian_mlp_policy.py
```diff
@@ -0,0 +1,196 @@
...
+                :math:`env_spec.action_space`.
+
+        Returns:
+            torch.Tensor: Calculated log likelihood value of the action given
+                observation. Shape is :math:`1`.
```
ryanjulian: Just call a scalar a scaler. shape 1 is pedantic and hard to understand

src/garage/torch/policies/tanh_gaussian_mlp_policy.py
```diff
@@ -0,0 +1,196 @@
...
+    def entropy(self, observation):
+        r"""Get entropy given observations.
+
+        Args:
+            observation (torch.Tensor): Observation from the environment.
+                Shape is :math:`env_spec.observation_space`.
+
+        Returns:
+            torch.Tensor: Calculated entropy values given observation.
+                Shape is :math:`1`.
```
ryanjulian: scalar

src/garage/torch/policies/tanh_gaussian_mlp_policy.py
ryanjulian: again, returning a tuple is equivalent to multiple returns. please use the correct style.

src/garage/torch/policies/gaussian_mlp_policy.py
```diff
@@ -88,58 +89,70 @@ def __init__(self,
...
+        r"""Get a single action given an observation.
 
         Args:
             observation (np.ndarray): Observation from the environment.
+                Shape is :math:`env_spec.observation_space`.
 
         Returns:
             tuple:
-                * np.ndarray: Predicted action.
+                * np.ndarray: Predicted action. Shape is
```
ryanjulian: please correct this docstring to use multiple returns documentation style

src/garage/torch/modules/gaussian_mlp_module.py
ryanjulian: torch.distribution doesn't even exist as a package.

please remember that these have to be resolved automatically by the documentation package. it's not sufficient to get it "kind-of" correct or "your know what i mean" correct.

https://pytorch.org/docs/stable/distributions.html#distribution

ryanjulian: is there a more specific interface you can use? perhaps anything from the Exponential family?

which interface do you actually rely on?

src/garage/torch/policies/tanh_gaussian_mlp_policy.py
ryanjulian: thanks @krzentner .

this doc still needs a shape.

src/garage/torch/policies/tanh_gaussian_mlp_policy.py
avnishn: Wasn't sure how to approach that, thanks for the clarification.

avnishn: Wasn't sure how to approach that, thanks for the clarification.

PR #1178, opened 2020-02-27
Fix VecEnvExecutor + off-policy algorithms

krzentner:
This requires separating the done and path complete signals.

This is mostly a temporary measure, so that off policy algorithms are fixed for this release.

### DIFF
--- a/src/garage/sampler/off_policy_vectorized_sampler.py
+++ b/src/garage/sampler/off_policy_vectorized_sampler.py
@@ -83,7 +83,7 @@ class OffPolicyVectorizedSampler(BatchSampler):
             obses = self._vec_env.reset()
         else:
             obses = self._last_obses
-        dones = np.asarray([True] * self._vec_env.num_envs)
+        completes = np.asarray([True] * self._vec_env.num_envs)
         running_paths = [None] * self._vec_env.num_envs
         n_samples = 0
 
@@ -92,7 +92,7 @@ class OffPolicyVectorizedSampler(BatchSampler):
             self.algo.es.reset()
 
         while n_samples < batch_size:
-            policy.reset(dones)
+            policy.reset(completes)
             if self.algo.input_include_goal:
                 obs = [obs['observation'] for obs in obses]
                 d_g = [obs['desired_goal'] for obs in obses]
@@ -109,7 +109,9 @@ class OffPolicyVectorizedSampler(BatchSampler):
                 actions, agent_infos = self.algo.policy.get_actions(
                     obs_normalized)
 
-            next_obses, rewards, dones, env_infos = self._vec_env.step(actions)
+            next_obses, rewards, dones, env_infos = \
+                self._vec_env.step(actions)
+            completes = env_infos['vec_env_executor.complete']
             self._last_obses = next_obses
             agent_infos = tensor_utils.split_tensor_dict_list(agent_infos)
             env_infos = tensor_utils.split_tensor_dict_list(env_infos)
--- a/src/garage/sampler/on_policy_vectorized_sampler.py
+++ b/src/garage/sampler/on_policy_vectorized_sampler.py
@@ -112,7 +112,8 @@ class OnPolicyVectorizedSampler(BatchSampler):
 
             policy_time += time.time() - t
             t = time.time()
-            next_obses, rewards, dones, env_infos = self._vec_env.step(actions)
+            next_obses, rewards, dones, env_infos = \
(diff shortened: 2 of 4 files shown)


src/garage/sampler/vec_env_executor.py
```diff
@@ -16,43 +22,77 @@ def __init__(self, envs, max_path_length):
...
+        Returns:
+            int: Number of environments
+
+        """
         return len(self.envs)
 
     @property
     def action_space(self):
-        """Read / write the action space."""
+        """Read / write the action space.
```
ryanjulian: nit: you can only read this.

src/garage/sampler/vec_env_executor.py
```diff
@@ -16,43 +22,77 @@ def __init__(self, envs, max_path_length):
...
+
+        Args:
+            action_n (np.ndarray): Array of actions.
+
+        Returns:
+            tuple: Tuple containing:
+                * observations (np.ndarray)
+                * rewards (np.ndarray)
+                * dones (np.ndarray): The done signal from the environment.
+                * completes (np.ndarray): Episode end (time limit reached or
```
ryanjulian: so it's a little weird that this used to implicitly implement gym.Env.step  and now does not (because of completes). Is it important that completes be returned in this tuple, or can it be put into env_infos?

src/garage/sampler/off_policy_vectorized_sampler.py
```diff
@@ -83,7 +83,7 @@ def obtain_samples(self, itr, batch_size=None, whole_paths=True):
             obses = self._vec_env.reset()
         else:
             obses = self._last_obses
-        dones = np.asarray([True] * self._vec_env.num_envs)
+        completes = np.asarray([True] * self._vec_env.num_envs)
```
avnishn: ah yes thank you for doing this

src/garage/sampler/on_policy_vectorized_sampler.py
```diff
@@ -106,13 +106,14 @@ def obtain_samples(self, itr, batch_size=None, whole_paths=True):
 
         while n_samples < batch_size:
             t = time.time()
-            policy.reset(dones)
+            policy.reset(completes)
```
avnishn: So just to be clear, a policy should always reset after a max path length is reached or if there is a genuine done signal @krzentner

ryanjulian:
If the premise is that this will fix the DDPG+TD3 regression, we should run those on MuJoCo3M to verify this. You might also want to talk to Max since he might be motivated to run these anyway (he wants this fix).

avnishn:
@mergifyio rebase

ryanjulian:
@krzentner is this ready to be merged?

src/garage/sampler/vec_env_executor.py
```diff
@@ -16,43 +22,80 @@ def __init__(self, envs, max_path_length):
...
+
+        Inserts an environment infor 'vec_env_executor.complete' containing the
+        episode end signal (time limit reached or done signal from
+        environment).
+
+        Args:
+            action_n (np.ndarray): Array of actions.
+
+        Returns:
+            tuple: Tuple containing:
```
ryanjulian: tuple returns are the same as multiple returns, so you can document them as:


```python
def step(self, action_n):
   r""" Step all environments using the provided actions.

    Inserts an...

    Note:
        In the shapes below, :math:`K` refers to the vectorization height this executor.

    Args:
        action_n (np.ndarray): Array of actions with shape :math:`(K, A*)`
   
    Returns:
        np.ndarray: Observations tensor with shape :math:`(K, O^*)`
        np.ndarray: Rewards tensor with shape :math:`(K,)`
        np.ndarray: Done signals tensor with shape :math:`(K,)`
        dict[str, np.ndarray]: Dictionary of environment infos, each with shape `(K, X*)`
    """

File: src/garage/sampler/vec_env_executor.py [hunk 6]
```diff
@@ -16,43 +22,80 @@ def __init__(self, envs, max_path_length):
...
+            env_infos[i]['vec_env_executor.complete'] = completes
+        return (obs, rewards, dones,
+                tensor_utils.stack_tensor_dict_list(env_infos))
 
     def reset(self):
-        """Reset all environments."""
+        """Reset all environments.
+
+        Returns:
+            np.ndarray: Observations
```
ryanjulian: `np.ndarray: Observations tensor with shape :math:`(K, O*)`

PR #1182, opened 2020-02-29
Torch utils for GPU usage

avnishn:
- Currently in torch, the standard way of handling
  gpu ussage across multiple objects is potentially
  error causing. The addition of a torch utils that
  allows the user to set a global gpu is therefore
  necessary to allow ease of use in adding gpu support
  for one's torch algorithm.
- One should utilize the garage.torch.utils functions
  instead of the standard torch functions for torch tensor
  creation in order to utilize the device specified by
  garage.torch.utils.Utils.device.
- One should use garage.torch.utils.Utils.set_gpu() in order
  to specify the default gpu to be used by torch.Tensors
  that are created with functions from garage.torch.utils.Utils.


src/garage/torch/utils.py
```diff
@@ -2,77 +2,233 @@
...
+                tensor
+            layout (torch.layout, optional): the desired layout of returned
+                Tensor
+            device (torch.device, optional): the desired device of returned
+                Tensor
+            requires_grad (bool, optional): If autograd should record
+                operations on the returned tensor
+
+        Returns:
+            `torch.Tensor`: A new Tensor filled with the value `0` placed on
```
lywong92: value '1'

src/garage/torch/utils.py
```diff
@@ -2,77 +2,233 @@
...
+                device if not None otherwise on `garage.torch.utils.device`
+
+        """
+        if device is not None:
+            device_placement = device
         else:
-            update(module, name, new_param)
+            device_placement = cls.device
```
naeioi: You can do `device_placement = device or cls.device`

src/garage/torch/utils.py
```diff
@@ -2,77 +2,233 @@
...
-        array_dict[key] = torch.FloatTensor(value)
-    return array_dict
-
-
-def torch_to_np(value_in):
-    """Convert PyTorch tensors to numpy arrays.
-
-    Args:
-        value_in (tuple): Tuple of data in PyTorch tensors.
+    use_gpu = False
```
krzentner: These are basically still just globals. I think the file was better as it was before, since the import pattern was more obvious. This just feels like trying to emulate Java in Python.

Of course, that requires telling pylint to ignore your globals, but I think that's fine.

avnishn: But the problem with this is that we have these variables in the global scope in the older version. While it be very unlikely for users to name their variables something like _USE_GPU, I feel that this is still a bad programming practice. 

Would you like me to change back to the older versioning of things with the globals? @krzentner

krzentner: Python "globals" are still affected by namespaces. It's only "global" to this module.

krzentner: So that isn't actually a problem. I think it was clearer in the older version, since these are actually acting like globals.

src/garage/torch/__init__.py
```diff
@@ -1 +1,4 @@
 """PyTorch-backed modules and algorithms."""
+from garage.torch.utils import Utils
```
ryanjulian: can you just export these all to `garage.torch` (i.e., here) and put them in a file called `_functions.py`?

src/garage/torch/utils.py
```diff
@@ -76,3 +116,35 @@ def update(m, name, param):
...
+
+def set_gpu_mode(mode, gpu_id=0):
+    """Set GPU mode and device ID.
+
+    Args:
+        mode (bool): Whether or not to use GPU
+        gpu_id (int): GPU ID
+
+    """
+    global _GPU_ID  # pylint: disable=global-statement
```
krzentner: I *think* you only need one ignore statement right before this line.

avnishn:
@mergifyio rebase

PR #1186, opened 2020-03-03
Fix torch.TRPO unpickling

krzentner:
I have not yet written tests for this fix, but it looks like it should work.

We probably ultimately want our own `garage.torch.Optimizer` type general enough to include this and `DifferentiableSGD`, but I believe this change is a reasonable compromise between garage and pytorch idioms.

Of course, this is mostly important since it fix resuming, which I think should block the next release.

### DIFF
--- a/src/garage/torch/optimizers/conjugate_gradient_optimizer.py
+++ b/src/garage/torch/optimizers/conjugate_gradient_optimizer.py
@@ -5,6 +5,8 @@ computes the optimal step size that will satisfy the KL divergence constraint.
 Finally, it performs a backtracking line search to optimize the objective.
 
 """
+import warnings
+
 from dowel import logger
 import numpy as np
 import torch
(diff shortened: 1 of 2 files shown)


src/garage/torch/optimizers/conjugate_gradient_optimizer.py
```diff
@@ -178,6 +180,42 @@ def step(self, f_loss, f_constraint):  # pylint: disable=arguments-differ
...
+    def __setstate__(self, state):
+        """Restore the optimizer state.
+
+        Args:
+            state (dict): State dictionary.
+
+        """
+        if 'hvp_reg_coeff' not in state['state']:
+            warnings.warn(
+                'Resuming ConjugateGradientOptimizer from corrupt state')
```
ryanjulian: hmm can you tell me more about why it's corrupt and what (if anything) i can do about it? this is just scary.

krzentner: This should only happen if you're resuming from a checkpoint made with a version of garage before this change. I'll write some tests to make sure that's the case.

krzentner: Okay, I've made it clearer what can be done to fix this problem, and made it a little less scary sounding.

krzentner:
I've added tests. I think this is good to merge.

PR #1187, opened 2020-03-03
Add offline meta-testing and per-task logging to MAML

naeioi:


naeioi:
@Mergifyio rebase

examples/torch/maml_trpo_ml10_meta_test.py
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env python3
+"""This is an example to do meta-testing on ML10 environment."""
+from metaworld.benchmarks import ML10
+
+from garage.experiment.offline_meta_test_helper \
```
ryanjulian: Please put this into the `garage.experiment` namespace

src/garage/_functions.py
```diff
@@ -1,10 +1,71 @@
...
+            log NaN as a placeholder.
+
+    Returns:
+        numpy.ndarray: Undiscounted returns.
+
+    """
+    traj_by_name = defaultdict(list)
+    for trajectory in batch.split():
+        try:
+            task_name = trajectory.env_infos['task_name'][0]
```
ryanjulian: can you leave an issue about unifying this interface?

src/garage/_functions.py
```diff
@@ -1,10 +1,71 @@
...
+            if task_name in traj_by_name:
+                trajectories = traj_by_name[task_name]
+                log_performance(
+                    itr,
+                    garage.TrajectoryBatch.concatenate(*trajectories),
+                    discount,
+                    prefix=task_name)
+            else:
+                with tabular.prefix(task_name + '/'):
+                    tabular.record('Iteration', np.nan)
```
ryanjulian: can you leave an issue on dowel about handling this case better?

src/garage/envs/task_id_wrapper.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def step(self, action):
+        """gym.Env step for the active task env.
+
+        Args:
+            action (np.ndarray): Action performed by the agent in the
+                environment.
+
+        Returns:
+            tuple:
```
ryanjulian: returning a tuple is equivalent to multiple return values. please use that style instead (it looks just like the Args section)

src/garage/envs/task_id_wrapper.py
```diff
@@ -0,0 +1,54 @@
+"""Wrapper for Metaworld environment to provide properly-named tasks."""
+
+import gym
+
+
+class TaskIdWrapper(gym.Wrapper):
```
ryanjulian: if this only works for metaworld, please put that in the name `MetaworldTaskIdWrapper`

src/garage/experiment/offline_meta_test_helper.py
```diff
@@ -0,0 +1,248 @@
+"""Offline meta-testing helper."""
```
ryanjulian: this file looks to me like it should live in scripts/

src/garage/torch/algos/_utils.py
```diff
@@ -110,7 +110,7 @@ def compute_advantages(discount, gae_lambda, max_path_length, baselines,
     deltas = (rewards + discount * F.pad(baselines, (0, 1))[:, 1:] - baselines)
     deltas = F.pad(deltas, (0, max_path_length - 1)).unsqueeze(0).unsqueeze(0)
 
-    advantages = F.conv2d(deltas, adv_filter, stride=1).squeeze()
+    advantages = F.conv2d(deltas, adv_filter, stride=1).view(rewards.shape)
```
ryanjulian: i also saw an update to this in #1188. Is this a bug?

Please confer with @yonghyuc and perhaps open a separate issue/PR so that it can be backported.

tests/garage/experiment/test_task_sampler.py
```diff
@@ -6,6 +6,20 @@
 
 from garage.envs.half_cheetah_vel_env import HalfCheetahVelEnv
 from garage.experiment import task_sampler
+from garage.experiment.task_sampler import FullSetTaskSampler
+
+
+@pytest.mark.large
+def test_round_robin_set_task_sampler():
```
ryanjulian: do we really need to load ML10 to test this??

tests/garage/torch/misc/test_algo_utils.py
```diff
@@ -0,0 +1,39 @@
+"""Tests for utilities in torch/algos/_utils.py."""
+import torch
+
+from garage.torch.algos import compute_advantages
+
+
+class TestAlgoUtils:
```
ryanjulian: just an FYI that pytest doesn't require you define a class for this (you can just use functions called `test_...`)

examples/torch/maml_trpo_ml10_meta_test.py
ryanjulian: please split lines with (), per PEP8

src/garage/torch/algos/_utils.py
naeioi: Yes, this is a bug that breaks when `rewards` has a shape of something like `(1, N)`. The expected shape of `advantages` is the same of that of `rewards`, but the `Tensor.seqeeze()` will squeeze out the the first dimension in this case.

naeioi: @yonghyuc Because this is a bug that impacts old release of garage, you need to make a backport PR for this. If you don't know how to do it, here is an example https://github.com/rlworkgroup/garage/pull/1079.

naeioi:
@ryanjulian OK. I will break this into several smaller PRs.

src/garage/torch/algos/maml.py
```diff
@@ -419,34 +467,30 @@ class MAMLTrajectoryBatch(
...
     +-----------------------+-------------------------------------------------+
-    | :math:`N \bullet [T]` | A dimension computed by flattening a            |
-    |                       | variable-length time dimension :math:`[T]` into |
-    |                       | a single batch dimension with length            |
-    |                       | :math:`sum_{i \in N} [T]_i`                     |
-    +-----------------------+-------------------------------------------------+
 
     Attributes:
+        paths (list[dict[str, np.ndarray or dict[str, np.ndarray]]]):
+            Nonflatten original paths from sampler.
```
krzentner: Wait, why is this necessary? It kinda defeats the point of the datastructure, and doesn't seem to be read anywhere.

src/garage/_functions.py
```diff
@@ -1,10 +1,71 @@
...
+from collections import defaultdict
+
 from dowel import tabular
 import numpy as np
 
+import garage
 from garage.misc.tensor_utils import discount_cumsum
 
 
+def log_multitask_performance(itr, batch, discount, task_names=None):
```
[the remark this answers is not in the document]
naeioi: Sure! Will do.

src/garage/torch/algos/_utils.py
naeioi: Backport PR opened at #1197

src/garage/_functions.py
naeioi: Opened an issue at https://github.com/rlworkgroup/dowel/issues/45

src/garage/_functions.py
naeioi: Opened at https://github.com/rlworkgroup/garage/issues/1198

src/garage/experiment/offline_meta_test_helper.py
```diff
@@ -0,0 +1,248 @@
...
+                        sys.exit()
+                    else:
+                        # In parent process
+                        children.append(pid)
+
+                for child in children:
+                    os.waitpid(child, 0)
+
+            if to_merge:
+                self._merge_csv(meta_train_dir, itrs)
```
krzentner: This file should probably be in `scripts/`, and should have a `if __name__ == '__main__'` block.

src/garage/envs/task_id_wrapper.py
krzentner: Wait, so is this class still necessary?

PR #1189, opened 2020-03-03
Refactor NPO

ahtsan:
This PR refactors NPO, as discussed in #1172 , in preparation of RL2NPO.


src/garage/misc/tensor_utils.py
```diff
@@ -163,6 +169,33 @@ def stack_tensor_dict_list(tensor_dict_list):
...
+
+    Args:
+        paths (list[dict]): List of dict to be stacked and padded.
+            Value of each dict will be shape of :math:`(D, S^*)`.
+        key (str): Key of the values in the paths to be stacked and padded.
+        max_len (int): Maximum length for padding.
+
+    Returns:
+        numpy.ndarray: Stacked and padded tensor. Shape: :math:`(N, D, S^*)`
+            where K is the len of input paths.
```
ryanjulian: K?

ryanjulian:
Can you describe in your commit message/subject in a little more detail *what* and *why* you are refactoring NPO?

src/garage/misc/tensor_utils.py
ahtsan: it's a typo, fixed

ahtsan:
> Can you describe in your commit message/subject in a little more detail _what_ and _why_ you are refactoring NPO?

Done

src/garage/misc/tensor_utils.py
```diff
@@ -23,20 +23,26 @@ def discount_cumsum(x, discount):
                                 axis=0)[::-1]
 
 
-def explained_variance_1d(ypred, y):
+def explained_variance_1d(ypred, y, valids):
```
ryanjulian: hmm perhaps valids should be optional?

ahtsan: Yea

tests/garage/tf/algos/test_batch_polopt2.py
```diff
@@ -18,8 +19,10 @@
 
 class TestBatchPolopt2(TfGraphTestCase):
 
-    @mock.patch.multiple(BatchPolopt2, __abstractmethods__=set())
     # pylint: disable=abstract-class-instantiated, no-member
+    # This test cause low memory with some reason
+    @pytest.mark.flaky
```
ryanjulian: that would be  really surprising. you seen this in the wild on master?

It seems likely that some other part of the test suite is using all if your memory, and this test just became a victim.

ahtsan: that makes sense, I will remove the decorator

PR #1190, opened 2020-03-03
Add RL2 Worker

ahtsan:
Add RL2 worker and test


src/garage/sampler/rl2_worker.py
```diff
@@ -0,0 +1,59 @@
+"""Worker class used in all Samplers."""
+import numpy as np
+
+from garage.sampler.worker import DefaultWorker
+
+
+class RL2Worker(DefaultWorker):
+    """Initialize a worker.
```
krzentner: Can you describe the class in more detail here?

ahtsan: Done

src/garage/sampler/__init__.py
```diff
@@ -18,7 +19,7 @@
 
 __all__ = [
     'BatchSampler', 'Sampler', 'ISSampler', 'singleton_pool', 'LocalSampler',
-    'RaySampler', 'RL2Sampler', 'SamplerWorker', 'ParallelVecEnvExecutor',
-    'VecEnvExecutor', 'OffPolicyVectorizedSampler',
+    'RaySampler', 'RL2Sampler', 'RL2Worker', 'SamplerWorker',
```
ryanjulian: @krzentner do you think this kind of stuff should go in the algorithm or in the common directories like this? i think algorithm.

ahtsan:
> Is RL2Sampler still needed if you have RL2Worker?

Not really. RL2Sampler uses the old API. I think we can remove it safely.

ryanjulian:
Please remove it if you are confident.

src/garage/sampler/__init__.py
ryanjulian: if KR doesn't answer in time, please just put it inside the RL2 module, just like you put RL2Env there.

ahtsan: Agree

ryanjulian: the whole idea of the worker interface is that algorithms can define their own workers, which are unique, and use them with this package, which is shared.

shared stuff goes here.

unique stuff should go in the algorithm.

krzentner: So far in garage, our algorithms don't have directories of their own. You could stick the classes the algorithm needs in that algorithm's file, or we could start using algorithm directories (like in rlkit). I think it would look weird for this to be the only algorithm with its own directory, but it would also be weird to place the class in the main algorithm's file.

I think it's fine where it is now, but at some point we do need to think more carefully about what we want to expose in the common directory and how we want to organize algorithm details like this.

ahtsan: Let RL2 be the first one?

ryanjulian:
Actually, I was asking for you to put it in garage/tf/algos/rl2.py -- I agree with KR that it's too early to put it in a directory. It's a very small class, you can just put it directly in the rl2.py module

ahtsan:
> Actually, I was asking for you to put it in garage/tf/algos/rl2.py -- I agree with KR that it's too early to put it in a directory. It's a very small class, you can just put it directly in the rl2.py module

Also for RL2Env?

ryanjulian:
yes. it would look like

```py
from garage.tf.algos.rl2 import RL2Env
from garage.tf.algos.rl2 import RL2Worker
```
where `RL2Env` and `RL2Worker` are both defined in `src/garage/tf/algos/rl2.py`

src/garage/tf/algos/rl2.py
```diff
@@ -1,9 +1,13 @@
...
+"""Module for RL2.
+
+This module contains RL2Worker and the environment wrapper for RL2.
+"""
 import akro
 import gym
 import numpy as np
 
 from garage.envs.env_spec import EnvSpec
+from garage.sampler.worker import DefaultWorker
```
ryanjulian: is DefaultWorker not exported in `garage.sampler`? Same question for `EnvSpec`

ryanjulian:
do you want to mark this as ready so it will merge?

src/garage/tf/algos/rl2.py
ahtsan: Will fix

PR #1192, opened 2020-03-04
Add log_multitask_performance

krzentner:
log_multitask_performance logs the performance of a batch of
trajectories sampled from multiple tasks.


krzentner:
Note that I simplified the API slightly from the version in `meta_squash`.

PR #1193, opened 2020-03-05
Fix flaky tests for ray sampler

avnishn:
- Ray sampler would routinely fail Travis CI and cause subsequent
  tests to fail.
- This is because the ray server would be reinited during
  newly added tests in non local ray mode. Local mode in ray makes
  code act serially, which is sufficient for our test suite.
- The newly added tests from the new sampler refactor were not apart of the ray_sampler test class,
  so ray would be inited in standard mode for these tests.
- Standard mode ray consumes more memory than is feasible for our
  travis compute instance.
- The fix was made by adding the tests to the ray_sampler test class,
  as opposed to them being standalone test functions.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -3,8 +3,8 @@ import pytest
 import ray
 
 
-@pytest.fixture(scope='module')
-def ray_local_test_fixture():
+@pytest.fixture(scope='function')
+def ray_local_session_fixture():
     """Initializes Ray and shuts down Ray in local mode.
 
     Yields:
@@ -22,8 +22,8 @@ def ray_local_test_fixture():
         ray.shutdown()
 
 
-@pytest.fixture(scope='module')
-def ray_test_fixture():
+@pytest.fixture(scope='function')
+def ray_session_fixture():
     """Initializes Ray and shuts down Ray.
 
     Yields:
--- a/conftest.py
+++ b/conftest.py
@@ -4,6 +4,25 @@ import ray
 
 
 @pytest.fixture(scope='module')
+def ray_local_test_fixture():
+    """Initializes Ray and shuts down Ray in local mode.
+
+    Yields:
+        None: Yield is for purposes of pytest module style.
+            All statements before the yield are apart of module setup, and all
+            statements after the yield are apart of module teardown.
+    """
+    if not ray.is_initialized():
+        ray.init(local_mode=True,
+                 ignore_reinit_error=True,
+                 log_to_driver=False,
+                 include_webui=False)
+    yield
+    if ray.is_initialized():
+        ray.shutdown()
+
+
+@pytest.fixture(scope='module')
 def ray_test_fixture():
     """Initializes Ray and shuts down Ray.
 
--- a/conftest.py
+++ b/conftest.py
@@ -1,0 +1,23 @@
+"""Place Pytest fixtures to be used across all tests here."""
+import pytest
+import ray
+
+
+@pytest.fixture(scope='module')
+def ray_test_fixture():
+    """Initializes Ray and shuts down Ray.
+
+    Yields:
+        None: Yield is for purposes of pytest module style.
+            All statements before the yield are apart of module setup, and all
+            statements after the yield are apart of module teardown.
+    """
+    if not ray.is_initialized():
+        ray.init(memory=52428800,
+                 object_store_memory=78643200,
+                 ignore_reinit_error=True,
+                 log_to_driver=False,
+                 include_webui=False)
+    yield
+    if ray.is_initialized():
+        ray.shutdown()
--- a/examples/tf/trpo_swimmer_ray_sampler.py
+++ b/examples/tf/trpo_swimmer_ray_sampler.py
@@ -6,8 +6,10 @@ sampler.
 Here it runs Swimmer-v2 environment with 40 iterations.
 """
 import gym
+import ray
 
-from garage.experiment import run_experiment
+from garage import wrap_experiment
+from garage.experiment.deterministic import set_seed
 from garage.np.baselines import LinearFeatureBaseline
 from garage.sampler import RaySampler
 from garage.tf.algos import TRPO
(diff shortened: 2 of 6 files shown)


tests/garage/sampler/test_ray_batched_sampler.py
```diff
@@ -78,81 +77,82 @@ def test_ray_batch_sampler(self):
         sampler1.shutdown_worker()
         sampler2.shutdown_worker()
 
+    @pytest.mark.flaky
```
ryanjulian: is this still flaky?

avnishn: no, thanks for the catch

avnishn:
> You can even define your fixture in its own module and import it in all of the tests which want to use ray, setting the scope to "session."
> 
> Then only the first test to use the fixture will bring up ray, and pytest will tear it down once the last test to use it is done.

So I went ahead and implemented this, but one problem I can see arising is that someone could write a piece of code that calls ray.shutdown() before the sessionwide call to ray.shutdown() is call. This would cause ray to get reinited in non_local mode, which will trigger the same ray issues from before.

ryanjulian:
At some point people just need to be careful. I think if they follow your example it should be fine. It's likely that would break your test, anyway.

ryanjulian:
I don't see the versions using fixtures here...

avnishn:
have not pushed it yet, sorry

ahtsan:
Is this PR going to solve the test failure regarding Ray in Travis?

avnishn:
@mergifyio rebase

conftest.py
```diff
@@ -0,0 +1,30 @@
...
+    """Initializes Ray in localmode before session starts."""
+    if not ray.is_initialized():
+        ray.init(memory=52428800,
+                 object_store_memory=78643200,
+                 ignore_reinit_error=True,
+                 log_to_driver=False)
+
+
+@pytest.fixture(scope='session', autouse=True)
+def cleanup_ray(request):
```
ryanjulian: Is there a reason you didn't use the clean-up pattern recommended by the pytest developers?

https://docs.pytest.org/en/latest/fixture.html#fixture-finalization-executing-teardown-code

avnishn:
I have updated this PR so that we are using a uses a pytest fixture style that does both setup and teardown in a single module - Reviewers, direct your attention to garage/conftest.py

conftest.py
```diff
@@ -0,0 +1,23 @@
+"""Place Pytest fixtures to be used across all tests here."""
+import pytest
+import ray
+
+
+@pytest.fixture(scope='module')
```
ryanjulian: what did you find about the interaction between fixture scopes and xdist?

tests/garage/sampler/test_ray_batched_sampler.py
```diff
@@ -47,7 +46,9 @@ def setup_method(self):
     def teardown_method(self):
         self.env.close()
 
-    def test_ray_batch_sampler(self):
+    def test_ray_batch_sampler(self, ray_test_fixture):
+        # pylint: disable=unused-argument
```
ryanjulian: in the future, just `del unused_arg`

tests/garage/tf/samplers/test_ray_batched_sampler_tf.py
```diff
@@ -51,7 +49,9 @@ def setup_method(self):
     def teardown_method(self):
         self.env.close()
 
-    def test_ray_batch_sampler(self):
+    def test_ray_batch_sampler(self, ray_test_fixture):
+        # pylint: disable=unused-argument
```
ahtsan: Same here, `del unused_arg`

tests/garage/tf/experiment/test_local_tf_runner.py
```diff
@@ -139,7 +140,9 @@ def test_make_sampler_local_sampler(self):
             assert isinstance(runner._sampler, LocalSampler)
             runner.train(n_epochs=1, batch_size=10)
 
-    def test_make_sampler_ray_sampler(self):
+    def test_make_sampler_ray_sampler(self, ray_test_fixture):
+        # pylint: disable=unused-argument
```
ahtsan: Same here, `del unused_arg`

tests/garage/sampler/test_ray_batched_sampler.py
```diff
@@ -156,3 +140,25 @@ def test_obtain_exact_trajectories():
         if count % n_traj_per_worker == 0:
             worker += 1
         assert (rollout.actions == per_worker_actions[worker]).all()
+
+
+def test_init_with_env_updates(ray_test_fixture):
+    # pylint: disable=unused-argument
```
ahtsan: Same here, `del unused_arg`

tests/garage/sampler/test_ray_batched_sampler.py
```diff
@@ -112,27 +114,9 @@ def test_update_envs_env_update():
...
-    sampler = RaySampler.from_worker_factory(workers,
-                                             policy,
-                                             envs=tasks.sample(n_workers))
-    rollouts = sampler.obtain_samples(0, 160, policy)
-    assert sum(rollouts.lengths) >= 160
-
-
-def test_obtain_exact_trajectories():
+def test_obtain_exact_trajectories(ray_test_fixture):
+    # pylint: disable=unused-argument
```
ahtsan: Same here, `del unused_arg`

tests/garage/sampler/test_ray_batched_sampler.py
```diff
@@ -79,8 +80,9 @@ def test_ray_batch_sampler(self):
         sampler2.shutdown_worker()
 
 
-@pytest.mark.flaky
-def test_update_envs_env_update():
+def test_update_envs_env_update(ray_test_fixture):
+    # pylint: disable=unused-argument
```
ahtsan: Same here, `del unused_arg`

conftest.py
```diff
@@ -0,0 +1,42 @@
...
+def ray_local_test_fixture():
```
ryanjulian: can you give us some info on why we would use the local version versus the non-local version?

conftest.py
```diff
@@ -0,0 +1,42 @@
...
+                 ignore_reinit_error=True,
+                 log_to_driver=False,
+                 include_webui=False)
+    yield
+    if ray.is_initialized():
+        ray.shutdown()
+
+
+@pytest.fixture(scope='module')
+def ray_test_fixture():
```
ryanjulian: i don't think "test_fixture" is really needed here. `ray` is probably the most canonical name but makes a name collision, so maybe you can use `ray_session` and `ray_local_session`?

examples/tf/trpo_swimmer_ray_sampler.py
```diff
@@ -6,28 +6,39 @@
...
 from garage.tf.experiment import LocalTFRunner
 from garage.tf.policies import GaussianMLPPolicy
 
-seed = 100
 
-
-def run_task(snapshot_config, *_):
-    """Run task.
+@wrap_experiment
+def tf_trpo_swimmer(ctxt=None, seed=1):
```
ryanjulian: isn't ctxt a required argument @krzentner?

examples/tf/trpo_swimmer_ray_sampler.py
```diff
@@ -6,28 +6,39 @@
...
-        *_ (object): Hyperparameters (unused).
+        ctxt (garage.experiment.ExperimentContext): The experiment
+            configuration used by LocalRunner to create the snapshotter.
+        seed (int): Used to seed the random number generator to produce
+            determinism.
+
 
     """
-    with LocalTFRunner(snapshot_config=snapshot_config) as runner:
+    # Since this is an example, we are running ray in a reduced state.
```
ryanjulian: can you explain what is reduced about this?

avnishn: the memory and object story memory available to ray in this example is the minimum amount possible. The runs perfectly fine in this state, as it doesn't use much memory anyways.

PR #1195, opened 2020-03-07
Add RL2, RL2NPO and RL2PPO

ahtsan:
- It also modifies local runner for necessary changes in order to run RL2
- Changes snapshotter to use cloudpickle for pickling lambda expression


examples/tf/rl2_halfcheetah.py
```diff
@@ -0,0 +1,85 @@
...
+    """
+    set_seed(seed)
+    with LocalTFRunner(snapshot_config=ctxt) as runner:
+        max_path_length = 100
+        meta_batch_size = 10
+        n_epochs = 50
+        episode_per_task = 4
+
+        # ---- For ML1-push
+        # env = RL2Env(ML1.get_train_tasks('push-v1'))
```
ryanjulian: if you'd like to include this, please make it a separate example

src/garage/experiment/local_runner.py
```diff
@@ -148,6 +148,9 @@ def __init__(self, snapshot_config, max_cpus=1):
         self.step_itr = None
         self.step_path = None
 
+        self._n_workers = None
```
ryanjulian: can you put these changes to LocalRunner in a separate PR so that @naeioi @zequnyu and @krzentner can review them?

src/garage/tf/algos/rl2.py
```diff
@@ -1,15 +1,340 @@
...
+    def policy(self):
+        """Policy.
+
+        Returns:
+            garage.Policy: Policy to be used.
+
+        """
+        return self._inner_algo.policy
+
+    @property
```
ryanjulian: is there a reason this needs to be public?

src/garage/tf/algos/rl2npo.py
```diff
@@ -0,0 +1,188 @@
...
+                 policy_ent_coeff=0.0,
+                 use_softplus_entropy=False,
+                 use_neg_logli_entropy=False,
+                 stop_entropy_gradient=False,
+                 entropy_method='no_entropy',
+                 flatten_input=True,
+                 fit_baseline='before',
+                 name='NPO'):
+        self._fit_baseline = fit_baseline
+        if fit_baseline != 'before':
```
ryanjulian: Why even accept a parameter at all? Just pass `'before'` to super()

src/garage/tf/algos/rl2ppo.py
```diff
@@ -0,0 +1,113 @@
...
+                 policy_ent_coeff=0.0,
+                 use_softplus_entropy=False,
+                 use_neg_logli_entropy=False,
+                 stop_entropy_gradient=False,
+                 entropy_method='no_entropy',
+                 flatten_input=True,
+                 fit_baseline='before',
+                 name='PPO'):
+        if optimizer is None:
+            optimizer = FirstOrderOptimizer
```
ryanjulian: i don't think you need to accept this parameter. PPO is defined as using a FirstOrderOptimizer.

src/garage/tf/algos/rl2npo.py
```diff
@@ -0,0 +1,188 @@
...
+                         optimizer=optimizer,
+                         optimizer_args=optimizer_args,
+                         policy_ent_coeff=policy_ent_coeff,
+                         use_softplus_entropy=use_softplus_entropy,
+                         use_neg_logli_entropy=use_neg_logli_entropy,
+                         stop_entropy_gradient=stop_entropy_gradient,
+                         entropy_method=entropy_method,
+                         flatten_input=flatten_input)
+
+    def optimize_policy(self, itr, samples_data):
```
ryanjulian: how does this differ from optimize_policy in regular NPO?

src/garage/tf/algos/rl2npo.py
```diff
@@ -0,0 +1,188 @@
...
+        tabular.record('{}/KL'.format(self.policy.name), policy_kl)
+        pol_ent = self._f_policy_entropy(*policy_opt_input_values)
+        tabular.record('{}/Entropy'.format(self.policy.name), np.mean(pol_ent))
+
+        ev = np_tensor_utils.explained_variance_1d(samples_data['baselines'],
+                                                   samples_data['returns'],
+                                                   samples_data['valids'])
+        tabular.record('{}/ExplainedVariance'.format(self.baseline.name), ev)
+
+    def _get_baseline_prediction(self, samples_data):
```
ryanjulian: is there a reason this isn't a method in NPO? then you can do a lot less in the inheritance.

src/garage/tf/algos/rl2.py
```diff
@@ -1,15 +1,340 @@
...
 from garage.sampler import DefaultWorker
 
 
+class RL2(MetaRLAlgorithm):
+    """RL^2.
+
+    Reference: https://arxiv.org/pdf/1611.02779.pdf.
+
+    Args:
+        inner_algo (garage.np.algos.RLAlgorithm): Inner algorithm.
```
ryanjulian: it seems that the actual type of `garage.tf.algos.BatchPolopt`, based on your assert below.

src/garage/tf/algos/rl2.py
```diff
@@ -1,15 +1,340 @@
...
+                                 lengths=lengths,
+                                 baselines=np.zeros_like(rewards),
+                                 agent_infos=agent_infos,
+                                 env_infos=env_infos,
+                                 paths=concatenated_path_in_meta_batch,
+                                 average_return=np.mean(undiscounted_returns))
+
+        return concatenated_path
+
+    def _concatenate_paths(self, paths):
```
ryanjulian: @krzentner can this not be achieved by your recent TrajectoryBatch helpers?

src/garage/tf/algos/rl2.py
```diff
@@ -1,15 +1,340 @@
...
+        """Max path length.
+
+        Returns:
+            int: Maximum path length in a trajectory.
+
+        """
+        return self._max_path_length
+
+
+def stack_paths(max_len, paths):
```
ryanjulian: seems like this should go into `garage.tf.utils`?

examples/tf/rl2_halfcheetah.py
```diff
@@ -0,0 +1,85 @@
...
+
+        env_spec = tasks.sample(1)[0]().spec
+        policy = GaussianGRUPolicy(name='policy',
+                                   hidden_dim=64,
+                                   env_spec=env_spec,
+                                   state_include_action=False)
+
+        baseline = LinearFeatureBaseline(env_spec=env_spec)
+
+        inner_algo = RL2PPO(
```
ryanjulian: this is a really strange API.

why not just have RL2PPO and RL2TRPO, which handle composing these objects automatically?

@krzentner PTAL

src/garage/experiment/local_runner.py
krzentner: I'm already pretty familiar with these changes, so I'm fine reviewing them here. I think they're fine.

src/garage/tf/algos/rl2.py
```diff
@@ -1,15 +1,340 @@
...
+            float: The average return in last epoch.
+
+        """
+        last_return = None
+
+        for _ in runner.step_epochs():
+            runner.step_path = runner.obtain_samples(
+                runner.step_itr,
+                env_update=self._task_sampler.sample(self._meta_batch_size))
+            tabular.record('TotalEnvSteps', runner.total_env_steps)
```
krzentner: Can we just stick this in `LocalRunner.step_epochs` while we're at it? I'm tired of finding an algorithm that forgot it, and now the X axis is silently something else.

src/garage/tf/algos/rl2.py
```diff
@@ -1,15 +1,340 @@
...
+        for path in paths:
+            path['returns'] = np_tensor_utils.discount_cumsum(
+                path['rewards'], self._discount)
+            path['lengths'] = [len(path['rewards'])]
+            if 'batch_idx' in path:
+                paths_by_task[path['batch_idx']].append(path)
+            elif 'batch_idx' in path['agent_infos']:
+                paths_by_task[path['agent_infos']['batch_idx'][0]].append(path)
+            else:
+                raise ValueError('Batch idx is required for RL2 but not found')
```
krzentner: Can you mention in this error message to use `RL2Worker`? As a user, I want the error message to tell me how to fix the problem.

src/garage/tf/algos/rl2.py
krzentner: I'm pretty sure this function can either use that, or `tensor_utils.concat_tensor_dict_list` to be 80% shorter.

src/garage/tf/algos/rl2.py
krzentner: Yeah, `algo.max_path_length` still gets used in a few places.

src/garage/tf/algos/rl2.py
krzentner: I think this function might already exist. Also, it should definitely be in `garage.misc.tensor_utils`, there's nothing TensorFlow specific here.

src/garage/tf/algos/rl2.py
krzentner: I'm pretty sure the actual type should be `garage.tf.algos.RL2NPO`.

src/garage/tf/algos/rl2npo.py
ahtsan: This fit baseline first, regular NPO fit baseline after

src/garage/tf/algos/rl2npo.py
ahtsan: NPO get baseline prediction in `BatchPolopt.process_samples()`, so it isn't really needed in NPO

src/garage/tf/algos/rl2.py
ryanjulian: SGTM

src/garage/tf/algos/rl2npo.py
ryanjulian: Wait, then why is there a constructor argument which tells NPO when to fit the baseline? It seems like this is a mix of 2 designs

1. Modify NPO to support some weird things (like fitting the baseline before), then pass only constructor params
1. Make special versions of NPO which do what you want.

I'm kind of in favor of 1, but I don't mind either as long as you are consistent.

src/garage/tf/algos/rl2npo.py
ryanjulian: see my comment above about picking a consistent implementation style

If you're using inheritance from NPO, you shouldn't be re-implementing functionality from NPO under *different* names. Inheritance is not a tool for arbitrarily picking and choosing functions from one class into another. It's a tool for *specializing* types from more general to more specific.

This means if `Foo.speak()` exists and says "Hello", and you want to make `Bar.speak()` say "Hello, Anson", you shouldn't just define `Bar._speak()` and call that instead! Now `Bar` has both `Bar.speak()` and `Bar._speak()`!!! If I pass `bar = Bar()` to a function `make_it_speak()` which was defined for `Foo`, it will still call `Bar.speak()` and get "Hello", which is not what you wanted. You haven't created a more specific type, you've created a new type entirely!

Now if `Bar` can `walk()` and `Foo` can't `walk()`, then yes you can add `Bar.walk()`. `Bar` can still `speak()`, so `class Bar(Foo)` still applies.

In your situation, you either have to make Foo general-enough that you can effectively inherit from it (e.g. like adding options to NPO), or you can abandon inheritance and use composition instead. Since we've already discussed why composition is hard right now, I suggest you make NPO more flexible for your purposes.

src/garage/tf/algos/rl2.py
ahtsan: fixed

src/garage/tf/algos/rl2.py
ahtsan: fixed

src/garage/tf/algos/rl2.py
ahtsan: fixed

src/garage/tf/algos/rl2.py
ahtsan: fixed

examples/tf/rl2_ppo_halfcheetah.py
```diff
@@ -0,0 +1,80 @@
...
+
+        env_spec = tasks.sample(1)[0]().spec
+        policy = GaussianGRUPolicy(name='policy',
+                                   hidden_dim=64,
+                                   env_spec=env_spec,
+                                   state_include_action=False)
+
+        baseline = LinearFeatureBaseline(env_spec=env_spec)
+
+        inner_algo_args = dict(
```
ryanjulian: can you create RL2PPO and RL2TRPO, which are basically just wrappers which have these args as constructor params rather than passing them as a dict? this will make a better API for uses (its also consistent with MAML).

examples/tf/rl2_ppo_halfcheetah.py
```diff
@@ -0,0 +1,80 @@
...
+                batch_size=32,
+                max_epochs=10,
+            ),
+            stop_entropy_gradient=True,
+            entropy_method='max',
+            policy_ent_coeff=0.02,
+            center_adv=False,
+        )
+
+        algo = RL2(inner_algo='PPO',
```
ryanjulian: please make the strings lowercase, i.e. `ppo` and `trpo`

examples/tf/rl2_trpo_halfcheetah.py
```diff
@@ -0,0 +1,74 @@
...
+    with LocalTFRunner(snapshot_config=ctxt) as runner:
+        max_path_length = 100
+        meta_batch_size = 10
+        n_epochs = 50
+        episode_per_task = 4
+
+        tasks = task_sampler.SetTaskSampler(lambda: RL2Env(
+            env=HalfCheetahVelEnv()))
+
+        env_spec = tasks.sample(1)[0]().spec
```
ryanjulian: is there a less nasty way to do this?

setup.cfg
```diff
@@ -18,7 +18,7 @@ extend-ignore =
     F841  # Unused variables are checked by pylint
 
 [tool:pytest]
-addopts = -n auto -rfE -s --strict-markers
+addopts = -rfE -s --strict-markers
```
ryanjulian: uhh auto is here for a reason, please do not remove it.

src/garage/tf/algos/_rl2trpo.py
```diff
@@ -0,0 +1,117 @@
+"""Trust Region Policy Optimization."""
+from garage.tf.algos._rl2npo import RL2NPO
+from garage.tf.optimizers import ConjugateGradientOptimizer
+from garage.tf.optimizers import PenaltyLbfgsOptimizer
+
+
+class RL2TRPO(RL2NPO):
```
ryanjulian: there is not much reason to use the names `RL2PPO` and RL2TRPO` for these because they are very simple and the user won't interact with them.

i think it would be better for RL2TRPO and RL2PPO to be subclasses of RL2 which allow the user to pass PPO/TRPO constructor args directly rather than using `inner_algo_args`. You can replace what you currently call RL2TRPO and RL2PPO which just different calls to RL2NPO.

Basically (I use *args for brevity, but please expand all of the args per usual)
```python
class RL2NPO(NPO):  # private, unchanged
      # ...


class RL2(MetaRLAlgorithm):  # public
    
    def __init__(self, *rl2_stuff, inner_algo, inner_algo_args):
        # ...
        if inner_algo == 'ppo'
            optimizer = FirstOrderOptimizer
            # etc...
        elif inner_algo == 'trpo'
            optimizer = ConjugateGradientOptimizer
            # etc...
        else:
            raise ValueError('inner_algo must be one of \'ppo\' or \'trpo\', but got \'{}\' instead'.format(inner_algo))
        
        self._inner_algo = RL2NPO(optimizer=optimizer, *inner_algo_args)


class RL2TRPO(RL2):   # public

    def __init__(self, *rl2_stuff, *trpo_stuff):
        # ...
        super.__init__(*rl2_stuff, inner_algo='trpo', inner_algo_args=trpo_stuff)


class RL2PPO(RL2):  # public

    def __init__(self, *rl2_stuff, *ppo_stuff):
        # ...
        super.__init__(*rl2_stuff, inner_algo='ppo', inner_algo_args=ppo_stuff)

Comment (ahtsan) [reply src/garage/tf/algos/_rl2trpo.py]: Should we expand `*rl2_stuff` in `RL2`? I think arguments for `PPO` and `TRPO` are not identifical, e.g. only TRPO has `kl_constraint`

Comment (ahtsan) [reply src/garage/tf/algos/_rl2trpo.py]: Actually `RL2` should be an ABC?

Comment (ryanjulian) [reply src/garage/tf/algos/_rl2trpo.py]: RL2 should not be an ABC, because someone could provide an inner_algo and instantiate it. we just won't advertise it.

note that in my example, RL2 has `inner_algo_args` (a dict, just like now) and `RL2PPO`/`RL2TRPO` has `*ppo_stuff`/`trpo_stuff` (i.e. expanded). Because RL2PPO is just like PPO, it should have all the PPO args expanded.

Comment (ryanjulian) [reply src/garage/tf/algos/_rl2trpo.py]: so basically, no, RL2's signature doesn't change

Comment (ahtsan) [reply src/garage/tf/algos/_rl2trpo.py]: In your example, logic about inner algorithm is done within RL2, while in `PPO` and `TRPO`, the logic is done in the constructor, i.e. before calling `super.init()`. To stick with garage API, I think it's better to make RL2 an ABC, and if one wants to add a new inner algorithm say `RL2XXX`, he would subclass RL2 and handle the logic, and we don't have to touch RL2 ever.

Comment (ahtsan) [reply src/garage/tf/algos/_rl2trpo.py]: @krzentner PTAL

**Comment by ryanjulian** [comment]
@Mergifyio rebase

File: src/garage/experiment/local_runner.py [hunk 19]
```diff
@@ -353,7 +387,9 @@ def log_diagnostics(self, pause_for_plot=False):
         """
         logger.log('Time %.2f s' % (time.time() - self._start_time))
         logger.log('EpochTime %.2f s' % (time.time() - self._itr_start_time))
+        tabular.record('TotalEnvSteps', self._stats.total_env_steps)
```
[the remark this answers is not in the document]
ahtsan: Good catch, will fix

PR #1197, opened 2020-03-10
Fix torch compute_advantages

naeioi:
Fix a bug that breaks `compute_advantage()` when rewards has a shape 
of like (1, N).  The expected shape of advantages is the same of that of 
rewards, but the Tensor.seqeeze() will wrongly squeeze out the 
the first dimension in such a case.


ryanjulian:
FYI mergify might not create a rebase until this is merged.

naeioi:
@Mergifyio rebase

ryanjulian:

PR #1201, opened 2020-03-11
Separate RUN_ARGS and BUILD_ARGS in Makefile

ryanjulian:
This fixes a bug in `make test` created by #1196


Makefile
```diff
@@ -117,7 +117,7 @@ run-ci:
 		-e MJKEY \
 		--memory 7500m \
```
krzentner: Do we still need to pass these here if we're passing them by via RUN_ARGS?

ryanjulian: `make test` passes them via RUN_ARGS because it depends on `run-headless`. This imposes the limit for `run-ci` which is a different target.

krzentner: Ah, I see.

PR #1214, opened 2020-03-12
Log per-task performance in MAML

naeioi:
This PR adds per-task logging to MAML. 

A wrapper named `MetaworldTaskNamesWrapper` is also added. It saves a `task_name` in `env_info` dict returned by `env.step(action)`. It also provides a `wrapper.task_names()` that returns all available tasks of the _wrapped_ Metaworld environment.


src/garage/torch/algos/maml.py
```diff
@@ -402,51 +400,47 @@ def evaluate_performance(self, itr, all_samples, loss_before, loss_after,
...
     | :math:`S^*`           | Single-step shape of a time-series tensor       |
     +-----------------------+-------------------------------------------------+
-    | :math:`N \bullet [T]` | A dimension computed by flattening a            |
-    |                       | variable-length time dimension :math:`[T]` into |
-    |                       | a single batch dimension with length            |
-    |                       | :math:`sum_{i \in N} [T]_i`                     |
-    +-----------------------+-------------------------------------------------+
 
     Attributes:
+        paths (list[dict[str, np.ndarray or dict[str, np.ndarray]]]):
```
naeioi: I add `paths` to `MAMLTrajectoryBatch` because I need it to construct `TrajectoryBatch` here https://github.com/rlworkgroup/garage/blob/281ada27351d07817969c9985cd3ab1eb080ea2d/src/garage/torch/algos/maml.py#L386

src/garage/envs/metaworld_task_names_wrapper.py
```diff
@@ -0,0 +1,58 @@
+"""A wrapper for Metaworld environment to expose names of all tasks."""
+import gym
+
+
+class MetaworldTaskNamesWrapper(gym.Wrapper):
```
krzentner: Can we just add this to metaworld? It looks pretty useful generally.

ryanjulian: i agree -- we can just add this feature to metaworld

src/garage/torch/algos/maml.py
krzentner: Ah, I see. This is definitely a code smell, but understandable for the time being.
In the future, I think we should change `TrajectoryBatch` to be a type that can be easily subclassed to add additional data (for example, using [`types.SimpleNamespace`](https://docs.python.org/3/library/types.html#types.SimpleNamespace), [`typing.NameTuple`](https://docs.python.org/3/library/typing.html#typing.NamedTuple), or [`dataclass`](https://github.com/ericvsmith/dataclasses).

krzentner: But for this PR I don't think we need to change it, and you can just keep paths.

src/garage/envs/metaworld_task_names_wrapper.py
naeioi: Will do.

src/garage/torch/algos/maml.py
ryanjulian: @krzentner if you feel like that's a fully-fleshed idea, please add an issue for it. i'm a little wary of type proliferation but i can see the benefits of subclassing.

src/garage/envs/metaworld_task_names_wrapper.py
naeioi: `MetaworldTaskNamesWrapper` moved to here https://github.com/rlworkgroup/metaworld/pull/62

naeioi:
This PR is now waiting for https://github.com/rlworkgroup/metaworld/pull/62 and https://github.com/rlworkgroup/garage/issues/1215. The CI will not pass until these two blockers are resolved.

naeioi:
@Mergifyio rebase

src/garage/torch/algos/maml.py
krzentner: I've opened an issue  #1226 for this.

naeioi:
@Mergifyio rebase

examples/torch/maml_ppo_half_cheetah_dir.py
```diff
@@ -1,25 +1,40 @@
 #!/usr/bin/env python3
 """This is an example to train MAML-VPG on HalfCheetahDirEnv environment."""
+# pylint: disable=no-value-for-parameter
```
ryanjulian: is this solved by providing default args to your launcher function?

tests/integration_tests/test_examples.py
```diff
@@ -119,3 +120,25 @@ def test_step_dm_control_env():
     assert subprocess.run(
         [EXAMPLES_ROOT_DIR / 'step_dm_control_env.py', '--n_steps', '1'],
         check=False).returncode == 0
+
+
+@pytest.mark.no_cover
```
ryanjulian: is it possible to provide default args in your examples so that the special cases here aren't necessary?

the general idea in this test is that most examples should be self-contained and can be covered by `test_algo_examples`

examples/torch/maml_ppo_half_cheetah_dir.py
naeioi: The default args are provided in `click.option(arg, default=value)`, and later, no arguments is needed to call `maml_ppo()`. But pylint cannot recognize this way in which click handles arguments

naeioi:
@Mergifyio rebase

tests/integration_tests/test_examples.py
naeioi: I have to reduce epochs and rollouts per task to 1 to save time.

naeioi:
* Add per-task logging to MAML.

* Add MetaworldTaskNamesWrapper wrapper. It saves a task_name in env_info dict returned by env.step(action). It also provides a wrapper.task_names() that returns all available tasks of the wrapped Metaworld environment.

PR #1216, opened 2020-03-12
Flatten env_info of HalfCheetah* environments

naeioi:
This PR flattens `env_info` of `HalfCheetahVel` and `HalfCheetahDir` to be a dict of only float. Previously, `env_info['task']` is a dict, but `env_info` should not have a value of dict type because it is not compatible with `TrajectoryBatch.concatenate()`.

Another option is to change `TrajectoryBatch` to make them compatible with each other. But since `env_info` already has flatten fields regarding reward (i.e. `reward_ctrl` and `reward_forward`) I think changing `HalfCheetah*` makes more sense.

### DIFF
--- a/src/garage/envs/half_cheetah_dir_env.py
+++ b/src/garage/envs/half_cheetah_dir_env.py
@@ -31,7 +31,6 @@ class HalfCheetahDirEnv(HalfCheetahEnvMetaBase):
 
     def __init__(self, task=None):
         task = task or {'direction': 1.}
-        self._task = task
         self._goal_dir = task['direction']
         super().__init__()
 
@@ -54,9 +53,8 @@ class HalfCheetahDirEnv(HalfCheetahEnvMetaBase):
                         control cost.
                     * reward_ctrl (float): The reward for acting i.e. the
                         control cost (always negative).
-                    * task (dict):
-                        * direction (float): 1.0 for forwards, -1.0 for
-                            backwards.
+                    * task_dir (float): Target direction. 1.0 for forwards,
+                        -1.0 for backwards.
 
         """
         xposbefore = self.sim.data.qpos[0]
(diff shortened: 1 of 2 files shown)


src/garage/envs/half_cheetah_dir_env.py
```diff
@@ -54,9 +53,8 @@ def step(self, action):
                         control cost.
                     * reward_ctrl (float): The reward for acting i.e. the
                         control cost (always negative).
-                    * task (dict):
-                        * direction (float): 1.0 for forwards, -1.0 for
-                            backwards.
+                    * task_dir (float): Target direction. 1.0 for forwards,
```
ryanjulian: should this be an int or bool instead?

ryanjulian:
@krzentner do you think that TrajectoryBatch is overly-restrictive here? I am tempted to keep it restrictive until we get evidence it absolutely must be changed.

naeioi:
@ryanjulian Actually, it is not that `TrajectoryBatch` is being overly restrictive on verifying the type, but this bug breaks `TrajectoryBatch.concatenate()`.

ryanjulian:
ah! can we make concatenate more flexible?

naeioi:
@ryanjulian Yes I can fix `TrajectoryBatch.concatenate()` instead. I prefer to attack on HalfCheetah though, because there are already fields that are flatten (i.e. `reward_ctrl` and `reward_forward`, instead of a single `reward` dict).

ryanjulian:
Sure, sounds good to me. Whichever you prefer.

naeioi:
@krzentner I want to hear your final opinion since you contributed HalfCheetah.

naeioi:
@Mergifyio rebase

src/garage/envs/half_cheetah_dir_env.py
krzentner: I *think* it's easier for it to be a float.

Really, I feel like task out to be an `np.ndarray`, and we should also have a `task_space` containing it.
But this is fine.

ahtsan: With new API (still under discussion), task would probably become a TaskSpec? @ryanjulian @naeioi

ryanjulian: Let's see how that API does in metaworld before we start propagating it.

krzentner:
> 
> 
> @krzentner do you think that TrajectoryBatch is overly-restrictive here? I am tempted to keep it restrictive until we get evidence it absolutely must be changed.

It seems to usually be fine? Let's leave it be for now. It would be nice to have a way to globally disable it though or something.

PR #1218, opened 2020-03-13
Add PEARL algorithm

lywong92:


examples/torch/pearl_ml1_push.py
```diff
@@ -0,0 +1,163 @@
+"""PEARL ML1 example."""
+
+import akro
+from metaworld.benchmarks import ML1
+import numpy as np
+
+from garage.envs import normalize
+from garage.envs.base import GarageEnv
```
ryanjulian: please use canonical imports for all of these -- that is, `from garage.envs import GarageEnv` etc.

examples/torch/pearl_ml1_push.py
```diff
@@ -0,0 +1,163 @@
...
+    embedding_mini_batch_size=64,
+    max_path_length=150,
+    reward_scale=10.,
+    use_information_bottleneck=True,
+    use_next_obs_in_context=False,
+    use_gpu=True,
+)
+
+
+def run_task(snapshot_config, *_):
```
ryanjulian: please use the `wrap_experiment` interface for this instead. that means each entry in `params` will become an argument to your experiment function here.

examples/torch/pearl_ml1_push.py
```diff
@@ -0,0 +1,163 @@
...
+                 worker_class=PEARLWorker)
+
+    worker_args = dict(deterministic=True, accum_context=True)
+    meta_evaluator = MetaEvaluator(runner,
+                                   test_task_sampler=test_env_sampler,
+                                   max_path_length=params['max_path_length'],
+                                   worker_class=PEARLWorker,
+                                   worker_args=worker_args,
+                                   n_test_tasks=params['num_test_tasks'])
+    pearl.evaluator = meta_evaluator
```
ryanjulian: is there no way to make this a constructor arg instead?

src/garage/_dtypes.py
```diff
@@ -452,24 +452,56 @@ class TimeStep(
     def __new__(cls, env_spec, observation, action, reward, next_observation,
                 terminal, env_info, agent_info):  # noqa: D102
 
-        # observation
+        # obseqrvation
```
ryanjulian: if you need these changes, please send them in a separate PR

examples/torch/pearl_ml1_push.py
```diff
@@ -0,0 +1,163 @@
...
+                                 hidden_sizes=[200, 200, 200])
+
+    space_a = akro.Box(low=-1,
+                       high=1,
+                       shape=(obs_dim + params['latent_size'], ),
+                       dtype=np.float32)
+    space_b = akro.Box(low=-1, high=1, shape=(action_dim, ), dtype=np.float32)
+    augmented_env = EnvSpec(space_a, space_b)
+
+    qf1 = ContinuousMLPQFunction(env_spec=augmented_env,
```
ryanjulian: if qf1 and qf2 are always the same, can you take a single parameter `qf` and just duplicate it to get qf1 and qf2?

examples/torch/pearl_ml1_push.py
```diff
@@ -0,0 +1,163 @@
...
+                            dtype=np.float32)
+    vf_env = EnvSpec(obs_space, action_space)
+
+    vf = ContinuousMLPQFunction(env_spec=vf_env,
+                                hidden_sizes=[net_size, net_size, net_size])
+
+    policy = TanhGaussianMLPPolicy(env_spec=augmented_env,
+                                   hidden_sizes=[net_size, net_size, net_size])
+
+    context_conditioned_policy = ContextConditionedPolicy(
```
ryanjulian: since PEARL doesn't support anything but this policy, I think it would be more natural to pass the policy and encoder and have PEARL internally construct a ContextConditionedPolicy. That would also hide a lot of this nasty space munging from the launcher

examples/torch/pearl_ml1_push.py
ahtsan: I think no. Under the current API, there is a circular reference -- algo takes evaluator, evaluator takes runner and runner needs algo.policy in `runner.make_sampler`, which requires `runner.setup(algo, ...)` to be called.

examples/torch/pearl_ml1_push.py
```diff
@@ -0,0 +1,163 @@
...
+    env_sampler = SetTaskSampler(lambda: GarageEnv(
+        normalize(ML1.get_train_tasks('push-v1'))))
+    env = env_sampler.sample(params['num_train_tasks'])
+
+    test_env_sampler = SetTaskSampler(lambda: GarageEnv(
+        normalize(ML1.get_test_tasks('push-v1'))))
+    #test_env = test_env_sampler.sample(params['num_train_tasks'])
+
+    runner = LocalRunner(snapshot_config)
+    obs_dim = int(np.prod(env[0]().observation_space.shape))
```
ryanjulian: is there a way we can hide this stuff inside the primitives or the algorithms? it seems it should be constant for every PEARL experiment.

examples/torch/pearl_ml1_push.py
ryanjulian: is there an issue open which would resolve this problem?

src/garage/sampler/pearl_worker.py
```diff
@@ -0,0 +1,99 @@
+"""Module for PEARLWorker."""
```
ryanjulian: did you forget to remove this file? I thought this was moved to pearl.py

examples/torch/pearl_ml1_push.py
krzentner: This is actually kinda difficult to do well in PyTorch, since you need to also specify how to re-initialize the model.
I guess we could add a `reinitialize` method to our pytorch models.

examples/torch/pearl_ml1_push.py
krzentner: Yeah, that would probably be better.

examples/torch/pearl_ml1_push.py
ahtsan: Would it be similar to Policy.clone() in TensorFlow branch?

examples/torch/pearl_ml1_push.py
ahtsan: Yes

krzentner: Okay, I didn't understand the real circular reference problem until yesterday, since I missed Anson's comment here.
I'll have it fixed soon, and then the `MetaEvaluator` can be constructed without the runner or algorithm.

examples/torch/pearl_ml1_push.py
krzentner: No, it's sorta the opposite.

lywong92: I can just use deepcopy actually.

examples/torch/pearl_ml1_push.py
lywong92: I need `obs_dim` and `action_dim` to make `env_spec` for qf and vf. I dont think there's a way around that right now.

ryanjulian: how about a classmethod on PEARL which takes the regular env_spec and returns the augmented one? that way, you can be guaranteed that you get the right spec.

src/garage/sampler/pearl_worker.py
lywong92: Yeah I will delete this.

examples/torch/pearl_ml1_push.py
lywong92: Are you suggesting to construct qf and vf inside PEARL as well? Because I need to pass the augmented space into qf and vf.

ryanjulian: I'm suggesting something like

```python
# g/t/a/pearl.py
class PEARL:

    @classmethod
    def augment_env_spec(cls, env_spec, latent_dim):
        # calculations...

        return EnvSpec(aug_obs, aug_act)

    @classmethod
    def get_encoder_spec(cls, env_spec, latent_dim):
        # calculations...

        return InOutSpec(in_space, out_space)
```

```python
# g/_dtypes.py
class InOutSpec:
    """Describes the input and output spaces of a primitive or module.

    Args:
        input_space (akro.Space): Input space of a module.
        output_space (akro.Space): Output space of a module.
    """
    def __init__(self, input_space, output_space):
        self._input_space = input_space
        self._output_space = output_space

    @property
    def input_space(self):
        """akro.Space: Input space of the module."""
        return self._input_space

    @property
    def output_space(self):
    """akro.Space: Output space of the module."""
        return self._output_space

    @property
    def transpose(self):
    """InOutSpec: A spec which has the input of this space as output, and vice-versa."""
        return InOutSpec(self._output_space, self._input_space)
```

```python
# g/e/env_spec.py
class EnvSpec(InOutSpec):
    """Describes the action and observation spaces of an environment.

    Args:
        observation_space (akro.Space): Observation space of the environment.
        action_space (akro.Space): Action space of the environment.
    """
    def __init__(self, observation_space, action_space):
        super().__init__(action_space, observation_space)

    @property
    def action_space(self):
    """akro.Space: Action space of the environment."""
        return self.input_space

    @property
    def observation_space(self):
    """akro.Space: Observation space of the environment."""
        return self.output_space
```

examples/torch/pearl_ml1_push.py
```diff
@@ -0,0 +1,135 @@
...
+    """
+    set_seed(seed)
+    # create multi-task environment and sample tasks
+    env_sampler = SetTaskSampler(lambda: GarageEnv(
+        normalize(ML1.get_train_tasks('push-v1'))))
+    env = env_sampler.sample(params['num_train_tasks'])
+
+    test_env_sampler = SetTaskSampler(lambda: GarageEnv(
+        normalize(ML1.get_test_tasks('push-v1'))))
+    snapshot_config = SnapshotConfig(snapshot_dir=os.path.join(
```
ryanjulian: is there a reason that you can't use the default here? AFAICT this is the default SnapshotConfig

examples/torch/pearl_ml1_push.py
```diff
@@ -0,0 +1,135 @@
...
+    reward_scale=10.,
+    use_information_bottleneck=True,
+    use_next_obs_in_context=False,
+    use_gpu=True,
+)
+
+
+# pylint: disable=unused-argument
+@wrap_experiment
+def torch_pearl_ml1_push(ctxt=None, seed=1, **params):
```
ryanjulian: please explicitly list the params to the function -- you can use defaults you have here as the default args. this is the style we're going for with the wrap_experiment API

examples/torch/pearl_ml1_push.py
```diff
@@ -0,0 +1,135 @@
...
+                                   max_path_length=params['max_path_length'],
+                                   worker_class=PEARLWorker,
+                                   worker_args=worker_args,
+                                   n_test_tasks=params['num_test_tasks'])
+    pearl.evaluator = meta_evaluator
+    runner.train(n_epochs=params['num_epochs'],
+                 batch_size=params['batch_size'])
+
+
+torch_pearl_ml1_push(**algo_params)
```
ryanjulian: this can be a no-argument function call if you use default params

src/garage/envs/env_spec.py
```diff
@@ -11,5 +13,44 @@ class EnvSpec:
...
+    def observation_space(self):
+        """Get observation space of the env.
+
+        Returns:
+            akro.Space: Observation space.
+
+        """
+        return self.output_space
+
+    @action_space.setter
```
ryanjulian: please do not have setters. this omission was intentional -- this is supposed to be immutable. was there any need for this somewhere?

lywong92: A lot of existing tests set `action_space` and `observation_space`, so they are failing with setters.

examples/torch/pearl_ml1_push.py
ryanjulian: an alternative (if you want to avoid re-documenting a ton of parameters), is just to pass most of these to PEARL instead of making them parameters of the experiment.

src/garage/torch/algos/pearl.py
```diff
@@ -1,6 +1,635 @@
...
+
+    PEARL, which stands for Probablistic Embeddings for Actor-Critic
+    Reinforcement Learning, is an off-policy meta-RL algorithm. It is built
+    on top of SAC using two Q-functions and a value function with an addition
+    of an inference network that estimates the posterior :math:`q(z \| c)`.
+    The policy is conditioned on the latent variable Z in order to adpat its
+    behavior to specific tasks.
+
+    Args:
+        env (list): Batch of sampled environment updates(EnvUpdate), which,
```
ryanjulian: `list[GarageEnv]`

src/garage/torch/algos/pearl.py
```diff
@@ -1,6 +1,635 @@
...
+            data during training (in trajectories).
+        eval_deterministic (bool): Whether to make policy deterministic during
+            evaluation.
+
+    """
+
+    # pylint: disable=too-many-statements
+    def __init__(self,
+                 env,
+                 policy_class,
```
ryanjulian: i think that both of these classes should have default args, to save most people time from looking them up (most people won't change these)

src/garage/torch/algos/pearl.py
```diff
@@ -1,6 +1,635 @@
...
+                 qf_lr=3E-4,
+                 vf_lr=3E-4,
+                 context_lr=3E-4,
+                 policy_mean_reg_coeff=1E-3,
+                 policy_std_reg_coeff=1E-3,
+                 policy_pre_activation_coeff=0.,
+                 soft_target_tau=0.005,
+                 kl_lambda=.1,
+                 optimizer_class=torch.optim.Adam,
+                 recurrent=False,
```
ryanjulian: does `recurrent` refer to the policy? the encoder? both? neither? i think it should be clear from the argument name.

src/garage/torch/algos/pearl.py
```diff
@@ -1,6 +1,635 @@
...
+                 embedding_batch_size=1024,
+                 embedding_mini_batch_size=1024,
+                 max_path_length=1000,
+                 discount=0.99,
+                 replay_buffer_size=1000000,
+                 reward_scale=1,
+                 update_post_train=1,
+                 eval_deterministic=True):
+
+        self.env = env
```
ryanjulian: is there a reason this can't be private?

src/garage/torch/algos/pearl.py
```diff
@@ -1,6 +1,635 @@
...
+        self._task_idx = None
+
+        encoder_spec = self.get_env_spec(env[0](), latent_dim, 'encoder')
+        encoder_in_dim = int(np.prod(encoder_spec.input_space.shape))
+        encoder_out_dim = int(np.prod(encoder_spec.output_space.shape))
+        context_encoder = encoder_class(input_dim=encoder_in_dim,
+                                        output_dim=encoder_out_dim,
+                                        hidden_sizes=encoder_hidden_sizes)
+
+        self.policy = policy_class(
```
ryanjulian: is there a reason this can't be private?

src/garage/torch/algos/pearl.py
```diff
@@ -1,6 +1,635 @@
...
+            # sample train tasks and optimize networks
+            self.train_once()
+            runner.step_itr += 1
+
+            logger.log('Evaluating...')
+            # evaluate
+            self.policy.reset_belief()
+            self.evaluator.evaluate(self)
+
+    def train_once(self):
```
ryanjulian: can you make this (and all other methods not part of `MetaRLAlgorithm`) private?

the new classmethods are an exception, of course.

src/garage/torch/algos/pearl.py
```diff
@@ -1,6 +1,635 @@
...
+    def get_env_spec(cls, env_spec, latent_dim, module):
+        """Get environment specs of encoder with latent dimension.
+
+        Args:
+            env_spec (garage.envs.EnvSpec): Environment specs.
+            latent_dim (int): Latent dimension.
+            module (str): Module to get environment specs for.
+
+        Returns:
+            garage.envs.EnvSpec: Encoder environment specs with latent
```
ryanjulian: looks like this function returns `InOutSpec` (EnvSpec is a subclass)

it might be more clear if you just split these.

also, if this isn't used from the launcher, it can just be private.

src/garage/torch/algos/pearl.py
```diff
@@ -1,6 +1,635 @@
...
+                           shape=(obs_dim + latent_dim, ),
+                           dtype=np.float32)
+        aug_act = akro.Box(low=-1,
+                           high=1,
+                           shape=(action_dim, ),
+                           dtype=np.float32)
+        return EnvSpec(aug_obs, aug_act)
+
+    @classmethod
+    def get_env_spec(cls, env_spec, latent_dim, module):
```
ryanjulian: switching your return type based on an enum (`module`) is often a sure sign that a methods should be split.

src/garage/torch/algos/pearl.py
```diff
@@ -1,6 +1,635 @@
...
+        self._reward_scale = reward_scale
+        self._update_post_train = update_post_train
+        self._eval_deterministic = eval_deterministic
+        self.evaluator = None
+        self._task_idx = None
+
+        encoder_spec = self.get_env_spec(env[0](), latent_dim, 'encoder')
+        encoder_in_dim = int(np.prod(encoder_spec.input_space.shape))
+        encoder_out_dim = int(np.prod(encoder_spec.output_space.shape))
+        context_encoder = encoder_class(input_dim=encoder_in_dim,
```
ryanjulian: sorry i wasn't clear -- my intention here was that you could use `InOutSpec` here instead of `input_dim` and `output_dim`.

examples/torch/pearl_ml1_push.py
ryanjulian: ...or just don't pass them if they are the same as the default

src/garage/envs/env_spec.py
ryanjulian: ah -- please instead file an issue about making these immutable

examples/torch/pearl_ml1_push.py
lywong92: wrap_experiment throws an error if I don't pass ctxt and seed. It calls another method, and these attributes don't have default values.

src/garage/torch/algos/pearl.py
lywong92: The encoder_class inherits from `MLPModule`, so I need to pass in input_dim and output_dim to construct it.

src/garage/torch/algos/pearl.py
```diff
@@ -1,6 +1,635 @@
...
+            out_dim = latent_dim
+        in_space = akro.Box(low=-1, high=1, shape=(in_dim, ), dtype=np.float32)
+        out_space = akro.Box(low=-1,
+                             high=1,
+                             shape=(out_dim, ),
+                             dtype=np.float32)
+        if module == 'encoder':
+            spec = InOutSpec(in_space, out_space)
+        elif module == 'vf':
+            spec = EnvSpec(in_space, out_space)
```
ryanjulian: is `in_space` for the vf not just `env_spec.observation_space`?

src/garage/torch/algos/pearl.py
```diff
@@ -1,6 +1,635 @@
...
+
+        """
+        obs_dim = int(np.prod(env_spec.observation_space.shape))
+        action_dim = int(np.prod(env_spec.action_space.shape))
+        if module == 'encoder':
+            in_dim = obs_dim + action_dim + 1
+            out_dim = latent_dim * 2
+        elif module == 'vf':
+            in_dim = obs_dim
+            out_dim = latent_dim
```
ryanjulian: reading the paper, isn't the prototype for the VF `r = V(s,z)`? i.e. `in_space = concat(o, z)` and `out_space = Box(1,)`?

examples/torch/pearl_ml1_push.py
```diff
@@ -0,0 +1,135 @@
...
+
+    # instantiate networks
+    net_size = params['net_size']
+
+    augmented_env = PEARL.augment_env_spec(env[0](), params['latent_size'])
+    qf = ContinuousMLPQFunction(env_spec=augmented_env,
+                                hidden_sizes=[net_size, net_size, net_size])
+
+    vf_env = PEARL.get_env_spec(env[0](), params['latent_size'], 'vf')
+    vf = ContinuousMLPQFunction(env_spec=vf_env,
```
ryanjulian: since the VF maps `r ~= V(o,z)` it can't be a Q-function (which has prototype `R ~= Q(s,a))`.

I understand why you used the hack you did, but in the future this should be a primitive in`g.torch.value_functions` module. Can you leave an issue explaining this?

src/garage/torch/algos/pearl.py
ryanjulian: @krzentner WDYT?

examples/torch/pearl_ml1_push.py
ryanjulian: @krzentner can you help here?

krzentner: Yeah, `wrap_experiment` requires explicitly listing the parameters. That could be changed, but I haven't added support for it yet.

You shouldn't need to pass `seed`, but `ctxt` is required.

examples/torch/pearl_ml1_push.py
krzentner: You should just be able to pass `ctxt` instead of `snapshot_config`. They implement the same fields.

src/garage/torch/algos/pearl.py
krzentner: I think this is fine. It's a reasonable place to have this logic.

examples/torch/pearl_ml1_push.py
lywong92: If I don't pass in `seed`, I get this error:

```
File "/home/lindawong/Desktop/github/garage/src/garage/sampler/worker.py", line 140, in worker_init
    deterministic.set_seed(self._seed + self._worker_number)
TypeError: unsupported operand type(s) for +: 'NoneType' and 'int
```'

Comment (krzentner) [reply examples/torch/pearl_ml1_push.py]: Oh, did you still call `set_seed` below?

File: src/garage/torch/algos/pearl.py (hunk 17 above)
Comment (lywong92) [reply src/garage/torch/algos/pearl.py]: `LocalRunner` needs access to `policy`.

File: src/garage/torch/algos/pearl.py (hunk 16 above)
Comment (lywong92) [reply src/garage/torch/algos/pearl.py]: I will make this private.

File: examples/torch/pearl_ml1_push.py (hunk 24 above)
Comment (lywong92) [reply examples/torch/pearl_ml1_push.py]: Okay, will do.

File: examples/torch/pearl_ml1_push.py [hunk 25]
```diff
@@ -0,0 +1,153 @@
...
+from garage.torch.algos import PEARL
+from garage.torch.algos.pearl import PEARLWorker
+from garage.torch.embeddings import MLPEncoder
+from garage.torch.policies import (ContextConditionedPolicy,
+                                   TanhGaussianMLPPolicy)
+from garage.torch.q_functions import ContinuousMLPQFunction
+import garage.torch.utils as tu
+
+
+# pylint: disable=unused-argument, dangerous-default-value
```
krzentner: Wait, which arg doesn't get used?

lywong92: I removed the arguments.

examples/torch/pearl_ml1_push.py
lywong92: No. I got the error above when I removed `seed` from the argument list. I removed `set_seed` as well.

lywong92: There's probably a bug in `wrap_experiment`. When other examples call `wrap_experiment`, they also pass in `seed`.

examples/torch/pearl_ml1_push.py
```diff
@@ -0,0 +1,148 @@
...
+from garage.torch.algos import PEARL
+from garage.torch.algos.pearl import PEARLWorker
+from garage.torch.embeddings import MLPEncoder
+from garage.torch.policies import (ContextConditionedPolicy,
+                                   TanhGaussianMLPPolicy)
+from garage.torch.q_functions import ContinuousMLPQFunction
+import garage.torch.utils as tu
+
+
+# pylint: disable=dangerous-default-value
```
ryanjulian: i think you can just use a tuple instead and this will go away?

src/garage/torch/algos/pearl.py
ryanjulian: can you implement this as a property instead? that way it is immutable

```python
def __init__(self, ...):
    self._policy = policy_class(...)

@property
def policy(self):
    """garage.torch.Policy: Most up-to-date policy."""
    return self._policy
```

tests/integration_tests/test_examples.py
```diff
@@ -23,7 +23,8 @@
     EXAMPLES_ROOT_DIR / 'torch/maml_ppo_half_cheetah_dir.py',
     EXAMPLES_ROOT_DIR / 'torch/maml_trpo_half_cheetah_dir.py',
     EXAMPLES_ROOT_DIR / 'torch/maml_vpg_half_cheetah_dir.py',
-    EXAMPLES_ROOT_DIR / 'torch/maml_trpo_ml10.py'
+    EXAMPLES_ROOT_DIR / 'torch/maml_trpo_ml10.py',
```
ryanjulian: i think your example tests should run automagically by virtue of being in `examples/` --- they only need to go here if they require special arguments

@gitanshu can explain

ryanjulian:
Don't forget to squash and write the commit message before merging!

PR #1220, opened 2020-03-14
Skip mujoco dependent tests for external PRs

gitanshu:


tests/garage/envs/dm_control/test_dm_control_env.py
```diff
@@ -10,6 +10,8 @@
 
 
 class TestDmControlEnv:
+
```
ryanjulian: can you mark the class, for your own convenience?

tests/garage/envs/dm_control/test_dm_control_tf_policy.py
```diff
@@ -11,6 +12,7 @@
 
 class TestDmControlTfPolicy(TfGraphTestCase):
 
+    @pytest.mark.mujoco
```
ryanjulian: can you mark the class or file?

Issue #1223, opened 2020-03-15
Separate TaskSampler instantiation from LocalRunner in MetaEvaluator


ahtsan:
Currently, LocalRunner instantiates TaskSampler and introduces circular reference between Algorithm, MetaEvaluator and LocalRunner. This is a blocking issue for meta-RL algorithms.

PR #1227, opened 2020-03-17
MetaEval changes from paper

krzentner:
Included is this is tests and everything needed to use the MetaEvaluator
as a field on an algorithm.

### DIFF
--- a/src/garage/experiment/meta_evaluator.py
+++ b/src/garage/experiment/meta_evaluator.py
@@ -1,16 +1,18 @@
 """Evaluator which tests Meta-RL algorithms on test environments."""
-from garage import log_performance, TrajectoryBatch
+
+from dowel import logger, tabular
+
+from garage import log_multitask_performance, TrajectoryBatch
+from garage.experiment.deterministic import get_seed
 from garage.sampler import LocalSampler
+from garage.sampler.worker import DefaultWorker
+from garage.sampler.worker_factory import WorkerFactory
 
 
 class MetaEvaluator:
     """Evaluates Meta-RL algorithms on test environments.
 
     Args:
-        runner (garage.experiment.LocalRunner): A runner capable of running
-            policies from the (meta) algorithm. Can be the same runner used by
-            the algorithm. Does not use runner.obtain_samples, and so does not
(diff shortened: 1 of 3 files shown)


src/garage/_functions.py
```diff
@@ -34,10 +34,14 @@ def log_multitask_performance(itr, batch, discount, name_map=None):
             task_name = trajectory.env_infos['task_name'][0]
         except KeyError:
             try:
+                task_id = None
```
krzentner: I think Keren has an alternative change to this logic, which looks a little better.
I just needed the `MetaEvaluator` to work on environments where all tasks are unnamed / continuous.

krzentner: Yeah, when Keren merges PR #1214 I'll remove the logic here.

ahtsan:
This will close #1223

src/garage/experiment/meta_evaluator.py
```diff
@@ -22,40 +24,61 @@ class MetaEvaluator:
...
         adapted_trajectories = []
+        logger.log('Sampling for adapation and meta-testing...')
+        if self._test_sampler is None:
+            self._test_sampler = LocalSampler.from_worker_factory(
+                WorkerFactory(seed=get_seed(),
+                              max_path_length=self._max_path_length,
+                              n_workers=1,
+                              worker_class=DefaultWorker),
+                agents=algo.get_exploration_policy(),
+                envs=self._test_task_sampler.sample(1))
```
ahtsan: Should it be 
```
...
envs=self._test_task_sampler.sample(self._n_test_tasks))
```
?

krzentner: Nope. The meta evaluator goes through one task at a time. See the next line of code.

ahtsan: Since in `meta_squash` branch it is, just want to make sure it's the correct implementation here

krzentner: Hmm. I'm fairly certain that part of the `meta_squash` branch is wrong, but thanks for checking.

PR #1231, opened 2020-03-20
RL2 meta-test

ahtsan:
* Fix RL2 exploration policy and adapted policy for correctness
* Add MetaEvaluator and n_epoch_per_eval inside RL2


examples/tf/rl2_ppo_halfcheetah_meta_test.py
```diff
@@ -0,0 +1,84 @@
+"""Example script to run RL2PPO meta test in HalfCheetah."""
+from garage import wrap_experiment
+from garage.envs.half_cheetah_vel_env import HalfCheetahVelEnv
```
ryanjulian: please use canonical imports here and below, e.g.`from garage.envs import HalfCheetahVelEnv` if it exists

src/garage/tf/algos/rl2.py
```diff
@@ -107,8 +116,63 @@ def get_exploration_policy(self):
                 used for meta-RL adaptation.
 
         """
-        return self._policy
+        self._policy.reset()
 
+        class NoResetPolicy:
```
ryanjulian: can you define this outside RL2 (but still in this file?)

Also, can you move the inner classes to the top of the file, before the RL2 class?

ahtsan: Do you mean like `RL2Env` and `RL2Worker` as well? or just `NoResetPolicy` and `RL2AdaptedPolicy`

ryanjulian: I mean that RL2 should be the last class defined in the file. This allows travelers to read the file from top to bottom without missing anything.

examples/tf/rl2_ppo_halfcheetah.py
```diff
@@ -73,4 +77,8 @@ def rl2_ppo_halfcheetah(ctxt=None, seed=1):
                      meta_batch_size)
 
 
-rl2_ppo_halfcheetah(seed=1)
+rl2_ppo_halfcheetah(seed=1,
```
ryanjulian: i don't think you have to pass many of these args because they have defaults?

examples/tf/rl2_ppo_halfcheetah_meta_test.py
```diff
@@ -0,0 +1,92 @@
...
+                     n_workers=meta_batch_size,
+                     worker_class=RL2Worker)
+
+        runner.train(n_epochs=n_epochs,
+                     batch_size=episode_per_task * max_path_length *
+                     meta_batch_size)
+
+
+rl2_ppo_halfcheetah_meta_test(seed=1,
+                              max_path_length=100,
```
ryanjulian: same comment on using default args

setup.cfg
```diff
@@ -18,7 +18,7 @@ extend-ignore =
     F841  # Unused variables are checked by pylint
 
 [tool:pytest]
-addopts = -n auto -rfE -s --strict-markers
+addopts = -rfE -s --strict-markers
```
ryanjulian: please remember to remove this once you are satisfied -- if you don't use auto your tests won't run fast enough.

tests/integration_tests/test_examples.py
```diff
@@ -151,3 +152,36 @@ def test_pearl_ml1_push():
...
                           check=False).returncode == 0
+
+
+@pytest.mark.no_cover
+@pytest.mark.timeout(400)
+def test_rl2_ppo_halfcheetah():
+    """Test rl2_ppo_halfcheetah.py."""
+    assert subprocess.run([
+        EXAMPLES_ROOT_DIR / 'tf/rl2_ppo_halfcheetah.py', '--n_epochs', '1',
+        '--episode_per_task', '1', '--meta_batch_size', '1'
```
naeioi: `wrap_experiment` does not handle console arguments automatically. You need to use `click`.

tests/garage/torch/algos/test_pearl.py
```diff
@@ -21,24 +22,25 @@
 class TestPEARL:
     """Test class for PEARL."""
 
+    @pytest.mark.large
```
naeioi: @lywong92 Given that you have an example `pearl_ml1_push.py`, I think this test here is doing the same thing, so it is not necessary. Am I correct?

ryanjulian:
Please consult with @gitanshu about examples tests before merging

tests/garage/torch/algos/test_pearl.py
gitanshu: We don't count examples towards test coverage though, so keep that in mind

tests/integration_tests/test_examples.py
ahtsan: fixed

tests/garage/torch/algos/test_pearl.py
ahtsan: @naeioi I think `pearl_ml1_push.py` is added to `LONG_RUNNING_EXAMPLES`

ryanjulian:
Please don't merge this until @gitanshu approves your example tests.

ahtsan:
@gitanshu PTAL

tests/integration_tests/test_examples.py
```diff
@@ -16,15 +16,16 @@
 ]
 
 LONG_RUNNING_EXAMPLES = [
-    EXAMPLES_ROOT_DIR / 'tf/ppo_memorize_digits.py',
-    EXAMPLES_ROOT_DIR / 'tf/dqn_pong.py',
-    EXAMPLES_ROOT_DIR / 'tf/trpo_cubecrash.py',
+    EXAMPLES_ROOT_DIR / 'tf/ppo_memorize_digits.py', EXAMPLES_ROOT_DIR /
```
gitanshu: @ahtsan the formatting looks weird here. I'm approving the PR but please fix this before merging.

ahtsan: Thanks. Fixed

src/garage/tf/algos/rl2.py
```diff
@@ -18,6 +18,262 @@
 from garage.tf.algos._rl2npo import RL2NPO
 
 
+class RL2Env(gym.Wrapper):
+    """Environment wrapper for RL2.
+
+    In RL2, observation is concatenated with previous action,
+    reward and terminal signal to form new observation.
+
+    Also, different tasks could have different observation dimension.
```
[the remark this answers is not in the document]
ahtsan: Fixed

tests/garage/experiment/test_deterministic.py
```diff
@@ -25,7 +25,7 @@ def test_deterministic_pytorch():
 def test_deterministic_tensorflow():
     """Test deterministic behavior of Tensorflow"""
     deterministic.set_seed(77)
```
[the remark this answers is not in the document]
ahtsan: Will do

PR #1232, opened 2020-03-20
Add RL2 support and examples for ML10/45

ahtsan:


ahtsan:
Should I add the tests and mark it as nightly? It requires Metaworld.

examples/tf/rl2_ppo_ml10.py
```diff
@@ -0,0 +1,81 @@
...
+    Args:
+        ctxt (garage.experiment.ExperimentContext): The experiment
+            configuration used by LocalRunner to create the snapshotter.
+        seed (int): Used to seed the random number generator to produce
+            determinism.
+
+    """
+    set_seed(seed)
+    with LocalTFRunner(snapshot_config=ctxt) as runner:
+        max_path_length = 100
```
ryanjulian: can you make these args of the experiment function instead? they can still have defaults.

examples/tf/rl2_ppo_ml45.py
```diff
@@ -0,0 +1,87 @@
...
+    Args:
+        ctxt (garage.experiment.ExperimentContext): The experiment
+            configuration used by LocalRunner to create the snapshotter.
+        seed (int): Used to seed the random number generator to produce
+            determinism.
+
+    """
+    set_seed(seed)
+    with LocalTFRunner(snapshot_config=ctxt) as runner:
+        max_path_length = 100
```
ryanjulian: can you make these arguments of the experiment function instead?

ryanjulian: FYI this is usually 150 for metaworld, but I guess go with whatever you used for ICML

src/garage/tf/algos/rl2.py
```diff
@@ -360,6 +356,19 @@ def spec(self):
         """
         return self._spec
 
+    def _pad_zeros(self, obs):
```
ryanjulian: does this not exist in `g.misc.tensor_utils`?

ryanjulian: https://github.com/rlworkgroup/garage/blob/master/src/garage/misc/tensor_utils.py#L91

ryanjulian:
Your examples should be automatically tested because they are in the examples folder. You can check in the Travis logs.

src/garage/tf/algos/rl2.py
```diff
@@ -185,9 +332,18 @@ def _process_samples(self, itr, paths):
             np_tensor_utils.stack_and_pad_tensor_dict_list(
                 concatenated_paths, self._inner_algo.max_path_length))
 
-        undiscounted_returns = log_performance(
-            itr, TrajectoryBatch.from_trajectory_list(self._env_spec, paths),
-            self._inner_algo.discount)
+        name_map = None
+        if hasattr(self._task_sampler._envs[0].env, 'all_task_names'):
```
ahtsan: @ryanjulian @krzentner Does this part look ok to you?

examples/tf/rl2_ppo_ml10.py
ahtsan: fixed

examples/tf/rl2_ppo_ml45.py
ahtsan: fixed

src/garage/tf/algos/rl2.py
ahtsan: fixed

examples/tf/rl2_ppo_ml45.py
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env python3
+"""Example script to run RL2 in ML45."""
+# pylint: disable=wrong-import-order
```
ryanjulian: why is this needed?

src/garage/tf/algos/rl2.py
```diff
@@ -10,14 +10,160 @@
...
+        obs_flat_dim = np.prod(env.observation_space.shape)
+        action_flat_dim = np.prod(env.action_space.shape)
+        if self._max_obs_dim is not None:
+            obs_flat_dim = self._max_obs_dim
+        return akro.Box(low=-np.inf,
+                        high=np.inf,
+                        shape=(obs_flat_dim + action_flat_dim + 1 + 1, ))
+
+    # pylint: disable=arguments-differ
+    def reset(self):
```
ryanjulian: not sure why this can't have the same signature as the parent class?

examples/tf/rl2_ppo_ml45.py
ahtsan: pylint and flake8 fight with each other

src/garage/tf/algos/rl2.py
ryanjulian: hmm would it be simpler to just check if it's from metaworld? Do we expect anyone else to be implementing that interface?

ryanjulian: I'm not terribly offended by this, but I think we should file an issue about cleaning it up. It seems there's a missing interface here for multitask, e.g. something like this could go into a multitask version of GarageEnv.

naeioi: I like the idea of a multitask version of GarageEnv. So far in docstring, I write `GarageEnv` but what I truly want is a `GarageMultiTaskEnv`.

src/garage/tf/algos/rl2.py
```diff
@@ -185,9 +332,18 @@ def _process_samples(self, itr, paths):
...
 
-        undiscounted_returns = log_performance(
-            itr, TrajectoryBatch.from_trajectory_list(self._env_spec, paths),
-            self._inner_algo.discount)
+        name_map = None
+        if hasattr(self._task_sampler._envs[0].env, 'all_task_names'):
+            names = [
+                env.env.all_task_names[0] for env in self._task_sampler._envs
+            ]
+            name_map = dict(zip(names, names))
```
naeioi: `name_map` does not change over the training. Can it can precomputed in `__init__()`?

naeioi:
You might also want to configure rl2 examples manually in `test_examples.py`, so it does not take too long to run.

ryanjulian:
Please consult with @gitanshu about examples tests before merging

src/garage/tf/algos/rl2.py
krzentner: This also isn't actually the correct type for `name_map`. It should map `int -> str` (i.e. `dict(enumerate(names))`).

PR #1233, opened 2020-03-20
Update baselines to latest master

gitanshu:


ryanjulian:
fixes #1233

gitanshu:
@avnishn @yonghyuc can you take care of this?

setup.py
```diff
@@ -57,7 +57,7 @@
 # Development dependencies (*not* included in 'all')
 EXTRAS['dev'] = [
     # Please keep alphabetized
-    'baselines @ https://api.github.com/repos/openai/baselines/tarball/f2729693253c0ef4d4086231d36e0a4307ec1cb3',  # noqa: E501
+    'baselines @ https://api.github.com/repos/openai/baselines/tarball/ea25b9e8b234e6ee1bca43083f8f3cf974143998',  # noqa: E501
```
ryanjulian: don't forget to use the updated style, which uses `GARAGE_GH_TOKEN` so we don't run afoul of GitHub API limits.

yonghyuc:
You also need to change `test_benchmark_trpo.py`

I check the code and I think you can change the code calling `trpo_mpi` like this.
```
    policy_network = 'mlp'
    trpo_mpi.learn(network=policy_network,
                   env=env,
                   total_timesteps=hyper_parameters['batch_size'] *
                                   hyper_parameters['n_epochs'],
                   timesteps_per_batch=hyper_parameters['batch_size'],
                   gamma=hyper_parameters['discount'],
                   lam=hyper_parameters['gae_lambda'],
                   max_kl=hyper_parameters['max_kl'],
                   cg_iters=10,
                   cg_damping=0.1,
                   vf_iters=5,
                   vf_stepsize=1e-3)
```

docker/Dockerfile.base.16.04
```diff
@@ -101,7 +101,8 @@ RUN pip install --upgrade pip
 # In this step only the presence of the file mjkey.txt is required, so we only
 # create an empty file
 RUN touch /root/.mujoco/mjkey.txt && \
-  pip install .[all,dev] && \
+  pip install .[all] && \
```
ryanjulian, line 104: please only change things which absolutely need to be changed to make this work. AFAICT we only need to do this in `ci-verify-conda` and nowhere else.

krzentner: This change is also necessary, or baselines will fail to install because tensorflow wasn't installed first. I'm going to copy it to my change.

krzentner:
I think this change should no longer be needed, now that I've merged #1282.

PR #1234, opened 2020-03-21
Add meta-testing to MAML

naeioi:


examples/torch/maml_ppo_half_cheetah_dir.py
```diff
@@ -48,6 +50,14 @@ def maml_ppo(ctxt, seed, epochs, rollouts_per_task, meta_batch_size):
 
     max_path_length = 100
 
+    tasks = SetTaskSampler(lambda: GarageEnv(
+        normalize(HalfCheetahDirEnv(), expected_action_scale=10.)))
```
ahtsan: is `expected_action_scale` necessary `HalfCheetahDirEnv`, or just for `MAML`?

src/garage/torch/algos/maml.py
```diff
@@ -137,6 +143,9 @@ def train_once(self, runner, all_samples, all_params):
                                                   kl_after.item(),
                                                   policy_entropy.mean().item())
 
+        if self._meta_evaluator:
+            self._meta_evaluator.evaluate(self)
```
ahtsan: Should it evaluate every n epochs? Evaluate in each epoch seems to introduce a huge overhead, especially when running large benchmarks like ML45.

naeioi: How do you suggest to handle dowel's caveat of not supporting dynamic tabular headers?

ahtsan: There will be ONE warning at the beginning. After talking to KR, we think it's fine by now.

examples/torch/maml_ppo_half_cheetah_dir.py
```diff
@@ -9,6 +9,8 @@
 from garage.envs.base import GarageEnv
 from garage.experiment import LocalRunner
 from garage.experiment.deterministic import set_seed
+from garage.experiment.meta_evaluator import MetaEvaluator
```
ryanjulian: nit: which of these can these use canonical imports? `from garage.experiment import MetaEvaluator` etc.

examples/torch/maml_trpo_half_cheetah_dir.py
```diff
@@ -9,6 +9,8 @@
 from garage.envs.base import GarageEnv
 from garage.experiment import LocalRunner
 from garage.experiment.deterministic import set_seed
+from garage.experiment.meta_evaluator import MetaEvaluator
```
ryanjulian: nit: canonical imports

examples/torch/maml_trpo_ml10.py
```diff
@@ -10,6 +10,8 @@
 from garage.envs.base import GarageEnv
 from garage.experiment import LocalRunner
 from garage.experiment.deterministic import set_seed
+from garage.experiment.meta_evaluator import MetaEvaluator
+from garage.experiment.task_sampler import EnvPoolSampler
```
ryanjulian: nit: canonical imports

tests/integration_tests/test_examples.py
```diff
@@ -151,3 +151,48 @@ def test_pearl_ml1_push():
         '--embedding_mini_batch_size', '2', '--max_path_length', '1'
     ],
                           check=False).returncode == 0
+
+
+@pytest.mark.nightly
+@pytest.mark.no_cover
+@pytest.mark.timeout(120)
+def test_maml_ml10():
```
ryanjulian: please have @gitanshu review this

adding new special test cases is generally discouraged. most example tests should be self-contained and run by virtue of living in examples/

examples/torch/maml_ppo_half_cheetah_dir.py
```diff
@@ -48,6 +50,14 @@ def maml_ppo(ctxt, seed, epochs, rollouts_per_task, meta_batch_size):
 
     max_path_length = 100
 
+    tasks = SetTaskSampler(lambda: GarageEnv(
```
ryanjulian: i think this name is a little misleading? how about `task_sampler`

examples/torch/maml_trpo_half_cheetah_dir.py
```diff
@@ -48,6 +50,14 @@ def maml_trpo(ctxt, seed, epochs, rollouts_per_task, meta_batch_size):
 
     max_path_length = 100
 
+    tasks = SetTaskSampler(lambda: GarageEnv(
```
ryanjulian: `task_sampler`?

examples/torch/maml_vpg_half_cheetah_dir.py
```diff
@@ -1,25 +1,42 @@
...
+import click
 import torch
 
+from garage import wrap_experiment
 from garage.envs import HalfCheetahDirEnv, normalize
 from garage.envs.base import GarageEnv
-from garage.experiment import LocalRunner, run_experiment
+from garage.experiment import LocalRunner
+from garage.experiment.deterministic import set_seed
+from garage.experiment.meta_evaluator import MetaEvaluator
```
ryanjulian: nit: canonical imports

examples/torch/maml_vpg_half_cheetah_dir.py
```diff
@@ -31,22 +48,31 @@ def run_task(snapshot_config, *_):
 
     baseline = LinearFeatureBaseline(env_spec=env.spec)
 
-    rollouts_per_task = 20
     max_path_length = 100
 
-    runner = LocalRunner(snapshot_config)
+    tasks = SetTaskSampler(lambda: GarageEnv(
```
ryanjulian: nit: `task_sampler`?

src/garage/torch/algos/maml.py
ryanjulian: I suggest a param similar to `evaluate_every_n_epochs` or whichever param name is used in the off-policy algos

tests/garage/torch/algos/test_maml.py
```diff
@@ -0,0 +1,97 @@
...
+        self.env.close()
+
+    def test_get_exploration_policy(self, set_params, test_params):
+        """Test if an independent copy of policy is returned."""
+        self.policy.apply(partial(set_params, 0.1))
+        adapt_policy = self.algo.get_exploration_policy()
+        adapt_policy.apply(partial(set_params, 0.2))
+
+        # Old policy should remain untouched
+        self.policy.apply(partial(test_params, 0.1))
```
ryanjulian: is this really the intended use of fixtures? can you make a more explicit loop/assert in the test body instead?

tests/garage/torch/algos/test_maml.py
```diff
@@ -0,0 +1,97 @@
...
+from garage.sampler import LocalSampler, WorkerFactory
+from garage.torch.algos import MAMLPPO
+from garage.torch.policies import GaussianMLPPolicy
+
+
+@pytest.fixture
+def set_params():  # noqa: D202
+    """Set parameters of a module."""
+
+    def _set_params(v, m):
```
ryanjulian: do these really need to be fixtures as opposed to just helper functions? pytest fixtures usually represent a resource, mocked class, or are used for some kind of metaprogramming,

examples/torch/maml_ppo_half_cheetah_dir.py
naeioi: I used `expected_action_scale=10.` for both `HalfCheetah` and Metaworld when I did the performance evaluation of MAML.

tests/integration_tests/test_examples.py
naeioi: I need to also set `meta_batch_size` to 1 to reduce time. Do you have an idea to achieve this without making a special test case?

tests/garage/torch/algos/test_maml.py
naeioi: Fixture removed.

naeioi: But I am still using `pytorch.nn.policy` because it is more succinct.

tests/integration_tests/test_examples.py
gitanshu: How long does it take to run 1 epoch without any other parameter change?

tests/integration_tests/test_examples.py
```diff
@@ -151,3 +147,48 @@ def test_pearl_ml1_push():
...
+    """Test maml_trpo_ml10.py"""
+    assert subprocess.run([
+        EXAMPLES_ROOT_DIR / 'torch/maml_trpo_ml10.py', '--epochs', '1',
```
gitanshu: We use this code to make it run for only 1 epoch
```
env = os.environ.copy()
env['GARAGE_EXAMPLE_TEST_N_EPOCHS'] = '1'
```
This is in line with the philosophy of having to modify examples as least as possible for the sake of tests

tests/integration_tests/test_examples.py
naeioi: 105s on my local machine.

Issue #1229, opened 2020-03-18
Change shape of Independent object in Gaussian MLP Module


yonghyuc:
Currently, the log likelihood from the GaussianMLPModule is single value that sum every log likelihood from each observation and action. However, for PPO and VPG, each individual log likelihood value is needed to calculate advantage value.

This is shape of log likelihood that we need for PPO and VPG

![image (1)](https://user-images.githubusercontent.com/43084978/76919035-f4c8e600-6884-11ea-8e0e-0f93a9847925.png)

But currently, the output of log likelihood is single value.

![image](https://user-images.githubusercontent.com/43084978/76919044-f98d9a00-6884-11ea-9468-eae57478b40b.png)

ryanjulian:
It seems to me this could be a mix-up of how batches/broadcasting is handled?

If I send a batch of samples to log_prob I should get a batch of LLs. This should apply not just to LL, but any other feedforward properties of the policy (e.g. entropy)

avnishn:
Closed by #1235

Issue #807, opened 2019-07-18
torch/Soft Actor Critic


avnishn:

ryanjulian:
You should probably collaborate with @lywong92 because this shares a lot of code with #804

avnishn:
Closed by #1235

PR #1235, opened 2020-03-21
Soft Actor Critic and Applications

avnishn:
This commit adds SAC to garage. The version of
SAC added however is built for collecting batches
of transitions between optimization steps, and not
single transitions, as the original SAC does.

See #807

Furthermore small issues are addressed by this PR such
as #1229 and expanding the robustness for datatypes for
the TanhGaussianMLP Policy.

### DIFF
--- a/src/garage/torch/algos/sac.py
+++ b/src/garage/torch/algos/sac.py
@@ -1,0 +1,414 @@
+"""This modules creates a sac model in PyTorch."""
+from collections import deque
+import copy
+
+from dowel import tabular
+import numpy as np
+import torch
+import torch.nn.functional as F
+
+from garage import log_performance
+from garage.np.algos.off_policy_rl_algorithm import OffPolicyRLAlgorithm
+import garage.torch.utils as tu
+
+
+class SAC(OffPolicyRLAlgorithm):
+    """A SAC Model in Torch.
+
+    Based on Soft Actor-Critic and Applications:
+        https://arxiv.org/abs/1812.05905
+
+    Soft Actor-Critic (SAC) is an algorithm which optimizes a stochastic
+    policy in an off-policy way, forming a bridge between stochastic policy
+    optimization and DDPG-style approaches.
+    A central feature of SAC is entropy regularization. The policy is trained
+    to maximize a trade-off between expected return and entropy, a measure of
+    randomness in the policy. This has a close connection to the
+    exploration-exploitation trade-off: increasing entropy results in more
+    exploration, which can accelerate learning later on. It can also prevent
+    the policy from prematurely converging to a bad local optimum.
+
+    Args:
+        policy(garage.torch.policy): Policy/Actor/Agent that is being optimized
+            by SAC.
+        qf1(garage.torch.q_function): QFunction/Critic used for actor/policy
+            optimization. See Soft Actor-Critic and Applications.
+        qf2(garage.torch.q_function): QFunction/Critic used for actor/policy
+            optimization. See Soft Actor-Critic and Applications.
+        replay_buffer(garage.replay_buffer): Stores transitions that
+            are previously collected by the sampler.
+        env_spec(garage.envs.env_spec.EnvSpec): The env_spec attribute of the
+            environment that the agent is being trained in. Usually accessable
+            by calling env.spec.
+        max_path_length(int): Max path length of the environment.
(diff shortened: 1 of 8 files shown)


src/garage/torch/algos/sac.py
```diff
@@ -0,0 +1,414 @@
...
+                self.episode_rewards.append(np.mean(path_returns))
+                for _ in range(self._gradient_steps):
+                    policy_loss, qf1_loss, qf2_loss = self.train_once()
+            last_return = log_performance(runner.step_itr,
+                                          self._obtain_evaluation_samples(
+                                              runner.get_env_copy(),
+                                              num_trajs=10),
+                                          discount=self.discount)
+            self._log_statistics(policy_loss, qf1_loss, qf2_loss)
+            tabular.record('TotalEnvSteps', runner.total_env_steps)
```
krzentner: I think the runner takes care of this now.

src/garage/torch/algos/sac.py
```diff
@@ -0,0 +1,414 @@
...
+        qf_lr(float): learning rate for q_function optimizers.
+        reward_scale (float): reward scale. Changing this hyperparameter
+            changes the effect that the reward from a transition will have
+            during optimization.
+        optimizer(torch.optim): optimizer to be used for policy/actor,
+            q_functions/critics, and temperature/entropy optimizations.
+        steps_per_epoch (int): Number of train_once calls per epoch.
+
+    """
+
```
krzentner: You should override the `sampler_cls` from `OffPolicyRLAlgorithm`, since it will crash if used.
The user can still override what you choose here using `runner.setup`.

src/garage/torch/algos/sac.py
```diff
@@ -0,0 +1,414 @@
...
+                                              runner.get_env_copy(),
+                                              num_trajs=10),
+                                          discount=self.discount)
+            self._log_statistics(policy_loss, qf1_loss, qf2_loss)
+            tabular.record('TotalEnvSteps', runner.total_env_steps)
+            runner.step_itr += 1
+
+        return np.mean(last_return)
+
+    def train_once(self, itr=None, paths=None):
```
krzentner: Can you just call this method something else (maybe `_train_once`), and remove the arguments? You're overloading `train()` anyways.

src/garage/torch/algos/sac.py
```diff
@@ -0,0 +1,414 @@
...
+        return policy_loss, qf1_loss, qf2_loss
+
+    def _get_log_alpha(self, **kwargs):
+        """Return the value of log_alpha.
+
+        This function exists in case there are versions of sac that need
+        access to a modified log_alpha, such as multi_task sac.
+
+        Args:
+            kwargs(dict): keyword args that can be used in retrieving the
```
krzentner: kwargs never get passed, so can you just delete it? Is it going to be used somewhere?

src/garage/torch/algos/sac.py
```diff
@@ -0,0 +1,414 @@
...
+
+        Args:
+            device (str): ID of GPU or CPU.
+
+        """
+        if device is None:
+            device = tu.global_device()
+        for net in self.networks:
+            net.to(device)
+        self.log_alpha = torch.Tensor([self._initial_log_entropy
```
krzentner: Why `_initial_log_entropy` instead of the current `log_alpha`?

src/garage/torch/algos/sac.py
```diff
@@ -0,0 +1,414 @@
...
+            min_buffer_size=int(1e4),
+            target_update_tau=5e-3,
+            policy_lr=3e-4,
+            qf_lr=3e-4,
+            reward_scale=1.0,
+            optimizer=torch.optim.Adam,
+            steps_per_epoch=1,
+    ):
+
+        self.policy = policy
```
lywong92: Can you make these attributes private?

src/garage/torch/algos/sac.py
avnishn: its going to be used in MTSAC.

src/garage/torch/algos/sac.py
avnishn: yes, sure thing!

src/garage/torch/algos/sac.py
krzentner: Ah, I see. That makes sense.

Issue #734, opened 2019-06-20
*Conv* models should accept uint8(images) and normalize


ahtsan:
Convolution network should accept placeholder with type uint8 (images) but garage is not. Example: https://github.com/rlworkgroup/garage/blob/master/src/garage/tf/policies/categorical_cnn_policy.py#L104

It should also normalize the images.

ahtsan:
Check out the remote branch `/fix_npo_conv` as a reference.

ahtsan:
Issue #794 and #795 should cover this bug, since it makes more sense that sampler should be responsible to normalize the inputs if inputs are images. In that case, policies do not need to normalize the inputs and will receive normalized images, with type `float32`. Therefore closing this issue.

ahtsan:
Policies should be responsible to normalize pixel inputs. Reopened.

PR #1236, opened 2020-03-21
Normalize images in CategoricalCNNPolicy

maliesa96:
Normalizes pixel values and accepts uint8 inputs (addresses [#734](https://github.com/rlworkgroup/garage/issues/734))

### DIFF
--- a/src/garage/envs/base.py
+++ b/src/garage/envs/base.py
@@ -41,10 +41,13 @@ class GarageEnv(gym.Wrapper):
         env_name (str): If the env_name is speficied, a gym environment
             with that name will be created. If such an environment does not
             exist, a `gym.error` is thrown.
+        is_image (bool): True if observations contain pixel values,
+            false otherwise. Setting this to true converts a gym.Spaces.Box
+            obs space to an akro.Image and normalizes pixel values.
 
     """
 
-    def __init__(self, env=None, env_name=''):
+    def __init__(self, env=None, env_name='', is_image=False):
         # Needed for deserialization
         self._env_name = env_name
         self._env = env
@@ -55,7 +58,8 @@ class GarageEnv(gym.Wrapper):
             super().__init__(env)
 
         self.action_space = akro.from_gym(self.env.action_space)
-        self.observation_space = akro.from_gym(self.env.observation_space)
+        self.observation_space = akro.from_gym(self.env.observation_space,
+                                               is_image=is_image)
         self.__spec = EnvSpec(action_space=self.action_space,
                               observation_space=self.observation_space)
 
--- a/src/garage/misc/tensor_utils.py
+++ b/src/garage/misc/tensor_utils.py
@@ -1,5 +1,4 @@
 """Utiliy functions for tensors."""
-import gym.spaces
 import numpy as np
 import scipy.signal
 
@@ -275,27 +274,21 @@ def truncate_tensor_dict(tensor_dict, truncated_len):
     return ret
 
 
-def normalize_pixel_batch(env_spec, observations):
+def normalize_pixel_batch(observations):
     """Normalize the observations (images).
 
-    If the observations appear to be flattened, attempts to unflatten them.
-
-    If the input are images, it normalized into range [0, 1].
+    Normalize pixel values to be between [0, 1].
 
     Args:
-        env_spec (garage.envs.EnvSpec): Environment specification.
         observations (numpy.ndarray): Observations from environment.
+            obses should be unflattened and should contain pixel
+            values.
 
     Returns:
         numpy.ndarray: Normalized observations.
 
     """
-    if isinstance(env_spec.observation_space, gym.spaces.Box):
-        if len(observations.shape) == 2:
-            observations = env_spec.observation_space.unflatten_n(observations)
-        if len(env_spec.observation_space.shape) == 3:
-            return [obs.astype(np.float32) / 255.0 for obs in observations]
-    return observations
+    return [obs.astype(np.float32) / 255.0 for obs in observations]
 
 
 def slice_nested_dict(dict_or_array, start, stop):
--- a/src/garage/sampler/off_policy_vectorized_sampler.py
+++ b/src/garage/sampler/off_policy_vectorized_sampler.py
@@ -7,6 +7,7 @@ It diffs from OnPolicyVectorizedSampler in two parts:
  from replay buffer, which only has buffer_batch_size.
  - It needs to add transitions to replay buffer throughout the rollout.
 """
+
 import itertools
 import pickle
 
@@ -93,15 +94,15 @@ class OffPolicyVectorizedSampler(BatchSampler):
 
         while n_samples < batch_size:
             policy.reset(completes)
-            input_obses = self.algo.env_spec.observation_space.flatten_n(obses)
-            obs_normalized = tensor_utils.normalize_pixel_batch(
-                self._env_spec, input_obses)
+            obs_space = self.algo.env_spec.observation_space
+            input_obses = obs_space.flatten_n(obses)
+
             if self.algo.es:
                 actions, agent_infos = self.algo.es.get_actions(
-                    itr, obs_normalized, self.algo.policy)
+                    itr, input_obses, self.algo.policy)
             else:
                 actions, agent_infos = self.algo.policy.get_actions(
-                    obs_normalized)
+                    input_obses)
 
             next_obses, rewards, dones, env_infos = \
                 self._vec_env.step(actions)
--- a/src/garage/tf/algos/dqn.py
+++ b/src/garage/tf/algos/dqn.py
@@ -1,9 +1,9 @@
 """Deep Q-Learning Network algorithm."""
+import akro
 from dowel import tabular
 import numpy as np
 import tensorflow as tf
 
-from garage.misc.tensor_utils import normalize_pixel_batch
 from garage.np.algos.off_policy_rl_algorithm import OffPolicyRLAlgorithm
 from garage.tf.misc import tensor_utils
 
@@ -234,12 +234,13 @@ class DQN(OffPolicyRLAlgorithm):
         next_observations = transitions['next_observation']
         dones = transitions['terminal']
 
-        # normalize pixel to range [0, 1] since the samples stored in the
-        # replay buffer are of type uint8 and not normalized, for memory
-        # optimization
-        observations = normalize_pixel_batch(self.env_spec, observations)
-        next_observations = normalize_pixel_batch(self.env_spec,
-                                                  next_observations)
+        if isinstance(self.env_spec.observation_space, akro.Image):
+            if len(self.env_spec.observation_space.shape) < 3:
+                observations = self.env_spec.observation_space.unflatten_n(
+                    observations)
+                next_observations = self.env_spec.observation_space.\
+                    unflatten_n(next_observations)
+
         loss, _ = self._train_qf(observations, actions, rewards, dones,
                                  next_observations)
 
--- a/src/garage/tf/baselines/gaussian_cnn_baseline.py
+++ b/src/garage/tf/baselines/gaussian_cnn_baseline.py
@@ -1,7 +1,8 @@
 """Gaussian CNN Baseline."""
-
+import akro
 import numpy as np
 
+from garage.misc.tensor_utils import normalize_pixel_batch
 from garage.np.baselines import Baseline
 from garage.tf.regressors import GaussianCNNRegressor
 
@@ -27,6 +28,16 @@ class GaussianCNNBaseline(Baseline):
             regressor_args=None,
             name='GaussianCNNBaseline',
     ):
+        if not isinstance(env_spec.observation_space, akro.Box) or \
+                not len(env_spec.observation_space.shape) in (2, 3):
+            raise ValueError(
+                '{} can only process 2D, 3D akro.Image or'
+                ' akro.Box observations, but received an env_spec with '
+                'observation_space of type {} and shape {}'.format(
+                    type(self).__name__,
+                    type(env_spec.observation_space).__name__,
+                    env_spec.observation_space.shape))
+
         super().__init__(env_spec)
         if regressor_args is None:
             regressor_args = dict()
@@ -38,6 +49,7 @@ class GaussianCNNBaseline(Baseline):
             name=name,
             **regressor_args)
         self.name = name
+        self.env_spec = env_spec
 
     def fit(self, paths):
         """Fit regressor based on paths.
@@ -47,6 +59,9 @@ class GaussianCNNBaseline(Baseline):
 
         """
         observations = np.concatenate([p['observations'] for p in paths])
+        if isinstance(self.env_spec.observation_space, akro.Image):
+            observations = normalize_pixel_batch(observations)
+
         returns = np.concatenate([p['returns'] for p in paths])
         self._regressor.fit(observations, returns.reshape((-1, 1)))
 
@@ -60,7 +75,11 @@ class GaussianCNNBaseline(Baseline):
             numpy.ndarray: Predicted value.
 
         """
-        return self._regressor.predict(path['observations']).flatten()
+        observations = path['observations']
+        if isinstance(self.env_spec.observation_space, akro.Image):
+            observations = normalize_pixel_batch(observations)
+
+        return self._regressor.predict(observations).flatten()
 
     def get_param_values(self):
         """Get parameter values.
--- a/src/garage/tf/misc/tensor_utils.py
+++ b/src/garage/tf/misc/tensor_utils.py
@@ -1,11 +1,27 @@
-from collections import Iterable
+"""Tensor utility functions for tensorflow."""
 from collections import namedtuple
+from collections.abc import Iterable
 
 import numpy as np
 import tensorflow as tf
 
 
+# pylint: disable=unused-argument
+# yapf: disable
+# pylint: disable=missing-return-doc, missing-return-type-doc
 def compile_function(inputs, outputs, log_name=None):
+    """Compiles a tensorflow function using the current session.
+
+    Args:
+        inputs (list[tf.Tensor]): Inputs to the function. Can be a list of
+            inputs or just one.
+        outputs (list[tf.Tensor]): Outputs of the function. Can be a list of
+            outputs or just one.
+        log_name (string): Name of the function. This is None by default.
+
+    Returns:
+        function: Compiled tensorflow function.
+    """
 
     def run(*input_vals):
         sess = tf.compat.v1.get_default_session()
@@ -14,9 +30,13 @@ def compile_function(inputs, outputs, log_name=None):
     return run
 
 
+# yapf: enable
+# pylint: enable=missing-return-doc, missing-return-type-doc
+# pylint: enable=unused-argument
+
+
 def get_target_ops(variables, target_variables, tau=None):
-    """
-    Get target variables update operations.
+    """Get target variables update operations.
 
     In RL algorithms we often update target network every n
     steps. This function returns the tf.Operation for updating
@@ -27,10 +47,13 @@ def get_target_ops(variables, target_variables, tau=None):
 
     Args:
         variables (list[tf.Variable]): Soure variables for update.
-        target_variable (list[tf.Variable]): Target variables to
+        target_variables (list[tf.Variable]): Target variables to
             be updated.
         tau (float): Fraction to update. Set it to be None for
             hard-update.
+
+    Returns:
+        tf.Operation: Operation for updating the target variables.
     """
     update_ops = []
     init_ops = []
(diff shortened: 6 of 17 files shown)


src/garage/tf/misc/tensor_utils.py
```diff
@@ -63,6 +64,14 @@ def filter_valids(t, valid, name='filter_valids'):
         t, tf.cast(tf.round(valid), tf.int32), 2, name=name)[1]
 
 
+def normalize_pixel_observation(env_spec, obs):
```
ahtsan: I believe we already have this, check out `garage.misc.tensor_utils.normalize_pixel_batch`

maliesa96: Yep though that takes in a batch of observations and returns a list of normalized observations since it's used by the sampler. This function just accepts one observation. Although `garage.misc.tensor_utils.normalize_pixel_batch` can be rewritten to call `normalize_pixel_observation` in a loop.

ahtsan: I see. Indeed, the difference is probably that one deals with python array while one deals with tf.Tensor.

ahtsan:
Could you write a test for the function you added?

maliesa96:
Done.

src/garage/tf/misc/tensor_utils.py
ahtsan: Would be great if you can provide more information in the docstring. Please refer to [this link](https://github.com/rlworkgroup/garage/blob/master/CONTRIBUTING.md) for guide.

maliesa96: Thanks, updated. I also followed the docstring format in `garage.misc.tensor_utils.normalize_pixel_batch`.

src/garage/tf/misc/tensor_utils.py
```diff
@@ -62,6 +63,25 @@ def filter_valids(t, valid, name='filter_valids'):
...
+    Args:
+        env_spec (garage.envs.env_spec.EnvSpec): Environment specification.
+        obs (numpy.ndarray): Observation to be normalized from environment.
+
+    Returns:
+        Tf.Tensor: Normalized tensor with same shape as obs.
+    """
+    
+    if isinstance(env_spec.observation_space, gym.spaces.Box):
+        if len(env_spec.observation_space.shape) == 3:
```
ryanjulian: what happens if I have a Box(3,) non-pixel observation space?

src/garage/tf/policies/categorical_cnn_policy.py
```diff
@@ -101,9 +102,9 @@ def __init__(self,
         self._initialize()
 
     def _initialize(self):
-        state_input = tf.compat.v1.placeholder(tf.float32,
+        state_input = tf.compat.v1.placeholder(tf.uint8,
                                                shape=(None, ) + self.obs_dim)
-
+        state_input = normalize_pixel_observation(self.env_spec, state_input)
```
ryanjulian: since the only valid input to this primitive is an image, you should probably verify it's an image here, rather than writing a conditional into the utility function

src/garage/tf/misc/tensor_utils.py
```diff
@@ -62,6 +63,25 @@ def filter_valids(t, valid, name='filter_valids'):
     return tf.dynamic_partition(
         t, tf.cast(tf.round(valid), tf.int32), 2, name=name)[1]
 
+def normalize_pixel_observation(env_spec, obs):
```
ryanjulian: there's a common convention that functions which *might* do something should be called `maybe_foo` rather than just foo.

with that said, see my comments below on why this function shouldn't be checking for input types anyway

tests/garage/tf/misc/test_tensor_utils.py
```diff
@@ -35,6 +38,22 @@ def test_compute_advantages(self):
 
         assert np.array_equal(adv, desired_val)
 
+    def test_normalize_pixel_observation(self):
```
ryanjulian: you will also need to test that CategoricalCNNPolicy is doing the right think here...

src/garage/tf/policies/categorical_cnn_policy.py
maliesa96: Is there a foolproof way of checking whether observations are images, since they are simply tensors? I was going off what was done in `garage.misc.tensor_utils.normalize_pixel_batch` (which also just checks the shape of the Box). I can check that the box boundaries are 255 all around which means they are likely pixels, though that also does not guarantee it. The other option is to a pass an optional flag to the policy that indicates whether or not the observations are images.

ryanjulian: there isn't, and that's one of the difficult parts of this PR.

without adding an `Image` type to akro (which i think is a good idea), there isn't a whole lot which can be done in this PR specifically to guarantee these types are safe.

however, what we *can* do is decide whose job type safety is. since CNN policies can't infer image types, once data reaches them, it must be too late to do anything about it. so it can't be the responsibility of the CNN policies to enforce type safety. similarly, helper functions couldn't possibly do this without types, either (which is the issue I raise here).

if we raise ourselves on level in the call tree, we will find ourselves in the algorithm. that means it must be the algorithm's job to enforce type safety -- it can find out whether the observation is an image by interrogating the environment class, which it actually has access to.

to summarize, there are two good options here. which you use is up to you.

1) Add an Image type to [akro](https://github.com/rlworkgroup/akro) (a package we maintain), get me or @krzentner to cut akro release 0.0.7, bump garage's akro version to 0.0.7, then ensure `GarageEnv` converts the EnvSpec of image-based environments to use the new `Image` type rather than `Box`)
2) Add a flag to the algorithm alerting it that it should expect an image input, check the observation space (inside the algorithm) to ensure it's consistent with an image, refuse to use CNN-based primitives unless this flag is turned on.

(1) Is the long-term solution, but there's something to be said for doing (2) first since it would allow you to close the issue faster.

maliesa96: Option 1 sounds cleaner to me. I'll create a PR in the akro repo for the new image type.

maliesa96:
Updated to use akro v0.0.7. Here are the major changes:

- ` normalize_pixel_observation` and `normalize_pixel_batch` no longer check the observation type. This is done by the callers by checking to see if the observation space is an `akro.Image` instance.

- Added a test to make sure `CategoricalCNNPolicy` isn't trying to normalize observations that aren't images.

- added an `is_image` flag to `GarageEnv`and `TfEnv` that gets passed to `akro.from_gym`.

ryanjulian:
@maliesa96 you appear to have an unclean branch now, as a results of (probably) using `git merge` or `git pull`, which we don't use in this repo.

You can solve this by using `git rebase -i master` and then squashing all but your head commit. In the future, don't use `git merge`, `git pull` or the "Update branch" button on GitHub. All of those generate merge commits, which we don't allow on master.

For more info and links to guides on how to use our git worfklow, check out the Git section of CONTRIBUTING.md.

ryanjulian:
I will review this as soon as you've updated the PR -- I'd like to review it now, but there are just too many touched files for me to figure out what to pay attention to.

maliesa96:
Weird, I had already done a rebase and squashed the merge commits.  I had two commits that I just squashed into one, did that fix it?

maliesa96:
Looks okay now I believe. My mistake, I should've used my master branch to track rlworkgroup/master and a feature branch for my changes.

ryanjulian:
Your branch still have conflicts which need resolving, but it looks much better now.

I understand that most people aren't used to working with multiple remotes...

src/garage/envs/base.py
```diff
@@ -41,10 +41,13 @@ class GarageEnv(gym.Wrapper):
         env_name (str): If the env_name is speficied, a gym environment
             with that name will be created. If such an environment does not
             exist, a `gym.error` is thrown.
+        is_image (bool): True if observations contain pixel values,
+            false otherwise. Setting this to true causes pixel values
+            to be normalized.
```
ryanjulian, line 46: doesn't it also cause the observation space to be cast to akro.Image?

src/garage/tf/policies/categorical_cnn_policy.py
```diff
@@ -101,9 +99,12 @@ def __init__(self,
         self._initialize()
 
     def _initialize(self):
-        state_input = tf.compat.v1.placeholder(tf.float32,
+        state_input = tf.compat.v1.placeholder(tf.uint8,
                                                shape=(None, ) + self.obs_dim)
 
+        if isinstance(self.env_spec.observation_space, akro.Image):
```
ryanjulian, line 105: a CategoricalCNNPolicy can only accept akro.Images, so I think you should remove these conditionals and instead add a guard in the constructor

```py
if not isinstance(env_spec.observation_space, akro.Image):
    raise ValueError('CategoricalCNNPolicy can only process akro.Image observations, but received an env_spec with observation_space of type {}'
    .format(type(env_spec.observation_space).__name__))
```

ryanjulian:
Your branch still have conflicts which need resolving, but it looks much better now.

I understand that most people aren't used to working with multiple remotes...

ryanjulian:
Note that you should be adding these checks to other CNN-based primitives, in both torch and tf:

tf/GaussianCNNBaseline
tf/DiscreteCNNQFunction
tf/GaussianCNNRegressor

There may be others. Any CNN primitives which accepts an EnvSpec should be checking this.

src/garage/tf/algos/dqn.py
```diff
@@ -240,9 +241,11 @@ def optimize_policy(self, itr, samples_data):
         # normalize pixel to range [0, 1] since the samples stored in the
         # replay buffer are of type uint8 and not normalized, for memory
         # optimization
-        observations = normalize_pixel_batch(self.env_spec, observations)
-        next_observations = normalize_pixel_batch(self.env_spec,
-                                                  next_observations)
+
+        if isinstance(self.env_spec.observation_space, akro.Image):
```
ryanjulian, line 245: DQN can only use Image observations, so your code should actually just be raising an exception when you get a non-Image observation space.

src/garage/tf/policies/categorical_cnn_policy.py
maliesa96: Does this apply to all CNN primitives? I was under the impression that CNNs can be used on non-image data as well.

ryanjulian: You can apply a CNNs to arbitrary 2D data, which of course doesn't have to be an image (but usually are).

However, for the purposes of type-safety, CNNs operate the type `Image` (i.e. arbitrary 2D data). If, for instance, someone wants their CNNs to operate on sound data, they could rearrange that data into a 2D grid, which we call an image.

maliesa96: Shouldn't non-image 2D data be classified as a Box? Using your sound example, if one were to input sound data as an `akro.Image`,  `CategoricalCNNPolicy` will normalize this data by dividing by 255 thinking it contains pixel values.

maliesa96: On second thought this might not be an issue since an akro.Image can only contain values between 0 and 255, though it would mean non-image data would have to be mapped to that range as well.

ahtsan: How about CNN could take either akro.Image or akro.Box, and only normalize when it's akro.image?

maliesa96: I agree @ahtsan. How does that sound @ryanjulian?

ryanjulian: My preference is to enforce a 1:1 mapping between types and networks (i.e. CNNs take images). The purpose of this change was to simplify the type relationships in this code by eliminating `Box` types, since they are hard to validate, not introduce a new, complicated set of cases.

Perhaps the definition of `akro.Image` is too restrictive, and we should provision for float-valued images or int-valued images, but that adds complexity which also has to be justified by identifying a real use case, not a hypothetical one.

In practical terms, if someone wanted to use a CNN with float-valued data (for instance), converting that data to uint8 has shown to have a small-to-negligible impact on performance in NNs. If this isn't sufficient for them, there's nothing stopping them from writing their own CategoricalCNNPolicy based largely on the existing one.

As for what to actually do, I think accepting `Box` is fine, as long as you validate that it's a 2D or 3D `Box`. The 2D box is assumed to be shaped `(W, H)` and the 3D box is assumed to be shaped `(W, H, C)`.

ryanjulian: @krzentner i'm curious to hear your thoughts

krzentner: I would also lean towards accepting `Box`, since many people will have environments that don't use the `Image` type, but they would expect garage to handle.

If they chose to use a CNN, their intent is fairly clear anyways.

ryanjulian: Alright -- so we can merge this when all the CNN types validate that the input is either a 2D/3D Box or an Image.

maliesa96: @ryanjulian sounds good. I assume they should also normalize images?

src/garage/sampler/off_policy_vectorized_sampler.py
```diff
@@ -93,15 +95,19 @@ def obtain_samples(self, itr, batch_size=None, whole_paths=True):
...
-            input_obses = self.algo.env_spec.observation_space.flatten_n(obses)
-            obs_normalized = tensor_utils.normalize_pixel_batch(
-                self._env_spec, input_obses)
+            obs_space = self.algo.env_spec.observation_space
+
+            if not isinstance(obs_space, akro.Image):
+                input_obses = obs_space.flatten_n(obses)
+            else:
+                input_obses = obses
+
```
maliesa96, line 104: @krzentner is this the expected behavior? Just want to make sure this change doesn't break anything.

maliesa96:
Updated. Summary of changes:

- GaussianCNNBaseline, CategoricalCNNPolicy, and DiscreteCNNQFunction all only accept `akro.Image` or 2D/3D `akro.Box` spaces.

- These primitives all normalize `akro.Image` observations internally, meaning OffPolicyVectorizedSampler and DQN no longer normalize pixel values themselves.

src/garage/sampler/off_policy_vectorized_sampler.py
krzentner: I think it's better to fix this in the policy. (i.e. assume that inputs into the policy *might* be flattened already).

maliesa96: Fixed. The sampler now flattens all inputs, and CNN policies unflatten them.

ryanjulian:
@maliesa96 let us know when this is ready

src/garage/tf/algos/dqn.py
```diff
@@ -236,10 +242,12 @@ def optimize_policy(self, itr, samples_data):
...
-        next_observations = normalize_pixel_batch(self.env_spec,
-                                                  next_observations)
+        # optimization. DiscreteCNNQFunction already does this internally.
+
+        if not isinstance(self.qf, DiscreteCNNQFunction):
+            observations = normalize_pixel_batch(observations)
+            next_observations = normalize_pixel_batch(next_observations)
+
         loss, _ = self._train_qf(observations, actions, rewards, dones,
                                  next_observations)
```
maliesa96, line 252: @ryanjulian does this part look okay? the idea is to avoid normalizing twice and only normalize if the Q function doesn't do so internally (which at the moment is only done by DiscreteCNNQFunction). Alternatively I can make every Q function normalize akro.Images, though I'm not sure that's always desirable. 

If this looks good then it's ready to be merged.

ryanjulian: this is hacky -- i would prefer if you make it consistent across the codebase. 

an argument can be made either way about whether the QFunction or the algorithm should be normalizing.

without doing a grand review of API design, we should just go with the decision which creates the last amount of change. in this case, that seems that it would be to remove normalization from DiscreteCNNQFunction.

ryanjulian:
note that if your CI/GitHub checks are not passing, I'm not going to do legwork for you to finish it up and merge.

please let me know when this is ready-to-merge (i.e. everything is green)

src/garage/tf/algos/dqn.py
ryanjulian: note that #734 actually tells you what to do here -- the models (FooCNNBar) should be doing the normalizing, not the algorithms.

src/garage/tf/algos/dqn.py
maliesa96: About this, this would actually break `examples/dqn_cartpole`. The observations in that example are Boxes relating to the position of the cart and pole, not RGB pixels.

maliesa96:
Yep I know. The builds were failing due to code that isn't mine (missing docstrings in files I modified), I wasn't sure if that meant I'd have to make those fixes. I'll make the changes and push.

ryanjulian:
Our policy is that if you a touch a file which needs updated due to stronger standards (e.g. docstrings, style), it's your responsibility to update it.

We usually only relax this when a PR touches a ton of files all at once.

src/garage/tf/baselines/gaussian_cnn_baseline.py
```diff
@@ -27,6 +28,15 @@ def __init__(
             regressor_args=None,
             name='GaussianCNNBaseline',
     ):
+        assert isinstance(env_spec.observation_space, akro.Box) and \
```
ryanjulian, line 31: an assert is not the right way to go here. that will crash the program. asserts are only for testing conditions which absolutely should not be logically possible unless something terrible has happened, not for validating user input.

Here you just want a simple `if` and raise a ValueError

ryanjulian, line 31: wait -- what if I get an `akro.Image`? is that not allowed?

src/garage/tf/misc/tensor_utils.py
```diff
@@ -59,8 +60,22 @@ def flatten_batch_dict(d, name=None):
...
     # Must round before cast to prevent floating-error
-    return tf.dynamic_partition(
-        t, tf.cast(tf.round(valid), tf.int32), 2, name=name)[1]
+    return tf.dynamic_partition(t,
+                                tf.cast(tf.round(valid), tf.int32),
+                                2,
+                                name=name)[1]
+
+
+def normalize_pixel_observation(obs):
```
ryanjulian, line 69: is this really a necessary helper function? its implementation is 1-line.

src/garage/tf/q_functions/discrete_cnn_q_function.py
```diff
@@ -86,6 +88,15 @@ def __init__(self,
                  output_b_init=tf.zeros_initializer(),
                  dueling=False,
                  layer_normalization=False):
+        assert isinstance(env_spec.observation_space, akro.Box) and \
```
ryanjulian, line 91: use a ValueError

src/garage/tf/policies/categorical_cnn_policy.py
```diff
@@ -76,6 +74,16 @@ def __init__(self,
         assert isinstance(env_spec.action_space, akro.Discrete), (
             'CategoricalCNNPolicy only works with akro.Discrete action '
             'space.')
+
+        assert isinstance(env_spec.observation_space, akro.Box) and \
```
ryanjulian, line 78: raise a ValueError

src/garage/tf/baselines/gaussian_cnn_baseline.py
maliesa96: akro.Image is a subclass of akro.Box, so that'll work fine.

ryanjulian: ah, thanks for pointing that out.

maliesa96:
Note that the Travis builds won't pass without merging #1313.

src/garage/misc/tensor_utils.py
```diff
@@ -275,23 +274,19 @@ def truncate_tensor_dict(tensor_dict, truncated_len):
     return ret
 
 
-def normalize_pixel_batch(env_spec, observations):
+def normalize_pixel_batch(observations):
```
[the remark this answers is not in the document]
ryanjulian: @maliesa96 is this resolved?

src/garage/tf/policies/categorical_cnn_policy.py
ryanjulian: err TypeError, i mean

src/garage/misc/tensor_utils.py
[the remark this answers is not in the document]
maliesa96: Yup.

maliesa96:
@ryanjulian Travis builds pass but won't run the mujoco tests because external PR containers don't include a mujoco key (explains the low code coverage). Does codecov not report the patch and project statuses when the PR is below the codecov threshold?

ryanjulian:
@Mergifyio rebase

ryanjulian:
@maliesa96 can you rebase this branch and ping me when the CI goes green? then I can merge it.

ryanjulian:
@maliesa96 never mind, i think i can do it myself

ryanjulian:
@Mergifyio rebase

ryanjulian:
@Mergifyio rebase

PR #1241, opened 2020-03-23
Add more PEARL examples

lywong92:


examples/torch/pearl_half_cheetah_vel.py
```diff
@@ -0,0 +1,146 @@
+#!/usr/bin/env python3
+"""PEARL HalfCheetahVel example."""
+
+from garage.envs import GarageEnv, normalize
+from garage.envs.half_cheetah_vel_env import HalfCheetahVelEnv
+from garage.experiment import LocalRunner, wrap_experiment
+from garage.experiment.deterministic import set_seed
+from garage.experiment.meta_evaluator import MetaEvaluator
```
ryanjulian: nit: which of these can use a canonical import, e.g. `from garage.experiment import MetaEvaluator`?

tests/integration_tests/test_examples.py
```diff
@@ -24,7 +24,8 @@
     EXAMPLES_ROOT_DIR / 'torch/maml_trpo_half_cheetah_dir.py',
     EXAMPLES_ROOT_DIR / 'torch/maml_vpg_half_cheetah_dir.py',
     EXAMPLES_ROOT_DIR / 'torch/maml_trpo_ml10.py',
-    EXAMPLES_ROOT_DIR / 'torch/pearl_ml1_push.py'
+    EXAMPLES_ROOT_DIR / 'torch/pearl_half_cheetah_vel.py',
```
ryanjulian: hmm is there a particular reason these are long-running? do they take a long time to setup?

usually example tests should run only for 1 epoch (the actual example can run for longer).

@gitanshu can help here

ryanjulian:
please consult with Gitanshu about the examples tests before submitting

tests/integration_tests/test_examples.py
gitanshu: Also if you're adding a test here because it takes too long to run, you'd have to add new test for it

examples/torch/pearl_half_cheetah_vel.py
lywong92: Only `LocalRunner` and `wrap_experiment`.

tests/integration_tests/test_examples.py
lywong92: GPU isn't available so these tests are taking a long time to run.

examples/torch/pearl_half_cheetah_vel.py
ryanjulian: I think MetaEvaluator is also available

@krzentner is there a reason you didn't export these TaskSamplers?

tests/integration_tests/test_examples.py
ryanjulian: does it take a huge time to run 1 iteration or many?

examples/torch/pearl_half_cheetah_vel.py
krzentner: Not in particular. It seemed likely we would move them at a later date.

ryanjulian: @lywong92 can you import them in `g/experiment/__init__.py` and update the imports for them to be canonical? 

examples are where most people will be getting their usage examples.

ryanjulian:
I'm waiting for @gitanshu 's approval for this one

PR #1249, opened 2020-03-24
Wrap exp name parameters

krzentner:
This change gives wrap_experiment an option that inserts parameters (and their values) into the name of the experiment directory, which is very convenient for keeping track of multiple experiments.

### DIFF
--- a/src/garage/experiment/experiment.py
+++ b/src/garage/experiment/experiment.py
@@ -1,6 +1,7 @@
 """Tools for running experiments with Garage."""
 # flake8: noqa
 import base64
+import collections
 import datetime
 import enum
 import functools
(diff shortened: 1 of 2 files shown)


src/garage/experiment/experiment.py
```diff
@@ -339,20 +340,26 @@ def my_experiment(ctxt, seed, lr=0.5):
         archive_launch_repo (bool): Whether to save an archive of the
             repository containing the launcher script. This is a potentially
             expensive operation which is useful for ensuring reproducibility.
+        name_parameters (str or None): Parameters to insert into the experiment
+            name. Should be either None (the default), 'all' (all parameters
+            will be used), or 'passed' (only passed parameters will be used).
+            The used parameters will be sorted, and inserted with their values
+            into the experiment name.
```
krzentner: Main documentation of the new parameter is here.

krzentner: And actually is slightly wrong (the parameters aren't sorted, they're used in the order of the function definition). I'll fix the doc in a minute.

src/garage/experiment/experiment.py
```diff
@@ -504,7 +549,11 @@ def my_experiment(ctxt, seed, lr=0.5):
         archive_launch_repo (bool): Whether to save an archive of the
             repository containing the launcher script. This is a potentially
             expensive operation which is useful for ensuring reproducibility.
-
+        name_parameters (str or None): Parameters to insert into the experiment
+            name. Should be either None (the default), 'all' (all parameters
+            will be used), or 'passed' (only passed parameters will be used).
+            The used parameters will be inserted in the order they appear in
+            the function definition.
```
krzentner: Main documentation is here.

PR #1250, opened 2020-03-24
Use a GH token in the CI to retrieve packages

ryanjulian:
Recently the CI has been crippled by rate-limiting from GitHub, which
limits API requests from unauthenticated IP addresses to 60/hour. This
has broken our ability to retrieve source tarballs for non-PyPI
packages from Travis, which can schedule many jobs onto the same IP
address. We have never hit this rate limit in the past, and the root
cause of the behavior change is unclear (e.g. it could be recent Travis
incidents causing many jobs to repeatedly requests resources as part of
restart or recovery, a policy change by GitHub, or a change in Travis
scheduling strategies, etc.).

This PR allows setup.py to use an optional GitHub Personal Access token
to authenticate downloads. This is only active if the environment
variable `GARAGE_GH_TOKEN` is present, and otherwise uses the token
`'git'`, which signifies an unauthenicated request when using the GitHub
API.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ jobs:
     - name: "Large tests"
       env: JOB_RUN_CMD="make ci-job-large"
     - if: type != pull_request OR head_repo = "rlworkgroup/garage"
-      name: "Mujoco based tests"
+      name: "MuJoCo-based tests"
       env:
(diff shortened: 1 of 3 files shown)


ryanjulian:

PR #1257, opened 2020-03-26
Remove MuJoCo key for normal and large test jobs

gitanshu:
to make sure that no mujoco based tests slip into
either normal or large test jobs, since both of
these jobs are run for external PRs where mujoco
key is not present.

* Exclude flaky tests from MuJoCo test job


gitanshu:
@mergifyio rebase

ryanjulian:
There may be a simple way of ensuring you're running in a container -- checking now

PR #1258, opened 2020-03-26
Fix experiment 'name' scope

gagankhandate:
Fix leaves `name` in scope for use [here](https://github.com/rlworkgroup/garage/blob/52049c330995b442ab3efaf312cbc789071a1990/src/garage/experiment/experiment.py#L419)

@zhanpenghe @krzentner


gagankhandate:
Great. Thanks KR & ZP!

PR #1263, opened 2020-03-30
Add ML1 example for RL2

ahtsan:


tests/integration_tests/test_examples.py
```diff
@@ -238,6 +239,18 @@ def test_rl2_ppo_halfcheetah():
                           check=False).returncode == 0
 
 
+@pytest.mark.mujoco
```
ryanjulian: why does this need to be in the manual test cases rather than automatically-collected above?

@gitanshu PTAL

ahtsan: I wanted to give it smaller parameters to save time. It's ok to just leave it automatically-collected as well.

ryanjulian: i think that n_epochs should always be 1 in tests?

I'm fine with it as long as @gitanshu is.

ahtsan:
@krzentner @gitanshu PTAL

tests/integration_tests/test_examples.py
krzentner: Yeah, automatically-collected examples are run so that the runner only runs one epoch, so I don't think this should really be necessary.

PR #1265, opened 2020-03-31
Add ML10 and ML45 Pearl examples

lywong92:


examples/torch/pearl_ml10.py
```diff
@@ -0,0 +1,169 @@
...
+                 n_workers=1,
+                 worker_class=PEARLWorker)
+
+    worker_args = dict(deterministic=True, accum_context=True)
+    meta_evaluator = MetaEvaluator(test_task_sampler=test_env_sampler,
+                                   max_path_length=max_path_length,
+                                   worker_class=PEARLWorker,
+                                   worker_args=worker_args,
+                                   n_test_tasks=num_test_tasks)
+    pearl.evaluator = meta_evaluator
```
krzentner: Can you move the construction of the meta evaluator into the PEARL algorithm?

src/garage/torch/algos/pearl.py
```diff
@@ -225,7 +231,7 @@ def __getstate__(self):
         data = self.__dict__.copy()
         del data['_replay_buffers']
         del data['_context_replay_buffers']
-        del data['evaluator']
+        del data['_evaluator']
```
krzentner, line 234: I think you should no longer need this.

PR #1270, opened 2020-04-01
Fix TF plotter

ahtsan:
Currently, the main thread and plotter thread share the same policy object, which might break
because of shape mismatch, e.g. in most cases main thread performs vectorized sampling
while plotter thread samples from a single environment. This PR basically clone the policy
for the plotter to avoid the mentioned issue.

### DIFF
--- a/src/garage/tf/models/base.py
+++ b/src/garage/tf/models/base.py
@@ -43,7 +43,7 @@ class BaseModel(abc.ABC):
             name (str): Name of the model.
 
         Return:
-            object: Output(s) of the model.
+            list[tf.Tensor]: Output(s) of the model.
 
         """
 
@@ -60,7 +60,7 @@ class BaseModel(abc.ABC):
         function model.
 
         Returns:
-            list[object]: Parameters.
+            list[tf.Tensor]: Parameters.
 
         """
 
@@ -69,7 +69,7 @@ class BaseModel(abc.ABC):
         """Set parameters of the Model.
 
         Args:
-            parameters (list[object]): Parameters.
+            parameters (list[tf.Tensor]): Parameters.
 
         """
 
--- a/src/garage/tf/models/base.py
+++ b/src/garage/tf/models/base.py
@@ -7,8 +7,7 @@ import tensorflow as tf
 
 
 class BaseModel(abc.ABC):
-    """
-    Interface-only abstract class for models.
+    """Interface-only abstract class for models.
 
     A Model contains the structure/configuration of a set of computation
     graphs, or can be understood as a set of networks. Using a model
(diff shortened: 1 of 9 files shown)


src/garage/tf/plotter/plotter.py
```diff
@@ -53,11 +53,11 @@ def __init__(self,
                  rollout=default_rollout):
         Plotter.__plotters.append(self)
         self._env = env
-        self._policy = policy
         self.sess = tf.compat.v1.Session() if sess is None else sess
-
         self.graph = tf.compat.v1.get_default_graph(
         ) if graph is None else graph
+        with self.sess.as_default(), self.sess.graph.as_default():
+            self._policy = policy.clone('plotter_policy')
```
zequnyu, line 60: Can this be done inside `LocalRunner`?

ahtsan: Not really, this policy has to be cloned and built (automatically during construction) under the session which the plotter would use, so it needs to be done here.

ahtsan: Unless we pass the current default tf.Session, then we can also pass a "cloned" policy.

zequnyu: Using default tf session will break the plotter (not this tf one), so maybe not.

src/garage/tf/models/base.py
```diff
@@ -32,78 +31,105 @@ class BaseModel(abc.ABC):
...
         implement `build()` directly.
 
         Args:
-            inputs: Tensor input(s) for the model.
+            inputs (object): Input(s) for the model.
+            name (str): Name of the model.
 
         Return:
-            output: Tensor output(s) of the model.
+            object: Output(s) of the model.
```
ryanjulian, line 46: Isn't this some number of tf.Tensors?

src/garage/tf/models/base.py
```diff
@@ -32,78 +31,105 @@ class BaseModel(abc.ABC):
...
-        """
-        Parameters of the Model.
+        """Parameters of the Model.
 
         The output of a model is determined by its parameter. It could be
         the weights of a neural network model or parameters of a loss
         function model.
+
+        Returns:
+            list[object]: Parameters.
```
ryanjulian, line 63: aren't these tf.Tensors?

src/garage/tf/models/base.py
```diff
@@ -32,78 +31,105 @@ class BaseModel(abc.ABC):
...
-        pass
 
     @parameters.setter
     def parameters(self, parameters):
-        """Set parameters of the Model."""
-        pass
+        """Set parameters of the Model.
+
+        Args:
+            parameters (list[object]): Parameters.
```
ryanjulian, line 72: aren't these tf.Tensors?

src/garage/tf/plotter/plotter.py
```diff
@@ -53,11 +53,11 @@ def __init__(self,
                  rollout=default_rollout):
         Plotter.__plotters.append(self)
         self._env = env
-        self._policy = policy
         self.sess = tf.compat.v1.Session() if sess is None else sess
-
         self.graph = tf.compat.v1.get_default_graph(
         ) if graph is None else graph
+        with self.sess.as_default(), self.sess.graph.as_default():
```
ryanjulian, line 59: based on the line immediately above, shouldn't this be `self.graph.as_default()`?

src/garage/tf/policies/categorical_lstm_policy.py
```diff
@@ -155,11 +165,28 @@ def _initialize(self):
 
     @property
     def vectorized(self):
-        """Vectorized or not."""
+        """Vectorized or not.
+
+        Returns:
+            Bool: True if primitive supports vectorized operations.
```
ryanjulian, line 171: `bool`

src/garage/tf/policies/categorical_lstm_policy.py
```diff
@@ -155,11 +165,28 @@ def _initialize(self):
...
+        """
         return True
 
     def dist_info_sym(self, obs_var, state_info_vars, name=None):
-        """Symbolic graph of the distribution."""
+        """Build a symbolic graph of the distribution parameters.
+
+        Args:
+            obs_var (tf.Tensor): Tensor input for symbolic graph.
+            state_info_vars (dict): Extra state information, e.g.
```
ryanjulian, line 181: dict of what?

src/garage/tf/policies/categorical_lstm_policy.py
```diff
@@ -224,31 +279,89 @@ def get_actions(self, observations):
 
     @property
     def recurrent(self):
-        """Recurrent or not."""
+        """Recurrent or not.
+
+        Returns:
+            Bool: True if policy is recurrent.
```
ryanjulian, line 285: bool

src/garage/tf/policies/categorical_lstm_policy.py
```diff
@@ -88,12 +89,21 @@ def __init__(self,
...
         self._state_include_action = state_include_action
         self._output_nonlinearity = output_nonlinearity
         self._output_w_init = output_w_init
         self._output_b_init = output_b_init
         self._hidden_state_init = hidden_state_init
+        self._hidden_state_init_trainable = hidden_state_init_trainable
         self._cell_state_init = cell_state_init
+        self._cell_stat_init_trainable = cell_state_init_trainable
+        self._forget_bias = forget_bias
+        self._layere_normalization = layer_normalization
```
ryanjulian, line 106: `self._layer_normalization`

PR #1271, opened 2020-04-02
Update README to include environments for benchmarking

ahtsan:
Adds list of environments for benchmarking, for off-policy/on-policy algorithms on pixel environments
and all other algorithms on state environments.

### DIFF
--- a/tests/benchmarks/README.md
+++ b/tests/benchmarks/README.md
@@ -5,6 +5,34 @@ We have an automated system that will run any benchmark file named "test_benchma
 
 The results are displayed [here](https://rlworkgroup.github.io/benchmarks/)
 
+# Environments
+| Observations/Actions | Discrete | Continuous |
+| --- | --- | --- |
+| On-policy & Pixel | *PIXEL_ENV_SET | \ |
+| Off-policy & Pixel | Atari1M | \ |
+| State | *STATE_ENV_SET | MuJoCo1M |
+
+```
+PIXEL_ENV_SET: [
+    'MemorizeDigits-v0',
+    'CubeCrash-v0',
+    ^'Acrobot-v1',
+    ^'MountainCar-v0',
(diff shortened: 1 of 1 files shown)


ahtsan:
We might also add environments for meta-RL and multi-task RL

meta-RL: HalfCheetahDir and ML10?

Not sure about multi-task RL.

tests/benchmarks/README.md
```diff
@@ -5,6 +5,34 @@ We have an automated system that will run any benchmark file named "test_benchma
 
 The results are displayed [here](https://rlworkgroup.github.io/benchmarks/)
 
+# Environments
+| Observations/Actions | Discrete | Continuous |
+| --- | --- | --- |
+| On-policy & Pixel | *PIXEL_ENV_SET | \ |
+| Off-policy & Pixel | Atari1M | \ |
+| State | *STATE_ENV_SET | MuJoCo1M |
```
naeioi, line 13: I suggest a more straight forward table

```suggestion
# Environments
|Algorithm	|Observation	|Action	|Environment Set|
|---|---|---|---|
|On-Policy	|Pixel	|Discrete	|PIXEL_ENV_SET|
|Off-Policy	|Pixel	|Discrete	|Atari1M|
|All	|Non-Pixel	|Discrete	|STATE_ENV_SET|
|All	|Non-Pixel	|Continuous	|MuJoCo1M|

```

ahtsan: That looks good. We should also include those for meta-RL and multitask-RL. Let's modify accordingly after we have a conclusion.

ryanjulian: I think it would be much clearer to create a nicely-labeled list in Markdown rather than using code fences.

ahtsan: The code fences are actually a table, which I think is clearer than a list

PR #1272, opened 2020-04-02
Remove some examples from exclusion list

gitanshu:
These examples were removed from exclusion list because
they run fine under the default test for examples, but
somehow made their way back there. Removing again.


ahtsan:
How about other half_cheetah examples?

PR #1279, opened 2020-04-04
Use correct max_path_length for off-policy eval

krzentner:


krzentner:
Sorry Chelsea and Dierdre, I think our contributor bot is broken.

PR #1283, opened 2020-04-06
Fix MAML for num_grad_update > 1

### LINKED ISSUE #1269: MAML does not work when num_grad_updates > 1

naeioi:
This fixes #1269 , that MAML breaks when `num_grad_update` is larger than 1.

This is because gradient information is cleared out after the first gradient update, but it is needed for later gradient updates.


naeioi:
@Mergifyio rebase

PR #1284, opened 2020-04-07
Downgrade sphinx version

krzentner:
Two days ago, sphinx doc released a new version, 3.0.0.
Unfortunately, there's an issue with it that causes building docs for
garage to crash.

Until [that issue](https://github.com/sphinx-doc/sphinx/issues/7422) is
closed, we need to avoid updating for the CI to work.


krzentner:
For reference, here's an example run of the CI failing: https://travis-ci.com/github/rlworkgroup/garage/jobs/315662150

ryanjulian:

PR #1293, opened 2020-04-09
Shutdown workers after training in resume

ahtsan:
We should shutdown workers properly after training.


ahtsan:
@avnishn We didn't call `_shutdown_worker` at the end.

PR #1298, opened 2020-04-10
Add the "MultiprocessingSampler"

krzentner:
This sampler is designed based on what I believe the ideal architecture
for an (batch) sampler would be.

Specifically, workers use a simple state machine, which is either
"streaming" or "not streaming." Receiving an update for the agent (and
possibly environment), enters the worker into the "streaming" state.
From there, it keeps computing and transmitting rollouts back to the
sampler until it either receives a "stop" signal or the sampler's
rollout queue fills up.

Meanwhile, the manager (a.k.a. sampler) operates by sending start
signals (with agent updates) to all of its workers, collecting incoming
rollouts until it has enough, then trasmitting stop signals to all of
the workers.

The main benefits of this design are:
  - The sampler can return the batch of samples as soon as it has
  enough. Critically, it doesn't need to wait for the slowest worker
  (which means one worker can be e.g. frozen or rendering an environment
  slowly).
  - All messages are clearly owned by a queue fixed-size queue. There's
  no potential memory leak from e.g. not calling `.get()` on an
  AsyncResult.
  - After finishing a rollout, the worker doesn't need to wait for the
  manager to send a message to start the next rollout. Especially for
  fixed-length rollouts, the manager may be very busy at the moment a
  worker finishes with a rollout. Without this feature, the worker needs
  to wait for the manager to receive its (and potentially other)
  rollouts before it can start the next rollout, which may be a
  noticeable amount of time.

The main issues with this design are:
  - Since the worker needs to keep running after sending a message back
  to the manager, this method needs some type of raw queue or async
  stream API. Currently, it just uses multiprocessing queues. It would
  be potentially faster to use joblib's queues, since they can memmap
  numpy arrays, but those queues probably aren't intended to be a public
  API (they are part of joblib's multiprocessing shim, but have a
  slightly different API than multiprocessing's Queues).
  - Since we're not using an off-the-shelf pool API, we need to start /
  stop the workers ourselves. This is often tricky to do right. The
  current method daemonizes the workers, and is extra careful in
  `shutdown_worker()` at the texpense of time.
  - It's not clear how to distribute this architecture, since very few
  (or none?) of the commonly used distributed Python APIs allow direct
  access to Queues or async Streams.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,6 +30,7 @@ REQUIRED = [
     'ray',
     'scikit-image',
     'scipy',
+    'setproctitle>=1.0',
     'tensorflow' + TF_VERSION,
     'tensorflow-probability',
     'torch>=1.0.0',
--- a/src/garage/sampler/__init__.py
+++ b/src/garage/sampler/__init__.py
@@ -3,6 +3,7 @@
 from garage.sampler.batch_sampler import BatchSampler
 from garage.sampler.is_sampler import ISSampler
 from garage.sampler.local_sampler import LocalSampler
+from garage.sampler.multiprocessing_sampler import MultiprocessingSampler
 from garage.sampler.off_policy_vectorized_sampler import (
     OffPolicyVectorizedSampler)
 from garage.sampler.on_policy_vectorized_sampler import (
@@ -23,6 +24,7 @@ __all__ = [
     'LocalSampler',
     'RaySampler',
     'SamplerWorker',
+    'MultiprocessingSampler',
     'ParallelVecEnvExecutor',
     'VecEnvExecutor',
     'OffPolicyVectorizedSampler',
--- a/src/garage/sampler/multiprocessing_sampler.py
+++ b/src/garage/sampler/multiprocessing_sampler.py
@@ -1,0 +1,391 @@
+"""A multiprocessing sampler which avoids waiting as much as possible."""
+from collections import defaultdict
+import itertools
+import multiprocessing as mp
+import queue
+
+import cloudpickle
+import setproctitle
+
+from garage import TrajectoryBatch
+from garage.misc.prog_bar_counter import ProgBarCounter
+from garage.sampler.sampler import Sampler
+
+
+class MultiprocessingSampler(Sampler):
+    """Sampler that uses multiprocessing to distribute workers.
+
+    Args:
+        worker_factory(WorkerFactory): Pickleable factory for creating
+            workers. Should be transmitted to other processes / nodes where
+            work needs to be done, then workers should be constructed
+            there.
+        agents(Agent or List[Agent]): Agent(s) to use to perform rollouts.
+            If a list is passed in, it must have length exactly
+            `worker_factory.n_workers`, and will be spread across the
+            workers.
+        envs(gym.Env or List[gym.Env]): Environment rollouts are performed
+            in. If a list is passed in, it must have length exactly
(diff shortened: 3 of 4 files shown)


krzentner:
Note that there's still a bug with this sampler that causes a deadlock on process exit. I don't know why yet.

ryanjulian:
Thanks for doing this.

I did some research on the joblib situation, and it turns out there's some history here. The way we use joblib was blessed when `stateful_pool` was first written, but later the joblib authors decided to drastically cut the public APIs, and also fork out parts of the project to other packages.

The core joblib API is now just a couple classes which help with embarrassingly-parallel for loops, memoization, and persisting large numpy arrays to disk and efficiently-unpickling them by memmapping the files, rather than loading them (`joblib.load`/`joblib.dump`). Hidden in the joblib API is MemmappingPool, which is a `multiprocessing.pool` which uses memmap to avoid actually pickling objects across queues. I don't think us using this API is particularly-offensive, though I'm not sure anyone has verified the performance difference. It should be fairly large for pixel environments.

It used to contain a better multiprocessing implementation and queues, this has been moved to loky: https://github.com/tomMoral/loky

It used to help you define more complicated data pipelines in Python than just parallel-for, but they now endorse Dask for this workflow: https://docs.dask.org/en/latest/. Dask is a very interesting API which might provide us a nice lightweight alternative to Ray. I suspect Dask does something like joblib's MemappingPool for local processes, but wasn't able to verify that after spending 10 mins reading the code/docs (important entrypoints here: https://github.com/dask/dask/blob/master/dask/local.py).

ryanjulian:
Anyway, I would love to see this merged, especially if it can allow us to delete stateful_pool (or get close).

I would prefer if it were called MultiprocessingSampler because, despite being longer, it makes clear that this is a sampler with a `multiprocessing` backend.

krzentner:
> 
> 
> Note that there's still a bug with this sampler that causes a deadlock on process exit. I don't know why yet.

I tracked down why this is happening. It turns out python multiprocessing queues block on program exit to flush all of their pending data to the underlying pipe. That doesn't work when the parent process isn't reading from that pipe anymore.

krzentner:
So I found that porting this sampler to the loky backend is actually fairly easy. Unfortunately, the result is massively slower than before, and it breaks ray in the process somehow. The result is in the [loky-sampler branch](https://github.com/rlworkgroup/garage/tree/loky-sampler) if anyone wants to check it out.

Porting this design to Dask appears to be fundamentally impossible without a major rewrite, since it Dask requires specifying the entire task graph (instead of just constructing queues in whatever way desired).

setup.py
```diff
@@ -30,6 +30,7 @@
     'ray',
     'scikit-image',
     'scipy',
+    'setproctitle==1.1.10',
```
ryanjulian, line 33: this is from ray, right? can you leave a comment to that effect

krzentner: It's actually not, it's just used by multiprocessing workers to label themselves as workers.

ryanjulian: ah -- is there a less-strict constraint we can add? <=, >=, etc?

krzentner: Yeah, I've changed it to just `>=1.0`.

krzentner:
Okay, I've added a test for pickling (which caught a bug), and also for some (but not all) of the subprocesses crashing.

PR #1300, opened 2020-04-11
Remove after_n_builds in codecov config

ryanjulian:


krzentner:
Yeah, `make ci-job-mujoco` is definitely flaky.

PR #1303, opened 2020-04-14
MultitaskMetaworldWrapper

avnishn:
- Supports Multitask benchmarks from Metaworld including MT10 and MT50.
- Wraps MT environments that are constructed using the from task api.
- environments are cycled through by calling .reset().
- environments are cycled through in a round robin fashion.


src/garage/envs/multi_task_meta_world_wrapper.py
```diff
@@ -0,0 +1,129 @@
+"""A wrapper for Metaworld MultiTask Benchmarks."""
```
ryanjulian, line 1: the file and main class name should match exactly

src/garage/envs/multi_task_meta_world_wrapper.py
```diff
@@ -0,0 +1,129 @@
...
+import gym
+
+from garage.envs import GarageEnv, round_robin_strategy
+
+
+class MTMetaWorldWrapper(gym.Wrapper):
```
ryanjulian, line 7: why not just MetaWorldWrapper?

ryanjulian:
can you explain how this is different from using `from_task` + `MultiEnvWrapper`?

src/garage/envs/__init__.py
```diff
@@ -4,11 +4,24 @@
 from garage.envs.base import Step
 from garage.envs.env_spec import EnvSpec
 from garage.envs.grid_world_env import GridWorldEnv
+from garage.envs.multi_env_wrapper import (MultiEnvWrapper,
+                                           round_robin_strategy,
```
ryanjulian, line 8: you shouldn't import round_robin_strategy, etc. into this package space -- only MultiEnvWrapper

PR #1306, opened 2020-04-15
Add Multitask metaworld benchmark support

avnishn:
- Add option to MultiEnvWrapper to account for task_one_hot
  that is already appended to metaworld multi task benchmark
  environments

- tests that initialize metaworld benchmark environments now
  use the metaworld from_task api

opened this separate pr from #1303  to potentially resolve mujoco key issues

### DIFF
--- a/src/garage/envs/multi_env_wrapper.py
+++ b/src/garage/envs/multi_env_wrapper.py
@@ -10,6 +10,8 @@ import akro
 import gym
 import numpy as np
 
+from garage.envs import GarageEnv
+
 
 def round_robin_strategy(num_tasks, last_task=None):
     """A function for sampling tasks in round robin fashion.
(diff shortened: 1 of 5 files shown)


yonghyuc:
How can I get the environment name for logging from the environment wrapper?
I think we passed `env_names` to the previous ` MultiEnvSamplingWrapper` (when we did benchmark for the paper). And get the current environment name to log the values.
You can check [mtsac_gpu branch for the `MTMetaWorldWrapper`](https://github.com/rlworkgroup/garage/blob/9df6be93271cafca354ddf922f21aa37396d63ce/src/garage/envs/multi_task_metaworld_wrapper.py#L25)

avnishn:
@yonghyuc I'll add something in to support this use case

src/garage/envs/multi_env_wrapper.py
```diff
@@ -50,18 +52,24 @@ class MultiEnvWrapper(gym.Wrapper):
             A list of objects implementing gym.Env.
         sample_strategy (function(int, int)):
             Sample strategy to be used when sampling a new task.
-
+        metaworld_mt (bool): Set if the envs that are passed Metaworld
```
ryanjulian, line 55: please make this name/description general, rather than referring to metaworld. what does it do, and what does using this flag require of underlying envs?

avnishn:
@yonghyuc you can now access task names by accessing the active_task_name property of the MultiEnvWrapper.

yonghyuc:
Looks good to me

src/garage/envs/multi_env_wrapper.py
avnishn: I think I have now added sufficient information

ryanjulian:
I don't see anything addressing making it not metalworld specific. Hence no review.

src/garage/envs/multi_env_wrapper.py
```diff
@@ -194,3 +215,18 @@ def _obs_with_one_hot(self, obs):
         """
         oh_obs = np.concatenate([self.active_task_one_hot, obs])
         return oh_obs
+
+    @property
+    def active_task_name(self):
```
krzentner, line 220: Is this actually needed anywhere?

src/garage/envs/multi_env_wrapper.py
```diff
@@ -137,6 +153,8 @@ def active_task_one_hot(self):
             numpy.ndarray: one-hot representation of active task
 
         """
+        if self._metaworld_mt:
```
krzentner, line 156: Is this method actually needed anywhere?

src/garage/envs/multi_env_wrapper.py
krzentner: Is it possible to remove the metaworld specific logic from this wrapper and call this `add_onehot`? If not, ignore the rest of my comments on this PR.

ryanjulian:
Uhh please send a PR to fix the remaining comments

PR #1307, opened 2020-04-15
Refactor PPO auto benchmark script

zequnyu:
What this PR brings in:

- Refactoring of PPO benchmarking script: tests are separated by algorithm implementation
- `benchmark` mark for pytest to isolate these tests
- @wrap_experiment and iterate_experiments extract `log_dir` config and `csv_to_json` to helper function to keep tests clean
- Instead of a giant JSON with complicated file structure, data is now separated and file structure is clear

This severs as preparation for automatic benchmarking. I think we can use similar schema for other algorithms. I'd like to avoid a huge PR, so probably better to bring up piece by piece.

### DIFF
--- a/tests/benchmark_helper.py
+++ b/tests/benchmark_helper.py
@@ -1,9 +1,13 @@
 """helper functions for benchmarks."""
+import csv
 import json
+import os
+import pathlib
 
 import matplotlib.pyplot as plt
 import numpy as np
 import pandas as pd
+import tensorflow as tf
 
 
 def create_json(csvs, trials, seeds, xs, ys, factors, names):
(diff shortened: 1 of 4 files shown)


yonghyuc:
With this new test script, do I need to run each version of algorithm (TensorFlow, PyTorch, and openai.baseline) in different test function?
test_auto_benchmark_ppo_garage_tf -> only run PyTorch PPO
test_auto_benchmark_ppo_baselines -> only run openai.baseline.ppo2
test_auto_benchmark_ppo_garage_pytorch -> only run TensorFlow PPO
Is this correct?

How can I plot the graph using this new test script? In the previous benchmark script, I can plot a graph per environment using the output from all three algorithm through `benchmark_helper.plot_average_over_trials`. So, if i use this benchmark script, do I need to run other script to plot a graph of results?

tests/benchmarks/auto/test_auto_benchmark_ppo.py
```diff
@@ -0,0 +1,251 @@
...
+from garage.tf.envs import TfEnv
+from garage.tf.experiment import LocalTFRunner
+from garage.tf.optimizers import FirstOrderOptimizer
+from garage.tf.policies import GaussianMLPPolicy as TF_GMP
+from garage.torch.algos import PPO as PyTorch_PPO
+from garage.torch.policies import GaussianMLPPolicy as PyTorch_GMP
+from tests import benchmark_helper
+
+hyper_parameters = {
+    'n_epochs': 1,
```
yonghyuc: I think `n_trials` has to be bigger than this value.

tests/benchmarks/auto/test_auto_benchmark_ppo.py
```diff
@@ -0,0 +1,251 @@
...
+                runner.setup(algo, env)
+                runner.train(n_epochs=hyper_parameters['n_epochs'],
+                             batch_size=hyper_parameters['batch_size'])
+
+        for env_id, seed, log_dir in benchmark_helper.iterate_experiments(
+                ppo_garage_tf,
+                tasks,
+                seeds,
+                use_tf=True,
+                xcolumn='TotalEnvSteps',
```
yonghyuc, line 119: I think if we call the `csv_to_json` function outside of `iterate_experiments`, we don't have to pass `xcolumn`, `xlabel`, `ycolumn`, and `ylabel` and we can reduce the number of parameters of `iterate_experiments`.

tests/benchmark_helper.py
```diff
@@ -84,3 +87,88 @@ def plot_average_over_trials(csvs, ys, plt_file, env_id, x_label, y_label,
...
+                        ylabel):
+    """Iterator to iterate experiments.
+
+    Also it saves csv to JSON format preparing for automatic benchmarking.
+
+    Args:
+        tasks (list[dict]): List of running tasks.
+        seeds (list[int]): List of seeds.
+        func (func): The experiment function.
+        use_tf (bool): Weather TF is used.
```
ryanjulian, line 102: whether

tests/benchmark_helper.py
```diff
@@ -84,3 +87,88 @@ def plot_average_over_trials(csvs, ys, plt_file, env_id, x_label, y_label,
...
+
+    Args:
+        name (str): The experiment name.
+
+    Returns:
+        str: Log directory.
+
+    """
+    if name is None:
+        return ('{data}/local/{prefix}'.format(data=os.path.join(
```
ryanjulian, line 145: you might find pathlib.Path useful for making this less ugly

tests/benchmarks/auto/test_auto_benchmark_ppo.py
zequnyu: The reason I considered to add this to the iterator is because we need to manually call `csv_to_json` for every benchmarking test, which is kinda tedious to me. We need to specify the column and label name anyway, so I'd prefer keeping as is.

zequnyu:
> 
> 
> With this new test script, do I need to run each version of algorithm (TensorFlow, PyTorch, and openai.baseline) in different test function?
> test_auto_benchmark_ppo_garage_tf -> only run PyTorch PPO
> test_auto_benchmark_ppo_baselines -> only run openai.baseline.ppo2
> test_auto_benchmark_ppo_garage_pytorch -> only run TensorFlow PPO
> Is this correct?
> 
> How can I plot the graph using this new test script? In the previous benchmark script, I can plot a graph per environment using the output from all three algorithm through `benchmark_helper.plot_average_over_trials`. So, if i use this benchmark script, do I need to run other script to plot a graph of results?

Hi @yonghyuc 
1. Yes, your understanding is correct.
2. I didn't implement that yet. The starting point of this refactoring is to prepare for the automatic benchmarking, so for now I don't aim to stop people using the existing scripts. However my plan is to refactor all the scripts step by step, and I believe by making minor effort we could replicate the plot functionality. By that time, we can delete those old scripts and replace with new one.

PR #1309, opened 2020-04-15
Unpin TF

ahtsan:
Unpinned TensorFlow, updated some dependency and added a bunch of docstrings.


setup.py
```diff
@@ -55,7 +54,7 @@
 EXTRAS['all'] = list(set(sum(EXTRAS.values(), [])))
 
 # dependencies for using gpu, not included in 'all'
-EXTRAS['gpu'] = ['tensorflow-gpu' + TF_VERSION]
+EXTRAS['gpu'] = ['tensorflow-gpu']
```
ryanjulian, line 57: you can remove this line because TF now support GPU without an extra

naeioi:
A general question: Can TensorFlow guarantee all 1.x functionalities if unpinned?

ryanjulian:
@naeioi yes, if you import from the path `tf.compat.v1`

src/garage/tf/q_functions/discrete_cnn_q_function.py
```diff
@@ -64,25 +64,26 @@ class DiscreteCNNQFunction(QFunction):
...
+
     """
 
     def __init__(self,
                  env_spec,
                  filter_dims,
                  num_filters,
                  strides,
-                 hidden_sizes=[256],
+                 hidden_sizes=(256),
```
naeioi, line 75: (256, )

setup.py
ahtsan: removed

ahtsan:
@ryanjulian Can you force merge it? It passed all tests but it asked for a lot of docstring since I touched a lot of files. Also, the missing tests are mostly from `garage.distributions`, which I'm going to replace it with `tfp.distributions`.

src/garage/tf/models/gaussian_cnn_model.py
```diff
@@ -86,8 +90,10 @@ class GaussianCNNModel(Model):
                exponential transformation
             - softplus: the std will be computed as log(1+exp(x))
         layer_normalization (bool): Bool for using layer normalization or not.
+
     """
 
+    # pylint: disable=dangerous-default-value
```
ryanjulian, line 96: please don't disable this -- it's a real problem.

which value is dangerous?

src/garage/tf/misc/tensor_utils.py
```diff
@@ -52,19 +53,21 @@ def flatten_batch(t, name='flatten_batch'):
...
-        t, tf.cast(tf.round(valid), tf.int32), 2, name=name)[1]
+    return tf.dynamic_partition(t,
+                                tf.cast(tf.round(valid), tf.int32),
+                                2,
+                                name=name)[1]
 
 
 def filter_valids_dict(d, valid, name=None):
-    with tf.name_scope(name, 'filter_valids_dict', [d, valid]):
+    with tf.name_scope('filter_valids_dict' if name is None else name):
```
ryanjulian, line 70: aren't these all equivalent to `name='filter_valids_dict'` etc.?

src/garage/tf/q_functions/discrete_cnn_q_function.py
```diff
@@ -160,14 +162,25 @@ def _initialize(self):
...
-        """Input tf.Tensor of the Q-function."""
+        """Get input.
+
+        Return:
+            tf.Tensor: QFunction Input.
+
+        """
         return self.model.networks['default'].input
 
+    # pylint: disable=arguments-differ
```
ryanjulian, line 183: reason? doesn't seem like this should different from the ABC

src/garage/tf/q_functions/discrete_mlp_q_function.py
```diff
@@ -101,14 +104,25 @@ def _initialize(self):
...
-        """Get input."""
+        """Get input.
+
+        Return:
+            tf.Tensor: QFunction Input.
+
+        """
         return self.model.networks['default'].input
 
+    # pylint: disable=arguments-differ
```
ryanjulian, line 125: why?

tests/benchmarks/garage/tf/policies/benchmark_categorical_cnn_policy.py
```diff
@@ -32,11 +32,11 @@
 }
 
 
-class TestBenchmarkCategoricalCNNPolicy:
+class BenchmarkCategoricalCNNPolicy:
     '''Compare benchmarks between garage and baselines.'''
 
     @pytest.mark.huge
```
ryanjulian, line 38: you can remove all the pytest marks

tests/garage/sampler/test_local_sampler.py
```diff
@@ -100,8 +100,8 @@ def test_update_envs_env_update():
     for rollout in rollouts.split():
         mean_rewards.append(rollout.rewards.mean())
         goals.append(rollout.env_infos['task'][0]['goal'])
-    assert np.var(mean_rewards) > 0
-    assert np.var(goals) > 0
+    assert len(mean_rewards) == 11
```
ryanjulian, line 103: umm variance is not the same thing as length -- why replace this condition when you can add it?

src/garage/tf/models/gaussian_cnn_model.py
ryanjulian: just change the lists to tuple...

src/garage/tf/misc/tensor_utils.py
ahtsan: I thought someone may give it a special name scope, i.e. name is not None

src/garage/tf/models/gaussian_cnn_model.py
ahtsan: I forgot to remove thesde

src/garage/tf/q_functions/discrete_cnn_q_function.py
ahtsan: it does, sometimes it takes one argument (state), sometimes two (state, action)

tests/garage/sampler/test_local_sampler.py
ahtsan: actually the original test is flaky -- sometimes all rewards are the same so variance is indeed 0. I will add these two checks back but with `>=` instead.

src/garage/tf/misc/tensor_utils.py
ahtsan: after checking, there's no usage that pass a name to these functions. I will remove the name argument.

ahtsan: e.g. in NPO
```
...
                policy_state_info_vars_flat = flatten_batch_dict(
                    policy_state_info_vars, name='policy_state_info_vars_flat')
                policy_old_dist_info_vars_flat = flatten_batch_dict(
                    policy_old_dist_info_vars,
                    name='policy_old_dist_info_vars_flat')
```

ryanjulian: ...but why not just use the default function argument to achieve that?

src/garage/tf/q_functions/discrete_cnn_q_function.py
ryanjulian: that should not be the case for a Q-function -- it should always be conditioned on (s,a)

who is conditioning a Q-function on just state?

ahtsan: discrete q-function only takes state in garage, and output all the Q values for all actions... sounds like this implementation is not right?

src/garage/tf/misc/tensor_utils.py
ahtsan: I see, I missed your point

src/garage/tf/q_functions/discrete_cnn_q_function.py
ahtsan: if that's the case, can we fix that in another PR?

ryanjulian:
Does this need force-push? We can fix the possible Q-function issue when we unify interfaces later.

PR #1310, opened 2020-04-16
 Refactor TRPO auto benchmark script

zequnyu:
Following #1307.

### DIFF
--- a/tests/benchmarks/auto/test_auto_benchmark_ppo.py
+++ b/tests/benchmarks/auto/test_auto_benchmark_ppo.py
@@ -119,7 +119,7 @@ def test_auto_benchmark_ppo_garage_tf():
 
 @pytest.mark.benchmark
 def test_auto_benchmark_ppo_baselines():
-    """Create baselines model and training.
+    """Create PPO baselines model and training.
 
     Training over different environments and seeds.
 
--- a/tests/benchmarks/auto/test_auto_benchmark_trpo.py
+++ b/tests/benchmarks/auto/test_auto_benchmark_trpo.py
@@ -1,0 +1,222 @@
+"""A regression test over TRPO Algorithms for automatic benchmarking.
+(garage-PyTorch-TRPO, garage-TensorFlow-TRPO, and baselines-TRPO)
+
+Unlike garage, baselines doesn't set max_path_length. It keeps steps the action
+until it's done. So we introduced tests.wrappers.AutoStopEnv wrapper to set
+done=True when it reaches max_path_length. We also need to change the
+garage.tf.samplers.BatchSampler to smooth the reward curve.
+"""
+import multiprocessing
+import random
+
+from baselines import logger as baselines_logger
+from baselines.bench import benchmarks
+from baselines.common import set_global_seeds
+from baselines.logger import configure
+from baselines.trpo_mpi import trpo_mpi
+import gym
+import pytest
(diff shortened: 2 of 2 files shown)


avnishn:
I know someone before was saying that pytest wasn't the way to go for auto benchmarking, but I feel like it provides a seamless way to manage running many experiments. We could write the infrastructure ourselves (it's minimal) but then we would have to maintain it.

Reasons I could see us writing the infrastructure ourselves includes the possibility that pytest reduces runtime performance, or introduces not easy to manage bugs that are caused by xdist. I don't see the latter issuer occuring but potentially, the former. I guess we won't know till we test it all.

zequnyu:
Agree. I thought about it last night and I'm totally ok with the current status. And it's probably easier to use pytest mark to control these tests, since all you need to trigger is `pytest -m benchmark`.

ryanjulian:
Ah but there's a couple problems with this logic in the case of benchmarking all of garage (e.g. the autobenchmark project)

1. pytest isn't designed for "tests" which run for hours or days -- the 95th percentile pytest case runs in a fraction of a second.
1. It has no provision for restarting a test which is interrupted. In this extremely rare case, you're assumed to just restart the test. This might be a deal-breaker for benchmarking.
1. No first-class distributed features. There exist plug-ins (e.g. xdist) for distributing test jobs across machines, but they are not a first-class feature and, again, they are designed to farm *many very short tests* across *few hosts*, not *few very very long tests* across *few hosts*.

I'm not saying pytest is bad, just that it doesn't bring a lot to the table. If you imagine kicking off the auto benchmarking suite by running `pytest -m benchmark`, I think you will quickly find this makes things unnecessarily hard and requires lots of stupid workarounds, and for what?

It seems what you like here is that you can quickly enumerate all the benchmarks by asking pytest to go discover tests marked with `benchmark`. We can duplicate this feature easily, e.g.

* by just putting all the benchmark scripts in a directory tree, and assuming every script in that tree is a benchmark, of if you want to put benchmark scripts in a arbitrary locations...
* keeping a manifest (file with a list) of all current benchmark scripts
* writing our own discovery script which looks for a special-purpose comment, mark, file name, etc., which would be like ~10 lines of bash in a Makefile

I understand you are hungry for something to catalog/discover/track all of the possible benchmark jobs, then supervise them as they run. You're right, this is a key element of this project -- but pytest really isn't a solution. There are many possibilities which would require a similar amount of work or less. These are generally called workflow managers -- they come in many flavors.

* CIs: TravisCI, Azure Pipelines, GitLab CI, GitHub Actions, CircleCI, Jenkins, TeamCity, etc. -- note that many of these (including Azure and Github Actions) will manage/track your jobs using the cloud service, but let you setup your local machines to actually run the jobs. 
* Cluster managers: slurm, kubernetes, etc.
* ML-specific solutions: ray tune, etc.
* Pipeline managers: Apache Airflow, etc. e.g. see https://github.com/pditommaso/awesome-pipeline

Many people have the problem of "I have x work, y machines, and want to figure out which xs to put on which ys and get all the results," so the design space is huge. This is part of what makes this project fun!

krzentner:
This looks good. I think it would also be fine to have each experiment definition in its own file in a directory, and have a single test / script to run all of them.

We already do this for examples, so it shouldn't be too complicated.

No matter what we do next, I think this is a step in the right direction.

Issue #1212, opened 2020-03-11
Resume integration test for each algorithm


ryanjulian:

ryanjulian:
Perhaps a better way to do this: change the LocalRunner implementation so that `runner.train` is equivalent to `runner.save/runner.resume`

PR #1316, opened 2020-04-17
Fix PEARL resume

### LINKED ISSUE #1314: How to resume training 'Pearl'?
Hello! I ran a example code: examples/torch/pearl_ml1_push.py and got a log data.
Then I wrote a resume code like this:
```
def` resume_task(snapshot_config, *_):
    runner = LocalRunner(snapshot_config=snapshot_config)
    runner.restore(from_dir='dir')
    runner.resume()

run_experiment(
    resume_task,
    log_dir='dir',
    snapshot_mode='last',
    seed=1,
)
```

but this doesn't work. The errors are the following:
image

How can I fix this problem?

naeioi:
This will close #1314


src/garage/torch/algos/pearl.py
```diff
@@ -165,6 +165,8 @@ def __init__(self,
         self._update_post_train = update_post_train
         self._task_idx = None
 
+        self._just_resume = False
```
krzentner, line 168: Can you call this `_is_resuming`? The meaning was a little unclear to me.

krzentner:
We should probably also *save* (and resume if they're present) the replay buffer's too by default. That can wait for another PR though.

naeioi:
@krzentner That will take a long time to save. Linda dropped reply_buffer because it is too expensive.

krzentner:
> 
> 
> @krzentner That will take a long time to save. Linda dropped reply_buffer because it is too expensive.

Yeah, that makes sense. Maybe in the future we should create a replay buffer backed up by a database (so we don't need to re-pickle every timestep every snapshot but can still resume training seamlessly).

naeioi:
@krzentner I agreed. An off-policy algorithm should be able to restart and absorb any collected samples as long as the task-set remains the same.

naeioi:
@ryanjulian It wouldn't fail. We have no test on per-algo resuming (yet)

naeioi:
@Mergifyio rebase

ryanjulian:
#1212 is relevant here

ryanjulian:
ohh i like that idea. maybe a SQLite replay buffer

naeioi:
@ryanjulian Resuming test for PEARL is particularly hard. It takes around 10 mins before the first pickle file is generated. I am afraid i need to add an additional travis to run just this.

ryanjulian:
ah okay, that's crazy. does it do that if you configure the replay buffer and networks to be tiny?

naeioi:
@ryanjulian actually that is the iteration time without saving replay buffer. I don't have a number if I save it

naeioi:
I think I can make a workaround testing - By pickling and unpickling a PEARL instance, without running any training. Doing so necessarily guarantees a successful unpickling. (but not sufficiently)

tests/garage/torch/algos/test_pearl.py
```diff
@@ -120,3 +123,37 @@ def test_pearl_ml1_push(self):
...
+                      qf=qf,
+                      vf=vf,
+                      num_train_tasks=5,
+                      num_test_tasks=5,
+                      latent_dim=5,
+                      encoder_hidden_sizes=[10, 10],
+                      test_env_sampler=test_env_sampler)
+
+        pickled = pickle.dumps(pearl)
+        pickle.loads(pickled)
```
ryanjulian, line 159: can you test to make sure your unpickled instance has all the data it's supposed to? you're 90% there lol.

naeioi:
@Mergifyio rebase

naeioi:
I've improved the code coverage as much as I can, but I still can't get a 90% diff coverage rate. I think we can just merge this as it is.

lywong92:
Sorry for the late reply. I couldn't run the algorithm on HalfCheetahVel or MetaWorld if the buffers were included. I got errors from pickling and unpickling. For simpler algorithms like HalfCheetahDir, it worked but still took around 10 minutes to save a snapshot. I had to resume before, but I just hacked the code.

PR #1317, opened 2020-04-18
Add value function for torch

yonghyuc:
Implement GaussianMLPValueFunction for PyTorch algorithms.

A value function computes loss and returns the value to an algorithm to update the weights. 
An algorithm has two optimizers, for policy and value function.

### DIFF
--- a/examples/torch/maml_ppo_half_cheetah_dir.py
+++ b/examples/torch/maml_ppo_half_cheetah_dir.py
@@ -11,9 +11,9 @@ from garage.envs.mujoco import HalfCheetahDirEnv
 from garage.experiment import LocalRunner, MetaEvaluator
 from garage.experiment.deterministic import set_seed
 from garage.experiment.task_sampler import SetTaskSampler
-from garage.np.baselines import LinearFeatureBaseline
 from garage.torch.algos import MAMLPPO
 from garage.torch.policies import GaussianMLPPolicy
+from garage.torch.value_functions import GaussianMLPValueFunction
 
 
 @click.command()
@@ -46,7 +46,10 @@ def maml_ppo(ctxt, seed, epochs, rollouts_per_task, meta_batch_size):
         output_nonlinearity=None,
     )
 
-    value_function = LinearFeatureBaseline(env_spec=env.spec)
+    value_function = GaussianMLPValueFunction(env_spec=env.spec,
+                                              hidden_sizes=(32, 32),
+                                              hidden_nonlinearity=torch.tanh,
+                                              output_nonlinearity=None)
 
     max_path_length = 100
 
--- a/src/garage/torch/algos/maml.py
+++ b/src/garage/torch/algos/maml.py
@@ -25,7 +25,6 @@ class MAML:
             computing loss.
         env (garage.envs.GarageEnv): A gym environment.
         policy (garage.torch.policies.Policy): Policy.
-        value_function (garage.np.baselines.Baseline): The value function.
         meta_optimizer (Union[torch.optim.Optimizer, tuple]):
             Type of optimizer.
             This can be an optimizer type such as `torch.optim.Adam` or a tuple
@@ -45,7 +44,6 @@ class MAML:
                  inner_algo,
                  env,
                  policy,
-                 value_function,
                  meta_optimizer,
                  meta_batch_size=40,
                  inner_lr=0.1,
@@ -63,7 +61,8 @@ class MAML:
         self._meta_evaluator = meta_evaluator
         self._policy = policy
         self._env = env
-        self._value_function = value_function
+        self._value_function = copy.deepcopy(inner_algo._value_function)
+        self._initial_vf_state = self._value_function.state_dict()
         self._num_grad_updates = num_grad_updates
         self._meta_batch_size = meta_batch_size
         self._inner_algo = inner_algo
@@ -149,6 +148,34 @@ class MAML:
         tu.update_module_params(self._old_policy, old_theta)
 
         return average_return
+
+    def _train_value_function(self, paths):
+        """Train the value function.
+
+        Args:
+            paths (list[dict]): A list of collected paths.
+
+        Returns:
+            torch.Tensor: Calculated mean scalar value of value function loss
+                (float).
+
+        """
+        # MAML resets a value function to its initial state before training.
+        self._value_function.load_state_dict(self._initial_vf_state)
+
+        obs = np.concatenate([path['observations'] for path in paths], axis=0)
+        returns = np.concatenate([path['returns'] for path in paths])
+        obs = torch.from_numpy(obs)
+        returns = torch.from_numpy(returns)
+
+        vf_loss = self._value_function.compute_loss(obs, returns)
+        # pylint: disable=protected-access
+        self._inner_algo._vf_optimizer.zero_grad()
+        vf_loss.backward()
+        # pylint: disable=protected-access
+        self._inner_algo._vf_optimizer.step()
+
+        return vf_loss
 
     def _obtain_samples(self, runner):
         """Obtain samples for each task before and after the fast-adaptation.
@@ -347,7 +374,7 @@ class MAML:
 
         Args:
             itr (int): Iteration number.
-            paths (list[dict]): A list of collected paths
+            paths (list[dict]): A list of collected paths.
 
         Returns:
             MAMLTrajectoryBatch: Processed samples data.
@@ -355,10 +382,10 @@ class MAML:
         """
         for path in paths:
             path['returns'] = tensor_utils.discount_cumsum(
-                path['rewards'], self._inner_algo.discount)
+                path['rewards'], self._inner_algo.discount).copy()
 
-        self._value_function.fit(paths)
-        obs, actions, rewards, valids, baselines \
+        self._train_value_function(paths)
+        obs, actions, rewards, _, valids, baselines \
             = self._inner_algo.process_samples(itr, paths)
         return MAMLTrajectoryBatch(paths, obs, actions, rewards, valids,
                                    baselines)
--- a/src/garage/torch/algos/maml_trpo.py
+++ b/src/garage/torch/algos/maml_trpo.py
@@ -3,7 +3,8 @@ import torch
 
 from garage.torch.algos import _Default, VPG
 from garage.torch.algos.maml import MAML
-from garage.torch.optimizers import ConjugateGradientOptimizer
+from garage.torch.optimizers import (ConjugateGradientOptimizer,
+                                     OptimizerWrapper)
 
 
 class MAMLTRPO(MAML):
@@ -66,11 +67,17 @@ class MAMLTRPO(MAML):
                  num_grad_updates=1,
                  meta_evaluator=None,
                  evaluate_every_n_epochs=1):
+
+        policy_optimizer = OptimizerWrapper(
+            (torch.optim.Adam, dict(lr=inner_lr)), policy)
+        vf_optimizer = OptimizerWrapper((torch.optim.Adam, dict(lr=inner_lr)),
+                                        value_function)
+
         inner_algo = VPG(env.spec,
                          policy,
                          value_function,
-                         optimizer=torch.optim.Adam,
-                         policy_lr=inner_lr,
+                         policy_optimizer=policy_optimizer,
+                         vf_optimizer=vf_optimizer,
                          max_path_length=max_path_length,
                          num_train_per_epoch=1,
                          discount=discount,
@@ -88,7 +95,6 @@ class MAMLTRPO(MAML):
         super().__init__(inner_algo=inner_algo,
                          env=env,
                          policy=policy,
-                         value_function=value_function,
                          meta_optimizer=meta_optimizer,
                          meta_batch_size=meta_batch_size,
                          inner_lr=inner_lr,
--- a/src/garage/torch/algos/ppo.py
+++ b/src/garage/torch/algos/ppo.py
@@ -2,6 +2,7 @@
 import torch
 
 from garage.torch.algos import VPG
+from garage.torch.optimizers import OptimizerWrapper
 
 
 class PPO(VPG):
@@ -10,13 +11,12 @@ class PPO(VPG):
     Args:
         env_spec (garage.envs.EnvSpec): Environment specification.
         policy (garage.torch.policies.base.Policy): Policy.
-        value_function (garage.np.baselines.Baseline): The value function.
-        optimizer (Union[type, tuple[type, dict]]): Type of optimizer.
-            This can be an optimizer type such as `torch.optim.Adam` or a
-            tuple of type and dictionary, where dictionary contains arguments
-            to initialize the optimizer
-            e.g. `(torch.optim.Adam, {'lr' = 1e-3})`
-        policy_lr (float): Learning rate for policy parameters.
+        value_function (garage.torch.value_functions.ValueFunction): The value
+            function.
+        policy_optimizer (garage.torch.optimizer.OptimizerWrapper): Optimizer
+            for policy.
+        vf_optimizer (garage.torch.optimizer.OptimizerWrapper): Optimizer for
+            value function.
         max_path_length (int): Maximum length of a single rollout.
         lr_clip_range (float): The limit on the likelihood ratio between
             policies.
@@ -41,8 +41,6 @@ class PPO(VPG):
             dense entropy to the reward for each time step. 'regularized' adds
             the mean entropy to the surrogate objective. See
             https://arxiv.org/abs/1805.00909 for more details.
-        minibatch_size (int): Batch size for optimization.
-        max_optimization_epochs (int): Maximum number of epochs for update.
 
     """
 
@@ -50,8 +48,8 @@ class PPO(VPG):
                  env_spec,
                  policy,
                  value_function,
-                 optimizer=torch.optim.Adam,
-                 policy_lr=3e-4,
+                 policy_optimizer=None,
+                 vf_optimizer=None,
                  max_path_length=500,
                  lr_clip_range=2e-1,
                  num_train_per_epoch=1,
@@ -62,15 +60,26 @@ class PPO(VPG):
                  policy_ent_coeff=0.0,
                  use_softplus_entropy=False,
                  stop_entropy_gradient=False,
-                 entropy_method='no_entropy',
-                 minibatch_size=128,
-                 max_optimization_epochs=10):
+                 entropy_method='no_entropy'):
+
+        if policy_optimizer is None:
+            policy_optimizer = OptimizerWrapper(
+                (torch.optim.Adam, dict(lr=2.5e-4)),
+                policy,
+                max_optimization_epochs=10,
+                minibatch_size=64)
+        if vf_optimizer is None:
+            vf_optimizer = OptimizerWrapper(
+                (torch.optim.Adam, dict(lr=2.5e-4)),
+                value_function,
+                max_optimization_epochs=10,
+                minibatch_size=64)
 
         super().__init__(env_spec=env_spec,
                          policy=policy,
                          value_function=value_function,
-                         optimizer=optimizer,
-                         policy_lr=policy_lr,
+                         policy_optimizer=policy_optimizer,
+                         vf_optimizer=vf_optimizer,
                          max_path_length=max_path_length,
                          num_train_per_epoch=num_train_per_epoch,
                          discount=discount,
@@ -80,9 +89,7 @@ class PPO(VPG):
                          policy_ent_coeff=policy_ent_coeff,
                          use_softplus_entropy=use_softplus_entropy,
                          stop_entropy_gradient=stop_entropy_gradient,
-                         entropy_method=entropy_method,
-                         minibatch_size=minibatch_size,
-                         max_optimization_epochs=max_optimization_epochs)
+                         entropy_method=entropy_method)
 
         self._lr_clip_range = lr_clip_range
 
--- a/src/garage/torch/algos/trpo.py
+++ b/src/garage/torch/algos/trpo.py
@@ -3,6 +3,7 @@ import torch
 
 from garage.torch.algos import VPG
 from garage.torch.optimizers import ConjugateGradientOptimizer
+from garage.torch.optimizers import OptimizerWrapper
 
 
 class TRPO(VPG):
@@ -11,7 +12,12 @@ class TRPO(VPG):
     Args:
         env_spec (garage.envs.EnvSpec): Environment specification.
         policy (garage.torch.policies.base.Policy): Policy.
-        value_function (garage.np.baselines.Baseline): The value function.
+        value_function (garage.torch.value_functions.ValueFunction): The value
+            function.
+        policy_optimizer (garage.torch.optimizer.OptimizerWrapper): Optimizer
+            for policy.
+        vf_optimizer (garage.torch.optimizer.OptimizerWrapper): Optimizer for
+            value function.
         max_path_length (int): Maximum length of a single rollout.
         num_train_per_epoch (int): Number of train_once calls per epoch.
         discount (float): Discount.
@@ -25,8 +31,6 @@ class TRPO(VPG):
             standardized before shifting.
         policy_ent_coeff (float): The coefficient of the policy entropy.
             Setting it to zero would mean no entropy regularization.
-        max_kl_step (float): The maximum KL divergence between old and new
-            policies.
         use_softplus_entropy (bool): Whether to estimate the softmax
             distribution of the entropy to prevent the entropy from being
             negative.
@@ -43,6 +47,8 @@ class TRPO(VPG):
                  env_spec,
                  policy,
                  value_function,
+                 policy_optimizer=None,
+                 vf_optimizer=None,
                  max_path_length=100,
                  num_train_per_epoch=1,
                  discount=0.99,
@@ -50,19 +56,26 @@ class TRPO(VPG):
                  center_adv=True,
                  positive_adv=False,
                  policy_ent_coeff=0.0,
-                 max_kl_step=0.01,
                  use_softplus_entropy=False,
                  stop_entropy_gradient=False,
                  entropy_method='no_entropy'):
 
-        optimizer = (ConjugateGradientOptimizer, {
-            'max_constraint_value': max_kl_step
-        })
+        if policy_optimizer is None:
+            policy_optimizer = OptimizerWrapper(
+                (ConjugateGradientOptimizer, dict(max_constraint_value=0.01)),
+                policy)
+        if vf_optimizer is None:
+            vf_optimizer = OptimizerWrapper(
+                (torch.optim.Adam, dict(lr=2.5e-4)),
+                value_function,
+                max_optimization_epochs=10,
+                minibatch_size=64)
 
         super().__init__(env_spec=env_spec,
                          policy=policy,
                          value_function=value_function,
-                         optimizer=optimizer,
+                         policy_optimizer=policy_optimizer,
+                         vf_optimizer=vf_optimizer,
                          max_path_length=max_path_length,
                          num_train_per_epoch=num_train_per_epoch,
                          discount=discount,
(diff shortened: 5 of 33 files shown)


src/garage/torch/value_functions/gaussian_mlp_value_function.py
```diff
@@ -0,0 +1,130 @@
+"""A value function based on a GaussianMLP model."""
+import torch
+from torch import nn
+
+from garage.torch.modules import GaussianMLPModule
+from garage.torch.value_functions.base import ValueFunction
+
+
+class GaussianMLPValueFunction(ValueFunction, GaussianMLPModule):
```
krzentner, line 9: Please don't inherit from `GaussianMLPModule`. We did this with policies, and it turned out to make it impossible to make the pytorch policies into a class hierarchy without some major hacks or undoing it. Instead, `ValueFunction` should inherit from `torch.nn.module`, and the `GaussianMLPModule` should just be a field.

src/garage/torch/value_functions/base.py
```diff
@@ -0,0 +1,71 @@
...
+        name (str): Value function name, also the variable scope.
+
+    """
+
+    def __init__(self, env_spec, name):
+        self._mdp_spec = env_spec
+        self.name = name
+
+    @abc.abstractmethod
+    def get_param_values(self):
```
krzentner, line 19: I'm not quite sure we need these methods right now. Let's try not having them, and we can add them later if needed.

src/garage/torch/algos/vpg.py
```diff
@@ -148,41 +157,42 @@ def train_once(self, itr, paths):
         obs_flat = torch.cat(filter_valids(obs, valids))
         actions_flat = torch.cat(filter_valids(actions, valids))
         rewards_flat = torch.cat(filter_valids(rewards, valids))
-        advantages_flat = self._compute_advantage(rewards, valids, baselines)
+        returns_flat = torch.cat(filter_valids(returns, valids))
```
krzentner, line 160: You have an extra line here.

examples/torch/maml_ppo_half_cheetah_dir.py
```diff
@@ -11,9 +11,9 @@
 from garage.experiment import LocalRunner, MetaEvaluator
```
ryanjulian, line 11: don't forget to update all the other torch examples

src/garage/torch/algos/maml.py
```diff
@@ -150,6 +149,32 @@ def train_once(self, runner, all_samples, all_params):
 
         return average_return
 
+    def _train_value_function(self, paths):
```
ryanjulian, line 152: @naeioi this open the possibility of applying MAML to both the VF and the policy. are you interested in trying this?

src/garage/torch/algos/trpo.py
```diff
@@ -72,7 +86,22 @@ def __init__(self,
...
+                         minibatch_size=minibatch_size,
+                         max_optimization_epochs=max_optimization_epochs)
+
+    def _train(self, obs_flat, actions_flat, rewards_flat, returns_flat,
+               advs_flat):
+        policy_loss = self._train_policy(obs_flat, actions_flat, rewards_flat,
+                                         advs_flat)
+        logger.log('Policy loss: {}'.format(policy_loss))
+
+        batch_dataset = BatchDataset((obs_flat, returns_flat),
```
ryanjulian, line 99: why was this necessary for minibatching for the value function, but not the policy?

src/garage/torch/algos/vpg.py
```diff
@@ -22,13 +23,18 @@ class VPG(BatchPolopt):
...
+            for policy. This can be an optimizer type such as
+            `torch.optim.Adam` or a tuple of type and dictionary, where
+            dictionary contains arguments to initialize the optimizer.
+            e.g. `(torch.optim.Adam, {'lr' = 1e-3})`
+        vf_optimizer (Union[type, tuple[type, dict]]): Type of optimizer
+            for value function. This can be an optimizer type such as
+            `torch.optim.Adam` or a tuple of type and dictionary, where
+            dictionary contains arguments to initialize the optimizer.
             e.g. `(torch.optim.Adam, {'lr' = 1e-3})`
-        policy_lr (float): Learning rate for policy parameters.
```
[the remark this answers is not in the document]
ryanjulian: yes. should probably add vf_lr for symmetry. it should default to policy_lr, IMO.

src/garage/torch/algos/vpg.py
```diff
@@ -449,9 +496,10 @@ def process_samples(self, itr, paths):
...
             for path in paths
         ])
-        baselines = torch.stack([
-            pad_to_last(self._get_baselines(path),
-                        total_length=self.max_path_length) for path in paths
+        returns = torch.stack([
+            pad_to_last(path['returns'], total_length=self.max_path_length)
+            for path in paths
         ])
+        baselines = self._value_function.predict(obs)
```
ryanjulian, line 503: please remove the predict interface.

the intention of this PR is to remove the fit/predict interface and retrieve the VF predictions by calling `forward` on the VF within the same computation graph as the policy.

src/garage/torch/value_functions/base.py
```diff
@@ -0,0 +1,71 @@
...
+
+    Args:
+        env_spec (garage.envs.env_spec.EnvSpec): Environment specification.
+        name (str): Value function name, also the variable scope.
+
+    """
+
+    def __init__(self, env_spec, name):
+        self._mdp_spec = env_spec
+        self.name = name
```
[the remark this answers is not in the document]
ryanjulian: but where will we find out what the correct prefix is?

src/garage/torch/value_functions/base.py
```diff
@@ -0,0 +1,71 @@
...
+            returns (torch.Tensor): Acquired returns with shape :math:`(N, )`.
+
+        Returns:
+            torch.Tensor: Calculated negative mean scalar value of
+                objective (float).
+
+        """
+
+    @abc.abstractmethod
+    def predict(self, obs):
```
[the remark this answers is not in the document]
ryanjulian: agreed. the idea in this PR is to just consider the VF as any other NN primitive in garage. in torch, that means you retrieve its values with forward. this will allow us to compute the policy loss with a single feedforward, for instance. it also allows us to apply things like MAML to the VF as well as the policy.

src/garage/torch/value_functions/base.py
```diff
@@ -0,0 +1,71 @@
...
+            obs (torch.Tensor): Observation from the environment
+                with shape :math:`(P, O*)`.
+
+        Returns:
+            torch.Tensor: Calculated baselines given observations with
+                shape :math:`(P, O*)`.
+
+        """
+
+    def log_diagnostics(self, paths):
```
ryanjulian, line 65: please remove this. we'll add it if we're sure we need it later.

src/garage/torch/value_functions/gaussian_mlp_value_function.py
ryanjulian: i agree.

src/garage/torch/value_functions/base.py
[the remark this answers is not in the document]
krzentner: I had assumed that whatever is calling `log_diagnostics` would know what the prefix is. Anyways, I don't have a super strong opinion on this point.

ryanjulian:
What do the MuJoCo3M benchmarks look like?

src/garage/torch/algos/maml.py
```diff
@@ -63,7 +61,8 @@ def __init__(self,
         self._meta_evaluator = meta_evaluator
         self._policy = policy
         self._env = env
-        self._value_function = value_function
+        self._value_function = copy.deepcopy(inner_algo._value_function)
+        self._lame_vf_state = self._value_function.state_dict()
```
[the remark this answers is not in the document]
yonghyuc: @naeioi

src/garage/torch/algos/maml.py
```diff
@@ -150,6 +149,32 @@ def train_once(self, runner, all_samples, all_params):
...
+        # MAML resets a value function to its initial state before training.
+        self._value_function.load_state_dict(self._lame_vf_state)
+
+        obs = np.concatenate([path['observations'] for path in paths], axis=0)
+        returns = np.concatenate([path['returns'] for path in paths])
+        obs = torch.from_numpy(obs)
+        returns = torch.from_numpy(returns)
+
+        vf_loss = self._value_function.compute_loss(obs, returns)
+        self._inner_algo._vf_optimizer.zero_grad()
```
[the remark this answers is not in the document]
yonghyuc: Change to `vf_optimizer`

src/garage/torch/algos/vpg.py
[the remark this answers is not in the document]
yonghyuc: fixed

src/garage/torch/value_functions/base.py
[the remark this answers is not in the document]
yonghyuc: Every value function returns distribution object?
If I remove the `predict` function, I need to call `self._value_function(obs).mean.flatten(02)`
but what if there is a value function that directly returns baseline value instead of distribution object?

src/garage/torch/value_functions/base.py
```diff
@@ -0,0 +1,71 @@
+"""Base class for all baselines."""
+import abc
+
+
+class ValueFunction(abc.ABC):
```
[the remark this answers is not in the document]
yonghyuc: But GaussianMLPValueFunction inherits the `GaussianMLPModule` which inherit the `nn.Module`
Is it okay to inherit same class from two different parents?
Or do I need to make the base class to other directory?

src/garage/torch/value_functions/gaussian_mlp_value_function.py
```diff
@@ -0,0 +1,130 @@
...
+            obs (torch.Tensor): Observation from the environment
+                with shape :math:`(N \dot [T], O*)`.
+            returns (torch.Tensor): Acquired returns with shape :math:`(N, )`.
+
+        Returns:
+            torch.Tensor: Calculated negative mean scalar value of
+                objective (float).
+
+        """
+        dist = self.forward(obs)
```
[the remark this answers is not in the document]
yonghyuc: fixed

src/garage/torch/value_functions/base.py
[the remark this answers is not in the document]
yonghyuc: Currently we use the `name` field to log the value using `tabular.record`.
You can check the code below.
https://github.com/rlworkgroup/garage/blob/04841e31df14642a975354b306501da7088f3bd9/src/garage/torch/algos/vpg.py#L190

src/garage/torch/value_functions/base.py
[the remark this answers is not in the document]
yonghyuc: fixed

src/garage/torch/value_functions/base.py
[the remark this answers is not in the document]
yonghyuc: fixed

src/garage/torch/value_functions/gaussian_mlp_value_function.py
yonghyuc: fixed

src/garage/torch/value_functions/base.py
yonghyuc: fixed

src/garage/torch/algos/trpo.py
yonghyuc: I thought for TRPO, there is the `cg_iters` field for `ConjugateGradientOptimizer` and this works very similar with `max_optimization_epochs`.
Is it okay to use minibatch for ConjugateGradientOptimizer?
If so, it need to use same minibatch dataset but with different optimizing iterations? (`cg_iters` and `max_optimization_epochs`)?

src/garage/torch/algos/vpg.py
yonghyuc: fixed

src/garage/torch/value_functions/base.py
yonghyuc: fixed

yonghyuc:
This is PPO benchmark result for Mujoco1M (5.5e+5)

HalfCheetah-v2_benchmark
Hopper-v2_benchmark
InvertedPendulum-v2_benchmark
Reacher-v2_benchmark
Swimmer-v2_benchmark
Walker2d-v2_benchmark

yonghyuc:
This is TRPO benchmark result for Mujoco1M (5.5e+5)

HalfCheetah-v2_benchmark
Hopper-v2_benchmark
InvertedPendulum-v2_benchmark
Reacher-v2_benchmark
Swimmer-v2_benchmark
Walker2d-v2_benchmark

src/garage/torch/value_functions/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""Baselines (value functions) which use NumPy as a numerical backend."""
```
krzentner, line 1: This docstring should read `"""Value Function which use PyTorch."""` (Or something)

src/garage/torch/algos/trpo.py
ryanjulian: you can't use minibatching with ConjugateGradientOptimizer, or rather it would be kind of useless. Minibatching only works with SGD-based optimization algorithms, but a CG optimizer doesn't use SGD to choose parameters.

yonghyuc: Yes, so I used mini-batch only for value_function because default optimizer for value function is adam. When I run benchmark, I got a better result with minibatch and `optimization_epochs`

src/garage/torch/value_functions/__init__.py
yonghyuc: I missed this one. Thanks

src/garage/torch/algos/maml.py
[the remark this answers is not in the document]
krzentner: Conceptually it's the same, it's just that "lame" sounds a little unprofessional. Either way is okay, I guess.

src/garage/torch/algos/vpg.py
krzentner: You still have an extra line here, as far as I can tell.

yonghyuc: I'm sorry I don't know what you are referring to. Could you clarify what the extra line means?

yonghyuc: I'm sorry I don't know what you are referring to. Could you clarify what the extra line means?

krzentner: Okay, looks like the diff was just glitched on my end. Sorry for the confusion.

src/garage/torch/value_functions/base.py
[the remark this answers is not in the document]
krzentner: Okay, it sounds like everyone else finds name fields to be overall useful, so I'm convinced we should keep them.

src/garage/torch/algos/maml.py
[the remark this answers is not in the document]
ryanjulian: it's just confusing. use `old` or `previous`

src/garage/torch/algos/ppo.py
```diff
@@ -42,16 +49,18 @@ class PPO(VPG):
             the mean entropy to the surrogate objective. See
             https://arxiv.org/abs/1805.00909 for more details.
         minibatch_size (int): Batch size for optimization.
-        max_optimization_epochs (int): Maximum number of epochs for update.
+        optimization_epochs (int): Number of epochs for update.
```
ryanjulian, line 52: please continue to use max_optimization_epochs. there's no reason this algorithm can't use early stopping in the future.

src/garage/torch/algos/trpo.py
```diff
@@ -36,33 +48,38 @@ class TRPO(VPG):
             dense entropy to the reward for each time step. 'regularized' adds
             the mean entropy to the surrogate objective. See
             https://arxiv.org/abs/1805.00909 for more details.
+        minibatch_size (int): Batch size for optimization.
+        optimization_epochs (int): Number of epochs for update.
```
ryanjulian, line 52: max_optimization_epochs

src/garage/torch/algos/trpo.py
ryanjulian: what if i use the non-default ConjugateGradientOptimizer? will it still minibatch? wouldn't that be incorrect?

src/garage/torch/algos/vpg.py
```diff
@@ -73,8 +83,8 @@ def __init__(
             use_softplus_entropy=False,
             stop_entropy_gradient=False,
             entropy_method='no_entropy',
-            minibatch_size=None,
-            max_optimization_epochs=1,
+            minibatch_size=100,
+            optimization_epochs=10,
```
ryanjulian, line 87: please use max_optimization_epochs

src/garage/torch/algos/vpg.py
```diff
@@ -432,7 +483,7 @@ def process_samples(self, itr, paths):
         for path in paths:
             if 'returns' not in path:
                 path['returns'] = tu.discount_cumsum(path['rewards'],
-                                                     self.discount)
+                                                     self.discount).copy()
```
ryanjulian, line 486: why copy this?

src/garage/torch/algos/vpg.py
```diff
@@ -191,8 +205,36 @@ def train_once(self, itr, paths):
...
+        r"""Train the policy and value function with minibatch.
+
+        Args:
+            obs_flat (torch.Tensor): Observation from the environment
+                with shape :math:`(N, O*)`.
+            actions_flat (torch.Tensor): Actions fed to the environment
+                with shape :math:`(N, A*)`.
+            rewards_flat (torch.Tensor): Acquired rewards
+                with shape :math:`(N, )`.
+            returns_flat (torch.Tensor): Acquired returns
```
ryanjulian, line 219: why do you need returns from the CPU? we have a function to calculate returns in the GPU.

src/garage/torch/algos/vpg.py
```diff
@@ -191,8 +205,36 @@ def train_once(self, itr, paths):
...
+                with shape :math:`(N, )`.
+            advs_flat (torch.Tensor): Advantage value at each step
```
ryanjulian, line 221: why continue to calculate advantages in the CPU? the point of this change is that we an feedforward the value function and calculate the returns and advantages in the GPU.

src/garage/torch/algos/vpg.py
```diff
@@ -191,8 +205,36 @@ def train_once(self, itr, paths):
...
+
+        """
+        dataset = BatchDataset(
+            (obs_flat, actions_flat, rewards_flat, returns_flat, advs_flat),
+            self._minibatch_size)
+
+        for epoch in range(self._optimization_epochs):
+            for obs, actions, rewards, returns, advs in dataset.iterate():
+                policy_loss = self._train_policy(obs, actions, rewards, advs)
+                vf_loss = self._train_value_function(obs, returns)
```
ryanjulian, line 232: it's not clear that it's algorithmically valid to modify the value function during policy training. if you are feeding act/obs through the value function to calculate predicted returns, these predicted returns will start using information from the current batch to calculate predicted returns, which is not valid. the baseline should only have information about the batches prior to this one.

where did you get the idea of doing this (vs just optimizing it after the policy)?

src/garage/torch/algos/vpg.py
yonghyuc: If not, I got the error
`E       ValueError: some of the strides of a given numpy array are negative. This is currently not supported, but will be added in future releases.`
This is becaue we revert the order of output `[::-1]` in the `discount_cumsum` function.
So I need to copy the array

src/garage/torch/algos/vpg.py
yonghyuc: Can you clarify what is the meaning of "from the CPU"?
I just get values from `tensor_utils.discount_cumsum`, flatten the values, and make minibatch.
Do I need to make another function in PyTorch to compute the returns?

src/garage/torch/algos/vpg.py
yonghyuc: I don't know this is related to above comments what you post, but I thought every value (advantages, baselines, returns, ...) are computed at the beginning of the training process in the `process_samples` function. So even I update the value_function with the policy, I don't need to re-compute the values(advantages and baselines) and it would be okay.

src/garage/torch/algos/vpg.py
ryanjulian: How did it work before? What changed?

src/garage/torch/algos/vpg.py
```diff
@@ -432,7 +483,7 @@ def process_samples(self, itr, paths):
         for path in paths:
             if 'returns' not in path:
```
ryanjulian, line 484: in what circumstance is `'returns'` not in the path?

src/garage/torch/algos/vpg.py
ryanjulian: Can you use some other numpy function to get a reversed view, other than negative indexing?

When you call `copy()` you break the gradient path. If your returns have a differentiable components, e.g. a differentiable reward augmentation, your augmented rewards will no longer be differentiable.

src/garage/torch/algos/vpg.py
ryanjulian: ok second read, this seems probably okay to me as long as `tu.discount_cumsum` is making differentiable operations.

are you sure that if this algorithm were moved to the GPU, that all of these tensors would go with it?

src/garage/torch/algos/vpg.py
```diff
@@ -191,8 +205,36 @@ def train_once(self, itr, paths):
             discount=self.discount)
         return np.mean(undiscounted_returns)
 
-    def _train(self, obs, actions, rewards, advantages):
-        r"""Train the algorithm with minibatch.
+    def _train(self, obs_flat, actions_flat, rewards_flat, returns_flat,
```
ryanjulian, line 208: it's strange for this function to take rewards, returns, and advantages, because returns can always be calculated from rewards, and advantages can always be calculated using returns and the value_function

i was confused because we usually reserve `process_samples` for operations which happen in the CPU, and use `train` for operations which occur in the computation graph (e.g. the GPU if you are running in GPU mode). this class blurs the line, and it's difficult to tell where the np.ndarrays stop and the torch.Tensors start.

src/garage/torch/algos/vpg.py
yonghyuc: This error is from [here](https://github.com/rlworkgroup/garage/blob/a2207eeff9fe7ff84c9db68202ffa9f0f233ac22/src/garage/torch/algos/vpg.py#L503).
Previously, there is only LinearFeatureBaseline which doesn't have to convert numpy.array to torch.tensor. But now we need to convert numpy.array to torch.tensor for GaussianMLPValueFunction.

yonghyuc: According to this [post](https://discuss.pytorch.org/t/torch-from-numpy-not-support-negative-strides/3663/11), problem is in PyTorch. PyTorch doesn't support numpy.array with negative stride.
For the gradient path, the returns values from the `tu.discount_cumsum` is numpy.array not a torch.tensor, so I think there is no gradient path for the values yet. 

Also I'm just curious that we use rewards from environment to compute the return values and there is module or functions that compute gradient for the rewards?

src/garage/torch/algos/vpg.py
yonghyuc: As I understand, I have to convert tensors using CUDA commands like 
```
if torch.cuda.is_availabe():
    tensors = tensors.cuda()
```
Are you talking about this codes?

src/garage/torch/algos/vpg.py
ryanjulian: the point here is that the returns can be differentiable.

src/garage/torch/algos/maml.py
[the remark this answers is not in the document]
naeioi: Sorry I am late to the party. I made this name. This is not an old or previous value function, but a value function without any training. Now I think `initial_vf_state` is the most appropriate name.

src/garage/torch/algos/vpg.py
yonghyuc: I'll take out returns and baselines from the `process_samples` and put these in `train_once`.
Because it has to compute `returns` and `advantages` using the whole trajectories, these values should be computed before separating values into mini-batches.

yonghyuc:
I added new OptimizerWrapper class. 
The OptimizerWrapper gets optimizer type(torch.optim.optimizer) and other parameters for mini batch. So, it is similar with torch.optim.optimizer (`zero_grad`, `step`) but it has `get_minibatch` fucntion to provide batch data to algorithm. 

Overall process is
algo gets mini batch dataset from OptimizerWrapper
OptimizerWrapperzero_grad() ->  inner torch.optim.optimizer.zero_grad()
algo computes loss and backward it
OptimizerWrapper.step() -> inner torch.optim.optimizer.step()

src/garage/torch/algos/vpg.py
```diff
@@ -295,8 +332,8 @@ def _compute_advantage(self, rewards, valids, baselines):
 
         if self._center_adv:
             means = advantage_flat.mean()
-            variance = advantage_flat.var()
-            advantage_flat = (advantage_flat - means) / (variance + 1e-8)
+            std = advantage_flat.std()
```
ryanjulian, line 335: why this change? it seems unrelated to value functions

yonghyuc:
@naeioi Could you check the changes and how they affect MAML algorithm?
I changed the MAML following the changes but I am worried about missing part. 
You can check main changes [here](https://github.com/rlworkgroup/garage/pull/1317#issuecomment-620270310)
Thanks!

@krzentner Could you check this PR for me? Thanks!

src/garage/torch/algos/trpo.py
ryanjulian: i saw in the CHANGELOG that Torch 1.5 started making it easier to add custom optimizers, but didn't look in detail.

src/garage/torch/algos/vpg.py
ryanjulian: i am asking whether all of these operations are performed as part of the torch computation graph.

tests/benchmarks/garage/tf/algos/benchmark_ppo.py
```diff
@@ -264,49 +247,3 @@ def run_garage_tf(env, seed, log_dir):
         dowel_logger.remove_all()
 
         return tabular_log_file
-
-
-def run_baselines(env, seed, log_dir):
```
zequnyu, line 269: Do we want to delete this? This should be the only place we could locate the baselines code.

tests/benchmarks/garage/tf/algos/benchmark_trpo.py
```diff
@@ -218,59 +215,3 @@ def run_garage(env, seed, log_dir):
         dowel_logger.remove_all()
 
         return tabular_log_file
-
-
-def run_baselines(env, seed, log_dir):
```
zequnyu, line 223: Same here

src/garage/torch/algos/maml.py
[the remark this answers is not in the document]
naeioi: Can you change this to `_initial_vf_state `?

tests/benchmarks/garage/tf/algos/benchmark_ppo.py
yonghyuc: It looks like the openai.baseline version is not support Tensorflow 2.0.

tests/benchmarks/garage/tf/algos/benchmark_trpo.py
yonghyuc: It looks like the openai.baseline version is not support Tensorflow 2.0.

src/garage/torch/algos/maml_trpo.py
```diff
@@ -3,7 +3,8 @@
 
 from garage.torch.algos import _Default, VPG
 from garage.torch.algos.maml import MAML
-from garage.torch.optimizers import ConjugateGradientOptimizer
+from garage.torch.optimizers import ConjugateGradientOptimizer, \
```
ryanjulian, line 6: please just make these separate lines. \ are discourages by PEP8

ryanjulian:
@Mergifyio rebase

ryanjulian:
@Mergifyio rebase

PR #1319, opened 2020-04-18
Make copies of env in worker_factory

naeioi:
When a single object is passed to prepare_worker_messages(), it should
make a list of copies of that object, but not a list of references.

This is a bug that makes test_update_envs_env_update() flaky. Without
this patch, rollouts will be exactly the same, and np.var(mean_rewards)
will be a small number that is close to zero. (~1e-32)

### DIFF
--- a/src/garage/sampler/local_sampler.py
+++ b/src/garage/sampler/local_sampler.py
@@ -1,4 +1,5 @@
 """Sampler that runs workers in the main process."""
+import copy
 
 from garage import TrajectoryBatch
 from garage.sampler.sampler import Sampler
@@ -31,7 +32,8 @@ class LocalSampler(Sampler):
         # pylint: disable=super-init-not-called
         self._factory = worker_factory
         self._agents = worker_factory.prepare_worker_messages(agents)
-        self._envs = worker_factory.prepare_worker_messages(envs)
+        self._envs = worker_factory.prepare_worker_messages(
+            envs, preprocess=copy.deepcopy)
         self._workers = [
             worker_factory(i) for i in range(worker_factory.n_workers)
         ]
@@ -78,7 +80,8 @@ class LocalSampler(Sampler):
 
         """
         agent_updates = self._factory.prepare_worker_messages(agent_update)
-        env_updates = self._factory.prepare_worker_messages(env_update)
+        env_updates = self._factory.prepare_worker_messages(
+            env_update, preprocess=copy.deepcopy)
         for worker, agent_up, env_up in zip(self._workers, agent_updates,
                                             env_updates):
             worker.update_agent(agent_up)
--- a/src/garage/sampler/worker_factory.py
+++ b/src/garage/sampler/worker_factory.py
@@ -1,6 +1,4 @@
 """Worker factory used by Samplers to construct Workers."""
-import copy
-
 import psutil
 
 from garage.sampler.worker import DefaultWorker
@@ -87,8 +85,7 @@ class WorkerFactory:
                     'Length of list doesn\'t match number of workers')
             return [preprocess(obj) for obj in objs]
         else:
-            obj = preprocess(objs)
-            return [copy.deepcopy(obj) for _ in range(self.n_workers)]
+            return [preprocess(objs) for _ in range(self.n_workers)]
 
     def __call__(self, worker_number):
         """Construct a worker given its number.
--- a/src/garage/sampler/worker_factory.py
+++ b/src/garage/sampler/worker_factory.py
@@ -1,4 +1,6 @@
 """Worker factory used by Samplers to construct Workers."""
+import copy
+
 import psutil
 
 from garage.sampler.worker import DefaultWorker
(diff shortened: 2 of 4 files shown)


ahtsan:
I am curious whether the preferred usage is to pass a single environment object to `LocalSampler.from_worker_factory` and make n copies for the workers, or pass `task.sample(n)`. `SetTaskSampler` creates n `SetTaskUpdate`, which will also construct the environment. Are they the same?

Single environment object:
```
    env = TfEnv(PointEnv())
    tasks = SetTaskSampler(lambda: TfEnv(PointEnv()))
    n_workers = 8
    workers = WorkerFactory(seed=100,
                            max_path_length=max_path_length,
                            n_workers=n_workers)
    sampler = LocalSampler.from_worker_factory(workers, policy, env))
    rollouts = sampler.obtain_samples(0,
                                      161,
                                      np.asarray(policy.get_param_values()),
                                      env_update=tasks.sample(n_workers))
```

Passing EnvUpdate
```
    env = TfEnv(PointEnv())
    tasks = SetTaskSampler(lambda: TfEnv(PointEnv()))
    n_workers = 8
    workers = WorkerFactory(seed=100,
                            max_path_length=max_path_length,
                            n_workers=n_workers)
    sampler = LocalSampler.from_worker_factory(workers, policy, tasks.sample(n_workers))
    rollouts = sampler.obtain_samples(0,
                                      161,
                                      np.asarray(policy.get_param_values()))
```

src/garage/sampler/local_sampler.py
```diff
@@ -31,7 +32,8 @@ def __init__(self, worker_factory, agents, envs):
         # pylint: disable=super-init-not-called
         self._factory = worker_factory
         self._agents = worker_factory.prepare_worker_messages(agents)
-        self._envs = worker_factory.prepare_worker_messages(envs)
+        self._envs = worker_factory.prepare_worker_messages(
+            envs, preprocess=copy.deepcopy)
```
ryanjulian, line 36: seems like this actually might be a bug in the worker factory, no?

@krzentner PTAL

krzentner: I don't think this is a bug in worker factory. Then again, I don't quite understand why this change is necessary either. Only one environment instance gets used at a time, so it should be fine if they're actually all the same environment, right?

krzentner: > Passing EnvUpdate
> 
> ```
>     env = TfEnv(PointEnv())
>     tasks = SetTaskSampler(lambda: TfEnv(PointEnv()))
>     n_workers = 8
>     workers = WorkerFactory(seed=100,
>                             max_path_length=max_path_length,
>                             n_workers=n_workers)
>     sampler = LocalSampler.from_worker_factory(workers, policy, tasks.sample(n_workers))
>     rollouts = sampler.obtain_samples(0,
>                                       161,
>                                       np.asarray(policy.get_param_values()))
> ```

I think this should be the preferred usage, if you can sample *all* of the tasks at the start.
If you can't, you need to use the other one though.

krzentner:
> Passing EnvUpdate
> 
> ```
>     env = TfEnv(PointEnv())
>     tasks = SetTaskSampler(lambda: TfEnv(PointEnv()))
>     n_workers = 8
>     workers = WorkerFactory(seed=100,
>                             max_path_length=max_path_length,
>                             n_workers=n_workers)
>     sampler = LocalSampler.from_worker_factory(workers, policy, tasks.sample(n_workers))
>     rollouts = sampler.obtain_samples(0,
>                                       161,
>                                       np.asarray(policy.get_param_values()))
> ```

I think this should be the preferred usage, if you can sample *all* of the tasks at the start.
If you can't, you need to use the other one though.

src/garage/sampler/local_sampler.py
krzentner: The problem is that if we make the worker factory always return separate objects, then the other samplers will end up constructing 2x as many environment instances.

krzentner: I guess one option would be for `prepare_worker_messages` to always return a list of serialized messages (instead of whatever the preprocessing function is). This would break some optimizations in the `RaySampler` though.

naeioi:
@Mergifyio rebase

naeioi:
@Mergifyio rebase

ryanjulian:
What's the path forward for this PR? Is it still needed? Does the actual fix lie in changing the API?

naeioi:
@ryanjulian This PR fixes the flaky problems in `test_update_envs_env_update`. This test has a weak assertion on `np.var(mean_rewards) > 0`. It should be a small positive epsilon, but not 0. 

This PR does not change the API, but it does change the semantics. Under old semantic, if a single instance is given to the worker factory, the environments in workers are references to a single environment instance. And that will make different workers generate exactly the same samples, which is not desired and is wrong.

ryanjulian:
Okay. I'm basically waiting for @krzentner 's approval, since she knows the sampler API best.

krzentner:
I think this change is acceptable as-is. It theoretically makes the other samplers slower, but it fixes a rather significant bug and I know how to improve the performance in the future while keeping the bug fixed.

PR #1321, opened 2020-04-19
Make pytest output less verbose

gitanshu:
https://docs.pytest.org/en/latest/capture.html

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,24 +27,24 @@ ci-job-precommit: assert-docker docs
 
 ci-job-normal: assert-docker
 	[ ! -f $(MJKEY_PATH) ] || mv $(MJKEY_PATH) $(MJKEY_PATH).bak
-	pytest --cov=garage -v -m \
+	pytest --cov=garage -m \
 	    'not nightly and not huge and not benchmark and not flaky and not large and not mujoco' --durations=20
 	coverage xml
 	bash <(curl -s https://codecov.io/bash)
 
 ci-job-large: assert-docker
 	[ ! -f $(MJKEY_PATH) ] || mv $(MJKEY_PATH) $(MJKEY_PATH).bak
-	pytest --cov=garage -v -m 'large and not mujoco' --durations=20
+	pytest --cov=garage -m 'large and not mujoco' --durations=20
 	coverage xml
 	bash <(curl -s https://codecov.io/bash)
 
 ci-job-mujoco: assert-docker
-	pytest --cov=garage -v -m 'mujoco and not flaky' --durations=20
+	pytest --cov=garage -m 'mujoco and not flaky' --durations=20
 	coverage xml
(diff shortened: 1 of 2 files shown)


ryanjulian:
I enabled this flag intentionally, to (1) make the test output less chatty (it's already over 5k lines, which is a problem) and (2) prevent people from writing tests which rely on stdout (e.g. print() testing) on the CI.

Is there a reason you need this disabled? I was already contemplating making the current test output *less* verbose by removing the pass/fail line for each test.

ryanjulian:
These look much better -- maybe we can also skip printing the coverage report?

It's not super useful because we have multiple test jobs.

ryanjulian:
Sorry to keep piling on here -- did we change some flag about warning reporting?

Our warning reports don't show which test cases have warnings, which makes it difficult to resolve them when they come from general framework code.

The example shows each warning associated with a test: https://docs.pytest.org/en/latest/warnings.html

gitanshu:
> These look much better -- maybe we can also skip printing the coverage report?
Looking into this

> did we change some flag about warning reporting?
looks like this has always been the behavior for us. We don't change anything regarding warnings anywhere though :thinking:

gitanshu:
@ryanjulian this is warnings summary without xdist: https://travis-ci.com/github/rlworkgroup/garage/jobs/321427488#L1444
VS
this is with xdist
https://travis-ci.com/github/rlworkgroup/garage/jobs/321419199#L1320

> Our warning reports don't show which test cases have warnings 

Seems like warnings not showing the test name is a side effect of pytest-xdist :crying_cat_face:

Makefile
```diff
@@ -27,24 +27,23 @@ ci-job-precommit: assert-docker docs
 
 ci-job-normal: assert-docker
 	[ ! -f $(MJKEY_PATH) ] || mv $(MJKEY_PATH) $(MJKEY_PATH).bak
-	pytest --cov=garage -v -m \
-	    'not nightly and not huge and not benchmark and not flaky and not large and not mujoco' --durations=0
+	pytest --cov=garage --cov-report=xml -m \
+	    'not nightly and not huge and not benchmark and not flaky and not large and not mujoco' --durations=20
 	coverage xml
```
gitanshu, line 32: Remove this

Makefile
```diff
@@ -27,24 +27,23 @@ ci-job-precommit: assert-docker docs
...
 	bash <(curl -s https://codecov.io/bash)
 
 ci-job-large: assert-docker
 	[ ! -f $(MJKEY_PATH) ] || mv $(MJKEY_PATH) $(MJKEY_PATH).bak
-	pytest --cov=garage -v -m 'large and not mujoco' --durations=0
+	pytest --cov=garage --cov-report=xml -m 'large and not mujoco' --durations=20
 	coverage xml
```
gitanshu, line 38: Remove this

PR #1323, opened 2020-04-20
Add empty mjkey.txt to non-MuJoCo test jobs

ryanjulian:
This should silence warnings during test collection.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,14 +26,14 @@ ci-job-precommit: assert-docker docs
 	scripts/travisci/check_precommit.sh
 
 ci-job-normal: assert-docker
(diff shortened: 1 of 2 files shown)


gitanshu:
Yeah I should really put this in comment that the reason we remove mujoco key instead of leaving it blank, which was the old behaviour, is because mujoco throws a warning if key is not found, but if there's an invalid key file, it just kills the process.

ryanjulian:
Hmm maybe we can instead just monkeypatch the part of mujoco_py which prints this on import, or moneypatch all of mujoco_py out of existence for the non-mujoco jobs:
https://docs.pytest.org/en/latest/monkeypatch.html#global-patch-example-preventing-requests-from-remote-operations

Issue #1204, opened 2020-03-11
Metaworld examples for MTSAC


ryanjulian:

ryanjulian:
@avnishn is this done yet?

PR #1332, opened 2020-04-21
MTSAC and examples

avnishn:
Addition of Multitask SAC and examples on MT10 and MT50 #1204

Addresses #1322 to remove metaworld specific code from MultienvWrapper

### DIFF
--- a/src/garage/envs/multi_env_wrapper.py
+++ b/src/garage/envs/multi_env_wrapper.py
@@ -64,13 +64,17 @@ class MultiEnvWrapper(gym.Wrapper):
             - 'del-onehot' assumes an one-hot task id is appended to
               observation, and it excludes that.
               Use case: metaworld environments with Task Embedding.
+        env_names (list(str)): The names of the environments corresponding to
+            envs. The index of an env_name must correspond to the index of the
+            corresponding env in envs. An env_name in env_names must be unique.
 
     """
 
     def __init__(self,
                  envs,
                  sample_strategy=uniform_random_strategy,
-                 mode='add-onehot'):
+                 mode='add-onehot',
+                 env_names=None):
         assert mode in ['vanilla', 'add-onehot', 'del-onehot']
 
         self._sample_strategy = sample_strategy
@@ -82,6 +86,12 @@ class MultiEnvWrapper(gym.Wrapper):
             if not isinstance(env, GarageEnv):
                 envs[i] = GarageEnv(env)
         super().__init__(envs[0])
+        if env_names is not None:
+            assert isinstance(env_names, list), 'env_names must be a list'
+            msg = ('env_names are not unique or there is not an env_name',
+                   'corresponding to each env in envs')
+            assert len(set(env_names)) == len(envs), msg
+        self._env_names = env_names
         self._task_envs = []
         for env in envs:
             if (env.observation_space.shape !=
@@ -91,8 +101,19 @@ class MultiEnvWrapper(gym.Wrapper):
             if env.action_space.shape != self.env.action_space.shape:
                 raise ValueError('Action space of all envs should be same.')
             self._task_envs.append(env)
-
         self.env.spec.observation_space = self.observation_space
+        self._spec = self.env.spec
+
+    @property
+    def spec(self):
+        """Describes the action and observation spaces of the wrapped envs.
+
+        Returns:
+            garage.envs.EnvSpec: the action and observation spaces of the
+                wrapped environments.
+
+        """
+        return self._spec
 
     @property
     def num_tasks(self):
@@ -124,6 +145,11 @@ class MultiEnvWrapper(gym.Wrapper):
             int: Index of active task.
 
         """
+        if hasattr(self.env, 'active_task_index'):
+            return self.env.active_task_index
+        else:
+            return self._active_task_index
+
         return self._active_task_index
 
     @property
@@ -156,8 +182,7 @@ class MultiEnvWrapper(gym.Wrapper):
         """
         self._observation_space = observation_space
 
-    @property
-    def active_task_one_hot(self):
+    def _active_task_one_hot(self):
         """One-hot representation of active task.
 
         Returns:
@@ -186,7 +211,7 @@ class MultiEnvWrapper(gym.Wrapper):
         if self._mode == 'vanilla':
             return obs
         elif self._mode == 'add-onehot':
-            return np.concatenate([obs, self.active_task_one_hot])
+            return np.concatenate([obs, self._active_task_one_hot()])
         else:  # self._mode == 'del-onehot'
             return obs[:-self._num_tasks]
 
@@ -205,30 +230,16 @@ class MultiEnvWrapper(gym.Wrapper):
         """
         obs, reward, done, info = self.env.step(action)
         if self._mode == 'add-onehot':
-            obs = np.concatenate([obs, self.active_task_one_hot])
+            obs = np.concatenate([obs, self._active_task_one_hot()])
         elif self._mode == 'del-onehot':
             obs = obs[:-self._num_tasks]
-        info['task_id'] = self._active_task_index
-        info['task_name'] = self.active_task_name
+        if 'task_id' not in info:
+            info['task_id'] = self._active_task_index
+        if self._env_names is not None:
+            info['task_name'] = self._env_names[self._active_task_index]
         return obs, reward, done, info
 
     def close(self):
         """Close all task envs."""
         for env in self._task_envs:
             env.close()
-
-    @property
-    def active_task_name(self):
-        """Return the name of the active task.
-
-        Returns:
-            str: The name of the task. If wrapping metaworld benchmark,
-                environments, this will be the name of the environment,
-                given by metaworld. Otherwise, the name is the unwrapped
-                environment's class name.
-
-        """
-        if hasattr(self.env, 'all_task_names'):
-            return self.env.all_task_names[0]
-        else:
-            return str(self.env.unwrapped)
--- a/src/garage/replay_buffer/base.py
+++ b/src/garage/replay_buffer/base.py
@@ -1,5 +1,4 @@
-"""
-This module implements a replay buffer memory.
+"""This module implements a replay buffer memory.
 
 Replay buffer is an important technique in reinforcement learning. It
 stores transitions in a memory buffer of fixed size. When the buffer is
@@ -16,8 +15,7 @@ import numpy as np
 
 
 class ReplayBuffer(metaclass=abc.ABCMeta):
-    """
-    Abstract class for Replay Buffer.
+    """Abstract class for Replay Buffer.
 
     Args:
         env_spec (garage.envs.EnvSpec): Environment specification.
@@ -26,6 +24,7 @@ class ReplayBuffer(metaclass=abc.ABCMeta):
     """
 
     def __init__(self, env_spec, size_in_transitions, time_horizon):
+        del env_spec
         self._current_size = 0
         self._current_ptr = 0
         self._n_transitions_stored = 0
@@ -42,7 +41,7 @@ class ReplayBuffer(metaclass=abc.ABCMeta):
         rollout_batch_size = len(episode_buffer['observation'])
         idx = self._get_storage_idx(rollout_batch_size)
 
-        for key in self._buffer.keys():
+        for key in self._buffer:
             self._buffer[key][idx] = episode_buffer[key]
         self._n_transitions_stored = min(
             self._size_in_transitions, self._n_transitions_stored +
@@ -50,17 +49,27 @@ class ReplayBuffer(metaclass=abc.ABCMeta):
 
     @abstractmethod
     def sample(self, batch_size):
-        """Sample a transition of batch_size."""
+        """Sample a transition of batch_size.
+
+        Args:
+            batch_size(int): The number of transitions to be sampled.
+
+        """
         raise NotImplementedError
 
     def add_transition(self, **kwargs):
-        """Add one transition into the replay buffer."""
+        """Add one transition into the replay buffer.
+
+        Args:
+            kwargs (dict(str, [numpy.ndarray])): Dictionary that holds
+                the transitions.
+
+        """
         transition = {k: [v] for k, v in kwargs.items()}
         self.add_transitions(**transition)
 
     def add_transitions(self, **kwargs):
-        """
-        Add multiple transitions into the replay buffer.
+        """Add multiple transitions into the replay buffer.
 
         A transition contains one or multiple entries, e.g.
         observation, action, reward, terminal and next_observation.
@@ -71,6 +80,7 @@ class ReplayBuffer(metaclass=abc.ABCMeta):
         Args:
             kwargs (dict(str, [numpy.ndarray])): Dictionary that holds
                 the transitions.
+
         """
         if not self._initialized_buffer:
             self._initialize_buffer(**kwargs)
@@ -80,7 +90,7 @@ class ReplayBuffer(metaclass=abc.ABCMeta):
 
         if len(self._episode_buffer['observation']) == self._time_horizon:
             self.store_episode()
-            for key in self._episode_buffer.keys():
+            for key in self._episode_buffer:
                 self._episode_buffer[key].clear()
 
     def _initialize_buffer(self, **kwargs):
@@ -93,7 +103,16 @@ class ReplayBuffer(metaclass=abc.ABCMeta):
         self._initialized_buffer = True
 
     def _get_storage_idx(self, size_increment=1):
-        """Get the storage index for the episode to add into the buffer."""
+        """Get the storage index for the episode to add into the buffer.
+
+        Args:
+            size_increment(int): The number of storage indeces that new
+                transitions will be placed in.
+
+        Returns:
+            numpy.ndarray: The indeces to store size_incremente transitions at.
+
+        """
         if self._current_size + size_increment <= self._size:
             idx = np.arange(self._current_size,
                             self._current_size + size_increment)
@@ -125,30 +144,39 @@ class ReplayBuffer(metaclass=abc.ABCMeta):
         return idx
 
     def _convert_episode_to_batch_major(self):
-        """
-        Convert the shape of episode_buffer.
+        """Convert the shape of episode_buffer.
 
         episode_buffer: {time_horizon, algo.rollout_batch_size, flat_dim}.
         buffer: {size, time_horizon, flat_dim}.
+
+        Returns:
+            dict: Transitions that have been formated to fit properly in this
+                replay buffer.
+
         """
         transitions = {}
-        for key in self._episode_buffer.keys():
+        for key in self._episode_buffer:
             val = np.array(self._episode_buffer[key])
             transitions[key] = val.swapaxes(0, 1)
         return transitions
 
     @property
     def full(self):
-        """Whether the buffer is full."""
+        """Whether the buffer is full.
+
+        Returns:
+            bool: True of the buffer has reachd its maximum size.
+                False otherwise.
+
+        """
         return self._current_size == self._size
 
     @property
     def n_transitions_stored(self):
-        """
-        Return the size of the replay buffer.
+        """Return the size of the replay buffer.
 
         Returns:
-            self._size: Size of the current replay buffer.
+            int: Size of the current replay buffer.
 
         """
         return self._n_transitions_stored
--- a/src/garage/replay_buffer/path_buffer.py
+++ b/src/garage/replay_buffer/path_buffer.py
@@ -190,3 +190,13 @@ class PathBuffer:
             first, second = seg_b, seg_a
         assert first.start <= second.start
         return first.stop >= second.start
+
+    @property
+    def n_transitions_stored(self):
+        """Return the size of the replay buffer.
+
+        Returns:
+            int: Size of the current replay buffer.
+
+        """
+        return int(self._transitions_stored)
--- a/src/garage/torch/algos/mtsac.py
+++ b/src/garage/torch/algos/mtsac.py
@@ -1,0 +1,211 @@
+"""This modules creates a MTSAC model in PyTorch."""
+import numpy as np
+import torch
+
+from garage import log_multitask_performance, TrajectoryBatch
+from garage.torch.algos import SAC
+import garage.torch.utils as tu
+
+
+class MTSAC(SAC):
+    """A MTSAC Model in Torch.
+
+    This MTSAC implementation uses is the same as SAC except for a small change
+    called "disentangled alphas". Alpha is the entropy coefficient that is used
+    to control exploration of the agent/policy. Disentangling alphas refers to
+    having a separate alpha coefficients for every task learned by the policy.
+    The alphas are accessed by using a the one-hot encoding of an id that is
+    assigned to each task.
+
+    Args:
+        policy (garage.torch.policy.Policy): Policy/Actor/Agent that is being
+            optimized by SAC.
+        qf1 (garage.torch.q_function.ContinuousMLPQFunction): QFunction/Critic
+            used for actor/policy optimization. See Soft Actor-Critic and
+            Applications.
+        qf2 (garage.torch.q_function.ContinuousMLPQFunction): QFunction/Critic
+            used for actor/policy optimization. See Soft Actor-Critic and
+            Applications.
+        replay_buffer (garage.replay_buffer.ReplayBuffer): Stores transitions
+            that are previously collected by the sampler.
+        env_spec (garage.envs.env_spec.EnvSpec): The env_spec attribute of the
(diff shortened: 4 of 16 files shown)


src/garage/envs/multi_env_wrapper.py
```diff
@@ -192,9 +199,9 @@ def step(self, action):
 
         """
         obs, reward, done, info = self.env.step(action)
-        if not self._metaworld_mt:
+        if not self._has_one_hots:
             obs = self._obs_with_one_hot(obs)
-        info['task_id'] = self._active_task_index
+        info['task_id'] = self.active_task_index
```
krzentner, line 204: This should be under `if not self._has_task_id`, right?

krzentner, line 204: Are these environment infos already present inside metaworld? If so, we probably shouldn't add them.

src/garage/envs/multi_env_wrapper.py
```diff
@@ -227,6 +234,6 @@ def active_task_name(self):
                 environment's class name.
 
         """
-        if self._metaworld_mt:
+        if self._has_name:
```
krzentner, line 237: Again, we don't even need this method. Feel Free to just delete it instead.

examples/torch/mtsac_MT10.py
```diff
@@ -0,0 +1,100 @@
...
+        seed (int): Used to seed the random number generator to produce
+            determinism.
+        use_gpu (bool): Used to enable ussage of GPU in training.
+        _gpu (int): The ID of the gpu (used on multi-gpu machines).
+
+    """
+    deterministic.set_seed(seed)
+    runner = LocalRunner(ctxt)
+    task_names = MT10.get_train_tasks().all_task_names
+    mt10_train_envs = MultiEnvWrapper([
```
ryanjulian, line 42: can you manually format this so it looks less strange? or just construct your list of envs in a different block before this one.

examples/torch/mtsac_MT10.py
```diff
@@ -0,0 +1,100 @@
...
+                                 hidden_nonlinearity=F.relu)
+
+    replay_buffer = SimpleReplayBuffer(env_spec=mt10_train_envs.spec,
+                                       size_in_transitions=int(1e6),
+                                       time_horizon=1)
+
+    timesteps = 20000000
+    batch_size = int(150 * mt10_train_envs.num_tasks)
+    num_evaluation_points = 500
+    epochs = timesteps // batch_size
```
ryanjulian, line 76: to simplify the example, just pass these numbers as literals rather than doing this calculation.

examples/torch/mtsac_MT10.py
```diff
@@ -0,0 +1,100 @@
+"""MTSAC implementation based on Metaworld. Benchmarked on MT10.
```
ryanjulian, line 1: please remove the capital letters from these example names.

examples/torch/mtsac_MT10.py
```diff
@@ -0,0 +1,100 @@
...
+from garage.torch.algos import MTSAC
+from garage.torch.policies import TanhGaussianMLPPolicy
+from garage.torch.q_functions import ContinuousMLPQFunction
+import garage.torch.utils as tu
+
+
+@click.command()
+@click.option('--seed', 'seed', type=int, default=1)
+@click.option('--use_gpu', 'use_gpu', type=bool, default=False)
+@click.option('--gpu', '_gpu', type=int, default=0)
```
ryanjulian, line 25: Why not default `--gpu` to `None` and then only assign it to a gpu if it is non-None?

src/garage/envs/multi_env_wrapper.py
```diff
@@ -52,22 +52,29 @@ class MultiEnvWrapper(gym.Wrapper):
             A list of objects implementing gym.Env.
         sample_strategy (function(int, int)):
             Sample strategy to be used when sampling a new task.
-        metaworld_mt (bool): Set if the envs that are passed Metaworld
-            environments from Multitask Benchmark Environments. Primarily
-            disables wrapper from augmenting observations with task one-hot
-            ids.
+        has_one_hots (bool): If all envs have a provide their own one-hot ids
```
ryanjulian, line 55: i still don't understand what's going on here. these seem to specify details of what's being passed to the wrapper, rather than what the wrapper should do.

if the wrapper needs to know something about something it is passed, it should introspect that.

do we really need 3 booleans to decide whether to concatenate the onehot or not???

src/garage/torch/algos/multi_task_sac.py
```diff
@@ -0,0 +1,246 @@
+"""This modules creates a MTSAC model in PyTorch."""
```
ryanjulian, line 1: if the name of the main class is MTSAC, the name of the module should be mtsac.py

src/garage/torch/algos/multi_task_sac.py
```diff
@@ -0,0 +1,246 @@
...
+            policy_lr=policy_lr,
+            qf_lr=qf_lr,
+            reward_scale=reward_scale,
+            optimizer=optimizer,
+            steps_per_epoch=steps_per_epoch,
+            num_evaluation_trajs=num_evaluation_trajs,
+            eval_env=eval_envs)
+        self._num_tasks = num_tasks
+        self._eval_envs = eval_envs
+        if self.use_automatic_entropy_tuning and not alpha:
```
ryanjulian, line 121: why is this public?

src/garage/torch/algos/multi_task_sac.py
```diff
@@ -0,0 +1,246 @@
...
+
+    def optimize_policy(self, itr, samples_data):
+        """Optimize the policy q_functions, and temperature coefficient.
+
+        Args:
+            itr (int): Iterations.
+            samples_data (list): Processed batch data.
+
+        Returns:
+            torch.Tensor: loss from actor/policy network after optimization.
```
ryanjulian, line 166: please provide shape information for all tensor inputs/outputs on public functions

src/garage/torch/algos/multi_task_sac.py
```diff
@@ -0,0 +1,246 @@
...
+
+        Returns:
+            torch.Tensor: log_alpha
+
+        """
+        obs = kwargs['obs']
+        log_alpha = self.log_alpha
+        one_hots = obs[:, :self._num_tasks]
+        ret = torch.mm(one_hots, log_alpha.unsqueeze(0).t()).squeeze()
+        assert ret.size() == torch.Size([self.buffer_batch_size])
```
ryanjulian, line 155: in what circumstance is this assert false?

src/garage/torch/algos/sac.py
```diff
@@ -69,6 +69,10 @@ class SAC(OffPolicyRLAlgorithm):
         optimizer(torch.optim): optimizer to be used for policy/actor,
             q_functions/critics, and temperature/entropy optimizations.
         steps_per_epoch (int): Number of train_once calls per epoch.
+        num_evaluation_trajs(int): The number of evaluation trajectories
+            used for computing eval stats at the end of every epoch.
+        eval_env(garage.envs.GarageEnv): environment(s) used for collecting
```
ryanjulian, line 74: if it can be plural, does that mean this also accepts lists?

src/garage/torch/algos/sac.py
```diff
@@ -69,6 +69,10 @@ class SAC(OffPolicyRLAlgorithm):
         optimizer(torch.optim): optimizer to be used for policy/actor,
             q_functions/critics, and temperature/entropy optimizations.
         steps_per_epoch (int): Number of train_once calls per epoch.
+        num_evaluation_trajs(int): The number of evaluation trajectories
```
ryanjulian, line 72: please always use unabbreviated words

src/garage/torch/algos/sac.py
```diff
@@ -231,24 +237,32 @@ def _get_log_alpha(self, **kwargs):
...
+            alpha_loss = (-(self._get_log_alpha(**get_log_alpha_kwargs)) *
                           (log_pi.detach() + self.target_entropy)).mean()
         return alpha_loss
 
-    def _actor_objective(self, obs, new_actions, log_pi_new_actions):
+    def _actor_objective(self,
+                         obs,
+                         new_actions,
+                         log_pi_new_actions,
+                         get_log_alpha_kwargs=None):
```
ryanjulian, line 265: what is the purpose of get_log_alpha_kwargs? this seems like something the function should be determining. how the hell would a caller know what the kwargs to a subroutine should be???

src/garage/torch/algos/sac.py
```diff
@@ -260,24 +274,30 @@ def _actor_objective(self, obs, new_actions, log_pi_new_actions):
...
                 actions on the TanhNormal distributions that they were sampled
                 from. Shape is (1, buffer_batch_size).
+            get_log_alpha_kwargs(dict): keyword args that can be used in
+                retrieving the log_alpha parameter.
 
         Returns:
             torch.Tensor: loss from the Policy/Actor.
 
         """
+        if not get_log_alpha_kwargs:
```
ryanjulian, line 284: same question as before. how on earth would a caller know what kwargs to provide to a subroutine of this function?

examples/torch/mtsac_MT10.py
ryanjulian, line 1: you should probably put metaworld in the name of this example file

examples/torch/mtsac_MT50.py
```diff
@@ -0,0 +1,100 @@
+"""MTSAC implementation based on Metaworld. Benchmarked on MT50.
```
ryanjulian, line 1: please put metaworld in the name of this example file

src/garage/torch/algos/sac.py
```diff
@@ -69,6 +69,10 @@ class SAC(OffPolicyRLAlgorithm):
         optimizer(torch.optim): optimizer to be used for policy/actor,
             q_functions/critics, and temperature/entropy optimizations.
         steps_per_epoch (int): Number of train_once calls per epoch.
+        num_evaluation_trajs(int): The number of evaluation trajectories
+            used for computing eval stats at the end of every epoch.
+        eval_env(garage.envs.GarageEnv): environment(s) used for collecting
+            evaluation trajectories. If None, a copy of the train env is used.
```
ryanjulian, line 75: Please make the english in this docstring consistent -- capitalization, etc.

src/garage/torch/algos/sac.py
```diff
@@ -69,6 +69,10 @@ class SAC(OffPolicyRLAlgorithm):
         optimizer(torch.optim): optimizer to be used for policy/actor,
```
ryanjulian, line 69: torch.optim is a package, not a type.

ryanjulian: please correct this error throughout the docstring. it's present in several places (replay buffer, q_function, etc.)

src/garage/torch/algos/sac.py
```diff
@@ -260,24 +274,30 @@ def _actor_objective(self, obs, new_actions, log_pi_new_actions):
...
-            alpha = self._get_log_alpha().exp()
+            alpha = self._get_log_alpha(**get_log_alpha_kwargs).exp()
         min_q_new_actions = torch.min(self.qf1(obs, new_actions),
                                       self.qf2(obs, new_actions))
         policy_objective = ((alpha * log_pi_new_actions) -
                             min_q_new_actions.flatten()).mean()
         return policy_objective
 
-    def _critic_objective(self, samples):
+    def _critic_objective(self, samples, get_log_alpha_kwargs=None):
```
ryanjulian, line 294: same question as before. how on earth would a caller know what kwargs to provide to a subroutine of this function?

src/garage/torch/algos/sac.py
```diff
@@ -367,6 +389,27 @@ def optimize_policy(self, itr, samples_data):
...
+            epoch(int): The current training epoch.
+
+        Returns:
+            float: The average return across self._num_evaluation_trajs
+                trajectories
+
+        """
+        last_return = log_performance(
+            epoch,
+            self._obtain_evaluation_samples(
```
ryanjulian, line 408: this indentation is tortured. just retreive the samples in a different block -- it costs nothing.

src/garage/torch/algos/sac.py
```diff
@@ -377,7 +420,7 @@ def _log_statistics(self, policy_loss, qf1_loss, qf2_loss):
 
         """
         with torch.no_grad():
-            tabular.record('alpha', self.log_alpha.exp().item())
+            tabular.record('alpha', self.log_alpha.exp().mean().item())
         tabular.record('policy_loss', policy_loss.item())
         tabular.record('qf_loss/{}'.format('qf1_loss'), float(qf1_loss))
```
ryanjulian, line 425: please make these tabular logs consistent with the rest of the codebase, i.e. use CamelCase and scope the various statistics within their respective primitives QFunction/Loss, etc.

src/garage/torch/algos/multi_task_sac.py
```diff
@@ -0,0 +1,246 @@
...
+        return policy_loss, qf1_loss, qf2_loss
+
+    def _evaluate_policy(self, epoch):
+        """Evaluate the performance of the policy via deterministic rollouts.
+
+            Statistics such as (average) discounted return and success rate are
+            recorded.
+
+        Args:
+            epoch(int): The current training epoch.
```
ryanjulian, line 214: note that in the format there's a space between then name and type. i'm not sure if the parser correctly handles no-space. you can check.

src/garage/torch/algos/sac.py
```diff
@@ -231,24 +237,32 @@ def _get_log_alpha(self, **kwargs):
         log_alpha = self.log_alpha
```
ryanjulian, line 237: why not just `return self.log_alpha`?

src/garage/torch/algos/multi_task_sac.py
```diff
@@ -0,0 +1,246 @@
...
+        Args:
+            kwargs(dict): keyword args that can be used in retrieving the
+                log_alpha parameter. Unused here.
+
+        Returns:
+            torch.Tensor: log_alpha
+
+        """
+        obs = kwargs['obs']
+        log_alpha = self.log_alpha
```
ryanjulian, line 152: what is the purpose of this line?

src/garage/envs/multi_env_wrapper.py
ryanjulian: if the task is already in env_infos, why even bother with using this wrapper with MTSAC? Why append a task one-hot here, only to unpack it from obs later in the algorithm?

krzentner: The point of this wrapper, as far as I'm concerned, is to automatically wrap a list of environments so that a naive single-task algorithm can be used as a multi-task algorithm, without any code changes.

ryanjulian: i agree -- that's why i think there's no reason to use it if there's going to be something called "MTSAC" -- if it's MT, it should just handle multiple envs natively.

krzentner: That's reasonable, but people also want to be able to use single-task algorithms with multiple envs (e.g. MT-PPO and MT-TRPO), which I think is the main reason this wrapper exists. Admittedly, MT-SAC is a weird case of an algorithm half-way adapted to multi-tasking.

krzentner: I still think it's best not to override the existing `task_id`.

examples/torch/mtsac_MT10.py
ryanjulian: e.g. `mtsac_metaworld_mt10.py`

src/garage/replay_buffer/base.py
```diff
@@ -37,13 +37,17 @@ def __init__(self, env_spec, size_in_transitions, time_horizon):
...
         rollout_batch_size = len(episode_buffer['observation'])
         idx = self._get_storage_idx(rollout_batch_size)
 
-        for key in self._buffer.keys():
-            self._buffer[key][idx] = episode_buffer[key]
+        try:
+            for key in self._buffer.keys():
+                self._buffer[key][idx] = episode_buffer[key]
+        except:
+            import ipdb; ipdb.set_trace()
```
ryanjulian, line 50: please remove this

src/garage/replay_buffer/path_buffer.py
```diff
@@ -190,3 +190,14 @@ def _segments_overlap(seg_a, seg_b):
...
+    @property
+    def n_transitions_stored(self):
+        """
+        Return the size of the replay buffer.
+
+        Returns:
+            int: Size of the current replay buffer.
+
+        """
+        return int(self._transitions_stored)
```
ryanjulian, line 203: why does it need to be cast??

src/garage/torch/algos/mtsac.py
```diff
@@ -0,0 +1,208 @@
...
+import numpy as np
+import torch
+
+from garage import log_multitask_performance, TrajectoryBatch
+from garage.torch.algos import SAC
+import garage.torch.utils as tu
+
+
+class MTSAC(SAC):
+    """A MTSAC Model in Torch.
```
ryanjulian, line 11: it's not a model though?

src/garage/torch/algos/mtsac.py
```diff
@@ -0,0 +1,208 @@
...
+            from Soft Actor-Critic Algorithms and Applications is used.
+        initial_log_entropy(float): initial entropy/temperature coefficient
+            to be used if use_automatic_entropy_tuning is True.
+        discount(float): Discount factor to be used during sampling and
+            critic/q_function optimization.
+        buffer_batch_size(int): The number of transitions sampled from the
+            replay buffer that are used during a single optimization step.
+        min_buffer_size(int): The minimum number of transitions that need to be
+            in the replay buffer before training can begin.
+        target_update_tau(float): coefficient that controls the rate at which
```
ryanjulian, line 56: Please make sure these are all properly-formed english phrases (capitalization rules, etc)

src/garage/torch/algos/mtsac.py
```diff
@@ -0,0 +1,208 @@
...
+        for env in self._eval_env:
+            eval_trajs.append(
+                self._obtain_evaluation_samples(
+                    env, num_trajs=self._num_evaluation_trajectories))
+        eval_trajs = TrajectoryBatch.concatenate(*eval_trajs)
+        last_return = log_multitask_performance(epoch, eval_trajs,
+                                                self.discount)
+        return last_return
+
+    def to(self, device=None):
```
ryanjulian, line 193: @krzentner is there a way we can easily make this available to all torch algorithms?

src/garage/torch/algos/sac.py
```diff
@@ -205,92 +214,144 @@ def train_once(self, itr=None, paths=None):
...
         access to a modified log_alpha, such as multi_task sac.
 
-        Args:
-            kwargs(dict): keyword args that can be used in retrieving the
-                log_alpha parameter. Unused here.
+        Note:
+            samples_data's entries should be torch.Tensor's with the following
+            shapes:
+                observation:
+                    (sac.env_spec.observation_space, sac.buffer_batch_size)
```
ryanjulian, line 240: Please use the style for documenting shapes established in `src/garage/_dtypes.py`

src/garage/torch/algos/sac.py
```diff
@@ -377,12 +473,13 @@ def _log_statistics(self, policy_loss, qf1_loss, qf2_loss):
...
         """
         with torch.no_grad():
-            tabular.record('alpha', self.log_alpha.exp().item())
-        tabular.record('policy_loss', policy_loss.item())
-        tabular.record('qf_loss/{}'.format('qf1_loss'), float(qf1_loss))
-        tabular.record('qf_loss/{}'.format('qf2_loss'), float(qf2_loss))
-        tabular.record('buffer_size', self.replay_buffer.n_transitions_stored)
-        tabular.record('local/average_return', np.mean(self.episode_rewards))
+            tabular.record('Train/alpha', self.log_alpha.exp().mean().item())
+        tabular.record('Train/PolicyLoss', policy_loss.item())
```
ryanjulian, line 477: Please use the scheme for logging keys/values used throughout the codebase, e.g. `{AlgoOrModelName}/Quantity/MaybeSubKey`

src/garage/torch/algos/mtsac.py
```diff
@@ -0,0 +1,208 @@
...
+        qf2(garage.torch.q_function.ContinuousMLPQFunction): QFunction/Critic
+            used for actor/policy optimization. See Soft Actor-Critic and
+            Applications.
+        replay_buffer(garage.replay_buffer.ReplayBuffer): Stores transitions
+            that are previously collected by the sampler.
+        env_spec(garage.envs.env_spec.EnvSpec): The env_spec attribute of the
+            environment that the agent is being trained in. Usually accessable
+            by calling env.spec.
+        num_tasks(int): The number of tasks being learned.
+        max_path_length(int): Max path length of the environment.
```
ryanjulian, line 35: this is a property of the algorithm, not the environment.

src/garage/torch/algos/sac.py
avnishn: In terms of organization with respect to the reset of the data that is recorded, this organization schema makes more sense. The Train/ stub helps makes the distinction between statistics collected at train time and at evaluation time.

ryanjulian: You can either rewrite the rest of the codebase to use your scheme, our you can adopt the one I describe here to stay consistent.

src/garage/torch/algos/mtsac.py
krzentner: This mostly "just works" if you just use parameters, but still requires some logic to re-target the optimizer. It's about all we can do for now without re-designing the optimizer API.

avnishn:
@mergifyio rebase

setup.cfg
```diff
@@ -15,6 +15,7 @@ per-file-ignores =
     examples/torch/*:N812
     src/garage/torch/*:N812,D403
     tests/garage/torch/*:N812,D
+    tests/integration_tests/test_examples.py:D
```
ryanjulian, line 18: integration tests require documentation because they are complicated. please remove.

krzentner: Yeah please fix whatever's wrong with the file. It should be fine already.

src/garage/envs/multi_env_wrapper.py
```diff
@@ -82,6 +86,12 @@ def __init__(self,
             if not isinstance(env, GarageEnv):
                 envs[i] = GarageEnv(env)
         super().__init__(envs[0])
+        if env_names is not None:
+            assert isinstance(env_names, list), 'env_names must be a list'
+            msg = ('env_names are not unique or there is not an env_name',
+                   'corresponding to each env in envs')
+            assert len(set(env_names)) == len(envs), msg
+        self._env_names = env_names
```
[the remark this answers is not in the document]
krzentner: I think it makes sense as is, personally.

src/garage/envs/multi_env_wrapper.py
```diff
@@ -124,6 +145,11 @@ def active_task_index(self):
             int: Index of active task.
 
         """
+        if hasattr(self.env, 'active_task_index'):
+            return self.env.active_task_index
+        else:
+            return self._active_task_index
+
         return self._active_task_index
```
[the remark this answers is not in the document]
avnishn: The field self._active_task_index will be potentially different when using meta world environments. The private field acts as a way to index into env_names and envs

src/garage/envs/multi_env_wrapper.py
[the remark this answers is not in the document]
avnishn: making env names a set means that each env_names won't be duplicated, because set entries are unique. Doing this allows us to verify that the number of unique names in env_names is the exact same as the number of envs that the wrapper is being instantiated with.

setup.cfg
avnishn: fixed
\

ryanjulian:
It looks like this PR is not merging because you have a significant uncovered branch in the algorithm (something about the combination of alpha and automatic entropy tuning).

src/garage/envs/multi_env_wrapper.py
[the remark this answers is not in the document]
naeioi: Ah. I missed that you turn env_names into a set.

src/garage/envs/multi_env_wrapper.py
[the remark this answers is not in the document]
naeioi: By saying _**using metaworld environments**_, are you referring to use `MT10.get_train_tasks()` and `MT10.from_task(list_of_tasks)`? In such cases, MultiEnvWrapper saves only one instance in its `self._task_envs`, so that `self._active_task_index` will always be 0.

avnishn:
@mergifyio rebase

ryanjulian:
@Mergifyio refresh

PR #1342, opened 2020-04-23
Refactoring of DDPG and TD3 auto benchmarking scripts

zequnyu:
Following #1307, #1310 and #1320.


ryanjulian:
@Mergifyio rebase

PR #1343, opened 2020-04-23
Model and Policy interfaces

ryanjulian:


src/garage/_primitives.py
```diff
@@ -0,0 +1,152 @@
+"""Policy interface."""
```
ahtsan, line 1: Primitive

src/garage/_primitives.py
```diff
@@ -0,0 +1,152 @@
...
+            satisfy.
+
+    """
+
+    def __init__(self, name, spec):
+        self._name = name
+        self._spec = spec
+
+    @abc.abstractmethod
+    def get_parameters(self):
```
krzentner, line 20: Moving forward, I think it would actually be better to replace these methods with a protocol, so that outside authors only need to implement e.g. two methods, not four.

src/garage/_primitives.py
ryanjulian: yes, something like that

src/garage/_primitives.py
ryanjulian: hmm how would a protocol save them implementing 2 methods?

krzentner: If the author is writing e.g. a `ValueFunction` using pytorch, they need to define `forward` and `compute_loss` (two methods). If we make a protocol that attempts to use `get_parameters` but falls back to `get_state_dict`, then the author doesn't need to know about `get_parameters`.

ryanjulian: oh hmm -- we would implement that in this class?

krzentner: Yeah, we would have a function (e.g. `garage.get_parameters`) that implements this protocol by attempting to use `get_parameters`, then falling back to `get_state_dict`.

ryanjulian:
ah, so you think this function should still be part of the interface, but maybe just not required all the time?

re: our offline discussion of protocols, i will document a plan here mostly for posterity:

If you actually declare yourself as implementing this ABC, then standard rules apply and Python will force you to implement this method. If you don't, we will still try our best with the protocol. I favor this for a few reasons:

1. If you are anxious about protocols and like hyper-explicit code, you can declare the ABC and Python will check for you. It's also good for documentation of the formal interface, and provides a guarantee of what you should do in case your use case isn't covered by the magic in the protocol (e.g. what if you want to add a policy in `numpy`, which has no built-in parameter container?). Providing the ABC might also help for typed codebases, but I'm not sure.
2. Declaring the function with empty implementation seems dangerous (masks missing protocol implementation because of duck typing) and `NotImplementedErrors` only show up at the worse possible time in runtime, when it's already too late. `pylint` will catch calling a method which doesn't exist or is abstract in the type hierarchy, but I don't think it will complain if the implementation is `NotImplementedError`.
3. Declaring the interface documents what do to if you want to override the protocol, e.g. your can support `get_state_dict` with the protocol and most people in torch will ignore this interface, but if they want to do something special when we retrieve parameters, they can implement this explicitly.

The downside is that we are supporting more than one way to do things (inheritance vs protocols), and most people learn to modify code by example. So it raises the question

> "Should be be explicit in our own implementation, or use protocols?"

 I thought about this more last night, and I think a consistent truism is that you should make idiomatic use of the abstractions in the library code itself, e.g. kind of like being a self-hosting compiler. The library itself should be the best catalog of usage examples, and it's also a living proof that what you're asking users to do actually works out in the real world. This means that the TF and torch trees will probably make use of the protocol, and the numpy tree will probably be implementing interfaces.

The problem with this rule, and what makes us different from most users, we have a large library of many of these things, and want to ensure they are implementing these interfaces. We have been thinking that we'd use ABCs to do that, but that no longer seems tractible.

ryanjulian:
Research on interface contracts:
zope.interface: https://zopeinterface.readthedocs.io/en/latest/README.html
smaller package: https://interface.readthedocs.io/en/latest/ (uses inheritance not class decorators, is icky)
roll your own: https://hub.packtpub.com/interfaces/ (ensure_interface example is executed at runtime, but could likely be adapted to be a define-time evaluation)

src/garage/_primitives.py
```diff
@@ -0,0 +1,152 @@
...
+        """Get parameters of this primitive.
+
+        Returns:
+            dict[str, np.ndarray]: A dictionary mapping from names to values of
```
ryanjulian, line 24: note: probably don't make this ndarray

src/garage/_primitives.py
```diff
@@ -0,0 +1,152 @@
...
+        """Set the parameters of this primtive.
+
+        Args:
+            parameters (dict[str, np.ndarray]): A dictionary mapping from names
+                to values of each parameter.
+
+        """
+
+    @property
+    def spec(self):
```
krzentner, line 40: I think `InOutSpec` is a good argument to pass into this API, but I'm not convinced it should actually be accessible.

src/garage/_primitives.py
```diff
@@ -0,0 +1,152 @@
...
+        """InOutSpec: The input/output spec for this primitive."""
+        return self._spec
+
+    @property
+    def name(self):
+        """str: Name of this primitive."""
+        return self._name
+
+    @abc.abstractmethod
+    def clone(self, name):
```
krzentner, line 50: Is there a reason (besides Tensorflow 1.0 being weird) that we should include this in the API? Why can't we just use the `copy.deepcopy` API?

src/garage/_primitives.py
```diff
@@ -0,0 +1,152 @@
...
+        Args:
+            name (str): Name for the cloned instance.
+
+        Returns:
+            Primitive: A fully-independent copy of this primitive.
+
+        """
+
+    @abc.abstractmethod
+    def __getstate__(self):
```
krzentner, line 62: Does this force this method to be redefined? I agree this method should *work* on this type, but I'm not 100% on how `abc.abstractmethod` works. If another class I inherit from implements this method do I need to re-implement it?

ryanjulian: yeah, also not sure

src/garage/_primitives.py
```diff
@@ -0,0 +1,152 @@
...
+        env_spec (garage.envs.EnvSpec): Environment spec this Policy
+            will satisfy.
+
+    """
+
+    def __init__(self, name, env_spec):
+        super().__init__(name, env_spec.transpose())
+
+    @abc.abstractmethod
+    def get_action(self, observation, policy_state=None):
```
krzentner, line 94: Why can `policy_state` be `None`, but I can also explicitly request a reset state? What's the semantic difference between passing `None` and passing the result of `get_reset_state`.

src/garage/_primitives.py
```diff
@@ -0,0 +1,152 @@
...
+        Returns:
+            np.ndarray: An action, conditioned on the observation, conforming
+                to :obj:`env_spec.action_space`. Shape is :math:`(A^*,)`.
+            object: State information. If not None, this should be passed to
+                the next call to `get_action`.
+
+        """
+
+    @abc.abstractmethod
+    def get_actions(self, observations, policy_states=None):
```
krzentner, line 112: What happens if the number of `observations` and `policy_states` is different? `ValueError`?

src/garage/_primitives.py
naeioi: Under two cases we might want to get the parameters of a primitives - a) get trainable params for optimizer, b) get all params for snapshot. Case b) should be covered by __setstate__ and __getstate__ so that leaves only case a). Then we need to make sure the usage of `get_parameters` through its name or documents.

src/garage/_primitives.py
naeioi: @krzentner Without this property, how can it be connected with other primitives? Given that the spec should be compatible between two primitives when they are connected.

src/garage/_primitives.py
naeioi: `abc.abstractmethod` enforces subclass to implements it.

src/garage/_primitives.py
```diff
@@ -0,0 +1,152 @@
...
+    def __setstate__(self, state):
+        """See Object.__setstate__.
+
+        Args:
+            state (dict): Unpickled state of this primitive.
+
+        """
+
+
+class Policy(Primitive):
```
naeioi, line 80: On the naming of `Primitive`. I have a feeling that a policy is a large thing that does not fit well with the nuance of the smallness of Primitive. What do you think of the name `Module`?

src/garage/_primitives.py
```diff
@@ -0,0 +1,152 @@
...
+
+    Args:
+        name (str): Name for logs, variable scopes, etc.
+        env_spec (garage.envs.EnvSpec): Environment spec this Policy
+            will satisfy.
+
+    """
+
+    def __init__(self, name, env_spec):
+        super().__init__(name, env_spec.transpose())
```
naeioi, line 91: what is a spec transpose?

src/garage/_primitives.py
naeioi: Does this `policy_state` apply to only recurrent policies?

src/garage/_primitives.py
ryanjulian: it is defined above

spec(in_space, out_space).transpose() == spec(out_space, in_space)

src/garage/_primitives.py
ryanjulian: i agree this should be removed and instead enforced with some other means, such as using a decorator which checks interfaces

src/garage/_primitives.py
ryanjulian: `Module` is already taken by a very specific abstraction in tf and torch, but I am thinking we should rename this Model.

src/garage/_primitives.py
ryanjulian: yes. perhaps we could allow it to be length 1 or length len(observation) -- i.e. broadcastable. but that seems complicated.

src/garage/_primitives.py
ahtsan: is this `Model` related to the `Model` defined in TF branch?

ryanjulian: this would replace `garage.tf.Model` which would be partially- or fully-replaced by `tf.Module`

A tf policy would inherit from `tf.Module` and implement `garage.Policy`, but (as KR has discussed), it may not actually inherit from `garage.Policy`.

src/garage/_primitives.py
ryanjulian: @krzentner This allows for calling the `get_action()` API without a `policy_state`, which is idiomatic for non-stateful policies (i.e. feedforward only). Whereas infrastructure code should always implement the full API (below), it would be annoying for short scripts and notebooks using non-stateful policies.

@naeioi It applies to all policies in the sense that it must be part of their function signature, but they are under no obligation to use it.

rollout example for general code:
```py
for _ in range(num_episodes):
    ps = p.get_reset_state()
    o = env.reset()

    done = False
    t = 0
    while not done and t < max_path_length:
        a, ps = policy.get_action(o, policy_state=ps)
        o, r, done, env_info = env.step(a)
```

rollout example for known-stateless code:
```python
for _ in range(num_episodes):
    o = env.reset()

    done = False
    t = 0
    while not done and t < max_path_length:
        a, _ = policy.get_action(o)
        o, r, done, env_info = env.step(a)

Comment (ryanjulian) [reply src/garage/_primitives.py:94]: Essentially, the goal of this API is to make policy modules stateless, and to force the users of those stateful modules to maintain the state themselves.

This avoids forcing stateful policy authors to maintain state between calls to `get_action`, which can make things like serialization and cloning fraught. For instance, the current TF RNNs maintain the RNN state internally, and assume that sequential calls to `get_actions` (between calls to `reset`) will always have the same batch size. The policy now has an implicit internal state machine, which the caller has to mentally-model themselves.

I agree this problem is thorny, and I'm open to hearing different perspectives on how we could solve it.

The current practice seems reasonable on the face of things, but it can get messy:
* If you forget to call `reset()`, because you designed your code for stateless policies, your policy isn't actually stateful in the way you want it to be. This API makes it clear that if you want state, you will have to keep it yourself. 
* Under this API, passing `policy_state=None` to an RNN policy will lead to a `ValueError`, whereas right now you can just happily call `get_actions` repeatedly and always be accumulating state improperly. Right now, we could raise a `RuntimeError` if you forget to call `reset()` the first time, but after that we have no idea if you should have `reset()` or not.
* RNN policies have slightly-different APIs during optimization, so RNN support has to be added as a special case in each algorithm. Unifying the API allows us to add stateful policy support right out of the gate.
* You can't clone/pickle+unpickle a stateful policy with one batch size, maintaining its internal state, and sample it with another batch size (e.g. 1) without resetting the state. For instance, this could make evaluation sampling difficult, because the optimization target policy usually has a large batch size, and the evaluation policy usually a small one.

Comment (krzentner) [reply src/garage/_primitives.py:94]: Okay, I'm convinced this API is about as good as we can do.

**Comment by krzentner** [comment]
Thinking about the API here, what our users appear to be doing in practice, and how it fits into where the rest of the field is going as well as how we think we should structure our own logging code, I still think it's best for Models not to own their own names. Instead, I think it's best for the Model's owner (typically the algorithm) to have a name for the model.

Algorithms assigning names may seem more difficult, but the algorithm is actually the source of truth for what a model is used for. If you have four q networks (common in modern off-policy methods), then it's easy to get a name wrong and be confused by incorrect logging. If the algorithm always logs e.g. qf_1 as `qf_1` then this problem disappears. We already do this halfway, since algorithms are responsible for naming the clones of models they use, resulting in some q functions with log names like `MLPQFunction` and some with `qf1_target`. If we want to fix that inconsistency, we either need a more sophisticated way of automatically deriving names, or algorithms should just specify the name themselves. Either way, calling clone and passing a name feels like an anti-pattern, since there are many different places creating names, and they can't necessarily coordinate.

It's also the case that in Tensorflow code, we've encountered users expecting Tensorflow `Model`s to "just work" (in that unpickling / deepcopying them multiple times shouldn't throw exceptions). If we allocate names automatically when there would be a conflict, then we can make this "just work," but we've been hesitant to do so since it would make the resulting scalar and graph logs difficult to interpret.

If our main concern here is that Tensorflow graph logs are going to look inconsistent with scalar logs (e.g. `qf1`'s scalar values would correspond to `MLPQFunction`'s graph), then I think we're better off with some API for lazily setting the name on first use (basically, add a property (or method) to set `self._name` once instead of passing it to `__init__` in `garage.tf.models.Model`). I don't think that would actually involve a large code change, and it would fix existing inconsistency in the logging.

File: src/garage/_primitives.py (hunk 4 above)
Comment (krzentner) [reply src/garage/_primitives.py:40]: It's true that `InOutSpec` is useful if I have two things with arbitrary shapes (e.g. a pre-trained model and an environment that model wasn't trained on), and I want to create a model to automatically connect them together. However, we've very rarely wanted to do that in practice.

I'm not saying this property is a bad idea, but it's much easier to add new properties later than to remove existing ones, and it's not clear this property is necessary in most cases. In the rare cases that we do need it, I think it would be fine to put this in a separate interface.

If the `spec` property is supposed to also have the same semantics as an environment's `spec` property, then my opinion that this should be in a separate interface is even stronger.

File: src/garage/_primitives.py (hunk 2 above)
Comment (krzentner) [reply src/garage/_primitives.py:20]: I disagree that this method should return only trainable parameters. Which parameters are trainable is already better handled by the specific neural network library code. I think there's actually a third case c) for updating a policy with an existing name (e.g. what's usually done in the samplers). In order to handle that use case, this method should return all parameters (not just the trainable ones). Having said that, I still think we should have a protocol instead, so we can e.g. use `torch.nn.Module.get_state_dict` if that's defined and `get_parameters` isn't.

Comment (ryanjulian) [reply src/garage/_primitives.py:20]: I agree. If we run with the presumption that the Model<-->Optimizer API will always be framework-specific (e.g. no comingling of TF models with numpy optimizers, or at least doing so will require extra work/adapters), then there's no need to deal with trainability in this API. We only need be concerned with ALL parameters.

Alternatively, if we go to make a universal optimizer API, we can revisit this one.

I agree that all of these should basically be protocols and that inheriting from ABCs will be optional and even discouraged. The classes here function to define the interface unambiguously.

File: src/garage/_primitives.py (hunk 4 above)
Comment (ryanjulian) [reply src/garage/_primitives.py:40]: So I think this really gets at the heart of what we are going to ask of models in garage.

BLUF: Do we want to make safe runtime model composition first-class or not? If we do, then I think we need `.spec` interface, but maybe I'm missing something.

By "runtime" I mean that someone can define higher-order models like `ContextConditionedPolicy(ContextEncoderModel, PolicyModel)` which don't require users to pass classes+args to their constructors, but instead accept instances and wire themselves using the `.spec` provided by the instance.

This proposal basically says a model:
1. Has a name (discussion in another thread)
2. Exposes zero-or-more parameters, which I can get/set
3. Can be "wired together" with other models safely at runtime, using specs as a runtime type checking system
  a. By validating an InOutSpec in its constructor
  b. By providing its own InOutSpec, allowing the "wirer" to generate a derived InOutSpec

IME, if we are going to have 3a, not having 3b can make things really difficult.

I also don't agree that we don't really have this problem yet. It already appears
in PEARL: https://github.com/rlworkgroup/garage/blob/master/examples/torch/pearl_ml1_push.py
in Task Embeddings: 
https://github.com/rlworkgroup/garage/pull/1171/files#diff-47f13f3e683bface4baa8af5f8d921c4R127
..and the current state makes things really awkward to implement, and encourages creating one-off monolithic models for simple compositions (e.g. `GaussianMLPMultitaskPolicy`).

Comment (krzentner) [reply src/garage/_primitives.py:40]: I think you have the APIs needed for 3a and 3b backwards from what I expect. Personally, I think we should prioritize 3b (generating derived `InOutSpec` for sub-models) over 3a (checking that all the `InOutSpec` we've been given look right).

Consider the example you give above: `ContextConditionedPolicy(ContextEncoderModel, PolicyModel)`. I think to make this example "just work", `ContextConditionedPolicy` should be generating the `InOutSpec` for `ContextEncoderModel` and `PolicyModel` from its own `InOutSpec` (which is just the transposed environment spec), and passing those generated specs down to those sub-models. Basically, specs should be processed in an "outside-in" order. That's the obvious way to ensure everything fits together (i.e. 3b). No part of that process (except for the environment itself) involves reading `.spec`, only passing it down as a constructor arg.

Alternatively, you can check that `.spec`s fit together, but not design the API to ensure they do from the start. That, to me, sounds unwise. Although it might allow for slightly better error messages, the checking is not meaningfully earlier than just running the resulting model on real data. Furthermore, if `.spec` doesn't match the actual data handled / produced by the model, then you may end up with false negatives or false positives in you checking.

It might also make sense to read a `.spec` field if you're given two things that *don't* fit together, and you're going to generate another model to glue them together. Personally, I think it would be weird for `ContextConditionedPolicy` to construct additional trainable matrices to fit its sub-models together, so I don't think it even needs to read `.spec`. It might make sense for e.g. PEARL itself to do that though.

I agree that some cases may benefit from having the `.spec` interface, and I definitely think we're going to have a `.spec` interface one way or another. The question I'm trying to get at really is if third-party author's models need to implement it, when we can't even come up with a good example of when we would need to read such a field.

If we bundle it in with the `Model` interface, then the answer appears to be that they must (and environments simply happen to implement an identical but unrelated interface). If it's separate, then when it's needed the function can say so explicitly.

**Comment by ryanjulian** [comment]
@krzentner I agree. Perhaps could you provide a small snippet of how "new style" logging might work?

I think there is a tension going on about where calls to `logger.log` appear. Figuring out where the calls will live will help us a lot with names.

Generally, associated with a model will be some statistics which we want to log. It would be nice if the model author could provide a definition of how to calculate those, and could even inherit them.

rllab experimented with a `log_diagnostics` protocol for this problem, which was unevenly applied. Today, we have a mix of styles in the codebase, but have mostly succeeded in making the *outward* appearance pretty consistent. If Models were simple, I would just say they should implement some interface like `get_statistics` to return a dict (or even overload `__repr__`) and then handle it nicely with dowel.

```python
# Algorithm
for m in self._models:
    logger.log(m)

# Model named FooPolicy
def get_statistics(self):
    # self._name is 
    return {
        'entropy': self._entropy,
        'min_std': self._min_std,
    }

# Dowel
def model_handler(m):
    with prefix(m.name):
        for k, v in m.get_statistics():
            logkv(k,v)

# Produces logs to
# FooPolicy/entropy
# FooPolicy/min_std
```

The problems with the above framing are that (1) all of the statistics are conditioned on feedforward data, but models are stateless, and (2) sometimes you want to associate with a model something it can't even calculate, like a loss.

`log_diagnostics` from rllab tried to solve (1) by making the interface have an argument for a data batch, with the signature `log_diagnostics(self, path_data)`. It doesn't address (2) at all.

This doesn't really make much sense in the example I showed above. It introduces the possibility of re-running expensive feedforwards just to collect statistics, thought that seems mostly a non-issue at least in TensorFlow, due to intelligent graph caching. 

```python
# Algorithm
for m in self._models:
    stats = m.get_statistics(batch)  # expensive?
    for k, v in stats.items():
        logger.logkv(k, v)

# Model named FooPolicy
def get_statistics(self, batch):
    # self._name is 'FooPolicy'
    stats = self._stats_forward(batch)
    fstr = self._name + '/{}'
    return {fstr.format(k): v for k, v in stats}

# Produces logs to
# FooPolicy/entropy
# FooPolicy/min_std
```

KR's proposal would look like this. Issue (1) is the same, issue (2) is a non-issue (just log your associated stat at the same time), but overall it asks for a lot of manual munging for a very common task. It is very nice that it's flat.
```python
# Algorithm
for name, m in self._models.items():
    with logger.prefix(name):
        stats = m.get_statistics(batch)  # expensive?
        for k, v in stats.items():
            logger.logkv(k, v)

    logger.logkv('FooPolicy/loss', self._last_policy_loss)

# Model named FooPolicy
def get_statistics(self, batch):
    # self._name doesn't exist
    return self._stats_forward(batch)

# Produces logs to
# FooPolicy/entropy
# FooPolicy/min_std
```

### No-argument get_statistics

#### Force the model to cache them
One possibility for no-argument `get_statistics` is that we simply force the model to cache what it needs to report statistics during feedforward. This immediately creates the problem of which feedward(s) to report stats for. For instance, in the case of minibatching, you'd want to accumulate across possibly hundreds of feedforwards. The upside here is that you can calculate these during the main `forward` pass, so avoid a possible performance penalty *and* having to craft/feed a batch for `get_statistics`. Perhaps this would be solved by a `reset_statistics`. This adds complexity to both the Model implementations AND API. 

#### Return them from forward()
Another possibility is for the model to return statistics in response to `forward`, though again the algorithm would have to accumulate these itself in the case of minibatching.

#### Injection: A Metrics interface on dowel
This one dispatches entirely with `get_statistics` in favor of a `Metric` API. Models register their metrics on construction. Each metric has an accumulation strategy, and the algorithm decides when to reset accumulation. Dowel probably would just reset most metrics at the end of a logging epoch.

```python
# Algorithm
def __init__(self, ...):
    self._policy_loss_metric = dowel.register_metric(self._policy.name + '/loss')

def train(self, ...)
    for m in minibatches:
        loss = forward(stuff)
        self._policy_loss_metric.report(loss)

# Model named FooPolicy
def __init__(self, ...):
    self._entropy_metric = dowel.register_metric(self._name + '/entropy')
    self._min_std_metric = dowel.register_metric(self._name + '/min_std')

def forward(stuff):
    # stuff happens
    self._entropy_metric.report(ent)
    self._min_std_metric.report(min_std)

# Produces logs to
# FooPolicy/entropy
# FooPolicy/min_std
# FooPolicy/loss
```

krzentner:
To be clear, I think our priorities with respect to logging (in order of importance) should be:
1. Getting all of the log data reliably in the right place.
2. Being easy to extend without internal garage knowledge.
3. Not doing too much extra compute.
4. Being able to easily debug / write logging code.


For me, 1. means a few things:
1a. Statistics about a component are logged with a self-consistent prefix (i.e. statistics for the `qf1` shouldn't be scattered across multiple prefixes).
1b. Prefixes of different components should be consistent with respect to each other (i.e. `qf1` and `qf1_target`, not `MLPQFunction` and `qf1_target`).

In my opinion, extending dowel makes 1., 2., 3., and 4. difficult, so I don't think that's the route to go.

Priority 3. is quite complicated, as Ryan points out above, since these statistics are very expensive to compute from scratch, but should be possible to cache from e.g. `forward`. Personally, I think either way we need to accumulate statistics across multiple `forward` calls. I don't have a very strong opinion on whether that should be done by the policy, algorithm, or optimizer, but I suspect the optimizer should somehow provide a hook for doing it. Maybe we should have a simple `StatisticsAccumulator` class that knows what to do here?

If we rule out using dowel, then where the name exists is mostly relevant for 1., 2., and 4.
From my perspective, I'm mostly thinking of these options:

Model owns name:
```python
class QF:
    ...
    def get_statistics(self):
        return {'{}/{}'format(self.name, key): value for (key, value) in [
                ('parameter_var', self._param_var),
                ('parameter_mean', self._parameter_mean)]}

qf1_stats = self._qf1.get_statistics()
logger.log(qf1_stats)
with logger.prefix(self._qf1.name):
    logger.logkv('loss', self._qf1_optimizer.run())
qf1_target_stats = self._qf1_target.get_statistics()
logger.log(qf1_target_stats)  # Breaks 1b.
```
Which normally outputs:
```
MLPQFunction/parameter_var: 0.52...
MLPQFunction/parameter_mean: 1e-4...
MLPQFunction/loss: 0.4...
qf1_target/parameter_var: 0.52...
qf1_target/parameter_mean: 1e-4...
qf1_target/loss: 0.4...
```
If we auto generate names (instead of requiring they be passed into `model.clone`) it might output:
```
MLPQFunction/parameter_var: 0.52...
MLPQFunction/parameter_mean: 1e-4...
MLPQFunction/loss: 0.4...
MLPQFunction_copy_1/parameter_var: 0.52...
MLPQFunction_copy_1/parameter_mean: 1e-4...
MLPQFunction_copy_1/loss: 0.4...
```

Algorithm owns name:
```python
class QF:
    ...
    def get_statistics(self):
        return {'parameter_var': self._param_var),
                'parameter_mean': self._param_mean}

with logger.prefix('qf1'):
    qf1_stats = self._qf1.get_statistics()
    logger.log(qf1_stats)
    logger.logkv('loss', self._qf1_optimizer.run())
with logger.prefix('qf1_target'):
    qf1_target_stats = self._qf1_target.get_statistics()
    logger.log(qf1_target_stats)
```
Which outputs:
```
qf1/parameter_var: 0.52...
qf1/parameter_mean: 1e-4...
qf1/loss: 0.4...
qf1_target/parameter_var: 0.52...
qf1_target/parameter_mean: 1e-4...
qf1_target/loss: 0.4...
```

Technically, the algorithm owning the name results in one more line of code in the algorithm (to set up the prefix for `qf1_target`, which wouldn't otherwise need it). However, it results in simpler code in the model, which we actually tend to have more implementations of. It could be argued that code in the algorithm is more expensive, but I think the resulting code is actually simpler anyways. Either way, the change is not very large, at most one line of extra formatting logic.

As currently proposed, models owning their names makes it difficult not to violate 1b, because the algorithm has made `qf1_target`, by cloning `qf1`. We don't have a canonical way of auto-generating names, so it uses a hardcoded `'qf1_target'`, which doesn't match with the name `qf1` has in that example. Still, there are various ways for us to solve that while still letting models own their own names (such as making `clone` (and maybe `__setstate__`) auto generate new names).

However, I think that the algorithm owning the name is a clear win for priorities 2. and 3.. If I didn't know anything about garage, and I was asked to write a q function, I wouldn't generally assume I need to have a `name` field, so it would be easier for me if I didn't have to learn that. 

Priority 3. is also easier, in my opinion, since I can find where logging is generated by searching for `'qf1'` in the code (as opposed to having to search for `parameter_var` and then walk up the call stack.

The main remaining issue, namely that 1. would be violated because the graph and scalar logs use different names, I believe can be solved with a different protocol.

src/garage/_primitives.py
ahtsan: `name` is needed because two different variables cannot exist in the same variable scope (i.e. same name). Is this not happening anymore in TF2?

src/garage/_primitives.py
```diff
@@ -0,0 +1,152 @@
...
+    @abc.abstractmethod
+    def get_reset_states(self, batch_size):
+        """Get the reset value of policy_states for a given batch size.
+
+        Args:
+            batch_size (int): Observation/action batch size for the calls to
+               `get_actions` this state will be used with.
+
+        Returns:
+            object: Policy state representing a newly-reset policy.
```
ahtsan, line 150: Should it return a list of dict? A list of `batch_size` dict, where each dict contains key-value pairs for `states`, e.g. `{'prev_actions': ..., 'prev_hidden_state': ..., 'prev_cell_state': ...}`

PR #1344, opened 2020-04-24
Move most dm_control tests to nightly

ryanjulian:


ryanjulian:
This doesn't achieve the goal of making the MuJoCo job any faster -- that appears to be driven by some very long-running tests.

PR #1353, opened 2020-04-28
Remove joblib and old-style samplers

krzentner:
Note that OffPolicyVectorizedSampler still exists, for now.


scripts/setup_linux.sh
```diff
@@ -144,18 +144,18 @@ sudo apt install -y \
...
-git checkout 0be4f3f75aebd9d24583ee86590a38e741db0904
-mkdir glfw-build
-cd glfw-build
-sudo cmake -DBUILD_SHARED_LIBS=ON -DGLFW_BUILD_EXAMPLES=OFF \
-  -DGLFW_BUILD_TESTS=OFF -DGLFW_BUILD_DOCS=OFF ..
-sudo make -j"$(nproc)"
-sudo make install
-cd "${GARAGE_DIR}"
+# sudo apt purge -y libglfw*
+# GLFW_DIR="$(mktemp -d)/glfw"
```
ryanjulian, line 148: this is needed on 18.04 and 16.04, but not 20.04 -- perhaps fork this into a new script?

krzentner: Oh whoops. I didn't think I'd staged this file.

src/garage/experiment/experiment.py
```diff
@@ -701,22 +701,28 @@ def make_launcher_archive(*, git_root_path, log_dir):
...
-    files_to_archive = subprocess.check_output(
+    git_files = subprocess.check_output(
         ('git', 'ls-files', '--others', '--exclude-standard', '--cached',
          '-z'),
         cwd=git_root_path).strip()
-    repo_size = sum([
-        os.stat(os.path.join(git_root_path, f)).st_size
-        for f in files_to_archive.split(b'\0')
-    ])
+    repo_size = 0
```
krzentner, line 708: This stuff should actually be in a separate PR. It fixes `wrap_experiment` when there are deleted files in the current launch repo.

examples/np/cem_cartpole.py
```diff
@@ -47,7 +47,7 @@ def cem_cartpole(ctxt=None, seed=1):
                    max_path_length=100,
                    n_samples=n_samples)
 
-        runner.setup(algo, env, sampler_cls=OnPolicyVectorizedSampler)
+        runner.setup(algo, env, sampler_cls=MultiprocessingSampler)
```
ryanjulian, line 50: what would the default be if you removed this? is it not okay?

krzentner: It is okay to remove this too. I think I'll replace it with `VecWorker` instead.

src/garage/experiment/experiment.py
avnishn: I have had this issue before. Very necessary

ryanjulian:
@Mergifyio rebase

ryanjulian:
This can now be merged

ryanjulian:
I <3 this PR

PR #1355, opened 2020-04-28
Fix setting model parameters

ahtsan:
Currently, when setting model parameters, the parameter name has to be an exact
match. This would not work when assigning a model's parameters with a model with
a different name. This PR fixes the bug by only matching the part of the parameter
name without the model name.

### DIFF
--- a/src/garage/tf/models/base.py
+++ b/src/garage/tf/models/base.py
@@ -339,9 +339,15 @@ class Model(BaseModel):
         """
         variables = self._get_variables()
         for name, var in variables.items():
-            if name in parameters:
-                var.load(parameters[name])
-            else:
+            found = False
+            # param name without model name
+            param_name = name[name.find(self.name) + len(self.name) + 1:]
+            for k, v in parameters.items():
+                if param_name in k:
+                    var.load(v)
+                    found = True
+                    continue
+            if not found:
                 warnings.warn('No value provided for variable {}'.format(name))
 
     @property
--- a/tests/garage/tf/models/test_model.py
+++ b/tests/garage/tf/models/test_model.py
@@ -19,7 +19,9 @@ class SimpleModel(Model):
     def network_output_spec(self):
         return ['state', 'action']
 
+    # pylint: disable=arguments-differ
     def _build(self, obs_input, name=None):
+        del name
         state = mlp(obs_input, self._output_dim, self._hidden_sizes, 'state')
         action = mlp(obs_input, self._output_dim, self._hidden_sizes, 'action')
         return state, action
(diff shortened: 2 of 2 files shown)


tests/garage/tf/models/test_model.py
```diff
@@ -304,3 +312,14 @@ def test_simple_model_is_pickleable_with_missing_parameters(self):
                 'SimpleModel/state/hidden_0/kernel:0']
             with pytest.warns(UserWarning):
                 model_pickled.build(state)
+
+    def test_model_set_parameters(self):
+        model1 = SimpleModel(output_dim=2, name='model1')
```
ryanjulian, line 317: hmm how does this work if the model is not in the top-level scope? i.e. what if it's in `/foo/bar/model1`?

src/garage/tf/models/base.py
```diff
@@ -339,9 +339,15 @@ def parameters(self, parameters):
         """
         variables = self._get_variables()
         for name, var in variables.items():
-            if name in parameters:
-                var.load(parameters[name])
-            else:
+            found = False
+            # param name without model name
+            param_name = name[name.find(self.name) + len(self.name) + 1:]
```
ryanjulian, line 344: what if the model name is not globally-unique? will this cause an error or incorrect behavior? i think we need to think about this really carefully.

ryanjulian:
did this cause a breaking bug somewhere?

src/garage/tf/models/base.py
ahtsan: it only compares the variable inside `model1` and `model2`, and match their parameters with the part of the names that exclude the model name. It doesn't involve any other models.

ryanjulian: What if I have Foo/model1 and Bar/model1 in my session?

tests/garage/tf/models/test_model.py
ahtsan: if model1 has `/foo/bar/model1/param1` and model2 has `/foo/bar/model2/param1`, then it match the parameters only looking at `/param1`. I think it's intuitive because the parameter name will be the sub-path of the model.

src/garage/tf/models/base.py
ahtsan: If you do model1_foo.parameters = model1_bar.parameters, their parameters will be the same, which is the intended usage. In the new NPO, we can do old_policy.model.parameters = new_policy.model.parameters to update policy.

ahtsan:
＠Mergifyio backport 2019.10

ahtsan:
@Mergifyio backport 2019.10

ahtsan:

PR #1357, opened 2020-04-28
Fix categorical primitives distribution argument

ahtsan:
Without providing the function argument `probs`, the default argument is `logits`
which is not the vanilla probability values.

### DIFF
--- a/src/garage/tf/models/categorical_gru_model.py
+++ b/src/garage/tf/models/categorical_gru_model.py
@@ -60,7 +60,7 @@ class CategoricalGRUModel(GRUModel):
                  hidden_b_init=tf.zeros_initializer(),
                  recurrent_nonlinearity=tf.nn.sigmoid,
                  recurrent_w_init=tf.initializers.glorot_uniform(),
-                 output_nonlinearity=None,
+                 output_nonlinearity=tf.nn.softmax,
                  output_w_init=tf.initializers.glorot_uniform(),
                  output_b_init=tf.zeros_initializer(),
                  hidden_state_init=tf.zeros_initializer(),
@@ -116,6 +116,5 @@ class CategoricalGRUModel(GRUModel):
         """
         outputs, step_output, step_hidden, init_hidden = super()._build(
             state_input, step_input, step_hidden, name=name)
-        outputs = outputs / tf.math.reduce_sum(outputs)
         dist = tfp.distributions.OneHotCategorical(probs=outputs)
         return dist, step_output, step_hidden, init_hidden
--- a/src/garage/tf/models/categorical_gru_model.py
+++ b/src/garage/tf/models/categorical_gru_model.py
@@ -60,7 +60,7 @@ class CategoricalGRUModel(GRUModel):
                  hidden_b_init=tf.zeros_initializer(),
                  recurrent_nonlinearity=tf.nn.sigmoid,
                  recurrent_w_init=tf.initializers.glorot_uniform(),
-                 output_nonlinearity=tf.nn.softmax,
+                 output_nonlinearity=None,
                  output_w_init=tf.initializers.glorot_uniform(),
                  output_b_init=tf.zeros_initializer(),
                  hidden_state_init=tf.zeros_initializer(),
@@ -75,12 +75,13 @@ class CategoricalGRUModel(GRUModel):
             hidden_b_init=hidden_b_init,
             recurrent_nonlinearity=recurrent_nonlinearity,
             recurrent_w_init=recurrent_w_init,
-            output_nonlinearity=output_nonlinearity,
+            output_nonlinearity=tf.nn.softmax,
             output_w_init=output_w_init,
             output_b_init=output_b_init,
             hidden_state_init=hidden_state_init,
             hidden_state_init_trainable=hidden_state_init_trainable,
             layer_normalization=layer_normalization)
+        self._output_nonlinearity = output_nonlinearity
 
     def network_output_spec(self):
         """Network output spec.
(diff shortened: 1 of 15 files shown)


ahtsan:
I can't fully explain yet, but if we are passing the output from MLP to the Categorical distribution as `probs` instead of `logits`, it learns better.

ahtsan:
Are tf.distribution part of 2020.03 release?

ryanjulian:
They aren't blocking the release, no.

How much better does it learn? How do you know?

```
logits: An N-D Tensor, N >= 1, representing the unnormalized log probabilities of a set of Categorical distributions. The first N - 1 dimensions index into a batch of independent distributions and the last dimension represents a vector of logits for each class. Only one of logits or probs should be passed in.
probs: An N-D Tensor, N >= 1, representing the probabilities of a set of Categorical distributions. The first N - 1 dimensions index into a batch of independent distributions and the last dimension represents a vector of probabilities for each class. Only one of logits or probs should be passed in.
```

Is your MLP output log-prob or absolute? are they guaranteed normalized?

ahtsan:
I think it depends on our interpretation? MLP is just an estimated function that we train to estimate whatever values. Would it be possible the values are too large for log-prob?

ryanjulian:
My main concern is that the documentation above implies that probs must be normalized (i.e. sum to 1).

is it just as effective if you normalize the output of the MLP before sending it to prob?

ahtsan:
It is. Updated.

src/garage/tf/models/categorical_mlp_model.py
```diff
@@ -72,4 +72,5 @@ def _build(self, state_input, name=None):
 
         """
         prob = super()._build(state_input, name=name)
-        return tfp.distributions.OneHotCategorical(prob)
+        prob = prob / tf.math.reduce_sum(prob)
```
ryanjulian, line 75: i think you should probably use https://www.tensorflow.org/api_docs/python/tf/norm `ord=1` rather than a division operator.

your implementation is dangerous if `prob` is very small or zero. i think tf.norm should handle this? but maybe not.

ahtsan: How about we manually set the `output_nonlinearity` of the MLP to softmax?

ahtsan: I realized we were already using softmax so the output of the MLP should already be normalized and sum to 1?

ahtsan: I can add a test

ryanjulian: softmax should sum to 1

src/garage/tf/models/categorical_gru_model.py
```diff
@@ -60,7 +60,7 @@ def __init__(self,
                  hidden_b_init=tf.zeros_initializer(),
                  recurrent_nonlinearity=tf.nn.sigmoid,
                  recurrent_w_init=tf.initializers.glorot_uniform(),
-                 output_nonlinearity=None,
```
ryanjulian, line 63: if this is necessary for the correct operation of the model, it should not be an option.

Alternatively, you could rename this something like `output_normalization_fn`, but I recommend just removing it.

src/garage/tf/models/categorical_mlp_model.py
```diff
@@ -47,10 +47,10 @@ def __init__(self,
                  output_dim,
                  name=None,
                  hidden_sizes=(32, 32),
-                 hidden_nonlinearity=tf.nn.relu,
+                 hidden_nonlinearity=tf.nn.tanh,
                  hidden_w_init=tf.initializers.glorot_uniform(),
                  hidden_b_init=tf.zeros_initializer(),
-                 output_nonlinearity=None,
+                 output_nonlinearity=tf.nn.softmax,
```
ryanjulian, line 53: see my comment above.

src/garage/tf/models/categorical_gru_model.py
ahtsan: Right, we should have enforced it for categorical primitives. Will fix.

ryanjulian: there's also no reason you can't have an output non-linearity followed by your normalizer. that seems like another way out of this problem

ahtsan: Indeed, how about applying softmax inside the model, while making `output_nonlineaity=None` by default? Then we can still have any output non-linearity as we want.

ryanjulian: lgtm

ahtsan:
@krzentner @naeioi PTAL

src/garage/tf/policies/categorical_gru_policy2.py
```diff
@@ -67,7 +67,7 @@ def __init__(self,
                  hidden_b_init=tf.zeros_initializer(),
                  recurrent_nonlinearity=tf.nn.sigmoid,
                  recurrent_w_init=tf.initializers.glorot_uniform(),
-                 output_nonlinearity=tf.nn.softmax,
+                 output_nonlinearity=None,
```
krzentner, line 70: Is this because it's already done by a submodel? If so, does it ever makes sense to set this parameter? If not, should we delete it?

ahtsan: Submodel already applied a softmax layer. This output_nonlinearity allows users to do whatever they want.
Currently all primitives are sharing the same protocol, i.e. hidden_w, hidden_b, hidden_nonlinearity, output_w, output_b, output_nonlinearity.
We can also decide to remove some of them if we think some primitives do not need them, but I think that's not what we should do at this moment.

krzentner: Okay, sounds good.

ahtsan:
@ryanjulian I don't think it needs backport? It wasn't in 2019.10 release.

ryanjulian:
@ahtsan you're right -- removed it

PR #1325, opened 2020-04-20
Create a 'benchmarks' extra, separate from 'dev'

ryanjulian:

PR #1368, opened 2020-05-01
Separate benchmark experiments into files

zequnyu:
For #1325, because some dependencies exist in some [files](https://github.com/rlworkgroup/garage/blob/4f8ed2cff8c4e72e56d9cf55ca5f83db3df61386/tests/garage/tf/models/test_gaussian_gru_model.py#L10), it's better to refactor the plotting helper function for benchmarking first, and this is the preparation step.

This PR separates experiments into single files, and gathers all benchmarks in one [file](https://github.com/rlworkgroup/garage/blob/c2ccfefa9b5a16a131168e6b9b1021252b44d40b/tests/benchmarks/auto/benchmarks.py) (module) and I plan to automate figure plotting in `iterate_experiments`. Following this step shortly I'll bring up a PR to refactor the plotting helper function. By then, we should be able to replace all of our old benchmarking code as well as getting #1325 done. In the future, when new benchmarking scripts are being implemented, they should be able to follow similar pattern.


zequnyu:
@yonghyuc Would you check if this reflects the correct change of your new value function?

ryanjulian:
@Mergifyio rebase

Issue #1367, opened 2020-04-30
Add DeprecationWarnings to run_experiment and old-style samplers


ryanjulian:
We should include these in v2020.04 as a courtesy, since we plan on deleting them in all future releases.

PR #1369, opened 2020-05-01
Add DeprecationWarnings to old-style samplers

ryanjulian:


ryanjulian:
Fixes #1367

ryanjulian:
@Mergifyio rebase

ryanjulian:
We should probably change the default. It will get us more bug reports on the new code, rather than old, stale bug reports on code we're going to delete anyway.

Issue #1365, opened 2020-04-30
HalfCheetahDirEnv meta-evaluation


bordeauxred:
Hi there, 

I tried to replicate the pearl results from the pearl paper, however for halfcheetah forward/backward (script attached as txt) I got, using the hyper parameters similar to the oyster repo https://github.com/katerakelly/oyster/blob/master/configs/cheetah-dir.json and https://github.com/katerakelly/oyster/blob/master/configs/default.py ) a strange result. 
The returns are in absolute values as expected,  however the  Cheetah is either running always forward or always backward within one evaluation phase. (so the return for forward is very high (x) and for backward very low (-x) or vice versa) 
As the directions are always sampled the sampling I tried a) to hack the sampling to include always a forward and a backward task and to shuffle the order b) to sample 8 test and train tasks to make sure each direction is at least seen once. The above described behavior of running always into the same direction within one meta-evaluation was observed in both cases. 
Do you have by chance an intuition what might cause this and how to fix it?
[pearl_half_cheetah_dir.txt](https://github.com/rlworkgroup/garage/files/4558838/pearl_half_cheetah_dir.txt)

ryanjulian:
@bordeauxred thanks for reporting a possible bug! It's user reports like yours which really help us make the software better.

I think we need a little more information from you to help us track it down.

Can you give us a report with the following template?

# Expected Behavior

# Observed Behavior

# Steps to Reproduce
1.
2.
3.

# Possible solutions or things you've tried (and whether they worked or not)

# More info

bordeauxred:
Expected:
I expected to see for halfcheetahdir (using the oyster hyper-parameters  https://github.com/katerakelly/oyster/blob/master/configs/cheetah-dir.json and https://github.com/katerakelly/oyster/blob/master/configs/default.py ))a learning curve as in the paper pearl paper:
![Screenshot from 2020-04-30 20-54-53](https://user-images.githubusercontent.com/28216059/80748237-e547e880-8b24-11ea-9cad-52a31505e669.png)

i instead got 
![hc_dir](https://user-images.githubusercontent.com/28216059/80751838-ccdacc80-8b2a-11ea-8519-900bcd7771fa.png)

steps to reproduce:
I can't attach .py files, thus I changed the format to .txt. Please change the type back to .py and you could run it. 
[pearl_half_cheetah_dir.txt](https://github.com/rlworkgroup/garage/files/4560858/pearl_half_cheetah_dir.txt)


Things I tried:
[...]

PR #1374, opened 2020-05-02
Fix PEARL.adapt_policy

krzentner:
Before it was using samples from the replay buffer to create the
context. Now it uses the exploration samples (as it should).

This fixes the behavior of PEARL on HalfCheetahDirEnv (and probably most
other meta-learning environments).

### DIFF
--- a/src/garage/torch/algos/pearl.py
+++ b/src/garage/torch/algos/pearl.py
@@ -592,7 +592,12 @@ class PEARL(MetaRLAlgorithm):
                 exploration_trajectories.
 
         """
-        context = self._sample_context(self._task_idx)
+        total_steps = sum(exploration_trajectories.lengths)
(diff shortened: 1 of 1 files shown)


krzentner:
This appears to fix #1365. See [the tensorboard dev runs 13 and 15 here.](https://tensorboard.dev/experiment/RI9kVkEZRyuk2xxnU2AgQA/)

krzentner:
There isn't a test that would have caught this. Unfortunately, this is difficult to detect with less than about an hour of GPU compute, because of how unstable meta-RL algorithms are when starting up.

I suppose one invariant that should be true is "meta evaluation should work before any training." If we checked that we would have detected this bug, but it likely wouldn't detect similar bugs.

krzentner:
@mergifyio rebase

PR #1379, opened 2020-05-04
Refactor NPO to use primitives with tfp.distribution

ahtsan:
Refactor NPO to use primitives with tfp.distribution.

### DIFF
--- a/src/garage/tf/algos/__init__.py
+++ b/src/garage/tf/algos/__init__.py
@@ -5,7 +5,9 @@ from garage.tf.algos.ddpg import DDPG
 from garage.tf.algos.dqn import DQN
 from garage.tf.algos.erwr import ERWR
 from garage.tf.algos.npo import NPO
+from garage.tf.algos.npo2 import NPO2
 from garage.tf.algos.ppo import PPO
+from garage.tf.algos.ppo2 import PPO2
 from garage.tf.algos.reps import REPS
 from garage.tf.algos.rl2 import RL2
 from garage.tf.algos.rl2ppo import RL2PPO
@@ -13,6 +15,7 @@ from garage.tf.algos.rl2trpo import RL2TRPO
 from garage.tf.algos.td3 import TD3
 from garage.tf.algos.tnpg import TNPG
 from garage.tf.algos.trpo import TRPO
+from garage.tf.algos.trpo2 import TRPO2
 from garage.tf.algos.vpg import VPG
 
 __all__ = [
@@ -22,7 +25,9 @@ __all__ = [
     'DQN',
     'ERWR',
     'NPO',
+    'NPO2',
     'PPO',
+    'PPO2',
     'REPS',
     'RL2',
     'RL2PPO',
@@ -30,5 +35,6 @@ __all__ = [
     'TD3',
     'TNPG',
     'TRPO',
+    'TRPO2',
     'VPG',
 ]
--- a/src/garage/tf/algos/batch_polopt.py
+++ b/src/garage/tf/algos/batch_polopt.py
@@ -241,13 +241,8 @@ class BatchPolopt(RLAlgorithm):
 
         lengths = np.asarray([v.sum() for v in valids])
 
-        ent = np.sum(self.policy.distribution.entropy(agent_infos) *
-                     valids) / np.sum(valids)
-
         self.episode_reward_mean.extend(undiscounted_returns)
 
-        tabular.record('Entropy', ent)
-        tabular.record('Perplexity', np.exp(ent))
         tabular.record('Extras/EpisodeRewardMean',
                        np.mean(self.episode_reward_mean))
 
--- a/src/garage/tf/algos/npo2.py
+++ b/src/garage/tf/algos/npo2.py
@@ -1,0 +1,562 @@
+"""Natural Policy Gradient Optimization."""
+from dowel import logger, tabular
+import numpy as np
+import tensorflow as tf
+
+from garage.misc import tensor_utils as np_tensor_utils
+from garage.tf.algos.batch_polopt import BatchPolopt
+from garage.tf.misc.tensor_utils import center_advs
+from garage.tf.misc.tensor_utils import compile_function
+from garage.tf.misc.tensor_utils import compute_advantages
+from garage.tf.misc.tensor_utils import discounted_returns
+from garage.tf.misc.tensor_utils import flatten_inputs
+from garage.tf.misc.tensor_utils import graph_inputs
+from garage.tf.misc.tensor_utils import new_tensor
+from garage.tf.misc.tensor_utils import positive_advs
+from garage.tf.optimizers import LbfgsOptimizer
+
+
+class NPO2(BatchPolopt):
+    """Natural Policy Gradient Optimization.
+
+    Args:
+        env_spec (garage.envs.EnvSpec): Environment specification.
+        policy (garage.tf.policies.base.Policy): Policy.
+        baseline (garage.tf.baselines.Baseline): The baseline.
+        scope (str): Scope for identifying the algorithm.
+            Must be specified if running multiple algorithms
+            simultaneously, each using different environments
+            and policies.
+        max_path_length (int): Maximum length of a single rollout.
+        discount (float): Discount.
+        gae_lambda (float): Lambda used for generalized advantage
+            estimation.
+        center_adv (bool): Whether to rescale the advantages
+            so that they have mean 0 and standard deviation 1.
+        positive_adv (bool): Whether to shift the advantages
+            so that they are always positive. When used in
+            conjunction with center_adv the advantages will be
+            standardized before shifting.
+        fixed_horizon (bool): Whether to fix horizon.
+        pg_loss (str): A string from: 'vanilla', 'surrogate',
+            'surrogate_clip'. The type of loss functions to use.
+        lr_clip_range (float): The limit on the likelihood ratio between
+            policies, as in PPO.
+        max_kl_step (float): The maximum KL divergence between old and new
+            policies, as in TRPO.
+        optimizer (object): The optimizer of the algorithm. Should be the
+            optimizers in garage.tf.optimizers.
+        optimizer_args (dict): The arguments of the optimizer.
+        policy_ent_coeff (float): The coefficient of the policy entropy.
+            Setting it to zero would mean no entropy regularization.
+        use_softplus_entropy (bool): Whether to estimate the softmax
+            distribution of the entropy to prevent the entropy from being
+            negative.
+        use_neg_logli_entropy (bool): Whether to estimate the entropy as the
+            negative log likelihood of the action.
+        stop_entropy_gradient (bool): Whether to stop the entropy gradient.
+        entropy_method (str): A string from: 'max', 'regularized',
+            'no_entropy'. The type of entropy method to use. 'max' adds the
+            dense entropy to the reward for each time step. 'regularized' adds
+            the mean entropy to the surrogate objective. See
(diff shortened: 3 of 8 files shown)


ahtsan:
Benchmark results will be uploaded once it's done.

src/garage/tf/algos/batch_polopt.py
```diff
@@ -241,13 +241,8 @@ def process_samples(self, itr, paths):
 
         lengths = np.asarray([v.sum() for v in valids])
 
-        ent = np.sum(self.policy.distribution.entropy(agent_infos) *
-                     valids) / np.sum(valids)
-
         self.episode_reward_mean.extend(undiscounted_returns)
 
-        tabular.record('Entropy', ent)
```
ryanjulian, line 249: why?

src/garage/tf/algos/npo2.py
```diff
@@ -0,0 +1,562 @@
...
+
+                # Maximize E[surrogate objective] by minimizing
+                # -E_t[surrogate objective]
+                if self.policy.recurrent:
+                    # filter valid obj
+                    obj = tf.boolean_mask(obj, i.valid_var)
+                loss = -tf.reduce_mean(obj)
+
+            # Diagnostic functions
+            self._f_policy_kl = compile_function(
```
ryanjulian, line 367: can you use tf.Session.make_callable instead?

src/garage/tf/algos/__init__.py
```diff
@@ -5,14 +5,17 @@
 from garage.tf.algos.dqn import DQN
 from garage.tf.algos.erwr import ERWR
 from garage.tf.algos.npo import NPO
+from garage.tf.algos.npo2 import NPO2
 from garage.tf.algos.ppo import PPO
+from garage.tf.algos.ppo2 import PPO2
```
ryanjulian, line 10: i would prefer to just delete the old one, as long as tests still pass and benchmarks look good

src/garage/tf/algos/npo2.py
```diff
@@ -0,0 +1,562 @@
...
+                obs_var = tf.compat.v1.placeholder(
+                    tf.float32,
+                    shape=[None, None, observation_space.flat_dim],
+                    name='obs')
+            else:
+                obs_var = observation_space.to_tf_placeholder(name='obs',
+                                                              batch_dims=2)
+            action_var = action_space.to_tf_placeholder(name='action',
+                                                        batch_dims=2)
+            reward_var = new_tensor(name='reward', ndim=2, dtype=tf.float32)
```
ryanjulian, line 223: can you switch these out for tf.compat.v1.placeholder?

src/garage/tf/algos/__init__.py
ahtsan: Sure

src/garage/tf/algos/npo2.py
ahtsan: Sure

src/garage/tf/algos/npo2.py
ahtsan: Sure

src/garage/tf/algos/npo2.py
```diff
@@ -0,0 +1,562 @@
...
+              use_neg_logli_entropy=False
+
+    """
+
+    def __init__(self,
+                 env_spec,
+                 policy,
+                 baseline,
+                 scope=None,
+                 max_path_length=500,
```
ryanjulian, line 86: this is a dangerous default -- 100 is more typical

src/garage/tf/algos/npo2.py
```diff
@@ -0,0 +1,562 @@
...
+                                          stop_entropy_gradient,
+                                          use_neg_logli_entropy,
+                                          policy_ent_coeff)
+
+        if pg_loss not in ['vanilla', 'surrogate', 'surrogate_clip']:
+            raise ValueError('Invalid pg_loss')
+
+        with self._name_scope:
+            self._optimizer = optimizer(**optimizer_args)
+            self._lr_clip_range = float(lr_clip_range)
```
ryanjulian, line 126: i don't think the following 3 need to be in a name scope...

src/garage/tf/algos/npo2.py
```diff
@@ -0,0 +1,562 @@
...
+                for k, shape in self.policy.state_info_specs
+            }
+            policy_state_info_vars_list = [
+                policy_state_info_vars[k] for k in self.policy.state_info_keys
+            ]
+
+        if self.policy.recurrent:
+            _obs_var = obs_var
+            # pylint: disable= protected-access
+            if self.policy._state_include_action:
```
ryanjulian, line 245: i don't see the relationship here -- you test `_state_include_action` and then indiscriminately augment the input with all of the state_info_vars

src/garage/tf/algos/npo2.py
```diff
@@ -0,0 +1,562 @@
...
+                k: tf.compat.v1.placeholder(tf.float32,
+                                            shape=[None] * 2 + list(shape),
+                                            name=k)
+                for k, shape in self.policy.state_info_specs
+            }
+            policy_state_info_vars_list = [
+                policy_state_info_vars[k] for k in self.policy.state_info_keys
+            ]
+
+        if self.policy.recurrent:
```
ryanjulian, line 242: can you remove this from the new policies and just _always_ augment with extra_state_var? if a policy is stateless, it can just have `len(state_info_keys) == 0`

src/garage/tf/algos/npo2.py
```diff
@@ -0,0 +1,562 @@
...
+                policy_entropy = pol_dist.entropy()
+
+            # This prevents entropy from becoming negative for small policy std
+            if self._use_softplus_entropy:
+                policy_entropy = tf.nn.softplus(policy_entropy)
+
+            if self._stop_entropy_gradient:
+                policy_entropy = tf.stop_gradient(policy_entropy)
+
+        if not self.policy.recurrent:
```
ryanjulian, line 406: would this still work if the policy were non-recurrent?

src/garage/tf/algos/npo2.py
```diff
@@ -0,0 +1,562 @@
...
+            aug_returns.append(path['returns'])
+        samples_data['rewards'] = np_tensor_utils.pad_tensor_n(
+            aug_rewards, self.max_path_length)
+        samples_data['returns'] = np_tensor_utils.pad_tensor_n(
+            aug_returns, self.max_path_length)
+
+        # Fit baseline
+        logger.log('Fitting baseline...')
+        if hasattr(self.baseline, 'fit_with_samples'):
+            self.baseline.fit_with_samples(paths, samples_data)
```
ryanjulian, line 455: we don't have any baselines that implement `fit_with_samples`, so i think you can safely remove this check

src/garage/tf/algos/npo2.py
ahtsan: ops I copied from the original NPO and wasn't aware. Will fix

src/garage/tf/algos/npo2.py
ahtsan: Sure

src/garage/tf/algos/npo2.py
ahtsan: This flag is for applying `boolean_mask` or not. For recurrent, inputs feeding to policies will be [N, T, S^*] while non-recurrent would be (N, S^*)

src/garage/tf/algos/npo2.py
ahtsan: looks like the comment above is for this block of comment? Yes if we have empty `self.policy.state_info_keys` for non-recurrent policies, we can always augment with `self.policy.state_info_keys`.

src/garage/tf/algos/npo2.py
ahtsan: Ok

src/garage/tf/algos/npo2.py
ahtsan: We have tests and benchmarks for NPO with non-recurrent policies.

src/garage/tf/algos/__init__.py
ryanjulian: the same goes with all of the primitives -- having DistributionFooPolicy/DistributionFooPolicy2 will just create confusion.

If you don't have the time to run the benchmarks before merging this, then feel free to merge it and send a PR to delete/rename everything later.

src/garage/tf/algos/npo2.py
ryanjulian: this is what i'm proposing. i want to delete the property `Policy.recurrent`

src/garage/tf/algos/npo2.py
ryanjulian: is there a reason they both can't take the same feed?

src/garage/tf/algos/npo2.py
ryanjulian: i'm asking if we can do the same thing for both recurrent and non-recurrent here

ryanjulian:
I am not exactly sure how much work my suggestions about removing `self.policy.recurrent` are -- but if it's a lot of work, you could implement them in #1104 instead.

src/garage/tf/algos/npo2.py
ahtsan: They both take the same feed -- (N, T, S^), and for non-recurrent policies, since

> Can you repost this as modifying, rather than renaming, NPO/PPO/TRPO? Alternatively, if you're not going to do that in this PR, can you send us a branch which does that?
> 
> This PR makes it difficult to compare your changes between the two versions of the algorithms.

That makes sense. I will repost it.

ahtsan:
> Can you repost this as modifying, rather than renaming, NPO/PPO/TRPO? Alternatively, if you're not going to do that in this PR, can you send us a branch which does that?
> 
> This PR makes it difficult to compare your changes between the two versions of the algorithms.

That makes sense. I will repost it.

src/garage/tf/algos/npo2.py
ahtsan: Sampler provides trajectories (N, T, S^) to the algorithm, and then passed to the policies. For recurrent policies, the shape matches. For non-recurrent policies, the shape has to be (N, S^) unless we augment it to be (N, 1, S^) to match the shape.

ahtsan: I think it's part of the API in garage that we need to clean up and unify -- Whether we use dense matrix for both recurrent and non-recurrent policies.

ryanjulian: how hard is it to make the non-recurrent policies accept `(N, T, S^)`? -- currently this loss function already computes assuming the inputs are `(N, T, S^)`

ahtsan: Indeed it computes advantage assuming inputs are `(N, T, S^)`, but then filter with valids to make it `(N * [T], S^)`. To make non-recurrent policies accept `(N, T, S^)`, we basically feed a placeholder with such shape, which means whenever the policy accepts data (observation/actions), we need to augment it from `(N, S^)` to `(N, 1, S^)`.

ryanjulian: Okay -- please look at it. If it's too much work, please file an issue about doing this before submitting.

ahtsan: Regarding augmentation with extra_state_var -- What if a recurrent policy with `state_include_action=False`?

ahtsan: I realized in that case `policy_state_info_list` will be empty (since `extra_state_var` is not added if `state_include_action=False`)

PR #1380, opened 2020-05-04
New plotting decorator for benchmarking

zequnyu:
This PR proposes a decorator for benchmark results plotting. This together with `iterate_experiments` can be used not only for the automatic benchmarking, but also for writing benchmark scripts generally.
For example, [BenchmarkVPG](https://github.com/rlworkgroup/garage/blob/20e2d6587ccf2a636cfb25486a30067f085543d6/tests/benchmarks/garage/tf/algos/benchmark_vpg.py#L41-L118) takes 77 lines to run VPG benchmarking. With the new approach, it would be 4 lines to do the same thing (figures will be generated per `env_id` with all algorithms running the `env_id` as lines):
```py
@benchmark
def ppo_benchmarks():
    iterate_experiments(ppo_garage_tf)
    iterate_experiments(ppo_garage_pytorch)
```
For automatic benchmarking, it would be:
```py
@benchmark(plot=False, auto=True)
def auto_benchmarks():
    iterate_experiments(trpo_garage_pytorch)
    ... # all experiments need to be auto benchmarking
```

### DIFF
--- a/tests/benchmarks/auto/benchmarks.py
+++ b/tests/benchmarks/auto/benchmarks.py
@@ -1,4 +1,4 @@
-"""Run automatic benchmarks."""
+"""Run automatic benchmarking."""
 from baselines.bench import benchmarks
 
 from tests.benchmarks.auto.experiments import ddpg_garage_tf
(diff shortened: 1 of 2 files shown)


zequnyu:
@Mergifyio rebase

PR #1384, opened 2020-05-05
Add max_eval_path_length to off-policy algorithms

krzentner:
Note that the default in this change is `max_path_length`, but maybe it should be 1000.


avnishn:
addresses #1384

PR #1394, opened 2020-05-06
Make runner._sampler optional

krzentner:
This makes the garage runner much more usable for writing algorithms
that don't need the sampler (e.g. fully off-policy algorithms, pure
behavioral cloning, and supervised learning).


krzentner:
Also, with this change, the runner *actually* respects the `RLAlgorithm` API, instead of accessing additional fields that don't technically exist on the `RLAlgorithm` (like `max_path_length` and `policy`).

PR #1396, opened 2020-05-07
Make wrap_exp's archiving handle deleted files

krzentner:
Without this, tar complains that the file was deleted.
I also made it skip archiving files larger than eight mebibytes, since
that's basically never source code.


zequnyu:
@Mergifyio rebase

krzentner:
@Mergifyio rebase

ryanjulian:
@Mergifyio rebase

krzentner:
I've updated the tests to cover every line, but I'm not sure if codecov can detect that, since I need to spawn a subprocess to test archiving.

PR #1397, opened 2020-05-07
Make ExplorationStrategies a type of Policy

krzentner:
This gets us one step closer to deleting OffPolicyVectorizedSampler.


krzentner:
I will un-draft this once I've confirmed that the tests pass.

src/garage/np/exploration_strategies/base.py
```diff
@@ -1,9 +1,72 @@
-class ExplorationStrategy:
-    def get_action(self, t, observation, policy, **kwargs):
+"""Exploration Strategy API used by off-policy algorithms."""
+import abc
+
+
+# This should be an ABC inheritting from garage.Policy, but that doesn't exist
+# yet.
+class ExplorationStrategy(abc.ABC):
```
ryanjulian, line 7: maybe we should be more general and call this a PolicyWrapper? There are reasons to wrap a policy, other than exploration.

benchmarks/experiments/algos/ddpg_garage_tf.py
```diff
@@ -46,14 +46,16 @@ def ddpg_garage_tf(ctxt, env_id, seed):
...
 
-        action_noise = OUStrategy(env.spec, sigma=hyper_parameters['sigma'])
-
         policy = ContinuousMLPPolicy(
             env_spec=env.spec,
             hidden_sizes=hyper_parameters['policy_hidden_sizes'],
             hidden_nonlinearity=tf.nn.relu,
             output_nonlinearity=tf.nn.tanh)
 
+        action_noise = OUStrategy(env.spec,
```
ryanjulian, line 55: continually the generality theme, maybe we should rename the algo param this is passed to `exploration_policy`?

src/garage/np/exploration_strategies/base.py
krzentner: So I do think a general policy interface is emerging, and once it's clear we should define a `PolicyWrapper` type, but I think it would be a little premature to do it in this change, before we have a `Policy` type in the first place.

benchmarks/experiments/algos/ddpg_garage_tf.py
krzentner: May as well.

krzentner:
> Quick question -- Could this lead to some unexpected behavior with pickling, or do we have all of our bases covered?

It should "just work" because the pickler knows that objects can appear multiple times and keeps a "memo" of already serialized objects. In practice, we just pass parameters to the sampler, and those go through the `get_param_values` and `set_param_values` interface.

The major concern with the current setup would be that if a multi-process sampler is used, the state of the `exploration_strategy` in the worker might not get saved. Currently, we assume that all state contained in the sampler is actually contained in the policy parameters, which isn't strictly correct. This is important to fix, but isn't changed either way by this change. I've filed issue #1399 to track this.

> Does the algorithm have to do any extra work to make sure this happens? In general, it's not necessarily aware that that the `exploration_strategy` is backed by the same params as `policy` (or is it?)
I don't think the algorithm should have to do any extra work. It should work either way if the params are shared or not.

benchmarks/src/garage_benchmarks/experiments/algos/ddpg_garage_tf.py
```diff
@@ -5,7 +5,7 @@
 from garage import wrap_experiment
 from garage.envs import normalize
 from garage.experiment import deterministic
-from garage.np.exploration_strategies import OUStrategy
+from garage.np.exploration_policies import OUPolicy
```
ryanjulian, line 8: Hmm, at the cost of being too verbose, should we name these OUExplorationPolicy, GaussianExplorationPolicy, etc?

Rationale: If you come across a random snippet of code, it's not clear that these are wrapping the main policy.

We could also use verb named, e.g. AddOUNoisePolicy AddGaussianNoisePolicy, etc.

benchmarks/src/garage_benchmarks/experiments/policies/continuous_mlp_policy.py
```diff
@@ -46,15 +46,17 @@ def continuous_mlp_policy(ctxt, env_id, seed):
...
-        action_noise = OUStrategy(env.spec, sigma=hyper_params['sigma'])
-
         policy = ContinuousMLPPolicy(
             env_spec=env.spec,
             name='ContinuousMLPPolicy',
             hidden_sizes=hyper_params['policy_hidden_sizes'],
             hidden_nonlinearity=tf.nn.relu,
             output_nonlinearity=tf.nn.tanh)
 
+        exploration_policy = OUStrategy(env.spec,
```
ryanjulian, line 56: did you miss a refactor here?

benchmarks/src/garage_benchmarks/experiments/q_functions/continuous_mlp_q_function.py
```diff
@@ -46,15 +46,17 @@ def continuous_mlp_q_function(ctxt, env_id, seed):
...
-        action_noise = OUStrategy(env.spec, sigma=hyper_params['sigma'])
-
         policy = ContinuousMLPPolicy(
             env_spec=env.spec,
             name='ContinuousMLPPolicy',
             hidden_sizes=hyper_params['policy_hidden_sizes'],
             hidden_nonlinearity=tf.nn.relu,
             output_nonlinearity=tf.nn.tanh)
 
+        exploration_policy = OUStrategy(env.spec,
```
ryanjulian, line 56: missed rename?

examples/tf/ddpg_pendulum.py
```diff
@@ -37,13 +37,13 @@ def ddpg_pendulum(ctxt=None, seed=1):
...
         env = TfEnv(gym.make('InvertedDoublePendulum-v2'))
 
-        action_noise = OUStrategy(env.spec, sigma=0.2)
-
         policy = ContinuousMLPPolicy(env_spec=env.spec,
                                      hidden_sizes=[64, 64],
                                      hidden_nonlinearity=tf.nn.relu,
                                      output_nonlinearity=tf.nn.tanh)
 
+        exploration_policy = OUPolicy(env.spec, policy, sigma=0.2)
```
ryanjulian, line 45: It might be better to call this OrnsteinUhlenbeckPolicy -- it's more googleable.

src/garage/np/exploration_strategies/exploration_strategy.py
```diff
@@ -1,10 +1,72 @@
-class ExplorationStrategy:
+"""Exploration Strategy API used by off-policy algorithms."""
+import abc
 
-    def get_action(self, t, observation, policy, **kwargs):
+
+# This should be an ABC inheritting from garage.Policy, but that doesn't exist
+# yet.
+class ExplorationStrategy(abc.ABC):
```
ryanjulian, line 7: wait -- why leave this AND ExplorationPolicy around?

krzentner: Whoops. Just git things.

benchmarks/src/garage_benchmarks/experiments/q_functions/continuous_mlp_q_function.py
krzentner: Good catch.

benchmarks/src/garage_benchmarks/experiments/policies/continuous_mlp_policy.py
krzentner: Yeah, the benchmarks directory got added after I made the initial change. Good catch.

benchmarks/src/garage_benchmarks/experiments/algos/ddpg_garage_tf.py
krzentner: I actually think the verb form without Policy at the end works best. What do you think?

ryanjulian: I agree

krzentner:
> why there are still two categories of primitives though? ExplorationPolicy + ExplorationStrategy.

Yeah, this was just git rebase not understanding file deletions / moves.

PR #1398, opened 2020-05-07
Fix LocalRunner test failure

krzentner:
I don't know how this got merged with this test broken.


krzentner:
This change is no longer relevant, since I merged #1401.

PR #1400, opened 2020-05-07
Refactor baseline benchmarks

zequnyu:


tests/benchmarks/garage/tf/baselines/benchmark_gaussian_cnn_baseline.py
```diff
@@ -1,147 +0,0 @@
...
-    :param env: Environment of the task.
-    :param seed: Random seed for the trial.
-    :param log_dir: Log dir path.
-    :return:
-    '''
-    deterministic.set_seed(seed)
-    config = tf.compat.v1.ConfigProto(allow_soft_placement=True,
-                                      intra_op_parallelism_threads=12,
-                                      inter_op_parallelism_threads=12)
-    sess = tf.compat.v1.Session(config=config)
```
zequnyu, line 96: @ahtsan Is it safe to use the default session?

benchmarks/benchmark_baselines.py
```diff
@@ -0,0 +1,33 @@
...
+from benchmarks.experiments.baselines import continuous_mlp_baseline_tf
+from benchmarks.experiments.baselines import gaussian_cnn_baseline_tf
+from benchmarks.experiments.baselines import gaussian_mlp_baseline_tf
+
+
+@benchmark
+def ppo_continuous_mlp_baseline_tf_benchmarks():
+    """Run experiments for Continuous MLP Baseline TF benchmarking."""
+    seeds = random.sample(range(100), 3)
+    iterate_experiments(continuous_mlp_baseline_tf, seeds=seeds)
```
ahtsan, line 14: can we make `env_ids` not optional for `iterate_experiments`? Having a default value seems not intuitive.

benchmarks/benchmark_baselines.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+@benchmark
+def ppo_gaussian_cnn_baseline_tf_benchmarks():
+    """Run experiments for Gaussian CNN Baseline TF benchmarking."""
+    env_ids = ['CubeCrash-v0', 'MemorizeDigits-v0']
```
ahtsan, line 20: Since we already have the [environment list](https://github.com/rlworkgroup/garage/tree/master/tests/benchmarks) for benchmarking for different kind of algorithms, should we make it a global dict? (so we won't need to worry about typo for `env_ids` )

tests/benchmarks/garage/tf/baselines/benchmark_gaussian_cnn_baseline.py
ahtsan: I think so within `LocalTFRunner` scope

benchmarks/benchmark_baselines.py
zequnyu: Okey, let's do it in another PR. I need some clarification for this.

zequnyu: Updated.

PR #1401, opened 2020-05-07
Deflake trpo cnn cubecrash

krzentner:
This test keeps causing builds to fail, because the success test is too strict.


krzentner:
I'm not sure. Honestly, we should remove the return value from `runner.train`, since it's not statistically useful. The right thing to do here is to compute an average over the last epochs, but it only trains for 10 total epochs, so I'm not even sure if this test is useful in general.

krzentner:
I'm just going to merge this, since it seems to make things less flaky, at least.

PR #1408, opened 2020-05-09
Add run script for benchmarking

zequnyu:
This PR adds a script to trigger automatic benchmarking.

Install: `cd benchmarks && pip install -e . `

List available benchmarks: `garage_benchmark list`
```
-----Algorithms-----
ddpg_benchmarks
her_benchmarks
ppo_benchmarks
td3_benchmarks
trpo_benchmarks
vpg_benchmarks

-----Policies-----
categorical_cnn_policy_tf_ppo_benchmarks
categorical_gru_policy_tf_ppo_benchmarks
categorical_lstm_policy_tf_ppo_benchmarks
categorical_mlp_policy_tf_ppo_benchmarks
continuous_mlp_policy_tf_ddpg_benchmarks
gaussian_gru_policy_tf_ppo_benchmarks
gaussian_lstm_policy_tf_ppo_benchmarks
gaussian_mlp_policy_tf_ppo_benchmarks

-----Baselines-----
continuous_mlp_baseline_tf_ppo_benchmarks
gaussian_cnn_baseline_tf_ppo_benchmarks
gaussian_mlp_baseline_tf_ppo_benchmarks

-----Q Functions-----
continuous_mlp_q_function_tf_ddpg_benchmarks

-----Automatic benchmarking-----
auto_ddpg_benchmarks
auto_ppo_benchmarks
auto_td3_benchmarks
auto_trpo_benchmarks
auto_vpg_benchmarks
```
Run benchmark(s): `garage_benchmark run ppo_benchmarks trpo_benchmarks`

### DIFF
--- a/benchmarks/src/garage_benchmarks/run_benchmarks.py
+++ b/benchmarks/src/garage_benchmarks/run_benchmarks.py
@@ -1,0 +1,119 @@
+#!/usr/bin/env python
+"""Script for running benchmarking.
+
+Examples:
+    # List all benchmark options
+    benchmark show
+
+    # Run selected benchmarks
+    benchmark run b_1 b_2 ...
+
+"""
+import inspect
+
+import click
+from garage_benchmarks import benchmark_algos
+from garage_benchmarks import benchmark_auto
+from garage_benchmarks import benchmark_baselines
+from garage_benchmarks import benchmark_policies
+from garage_benchmarks import benchmark_q_functions
+
+
+@click.group()
+def cli():
+    """The main command group."""
+
+
+@click.command()
+def list():  # pylint: disable=redefined-builtin
+    """List all benchmarks."""
+    _echo_run_names('Algorithms', _get_runs_dict(benchmark_algos))
+    _echo_run_names('Policies', _get_runs_dict(benchmark_policies))
+    _echo_run_names('Baselines', _get_runs_dict(benchmark_baselines))
+    _echo_run_names('Q Functions', _get_runs_dict(benchmark_q_functions))
+    _echo_run_names('Automatic benchmarking', _get_runs_dict(benchmark_auto))
+
+
+@click.command()
+@click.argument('names', nargs=-1)
+def run(names):
+    """Run selected benchmarks.
+
+    Args:
+        names (tuple): Benchmark names.
+
+    Raises:
+        BadParameter: if any run name is invalid or duplicated.
+
+    """
+    if not names:
+        raise click.BadParameter('Empty names!')
+
+    if len(names) != len(set(names)):
+        raise click.BadParameter('Duplicate names!')
+
+    options = _get_all_options()
+
+    for name in names:
+        if name not in options:
+            raise click.BadParameter(
+                'Invalid run name! Make sure every name can be found in '
+                '`garage_benchmark list`!')
+
+    for name in names:
+        options[name]()
+
+
+cli.add_command(list)
+cli.add_command(run)
+
+
+def _get_all_options():
+    """Return a dict containing all benchmark options.
+
+    Dict of (str: obj) representing benchmark name and its function object.
+
+    Returns:
+        dict: Benchmark options.
+
+    """
+    d = {}
+    d.update(_get_runs_dict(benchmark_algos))
+    d.update(_get_runs_dict(benchmark_policies))
+    d.update(_get_runs_dict(benchmark_baselines))
+    d.update(_get_runs_dict(benchmark_q_functions))
+    d.update(_get_runs_dict(benchmark_auto))
+    return d
+
+
+def _get_runs_dict(module):
+    """Return a dict containing benchmark options of the module.
+
+    Dict of (str: obj) representing benchmark name and its function object.
+
+    Args:
+        module (object): Module object.
+
+    Returns:
+        dict: Benchmark options of the module.
+
+    """
+    d = {}
+    for name, obj in inspect.getmembers(module):
+        if inspect.isfunction(obj) and name.endswith('benchmarks'):
+            d[name] = obj
+    return d
+
+
+def _echo_run_names(header, d):
+    """Echo run names to the command line.
+
+    Args:
+        header (str): The header name.
+        d (dict): The dict containing benchmark options.
+
+    """
+    click.echo('-----' + header + '-----')
+    for name in d:
+        click.echo(name)
+    click.echo()
--- a/benchmarks/setup.py
+++ b/benchmarks/setup.py
@@ -17,4 +17,9 @@ REQUIRED = [
 setup(name='garage_benchmarks',
       packages=find_packages(where='src'),
       package_dir={'': 'src'},
-      install_requires=REQUIRED)
+      install_requires=REQUIRED,
+      include_package_data=True,
+      entry_points='''
+              [console_scripts]
+              garage_benchmark=garage_benchmarks.run_benchmarks:cli
+          ''')
--- a/benchmarks/src/garage_benchmarks/benchmark_auto.py
+++ b/benchmarks/src/garage_benchmarks/benchmark_auto.py
@@ -1,0 +1,48 @@
+"""Automatic benchmarking."""
+from garage_benchmarks.experiments.algos import ddpg_garage_tf
+from garage_benchmarks.experiments.algos import ppo_garage_pytorch
+from garage_benchmarks.experiments.algos import ppo_garage_tf
+from garage_benchmarks.experiments.algos import td3_garage_tf
+from garage_benchmarks.experiments.algos import trpo_garage_pytorch
+from garage_benchmarks.experiments.algos import trpo_garage_tf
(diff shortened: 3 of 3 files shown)


benchmarks/auto_benchmarking.py
```diff
@@ -0,0 +1,88 @@
+#!/usr/bin/env python3
+"""Run automatic benchmarking."""
+# pylint: disable=wrong-import-order
+import warnings
+
+import click
+
+# pylint: disable=unused-import
+import garage.config  # noqa: F401
```
ryanjulian, line 9: I don't think this exists?

src/garage/config.py
```diff
@@ -0,0 +1,6 @@
+"""Config garage project path."""
+import pathlib
+import sys
```
ryanjulian, line 3: Oh -- please don't do this. Appending the PYTHONPATH using sys is brittle and doesn't work in many scenarios. 

If you can guarantee that your script will be called from the root of the repo, the benchmarks package should exist by virtue of being a directory.

Everything which appears in the `garage.` namespace has to be in src/garage, and the fact that it's difficult to import garage packages without installing using some environment manager is actually intentional (it helps us catch bugs which only happen for people who install the package )

How are you planning on running this code? That will inform how to deal with the namespaces.

krzentner: Honestly, I think the more correct way of handling the needs here is to have another `setup.py` in `benchmarks`, that can install a package called `garage_benchmarks` of something. If that were the case, to run the benchmark script you would first do `pip install -e ./benchmarks` from the garage root directory.

It would require moving the files in benchmarks into a subdirectory though, like `benchmarks/garage_benchmarks`.

krzentner: I've implemented what I describe here in #1411.

benchmarks/auto_benchmarking.py
```diff
@@ -0,0 +1,88 @@
+#!/usr/bin/env python3
```
ryanjulian, line 1: perhaps you can add this as a subcommand of the existing `garage` script in `scripts/`?

e.g. so that you could do `garage benchmark foo bar baz`

benchmarks/auto_benchmarking.py
```diff
@@ -0,0 +1,88 @@
...
+
+@benchmark(plot=False, auto=True)
+def auto_vpg_benchmarks():
+    """Run experiments for VPG benchmarking."""
+    iterate_experiments(vpg_garage_pytorch, MuJoCo1M_ENV_SET)
+    iterate_experiments(vpg_garage_tf, MuJoCo1M_ENV_SET)
+
+
+@click.command()
+@click.argument('runs')
```
ryanjulian, line 61: I think it would be less surprising here if you used a variadic (variable-length) argmument here, rather than forcing the user to format a list on the terminal. That would allow the user to call

`python3 benchmarks/auto_benchmarking.py auto_vpg_benchmarks auto_trpo_benchmarks`

or alternatively, for just one set:

`python3 benchmarks/auto_benchmarking.py auto_trpo_benchmarks`

benchmarks/auto_benchmarking.py
```diff
@@ -0,0 +1,88 @@
...
+
+    """
+    benchmarks = [
+        auto_ddpg_benchmarks, auto_ppo_benchmarks, auto_td3_benchmarks,
+        auto_trpo_benchmarks, auto_vpg_benchmarks
+    ]
+    runs = [b for b in benchmarks if b.__name__ in runs]
+
+    if not runs:
+        warnings.warn('No benchmark to run! Make sure you pass a list'
```
ryanjulian, line 78: if you are going to constrain the user to a hard-to-identify set, your error message should include a list of the valid options.

benchmarks/auto_benchmarking.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+@benchmark(plot=False, auto=True)
+def auto_vpg_benchmarks():
+    """Run experiments for VPG benchmarking."""
+    iterate_experiments(vpg_garage_pytorch, MuJoCo1M_ENV_SET)
+    iterate_experiments(vpg_garage_tf, MuJoCo1M_ENV_SET)
+
+
+@click.command()
```
ryanjulian, line 60: since it is already immediately-available, i think this command should have the option to list the available benchmarks, e.g.

`garage benchmarks list` -- lists the available benchmark jobs
`garage benchmarks run foo_benchmark bar_benchmark` -- runs foo_benchmark and bar_benchmark

benchmarks/auto_benchmarking.py
```diff
@@ -0,0 +1,88 @@
...
+    """Run experiments for PPO benchmarking."""
+    iterate_experiments(ppo_garage_pytorch, MuJoCo1M_ENV_SET)
+    iterate_experiments(ppo_garage_tf, MuJoCo1M_ENV_SET)
+
+
+@benchmark(plot=False, auto=True)
+def auto_td3_benchmarks():
+    """Run experiments for TD3 benchmarking."""
+    td3_env_ids = [
+        env_id for env_id in MuJoCo1M_ENV_SET if env_id != 'Reacher-v2'
```
maliesa96, line 40: Is there a reason we're skipping Reacher v2?

benchmarks/auto_benchmarking.py
maliesa96: If this is done then adding an example to the README would be great. Maybe garage_benchmarks should have its own README?

zequnyu: Ye I haven't started to write any high level docs. I'll definitely add those.

benchmarks/auto_benchmarking.py
zequnyu: Perhaps @krzentner or @ahtsan is the right person to answer it?

benchmarks/auto_benchmarking.py
zequnyu: @ryanjulian After some thinking I think it's probably better not to put this script in `garage`? My point is the installation for benchmarking is purely optional, so it makes sense to me to create a script under `benchmarks` folder. Though I think I could also make it useful for general benchmarking, which also justifies the point. So something like this:
```
benchmark list
benchmark run b1 b2
...
````

**Comment by zequnyu** [comment]
This is ready for another review.

Install: `pip install -e . `

Show benchmarks: `benchmark show`
```
-----Algorithms-----
ddpg_benchmarks
her_benchmarks
ppo_benchmarks
td3_benchmarks
trpo_benchmarks
vpg_benchmarks

-----Policies-----
categorical_cnn_policy_tf_ppo_benchmarks
categorical_gru_policy_tf_ppo_benchmarks
categorical_lstm_policy_tf_ppo_benchmarks
categorical_mlp_policy_tf_ppo_benchmarks
continuous_mlp_policy_tf_ddpg_benchmarks
gaussian_gru_policy_tf_ppo_benchmarks
gaussian_lstm_policy_tf_ppo_benchmarks
gaussian_mlp_policy_tf_ppo_benchmarks

-----Baselines-----
continuous_mlp_baseline_tf_ppo_benchmarks
gaussian_cnn_baseline_tf_ppo_benchmarks
gaussian_mlp_baseline_tf_ppo_benchmarks

-----Q Functions-----
continuous_mlp_q_function_tf_ddpg_benchmarks

-----Automatic benchmarking-----
auto_ddpg_benchmarks
auto_ppo_benchmarks
auto_td3_benchmarks
auto_trpo_benchmarks
auto_vpg_benchmarks
```
Run benchmark(s): `benchmark run ppo_benchmarks trpo_benchmarks`

File: benchmarks/auto_benchmarking.py (hunk 3 above)
Comment (ryanjulian) [reply benchmarks/auto_benchmarking.py]: yes, we can consider making it part of the `garage` command later.

i recommend you make the script named `garage-benchmark` because who knows what other executable someone has on their machine named "benchmark".

you can add this line to your setup.py to ensure your command is installed in the environment (without having to use ./ or add to PATH): https://github.com/rlworkgroup/garage/blob/master/setup.py#L100

Comment (zequnyu) [reply benchmarks/auto_benchmarking.py]: [This](https://github.com/rlworkgroup/garage/blob/ecd749fe76be759a9e402b88810eb99a566c261e/benchmarks/setup.py#L22-L25) should do the same magic. You just make sure you're in `/benchmarks`.

File: benchmarks/src/garage_benchmarks/run_benchmarks.py [hunk 8]
```diff
@@ -0,0 +1,119 @@
...
+    if len(names) != len(set(names)):
+        raise click.BadParameter('Duplicate names!')
+
+    options = _get_all_options()
+
+    for name in names:
+        if name not in options:
+            raise click.BadParameter(
+                'Invalid run name! Make sure every name can be found in '
+                '`benchmark list`!')
```
gitanshu, line 61: I guess this should be
```suggestion
                '`garage_benchmark list`!')
```

zequnyu:
@Mergifyio rebase

PR #1409, opened 2020-05-10
Refactor policy and q-function benchmarking

zequnyu:
This should be the last PR before getting  #1325 done.

This also contains a fix for #1411 `setup.py`'s location and moves everything to `/src`.


krzentner:
I think the location you put `setup.py` is really weird. Usually, we would put `setup.py` in `/benchmarks`, and add `packages=find_packages(where='src'), package_dir={'': 'src'},` to the `setup()` call.

Moving the files into `src` is generally recommended, so I'm fine with this change, but you should probably fix where you've put `setup.py`.

PR #1410, opened 2020-05-11
Refactor NPO and REPS to use primitives with tfp.distribution

ahtsan:
This PR does the following:
* Rename primitives with tfp.distribution version.
* Refactor NPO and REPS to use primitives with tfp.distribution.
* Modify primitives so all primitives expect trajectory input with
  shape (batch, time, obs_dim).
* Update tests accordingly.

P.S. The only old primitive that we will keep is `GaussianMLPModel` because it's still being used by `GaussianMLPRegressor`. It's named to `GaussianMLPModel2`.

<hr>

Benchmarks for all primitives (old version v.s. tfp.distribution version) look good.
*garage2 = tfp.distribution version

**CategoricalMLPPolicy**

**CategoricalGRUPolicy**

**CategoricalLSTMPolicy**

**CategoricalCNNPolicy**

**GaussianMLPPolicy**

**GaussianGRUPolicy**

**GaussianLSTMPolicy**


krzentner:
So garage2 in the above is using tfp? It looks like it fixes some bugs.

ahtsan:
> So garage2 in the above is using tfp? It looks like it fixes some bugs.

Yes, seem so. The benchmark is surprisingly good.

examples/np/cem_cartpole.py
```diff
@@ -35,7 +35,9 @@ def cem_cartpole(ctxt=None, seed=1):
         policy = CategoricalMLPPolicy(name='policy',
                                       env_spec=env.spec,
                                       hidden_sizes=(32, 32))
-
+        input_var = env.observation_space.to_tf_placeholder(name='obs',
```
krzentner, line 38: Is there a strong reason this can't happen in the policy? It makes the experiment launcher a lot less obvious to me.

examples/np/cma_es_cartpole.py
```diff
@@ -32,7 +32,9 @@ def run_task(snapshot_config, *_):
         policy = CategoricalMLPPolicy(name='policy',
                                       env_spec=env.spec,
                                       hidden_sizes=(32, 32))
-
+        input_var = env.observation_space.to_tf_placeholder(name='obs',
```
krzentner, line 35: Same here.

examples/np/cem_cartpole.py
ahtsan: Indeed, we can. It's just it no longer is a `np` algorithm (which it actually isn't in the first place, given it's using tf primitives)

ahtsan: or we can use a `np` primitive?

ahtsan:
> I love 99% of this change. Two small things to check:
> 
>     * In the CEM experiment launchers, `policy.build` is called manually. Is that avoidable? It's not that big a deal to me anyways, since CEM isn't all that popular, and probably shouldn't use tf policies anyways.
> 
>     * Have you coordinated with Keren who / when TENPO is going to be ported to the new primitives? Now that I've seen the port here, it doesn't look to hard, but we also should merge all of this before the release.

I haven't talked to @naeioi yet, will do soon. Yes, I think it shouldn't be hard and won't take long.

src/garage/tf/policies/categorical_cnn_policy.py
```diff
@@ -1,15 +1,18 @@
...
-from garage.tf.models import CNNModel
-from garage.tf.models import MLPModel
-from garage.tf.models import Sequential
-from garage.tf.policies.policy import StochasticPolicy
+from garage.tf.models import CategoricalCNNModel
+from garage.tf.policies.policy import StochasticPolicy2
 
 
-class CategoricalCNNPolicy(StochasticPolicy):
+class CategoricalCNNPolicy(StochasticPolicy2):
```
ryanjulian, line 15: why still have StochasticPolicy2 if you are getting rid of CategoricalCNNPolicy2, etc?

ahtsan: `StochasticPolicy` has `dist_info_sym` which is needed in `GaussianMLPTaskEmbeddingPolicy`.
I am refactoring `GaussianMLPTaskEmbeddingPolicy` rn so `StochasticPolicy2` will be gone later as well.

examples/np/cem_cartpole.py
ryanjulian: I think we will have to consider this relationship later as we write the primitive interfaces, so for now just do whatever is easiest. This algorithm is most interesting to KR and I, and fewer users actually use it, anyway.

ryanjulian:
🙌

avnishn:
@ahtsan im not sure if I understand the results of the gaussian mlp policy on half cheetah env. How many time steps are there in an evaluation iteration? The average returns for half cheetah seem low depending on how many timesteps the policy was trained on

ahtsan:
@avnishn Do you mean `batch_size`? It's 2048.

avnishn:
Yeah that's what I was looking for, thanks :). So the returns look a little low for 250,000 timesteps trained on. Can someone else besides me look at this. @yonghyuc @krzentner @ryanjulian @naeioi

ahtsan:
I can try again with a set of good hyperparameters. My purpose of the results above is to show that under the **same** set of hyperparameters, new version performs at least as good as the old verison.

PR #1411, opened 2020-05-11
Create garage_benchmarks package

krzentner:
I think this should address @zequnyu 's usecase. Basically all I did was move everything in `garage/benchmarks` to `garage/benchmarks/garage_benchmarks`, updated the imports, then added a `setup.py` to allow `pip install -e ./benchmarks` to work.


ryanjulian:
Another option is to make this `garage.benchmarks` but only install it via an option. We can consider that later -- this is great for now.

krzentner:
> 
> 
> This makes sense. Seems to me `setup.py` is a good place for #1325?

I hadn't thought of that, but yes, I think that's completely right.

PR #1413, opened 2020-05-11
Doc modernization

krzentner:
This change is stacked on the update to the examples, so it looks really big, but the doc change is the second commit and is fairly tall.

At least for the time being, a server with the built docs is available [on the VPN here](http://172.18.1.9:8000/index.html).


krzentner:
Note that this is only about the first third or so of the updates I think we should do before the release.

docs/user/installation.rst
```diff
@@ -76,10 +57,48 @@ Alternatively, you can add garage in the pip section of your `environment.yml`
     pip install garage
 
 
+Install System Dependencies (Optional)
```
gitanshu, line 60: I feel like this section talks about both system deps and installing mujoco stuff, but the heading obviously doesn't convey that. And the content mostly says that you need system deps only for mujoco etc stuff. So maybe we can call this `Optional dependencies`?

docs/user/installation.rst
```diff
@@ -76,10 +57,48 @@ Alternatively, you can add garage in the pip section of your `environment.yml`
...
+
+.. code-block:: bash
+
+    ./scripts/setup_linux.sh --mjkey path-to-your-mjkey.txt --modify-bashrc
+
+- On macOS:
+
+.. code-block:: bash
+
+    ./scripts/setup_macos.sh --mjkey path-to-your-mjkey.txt --modify-bashrc
```
maliesa96, line 88: The --modify-bashrc argument is invalid in setup_macox.sh. This should be
` ./scripts/setup_macos.sh --mjkey path-to-your-mjkey.txt --modify-shellrc`

krzentner: Good catch, thanks!

ryanjulian:
<3

PR #1414, opened 2020-05-12
Add function to upload auto benchmarking result to GCP

zequnyu:
Also does:
 - When running the same benchmark repeatedly, it stores to a separate folder instead of deleting the previous one.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,7 +12,7 @@ repos:
     -   id: check-pylint-normal
         name: Check pylint (except unit tests)
         types: [file, python]
-        exclude: ^tests/garage/.*$  # exclude unit tests
+        exclude: (^tests/garage/.*$|^benchmarks/)  # exclude unit tests and benchmarks
         require_serial: true  # pylint does its own parallelism
         language: system
         entry: pylint
--- a/benchmarks/src/garage_benchmarks/helper.py
+++ b/benchmarks/src/garage_benchmarks/helper.py
@@ -35,8 +35,8 @@ import json
 import os
 import pathlib
 import random
-import shutil
 
+from google.cloud import storage
 import matplotlib.pyplot as plt
 import numpy as np
 import tensorflow as tf
@@ -44,6 +44,8 @@ import tensorflow as tf
 _plot = None
 _log_dir = None
 _auto = False
+
+_bucket = storage.Client().bucket('resl-garage-benchmarks')
 
 
 def benchmark(exec_func=None, *, plot=True, auto=False):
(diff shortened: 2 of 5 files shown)


benchmarks/src/garage_benchmarks/helper.py
```diff
@@ -72,8 +72,12 @@ def wrapper_func():
         plt.close('all')
 
         _log_dir = _get_log_dir(exec_func.__name__)
+
         if os.path.exists(_log_dir):
-            shutil.rmtree(_log_dir)
+            count = 1
+            while os.path.exists(_log_dir + '(' + str(count) + ')'):
```
ryanjulian, line 78: maybe make this path a little easier to type on the terminal (i.e. no (), which have to be escaped)

benchmarks/src/garage_benchmarks/helper.py
```diff
@@ -211,3 +218,24 @@ def _export_to_json(json_name, xs, xlabel, ys, ylabel, ys_std):
                  y_max=(ys + ys_std).tolist(),
                  xlabel=xlabel,
                  ylabel=ylabel), json_file)
+
+
+def _upload_to_gcp_storage(exec_dir):
```
ryanjulian, line 223: i think you should be more conservative here and upload the entire experiment directory.

zequnyu: Can you develop your point? The rest is not being used by the auto benchmarking.

ryanjulian: What if you want to give users the opportunity to download the pre-trained policy from the website?
What if we see a regression and want to dig deeper and see if it was a code problem, or an unlucky seed?

krzentner: Yeah, I agree. As long as we use `snapshot_mode='last'`, we don't have any reason not to upload the entire experiment directory. It'll definitely help with debugging, and will cost us mere cents / month.

.pre-commit-config.yaml
```diff
@@ -12,7 +12,7 @@ repos:
     -   id: check-pylint-normal
         name: Check pylint (except unit tests)
         types: [file, python]
-        exclude: ^tests/garage/.*$  # exclude unit tests
+        exclude: (^tests/garage/.*$|^benchmarks/)  # exclude unit tests
```
ryanjulian, line 15: if you are going to do this because you feel some rules shouldn't apply to the directory, please add a custom .pylintrc to the `benchmarks/` dir and a custom pylint job to pre-commit, as we already do for tests.

zequnyu:
@Mergifyio rebase

PR #1417, opened 2020-05-14
Replace tf.dense layer with tf.keras.Dense 

ahtsan:


ahtsan:
so I realized to make this change, it's more than just renaming -- keras layer has a different behavior than tf.layers.dense. We need to store a keras.Dense instance and reuse it for accessing the underlying variables.

ryanjulian:
Would your problem go away if we deleted `mlp.py` and instead moved this into MLPModel? Then MLPModel can own all the Dense layers. Users of mlp.py would instead have to use MLPModel.

ahtsan:
Yes, that’s what I figured out yesterday — essentially we manage the keras layer object in MLPModel, and replace mlp with MLPModel across the codebase.

ryanjulian:
Seems not so bad:
```sh
$ rg "import mlp" -l
src/garage/tf/models/mlp_dueling_model.py
src/garage/tf/models/gaussian_mlp_model.py
src/garage/tf/models/mlp_model.py
src/garage/tf/models/gaussian_mlp_model2.py
src/garage/tf/models/gaussian_cnn_model.py
src/garage/tf/models/mlp_merge_model.py
tests/garage/tf/models/test_mlp.py
tests/garage/tf/models/test_model.py
tests/garage/tf/models/test_mlp_concat.py
```

PR #1419, opened 2020-05-15
Fix pixel normalization in CNN Q functions

### LINKED ISSUE #1418: CNN Q functions don't normalize images correctly
Observation inputs should be passed to the initial placeholder (so that they are correctly normalized) rather than passing them directly to the model.

maliesa96:
Pixel observations are now passed to the initial (unaugmented) placeholder rather directly to the model (fixes #1418)

Also updated the tests to reflect this.

### DIFF
--- a/src/garage/tf/q_functions/continuous_cnn_q_function.py
+++ b/src/garage/tf/q_functions/continuous_cnn_q_function.py
@@ -3,7 +3,7 @@ import akro
 import tensorflow as tf
 
 from garage.tf.models import CNNMLPMergeModel
-from garage.tf.q_functions.q_function import QFunction
+from garage.tf.q_functions import QFunction
 
 
 class ContinuousCNNQFunction(QFunction):
@@ -66,7 +66,6 @@ class ContinuousCNNQFunction(QFunction):
             of output dense layer(s) in the MLP. The function should return
             a tf.Tensor.
         layer_normalization (bool): Bool for using layer normalization or not.
-
     """
 
     def __init__(self,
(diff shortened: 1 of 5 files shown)


maliesa96:
Done. I just added a couple of lines that make sure the right placeholder is getting passed to the model depending on whether or not the observation is an image.

src/garage/tf/q_functions/continuous_cnn_q_function.py
```diff
@@ -184,11 +191,10 @@ def get_qval(self, observation, action):
                 corresponding to each (obs, act) pair.
 
         """
-        if isinstance(self._env_spec.observation_space, akro.Image):
-            if len(observation.shape) <= 3:
-                observation = self._env_spec.observation_space.unflatten(
-                    observation)
-            observation = observation / 255.0
+        if len(observation[0].shape) < len(self._obs_dim):
```
ahtsan, line 194: What does this line do?

ryanjulian:

ryanjulian:
@maliesa96 can you ensure that any relevant fixes get backported to release-2019.10?

src/garage/tf/q_functions/continuous_cnn_q_function.py
maliesa96: Unflattens observations that have been flattened (`observation` has a batch dimension). Ideally this should be `get_qvals(observations, actions)` to avoid this confusion, though I left it as is to remain consistent with the other primitives.

maliesa96: Sure I'll run DQN.

maliesa96:
2019.10 wasn't doing any pixel normalization so its not an issue in that release. Do we still want to backport?

ryanjulian:
as long as this bug definitely didn't affect the branch -- no. feel free to close the issue i filed if that's the case.

PR #1423, opened 2020-05-20
Force Python 3.5 in the CI

ryanjulian:


Makefile
```diff
@@ -13,7 +13,7 @@ DATA_PATH ?= $(shell pwd)/data
 MJKEY_PATH ?= ~/.mujoco/mjkey.txt
 
 test:  ## Run the CI test suite
-test: RUN_CMD = nice -n 11 pytest -v -m 'not huge and not flaky' --durations=0
+test: RUN_CMD = nice -n 11 pytest -v -n 5 -m 'not huge and not flaky' --durations=0
```
krzentner, line 16: Why `pytest -n 5`?

ryanjulian: I found that this hits max load average on my 10-core workstation.

krzentner: Okay, I guess that's as good a reason as any.

ryanjulian:
I don't have time this week to fix these test breakages, if someone wants to pick this up.

Makefile
ryanjulian, line 16: @gitanshu i did some experiments on my workstation and determined that `-n auto` is probably the best setting for this

tests/helpers.py
```diff
@@ -250,3 +253,34 @@ def max_pooling(_input, pool_shape, pool_stride):
                                pool_shape, k])
 
     return results
+
+
+# Taken from random.choices in Python 3.6 source since it's not available in
```
ryanjulian, line 258: nice!

ryanjulian:
This needs backport into 2019.10.

FYI, since at this point we have missed a full release cycle, I'm going to extend the maintenance window for 2019.10 by another 4 months.

ryanjulian:
Can you explain why this needs force-pushing? Just touches a lot of files?

If that's the case, just let me know when it's ready.

gitanshu:
> Can you explain why this needs force-pushing? Just touches a lot of files?
> 
> If that's the case, just let me know when it's ready.

It's only 1 file: tests/helpers.py that has a lot of pylint warnings because of undocumented params etc. I am not sure why we use a separate pylintrc for tests/garage only and not other subdirs of tests/

@krzentner @avnishn @maliesa96 this PR needs to be reviewed again

ryanjulian:
@gitanshu 
the pylintrc's really only differ in their documentation rules and a couple other small details. the main difference is in documentation requirements, so I'll focus on that.

the rationale is as follows:
* `src/`, `examples/`, `scripts/` follow the strictest .pylintrc (full docstrings), because they are "production" code which is likely to get read by users, either to understand garage or to fix a bug
* `tests/garage` is composed of unit tests, which should be self-documenting and are not called by anyone other than the test runner. unit tests have external caller, they don't need documentation meant to guide external callers on correct usage.
* `tests/helpers`, `tests/fixtures`, `tests/integration_tests`, etc. are all either (1) helper code for tests, or (2) integration tests, which tend to be more complex. in the former case, this code has external callers who need documentation and, more urgently, correct usage of these helpers is essential to the correctness of the tests which use them. put another way, if you mis-use something in `tests/helpers.py` as a test author, you may inadvertently help someone create a false-positive or false-negative tests. this has burned @krzentner and I many times before. integration tests have a habit of being large and complicated, and can probably benefit from additional documentation as well.

docker/Dockerfile.base.18.04
```diff
@@ -63,6 +74,9 @@ RUN mkdir -p /root/.mujoco && \
   ln -s $HOME/.mujoco/mujoco200_linux $HOME/.mujoco/mujoco200
   ENV LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/root/.mujoco/mujoco200/bin
 
+
+RUN python3.5 -m ensurepip && python3.5 -m pip install --upgrade pip
```
ryanjulian, line 78: for `RUN` statements, please use the convention of one statement per line, terminated with `&& \`

ryanjulian, line 78: how is this distinct from line 102 below?

gitanshu: line 102 upgrades pip inside virtual env. This line installs basically replaces install of `python3-pip` package since it the python install doesn't come with pip. We use the system install of pip to do a system install of pipenv later

ryanjulian: can you leave a comment to this effect?

ryanjulian:
Let me know when this is ready.

ryanjulian:
Looks like the MuJoCo job is blocking you. I suggest you just add a second MuJoCo job (perhaps have marks `mujoco` for unit tests and `mujoco_long` for long MuJoCo tests). 

Even if we get that to pass here, we don't want to commit a flaky CI pipeline. We have a combination of wasteful tests and increasingly-overtaxed testing capacity, and this PR isn't going to fix that.

gitanshu:
@ryanjulian this is ready to go

ryanjulian:
@mergifyio backport release-2019.10

PR #1424, opened 2020-05-20
Move to multi-stage docker builds

gitanshu:


ryanjulian:
A TODO list for this PR:
* [ ] Convert the 16.04 Dockerfile
* [ ] Name targets into the rlworkgroup/ namespace
* [ ] Figure out how to do the run-nvidia target (rather than forking the whole file, perhaps just use the pattern below and do it by passing an arg from inside the Makefile)

```Docker
ARG PARENT_IMAGE=ubuntu:18.04
FROM $PARENT_IMAGE as garage-18.04
```

```Makefile
build-headless: TAG ?= rlworkgroup/garage-headless:latest
build-headless: docker/Dockerfile
	docker build \
		-f docker/Dockerfile \
		--target garage-dev-18.04 \
		-t ${TAG} \
		${BUILD_ARGS} .

build-nvidia: TAG ?= rlworkgroup/garage-nvidia:latest
build-nvidia: docker/Dockerfile
	docker build \
		-f docker/Dockerfile \
		--target garage-dev-18.04 \
		-t ${TAG} \
                 --build_arg PARENT_IMAGE=nvidia/cuda:10.2-runtime-ubuntu18.04
		${BUILD_ARGS} .
```

This would hopefully allow us to support a ton of variants of the nvidia image by just adding some make targets. Alternatively, we can duplicate the file for each variant (of parent image) we want.

ryanjulian:
You should also probably work https://github.com/rlworkgroup/garage/pull/1423 into this PR -- though it's probably easier to submit that one first, because it requires some code fixes.

.dockerignore
```diff
@@ -1 +1,3 @@
 data
+.git
```
ryanjulian, line 2: this doesn't work for the garage-dev target though, right? or does COPY not obey .dockerignore

docker/Dockerfile
```diff
@@ -0,0 +1,156 @@
...
+COPY --chown=garage-user:garage-user src/garage/__init__.py $HOME/code/garage/src/garage/__init__.py
+COPY --chown=garage-user:garage-user setup.py $HOME/code/garage/setup.py
+WORKDIR $HOME/code/garage
+
+# Prevent pip from complaining about available upgrades
+RUN pip install --upgrade pip
+
+# Pre-build pre-commit env
+COPY --chown=garage-user:garage-user .pre-commit-config.yaml $HOME/code/garage
+RUN ls -la && echo "$(id -u)" && ls -la $HOME
```
ryanjulian, line 130: remove this?

docker/Dockerfile
```diff
@@ -0,0 +1,156 @@
...
+ENV HOME /home/garage-user
+WORKDIR $HOME
+
+# MuJoCo 2.0 (for dm_control)
+RUN mkdir -p $HOME/.mujoco && \
+  wget https://www.roboti.us/download/mujoco200_linux.zip -O mujoco.zip && \
+  unzip mujoco.zip -d $HOME/.mujoco && \
+  rm mujoco.zip && \
+  ln -s $HOME/.mujoco/mujoco200_linux $HOME/.mujoco/mujoco200
+  ENV LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$HOME/.mujoco/mujoco200/bin
```
ryanjulian, line 66: bad indent?

docker/Dockerfile
```diff
@@ -0,0 +1,156 @@
...
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
+
+# Prevent pip from complaining about available upgrades
+RUN pip install --upgrade pip
+
+# We need a MuJoCo key to install mujoco_py
+# In this step only the presence of the file mjkey.txt is required, so we only
+# create an empty file
+RUN touch $HOME/.mujoco/mjkey.txt && \
+  pip install garage[all]==2020.5rc1 && \
```
ryanjulian, line 80: this should probably just be `garage[all]` since we want the container to track whatever people will download from pip. the dev version uses the lastest

docker/Dockerfile
```diff
@@ -0,0 +1,156 @@
...
+RUN python3 -m virtualenv --python=/usr/bin/python3 $VIRTUAL_ENV
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
+
+# Prevent pip from complaining about available upgrades
+RUN pip install --upgrade pip
+
+# We need a MuJoCo key to install mujoco_py
+# In this step only the presence of the file mjkey.txt is required, so we only
+# create an empty file
+RUN touch $HOME/.mujoco/mjkey.txt && \
```
ryanjulian, line 79: a nice feature here if to check the secrets for MJKEY and it it's non empty, pre-build the mujoco_py bindings. currently the container spends the first minute of every run building them.

```Dockerfile
# using the new secrets feature
RUN --mount=type=secret,id=mjkey,dst=$HOME/.mujoco/mjkey.txt && \
    pip install garage[all] && \
    if [[ -s $HOME/.mujoco/mjkey.txt ]]; then python -c 'import mujoco_py'; fi && \
    rm -r $HOME/.cache/pip
```

docker/Dockerfile
```diff
@@ -0,0 +1,156 @@
...
+  cd glfw-3.3 && \
+  mkdir glfw-build && \
+  cd glfw-build && \
+  cmake -DBUILD_SHARED_LIBS=ON -DGLFW_BUILD_EXAMPLES=OFF -DGLFW_BUILD_TESTS=OFF -DGLFW_BUILD_DOCS=OFF .. && \
+  make -j"$(nproc)" && \
+  make install && \
+  cd ../../ && \
+  rm -rf glfw
+
+RUN groupadd -g 999 garage-user && \
```
ryanjulian, line 53: perhaps we can do something like this

```Dockerfile
ARG user=garage-user
ARG uid=999
RUN groupadd -g $uid $user && \
  useradd -m -r -u $uid -g $user $user && \
  chown -R $user:$user /home/$user
USER $user
ENV HOME /home/$user
WORKDIR $HOME
```

Then on the build command line (i.e. Makefile) we can do 
```sh
docker build --build_arg user="${USER}" --build_arg uid="${UID}" .
```
and the user/UID inside the container will match that of the building user. For people using the Dockerfile locally, this means files in `data/` will be created/owned by the user who built the dockerfile, which would cut down on weird permissions issues

docker/Dockerfile
```diff
@@ -0,0 +1,192 @@
...
+# create an empty file
+RUN touch $HOME/.mujoco/mjkey.txt
+RUN --mount=type=secret,id=mjkey,dst=$HOME/.mujoco/mjkey.txt pip install garage && \
+    pip install garage[all] && \
+    if [[ -s $HOME/.mujoco/mjkey.txt ]]; then python -c 'import mujoco_py'; fi && \
+    rm -r $HOME/.cache/pip
+
+#RUN touch $HOME/.mujoco/mjkey.txt && \
+#  pip install garage && \
+#  pip install garage[all] && \
```
ryanjulian, line 107: remove?

Makefile
```diff
@@ -10,13 +10,26 @@ SHELL := /bin/bash
 DATA_PATH ?= $(shell pwd)/data
 # Set the environment variable MJKEY with the contents of the file specified by
 # MJKEY_PATH.
-MJKEY_PATH ?= ~/.mujoco/mjkey.txt
+MJKEY_PATH ?= ${HOME}/.mujoco/mjkey.txt
+BUILD_ARGS = --build-arg user="${USER}" --build-arg uid=${shell id -u}
```
ryanjulian, line 14: look like maybe you can remove this variable, since it doesn't need to get overridden?

docker/Dockerfile
```diff
@@ -0,0 +1,192 @@
+# syntax=docker/dockerfile:experimental
+ARG USER=garage-user
+ARG UID=999
+ARG PARENT_IMAGE=ubuntu:18.04
+
+FROM $PARENT_IMAGE AS garage-18.04
```
ryanjulian, line 6: these should all be in the `rlworkgroup/` namespace, for when they get uploaded to Docker Hub

Makefile
```diff
@@ -10,13 +10,33 @@ SHELL := /bin/bash
...
 # Set the environment variable MJKEY with the contents of the file specified by
 # MJKEY_PATH.
-MJKEY_PATH ?= ~/.mujoco/mjkey.txt
-
-test:  ## Run the CI test suite
-test: RUN_CMD = nice -n 11 pytest -v -n auto -m 'not huge and not flaky' --durations=0
+MJKEY_PATH ?= ${HOME}/.mujoco/mjkey.txt
+
+# BuildKit is required for the --secret flag in docker. We use it to mount mjkey.txt during build time
+#export DOCKER_BUILDKIT=1
```
ryanjulian, line 16: how should we deal with this long-term?

Makefile
```diff
@@ -137,19 +165,19 @@ run-ci:
 		${TAG} ${RUN_CMD}
 
 run-headless: ## Run the Docker container for headless machines
-run-headless: CONTAINER_NAME ?= garage-headless
+run-headless: TAG ?= rlworkgroup/garage-headless
```
ryanjulian, line 168: thank you

docker/entrypoint-verify-envs.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+# Dockerfile entrypoint
+set -e
+
+pip install --upgrade pip setuptools
+pip install pipenv
+pipenv --python=3.5
```
ryanjulian, line 7: can we keep this in the Makefile? It allows us to tweak the entire CI pipeline by only editing one file.

docker/Dockerfile
```diff
@@ -0,0 +1,203 @@
...
+# Build and install the sdist
+RUN python3.5 setup.py sdist && \
+    cp dist/*.tar.gz dist/garage.tar.gz && \
+    pip install --upgrade --upgrade-strategy eager dist/garage.tar.gz[all,dev]
+
+RUN cd benchmarks && python3.5 setup.py sdist && \
+    cp dist/*.tar.gz dist/benchmarks.tar.gz && \
+    pip install --upgrade --upgrade-strategy eager dist/benchmarks.tar.gz
+
+# Nvidia target #######################################################################################################
```
ryanjulian, line 189: If you're going to use fences, please make them 80 chars wide.

.dockerignore
gitanshu: COPY obeys `.dockerignore`, but we need the `.git` folder since there are some tests which need it, the ones that test the archival of source code inside the experiment directory. So I don't have `.git` in `.dockerignore` anymore

docker/entrypoint-verify-envs.sh
gitanshu: Yes. I made this file but didn't end up using it anyway. Removing it.

docker/Dockerfile
gitanshu: Okay!

Makefile
```diff
@@ -10,13 +10,33 @@ SHELL := /bin/bash
...
+		--target garage-test-18.04 \
+		--secret id=mjkey,src=$(MJKEY_PATH) \
+		-t rlworkgroup/garage-test-18.04 \
+		${BUILD_ARGS} \
+		.
+
+test:  ## Run the garage-test-18.04 docker target
 test: RUN_ARGS = --memory 7500m --memory-swap 7500m
-test: run-headless
+test: CONTAINER_NAME ?= garage-test
```
ryanjulian, line 30: rather than overriding the container name, you should be able to capture it during build (return value) and re-use it during run.

Makefile
```diff
@@ -98,28 +117,38 @@ ci-deploy-docker: assert-travis
 	docker push rlworkgroup/garage-ci
 
 build-ci: TAG ?= rlworkgroup/garage-ci:latest
-build-ci: docker/docker-compose-ci.yml
-	TAG=${TAG} \
-	docker-compose \
-		-f docker/docker-compose-ci.yml \
-		build \
-		${BUILD_ARGS}
+build-ci: docker/Dockerfile
```
ryanjulian, line 120: something like 

```Makefile
build-ci:
    CONTAINER_NAME := $(shell \
    docker build \
        -- progress=plain \
        --build_arg BUILDKIT_INLINE_CACHE=1 \
        --cache-from rlworkgroup/garage-ci:latest \ 
        -f docker/Dockerfile \ 
        --target garage-dev-18.04 \
        -t ${TAG} \
        ${BUILD_ARGS} .
    )
    export CONTAINER_NAME
```

Makefile
```diff
@@ -138,30 +167,33 @@ run-ci:
 
 run-headless: ## Run the Docker container for headless machines
 run-headless: CONTAINER_NAME ?= garage-headless
+run-headless: user ?= $$USER
 run-headless: build-headless
 	docker run \
 		-it \
 		--rm \
-		-v $(DATA_PATH)/$(CONTAINER_NAME):/root/code/garage/data \
+		-v $(DATA_PATH)/$(CONTAINER_NAME):/home/$(user)/code/garage/data \
```
ryanjulian, line 175: you are not guaranteed that this is where garage code (or even home directories) are.

i think you should just put it in $(PWD)/garage_data or something.

docker/Dockerfile.base.16.04
```diff
@@ -51,9 +51,6 @@ RUN \
   apt-get clean && \
   rm -rf /var/lib/apt/lists/*
 
-# Build GLFW because the Ubuntu 16.04 version is too old
```
ryanjulian, line 54: hmm please don't remove these breadcrumbs explaining why we bother building glfw from source

Makefile
gitanshu: The container name is a runtime thing though. The image built during the build phase doesn't have a container name by itself.

Makefile
gitanshu: Not sure I understand this. Do you mean change it to 
`-v $(PWD)/garage_data:/home/$(user)/code/garage/date`

docker/Dockerfile.base.16.04
gitanshu: Done

Makefile
gitanshu: the current makefile sets a default container name, which you can override.
So you can do `make run-headless RUN_CMD="foo"` and it'll make a container called `garage-headless`
and then `CONTAINER_NAME='my-experiment-bar' make run-headless RUN_CMD="bar"` to make another container. And they'll run in parallel.

gitanshu: OTOH, if we don't set any name by default, you can always do `make run-headless ...` and it'll create a new container with a random name, and they all run in parallel. But those names don't have any meaning. And your experiment directory would go in some directory like `$pwd/data/trusty_pangolin`.

gitanshu: Yet another option is we keep the CONTAINER_NAME variable but set default value as empty instead of `garage-headless`, so you can override if you want, but you don't have to.

ryanjulian: perhaps we should remove naming the experiment directories by container name (since experiments and containers are usually 1:1 anyway)

i would advocate a default-empty container name, so that our system usually functions like docker normally would (people don't usually set container names)

ryanjulian:
How far is this from merging? I think this PR has dragged on too long. If it's not possible to do in one PR, we should submit what we have and keep working.

PR #1429, opened 2020-05-20
Mention OS support in README.md

### LINKED ISSUE #1427: Cannot install garage on Windows 10
I am trying to run some of the sample code to get a handle on another project, and I am encountering issues where at least one of the dependencies (ray, in my case) is not compatible with windows. Are there any workarounds or plans to get Windows compatibility in the future?

ryanjulian:
Fixes #1427

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,9 @@ The table below summarizes the algorithms available in garage.
 | MTTRPO                 | PyTorch, TensorFlow |
 
 ## Supported Tools and Frameworks
-garage supports Python 3.5+.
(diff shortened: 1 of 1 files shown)


maliesa96:
I just installed garage on macOS 10.15.4 (which is the latest version of macOS) and I've had no issues so far. Maybe we can specify that?

ryanjulian:
@maliesa96 There's a field in `scripts/setup_osx.sh` for this. You can send a PR to update it.

PR #1436, opened 2020-05-27
Extract long running mujoco tests in separate job

gitanshu:


ryanjulian:
You may need to update .codecov.yml?

gitanshu:
> You may need to update .codecov.yml?

Oh yeah. Good catch. Did that now.

PR #1444, opened 2020-05-28
Fix old and new distribution synchronization in on-policy algorithms

ahtsan:
In on-policy algorithms, a policy object is cloned to represent the old distribution, to be used
when calculating KL divergence between old and new policies in each optimization epoch.
The cloned policy should have the same parameters in the first place, otherwise it will result in
longer training time and large optimization error in the first epoch. This PR fixes this issue.

This also fixes RL2 to update old policy parameters at the end of each optimization epoch.

### DIFF
--- a/src/garage/tf/algos/_rl2npo.py
+++ b/src/garage/tf/algos/_rl2npo.py
@@ -100,6 +100,7 @@ class RL2NPO(NPO):
                                                    samples_data['returns'],
                                                    samples_data['valids'])
         tabular.record('{}/ExplainedVariance'.format(self.baseline.name), ev)
+        self._old_policy.model.parameters = self.policy.model.parameters
 
     def _get_baseline_prediction(self, samples_data):
         """Get baseline prediction.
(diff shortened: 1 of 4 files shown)


ryanjulian:
Is there a test which could have caught this bug?

ahtsan:
Maybe we can set timeout threshold to catch any unexpected behavior causing training to be slower.

ahtsan:
> It looks like the `RL2NPO.optimize_policy` function could be pretty easily refactored to call the superclass method which already has the `self._old_policy.model.parameters = self.policy.model.parameters` code. But this looks like it works, so I suppose that's an issue for another time.

That's a good catch. Let me explain why they do not share the same code in the first place.

It's a common practice to fit baseline after optimization in regular RL, as in `NPO`, while `RL2NPO` is an exception because it's used in meta-RL, where the task keeps changing in between each epoch. With the face that we only have non-recurrent baseline, as a result baseline has to be fitted before optimization, otherwise it will be fed with samples from a different task, which is wrong. 

Therefore we did not incorporate this particular logic into `NPO` to provide options for fitting baseline before / after, as it is not a common use case. But yes, it is error-prone and causes this bug. Hopefully we will have recurrent baseline comes into place so we don't have to fit baseline before optimization, then we can make them share the same code.

ryanjulian:
@mergifyio rebase

PR #1451, opened 2020-05-29
Update installation instructions for zsh users

haydenshively:
Because zsh likes to be weird, arguments like `.[all,dev]` get treated as commands and simply don't work. When installing, users must instead put quotes around the arguments (i.e. `pip install garage"[mujoco,dm_control]"`). I've updated the docs to include this information, since new Macs come with zsh installed by default.

### DIFF
--- a/docs/user/installation.rst
+++ b/docs/user/installation.rst
@@ -93,7 +93,7 @@ pip:
 
 .. code-block:: bash
 
-    pip install garage[mujoco,dm_control]
+    pip install 'garage[mujoco,dm_control]'
 
 Extra Steps for Garage Developers
 ---------------------------------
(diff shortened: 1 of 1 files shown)


docs/user/installation.rst
```diff
@@ -93,7 +93,10 @@ pip:
 
 .. code-block:: bash
 
+    # bash
```
ryanjulian, line 96: `pip install 'garage[mujoco,dm_control]'` works in both bash and zsh, so it's preferable to just use that

ryanjulian: same comment for all of the below

ryanjulian:
When you're ready, if you have 2 approving reviews (1 from maintainers), pass the CI, and pass codecov, you can apply the label `ready-to-merge` and the auto-merger will rebase your PR when it's ready and submit it for you.

haydenshively:
Anybody know why the build timed out? I only changed the install doc

ryanjulian:
The conda/pipenv job is broken on master by an upstream dependency. I will just force-push this for you.

PR #1456, opened 2020-05-29
Refactor GaussianMLPBaseline to use tfp

ahtsan:
Benchmark

launcher script:
https://github.com/rlworkgroup/garage/blob/gaussian_mlp_baseline_tfp_test/tests/benchmarks/benchmark_gaussian_mlp_policy.py

results: (Garage = baseline without tfp, Garage2 = baseline with tfp)
HalfCheetah-v2_benchmark_mean
Hopper-v2_benchmark_mean
InvertedDoublePendulum-v2_benchmark_mean
InvertedPendulum-v2_benchmark_mean
Reacher-v2_benchmark_mean
Swimmer-v2_benchmark_mean
Walker2d-v2_benchmark_mean


src/garage/tf/models/gaussian_mlp_model.py
```diff
@@ -74,6 +74,8 @@ class GaussianMLPModel(Model):
                exponential transformation
             - softplus: the std will be computed as log(1+exp(x))
         layer_normalization (bool): Bool for using layer normalization or not.
+        return_distribution (bool): If True, return distribution object.
```
ryanjulian, line 77: this is really nasty. is this temporary?

ryanjulian: if they really need to be supported in parallel, why not just return everything and use what you need?

ahtsan: Yes, this is temporary and the option that returns raw tensors will no longer be needed when regressors accept full trajectories of shape [N, T, *dims]. I don't want to return everything which would break other usage of GaussianMLPModel.

ryanjulian: I'm sorry, but conditionals controlling returns is so bad I don't want it anywhere in the codebase, even temporarily. I think you need to find another way.

Remember that this code might end up in a release which is supported for 8-12 months.

You can:
* Refactor all the callsites to expect a third return value
* Keep two classes around until you can remove the other variant
* Access the distribution through a property on the network

ryanjulian: Option 1 doesn't seem that bad:
```sh
rjulian@nyquist:~/code/garage$ rg GaussianMLPModel -l | wc -l
16
```

ahtsan: Option 1 sounds good. Will do

src/garage/tf/models/gaussian_mlp_model.py
```diff
@@ -159,6 +168,8 @@ def _build(self, state_input, name=None):
 
         Returns:
             tfp.distributions.MultivariateNormalDiag: Distribution.
+            tf.tensor: Mean.
+            tf.Tensor: log_stds.
```
naeioi, line 172: Log of standard variance.

src/garage/tf/regressors/gaussian_mlp_regressor.py
```diff
@@ -130,60 +130,45 @@ def __init__(self,
             std_parameterization='exp',
             layer_normalization=layer_normalization)
 
+        # model for old distribution, used when trusted region is on
+        self._old_model = self.model.clone(name='model_for_old_dist')
```
naeioi, line 134: `model.clone()` does not copy the parameter. This is problematic later when calculating the KL divergence using the old model.

src/garage/tf/regressors/gaussian_mlp_regressor_model.py
```diff
@@ -178,15 +179,67 @@ def _build(self, state_input, name=None):
...
+                vanilla_dist, means_var, log_stds_var, x_mean_var, x_std_var,
+                y_mean_var, y_std_var)
+
+    def clone(self, name):
+        """Return a clone of the model.
 
-        return (means_var, log_stds_var, std_param, normalized_mean,
-                normalized_log_std, x_mean_var, x_std_var, y_mean_var,
-                y_std_var, dist)
+        It only copies the configuration of the primitive,
```
naeioi, line 209: I think we also need to copy the parameters.

src/garage/tf/regressors/gaussian_mlp_regressor.py
ryanjulian: There are cases where you don't want the params (e.g. because you want to create dueling Q-functions), but I think the params should be copied by default.

I think maybe you want a signature like `model.clone(reinitialize=False)`

src/garage/tf/models/gaussian_mlp_model.py
ahtsan: covariance?

src/garage/tf/regressors/gaussian_mlp_regressor.py
ahtsan: Good idea. I will add this in another PR.

src/garage/tf/regressors/gaussian_mlp_regressor_model.py
ahtsan: Will do in another PR.

src/garage/tf/regressors/gaussian_mlp_regressor.py
ahtsan: However in most cases, parameters are not initialized when `model.clone()` is called, before the model is not built yet. We may also have to enforce `mode.clone()` can only be called after the model is built.

Issue #1391, opened 2020-05-06
Remove BatchPolopt


ryanjulian:
This is really a bundle of helper functions masquerading as a superclass. It makes the inheritance tree unnecessarily-deep and obscures implementation details of the algorithms which use it.

We should restrict our use of inheritance to the most obvious/terse relationships (e.g. VPG<--PPO), and not use it to group helper functions shared by a set of classes.

It can be replaced by moving its helper functions (namely, process_samples) to a function package, and moving the main algo loop into each algorithm.

naeioi:
This issue has actually two independent parts - the Numpy BatchPolopt and the TensorFlow BatchPolopt. The numpy one is inherited by both numpy and torch trees. If we were to remove BatchPolopt, we better break it to two issues.

This idea of replacing BatchPolopt with a package of helpers looks tricky to me. There're many accesses to the `self` object. A helper will be much more verbose. We need more discussion on the plan for this issue.

ryanjulian:
`BatchPolopt` is basically
1. Constructor which saves some fields
1. Short `train` loop (and `train_once` in the case of tf)
1. `process_samples`

(1) and (2) can be copied into the subclasses*, increasing clarity a lot. 
(3) can be a helper function.

* - `discount` and `max_path_length` should probably be put into RLAlgorithm

There's no reason one issue can't be addressed by a series of PRs. Feel free to edit it with a checklist, if you want.

ryanjulian:
Let's not overestimate the magnitude here:

```sh
rjulian@nyquist:~/code/garage$ rg BatchPolopt src -l
src/garage/np/algos/__init__.py
src/garage/np/algos/cem.py
src/garage/np/algos/cma_es.py
src/garage/np/algos/batch_polopt.py
src/garage/torch/algos/vpg.py
src/garage/tf/algos/__init__.py
src/garage/tf/algos/reps.py
src/garage/tf/algos/npo.py
src/garage/tf/algos/batch_polopt.py
```

The two versions of this class have 5 usages total.

AiRuiChen:
> `BatchPolopt` is basically
> 
> 1. Constructor which saves some fields
> 2. Short `train` loop (and `train_once` in the case of tf)
> 3. `process_samples`
> 
> (1) and (2) can be copied into the subclasses*, increasing clarity a lot.
> (3) can be a helper function.
> 
> * * `discount` and `max_path_length` should probably be put into RLAlgorithm
> 
[...]

PR #1486, opened 2020-05-31
Refactor np BatchPolopt

AiRuiChen:
Partial fix of [#1391 ](https://github.com/rlworkgroup/garage/issues/1391).

Algorithms now inherit directly from RLAlgorithms instead of np BatchPolopt. Once all refactoring is done BatchPolopt will be deleted.

Plans for future commits:
1. Refactor tf BatchPolopt
2. Fix reference to BatchPolopt from tests/fixtures. 
- Do I need to do this? If so, should `InstrumentedBatchPolopt` be removed or refactored as well?
3. Delete BatchPolopt


src/garage/np/algos/cem.py
```diff
@@ -62,6 +70,11 @@ def __init__(self,
         self.n_params = None
         self._initialize()
 
+        if policy.vectorized:
```
krzentner, line 73: Do we even have any non-vectorized policies? I think we should just always use `BatchSampler`.

src/garage/torch/algos/vpg.py
```diff
@@ -97,12 +103,12 @@ def __init__(
...
-                         baseline=value_function,
-                         discount=discount,
-                         max_path_length=max_path_length,
-                         n_samples=num_train_per_epoch)
+        super().__init__(env_spec=env_spec, discount=discount)
+
+        if policy.vectorized:
+            self.sampler_cls = OnPolicyVectorizedSampler
+        else:
+            self.sampler_cls = BatchSampler
```
krzentner, line 111: Same here.

src/garage/np/algos/rl_algorithm.py
```diff
@@ -9,10 +9,17 @@ class RLAlgorithm(abc.ABC):
...
         initialize a sampler.
 
+    Args:
+        env_spec (garage.envs.EnvSpec): Environment specification.
+        discount (float): Discount.
     """
 
     # pylint: disable=too-few-public-methods
 
+    def __init__(self, env_spec, discount):
```
krzentner, line 19: I think it would be best for `RLAlgorithm` to not have any constructor. Just copy the two lines into every class that needs them.

src/garage/np/algos/rl_algorithm.py
```diff
@@ -9,10 +9,17 @@ class RLAlgorithm(abc.ABC):
         If the field sampler_cls exists, it will be by LocalRunner.setup to
         initialize a sampler.
 
+    Args:
+        env_spec (garage.envs.EnvSpec): Environment specification.
+        discount (float): Discount.
```
krzentner, line 14: By adding two more fields, you've basically doubled the size of one of the most commonly used APIs in garage. Keeping this API small is very important, so please remove these fields.

src/garage/misc/process_samples_utils.py
```diff
@@ -0,0 +1,80 @@
...
+    ])
+
+    env_infos = [path['env_infos'] for path in paths]
+    env_infos = tensor_utils.stack_tensor_dict_list(
+        [tensor_utils.pad_tensor_dict(p, max_path_length) for p in env_infos])
+
+    valids = [np.ones_like(path['returns']) for path in paths]
+    valids = tensor_utils.pad_tensor_n(valids, max_path_length)
+
+    undiscounted_returns = log_performance(
```
ryanjulian, line 69: can you move this into the algo as well?

tests/fixtures/algos/dummy_algo.py
```diff
@@ -20,8 +20,8 @@ def __init__(self, env_spec, policy, baseline):
                          max_path_length=1,
                          n_samples=10)
 
-    def train_once(self, itr, paths):
+    def train_once(self, itr, paths):  # noqa: D102
```
ryanjulian, line 23: please provide documentation (it's required for test fixtures)

src/garage/misc/process_samples_utils.py
```diff
@@ -0,0 +1,80 @@
...
+        baseline(garage.np.baselines.Baseline): Baseline for GAE
+            (Generalized Advantage Estimation).
+        discount (float): Environment reward discount.
+        env_spec (garage.envs.EnvSpec): Environment specification.
+        episode_reward_mean(collections.deque): Deque to record episode
+            reward mean metric
+
+    Returns:
+        dict: Processed sample data, with key
+            * average_return: (float)
```
ryanjulian, line 27: i think it has many more fields than this...

AiRuiChen: My bad. I will update this.

src/garage/misc/process_samples_utils.py
AiRuiChen: > can you move this into the algo as well?

Sure!

ryanjulian:
@krzentner I requested those two be put in RLAlgorithm. 

I think that env_spec is an interesting discussion (how much do we want to spread akro into the API?), but every RL process has to have a discount factor. One could argue this is truly a property of the environment, but it's not part of any environment interfaces I know of.

src/garage/misc/process_samples_utils.py
AiRuiChen: > can you move this into the algo as well?

Sorry, I am actually a little confused here. Do you mean moving line 66-69 into the algo? Or move everything starting from line 69? It seems to me that everything up to line 67 is "preprocessing", and codes starting from line 69 are actually processing the samples. 

Thanks in advance for the clarification!

krzentner:
> but every RL process has to have a discount factor.

That's not really true? See "Average Return Reinforcement Learning" by Sridhar Mahadevan.
Having said that, these methods are equivalent to just setting the discount to 1.0, I suppose.

Anyways, I'm mostly against the *principle* of expanding the `RLAlgorithm` API, since it's the closest thing we have to a god-class (aside from `LocalRunner`), and we have many implementations of it. We've worked really hard to decrease the amount of global state that has to be there, and I haven't seen any argument for why adding it back would make things easier.

To begin with, we should try really hard to avoid passing `RLAlgorithm` instances into other functions, which is the only way this field would get read. When we've passed `RLAlgorithm` instances into other objects, it has lead to hidden global state, bizarre setup difficulties, performance problems, and resume bugs. I made this mistake once when I designed the `MetaEvaluator`, and had to patch several of those problems at the last minute. I don't think we should set ourselves up to make the same mistake again.

> env_spec

It might be reasonable to have this field? It's actually required by `LocalRunner` if you want it to control the sampler (unlike `.discount`, which is never used by any API). It's not needed if you never call `runner.obtain_samples`, which is checked by [this unit test](https://github.com/rlworkgroup/garage/blob/master/tests/garage/experiment/test_local_runner.py#L55). `EnvSpec` is also a more complex API to explain to users, which makes me a little nervous.

Having said that, I think we also shouldn't have the runner control the sampler, since it's involved basically duplicating the entire sampler interface in the runner, and some algorithms still reach into private `LocalRunner` fields to access additional parts of the sampler directly.

If we do add these fields, I think we should make a super-class called `Algorithm` that doesn't have these fields, and ensure that `LocalRunner` only uses that interface. We should also update the `Implementing New Algorithms` docs section to mention these new mandatory fields, and correct the code samples so that they are correct and still run.

ryanjulian:
I agree with the course-of-action about creating `Algorithm` when the need arises, e.g. if/when we add the ILAlgorithm (immitation learning) class.

When I said to have `discount` and possibly `env_spec` properties on RLAlgorithm, I meant those mostly as a service to algo implementers to remind them that `self.discount` and `self.env_spec` should exist. I agree that passing around instances of RLAlgorithm (especially as `self`) is probably a sign that something is wrong, because an RLAlgorithm is basically a `main()` function.

I think it is helpful to ask ourselves "why does this class even exist, if its own purpose is to implement `train()`? why not just have a `train()` protocol?" and honestly, I don't have a great answer. Currently, it's because there are parts of an RL algorithm which imply resource allocation, and classes are a handy mnemonic for "allocate some resources (constructor)" then "use those resources (train)". We could remove the "allocate some resources" part and force the `train()` function to either allocate the resources or accept some context, but that makes the save/resume story harder (as opposed to just telling an algorithm to implement the pickle protocol to get access to save/resume). For the time being, I think it's helpful to allow the user to model their training process as a class with a very small interface, because it gives them a default container for statefulness, despite not being very Jupyter-friendly.

Anyway, this all quite some high-flying discussion for a simple refactor. Since it's clear that this PR will make the codebase better, I suggest we just:
* Leave RLAlgorithm untouched
* Push `env_spec` and `discount` into the child classes (by copy-paste, not using inheritance)
* Figure out the API bits later

src/garage/np/algos/rl_algorithm.py
ryanjulian: I agree

src/garage/np/algos/cem.py
ryanjulian: All policies are now vectorized, AFAICT, so this check can be removed

ryanjulian:
@AiRuiChen i think you can follow KR's guidance and just not update the interfaces on RLAlgorithm

I think InstrumentedBatchPolopt can be totally removed without fear.

krzentner:
> We could remove the "allocate some resources" part and force the `train()` function to either allocate the resources or accept some context, but that makes the save/resume story harder (as opposed to just telling an algorithm to implement the pickle protocol to get access to save/resume).

Honestly not completely sure what you have in mind here, but that sounds like a discussion for another time.

> Anyway, this all quite some high-flying discussion for a simple refactor. Since it's clear that this PR will make the codebase better, I suggest we just:
>     - Leave RLAlgorithm untouched
>     - Push `env_spec` and `discount` into the child classes (by copy-paste, not using inheritance)
>     - Figure out the API bits later

I'm definitely happy with this approach. In the worst case we'll have copied a few lines of code extra.

And yeah, `InstrumentedBatchPolopt` has already taken up too much of our time. Just delete it and any tests that depend on it.

src/garage/misc/process_samples_utils.py
ryanjulian: I am suggesting that the call to `log_performance` is really critical to all algorithms functioning correctly (it is actually calculating the performance of the algorithm) and should be moved into the algos.

src/garage/misc/process_samples_utils.py
```diff
@@ -0,0 +1,80 @@
+"""Utiliy functions for batch processing samples."""
```
ryanjulian, line 1: please don't create anything new `garage.misc` -- we are trying to delete this package.

I think you can put it in `src/garage/np/_functions.py` and then export it to `garage.np.process_samples_for_policy_gradient()` by editing `src/garage/np/__init__.py`

src/garage/misc/process_samples_utils.py
```diff
@@ -0,0 +1,80 @@
...
+from dowel import tabular
+import numpy as np
+
+from garage import log_performance, TrajectoryBatch
+from garage.misc import tensor_utils
+
+
+def preprocess_samples_for_policy_gradient(itr, paths, max_path_length,
```
ryanjulian, line 9: This is also used by CEM and CMA-ES, neither of which are policy gradient algos (my bad). 

Let's be more descriptive, it is basically "paths_to_tensors"

src/garage/misc/process_samples_utils.py
```diff
@@ -0,0 +1,80 @@
...
+    Returns:
+        dict: Processed sample data, with key
+            * average_return: (float)
+
+    """
+    baselines = []
+    returns = []
+
+    if hasattr(baseline, 'predict_n'):
+        all_path_baselines = baseline.predict_n(paths)
```
ryanjulian, line 34: please calculate the baseline values in the algorithm, then pass them to this function,

algorithm:
```py
baseline_predictions = baseline.predict_n(paths)

paths = process_samples_for_policy_gradient(..., baseline_predictions)
```

src/garage/misc/process_samples_utils.py
```diff
@@ -0,0 +1,80 @@
...
+
+    Args:
+        itr (int): Iteration number.
+        paths (list[dict]): A list of collected paths.
+        max_path_length (int): Maximum length of a single rollout.
+        baseline(garage.np.baselines.Baseline): Baseline for GAE
+            (Generalized Advantage Estimation).
+        discount (float): Environment reward discount.
+        env_spec (garage.envs.EnvSpec): Environment specification.
+        episode_reward_mean(collections.deque): Deque to record episode
```
ryanjulian, line 22: please move everything involving `episode_reward_mean` into the algorithm (both maintaining the deque and the logging)

src/garage/np/algos/cem.py
```diff
@@ -43,15 +47,19 @@ def __init__(self,
                  best_frac=0.05,
                  extra_std=1.,
                  extra_decay_time=100):
-        super().__init__(env_spec, policy, baseline, discount, max_path_length,
-                         n_samples)
-
-        self.n_samples = n_samples
         self.best_frac = best_frac
+        self.baseline = baseline
```
ryanjulian, line 51: to the extend possible (i know there is still some code which assumes public attrs), please prefix all of these with `_` to make them private (e.g. `self._baseline = baseline`)

src/garage/np/algos/cma_es.py
```diff
@@ -34,14 +38,16 @@ def __init__(self,
...
-                         policy=policy,
-                         baseline=baseline,
-                         discount=discount,
-                         max_path_length=max_path_length,
-                         n_samples=n_samples)
-
+        self.env_spec = env_spec
+        self.discount = discount
         self.sigma0 = sigma0
+        self.n_samples = n_samples
```
ryanjulian, line 44: please make as many of these possible private

src/garage/torch/algos/vpg.py
```diff
@@ -79,13 +80,20 @@ def __init__(
         self._use_softplus_entropy = use_softplus_entropy
         self._stop_entropy_gradient = stop_entropy_gradient
         self._entropy_method = entropy_method
+        self.env_spec = env_spec
```
ryanjulian, line 83: please make as many of these possible private

src/garage/np/_functions.py
```diff
@@ -0,0 +1,74 @@
+"""Utility functions for NumPy-based Reinforcement learning algorithms."""
+import numpy as np
+
+from garage.misc import tensor_utils
+
+
+def process_samples_for_policy_gradient(paths_to_tensors, max_path_length,
```
ryanjulian, line 7: can you name this function `paths_to_tensors`?

src/garage/np/_functions.py
```diff
@@ -0,0 +1,74 @@
...
+                                        baseline_predictions, discount):
+    """Return processed sample data based on the collected paths_to_tensors.
+
+    Args:
+        paths_to_tensors (list[dict]): A list of collected paths_to_tensors.
```
ryanjulian, line 12: sorry, my previous comment was unclear. this arg can be named `paths` (that's the convention around the codebase) and the function can be named `paths_to_tensors`

PR #1492, opened 2020-06-02
Delete np/batch_polopt.py

ryanjulian:
I think this was forgotten in #1486


ryanjulian:
@AiRuiChen we usually try to keep the codebase really tight (e.g. no unused code) because people often download from git, so imagine every day is a release.

PR #1495, opened 2020-06-02
Remove garage.misc.prog_bar_counter and pyprind

ryanjulian:
These are easily replaced by click.progressbar


ryanjulian:
@mergifyio rebase

Issue #1062, opened 2019-11-21
Use new optimizer_args interface for tf and np algorithms


ryanjulian:
See #1030 for details.

ryanjulian:
Closed by #1496

PR #1496, opened 2020-06-02
New optimizer_args tf interface + test optimizer type

irisliucy:
This PR addresses issue #1062
1) removes `optimizer_args` pattern from tf algorithm constructors in `tf/algos` , `tf/optimizers` and `tf/regressors`, 
2) make members of algorithm class _private if haven't done so, 
3) test tf optimizer interface

The tf optimizer interface `garage/tf/algos/_utils.py` follows the compatibility of TensorFlow Core v2.2.0 and is similar to torch's optimizer interface but accepts an optimizer type only. The algorithm internally constructs an optimizer using a new utility function `_make_optimizer`.


benchmarks/src/garage_benchmarks/experiments/algos/ppo_garage_tf.py
```diff
@@ -68,7 +68,7 @@ def ppo_garage_tf(ctxt, env_id, seed):
                       optimizer_args=dict(
                           batch_size=32,
                           max_epochs=10,
-                          tf_optimizer_args=dict(learning_rate=3e-4),
+                          optimizer_args=dict(learning_rate=3e-4),
```
ryanjulian, line 71: hmm this is confusing -- optimizer_args has an optimizer_args inside of it?

perhaps we should just make these inner `optimizer_args` part of the outer `optimizer_args` and then just pass them directly.

src/garage/tf/algos/__init__.py
```diff
@@ -1,4 +1,6 @@
 """Tensorflow implementation of reinforcement learning algorithms."""
+from garage.tf.algos._utils import _Default  # noqa: F401
```
ryanjulian, line 2: you solve these by adding the symbols in question to `__all__` below

src/garage/tf/algos/_utils.py
```diff
@@ -0,0 +1,33 @@
+"""Utility functions used by tensorflow algorithms."""
```
ryanjulian, line 1: please put this in a module `_functions.py` instead (we are trying to move to this pattern across the codebase

src/garage/tf/algos/_utils.py
```diff
@@ -0,0 +1,33 @@
...
+    Args:
+        optimizer_type (type): Type of optimizer.
+        kwargs (dict): Other keyword arguments to initialize optimizer. This
+            is not used when `optimizer_type` is tuple.
+
+    Returns:
+        tf.compat.v1.train.Optimizer: Constructed optimizer.
+    """
+    opt_args = {}
+    for name, arg in kwargs.items():
```
ryanjulian, line 28: You can use a dict comprehension

opt_args = `{k: v.val if isinstance(v, _Default) else arg for k, v in kwargs.items()}`

src/garage/tf/algos/dqn.py
```diff
@@ -69,15 +70,15 @@ def __init__(self,
                  reward_scale=1.,
                  smooth_return=True,
                  name='DQN'):
-        self.qf_lr = qf_lr
-        self.qf_optimizer = qf_optimizer
-        self.name = name
-        self.target_network_update_freq = target_network_update_freq
-        self.grad_norm_clipping = grad_norm_clipping
-        self.double_q = double_q
+        # self.qf_lr = qf_lr
```
ryanjulian, line 73: remove this?

src/garage/tf/regressors/categorical_mlp_regressor.py
```diff
@@ -1,8 +1,10 @@
 """A regressor based on MLP with Normalized Inputs."""
+# pylint: disable=no-value-for-parameter
```
ryanjulian, line 2: please only disable lint rules directly at the affected line

src/garage/tf/algos/_utils.py
irisliucy: Do you mean the module in `src/garage/_functions.py`? or simply create a `_functions.py` under `src/garage/tf`?

The interface is very similar to the one in torch except it accepts optimizer type only, not tuple. And if we decide to use `src/garage/_functions.py`, I can move both tf, torch `make_optimizer` functions under the interface

ryanjulian: If your function needs tf, put it in `src/garage/tf/_functions.py` and export to `garage.tf`

If your function is not specific to tf (and can be share by torch), please move the shared implementation to `src/garage/_function.py` and export to `garage` package.

benchmarks/src/garage_benchmarks/experiments/algos/ppo_garage_tf.py
irisliucy: Cleaned out nested `optimizer_args `

src/garage/tf/algos/_utils.py
irisliucy: I refactored the tf and pytorch shared implementation and moved them to `src/garage/_function.py` for export.

src/garage/tf/algos/dqn.py
irisliucy: removed

src/garage/tf/algos/_utils.py
ryanjulian: Did you upload your new version yet? I don't see it here

src/garage/torch/algos/__init__.py
```diff
@@ -1,9 +1,9 @@
 """PyTorch algorithms."""
-from garage.torch.algos._utils import _Default  # noqa: F401
-from garage.torch.algos._utils import compute_advantages  # noqa: F401
-from garage.torch.algos._utils import filter_valids  # noqa: F401
-from garage.torch.algos._utils import make_optimizer  # noqa: F401
-from garage.torch.algos._utils import pad_to_last  # noqa: F401
+from garage.torch.algos._utils import _Default
+from garage.torch.algos._utils import compute_advantages
+from garage.torch.algos._utils import filter_valids
+from garage.torch.algos._utils import make_optimizer
```
ryanjulian, line 5: I think it's better if we put all of these in `garage.torch` rather than `garage.torch.algos`

ryanjulian:
Awesome work!

src/garage/tf/algos/_utils.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def __init__(self, val):
+        self.val = val
+
+
+def make_optimizer(optimizer_type, **kwargs):
+    """Create an optimizer for tensorflow algos.
+
+    Args:
+        optimizer_type (type): Type of optimizer.
```
naeioi, line 20: `optimizer_type` can be either a `type`, or `tuple[type, dict]`. This selection can be expressed by `Union[type, tuple[type, dict]]`

src/garage/tf/algos/_utils.py
```diff
@@ -0,0 +1,31 @@
...
+        kwargs (dict): Other keyword arguments to initialize optimizer. This
+            is not used when `optimizer_type` is tuple.
```
naeioi, line 22: You didn't implement the case when `optimizer_type` is a tuple.

Here is the use case - `optimizer_type` will be specified by the user, and `kwargs` is provided by the algorithm as a set of sane defaults. User can override default by passing `optimizer_type` as an optimizer type and a parameter dict. If two parameters of the same name appear in both `optimizer_type` and `kwargs`, take the one in `optimizer_type`.

src/garage/tf/algos/_utils.py
```diff
@@ -0,0 +1,31 @@
+"""Utility functions used by tensorflow algorithms."""
+
+
+class _Default:  # pylint: disable=too-few-public-methods
```
naeioi, line 4: The semantic of `_Default` is **_allowing_ a parameter to be missing and use default instead**.

In `make_optimizer()`, if a parameter in kwargs is not wrapped with `_Default`, and it also appears in `optimizer_type`, a `ValueError` should be raised.

See here https://github.com/rlworkgroup/garage/blob/3d9c1b27a39addb21c8df4fd6489c19eee5c6aeb/src/garage/torch/algos/_utils.py#L43

src/garage/tf/optimizers/first_order_optimizer.py
```diff
@@ -30,8 +31,8 @@ class FirstOrderOptimizer:
     """
 
     def __init__(self,
-                 tf_optimizer_cls=None,
-                 tf_optimizer_args=None,
+                 optimizer=None,
+                 optimizer_args=None,
```
naeioi, line 35: Can you replace `optimizer_args` with `learning_rate`? This is the only parameter we want to tune. This also aligns with your change at `benchmarks/src/garage_benchmarks/experiments/algos/ppo_garage_tf.py`

irisliucy: I merged default `optimizer_args` with `tf_optimizer_args` to new `optimizer_args`. Hence, now `optimizer_args` will accept more than `learning_rate` but also other parameters.

irisliucy: For easier usage, I replaced `optimizer_args` with `learning_rates` across the codebase

benchmarks/src/garage_benchmarks/experiments/algos/ppo_garage_tf.py
```diff
@@ -52,7 +52,7 @@ def ppo_garage_tf(ctxt, env_id, seed):
                 optimizer_args=dict(
                     batch_size=32,
                     max_epochs=10,
-                    tf_optimizer_args=dict(learning_rate=3e-4),
+                    learning_rates=dict(learning_rate=3e-4),
```
ryanjulian, line 55: i apologize for not being clear -- there are optimizer_args other than learning rate.

What i meant was that we would not have the dict at all, i.e. don't support passing arbitrary kwargs down to the tf.Optimizer instance.

e.g.

```
optimizer_args=dict(
    batch_size=32,
    max_epochs=10,
    learning_rate=3e-4,
)
```

irisliucy:
I push the unified tf and PyTorch optimizer interface not long ago. There are pickling tests i.e. `garage/tf/algos/test_dqn.py` failing and I'm looking into the reason why. Any inputs will be appreciated. Meanwhile, I'll address the rest of the comments.

src/garage/torch/algos/__init__.py
irisliucy: Moved `garage.torch.algos._utils` to `garage.torch._utils` and export `compute_advantages`, `filter_valids`, and `pad_to_last` to `garage.torch`.

ryanjulian:
Be careful with merge commits! They really mess with our CI (which only assumes rebases).

Only use `git pull --rebase` and never use the "Update branch" button on GitHub.

irisliucy:
I resolved the python pickle issues on the tf algorithm when using `make_optimizer()`. After looking into the issue, it may be cause when python cannot pickle lambda expressions in tf.optimizer. I found that by wrapping the `make_optimizer()` under named functions/ tf.name_scope, as well as using the local variables such as `optimizer` works. It'll look like this:

```
with tf.name_scope(‘init_optimizer’):
            optimizer = make_optimizer(self._optimizer, **self._optimizer_args)
            train_optimizer = optimizer.minimize(
                    …
             )

```

ryanjulian:
Hmm -- can `cloudpickle` not pickle these lambdas? Perhaps the tests are still using `pickle`, but we switched to `cloudpickle` for the implementations?

src/garage/_functions.py
```diff
@@ -8,6 +8,61 @@
...
+
+
+def make_optimizer(optimizer_type, module=None, **kwargs):
+    """Create an optimizer for pyTorch & tensorflow algos.
+
+    Args:
+        optimizer_type (Union[type, tuple[type, dict]]): Type of optimizer.
+            This can be an optimizer type such as 'torch.optim.Adam' or a
+            tuple of type and dictionary, where dictionary contains arguments
+            to initialize the optimizer e.g. (torch.optim.Adam, {'lr' = 1e-3})
```
naeioi, line 30: `lr: 1e-3` is the correct python grammar.

src/garage/tf/algos/td3.py
```diff
@@ -106,6 +107,16 @@ def __init__(
         self._action_loss = None
 
         self.target_qf2 = qf2.clone('target_qf2')
+        self._policy_optimizer = policy_optimizer
+        self._qf_optimizer = qf_optimizer
+        self._policy_lr = policy_lr
+        self._qf_lr = qf_lr
+        # self._policy_optimizer = make_optimizer(policy_optimizer,
```
naeioi, line 114: Comments should be deleted

PR #1497, opened 2020-06-02
Refactor tf BatchPolopt

AiRuiChen:
Complete fix of [#1391 ](https://github.com/rlworkgroup/garage/issues/1391).

TF version of BatchPolopt is now removed and inheritance relationship now looks cleaner.


src/garage/tf/_functions.py
```diff
@@ -0,0 +1,99 @@
...
+        max_path_length (int): Maximum length of a single rollout.
+        baseline_predictions(numpy.ndarray): : Predicted value of GAE
+            (Generalized Advantage Estimation) Baseline.
+        discount (float): Environment reward discount.
+        gae_lambda (float): Lambda used for generalized advantage
+            estimation.
+
+    Returns:
+        dict: Processed sample data, with key
+            * observations: (numpy.ndarray)
```
ryanjulian, line 23: does this return numpy arrays or tf.tensors? If it returns numpy arrays, can't you just use `garage.np.paths_to_tensors`?

src/garage/tf/algos/npo.py
```diff
@@ -156,13 +170,113 @@ def init_opt(self):
...
+            paths (list[dict]): A list of collected paths.
+
+        Returns:
+            numpy.float64: Average return.
+
+        """
+        # -- Stage: Calculate baseline
+        if self._flatten_input:
+            paths = [
+                dict(
```
ryanjulian, line 208: couldn't this be way shorter with a single helper variable?

src/garage/tf/algos/npo.py
```diff
@@ -156,13 +170,113 @@ def init_opt(self):
...
+
+        Args:
+            paths (list[dict]): A list of collected paths.
+
+        """
+        logger.log('Logging diagnostics...')
+        self.policy.log_diagnostics(paths)
+        self._baseline.log_diagnostics(paths)
+
+    # pylint: disable=unused-argument
```
AiRuiChen, line 272: Parameter `itr` is not used in the original `BatchPolopt` implementation, yet files in the tf/ directory still follows this common signature.  See this [workaround attempt](https://github.com/rlworkgroup/garage/blob/master/src/garage/np/algos/off_policy_rl_algorithm.py#L138). 

We can create a separate PR to refactor this.

ryanjulian, line 272: is there a reason to not just remove itr?

PR #1504, opened 2020-06-02
Remove tf batchpolopt

AiRuiChen:
Ditto PR of [#1497](https://github.com/rlworkgroup/garage/pull/1497). I was not able to reopen the last PR (github says no new commits found in that branch). Please look at this PR instead. Sorry for the inconvenience.

Complete fix of [Issue#1391](https://github.com/rlworkgroup/garage/issues/1391)

TF version of BatchPolopt is now removed and inheritance relationship now looks cleaner.


src/garage/tf/_functions.py
```diff
@@ -0,0 +1,99 @@
...
+    total_steps = 0
+
+    for idx, path in enumerate(paths):
+        total_steps += len(path['rewards'])
+        path_baselines = np.append(baseline_predictions[idx], 0)
+        deltas = (path['rewards'] + discount * path_baselines[1:] -
+                  path_baselines[:-1])
+        path['advantages'] = np_tensor_utils.discount_cumsum(
+            deltas, discount * gae_lambda)
+        path['deltas'] = deltas
```
AiRuiChen, line 45: @ryanjulian This helper functions differs from the np version here. So I didn't reuse the np version.

src/garage/tf/algos/npo.py
```diff
@@ -156,13 +170,101 @@ def init_opt(self):
...
+
+        Args:
+            paths (list[dict]): A list of collected paths.
+
+        """
+        logger.log('Logging diagnostics...')
+        self.policy.log_diagnostics(paths)
+        self._baseline.log_diagnostics(paths)
+
+    # pylint: disable=unused-argument
```
AiRuiChen, line 260: @ryanjulian 

Parameter itr is not used in the original BatchPolopt implementation, yet files in the tf/ directory still follows this common signature. See this [workaround attempt](https://github.com/rlworkgroup/garage/blob/master/src/garage/np/algos/off_policy_rl_algorithm.py#L138). 

I think refactoring this function signature isn't necessary a part of "remove BatchPolopt". Maybe we can create a separate PR to refactor this?

ryanjulian: Yeah -- just leave it for now and get this PR merged.

src/garage/tf/_functions.py
ryanjulian: alright no problem. I think they'll be unified when we move to TrajectoryBatch soon

src/garage/tf/algos/npo.py
```diff
@@ -156,13 +170,101 @@ def init_opt(self):
...
     def optimize_policy(self, itr, samples_data):
         """Optimize policy.
 
         Args:
             itr (int): Iteration number.
             samples_data (dict): Processed sample data.
-                See process_samples() for details.
+                See _functions.paths_to_tensors() for details.
```
ryanjulian, line 267: We use the _-prefixed package names to help users import functions from their canonical package paths rather than their full paths. In this case, I think that means you should point them to `garage.tf.paths_to_tensors()`

src/garage/tf/algos/npo.py
```diff
@@ -404,7 +506,7 @@ def _fit_baseline_with_data(self, samples_data):
 
         Args:
             samples_data (dict): Processed sample data.
-                See process_samples() for details.
+                See _functions.paths_to_tensors() for details.
```
ryanjulian, line 509: Use the canonical path

src/garage/tf/algos/npo.py
```diff
@@ -433,14 +535,14 @@ def _fit_baseline_with_data(self, samples_data):
...
-        self.baseline.fit(paths)
+        self._baseline.fit(paths)
 
     def _policy_opt_input_values(self, samples_data):
         """Map rollout samples to the policy optimizer inputs.
 
         Args:
             samples_data (dict): Processed sample data.
-                See process_samples() for details.
+                See _functions.paths_to_tensors() for details.
```
ryanjulian, line 545: Use the canonical path

src/garage/tf/algos/reps.py
```diff
@@ -1,18 +1,24 @@
...
 from dowel import logger, tabular
 import numpy as np
 import scipy.optimize
 import tensorflow as tf
 
-from garage.tf.algos.batch_polopt import BatchPolopt
+from garage import log_performance, TrajectoryBatch
+from garage.np.algos import RLAlgorithm
+from garage.sampler import OnPolicyVectorizedSampler
+from garage.tf import _functions
```
ryanjulian, line 12: you can use `from garage.tf import paths_to_tensors`

src/garage/tf/algos/npo.py
```diff
@@ -1,11 +1,16 @@
...
 from dowel import logger, tabular
 import numpy as np
 import tensorflow as tf
 
+from garage import log_performance, TrajectoryBatch
 from garage.misc import tensor_utils as np_tensor_utils
-from garage.tf.algos.batch_polopt import BatchPolopt
+from garage.np.algos import RLAlgorithm
+from garage.sampler import OnPolicyVectorizedSampler
+from garage.tf import _functions
```
ryanjulian, line 13: you should use `from garage.tf. import paths_to_tensors` to serve as a good example

src/garage/tf/algos/npo.py
```diff
@@ -156,13 +170,100 @@ def init_opt(self):
...
+
+        Args:
+            paths (list[dict]): A list of collected paths.
+
+        """
+        logger.log('Logging diagnostics...')
+        self.policy.log_diagnostics(paths)
+        self._baseline.log_diagnostics(paths)
+
+    # pylint: disable=unused-argument
```
naeioi, line 259: If a parameter is not used, prefer `del param` over pylint exception rule.

ryanjulian:
@AiRuiChen if you resolve the conflict and label this `ready-to-merge` then Mergify will take care of the rest :)

src/garage/tf/algos/npo.py
AiRuiChen: @ryanjulian @naeioi 

I am working on a small PR to fix this. Found that the `itr` param of `optimize_policy()` is only used [here](https://github.com/rlworkgroup/garage/blob/master/src/garage/tf/algos/td3.py#L324). So simply removing `itr` from `optimize_policy()`will not work. 

In this context, should I just do `del itr`? Although `itr` is unused now, passing it to`optimize_policy()` can support potential future functionalities (e.g. logging) anyways.

What do you think?

ryanjulian: optimize_policy is not a public interface (i.e. these methods should actually be `_optimize_policy`, so it's free to be different between each algorithm).

ryanjulian: in general you should commit for the code you have, not the code you think you might have in the future.

i recommend removing it.

AiRuiChen: `TD3` inherits from `DDPG`, and it seems unreasonable to have `TD3` overriding its parent's `optimize_policy()` function with a different signature. Both `optimize_policy()` are doing the same thing, but with different implementations and different inputs. So I also don't see how I can give them different names to avoid this inheritance conflict.

Is there a reason why `TD3` inherits from `DDPG`? If they don't have any special relationship (such as in an algorithm sense), I suggest:

1.  make `TD3` inherit from `OffPolicyRLAlgorithm` directly. 
2.  delete the abstract  `optimize_policy()` from `OffPolicyRLAlgorithm. As you suggested, this function should be private and sub-classes should not share the same interface.

Any thoughts?

ryanjulian: @maliesa96 and @krzentner  are working on removing OffPolicyRLAlgorithm entirely (for the exact same reasons we removed BatchPolopt), so I wouldn't spend too much time fretting over it.

If there a reason you can't remove itr from `NPO.optimize_policy()`? Is there an abstract `opimitze_policy` in NPO's inheritance tree?

AiRuiChen: NPO doesn't have this issue. The conflict only exists in TD3 and DDPG inheritance tree.

So I guess I will make TD3 inherit from `OffPolicyRLAlgorithm`?

ryanjulian: That is probably harder than it looks (because of some bad design choices in OffPolicyRLAlgorithm, previously mentioned). 

What's the goal of your new PR? Eliminating `itr` from `optimizer_policy` everywhere?

AiRuiChen: Ah I see. Yes the goal is to remove unused parameters from `optimize_policy()`. Most of them don't use `itr`, and [this one](https://github.com/rlworkgroup/garage/blob/master/src/garage/tf/algos/ddpg.py#L323) doesn't use neither `itr` nor `samples_data`.

This is not something urgent, as we can let `del itr` handle it and wait till `OffPolicyRLAlorithm` is refactored.

ryanjulian: Yeah, the story of `optimize_policy` is that someone tried to create an `RLAlgorithm` interface which included it, applied it (using `BatchPolopt` and `OffPolicyRLAlgorithm`), and then realized it doesn't really make sense for many algorithms.

The new `RLAlgorithm` interface only includes `train`. The motivation for this tiny interface, which forces each algorithm to do a lot, is that it's very difficult to define a single, universal control flow for an RL algorithm, and libraries which remove control flow from the user tend to make things difficult to understand.

Removing `BatchPolopt` and `OffPolicyRLAlgorithm` is basically unwinding this attempt. Going forward, anything which looks like `optimize_policy` can become a private function (`_optimize_policy`) inside the algorithms which were affected, only accessible. Whether `itr` appears in the signature is up to the algorithm, because it's a private function.

One of the biggest complaints from users about other RL libraries (why they don't adopt them and keep writing messy one-off code which often has bugs) is that they have to jump across 10 files just to understand what is happening in the algorithm. For the same reason, we're removing algorithm superclasses like BatchPolopt, so you can read the algorithm file and understand everything in an algorithm on "one pane of glass". The only reason we use inheritance on algorithms is for the most straightforward, shortest kind of inheritance, like when a single method gets overloaded in a simple way.

Perhaps a positive change would be to make everything which isn't `train` on all the algorithms underscore-private. This is also not urgent, but in retrospect the current situation tricks people into thinking there are public interfaces there when there are not.

AiRuiChen: I am happy to put these in my PR to make functions except `train()` private. 

However, changing all `optimize_policy` to `_optimize_policy` won't solve the inheritance conflict in `TD3`. Because now `TD3` and `DDPG` both have `_optimize_policy`, but sub-class `TD3` has a different `_optimize_policy` signature.

So we either remove this inheritance relationship between `TD3` and `DDPG`, or give [class-specific](https://stackoverflow.com/questions/20261517/inheritance-of-private-and-protected-methods-in-python) names to `_optimize_policy` (e.g. `_TD3__optimize_policy`).

Please see my attempt to remove inheritance relationship: [#1516](https://github.com/rlworkgroup/garage/pull/1516)

Sorry this simple PR turns into this huge discussion. Really appreciate your help!!!

PR #1510, opened 2020-06-03
Ensure TF and torch are optional

ryanjulian:
This PR ensures that all packages in garage (except garage.tf and
garage.torch, respectively) can be imported without installing either
of the numerical frameworks.

### DIFF
--- a/src/garage/sampler/__init__.py
+++ b/src/garage/sampler/__init__.py
@@ -19,7 +19,7 @@ from garage.sampler.worker import Worker
 from garage.sampler.worker_factory import WorkerFactory
 
 __all__ = [
-    'BatchSampler', 'Sampler', 'ISSampler', 'singleton_pool', 'LocalSampler',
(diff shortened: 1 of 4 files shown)


src/garage/sampler/__init__.py
```diff
@@ -18,6 +18,10 @@
 from garage.sampler.worker import Worker
 from garage.sampler.worker_factory import WorkerFactory
 
+# ISSampler requires TensorFlow
+if 'tensorflow' in sys.modules:
+    from garage.sampler.is_sampler import ISSampler
+
```
haydenshively, line 24: If ISSampler requires TensorFlow, shouldn't `__all__ = [...]` be adjusted accordingly? Right now `'ISSampler'` is in `__all__` regardless of whether TensorFlow is available

ryanjulian: Great catch, updating that now.

ryanjulian:
@krzentner added your suggestions as #1511

ryanjulian:
@merifyio rebase

Issue #1501, opened 2020-06-02
Refactor away garage.tf.experiment.LocalTFRunner


ryanjulian:
This can instead be handled using conditional behavior in garage.experiment.LocalRunner

PR #1513, opened 2020-06-04
Move LocalTFRunner to garage.experiment

ryanjulian:
This will help our effort to make NN libraries optional.

Partially addresses #1501

### DIFF
--- a/src/garage/experiment/local_tf_runner.py
+++ b/src/garage/experiment/local_tf_runner.py
@@ -5,11 +5,17 @@ pipelines data between sampler and algorithm during training.
 """
 from dowel import logger
 import psutil
-import tensorflow as tf
 
 from garage.experiment import LocalRunner
 from garage.sampler import DefaultWorker
-from garage.tf.samplers import TFWorkerClassWrapper
+
+tf = False
+TFWorkerClassWrapper = False
+try:
+    import tensorflow as tf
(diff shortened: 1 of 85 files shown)


ryanjulian:
You're right -- KR and I are used to the Google convention (e.g. HttpServer instead of HTTPServer).

ryanjulian:
@mergifyio rebase

ryanjulian:
@mergifyio rebase

ryanjulian:
@mergifyio rebase

src/garage/experiment/local_tf_runner.py
```diff
@@ -5,11 +5,17 @@
...
-import tensorflow as tf
 
 from garage.experiment import LocalRunner
 from garage.sampler import DefaultWorker
-from garage.tf.samplers import TFWorkerClassWrapper
+
+tf = False
+TFWorkerClassWrapper = False
+try:
+    import tensorflow as tf
```
krzentner, line 15: To be honest, I'm really not a fan of this approach, since it means that tf still gets imported if it can, even if I don't want it to be imported (i.e. I have tf installed, but only want to use `garage.torch`).

It'll do for now though.

ryanjulian: it is really hard to do this while keeping the static analyzers happy and not accidentally creating huge messes in the import graph (I tried).

My impression of the 90% use case here is that users who aren't interested in TF won't install it (and eventually there will be `garage[tf]` and `garage[torch]`, which is what this PR is working towards).

Issue #766, opened 2019-07-03
Change CNN API to use tuples for defining kernels


nish21:
Currently,
1) Garage currently only supports filters of same dimensions Eg: A filter should be of shape 
`(n, n)`. It should support filters of shape `(m, n)` as well. 

2) CNN currently accepts two inputs `filter_dims` and `num_filters` to specify the dimensions and number of filters. So, to create a CNN with `n1, n2` filters of shape `(s1, s1)` 
and `(s2, s2)` respectively, we need to pass it as `filter_dims = (s1, s2)` and 
`num_filters = (n1, n2)`. This should be updated to accept a single parameter of type `List(Tuple(Tuple, Int))`. So that it can be created as 
`filters = [((s1, s1), n1), ((s2, s3), n2)]`

PR #1515, opened 2020-06-04
Change CNN API to use tuples for defining kernels

yeukfu:
Fix [issue 766](https://github.com/rlworkgroup/garage/issues/766). Change CNN API to use tuples for defining kernels. 

1. The new API will have only one parameter `filters` instead of `filter_dims` and `num_filters`.
The type of `filters` is `tuple(int, tuple(tuple(int)))`. For example, `((3, (3, 5)), (32, (3, 3)))` means there are two convolutional layers. The filter for the first layer have 3 channels and its shape is (3 x 5), while the filter for the second layer have 32 channels and its shape is (3 x 3).

2. Fix problems in test_cnn.py. In specific, change the computing of the output size with 'SAME' padding.

3. Fix max_pooling function in tests/helpers.py.


benchmarks/src/garage_benchmarks/experiments/baselines/gaussian_cnn_baseline.py
```diff
@@ -10,8 +10,7 @@
 from garage.tf.policies import CategoricalCNNPolicy
 
 params = {
-    'conv_filters': (32, 64, 64),
-    'conv_filter_sizes': (5, 3, 2),
+    'conv_filters': (((5, 5), 32), ((3, 3), 64), ((2, 2), 64)),
```
ahtsan, line 13: I love this change. This is more readable.

src/garage/tf/models/categorical_cnn_model.py
```diff
@@ -20,13 +20,11 @@ class CategoricalCNNModel(Model):
...
         output_dim (int): Dimension of the network output.
-        filter_dims (tuple[int]): Dimension of the filters. For example,
-            (3, 5) means there are two convolutional layers. The filter
-            for first layer is of dimension (3 x 3) and the second one is of
-            dimension (5 x 5).
-        num_filters (tuple[int]): Number of filters. For example, (3, 32) means
-            there are two convolutional layers. The filter for the first layer
-            has 3 channels and the second one with 32 channels.
+        filters (tuple(tuple(tuple(int), int))): Dimension and number of
+            filters. For example, (((3, 5), 3), ((3, 3), 32)) means there are
```
ryanjulian, line 24: I think that it would be easier to read if this order were reversed, especially when formatted as a block, e.g.

```
filters = (
    (3, (3, 5)),   # 3 x (3, 5)
    (32, (3, 3)),  # 32 x (3, 3)
)
```

src/garage/tf/regressors/gaussian_cnn_regressor_model.py
```diff
@@ -5,21 +5,21 @@
 from garage.tf.models import GaussianCNNModel
 
 
+# pylint: disable=missing-param-doc, missing-type-doc
+# pylint: disable=missing-return-doc, missing-return-type-doc
+# pylint: disable=differing-param-doc, differing-type-doc
```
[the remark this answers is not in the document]
yeukfu: I will fix it

src/garage/tf/models/categorical_cnn_model.py
yeukfu: That makes sense. Will do.

src/garage/tf/regressors/gaussian_cnn_regressor_model.py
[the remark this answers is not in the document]
yeukfu: This class is inherited from GaussianCNNModel. There are some parameters not explictly seen here. It is more appropriete to remove these from the param doc or add pylint-disable?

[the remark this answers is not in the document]
ahtsan: Please add these parameters, for example see https://github.com/rlworkgroup/garage/blob/master/src/garage/tf/regressors/gaussian_mlp_regressor_model.py

tests/helpers.py
```diff
@@ -235,24 +237,32 @@ def _construct_image_vector(_input, batch, w, h, filter_width, filter_height,
...
+                    col = j * pool_shape
                     results[b][i][j][k] = np.max(
-                        _input[b, col:col + pool_shape, row:row +  # noqa: W504
+                        _input[b, row:row + pool_shape, col:col +  # noqa: W504
                                pool_shape, k])
 
     return results
 
 
+# pylint: disable=missing-raises-doc
```
ahtsan, line 265: Why need this?

tests/helpers.py
```diff
@@ -10,6 +10,8 @@
 from tests.quirks import KNOWN_GYM_RENDER_NOT_IMPLEMENTED
 
 
+# pylint: disable=missing-param-doc, missing-type-doc
+# pylint: disable=missing-return-doc, missing-return-type-doc
```
ahtsan, line 14: Please add docstring for this function. We usually will not disable pylint unless necessary. You can explore the codebase and look for examples.

src/garage/tf/regressors/gaussian_cnn_regressor_model.py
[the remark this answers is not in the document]
yeukfu: got it

tests/helpers.py
yeukfu: I have added docstring for most functions. However, there are several functions I need more time to understand its parameters and returns for example recurrent_step_lstm, recurrent_step_gru.

ryanjulian: @yeukfu @ahtsan should be able to help you fill those out.

yeukfu: great

benchmarks/src/garage_benchmarks/experiments/baselines/gaussian_cnn_baseline.py
```diff
@@ -10,8 +10,7 @@
 from garage.tf.policies import CategoricalCNNPolicy
 
 params = {
-    'conv_filters': (32, 64, 64),
-    'conv_filter_sizes': (5, 3, 2),
+    'conv_filters': ((32, (5, 5)), (64, (3, 3)), (64, (2, 2))),
```
ryanjulian, line 13: it would be nice to format these as below in all examples, to demonstrate to users better what the argument means

```
(
    (32, (5, 5)), 
    (64, (3, 3)), 
    (64, (2, 2)),
)  # yapf: disable
```

benchmarks/src/garage_benchmarks/experiments/policies/categorical_cnn_policy.py
```diff
@@ -10,8 +10,7 @@
 from garage.tf.policies import CategoricalCNNPolicy
 
 hyper_params = {
-    'conv_filters': (32, 64, 64),
-    'conv_filter_sizes': (5, 3, 2),
+    'conv_filters': ((32, (5, 5)), (64, (3, 3)), (64, (2, 2))),
```
ryanjulian, line 13: formatting

examples/tf/dqn_pong.py
```diff
@@ -66,8 +66,8 @@ def dqn_pong(ctxt=None, seed=1, buffer_size=int(5e4), max_path_length=None):
         replay_buffer = PathBuffer(capacity_in_transitions=buffer_size)
 
         qf = DiscreteCNNQFunction(env_spec=env.spec,
-                                  filter_dims=(8, 4, 3),
-                                  num_filters=(32, 64, 64),
+                                  filters=((32, (8, 8)), (64, (4, 4)),
```
ryanjulian, line 69: formatting

examples/tf/ppo_memorize_digits.py
```diff
@@ -34,20 +34,20 @@ def ppo_memorize_digits(ctxt=None, seed=1, batch_size=4000):
         env = GarageEnv(normalize(gym.make('MemorizeDigits-v0')),
                         is_image=True)
         policy = CategoricalCNNPolicy(env_spec=env.spec,
-                                      num_filters=(32, 64, 64),
-                                      filter_dims=(5, 3, 2),
+                                      filters=((32, (5, 5)), (64, (3, 3)),
```
ryanjulian, line 37: formatting

src/garage/tf/models/gaussian_cnn_model.py
```diff
@@ -54,13 +52,11 @@ class GaussianCNNModel(Model):
...
         std_share_network (bool): Boolean for whether mean and std share
             the same network.
-        std_filter_dims(tuple[int]): Dimension of the filters. For example,
-            (3, 5) means there are two convolutional layers. The filter
-            for first layer is of dimension (3 x 3) and the second one is of
-            dimension (5 x 5).
-        std_num_filters(tuple[int]): Number of filters. For example, (3, 32)
-            means there are two convolutional layers. The filter for the first
-            layer has 3 channels and the second one with 32 channels.
+        std_filters (tuple(int, tuple(tuple(int)))): Number and dimension of
```
ryanjulian, line 55: the type uses square brackets and capitals for the types, e.g. 

`Tuple[Tuple[int, Tuple[int, int]], ...]`

More importantly, i think you may have gotten the type spec slightly wrong.

Don't worry, these will be easier to catch when we add a type checker.


see https://docs.python.org/3/library/typing.html for a reference

yeukfu: Got it. Thanks

PR #1520, opened 2020-06-05
Policy build automatically in constructor and __setstate__

ahtsan:
Currently, policies have to be built by the owner explicitly. However, this can be done as policies have information about the input placeholder(s), and this will make it much easier to use policies. As a result, policies will be ready to be used right after instantiation, and owners, e.g. algorithms, do not have to build them.

### DIFF
--- a/src/garage/tf/algos/npo.py
+++ b/src/garage/tf/algos/npo.py
@@ -339,6 +339,16 @@ class NPO(RLAlgorithm):
                 policy_state_info_vars[k] for k in self.policy.state_info_keys
             ]
 
+        augmented_obs_var = obs_var
+        for k in self.policy.state_info_keys:
+            extra_state_var = policy_state_info_vars[k]
+            extra_state_var = tf.cast(extra_state_var, tf.float32)
+            augmented_obs_var = tf.concat([augmented_obs_var, extra_state_var],
+                                          -1)
+
+        self.policy.build(augmented_obs_var, name='policy')
+        self._old_policy.build(augmented_obs_var, name='policy')
+
         policy_loss_inputs = graph_inputs(
             'PolicyLossInputs',
             action_var=action_var,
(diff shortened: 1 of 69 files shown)


ryanjulian:
What happens if my policy is inside another policy?

e.g. consider `policy = A(b=B())` . If `B()` needs some tensors from A to be built, doesn't this fail?

I'd like to see a hierarchical policy model (e.g. task embeddings) added before we merge this.

ahtsan:
> What happens if my policy is inside another policy?
> 
> e.g. consider `policy = A(b=B())` . If `B()` needs some tensors from A to be built, doesn't this fail?
> 
> I'd like to see a hierarchical policy model (e.g. task embeddings) added before we merge this.

Does this happen in task embeddings? @naeioi

ahtsan:
There is no a strong reason why it should be merged. If it makes composition harder, I don't see a reason why we should merge this in the first place. Let's come back to this when we design a better `garage.Model` later.

ryanjulian:
I think KR's point is that if it make the user experience better today, we should merge it.

Any improvements to garage.Model later are theoretical.

ahtsan:
> I think KR's point is that if it make the user experience better today, we should merge it.
> 
> Any improvements to garage.Model later are theoretical.
> 
> So long as all of the TF primitives can still be built after construction, I don't actually have a problem with this PR.
> 
> In general, I think that the algorithms should hold the placeholders associated with feeding forward the loss function(s). It's fine for the primitives to have different placeholders for feedforwards associated with numpy operations (e.g. get_action)

That's a good division of use cases. I will incorporate that into this PR.

ahtsan:
@ryanjulian It's ready for review

src/garage/tf/algos/npo.py
```diff
@@ -338,18 +339,15 @@ def _build_inputs(self):
...
         for k in self.policy.state_info_keys:
             extra_state_var = policy_state_info_vars[k]
             extra_state_var = tf.cast(extra_state_var, tf.float32)
             augmented_obs_var = tf.concat([augmented_obs_var, extra_state_var],
                                           -1)
 
-        self.policy.build(augmented_obs_var)
-        self._old_policy.build(augmented_obs_var)
-        self._old_policy.model.parameters = self.policy.model.parameters
+        self.policy.build(augmented_obs_var, name='policy')
```
ryanjulian, line 349: wouldn't it be more canonical to capture the network here (as the return value) rather than always using `self.policy.networks['foo'])`? 

perhaps we should make `Model.networks` private

ahtsan: Agree, we should make policies return the network when built, and cache it. Will fix that.

ahtsan:
> Should we also make garage.tf.Model.networks private? That would also alert people that accessing networks isn't really correct usage.

Definitely

src/garage/tf/policies/gaussian_mlp_policy.py
```diff
@@ -142,21 +142,39 @@ def __init__(self,
...
-        """Build model.
+        """Build policy.
 
         Args:
-          state_input (tf.Tensor): State input.
-          name (str): Name of the model, which is also the name scope.
+            state_input (tf.Tensor) : State input.
+            name (str): Name of the policy, which is also the name scope.
+
+        Returns:
```
naeioi, line 170: The docstring of returns is wrong - `build()` returns a namedtuple, not a tuple.

ahtsan: ah, probably same for all the policies since they all now returned a `network` namedtuple. I'll fix them all together in another PR.

Issue #1141, opened 2020-01-21
Torch policies should not use multiple inheritance


krzentner:
All of the existing Torch policies are `torch.nn.Module`, because they inherit from `garage.torch.modules.MLPModule`. `torch.policies.Policy` is also a `torch.nn.Module`, but it cannot call `super().__init__()`, because multiple inheritance causes that to usually (but not always) call `MLPModule.__init__()`.

PR #1524, opened 2020-06-08
Fix abstract replay buffer interface

maliesa96:
Leaving this open for us to discuss the improved abstract replay buffer interface. I've kept the initial API as simple as possible. As is, `PathBuffer` can be easily modified to subclass `ReplayBuffer`, with the caveat that if an algorithm calls `ReplayBuffer.add_transitions()`, future calls to `sample_path` would no longer be valid, which I think makes this a little hacky. On the other hand, we could scrap `add_transitions` altogether.

### DIFF
--- a/src/garage/replay_buffer/replay_buffer.py
+++ b/src/garage/replay_buffer/replay_buffer.py
@@ -12,173 +12,73 @@ algorithms.
 import abc
 from abc import abstractmethod
 
-import numpy as np
+from typing import Sequence, Any, Union
+import queue
 
 
-class ReplayBuffer(metaclass=abc.ABCMeta):
-    """Abstract class for Replay Buffer.
+class ReplayBuffer(abc.ABC):
 
-    Args:
-        env_spec (garage.envs.EnvSpec): Environment specification.
-        size_in_transitions (int): total size of transitions in the buffer
-        time_horizon (int): time horizon of rollout.
-
-    """
-
-    def __init__(self, env_spec, size_in_transitions, time_horizon):
-        del env_spec
-        self._current_size = 0
-        self._current_ptr = 0
-        self._n_transitions_stored = 0
-        self._time_horizon = time_horizon
-        self._size_in_transitions = size_in_transitions
-        self._size = size_in_transitions // time_horizon
-        self._initialized_buffer = False
-        self._buffer = {}
-        self._episode_buffer = {}
-
-    def store_episode(self):
-        """Add an episode to the buffer."""
-        episode_buffer = self._convert_episode_to_batch_major()
-        rollout_batch_size = len(episode_buffer['observation'])
-        idx = self._get_storage_idx(rollout_batch_size)
-
-        for key in self._buffer:
-            self._buffer[key][idx] = episode_buffer[key]
-        self._n_transitions_stored = min(
-            self._size_in_transitions, self._n_transitions_stored +
-            self._time_horizon * rollout_batch_size)
-
-    @abstractmethod
-    def sample(self, batch_size):
-        """Sample a transition of batch_size.
-
-        Args:
-            batch_size(int): The number of transitions to be sampled.
-
+    @abc.abstractmethod
+    def insert(self, samples: Union[Any, Sequence[Any]]
+               ) -> Union[Any, Sequence[str]]:
+        """Insert samples into buffer.
+            Returns a UUID that idenitifies each sample in the data store.
         """
-        raise NotImplementedError
 
-    def add_transition(self, **kwargs):
-        """Add one transition into the replay buffer.
+    @abc.abstractmethod
+    def sample(self, batch_size: int) -> Union[Sequence[Any], Any]:
+        """Get samples from buffer."""
 
-        Args:
-            kwargs (dict(str, [numpy.ndarray])): Dictionary that holds
-                the transitions.
 
-        """
-        transition = {k: [v] for k, v in kwargs.items()}
-        self.add_transitions(**transition)
+class PathBuffer(ReplayBuffer):
 
-    def add_transitions(self, **kwargs):
-        """Add multiple transitions into the replay buffer.
+    def __init__(data_store, evic_policy_cls=FIFOEvictionPolicy):
+        self.evic_policy = evic_policy_cls(data_store)
+        self.data_store = data_store
 
-        A transition contains one or multiple entries, e.g.
-        observation, action, reward, terminal and next_observation.
-        The same entry of all the transitions are stacked, e.g.
-        {'observation': [obs1, obs2, obs3]} where obs1 is one
-        numpy.ndarray observation from the environment.
+    def insert(self, samples: TrajectoryBatch) -> Sequence[int]:
+        return self.evic_policy.insert(samples.to_list())
 
-        Args:
-            kwargs (dict(str, [numpy.ndarray])): Dictionary that holds
-                the transitions.
+    def sample(self, n: int) -> TrajectoryBatch:
+        return TrajectoryBatch.from_list(self.data_store.sample(n))
 
-        """
-        if not self._initialized_buffer:
-            self._initialize_buffer(**kwargs)
 
-        for key, value in kwargs.items():
-            self._episode_buffer[key].append(value)
+class DataStore(abc.ABC):
 
-        if len(self._episode_buffer['observation']) == self._time_horizon:
-            self.store_episode()
-            for key in self._episode_buffer:
-                self._episode_buffer[key].clear()
+    @abc.abstractmethod
+    def insert(self, samples) -> Sequence[str]:
+        """Insert samples into datastore. Returns keys/UUIDs for each sample."""
 
-    def _initialize_buffer(self, **kwargs):
-        for key, value in kwargs.items():
-            self._episode_buffer[key] = list()
-            values = np.array(value)
-            self._buffer[key] = np.zeros(
-                [self._size, self._time_horizon, *values.shape[1:]],
-                dtype=values.dtype)
-        self._initialized_buffer = True
+    @abc.abstractmethod
+    def remove(self, key: str):
+        """Remove samples"""
 
-    def _get_storage_idx(self, size_increment=1):
-        """Get the storage index for the episode to add into the buffer.
-
-        Args:
-            size_increment(int): The number of storage indeces that new
-                transitions will be placed in.
-
-        Returns:
-            numpy.ndarray: The indeces to store size_incremente transitions at.
-
-        """
-        if self._current_size + size_increment <= self._size:
-            idx = np.arange(self._current_size,
-                            self._current_size + size_increment)
-        elif self._current_size < self._size:
-            overflow = size_increment - (self._size - self._current_size)
-            idx_a = np.arange(self._current_size, self._size)
-            idx_b = np.arange(0, overflow)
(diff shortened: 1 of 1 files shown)


src/garage/replay_buffer/replay_buffer.py
```diff
@@ -12,173 +12,75 @@
...
+        Raises:
+            NotYetImplementedError: If the replay buffer does not support
+                path sampling.
         """
-        transition = {k: [v] for k, v in kwargs.items()}
-        self.add_transitions(**transition)
-
-    def add_transitions(self, **kwargs):
-        """Add multiple transitions into the replay buffer.
+        raise NotImplementedError
```
ryanjulian, line 54: ABCs don't need `NotImplementedError` or `pass` -- just leave them blank

src/garage/replay_buffer/replay_buffer.py
```diff
@@ -12,173 +12,75 @@
...
+            batch_size (int): The number of transitions to be sampled.
 
+        Returns:
+            SampleBatch: batch_size transitions sampled.
         """
-        raise NotImplementedError
 
-    def add_transition(self, **kwargs):
-        """Add one transition into the replay buffer.
+    def sample_path(self):
```
ryanjulian, line 44: mark with `@abc.abstractmethod`

src/garage/replay_buffer/replay_buffer.py
```diff
@@ -12,173 +12,75 @@
 import abc
 from abc import abstractmethod
```
ryanjulian, line 13: please use only package-level imports for non-garage packages (i.e. use this as `@abc.abstractmethod`)

src/garage/replay_buffer/replay_buffer.py
```diff
@@ -12,173 +12,75 @@
...
 
 class ReplayBuffer(metaclass=abc.ABCMeta):
     """Abstract class for Replay Buffer.
 
     Args:
         env_spec (garage.envs.EnvSpec): Environment specification.
-        size_in_transitions (int): total size of transitions in the buffer
-        time_horizon (int): time horizon of rollout.
-
+        capacity_in_transitions (int): capacity of the replay buffer,
```
ryanjulian, line 21: we should keep this terminology consistent in the interfaces

do we have Transitions, TimeSteps, Steps, or Samples? 

I slightly prefer TimeSteps and I think we already have that in dtypes, so this would be `capacity_in_time_steps`. And `SampleBatch` should actually be `TimeStepBatch`

If we thinks that's too wordy, we should look at the entire termnology family.

src/garage/replay_buffer/replay_buffer.py
```diff
@@ -12,173 +12,75 @@
...
+            measured in transitions.
     """
 
-    def __init__(self, env_spec, size_in_transitions, time_horizon):
-        del env_spec
+    def __init__(self, env_spec, capacity_in_transitions):
```
ryanjulian, line 25: I think that interfaces should be constructor-less and have no default implementations. If there's an obvious typical implementation that can be a simple abstract class inherited from the interface.

For instance, there's no reason a replay buffer doesn't need to know the capacity up front, in which case forcing this constructor args is just a useless burden.

src/garage/replay_buffer/replay_buffer.py
```diff
@@ -12,173 +12,75 @@
...
-        for key in self._buffer:
-            self._buffer[key][idx] = episode_buffer[key]
-        self._n_transitions_stored = min(
-            self._size_in_transitions, self._n_transitions_stored +
-            self._time_horizon * rollout_batch_size)
 
     @abstractmethod
-    def sample(self, batch_size):
-        """Sample a transition of batch_size.
+    def sample_transitions(self, batch_size):
```
ryanjulian, line 34: `sample_time_step_batch` or simply `sample_batch`

src/garage/replay_buffer/replay_buffer.py
```diff
@@ -12,173 +12,75 @@
...
-        """Add multiple transitions into the replay buffer.
+        raise NotImplementedError
 
-        A transition contains one or multiple entries, e.g.
-        observation, action, reward, terminal and next_observation.
-        The same entry of all the transitions are stacked, e.g.
-        {'observation': [obs1, obs2, obs3]} where obs1 is one
-        numpy.ndarray observation from the environment.
+    @abstractmethod
+    def add_path(self, path):
```
ryanjulian, line 57: `add_trajectory` since the dtype is `TrajectoryBatch`

consider defining a dtype `Trajectory` if the singleton case appears often.

src/garage/replay_buffer/replay_buffer.py
```diff
@@ -12,173 +12,75 @@
...
-    def _convert_episode_to_batch_major(self):
-        """Convert the shape of episode_buffer.
-
-        episode_buffer: {time_horizon, algo.rollout_batch_size, flat_dim}.
-        buffer: {size, time_horizon, flat_dim}.
-
-        Returns:
-            dict: Transitions that have been formated to fit properly in this
-                replay buffer.
+            transitions (dict[str, np.ndarray] or TrajectoryBatch): transitions to be added to the buffer.
```
ryanjulian, line 69: time_steps

src/garage/replay_buffer/replay_buffer.py
```diff
@@ -12,173 +12,75 @@
...
-            values = np.array(value)
-            self._buffer[key] = np.zeros(
-                [self._size, self._time_horizon, *values.shape[1:]],
-                dtype=values.dtype)
-        self._initialized_buffer = True
-
-    def _get_storage_idx(self, size_increment=1):
-        """Get the storage index for the episode to add into the buffer.
+    @abstractmethod
+    def add_transitions(self, transitions):
```
ryanjulian, line 65: `add_time_step_batch`

ryanjulian, line 65: i think this should accept only `TimeStepBatch` or possible `Union[TimeStep, TimeStepBatch]`

src/garage/replay_buffer/replay_buffer.py
ryanjulian: if `Trajectory` dtype exists, this could accept `Union[Trajectory, TrajectoryBatch]`

src/garage/replay_buffer/replay_buffer.py
```diff
@@ -12,173 +12,75 @@
...
     def full(self):
-        """Whether the buffer is full.
-
-        Returns:
-            bool: True of the buffer has reachd its maximum size.
-                False otherwise.
-
-        """
-        return self._current_size == self._size
+        """bool: True if the buffer has reached its maximum capacity."""
```
ryanjulian, line 75: Hmm does this *really* need to be in the interface? what is it used for?

src/garage/replay_buffer/replay_buffer.py
```diff
@@ -12,173 +12,75 @@
 import abc
 from abc import abstractmethod
 
-import numpy as np
-
 
 class ReplayBuffer(metaclass=abc.ABCMeta):
```
ryanjulian, line 16: Immediately I see an issue here where there is 2 interfaces: a `TimeStep`-oriented one and `Trajectory`-oriented one.

Do we want to:
* Force all replay buffers to implement both semantics?
* Standardize on adding/sampling time steps?
* Standardize on adding/sampling trajectories? Note this would make single-stepping sampling a pain.
* Choose one interface as the universal case, define a more-specialized interface for the less-universal case? In this case, I think `TimeStep`-oriented is the universal case and `Trajectory`-oriented is the specific case.
* Define two different interfaces, entirely.

src/garage/replay_buffer/replay_buffer.py
krzentner: My personal opinion on this topic is very simple: `__init__` should not be part of an API, ever. When I'm constructing something, I know exactly what type it is, so the API doesn't mean anything. The only reason to have `__init__` be part of an API is that you want to have default method implementations in terms of private fields, which I think is an anti-pattern.

We have some places where we've done this already in garage, but it's resulted in difficult to explain or fix bugs (see torch policies / issue #1141). The problem is that implementing multiple interfaces on the same type (which is very common in python, once you start looking) then requires reasoning about multiple inheritance, which will always be somewhat broken.

ryanjulian:
@krzentner since the goal of this PR is to establish a durable ReplayBuffer API, can you give us some of your thoughts on what you think that should look like?

I agree the data backend API should be as simple as possible. We could go so far as making it agnostic of the stored datatype (doesn't care if it's storing trajectories or samples).

maliesa96:
@ryanjulian Regarding the issue you point out, the current interface assumes we're going with option 1 (force replay buffers to implement both semantics), mainly because it was simpler to get this PR up.
I think the cleaner approach is to have separate APIs for TimeStep and Trajectory based replay buffers. This removes one layer of abstraction, but it makes reading code easier because I'd be able to tell whether an algorithm uses one type of buffer or the other simply based on the calls it makes. It also makes implementations of the buffers more straightforward, they either deal with TrajectoryBatch or TimeSteps, not both. This way we can also keep the simpler PathBuffer implementation as is like @krzentner mentioned.

ryanjulian:
Perhaps we should provide a `ReplayBuffer` interface which is data-agnostic, and further `PathBuffer` and `SampleBuffer` interfaces which specify the data types which are being stored.

```python
class ReplayMemory(abc.ABC):

    @abc.abstractmethod
    def puts(self, samples: Sequence[Any]): -> Sequence[int]

    @abc.abstractmethod
    def gets(self, indices): -> Sequence[Any]

    @abc.abtractemthod
    def evicts(self, indices): --> Sequence[int]

    @property
    @abc.abstractmethod
    def size(self): --> int

    @property
    @abc.abstractmethod
    def max_index(self): --> int

    @property
    @abc.abstractmethod
    def min_index(self): --> int


class ReplayBuffer:

    @abc.abstractmethod
    def puts(self, samples: Sequence[Any]): -> Sequence[int]

    @abc.abstractmethod
    def sample(self, n: int): -> Sequence[Any]
    

class FIFOReplayBuffer(ReplayBuffer):

    def __init__(self, memory: ReplayMemory, capacity_in_samples: int):
        self._capacity = capacity_in_samples
        self._memory = memory

    def puts(self, samples: Sequence[Any]): -> Sequence[int]
        maybe_overflow = self._memory.size + len(samples) - self._capacity
        if maybe_overflow > 0:
            self._memory.evicts(list(range(min_index, min_index + maybe_overflow)))

        return self._memory.puts(samples)

    def sample(self, n: int): -> Sequence[Any]
        indices = random.sample(range(self._memory.min_index, self._memory.max_index), n)
        return self._memory.gets(indices)
    

def FIFOPathBuffer(FIFOReplayBuffer):

    def puts(self, samples: TrajectoryBatch): -> Sequence[int]
        return super().puts(samples.to_list())

    def sample(self, n: int): -> TrajectoryBatch
        trajs = super().sample(n)
        return TrajectoryBatch.from_list(trajs)


def FIFOTimeStepBuffer(FIFOReplayBuffer):

    def puts(self, samples: TimeStepBatch): -> Sequence[int]
        return super().puts(samples.to_list())

    def sample(self, n: int): -> TimeStepBatch
        ts = super().sample(n)
        return TimeStepBatch.from_list(ts)

PR #1527, opened 2020-06-09
Refactor torch util functions

irisliucy:
I combined `garage.torch._utils` and `garage.torch.utils` into `garage.torch._functions` and exported the utility functions to `garage.torch`.


docs/user/experiments.rst
```diff
@@ -215,7 +215,7 @@ In order to enable the GPU for PyTorch, add the following code snippets to the e
 .. code-block:: python
 
   import torch
-  import garage.torch.utils as tu
+  import garage.torch._functions as tu
```
ryanjulian, line 218: these should be access from `garage.torch`

examples/torch/mtsac_metaworld_ml1_pick_place.py
```diff
@@ -20,10 +20,10 @@
 from garage.experiment import deterministic, LocalRunner
 from garage.replay_buffer import PathBuffer
 from garage.sampler import LocalSampler
+import garage.torch._functions as tu
 from garage.torch.algos import MTSAC
 from garage.torch.policies import TanhGaussianMLPPolicy
 from garage.torch.q_functions import ContinuousMLPQFunction
```
ryanjulian, line 26: please import these utils from garage.torch

src/garage/torch/algos/ddpg.py
```diff
@@ -9,7 +9,7 @@
 from garage import _Default, make_optimizer
 from garage.np.algos.off_policy_rl_algorithm import OffPolicyRLAlgorithm
 from garage.replay_buffer import PathBuffer
-import garage.torch.utils as tu
+from garage.torch import dict_np_to_torch, torch_to_np
```
ryanjulian, line 12: yes -- please use all imports like this

src/garage/torch/__init__.py
```diff
@@ -1,6 +1,21 @@
 """PyTorch-backed modules and algorithms."""
-from garage.torch._utils import compute_advantages
-from garage.torch._utils import filter_valids
-from garage.torch._utils import pad_to_last
+from garage.torch._functions import _DEVICE
```
ryanjulian, line 2: don't import _ variables -- those are private

irisliucy: Gotcha, will remove all private variables import.

irisliucy:
@Merifyio rebase

ryanjulian:
remember, rather than babysitting your PR, you can add the label `ready-to-merge` and Mergify will do it for you. this also helps the CI run faster (by not scheduling jobs which don't need to be run)

PR #1528, opened 2020-06-09
Refactor torch policies

### LINKED ISSUE #1141: Torch policies should not use multiple inheritance

krzentner:
This way there's much less code, no broken multiple inheritance, and how
to implement new torch policies is actually explainable to new users.

Fixes #1141

Note that this change is not yet complete. I still need to fix the tests and make sure that the algorithms still work.

### DIFF
--- a/src/garage/torch/policies/deterministic_mlp_policy.py
+++ b/src/garage/torch/policies/deterministic_mlp_policy.py
@@ -9,7 +9,7 @@ from garage.torch.modules import MLPModule
 from garage.torch.policies.policy import Policy
 
 
-class DeterministicMLPPolicy(Policy, MLPModule):
+class DeterministicMLPPolicy(Policy):
     """Implements a deterministic policy network.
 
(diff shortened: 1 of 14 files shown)


src/garage/torch/policies/deterministic_mlp_policy.py
```diff
@@ -25,15 +25,27 @@ def __init__(self, env_spec, name='DeterministicMLPPolicy', **kwargs):
             kwargs : Additional keyword arguments passed to the MLPModule.
 
         """
+        Policy.__init__(self, env_spec, name)
```
ryanjulian, line 28: why not `super()`?

Issue #1057, opened 2019-11-20
Define TrajectoryBatch and SampleBatch data structures


ryanjulian:

naeioi:
Any background on SampleBatch? @ryanjulian @krzentner

ryanjulian:
It's a batch of Sample objects. I'm not sure it needs to exist. 

It is used when you wanted to access the fields a batch of samples collectively. For instance, if you want a tensor (numpy array) of all the sampled actions to feed to a neural network

You can actually already see this happening in PathBuffer, but it's not a well-defined data structure:
https://github.com/rlworkgroup/garage/blob/master/src/garage/replay_buffer/path_buffer.py#L92
https://github.com/rlworkgroup/garage/blob/master/src/garage/torch/algos/ddpg.py#L181
https://github.com/rlworkgroup/garage/blob/master/src/garage/torch/algos/ddpg.py#L223

After we define this, `optimize_policy` in DDPG above can accept `samples_data: garage.SampleBatch` instead of `samples_data: Dict[str, np.ndarray]`

```python
BATCH_SIZE = 1024
sb = replay_buffer.sample(1024)
print(sb.actions)  # -> np.ndarry with shape (1024, *env.action_space.low.shape)
```

ryanjulian:
Basically this would take what's already happening and attach a type to it, so the code is more clear and the correct keys are guaranteed to exist.

It can also define helper functions, e.g. `.to()` in the case of torch, to help reduce the proliferation of helper functions. Most of "utils" and "tensor_utils" modules just provide different views on the same data, or do things like convert data to torch.Tensor, etc.

naeioi:
Can I say that BatchSampler is the data structure of path buffer? It provides utility methods for off-policy algorithms (e.g. sample a subset of data).

As for TrajectoryBatch, it is already in _dtype. What do we need to do with it in this issue?

ryanjulian:
`TrajectoryBatch` is already defined. It needs to replace `paths` throughout the codebase, but that's a different issue (#1110 )

`SampleBatch` is the analogous data structure to `TrajectoryBatch`, but for sample-oriented (e.g. off-policy) algorithms. It defines a dtype for the interface between replay buffers and algorithms.

PathBuffer is just one implementation of a ReplayBuffer, whose API we need to define ( #1468 ), but the API will basically look like `PathBuffer`.

What needs to be done here is for someone to add the `SampleBatch` dtype and tests for it.

[...]

PR #1529, opened 2020-06-09
Add TimeStepBatch

AiRuiChen:
This addresses [Issue #1057 ](https://github.com/rlworkgroup/garage/issues/1057).

This definition of SampleBatch follows the format of TrajectoryBatch, and provides similar helper functions.

Future plans:
1. add other helper functions (looking for suggestions!)
2. use SampleBatch in PathBuffer and change other usages accordingly.

### DIFF
--- a/src/garage/_dtypes.py
+++ b/src/garage/_dtypes.py
@@ -22,6 +22,8 @@ class TrajectoryBatch(
     # pylint: disable=missing-return-doc, missing-return-type-doc, missing-param-doc, missing-type-doc  # noqa: E501
     r"""A tuple representing a batch of whole trajectories.
 
+    Data type for on-policy algorithms.
+
     A :class:`TrajectoryBatch` represents a batch of whole trajectories
     produced when one or more agents interacts with one or more environments.
 
@@ -310,7 +312,7 @@ class TrajectoryBatch(
                     current environment).
                 * rewards (np.ndarray): Array of rewards of shape (T,) (1D
                     array of length timesteps).
-                * dones (np.ndarray): Array of rewards of shape (T,) (1D array
+                * dones (np.ndarray): Array of dones of shape (T,) (1D array
                     of length timesteps).
                 * agent_infos (dict[str, np.ndarray]): Dictionary of stacked,
                     non-flattened `agent_info` arrays.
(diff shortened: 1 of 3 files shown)


src/garage/_dtypes.py
```diff
@@ -559,3 +559,214 @@ def output_space(self):
...
+            current environment).
+        rewards (numpy.ndarray): Array of rewards of shape (batch_size,) (1D
+            array of length batch_size).
+        next_observation (numpy.ndarray): Non-flattened array of next
+            observations. Has shape (batch_size, S^*). next_observations[i] was
+            observed by the agent after taking actions[i].
+        terminals (numpy.ndarray): A boolean numpy array of shape
+            shape (batch_size,) containing the termination signals for all
+            transitions in this batch.
+        batch_size (int): The size of the batch
```
ryanjulian, line 587: this is inferrable from len(terminals) (or anything else) so it doesn't need to be included.

src/garage/_dtypes.py
```diff
@@ -559,3 +559,214 @@ def output_space(self):
 
         """
         return self._output_space
+
+
+class SampleBatch(
```
ryanjulian, line 564: We need to figure out naming here, since this is the batch version of `TimeStep` above. Either this is a `TimeStepBatch` or the dtype above is called `Sample`

I favor `TimeStepBatch` because sample is an overloaded term (as is `Step`).

src/garage/_dtypes.py
```diff
@@ -559,3 +559,214 @@ def output_space(self):
...
+        collections.namedtuple('SampleBatch', [
+            'observations', 'actions', 'rewards', 'next_observations',
+            'terminals', 'batch_size'
+        ])):
+    # pylint: disable=missing-param-doc, missing-type-doc
+    """A tuple representing a batch of samples.
+
+    Attributes:
+        observations (numpy.ndarray): Non-flattened array of observations.
```
ryanjulian, line 573: this will need env_info and agent_info fields

src/garage/_dtypes.py
```diff
@@ -559,3 +559,214 @@ def output_space(self):
...
+            transitions in this batch.
+        batch_size (int): The size of the batch
+
+    Raises:
+        ValueError: If any of the above attributes do not conform to their
+            prescribed types and shapes.
+    """
+    __slots__ = ()
+
+    def __new__(cls, observations, actions, rewards, next_observations,
```
ryanjulian, line 595: why not take env_spec so you can verify validate the observations and actions? we should either do this in all dtypes or none of them.

src/garage/_dtypes.py
AiRuiChen: If we remove batch_size, we would assume that all param comes with the same length. In other words we can't verify the length of inputs, and `TimeStepBatch` will rely on its producer to pass in arrs with equal length. Is this acceptable?

ryanjulian: take a look at what TrajectoryBatch does.

basically, it verifies all inputs are the same length

ahtsan:
Is it true that `TrajectoryBatch` is for on-policy algorithms, while `TimeStepBatch` is for off-policy algorithms? We should have more docstring about the major difference and preferred usage for these types so new users can have a better understanding.
Other than that, LGTM.

ryanjulian:
@ahtsan in general, yes.

AiRuiChen:
> @ahtsan in general, yes.

A bit confused by this "in general" -- are there any special exceptions? 

I will include the difference and preferred usage in docstring.

ryanjulian:
You could also expect to use TimeStepBatch in imitation learning and batch-RL. While both of these are technically "off-policy," people don't usually refer to them as such.

AiRuiChen:
> Is it true that `TrajectoryBatch` is for on-policy algorithms, while `TimeStepBatch` is for off-policy algorithms? We should have more docstring about the major difference and preferred usage for these types so new users can have a better understanding.
> Other than that, LGTM.

Just updated docstrings. I am also responsible for writing the documentation of these data types, so I can elaborate more in the documentation.

If this looks good to you, please approve it;)

PR #1552, opened 2020-06-10
Refactor out OffPolicyRLAlgorithm

### LINKED ISSUE #1392: Remove OffPolicyRLAlgorithm
This is really a bundle of helper functions masquerading as a superclass. It makes the inheritance tree unnecessarily-deep and obscures implementation details of the algorithms which use it.

We should restrict our use of inheritance to the most obvious/terse relationships (e.g. DDPG<--TD3), and not use it to group helper functions shared by a set of classes.

It can be replaced by moving its helper functions (namely, process_samples) to a function package, and moving the main algo loop into each algorithm.

maliesa96:
closes #1392


src/garage/_functions.py
```diff
@@ -159,3 +159,64 @@ def log_performance(itr, batch, discount, prefix='Evaluation'):
             tabular.record('SuccessRate', np.mean(success))
 
     return undiscounted_returns
+
+
+def process_samples(itr, paths):
```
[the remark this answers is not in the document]
maliesa96: Good catch thanks!

src/garage/_functions.py
```diff
@@ -159,3 +159,64 @@ def log_performance(itr, batch, discount, prefix='Evaluation'):
...
+    Returns:
+        dict: Processed sample data, with keys
+            * undiscounted_returns (list[float])
+            * success_history (list[float])
+            * complete (list[bool])
+
+    """
+    del itr
+
+    success_history = [
```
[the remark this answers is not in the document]
maliesa96: Looks like `garage.np._functions` already has a similar utility function, I'll put these there.

The data doesn't come from the replay buffer, it comes from `OffPolicyVectorizedSampler` right after the sampler puts these paths into the buffer. They are paths, I refactored OffPolicyVectorizedSampler to use PathBuffer, so all data that gets passed to `process_samples` is path-based, at least until OffPolicyVectorizedSampler gets deleted.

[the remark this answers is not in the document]
maliesa96: To add, this data doesn't get used by the optimizer, its just for logging and diagnostic purposes.

src/garage/_functions.py
```diff
@@ -159,3 +159,64 @@ def log_performance(itr, batch, discount, prefix='Evaluation'):
...
+
+    Returns:
+        TrajectoryBatch: Evaluation trajectories, representing the best
+            current performance of the algorithm.
+
+    """
+    paths = []
+    # Use a finite length rollout for evaluation.
+
+    for _ in range(num_trajs):
```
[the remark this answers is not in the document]
maliesa96: `SAC._evaluate_policy()` just calls this function, followed by `log_performance`.

src/garage/__init__.py
```diff
@@ -16,5 +18,7 @@
     'TrajectoryBatch',
     'log_multitask_performance',
     'log_performance',
+    'obtain_evaluation_samples',
+    'process_samples',
```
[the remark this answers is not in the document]
maliesa96: Moved these to `garage.np._functions`

src/garage/np/_functions.py
```diff
@@ -1,7 +1,67 @@
 """Utility functions for NumPy-based Reinforcement learning algorithms."""
 import numpy as np
 
+from garage._dtypes import TrajectoryBatch
 from garage.misc import tensor_utils
+from garage.sampler.utils import rollout
+
+
+def process_paths(paths):
```
ryanjulian, line 9: i still think this is an excessively-vague name to be in `garage.np` -- if you tab-completed it, you wouldn't know what it does.

How about `samples_to_tensors` which is what it appears to be doing.

src/garage/np/_functions.py
```diff
@@ -1,7 +1,67 @@
...
+    samples_data = dict(undiscounted_returns=undiscounted_returns,
+                        success_history=success_history,
+                        complete=complete)
+
+    return samples_data
+
+
+def obtain_evaluation_samples(policy, env, max_path_length=1000,
+                              num_trajs=100):
+    """Sample the policy for 10 trajectories and return average values.
```
ahtsan, line 39: I think It shouldn't be 10 trajectories but `num_trajs`.

src/garage/tf/algos/ddpg.py
```diff
@@ -248,6 +257,39 @@ def __setstate__(self, state):
...
+                for path in runner.step_path:
+                    path['rewards'] *= self._reward_scale
+                last_return = self.train_once(runner.step_itr,
+                                              runner.step_path)
+                if (cycle == 0 and self.replay_buffer.n_transitions_stored >=
+                        self._min_buffer_size):
+                    runner.enable_logging = True
+                    log_performance(runner.step_itr,
+                                    obtain_evaluation_samples(
+                                        self.policy, runner.get_env_copy()),
```
ahtsan, line 287: Should it pass a copy of the policy? Otherwise, it might alter policy's state if there's any, e.g. hidden state for recurrent policies, but it might not be an issue since this is called after the training, at the end of a training epoch.

Perhaps a safer approach is to make a policy copy inside `obtain_evaluation_samples()` so it doesn't modify the policy unexpectedly.

ryanjulian: This is a valid question, though it's worth nothing that we don't have RNN support for off-policy algos yet (it's possible but is a special training procedure)

src/garage/np/_functions.py
maliesa96: Thanks, fixed. I agree generally speaking, though this refactor didn't involve much logic changing, mainly moving things out of `OffPolicyRLAlgorithm`.

src/garage/np/_functions.py
maliesa96: Sure, fixed.

maliesa96:
@Mergifyio rebase

PR #1569, opened 2020-06-16
Return log std in agent_info

### LINKED ISSUE #1509: log_std in agent info is wrong
After the tfp patch, log_std in agent info is actually standard deviation, but not the logarithm of it.

ahtsan:
Fix #1509

### DIFF
--- a/src/garage/tf/embeddings/gaussian_mlp_encoder.py
+++ b/src/garage/tf/embeddings/gaussian_mlp_encoder.py
@@ -125,12 +125,10 @@ class GaussianMLPEncoder(StochasticEncoder, StochasticModule):
         """
         with tf.compat.v1.variable_scope(self._name) as vs:
             self._variable_scope = vs
-            self._dist, _, _ = self.model.build(embedding_input, name=name)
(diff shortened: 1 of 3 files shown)


src/garage/tf/embeddings/gaussian_mlp_encoder.py
```diff
@@ -127,8 +127,10 @@ def build(self, embedding_input, name=None):
             self._variable_scope = vs
             self._dist, _, _ = self.model.build(embedding_input, name=name)
             self._f_dist = tf.compat.v1.get_default_session().make_callable(
-                [self._dist.sample(), self._dist.loc,
-                 self._dist.stddev()],
+                [
+                    self._dist.sample(), self._dist.loc,
+                    tf.math.log(self._dist.stddev())
```
ryanjulian, line 132: can't you capture the `mean` `log_std` in the call to build above, rather than referring to `dist.loc` and running `tf.log` every time?

ahtsan: That’s right, will fix

PR #1570, opened 2020-06-16
Fix dones bug in VecEnvExecutor

### LINKED ISSUE #1435: VecEnvExecutor incorrectly copies dones array
`dones` should be a `copy.deepcopy` rather than a reference to `completes`, though changing this to a deepcopy breaks `OnPolicyVectorizedSampler` when used on nonterminating environments.

maliesa96:
closes #1435.

### DIFF
--- a/src/garage/sampler/off_policy_vectorized_sampler.py
+++ b/src/garage/sampler/off_policy_vectorized_sampler.py
@@ -114,9 +114,8 @@ class OffPolicyVectorizedSampler(BatchSampler):
 
             actions, agent_infos = policy.get_actions(input_obses)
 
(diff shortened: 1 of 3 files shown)


maliesa96:
@mergifyio rebase

maliesa96:
@ryanjulian I did, but @krzentner  would rather I do that after the release because removing `OffPolicyVectorizedSampler` right now would break resuming training for off-policy algos, so I just went with the simple fix for now.

ryanjulian:
@mergifyio rebase

ryanjulian:
@mergifyio backport release-2019.10

PR #1576, opened 2020-06-17
Make docker image smaller

gitanshu:


gitanshu:
@Mergifyio rebase

docker/Dockerfile
```diff
@@ -91,19 +91,13 @@ RUN python -m virtualenv $VIRTUAL_ENV
...
 # Prevent pip from complaining about available upgrades inside virtualenv
-RUN pip install --upgrade pip setuptools wheel
+RUN pip install --upgrade pip setuptools wheel && \
+  rm -r $HOME/.cache/pip
 
 # We need a MuJoCo key to install mujoco_py
 # In this step only the presence of the file mjkey.txt is required, so we only
 # create an empty file
 RUN touch $HOME/.mujoco/mjkey.txt
-RUN pip install garage && \
```
avnishn, line 100: where will garage be installed in the container now?

docker/Dockerfile
```diff
@@ -91,19 +91,13 @@ RUN python -m virtualenv $VIRTUAL_ENV
...
-    pip install garage[all] && \
-    rm -r $HOME/.cache/pip
-
-COPY --chown=$user:$user docker/entrypoint-runtime.sh $HOME/code/garage/docker/entrypoint-runtime.sh
```
ryanjulian, line 104: if you remove this, can i base a new image on this one without declaring an entrypoint?

e.g.
```Docker
FROM garage-18.04

RUN pip install requests
```

docker/Dockerfile
```diff
@@ -156,15 +150,16 @@ COPY --chown=$user:$user .pre-commit-config.yaml $HOME/code/garage
...
 
 # Install deps (but not the code)
-# Note that we don't delete the pip cache for dev builds
 RUN rm -rf $VIRTUAL_ENV && \
   python -m virtualenv $VIRTUAL_ENV && \
   pip install --upgrade pip setuptools wheel && \
   pip install .[all,dev] && \
-  rm $HOME/.mujoco/mjkey.txt
+  rm $HOME/.mujoco/mjkey.txt && \
+  rm -r $HOME/.cache/pip
```
ryanjulian, line 162: does this make https://github.com/rlworkgroup/garage/pull/1576/files#diff-ebacf6f6ae4ee68078bb16454b23247dR170 run slower?

docker/Dockerfile
ryanjulian: ...and by what mechanism does the entrypoint make the image bigger, anyway?

gitanshu: It doesn't, I put it back ┬━┬ ノ( ゜-゜ノ)

docker/Dockerfile
gitanshu: Line 170 should find everything in venv and wouldn't have to go to pip cache. It wasn't affected in my local test

docker/Dockerfile
ryanjulian: doesn't this make this container useless as a standalone?

gitanshu: Okay so my plan was to fix this for now so that CI becomes faster and then add a separate target that installs garage from pypi

gitanshu: @avnishn it's installed at the end of Dockerfile, using setup.py

ryanjulian: okay -- can you do that in this PR?

we should probably make that the default target, something like `rlworkgroup/garage` where `rlworkgroup/garage:2020.06.0` corresponds to the releases, etc.

gitanshu: This was a problem when our latest release on pypi was 2019.10 because the deps were different. So docker would end up installing both, say TF 1.15 and TF 2, which are like 500 MB each at download and much bigger at install. Now our pypi release and master have almost same deps, eventually it'll drift again.

ryanjulian: I agree. we will need some solution to this, but not right now.

Perhaps file an issue about version-specific Dockerfile?

gitanshu: The solution I had for this depends on BuildKit since it treats multi-stage builds as a tree instead of a line. With regular docker builder, this would probably need separate dockerfiles for separate targets. Which is why I wasn't doing it in this PR. I'll put back the pypi install for now.

ryanjulian: I'm okay removing this for now as long as you leave an issue, especially if it makes the builds faster.

Having a nice docker hub is theoretical, having a faster  build is very much a problem right now.

gitanshu:
opened #1625 to track pending docker related tasks not addressed by this PR

gitanshu:
@Mergifyio rebase

PR #1578, opened 2020-06-18
Documentation on testing

irisliucy:
A PR on testing documentation.

### DIFF
--- a/docs/user/testing.md
+++ b/docs/user/testing.md
@@ -4,7 +4,7 @@ In this section, we will talk about how to test different modules and algorithms
 
 In garage, we use [pytest](https://docs.pytest.org/en/stable/getting-started.html) to carry out tests.
 
-## Test modules
+## Test Modules
 
 All test files are stored under `tests/` in the garage respository.
 
@@ -18,10 +18,10 @@ Test modules of garage are structured in the following ways:
 - modules in `tests/` and `tests/fixtures` contain helper codes and fixtures which make writing good tests easier.
 
 ```bash
-├── tests
-│   ├── fixtures
-│   ├── garage
-│   └── integration_tests
+tests
+├── fixtures
+├── garage
+├── integration_tests
 ├── helpers.py
 └── (helper_files.py)
 ```
@@ -40,7 +40,14 @@ $ conda activate myenv
 $ source myenv/bin/activate
 ```
 
-Next, we need to install pytest. Generally, pytest should have already been installed upon garage installation.
+Next, we need to install pytest. Generally, pytest should have already been installed upon garage's `dev` installation. To install the garage environment for testing, you will need to install the `garage[all,dev]` for dependencies such as pytest, mujoco and so on.
+
+Also, you may want to check out our installation guide first before diving into tests. It is also recommened to check out official [pytest](https://docs.pytest.org/en/stable/getting-started.html) documentation.
+
+```bash
+$ cd path/to/garage/
+$ pip install -e '.[all,dev]'
+```
 
 To get started, run pytest as follows:
 
@@ -48,9 +55,9 @@ To get started, run pytest as follows:
 $ pytest
 ```
 
-If you haven't done so, please go ahead and check out our installation guide first. It is also recommened to check out official [pytest](https://docs.pytest.org/en/stable/getting-started.html) documentation.
-
 ## Basic Pytest Usage
+
+Congrats! Now you're ready for testing garage!
 
 We will start with a simple test. To be consistent with pytest requirement and garage modules, we name our test `.py` file with pre-fix `test_`.
 
@@ -91,7 +98,7 @@ class TestVPG(...):
             env.close()
 ```
 
-## Run tests
+## Run Tests
 
 In general, we can start running tests simply by:
 
@@ -101,7 +108,7 @@ $ pytest
 
 However, in most of the cases, we simply don't have the time to test everything, plus Travis CI will take care of the  majority of tests upon the deployment workflow. We can use the following ways to carry out specific tests to make life easier.
 
-### Specifying tests / selecting tests
+### Specifying Tests / Selecting Tests
 
 Pytest supports several ways to run and select tests.
 
@@ -130,15 +137,17 @@ $ pytest -k test_ppo_pendulum_continuous_baseline
 
 ```
 
-## Useful garage methods
+## Useful Pytest Methods
 
-Garage uses two helper functions `setup_method` and `teardown_method` to set up and close down test environment properly. For details, check [this](https://docs.pytest.org/en/2.8.7/xunit_setup.html).
+Below are the pytest methods and functions that we found helpful for testing garage.
 
-### Set-up and tear-down methods
+### Setup and teardown methods
 
 `setup_method` is called before every tests to set up the test environment. It setups any state tied to the execution of the given method in a class.  `setup_method` is invoked for every test method of a class.
 
 `teardown_method` is called after every tests to teardown any state that was previously setup with a `setup_method`.
+
+For details on `setup_method` and `teardown_method`, check [this](https://docs.pytest.org/en/2.8.7/xunit_setup.html).
 
 ```Python
 class TestSampleClass:
@@ -153,9 +162,7 @@ class TestSampleClass:
 
 ```
 
-## Useful pytest commands
-
-### Parametrized Test Functions
+### Parametrized test functions
 
 Parametrized test function is a  delightful solution to save us from tedious testing in same scenarios with different parameters. We can simply specify the name of the arguments that will be pass to the test function and a list of arguments corresponding to the names.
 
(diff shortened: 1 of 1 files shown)


irisliucy:
@Mergifyio rebase

irisliucy:
@ahtsan Thanks for the note! `setup_method` and `teardown_method` are included now.

docs/user/testing.md
```diff
@@ -0,0 +1,151 @@
...
+We begin by creating a file called `test_vpg.py` and put it under the tree `garage/tests/garage/tf/algos/`.
+
+We want to test VPG in the cart pole environment. Hence, we create a unitest named `test_vpg_cartpole`. Inside the function, we define the environemnt, the policy and the baselien and feed them to the VPG algorithm and run the experiment. Finally, we make sure the return value is identical to our expectation by using `assert`.
+
+Sometimes, we modele multiple tests into a class for modular and scalable structure. There is no need to subclass or anything, but make sure the prefix of the class starts with `Test`, otherwise the class will be skipped.
+
+```python
+# test_vpg.py
+
+import pytest
```
ryanjulian, line 53: is this import unused?

docs/user/testing.md
```diff
@@ -0,0 +1,151 @@
...
+... # import other dependencies
+
+class TestVPG(...):
+    def test_vpg_cartpole(self):
```
ryanjulian, line 57: new xunit-style tests are not encouraged. especially if a test doesn't need `setup_method` and `teardown_method`, you can just use functions which start with `test_` outside of any class.

docs/user/testing.md
```diff
@@ -0,0 +1,151 @@
...
+```bash
+#!/bin/bash
+$ pytest
+```
+
+However, in most of the cases, we simply don't have the time to test everything, plus Travis CI will take care of the  majority of tests upon the deployment workflow. We can use the following ways to carry out specific tests to make life easier.
+
+### Specifying tests / selecting tests
+
+Pytest supports several eays to run and select tests.
```
ryanjulian, line 95: ways

docs/user/testing.md
```diff
@@ -0,0 +1,151 @@
...
+#!/bin/bash
+$ pytest tests/garage/tf/algo/
+```
+
+#### Run tests in a module
+
+Run a test on a particular module by specifying a file path.
+
+```bash
+#!/bin/bash
```
ryanjulian, line 111: no need for these hash-bangs

docs/user/testing.md
```diff
@@ -0,0 +1,151 @@
+# Testing
+
+In this section, we will talk about how to test different modules and algorithms in garage.
+
+In garage, we use [pytest](https://docs.pytest.org/en/stable/getting-started.html) to carry out tests.
+
+## Test modules
+
+All test files are stored under `garage/tests` in the garage respository.
```
ryanjulian, line 9: can you update all of these paths to remove the first `garage/` -- consider "root" to be inside the garage folder. so the path would be `tests/`

ryanjulian: to be more specific here:

`tests/` contains all tests and supporting code
`tests/garage` contains unit tests
`tests/integration_tests` contains integration tests
modules in `tests/` and `tests/fixtures` contain helper code and fixtures which make writing good tests easier.

docs/user/testing.md
ryanjulian: sorry, a shebang is the `#!` line at the top of the file

docs/user/testing.md
irisliucy: Got it, I added this note on the doc too.

docs/user/testing.md
```diff
@@ -0,0 +1,182 @@
...
+
+Test modules of garage are structured in the following ways:
+
+- `tests/` contains all tests and supporting code
+- `tests/garage` contains unit tests
+- `tests/integration_tests` contains integration tests
+- modules in `tests/` and `tests/fixtures` contain helper codes and fixtures which make writing good tests easier.
+
+```bash
+├── tests
```
ryanjulian, line 21: nice!

docs/user/testing.md
```diff
@@ -0,0 +1,182 @@
...
+Test modules are organized in the same way as the garage main repository. Ideally, all main modules and files under `src/garage/` should be covered with test cases.
+
+Test modules of garage are structured in the following ways:
+
+- `tests/` contains all tests and supporting code
+- `tests/garage` contains unit tests
+- `tests/integration_tests` contains integration tests
+- modules in `tests/` and `tests/fixtures` contain helper codes and fixtures which make writing good tests easier.
+
+```bash
```
ryanjulian, line 20: note that if you just want fixed-width fonts, you can use ``` (code fences) without a programming language name

docs/user/testing.md
ryanjulian: shouldn't helpers.py and helper_files.py have parent tests/, instead of a sibling of tests/? perhaps i don't understand the diagram

docs/user/testing.md
```diff
@@ -0,0 +1,182 @@
...
+├── tests
+│   ├── fixtures
+│   ├── garage
+│   └── integration_tests
+├── helpers.py
+└── (helper_files.py)
+```
+
+## Prerequisites
```
ryanjulian, line 29: i think you should mention that none of this will work unless you install `garage[dev]`

docs/user/testing.md
```diff
@@ -0,0 +1,182 @@
...
+Run tests by keyword expressions. This is useful for running particular test function(s).
+
+```bash
+$ pytest -k test_ppo_pendulum_continuous_baseline
+
+```
+
+## Useful garage methods
+
+Garage uses two helper functions `setup_method` and `teardown_method` to set up and close down test environment properly. For details, check [this](https://docs.pytest.org/en/2.8.7/xunit_setup.html).
```
ryanjulian, line 135: this is actually just a feature of pytest (xunit support), but here i think you imply garage introduced this specifically.

docs/user/testing.md
irisliucy: You're right, I made a mistake here. They are all children of parent `tests/`

docs/user/testing.md
irisliucy: Right! I think I'll also include  `pip install 'garage[all,dev]'` since there are examples and tests which require mujoco and dm_control package.

docs/user/testing.md
irisliucy: Cool. I've markdown lint installed and it keeps bugging me that a language should be next to fenced codes. I think including a language will have some syntax highlighting as well.

ryanjulian: LGTM! We should add markdown lint to pre-commit!

ryanjulian: oh great point! thanks!

irisliucy:
@Mergifyio rebase

irisliucy:
@Mergifyio rebase

maliesa96:
One change I would recommend is adding a section for running marked tests.
```
@pytest.mark.mujoco
def test_something():
      ...
```

can be run with `pytest -m "mujoco"`. LGTM otherwise.

docs/user/testing.md
```diff
@@ -0,0 +1,189 @@
...
+```bash
+$ pytest -k test_ppo_pendulum_continuous_baseline
+
+```
+
+## Useful Pytest Methods
+
+Below are the pytest methods and functions that we found helpful for testing garage.
+
+### Setup and teardown methods
```
krzentner, line 144: Instead of documenting these methods (which are generally not recommended), can you document `pytest.fixture`? There's a good example of teardown with fixtures in `tests/fixtures/sampler/ray_fixtures.py`, and a few other good examples throughout the tests directory.

docs/user/testing.md
```diff
@@ -0,0 +1,189 @@
...
+
+We will start with a simple test. To be consistent with pytest requirement and garage modules, we name our test `.py` file with pre-fix `test_`.
+
+Let's write a simple test case for the vpg algorithm `src/garage/tf/algos/vpg.py`.
+
+We begin by creating a file called `test_vpg.py` and put it under the tree `tests/garage/tf/algos/`.
+
+We want to test VPG in the cart pole environment. Hence, we create a unitest named `test_vpg_cartpole`. Inside the function, we define the environemnt, the policy and the baselien and feed them to the VPG algorithm and run the experiment. Finally, we make sure the return value is identical to our expectation by using `assert`.
+
+In new xunit-style tests, multiple tests are modeled into a class for modular and scalable structure. There is no need to subclass or anything, but make sure the prefix of the class starts with `Test`, otherwise the class will be skipped. Note that it's not encouraged to use this style, especially when a test doesn't require `setup_method` and `teardown_method`.
```
gitanshu, line 70: To me, the first sentence sounds like that's the recommended way, while it's not.

Things I think that need to be mentioned are tests can be top level functions, don't subclass unittest.TestCase and basically follow pytest's recommendation.

ryanjulian:
@Mergifyio rebase

ryanjulian:
@Mergifyio rebase

ryanjulian:
@Mergifyio rebase

ryanjulian:
@Mergifyio rebase

PR #1583, opened 2020-06-19
Don't collect too many rollouts in RaySampler

avnishn:
There is an edge use case where samples aren't collected from workers
because `num_samples` have already been collected from some subset of
the sampler workers, then the other workers should have their calls
to rollout canceled. At the moment, the fix for this is to only call
rollout on workers if rollout should be called on them. For example
If there are 2 workers, and they are supposed to each collect 16
samples and then a call to obtain samples is made asking for only 16
samples, then only the first sampler worker will be started. The best
way for us to handle in the future is to shutdown workers using some
ipc, but handling starting workers the way that is mentioned above
is the best fix for a backport.

### DIFF
--- a/tests/garage/sampler/test_ray_batched_sampler.py
+++ b/tests/garage/sampler/test_ray_batched_sampler.py
@@ -58,15 +58,37 @@ class TestSampler:
         assert (trajs1[0]['observations'].shape == np.array(
             trajs2[0]['observations']).shape == (6, ))
         traj2_action_shape = np.array(trajs2[0]['actions']).shape
-        assert (trajs1[0]['actions'].shape == traj2_action_shape == (6, ))
-        assert (sum(trajs1[0]['rewards']) == sum(trajs2[0]['rewards']) == 1)
+        assert trajs1[0]['actions'].shape == traj2_action_shape == (6, )
+        assert sum(trajs1[0]['rewards']) == sum(trajs2[0]['rewards']) == 1
 
         true_obs = np.array([0, 1, 2, 6, 10, 14])
         true_actions = np.array([2, 2, 1, 1, 1, 2])
         true_rewards = np.array([0, 0, 0, 0, 0, 1])
         for trajectory in trajs1:
-            assert (np.array_equal(trajectory['observations'], true_obs))
-            assert (np.array_equal(trajectory['actions'], true_actions))
-            assert (np.array_equal(trajectory['rewards'], true_rewards))
+            assert np.array_equal(trajectory['observations'], true_obs)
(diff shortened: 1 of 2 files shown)


ryanjulian:
I disagree that an IPC to stop workers is the best solution. That kind of synchronization can get really messy.

Why not just let everything run and throw out the extra samples?

avnishn:
Sure I will add a similar test to master. In the interest of cutting the release and backport, can I do that afterwards, or do you need it now?

ryanjulian:
We don't merge bugfixes without tests which prove the bug is fixed, per CONTRIBUTING.md

tests/garage/sampler/test_ray_batched_sampler.py
```diff
@@ -58,15 +58,42 @@ def test_ray_batch_sampler(self):
...
+        There is an edge use case where samples aren't collected from workers
+        because `num_samples` have already been collected from some subset of
+        the sampler workers, then the other workers should have their calls
+        to rollout canceled. At the moment, the fix for this is to only call
+        rollout on workers if rollout should be called on them. For example
+        If there are 2 workers, and they are supposed to each collect 16
+        samples and then a call to obtain samples is made asking for only 16
+        samples, then only the first sampler worker will be started. The best
+        way for us to handle in the future is to shutdown workers using some
+        ipc, but handling starting workers the way that is mentioned above
```
ryanjulian, line 86: the best way is not to use IPC. Please remove that part of the comment.

tests/garage/sampler/test_ray_batched_sampler.py
```diff
@@ -58,15 +58,42 @@ def test_ray_batch_sampler(self):
...
+            assert np.array_equal(trajectory['observations'], true_obs)
+            assert np.array_equal(trajectory['actions'], true_actions)
+            assert np.array_equal(trajectory['rewards'], true_rewards)
         sampler1.shutdown_worker()
         sampler2.shutdown_worker()
+
+    def test_ray_sampler_idle_workers(self):
+        """Addresses issue #1349
+
+        There is an edge use case where samples aren't collected from workers
```
ryanjulian, line 77: This isn't really an edge case.

ryanjulian:
@avnishn Is this already fixed on master?

avnishn:
Sorry for the late response. @krzentner can confirm that this is fixed on master.

ryanjulian:
Is the API contract that the sampler returns exactly as many samples as asked for, or greater than or equal to?

avnishn:
> Is the API contract that the sampler returns exactly as many samples as asked for, or greater than or equal to?

Greater than or equal to.

PR #1587, opened 2020-06-19
Drop python 3.5 support

### LINKED ISSUE #1441: Remove Python 3.5 support
2020.06 will be the last version of garage to support Python 3.5.

gitanshu:
python 3.5 reaches end-of-life on 2020-09-13. So we can drop 3.5 support for 2020.09 release of garage.

Also revert commit 7f573eecba7b869fbb1f00126cd031a878bfd778.

closes #1441


docker/Dockerfile
```diff
@@ -39,19 +29,13 @@ RUN \
     # OpenAI baselines
     libopenmpi-dev \
     # virtualenv
-    python3.5 \
-    python3.5-dev \
-    python3.5-tk \
-    python3.5-venv && \
+    python3 \
```
ryanjulian, line 32: can you pin this to python3.6, so we don't get caught off guard as before? i think deadsnakes has specific versions if apt doesn't

you could also consider just using pyenv to isolate yourself from Ubuntu.

gitanshu: Yeah pyenv is a good idea. I am using that now.

README.md
```diff
@@ -53,7 +53,7 @@ The table below summarizes the algorithms available in garage.
 | MTTRPO                 | PyTorch, TensorFlow |
 
 ## Supported Tools and Frameworks
-garage supports Python 3.5+
+Starting from version 2020.09, garage supports Python 3.6+. 2020.06 is the last version of garage that supports Python 3.5.
```
ryanjulian, line 56: was?

ryanjulian:
You can merge this, now that the release branch has been cut

docker/Dockerfile
```diff
@@ -7,16 +7,6 @@ FROM $PARENT_IMAGE AS garage-18.04
 # > At the moment, setting "LANG=C" on a Linux system *fundamentally breaks Python 3*, and that's not OK.
 ENV LANG C.UTF-8
```
krzentner, line 8: BTW, I don't think the C.UTF-8 logic is necessary anymore.

ryanjulian: Yeah -- i think this was fixed in Py36

gitanshu: Looks like it was fixed in 3.7 https://bugs.python.org/issue19846#msg308567

ryanjulian: Ah, i guess we'll just have to live with it then.

ryanjulian:
@Mergifyio rebase

gitanshu:
@Mergifyio rebase

PR #1588, opened 2020-06-19
Update setup_linux to ignore purge errors

maliesa96:
`sudo apt purge -y` breaks the script if the package is not found, at least on ubuntu 20.04.

### DIFF
--- a/scripts/setup_linux.sh
+++ b/scripts/setup_linux.sh
@@ -144,18 +144,23 @@ sudo apt install -y \
 
 # Build GLFW because the Ubuntu 16.04 version is too old
 # See https://github.com/glfw/glfw/issues/1004
-sudo apt purge -y libglfw*
-GLFW_DIR="$(mktemp -d)/glfw"
-git clone https://github.com/glfw/glfw.git "${GLFW_DIR}"
-cd "${GLFW_DIR}"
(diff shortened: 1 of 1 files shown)


ryanjulian:
Did you verify building libglfw from source is necessary on 20.04?

gitanshu:
[Ubuntu 20.04 comes with libglfw3-* 3.3.2-1](https://packages.ubuntu.com/focal/libglfw3-dev). The latest we need for garage is 3.3.0. So we can skip building it. 

Also, our setup script install the master branch while the dockerfile installs 3.3.0. We should make that consistent.

maliesa96:
> Did you verify building libglfw from source is necessary on 20.04?

I don't think its an issue to build from source even if it is unnecessary (as long as its a stable release that doesn't cause issues), but I updated the script to avoid doing so anyways.

maliesa96:
> [Ubuntu 20.04 comes with libglfw3-* 3.3.2-1](https://packages.ubuntu.com/focal/libglfw3-dev). The latest we need for garage is 3.3.0. So we can skip building it.
> 
> Also, our setup script install the master branch while the dockerfile installs 3.3.0. We should make that consistent.

makes sense, I updated the script to point to the commit hash of the 3.3 release.

ryanjulian:
@Mergifyio rebase

Issue #46, opened 2018-06-11
Bullet support


ryanjulian:
We would like to add support for the [Bullet physics engine](https://github.com/bulletphysics/bullet3) to rllab. Thankfully, the Bullet team have recently provided Python bindings in the form of pybullet, and even provides [examples](https://github.com/bulletphysics/bullet3/tree/2.87/examples/pybullet/gym/pybullet_envs) of how to implement the `gym.Env` interface (from OpenAI Gym) using pyBullet.

This task is to add `pybullet` to the rllab `conda` environment, and implement a class (similar to `GymEnv`, e.g. `BulletEnv`) which allows any rllab algorithm to learn against pybullet environments. You will also need to implement the plot interface, if pybullet does not already, which shows the user a 3D animation of the environment. Essentially, you should duplicate the experience of running one of the MuJoCo-based examples (e.g. [trpo_swimmer.py](https://github.com/ryanjulian/rllab/blob/integration/examples/trpo_swimmer.py)), but using a Bullet environment instead. You should include examples (in `examples/` and `sandbox/rocky/tf/launchers/`) of launcher scripts which use an algorithms (suggestion: TRPO) to train the `KukaGymEnv` environment.

This is conceptually the same as [GymEnv](https://github.com/ryanjulian/rllab/blob/integration/rllab/envs/gym_env.py), which allows rllab users to import any OpenAI Gym environment and learn against them. In fact, pybullet environments implement the Gym interface, so in theory we should be done as soon as we can import pybullet. In practice, our constructor for Gym environments only takes the string name (e.g. "Humanoid-v1") of a Gym environment, not the class of a Gym environment. The pybullet environments do not have string shortcuts because they are not part of the official Gym repository. Furthermore, we'd like to use other unofficial Gym environments in rllab, but it is currently difficult for the same reason.

So you might structure this task as two pull requests (1) adding pybullet to the conda environment and (2) Modifying `GymEnv` to support arbitrary environments which implement the `gym.Env` interface (attempted in https://github.com/ryanjulian/rllab/pull/12).

Consider this a professional software engineering task, and provide a high-quality solution which does not break existing users, minimizes change, and is stable. Please always use [PEP8](https://www.python.org/dev/peps/pep-0008/) style in your code, and format it using [YAPF](https://github.com/google/yapf) (with the PEP8 setting). Submit your pull request against the `integration` branch of this repository.

Some notes:
[...]

PR #1589, opened 2020-06-19
Add Pybullet support

AiRuiChen:
Initial step to address FR [#46](https://github.com/rlworkgroup/garage/issues/46).

Future steps may include:
1. Request upstream update of Pybullet envs API (e.g. [KukaGymEnv](https://github.com/bulletphysics/bullet3/blob/master/examples/pybullet/gym/pybullet_envs/bullet/kukaGymEnv.py#L78) won't run in garage now because it unconditionally `resetSimulation()`, which globally resets parallel environment instances. ) 
_or_
Write our own bullet interface
2. Figure out pybullet client/server architecture and possibly support multiple simulations with one server. (Using `SHARED_MEMORY` connection mode maybe?)


examples/step_bullet_env.py
```diff
@@ -0,0 +1,32 @@
...
+
+This example requires that garage[pybullet] be installed.
+"""
+import argparse
+
+from pybullet_envs.bullet.kukaGymEnv import KukaGymEnv
+
+from garage.envs.pybullet import PybulletEnv
+
+parser = argparse.ArgumentParser()
```
ryanjulian, line 12: please use click for argument parsing, as in the other examples

examples/step_bullet_env.py
```diff
@@ -0,0 +1,32 @@
...
+# Reset the environment and launch the viewer
+env.reset()
+env.render()
+
+# Step randomly until interrupted
+steps = 0
+while True:
+    if steps == args.n_steps:
+        break
+    env.step(env.action_space.sample())
```
ryanjulian, line 31: you shouldn't step the environment if it returns `done` (the termination signal)

examples/tf/trpo_cartpole_bullet.py
```diff
@@ -0,0 +1,54 @@
...
+from garage.envs.pybullet import PybulletEnv
+from garage.experiment import LocalTFRunner
+from garage.experiment.deterministic import set_seed
+from garage.np.baselines import LinearFeatureBaseline
+from garage.tf.algos import TRPO
+from garage.tf.policies import CategoricalMLPPolicy
+
+
+@wrap_experiment
+def trpo_cartpole(ctxt=None, seed=1):
```
ryanjulian, line 22: please name the experiment according to the example name

src/garage/envs/pybullet/__init__.py
```diff
@@ -0,0 +1,13 @@
+"""Wrappers for the Pybullet based gym environments.
+
+See https://github.com/bulletphysics/bullet3/tree/master/examples/pybullet
+"""
+try:
+    import pybullet_envs  # noqa: F401
+except Exception as e:
+    raise ImportError("To use garage's pybullet wrappers, please install "
+                      'garage[pybullet].')
```
ryanjulian, line 9: Please stick to single quotes

also, i think you should remove the trailing period, since some people might get confused copy-pasting.

src/garage/envs/pybullet/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+
+from garage.envs.pybullet.pybullet_env import PybulletEnv
```
ryanjulian, line 11: pybullet styles themselves as PyBullet, so this should either be `PyBulletEnv`, or simply `BulletEnv` is probably sufficient (since PyBullet is just the Python interface for Bullet).

tests/garage/envs/pybullet/test_pybullet_env.py
```diff
@@ -0,0 +1,106 @@
...
+from pybullet_envs.bullet.racecarZEDGymEnv import RacecarZEDGymEnv
+from pybullet_utils.bullet_client import BulletClient
+import pytest
+
+from garage.envs.pybullet import PybulletEnv
+from tests.helpers import step_env
+
+
+@pytest.fixture
+def envs():
```
ryanjulian, line 20: i think you want a parameterized test here, not a fixture. Fixtures are generally for shared setup, rather than enumerating tests.

src/garage/envs/bullet/__init__.py
```diff
@@ -0,0 +1,13 @@
+"""Wrappers for the py_bullet based gym environments.
+
+See https://github.com/bulletphysics/bullet3/tree/master/examples/pybullet
+"""
+try:
+    import pybullet_envs  # noqa: F401
+except Exception as e:
+    raise ImportError('To use garage\'s py_bullet wrappers, please install '
```
ryanjulian, line 8: pybullet?

tests/garage/envs/bullet/test_bullet_env.py
```diff
@@ -0,0 +1,92 @@
...
+import pickle
+
+from pybullet_envs.bullet.cartpole_bullet import CartPoleBulletEnv
+from pybullet_envs.bullet.cartpole_bullet import CartPoleContinuousBulletEnv
+from pybullet_envs.bullet.kukaCamGymEnv import KukaCamGymEnv
+from pybullet_envs.bullet.kukaGymEnv import KukaGymEnv
+from pybullet_envs.bullet.minitaur_duck_gym_env import MinitaurBulletDuckEnv
+from pybullet_envs.bullet.minitaur_gym_env import MinitaurBulletEnv
+from pybullet_envs.bullet.racecarGymEnv import RacecarGymEnv
+from pybullet_envs.bullet.racecarZEDGymEnv import RacecarZEDGymEnv
```
ryanjulian, line 11: maybe use this? https://github.com/bulletphysics/bullet3/blob/master/examples/pybullet/gym/pybullet_envs/__init__.py#L230

examples/tf/trpo_cartpole_bullet.py
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env python3
+"""This is an example to train a task with TRPO algorithm.
+
+Here it runs CartPoleBulletEnv environment with 100 iterations.
+
+Results:
+    AverageReturn: 100
+    RiseTime: itr 13
```
naeioi, line 8: These numbers were tested in `trpo_cartpole.py`. You need to update them (or delete them)

tests/garage/envs/bullet/test_bullet_env.py
```diff
@@ -0,0 +1,90 @@
...
+    """Test Bullet environments can step"""
+
+    for env_id in env_ids:
+        # extract id string
+        env_id = env_id.replace('- ', '')
+        if env_id == 'KukaCamBulletEnv-v0':
+            # Kuka environments calls py_bullet.resetSimulation() in reset()
+            # unconditionally, which globally resets other simulations. So
+            # only one Kuka environment is tested.
+            continue
```
naeioi, line 24: why does the reset of `KukaCamBulletEnv-v0` matter? every iteration creates a new bullet environment, so the reset in the last iteration should not impact the next iteration. Or am I missing something?

tests/garage/envs/bullet/test_bullet_env.py
```diff
@@ -0,0 +1,90 @@
...
+@pytest.mark.parametrize('env_ids', [pybullet_envs.getList()])
+def test_pickleable(env_ids):
+    """Test Bullet environments are pickleable"""
+    for env_id in env_ids:
+        # extract id string
+        env_id = env_id.replace('- ', '')
+        if env_id == 'RacecarZedBulletEnv-v0':
+            env = BulletEnv(gym.make(env_id, renders=False))
+        else:
+            env = BulletEnv(gym.make(env_id))
```
naeioi, line 51: Why not always set `renders=False`?

AiRuiChen: By default all classes don't render the visualization. The only exception is class `RacecarZedBulletEnv`. 

However, some classes have the parameter name as `render` instead of `renders`. See [here](https://github.com/bulletphysics/bullet3/blob/master/examples/pybullet/gym/pybullet_envs/minitaur/envs/minitaur_gym_env.py#L92). Some classes don't have `render` for their constructors. See [here](https://github.com/bulletphysics/bullet3/blob/master/examples/pybullet/gym/pybullet_envs/gym_pendulum_envs.py#L10).

So the if/else logic here is necessary anyways. It's just easier to make an exception for `RacecarZedBulletEnv` instead of enumerating the other cases.

tests/garage/envs/bullet/test_bullet_env.py
AiRuiChen: I agree. Honestly I don't know why this is failing.
 
If two Kuka environments are tested, the second Kuka environment will complain about "py_bullet not connected to server" in its `reset()`. 

Some thoughts:
Kuka environment implementation doesn't store client-id, thus it assumes one running environment at a time. If there are two parallel Kuka environments, calling reset/disconnect will affect the other one. 

Is it possible that the first environment is garbage collected while the second one is running?  

My plan was to introduce bullet first. If we plan to use Kuka environment in a parallel scenario, I will spend more time and try to understand this.

tests/garage/envs/bullet/test_bullet_env.py
naeioi: I see. This inconsistency in pybullet is really painful.

tests/garage/envs/bullet/test_bullet_env.py
naeioi: My theory: `pybullet.resetSimulation()` closes both the client and the server, while `client.resetSimulation()` closes only the client. 

Please go ahead with this broken `KukaCamBulletEnv-v0` and fix it later. 

If multiple instances are problematic, this env cannot be used for training, since the sampler will create multiple instances. We should probably document this behaviour somewhere, or even disable `KukaCamBulletEnv-v0` until the problem is fixed.

ryanjulian:
@Mergifyio rebase

PR #1590, opened 2020-06-19
Add isort to pre-commit 

irisliucy:
isort is now added to pre-commit to save us from sorting import orders manually!

Packages Import order is ordered in sections like this:
```
FUTURE
STDLIB
THIRDPARTY
FIRSTPARTY
LOCALFOLDER
``` 

The sorting is tested against this [example]( https://github.com/PyCQA/flake8-import-order/blob/master/tests/test_cases/complete_google.py)

Note, It's not perfect:
1. We need to explicitly specify third party packages for now, in `setup.cfg` under the `[tool:isort]` section > `known_third_party`. Good news is the isort release 5.0 may solve the issue 👍 

2. The relative path is not sorted as desired. BUT, in garage, we are not using relative path anyway

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,0 +1,2 @@
+[settings]
+profile=google
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,2 +1,0 @@
-[settings]
-profile=google
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -54,6 +54,16 @@ repos:
     -   id: pretty-format-json               # Checks that all your JSON files have keys that are sorted and indented.
     -   id: trailing-whitespace              # Trims trailing whitespace.
 
+-   repo: https://github.com/markdownlint/markdownlint
+    rev: v0.9.0
+    hooks:
+      - id: markdownlint                     # Check markdown files and flag style issues
+        name: Markdownlint
+        description: Run markdownlint on your Markdown files
+        entry: mdl
+        language: ruby
+        files: \.(md|mdown|markdown)$
+
 # First pass: check format
 -   repo: https://github.com/pre-commit/mirrors-yapf
     rev: v0.28.0
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,6 +25,16 @@ repos:
         entry: pylint
         args: [--rcfile=tests/garage/.pylintrc]
 
+-   repo: https://github.com/timothycrosley/isort
+    rev: 4.3.21-2
+    hooks:
+    -   id: isort
+        name: isort
+        entry: isort
+        require_serial: true
+        language: python
+        types: [python]
+
 -   repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v2.2.3
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -31,7 +31,6 @@ repos:
     -   id: isort
         name: isort
         entry: isort
-        require_serial: true
         language: python
         types: [python]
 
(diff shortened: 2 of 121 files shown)


.pre-commit-config.yaml
```diff
@@ -25,6 +25,16 @@ repos:
         entry: pylint
         args: [--rcfile=tests/garage/.pylintrc]
 
+-   repo: https://github.com/timothycrosley/isort
+    rev: 4.3.21-2
+    hooks:
+    -   id: isort
+        name: isort
+        entry: isort
+        require_serial: true
```
ryanjulian, line 34: i assume isort is already parallel?

.pre-commit-config.yaml
```diff
@@ -25,6 +25,16 @@ repos:
         entry: pylint
         args: [--rcfile=tests/garage/.pylintrc]
 
+-   repo: https://github.com/timothycrosley/isort
```
ryanjulian, line 28: did the official hook end up not being very good?

irisliucy: I think this is the official one. The pre-commit one is depreciated: https://github.com/pre-commit/mirrors-isort

setup.cfg
```diff
@@ -22,6 +22,14 @@ extend-ignore =
     D107  # We document __init__ in the class docstring
     F841  # Unused variables are checked by pylint
 
+# isort config
+[tool:isort]
+sections=FUTURE,STDLIB,THIRDPARTY,FIRSTPARTY,LOCALFOLDER
+force_sort_within_sections = True
+order_by_type = False
+known_first_party = garage
+known_third_party = numpy, pytest, torch, tensorflow    # specify third party explicitly for sorting
```
ryanjulian, line 31: from setup.py:

`akro,cached_property,click,cloudpickle,cma,dowel,gym,joblib,numpy,psutil,pyglet,dateutil,ray,skimage,scipy,setproctitle,tensorflow,tensorflow_probability,torch,torchvision`

.isort.cfg
```diff
@@ -0,0 +1,2 @@
+[settings]
+profile=google
```
gitanshu, line 2: I guess this should be in setup.cfg too?

irisliucy:
@ryanjulian Yes, it recursively changes like hundreds of files and I need to manually add the unadded third-party packages to `known_third_party`.

ryanjulian:
Basically your goal here should be to get the config to not change anything (except for errors which slipped through).

.isort.cfg
gitanshu: Or rather, we should just copy the settings we need from the google profile to keep it explicit

```
"force_single_line": True,
"force_sort_within_sections": True,
"lexicographical": True,
"single_line_exclusions": ("typing",),
```
([Taken from here](https://timothycrosley.github.io/isort/reference/isort/profiles/))

irisliucy:
@Mergifyio rebase

.isort.cfg
irisliucy: Right, removed `.isort.cfg` and explicitly mentioned these params in `setup.cfg`.

benchmarks/src/garage_benchmarks/experiments/policies/__init__.py
```diff
@@ -1,20 +1,20 @@
...
-    categorical_mlp_policy)
-from garage_benchmarks.experiments.policies.continuous_mlp_policy import (
-    continuous_mlp_policy)
-from garage_benchmarks.experiments.policies.gaussian_gru_policy import (
-    gaussian_gru_policy)
-from garage_benchmarks.experiments.policies.gaussian_lstm_policy import (
-    gaussian_lstm_policy)
-from garage_benchmarks.experiments.policies.gaussian_mlp_policy import (
-    gaussian_mlp_policy)
+from garage_benchmarks.experiments.policies.categorical_cnn_policy import \
```
ryanjulian, line 2: please configure isort to use () for line breaks rather than \

src/garage/torch/value_functions/__init__.py
```diff
@@ -1,6 +1,6 @@
 """Value functions which use PyTorch."""
+from garage.torch.value_functions.gaussian_mlp_value_function import \
```
ryanjulian, line 2: note that anything with `noqa: I100` is has probably been ordered that way to prevent a circular import. we should disable isort on those lines.

PR #1591, opened 2020-06-19
Refactoring CategoricalMLPRegressor with tfp

yeukfu:
Refactoring categorical mlp regressor with tfp. [issue 1372](https://github.com/rlworkgroup/garage/issues/1372).

Below is the benchmark with [launcher script](https://github.com/rlworkgroup/garage/blob/test_categorical_mlp_baseline/tests/benchmarks/benchmark_categotical_mlp_policy.py).

environments:
` 'LunarLander-v2',
    'Assault-ramDeterministic-v4',
    'Breakout-ramDeterministic-v4',
    'ChopperCommand-ramDeterministic-v4',
    'Tutankham-ramDeterministic-v4',
Tutankham-ramDeterministic-v4_benchmark_mean

   
results: (Garage = baseline without tfp, Garage2 = baseline with tfp)

LunarLander-v2_benchmark_mean
Breakout-ramDeterministic-v4_benchmark_mean
ChopperCommand-ramDeterministic-v4_benchmark_mean

### DIFF
--- a/src/garage/tf/regressors/categorical_mlp_regressor.py
+++ b/src/garage/tf/regressors/categorical_mlp_regressor.py
@@ -140,8 +140,8 @@ class CategoricalMLPRegressor(StochasticRegressor):
 
             y_hat = self._network.y_hat
 
-            dist = self._network.categorical_dist
-            old_dist = self._old_network.categorical_dist
+            dist = self._network.dist
+            old_dist = self._old_network.dist
 
             mean_kl = tf.reduce_mean(old_dist.kl_divergence(dist))
 
--- a/src/garage/tf/regressors/categorical_mlp_regressor.py
+++ b/src/garage/tf/regressors/categorical_mlp_regressor.py
@@ -4,10 +4,10 @@ import numpy as np
 import tensorflow as tf
 
 from garage import make_optimizer
-from garage.tf.distributions import Categorical
 from garage.tf.misc import tensor_utils
-from garage.tf.models import NormalizedInputMLPModel
 from garage.tf.optimizers import ConjugateGradientOptimizer, LbfgsOptimizer
+from garage.tf.regressors.categorical_mlp_regressor_model import (
+    CategoricalMLPRegressorModel)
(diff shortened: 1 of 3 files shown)


src/garage/tf/regressors/categorical_mlp_regressor.py
```diff
@@ -274,7 +218,7 @@ def vectorized(self):
     @property
     def distribution(self):
         """garage.tf.distributions.DiagonalGaussian: Distribution."""
-        return self._dist
+        return self._network.dist
```
ahtsan, line 221: This name should be consistent with the one you specified inside the model (in this case `categorical_dist`).
I suggest renaming the name in `network_output_spec` to just `dist` since most of the models use `dist` as well.

Is there a test that will make sure `self._network.dist` exists?

zequnyu:
@yeukfu Is there any reason you implemented your benchmark scripts? FYI [this](https://github.com/rlworkgroup/garage/blob/master/benchmarks/src/garage_benchmarks/helper.py) will do most of the work if you only need to benchmark garage implementations.

src/garage/tf/regressors/categorical_mlp_regressor.py
yeukfu: Got it. Will do.
I will add the test to test `self._network.dist`.

yeukfu:
> @yeukfu Is there any reason you implemented your benchmark script? FYI [this](https://github.com/rlworkgroup/garage/blob/master/benchmarks/src/garage_benchmarks/helper.py) will do most of the work if you only need to benchmark garage implementations.
That is helperful! A small question, I try to run a benchmark with garage_benchmark. But it seems that I need credential to access GCP. Can this run locally?

ahtsan:
> @yeukfu Is there any reason you implemented your benchmark script? FYI [this](https://github.com/rlworkgroup/garage/blob/master/benchmarks/src/garage_benchmarks/helper.py) will do most of the work if you only need to benchmark garage implementations.

I think one reason is that he needs both before-change and after-change version for comparison, but the link would also be helpful.

PR #1592, opened 2020-06-21
Unify policy interface across the codebase

ahtsan:
This PR modifies `garage.np.policies.Policy`, which serves as a base policy class for np, TF and
Torch trees and unifies policy interface across the codebase.

### DIFF
--- a/src/garage/np/policies/policy.py
+++ b/src/garage/np/policies/policy.py
@@ -3,15 +3,7 @@ import abc
 
 
 class Policy(abc.ABC):
-    """Base classe for policies based on numpy.
-
-    Args:
-        env_spec (garage.envs.env_spec.EnvSpec): Environment specification.
-
-    """
-
-    def __init__(self, env_spec):
-        self._env_spec = env_spec
+    """Base class for policies based on numpy."""
 
     @abc.abstractmethod
     def get_action(self, observation):
@@ -20,32 +12,57 @@ class Policy(abc.ABC):
         Args:
             observation (np.ndarray): Observation from the environment.
 
-        Returns:
-            (np.ndarray): Action sampled from the policy.
+        """
+
+    @abc.abstractmethod
+    def get_actions(self, observations):
+        """Get actions given observations.
+
+        Args:
+            observations (torch.Tensor): Observations from the environment.
 
         """
 
-    def reset(self, dones=None):
+    def reset(self, do_resets=None):
         """Reset the policy.
 
-        If dones is None, it will be by default np.array([True]) which implies
-        the policy will not be "vectorized", i.e. number of parallel
-        environments for training data sampling = 1.
+        This is effective only to recurrent policies.
+
+        do_resets is an array of boolean indicating
+        which internal states to be reset. The length of do_resets should be
+        equal to the length of inputs, i.e. batch size.
 
         Args:
-            dones (numpy.ndarray): Bool that indicates terminal state(s).
+            do_resets (numpy.ndarray): Bool array indicating which states
+                to be reset.
 
         """
 
     @property
     def observation_space(self):
         """akro.Space: The observation space of the environment."""
-        return self._env_spec.observation_space
 
     @property
     def action_space(self):
         """akro.Space: The action space for the environment."""
-        return self._env_spec.action_space
+
+    @property
+    def env_spec(self):
+        """Policy environment specification.
+
+        Returns:
+            garage.EnvSpec: Environment specification.
+
+        """
+
+    @property
+    def name(self):
+        """Name of policy.
+
+        Returns:
+            str: Name of policy
+
+        """
 
     def log_diagnostics(self, paths):
         """Log extra information per iteration based on the collected paths.
@@ -56,41 +73,23 @@ class Policy(abc.ABC):
         """
 
     @property
-    def state_info_keys(self):
-        """Get keys describing policy's state.
+    def state_info_specs(self):
+        """State info specification.
 
         Returns:
-            List[str]: keys for the information related to the policy's state
-            when taking an action.
+            List[str]: keys and shapes for the information related to the
+                module's state when taking an action.
 
         """
         return list()
 
-    def terminate(self):
-        """Clean up operation."""
-
-
-class StochasticPolicy(Policy):
-    """Base class for stochastic policies implemented in numpy."""
-
     @property
-    @abc.abstractmethod
-    def distribution(self):
-        """Get the distribution of the policy.
+    def state_info_keys(self):
+        """State info keys.
 
         Returns:
-            garage.tf.distribution: The distribution of the policy.
+            List[str]: keys for the information related to the module's state
+                when taking an input.
 
         """
-
-    @abc.abstractmethod
-    def dist_info(self, obs, state_infos):
-        """Return the distribution information about the actions.
-
-        Args:
-            obs (np.ndarray): observation values
-            state_infos (dict): a dictionary whose values should contain
-                information about the state of the policy at the time it
-                received the observation
-
-        """
+        return [k for k, _ in self.state_info_specs]
--- a/src/garage/np/policies/policy.py
+++ b/src/garage/np/policies/policy.py
@@ -12,6 +12,10 @@ class Policy(abc.ABC):
         Args:
             observation (np.ndarray): Observation from the environment.
 
+        Returns:
+            Tuple[np.ndarray, dict[str,np.ndarray]]: Actions and extra agent
+                infos.
+
         """
 
     @abc.abstractmethod
@@ -20,6 +24,10 @@ class Policy(abc.ABC):
 
         Args:
             observations (torch.Tensor): Observations from the environment.
+
+        Returns:
+            Tuple[np.ndarray, dict[str,np.ndarray]]: Actions and extra agent
+                infos.
 
         """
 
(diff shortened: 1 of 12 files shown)


src/garage/np/policies/policy.py
```diff
@@ -20,77 +12,84 @@ def get_action(self, observation):
...
 
-class StochasticPolicy(Policy):
-    """Base class for stochastic policies implemented in numpy."""
+        """
 
     @property
-    @abc.abstractmethod
-    def distribution(self):
-        """Get the distribution of the policy.
+    def state_info_specs(self):
```
krzentner, line 76: This is really only ever been useful in the TensorFlow branch. Is it really necessary?

src/garage/tf/policies/policy.py
```diff
@@ -24,34 +25,16 @@ def get_action(self, observation):
...
-
-    @property
-    def vectorized(self):
-        """Boolean for vectorized.
-
-        Returns:
-            bool: Indicates whether the policy is vectorized. If True, it
-                should implement get_actions(), and support resetting with
-                multiple simultaneous states.
+            observations (torch.Tensor): Observations from the environment.
```
krzentner, line 35: Wait what? This return type doesn't make any sense here.

src/garage/torch/policies/policy.py
```diff
@@ -20,18 +22,10 @@ def __init__(self, env_spec, name):
...
-            observation (torch.Tensor): Observation from the environment.
-
-        Returns:
-            tuple:
-                * torch.Tensor: Predicted action.
-                * dict:
-                    * list[float]: Mean of the distribution
-                    * list[float]: Log of standard deviation of the
-                        distribution
+            observation (np.ndarray): Observation from the environment.
```
krzentner, line 28: There probably still should be documentation of the return types (even if the current docstring is wrong, it should be `Tuple[np.ndarray, dict[str,np.ndarray]]`).

src/garage/np/policies/policy.py
```diff
@@ -20,77 +12,84 @@ def get_action(self, observation):
...
 
     @property
     def action_space(self):
         """akro.Space: The action space for the environment."""
-        return self._env_spec.action_space
 
-    def log_diagnostics(self, paths):
-        """Log extra information per iteration based on the collected paths.
+    @property
+    def env_spec(self):
```
krzentner, line 50: Honestly not convinced this property should exist here. It's only used in the TensorFlow branch, so why not put it on `garage.tf.Policy` instead?

src/garage/torch/policies/policy.py
```diff
@@ -42,14 +36,6 @@ def get_actions(self, observations):
         Args:
             observations (torch.Tensor): Observations from the environment.
```
krzentner, line 38: Also would be good to have (correct) return type docs here.

tests/fixtures/policies/dummy_policy.py
```diff
@@ -88,9 +87,29 @@ def vectorized(self):
         """
         return True
 
+    @property
+    def observation_space(self):
```
krzentner, line 91: Aren't these the default definitions? Why are they repeated here?

ahtsan: One major issue I encountered is that, we can't make `garage.np.policy` an abstract class as `garage.tf.policy` has to inherit from another class. We can only make `garage.np.policy` an interface.

src/garage/np/policies/policy.py
ahtsan: > It's good to see this interface finally coming together!

Thanks for your comments! I'm not familiar with torch tree so your comments are really important and helpful.

ahtsan:
> It's good to see this interface finally coming together!

Thanks for your comments! I'm not familiar with torch tree so your comments are really important and helpful.

tests/fixtures/policies/dummy_policy.py
ahtsan: Do you think there is a workaround/better way to handle this?

krzentner: I'm not quite sure I understand the issue, to be honest. `garage.np.Policy` *does* inherit from `abc.ABC`, so I thought that made it an abstract class. Also, `DummyPolicy` here has nothing to do with `garage.tf.Policy`, so I'm doubly confused. Does anything break if you delete this method definition on `DummyPolicy`?

ahtsan: oh, it's just `garage.np.Policy` cannot have a constructor in `TensorFlow` branch, so we can't have default property.

ahtsan: Yeah, some tests need `DummyPolicy.observation_space`.

src/garage/np/policies/policy.py
```diff
@@ -20,77 +12,84 @@ def get_action(self, observation):
...
-            List[str]: keys for the information related to the policy's state
-            when taking an action.
+            str: Name of policy
 
         """
-        return list()
 
-    def terminate(self):
-        """Clean up operation."""
+    def log_diagnostics(self, paths):
```
ryanjulian, line 67: we don't actually use this protocol so i think we should remove it.

tests/fixtures/policies/dummy_policy.py
krzentner: Oh, I see what you're getting at. If we're going to have a public `env_spec` property, couldn't we just have default definitions for these properties that use the `.env_spec` property?

ryanjulian:
FWIW, the garage-wide policy API should probably look something like this

```py
import abc

class Policy(abc.ABC):
    """Policy interface."""

    @abc.abstractmethod
    def get_action(self, observation, policy_state=None):
        r"""Get action conditioned on observation.
        Args:
            observation (np.ndarray): Observation from an environment,
                conforming to :obj:`observation_space`. Shape is
                :math:`(O*,)`.
            policy_state (object): State information.
        Returns:
            np.ndarray: An action, conditioned on the observation, conforming
                to :obj:`env_spec.action_space`. Shape is :math:`(A^*,)`.
            object: State information. If not None, this should be passed to
                the next call to `get_action`.
        """

    @abc.abstractmethod
    def get_actions(self, observations, policy_states=None):
        r"""Get actions conditioned on observations.
        Note:
            This API is assumed to be vectorized, unless otherwise noted.
        Args:
            observations (np.ndarray): Array of observations of shape
                :math:`(N, O^*)` containing observations from the environment.
                These should conform to :obj:`observation_space`.
            policy_states (object): State information.
        Returns:
            np.ndarray: An array of actions of shape :math:`(N, A*)`,
                conforming to :obj:`env_spec.action_space`.
            object: State information. If not None, this should be passed to
                the next call to `get_actions`.
        """

    @abc.abstractmethod
    def get_reset_state(self):
        """Get the initial (reset) value of policy_state.
        Returns:
            object: Policy state information, to be used with get_action.
        """

    @abc.abstractmethod
    def get_reset_states(self, batch_size):
        """Get the reset value of policy_states for a given batch size.
        Args:
            batch_size (int): Observation/action batch size for the calls to
               `get_actions` this state will be used with.
        Returns:
            object: Policy state representing a newly-reset policy.
        """
```

tests/fixtures/policies/dummy_policy.py
ryanjulian: are you suggesting:

```py
# garage.tf.Policy
import numpy as np

from garage import Policy as IPolicy


class Policy(IPolicy):

    def __init__(self, env_spec, name):
        self._name = name
        self._env_spec = env_spec

    @property
    def observation_space(self):
        return self._env_spec.observation_space

    @property
    def action_space(self):
        return self._env_spec.action_space

        return self.get_actions(np.expand_dims(observation))[0]

    def get_reset_state(self):
        return self.get_reset_states(1)[0]
```

krzentner: So I think that `get_reset_state` and friends are way outside the scope of this PR. I'm just trying to give some minor suggestions here.

I don't actually think we need the `env_spec`, `observation_space`, or `action_space` properties, to be honest. The entire torch branch currently doesn't use them (and there doesn't really even seem to be anywhere they would be useful to use). Mostly this is just because I strongly prefer to not have unused (or mostly unused) methods in regularly implemented APIs.

However, in the interest of making a small, simple change with this PR, I think the following makes sense, at least.

```python

# garage.np.Policy

class Policy(abc.ABC):

    @abc.abstractproperty
    def env_spec(self):
        pass

    @property
    def observation_space(self):
        return self.env_spec.observation_space

    @property
    def action_space(self):
        return self.env_spec.action_space

# tests.fixtures.policies.DummyPolicy

from garage.np.policies import Policy as BasePolicy

class DummyPolicy(BasePolicy):

    def __init__(self, env_spec):
        self._env_spec = env_spec

    @property
    def env_spec(self):
        return self._env_spec
```

krzentner: Honestly, I was just trying to avoid some duplicated code with this comment. I don't really get what's hard here.

ahtsan: Currently in tf tree, observation_space and action_space are often used (e.g. to flatten / unflatten observations) so i think maybe it's good time to figure out if it should be included in the interface later when we really refactor the policy interface.

I think you proposed a good way to eliminate duplicated code specifically for the existing policy interface, although we will change it later anyway. For this PR, I think the objective is to unify policy interface across the codebase and remove duplicated code.

ryanjulian: @krzentner sorry i didn't mean we would implement the interface in this PR, just that quibbling over the details for a long time doesn't matter much if we are going to rip out much of this in the next ~2 weeks

ahtsan:
@Merifyio rebase

ahtsan:
@Mergifyio rebase

ahtsan:
@Mergifyio rebase

ahtsan:
@krzentner It's ready for review

PR #1597, opened 2020-06-22
Fix bug when calling CategoricalCNNPolicy

yeukfu:


yeukfu:
> Does this bug apply to `release-2019.10`?

I don't see the bug in `release-2019.10`.

ryanjulian:
@Mergifyio rebase

PR #1600, opened 2020-06-23
Add image observations doc

### LINKED ISSUE #1531: Docs page "Use image observations"
See #1426

maliesa96:
Closes #1531


docs/user/pixel_observations.md
```diff
@@ -0,0 +1,122 @@
...
+
+Garage includes various environment wrappers that are particularly useful when working with pixel observations. You may use multiple wrappers simultaneously, though the order in which you use them matters. See the pixel-specific wrappers section for more details on this.
+
+### GarageEnv
+When wrapping image-based environments in `GarageEnv`, `is_image` should be set to `True`:
+```python
+env = GarageEnv(gym.make("MemorizeDigits-v0"), is_image=True)
+```
+This converts the `gym.spaces.Box` observation space to an `akro.Image` space, which then:
+- Tells CNN primitives in garage to normalize pixel values to between 0 and 1. This typically decreases training time, particularly in environments where observation values vary widely. Note that this is only done internally within the primitives, so calls to `env.step()` will still return raw pixel values.
```
ahtsan, line 44: Do we store images as `int` in replay buffer and only normalize them when passing to policy, so as to save memory? If so, I think it's also worth mentioning as it is an important technique in processing spatial data.

docs/user/pixel_observations.md
```diff
@@ -0,0 +1,122 @@
...
+
+```python
+from gym.wrappers.pixel_observation import PixelObservationWrapper as gymWrapper
+from garage.envs.wrappers import PixelObservationWrapper as garageWrapper
+
+env = gym.make("Pendulum_v0")
+# env.observation_space = Box(3,)
+env.reset() # needed when using gym's wrapper
+gym_wrapped = gymWrapper(env)
+# gym_wrapped.observation_space = Dict(pixels:Box(1000, 1000, 3))
```
krzentner, line 62: May as well use actual space syntax here (i.e. `gym_wrapper.observation_space == Dict({'pixels':Box(1000, 1000, 3)})`

docs/user/pixel_observations.md
```diff
@@ -0,0 +1,122 @@
...
+env = gym.make('Env')
+# env.observation_space.shape = (100, 100, 3)
+env_wrapped = Grayscale(env)
+# env_wrapped.observation_space.shape = (100, 100)
+```
+#### StackFrames
+
+Frame stacking is a technique used in RL to create observations from which temporal data can be inferred. As a concrete example, consider the `Pendulum-v0` environment where the agent must balance the pendulum vertically by applying some amount of torque to the shaft. To do this reliably, the agent needs to know both the current position of the pendulum and the direction in which it is moving. The former is given by the current image observation, but the latter cannot be inferred from one image alone. To combat this, you can stack N (say, 3) images into one , which would embed this information into one observation.
+
+Frames are stacked on a rolling basis. When stacking N frames, the N most recent frames are returned as the observation.
```
krzentner, line 92: I think it would be good to mention that stacking increases the number of *channels* of the observation (it doesn't make the image taller or wider).

docs/user/pixel_observations.md
```diff
@@ -0,0 +1,122 @@
...
+def pixel_observations_example(ctxt=None, seed=1, buffer_size=int(1e4)):
+	set_seed(seed)
+
+	with LocalTFRunner(snapshot_config=ctxt) as runner:
+
+	env = gym.make('Pendulum-v0')
+	env = PixelObservation(env) # goes first
+	env = Grayscale(env)
+	env = Resize(env, 86, 86)
+	env = StackFrames(env, 2)
```
krzentner, line 115: You should probably mention that `StackFrames` should go after `Grayscale`.

docs/user/pixel_observations.md
```diff
@@ -0,0 +1,122 @@
...
+ Gym environments that do not return pixel observations by default can be used with `garage.envs.wrappers.PixelObservationWrapper`, which overrides the default observation and instead returns a tensor of RGB values by making calls to `env.render(mode="rgb_array")`. See the section on this wrapper below for more on this.
+
+## Environment Wrappers
+
+Garage includes various environment wrappers that are particularly useful when working with pixel observations. You may use multiple wrappers simultaneously, though the order in which you use them matters. See the pixel-specific wrappers section for more details on this.
+
+### GarageEnv
+When wrapping image-based environments in `GarageEnv`, `is_image` should be set to `True`:
+```python
+env = GarageEnv(gym.make("MemorizeDigits-v0"), is_image=True)
```
ryanjulian, line 41: we always use single quotes in garage code, so let's keep that consistent here

docs/user/pixel_observations.md
maliesa96: So we don't explicitly cast non-int values to ints for images before storing them in the buffer (though maybe we should). We assume that, if the input is an image, the environment will output ints. I'll mention that the user should make sure this is the case though, for the reasons you mentioned.

docs/user/pixel_observations.md
maliesa96: Fixed

docs/user/pixel_observations.md
maliesa96: Good call, done.

maliesa96:
Ready for another review. I added a section on using matplotlib to visualize observations after using the wrappers.

maliesa96:
@Mergifyio rebase

maliesa96:
@mergifyio rebase

ryanjulian:
@Mergifyio rebase

ryanjulian:
@Mergifyio rebase

PR #1603, opened 2020-06-23
Export $USER in Dockerfile

gitanshu:
docker doesn't export $USER by default. Exporting it
here so that any image using garage as parent, can
make use of it.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -71,6 +71,7 @@ ARG uid=999
 RUN groupadd -g $uid $user && \
     useradd -m -r -u $uid -g $user $user && \
     chown -R $user:$user /home/$user
+ENV USER $user
 USER $user
 ENV HOME /home/$user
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -3,10 +3,6 @@ ARG PARENT_IMAGE=ubuntu:18.04
 # Garage base target ###########################################################
 FROM $PARENT_IMAGE AS garage-18.04
 
-# http://bugs.python.org/issue19846
-# > At the moment, setting "LANG=C" on a Linux system *fundamentally breaks Python 3*, and that's not OK.
-ENV LANG C.UTF-8
```
ryanjulian, line 8: hmm probably keep this until your dockerfile doesn't use Python 3.5.

gitanshu: Oh right. I'll do this in the python 3.5 removal PR only then.

gitanshu:
@Mergifyio rebase

PR #1604, opened 2020-06-23
Add documentation for benchmarking

yeukfu:
Documentation for benchmarking, [issue #1546](https://github.com/rlworkgroup/garage/issues/1546), based on previous [readme](https://github.com/rlworkgroup/garage/blob/master/benchmarks/README.md) in benchmarks.

### DIFF
--- a/docs/user/benchmarking.md
+++ b/docs/user/benchmarking.md
@@ -4,7 +4,9 @@ This guide explains how to write and run garage benchmark scripts.
 
 ## Prerequisites
 
-Firstly, you should have installed garage. If not, you can see instructions in the [latest documentation](https://garage.readthedocs.io/en/latest/user/installation.html) to install garage.
+Firstly, you should have installed garage. If not, you can see instructions in
+the [latest documentation](https://garage.readthedocs.io/en/latest/user/installation.html)
+to install garage.
 
 ## Install `garage-benchmarks`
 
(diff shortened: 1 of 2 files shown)


docs/user/benchmarking.md
```diff
@@ -0,0 +1,89 @@
...
+
+This guide explains how to write and run garage benchmark scripts.
+
+## Prerequisites
+
+Firstly, you should have installed garage. If not, you can see instructions in the [latest documentation](https://garage.readthedocs.io/en/latest/user/installation.html) to install garage.
+
+Install `google-cloud-storage`:
+
+`pip install google-cloud-storage`
```
zequnyu, line 11: This has been [covered](https://github.com/rlworkgroup/garage/blob/19f47bba3baf9575d2cce3aa9d5d8af102fea50b/benchmarks/setup.py#L14) in the next step.

yeukfu: Got it. I will remove it.

docs/user/benchmarking.md
```diff
@@ -0,0 +1,89 @@
...
+See more [examples](https://github.com/rlworkgroup/garage/tree/master/benchmarks/src/garage_benchmarks/experiments).
+```py
+@wrap_experiment
+def trpo_garage_pytorch(ctxt, env_id, seed):
+    ...
+```
+
+**Define an execution function using `@benchmark` and `iterate_experiments()`**
+
+Plotting is enabled by default. You can explicitly disable it by defining `@benchmark(plot=False)`. Automatic benchmarking is diabled by default. You can explicitly disable it by defining `@benchmark(auto=True)`.
```
gitanshu, line 33:
```suggestion
Plotting is enabled by default. You can explicitly disable it by defining `@benchmark(plot=False)`. Automatic benchmarking is disabled by default. You can explicitly enable it by defining `@benchmark(auto=True)`.
```

Also, I guess this demands explaining what Automatic benchmarking is and why someone would want to enable it. Since these are user docs, maybe we don't need to mention automatic benchmarking here at all, but I am not sure about that opinion.

docs/user/benchmarking.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Install `garage-benchmarks`
+
+`cd garage/benchmarks && pip install -e .`
+
+## Write garage benchmark scripts
+
+**Define an experiment function**
+
+Use `@wrap_experiment` to define your experiment. `ctxt` is used for `@wrap_experiment`.
```
gitanshu, line 21: Maybe this should link to the documentation of using wrap_experiment?
Perhaps https://garage.readthedocs.io/en/stable/user/experiments.html#running-experiments-on-gpu-cpu ?

docs/user/benchmarking.md
yeukfu: Thanks for pointing out this. I will remove Automatic benchmarking.

docs/user/benchmarking.md
yeukfu: Agree. Will do.

ryanjulian:
@Mergifyio rebase

ryanjulian:
I think you need to update `docs/index.rst` so that it appears in the table of contents.

ryanjulian:
Don't forget to add an acknowledgement for yourself at the bottom of the page!  See https://github.com/rlworkgroup/garage/pull/1638 for an example.

ryanjulian:
@Mergifyio refresh

PR #1605, opened 2020-06-23
Torch discrete CNN policy

irisliucy:
The main goal of this PR is to add discrete CNN policy for #1273 

It includes:
- [x] `modules/cnn_module`
- [x]  `modules/categorical_cnn_module`
- [x]  `policies/categorical_cnn_policy`
- [x] `policies/discrete_policy`
- [x] tests for the above modules
  - [x]  `test_cnn_module`
  - [x]  `test_categorical_cnn_module`
  - [x]  `test_categorical_cnn_policy`


src/garage/torch/modules/categorical_cnn_module.py
```diff
@@ -0,0 +1,152 @@
...
+            *inputs: Input to the module.
+
+        Returns:
+            torch.distributions.Categorical: Policy distribution.
+
+
+        """
+        cnn_output = self._cnn_module(*inputs)
+        mlp_output = self._mlp_model(cnn_output)
+        if not isinstance(mlp_output, Categorical):
```
ryanjulian, line 150: hmm this seems hacky -- why would an MLPModel output a categorical?

src/garage/torch/modules/cnn_module.py
```diff
@@ -0,0 +1,355 @@
+"""CNN Module."""
+import copy
+
+from torch import nn
+
+
+# pylint: disable=unused-argument
+class CNNBaseModule(nn.Module):
```
ryanjulian, line 8: why not just CNNModule?

src/garage/torch/modules/cnn_module.py
```diff
@@ -0,0 +1,355 @@
...
+            self._check_valid_padding(paddings, kernel_sizes)
+        else:
+            padding = 0
+
+        in_channel = input_var.shape[1]  # read in N, C, H, W
+        prev_channel = in_channel
+        for index, (channel, kernel_size, stride) in enumerate(
+                zip(hidden_channels, kernel_sizes, strides)):
+
+            hidden_layers = nn.Sequential()
```
ryanjulian, line 83: maybe you can define this as a _CNNCell helper?

src/garage/torch/modules/cnn_module.py
```diff
@@ -0,0 +1,355 @@
...
+                     kernel_size=kernel_size,
+                     stride=stride,
+                     padding=padding,
+                     padding_mode=padding_mode,
+                     dilation=dilation,
+                     bias=bias)
+
+
+# pylint: disable=missing-return-type-doc
+def flatten(x):
```
ryanjulian, line 226: please put this in garage.torch._functions (it will need a more descriptive name)

src/garage/torch/policies/categorical_cnn_policy.py
```diff
@@ -0,0 +1,138 @@
...
+class CategoricalCNNPolicy(StochasticPolicy):
+    """CategoricalCNNPolicy.
+
+    A policy that contains a CNN and a MLP to make prediction based on
+    a categorical distribution.
+
+    It only works with akro.Discrete action space.
+
+    Args:
+        env_spec (garage.envs.env_spec.EnvSpec): Environment specification.
```
ryanjulian, line 19: is this actually used?

src/garage/torch/policies/deterministic_mlp_policy.py
```diff
@@ -36,51 +32,42 @@ def __init__(self, env_spec, name='DeterministicMLPPolicy', **kwargs):
...
         """Get a single action given an observation.
-
         Args:
             observation (np.ndarray): Observation from the environment.
-
         Returns:
             tuple:
                 * np.ndarray: Predicted action.
                 * dict:
                     * list[float]: Mean of the distribution
```
ryanjulian, line 51: i think this is actually an np.ndarray (in both cases)

src/garage/torch/policies/deterministic_mlp_policy.py
```diff
@@ -36,51 +32,42 @@ def __init__(self, env_spec, name='DeterministicMLPPolicy', **kwargs):
...
         """Get actions given observations.
-
         Args:
             observations (np.ndarray): Observations from the environment.
-
         Returns:
             tuple:
                 * np.ndarray: Predicted actions.
                 * dict:
                     * list[float]: Mean of the distribution
```
ryanjulian, line 67: isn't this an np.ndarray in both cases?

src/garage/torch/modules/cnn_module.py
```diff
@@ -0,0 +1,276 @@
...
+            of intermediate dense layer(s). The function should return a
+            torch.Tensor.
+        hidden_b_init (callable): Initializer function for the bias
+            of intermediate dense layer(s). The function should return a
+            torch.Tensor.
+        paddings (tuple[int]):  Zero-padding added to both sides of the input
+        padding_mode (str): The type of padding algorithm to use,
+            either 'SAME' or 'VALID'.
+        max_pool (bool): Bool for using max-pooling or not.
+        pool_shapes (tuple[int]): Dimension of the pooling layer(s). For
```
naeioi, line 41: Reading from the implementation, I think this argument should be renamed to `pool_shape` - You should mention that all pooling layers are of the same shape.

src/garage/torch/modules/cnn_module.py
```diff
@@ -0,0 +1,276 @@
...
+            example, (2, 2) means that all the pooling layers have
+            shape (2, 2).
+        pool_strides (tuple[int]): The strides of the pooling layer(s). For
```
naeioi, line 44: `pool_stride`

src/garage/torch/modules/cnn_module.py
```diff
@@ -0,0 +1,276 @@
...
+            filter_sizes (tuple[int]): filter sizes
+
+        Raises:
+            ValueError: if the padding size is not equal to the expected shape
+
+        """
+        for index, (padding,
+                    filter_size) in enumerate(zip(paddings, filter_sizes)):
+            if padding != 0:  # only check when padding value is not 0
+                expected_padding = int((filter_size - 1) / 2)  # P = (F - 1) /2
```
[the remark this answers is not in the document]
naeioi: Why does padding and kernel size need to follow this restriction anyway?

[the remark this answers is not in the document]
irisliucy: My initial idea is to set zero padding to be P=(F−1)/2 so that when the padding mode is `zero`, it will check whether the padding size fits this formula. 

But it is only useful in a marginal case when stride=1. Now I think I'll remove this function.

irisliucy:
@Mergifyio rebase

irisliucy:
@Mergifyio rebase

src/garage/torch/_functions.py
```diff
@@ -187,6 +188,23 @@ def flatten_batch(tensor):
     return tensor.reshape((-1, ) + tensor.shape[2:])
 
 
+# pylint: disable=missing-return-type-doc
```
ryanjulian, line 191: i think you are getting this message because you didn't specify a return type.

isn't it just

```python
"""
...

Returns:
    torch.Tensor: Reshaped view of that data (analogous to numpy.reshape)
"""
```

src/garage/torch/_functions.py
```diff
@@ -273,3 +291,36 @@ def product_of_gaussians(mus, sigmas_squared):
     sigma_squared = 1. / torch.sum(torch.reciprocal(sigmas_squared), dim=0)
     mu = sigma_squared * torch.sum(mus / sigmas_squared, dim=0)
     return mu, sigma_squared
+
+
+class TransposeImage(gym.ObservationWrapper):
+    """Transpose observation space for image observation in PyTorch."""
```
ryanjulian, line 297: i dont think a wrapper is the best way to approach this, but i might be wrong.

why not just transpose all of the observations as you receive them? what is the cost of doing this in the differentiable path, vs during sampling (what you do here)?

src/garage/torch/policies/__init__.py
```diff
@@ -1,4 +1,5 @@
 """PyTorch Policies."""
+from garage.torch.policies.categorical_cnn_policy import (CategoricalCNNPolicy)
```
ryanjulian, line 2: is there a reason this is in parens?

src/garage/torch/policies/discrete_policy.py
```diff
@@ -0,0 +1,70 @@
...
+import abc
+
+import akro
+import torch
+
+from garage.torch import global_device
+from garage.torch.policies.policy import Policy
+
+
+class DiscretePolicy(Policy, abc.ABC):
```
ryanjulian, line 11: hmm is there a reason we need a DiscretePolicy API? how is this different from a policy?

tests/garage/tf/models/test_categorical_cnn_model.py
```diff
@@ -31,6 +31,7 @@ def test_dist(self):
                                     strides=(1, ),
                                     padding='VALID')
         dist = model.build(self._input_ph).dist
+        assert dist is None
```
ryanjulian, line 34: hmm is this debugging you forgot to remove?

tests/garage/torch/policies/test_categorical_cnn_policy.py
```diff
@@ -0,0 +1,121 @@
...
+import cloudpickle
+import pytest
+
+from garage.envs import GarageEnv
+from garage.torch._functions import TransposeImage
+from garage.torch.policies import CategoricalCNNPolicy
+from tests.fixtures.envs.dummy import DummyDiscretePixelEnv
+
+
+class TestCategoricalCNNPolicies:
```
ryanjulian, line 11: it seems this only tests CategoricalCNNPolicy -- so `TestCategoricalCNNPolicy`?

tests/garage/torch/modules/test_cnn_module.py
```diff
@@ -0,0 +1,183 @@
...
+        current_size = self.input_width
+        for (filter_size, stride, padding) in zip(kernel_sizes, strides,
+                                                  paddings):
+            # padding = float((filter_size - 1) / 2) # P = (F - 1) /2
+            current_size = int(
+                (current_size - filter_size + padding * 2) /
+                stride) + 1  # conv formula = ((W - F + 2P) / S) + 1
+        flatten_shape = current_size * current_size * hidden_channels[-1]
+
+        assert output.shape == (self.batch_size, flatten_shape)
```
ryanjulian, line 117: can you test the value here in addition to the shape? did the CNN produce the correct output?

tests/garage/torch/modules/test_cnn_module.py
```diff
@@ -0,0 +1,183 @@
...
+        current_size = self.input_width
+        for (filter_size, stride, padding) in zip(kernel_sizes, strides,
+                                                  paddings):
+            # padding = float((filter_size - 1) / 2) # P = (F - 1) /2
+            current_size = int(
+                (current_size - filter_size + padding * 2) /
+                stride) + 1  # conv formula = ((W - F + 2P) / S) + 1
+        flatten_shape = current_size * current_size * hidden_channels[-1]
+
+        assert output1.shape == (self.batch_size, flatten_shape)
```
ryanjulian, line 80: can you test values here, and not just shapes? see the TF tree tests for some examples you can probably copy.

tests/garage/torch/test_functions.py
```diff
@@ -67,6 +70,30 @@ def test_product_of_gaussians():
     assert output[1] == 1 / size
 
 
+def test_flatten_to_single_vector():
+    """ Test `flatten` function
```
ryanjulian, line 74: none of these docstrings should be required (and really aren't encouraged, tbh)

src/garage/torch/_functions.py
```diff
@@ -273,3 +291,36 @@ def product_of_gaussians(mus, sigmas_squared):
...
+    def __init__(self, env=None):
+        """Transpose image observation space.
+
+        Reshape the input observation shape from (H, W, C) into (C, H, W)
+        in pytorch format.
+
+        Args:
+            env (garage.envs): environment
+        """
+        super(TransposeImage, self).__init__(env)
```
krzentner, line 308: You don't actually need to pass any parameters to `super()` in our versions of python.

src/garage/torch/modules/cnn_module.py
```diff
@@ -0,0 +1,252 @@
...
+                     out_channels=out_channels,
+                     kernel_size=kernel_size,
+                     stride=stride,
+                     padding=padding,
+                     padding_mode=padding_mode,
+                     dilation=dilation,
+                     bias=bias)
+
+
+class _NonLinearity(nn.Module):
```
krzentner, line 217: This should probably go in `garage.torch._functions`, since it's definitely wanted elsewhere.

src/garage/torch/policies/discrete_policy.py
krzentner: Given the contents of the methods here, I think this should actually inherit from `StochasticPolicy`.

krzentner: I definitely think this should just be `StochasticPolicy`, the dictionary should just be empty.

src/garage/torch/_functions.py
irisliucy: The reason for `TransposeImage` is to transform the gym `observation_space`, and not just the observation itself so that `observation_space.shape` and `observation` is consistent. And it is useful for manipulating dimension and observation input from `get_action` / `get_actions` in Pytorch format. 

And you're right, transforming the observation space is expensive if we do it rapidly. But in fact, we just need to perform once at the beginning when I get the `env`.

irisliucy: If I don't transform the observation space, I need to (remember to) transpose the `observation` / `shape` every time in pytorch.

irisliucy:
@Merifyio rebase

src/garage/torch/policies/stochastic_policy.py
```diff
@@ -25,12 +26,13 @@ def get_action(self, observation):
                     * np.ndarray[float]: Mean of the distribution
                     * np.ndarray[float]: Standard deviation of logarithmic
                         values of the distribution.
-
         """
         with torch.no_grad():
             if not isinstance(observation, torch.Tensor):
                 observation = torch.as_tensor(observation).float().to(
                     global_device())
+            if isinstance(self._env_spec.observation_space, akro.Image):
```
krzentner, line 34: This is kinda a really weird place to put this logic. I would expect it to be in `CNNModule`, since it's already essentially image-specific. It's not a big issue either way though.

src/garage/torch/modules/cnn_module.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+# pylint: disable=unused-argument
+class CNNModule(nn.Module):
+    """Convolutional neural network (CNN) model in pytorch.
+
+    Args:
+        input_var (pytorch.tensor): Input tensor of the model.
+            Based on 'NCHW' data format: [batch_size, channel, height, width].
+        kernel_sizes (tuple[int]): Dimension of the conv filters.
```
yeukfu, line 16: It will be better for me if the `kernel_sizes` is consistent to the [TF tree](https://github.com/rlworkgroup/garage/blob/e2c890d404eb8c40e90b480d2cb931cb21dbea74/src/garage/tf/models/cnn.py#L24).

irisliucy:
@Merifyio rebase

PR #1606, opened 2020-06-23
Add TensorBoard logging doc

### LINKED ISSUE #1536: Docs page "Monitor your experiments with TensorBoard"
See #1426

AiRuiChen:
Closes #1536


docs/user/monitor_experiments_with_tensorboard.md
```diff
@@ -0,0 +1,178 @@
...
+
+To start using dowel, import `logger`, dowel's logging facility:
+
+`from dowel import logger`
+
+`logger` takes in many different types of input and directs them to the correct output.
+
+Let's start by adding some outputs to logger:
+
+```bash
```
ryanjulian, line 22: so this is done for you if you use `@wrap_experiment` -- no need to recapitulate the dowel documentation.

You might want to cover adding your own output inside an experiment function though (usage is the same).

docs/user/monitor_experiments_with_tensorboard.md
```diff
@@ -0,0 +1,178 @@
...
+
+You can find more TensorBoard dashboard details [here](https://www.tensorflow.org/tensorboard/get_started).
+
+## Share Your TensorBoard Logs
+
+[TensorBoard Dev](https://tensorboard.dev/) allows you to share your TensorBoard results with others.
+
+Upload your logs by:
+```bash
+$ tensorboard dev upload --logdir logs \
```
ryanjulian, line 162: probably use an example command here which is likely to work with one of the garage examples (e.g. logs are sent to `data/local`

docs/user/monitor_experiments_with_tensorboard.md
```diff
@@ -0,0 +1,178 @@
...
+
+```bash
+@wrap_experiment(log_dir='my_dir')
+def my_experiment(ctxt, seed):
+```
+For a complete list of arguments, you can refer to the definition of `wrap_experiment` [here](https://github.com/rlworkgroup/garage/blob/master/src/garage/experiment/experiment.py).
+
+Congrats! Now you have finished setting up TensorBoard logging. You can run your experiment and find your TensorBoard logs under the log directory.
+
+By default, your `log_dir` is `garage/data/local/experiment/{experiment_name}_{experiment_id}`
```
ryanjulian, line 131: logs actually go to `$(pwd)/data/local` and remember that most users are not downloading the repo (they install from pip), so the prefix `garage/` makes no sense.

docs/user/monitor_experiments_with_tensorboard.md
```diff
@@ -0,0 +1,178 @@
...
+    Args:
+        ctxt (garage.experiment.ExperimentContext): The experiment
+            configuration used by LocalRunner to create the snapshotter.
+    """
+```
+`@wrap_experiment` helps create experiment log directories and set up logger outputs (e.g. `logger.add_output(dowel.TensorBoardOutput(log_dir))`). After the setup process, `my_experiment()` is called to start the actual experiment.
+
+Note that `my_experiment()` now receives a new input `ctxt`. This is required for  the wrapped function to receive experiment configuration created by the wrapper.
+
+This decorater can also be invoked **with arguments**:
```
ryanjulian, line 121: A lot of this is probably out-of-scope for  this doc and should be forked into its own doc about `wrap_experiment`, but if you're going to mention it, you should say that the arguments are intended to be your experiment's hyperparameters, and the hparams your experiment is invoked with will be automatically saved into that experiment's log directory.

docs/user/monitor_experiments_with_tensorboard.md
```diff
@@ -0,0 +1,178 @@
...
+
+Finally, log your scalar values with `logger`:
+
+`logger.log(tabular)`
+
+### Log **histograms** to Tensorboard
+
+To log histograms to TensorBoard, you would need to pass [`dowel.Histogram`](https://github.com/rlworkgroup/dowel/blob/master/src/dowel/histogram.py) to `tabular`.
+
+`dowel.Histogram` is implemented as a typed view of a numpy array. It will accept input that `numpy.asarray` will.
```
ryanjulian, line 80: you can also log distribution types from `scipy.stats`, I believe.

docs/user/monitor_experiments_with_tensorboard.md
```diff
@@ -0,0 +1,178 @@
...
+
+[TensorBoard](https://www.tensorflow.org/tensorboard) is a powerful tool that allows you to monitor and visualize an experiment. This section introduces how to set up TensorBoard logging in garage, and how to view/share your TensorBoard logs.
+
+## Prerequisite
+The TensorBoard package should be installed upon garage's installation. It can be also manually installed via:
+```bash
+$ pip install tensorboard
+```
+
+## Setup Your Experiment with `@wrap_experiment`
```
ryanjulian, line 11: i think you can just link out to the other documentation on how to use wrap_experiment. this doc can be very short, in other words

ryanjulian:
I think you need to update `docs/index.rst` so this appears in the table of contents

AiRuiChen:
> I think you need to update `docs/index.rst` so this appears in the table of contents

Please see my latest commit. I also added docs that are newly created. 

Let me know if this isn't what you mean;)

ryanjulian:
Don't forget to add an acknowledgement for yourself at the bottom of the page!  See https://github.com/rlworkgroup/garage/pull/1638 for an example.

ryanjulian:
@AiRuiChen i changed up how the TOC works so I think you will need to update this, despite your new commit.

I apologize for the back-and-forth. This is the first time we're really getting serious :).

docs/user/monitor_experiments_with_tensorboard.md
```diff
@@ -0,0 +1,132 @@
...
+pip install tensorboard
+```
+
+### Setup Your Experiment
+
+Use `@wrap_experiment` to setup your experiment.
+This will automatically setup TensorBoard output and allocate an experiment
+directory for you.
+
+Find more about how to use `@wrap_experiment` [here](https://github.com/rlworkgroup/garage/blob/master/docs/user/experiments.rst).
```
yeukfu, line 23: Relative link could also be used here. For example, `[here](experiments)`.

AiRuiChen: Got it. Thank you!

yeukfu: It seems that `[here](experiments.rst)` doesn't work when html generated. `[here](experiments)` works for me. You can see docs/readthedocs.org:garage in the check to preview it.

AiRuiChen: Changed to `[here](experiments)` and it works now!

Issue #380, opened 2018-11-02
torch/BC


ryanjulian:
Behavioral cloning in torch and associated primitives for imitation learning

PR #1607, opened 2020-06-23
Behavioral Cloning

krzentner:
Fixes / Implements #380

### DIFF
--- a/examples/torch/bc_point.py
+++ b/examples/torch/bc_point.py
@@ -1,0 +1,94 @@
+#!/usr/bin/env python3
+"""Example of using Behavioral Cloning."""
+import click
+import numpy as np
+
+from garage import wrap_experiment
+from garage.envs import GarageEnv, PointEnv
+from garage.experiment import LocalRunner
+from garage.torch.algos import BC
+from garage.torch.policies import GaussianMLPPolicy, Policy
+
+
+class OptimalPolicy(Policy):
+    """Optimal policy for PointEnv.
+
+    Args:
+        env_spec (EnvSpec): The environment spec.
+        goal (np.ndarray): The goal location of the environment.
+
+    """
+
+    # No forward method
+    # pylint: disable=abstract-method
+
+    def __init__(self, env_spec, goal):
+        super().__init__(env_spec, 'OptimalPolicy')
+        self.goal = goal
+
+    def get_action(self, observation):
+        """Get action given observation.
+
+        Args:
+            observation (np.ndarray): Observation from PointEnv. Should have
+                length at least 2.
+
+        Returns:
+            tuple:
+                * np.ndarray: Optimal action in the environment. Has length 2.
+                * dict[str, np.ndarray]: Agent info (empty).
+
+        """
+        return self.goal - observation[:2], {}
+
+    def get_actions(self, observations):
+        """Get actions given observations.
+
+        Args:
+            observations (np.ndarray): Observations from the environment.
+                Has shape :math:`(B, O)`, where :math:`B` is the batch
+                dimension and :math:`O` is the observation dimensionality (at
+                least 2).
+
+        Returns:
+            tuple:
+                * np.ndarray: Batch of optimal actions.
+                    Has shape :math:`(B, 2)`, where :math:`B` is the batch
+                    dimension.
+                Optimal action in the environment.
+                * dict[str, np.ndarray]: Agent info (empty).
+
+        """
+        return (self.goal[np.newaxis, :].repeat(len(observations), axis=0) -
+                observations[:, :2]), {}
+
+
+@click.command()
+@click.option('--loss', type=str, default='log_prob')
+@wrap_experiment
+def bc_point(ctxt=None, loss='log_prob'):
+    """Run Behavioral Cloning on garage.envs.PointEnv.
+
+    Args:
+        ctxt (ExperimentContext): Provided by wrap_experiment.
+        loss (str): Either 'log_prob' or 'mse'
+
+    """
+    runner = LocalRunner(ctxt)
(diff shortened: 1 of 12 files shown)


src/garage/torch/algos/bc.py
```diff
@@ -0,0 +1,190 @@
...
+    """Behavioral Cloning.
+
+    Based on Model-Free Imitation Learning with Policy Optimization:
+        https://arxiv.org/abs/1605.08478
+
+    Args:
+        env_spec (garage.envs.EnvSpec): Specification of environment.
+        learner (garage.torch.Policy): Policy to train.
+        batch_size (int): Size of optimization batch.
+        source (garage.Policy or Generator[garage.TimeStepBatch]): Expert to
```
ryanjulian, line 27: out-of-scope for this PR: what do you think about defining a Dataset API, which would unify things like this (in addition to accessing the replay buffer, or on-policy sampling).

basically `TimeStepDataset = Generator[garage.TimeStepBatch]` and `TrajectoryDataset = Generator[garage.TrajectoryBatch]`

alternatively, do this but use the torch and tf first-party APIs

ryanjulian:
You LOVE to see it!

src/garage/torch/algos/bc.py
krzentner: Yeah, I'm not sure if the first party APIs are best to use or not at this point. It seems unlikely we'd go wrong using a more restrictive API for the mean time.

examples/torch/bc_point.py
```diff
@@ -0,0 +1,86 @@
...
+        Args:
+            observation (np.ndarray): Observation from the environment.
+
+        Returns:
+            tuple:
+                * np.ndarray: Action.
+                * dict[str, np.ndarray]: Agent info (empty).
+
+        """
+        return self.goal - observation[:2], {}
```
ahtsan, line 40: Does this assume observation's dimension is [batch, time, *dims]? I think we should mention it here.

examples/torch/bc_point.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def __init__(self, env_spec, goal):
+        super().__init__(env_spec, 'OptimalPolicy')
+        self.goal = goal
+
+    def get_action(self, observation):
+        """Get action given observation.
+
+        Args:
+            observation (np.ndarray): Observation from the environment.
```
avnishn, line 32: can you document this shape please

examples/torch/bc_point.py
```diff
@@ -0,0 +1,86 @@
...
+                * dict[str, np.ndarray]: Agent info (empty).
+
+        """
+        return self.goal - observation[:2], {}
+
+    def get_actions(self, observations):
+        """Get actions given observations.
+
+        Args:
+            observations (np.ndarray): Observations from the environment.
```
avnishn, line 46: can you document this shape?

avnishn: I think the variables to be used to describe the shape are outlined in a table under garage/_dtypes.py

examples/torch/bc_point_deterministic_policy.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def __init__(self, env_spec, goal):
+        super().__init__(env_spec, 'OptimalPolicy')
+        self.goal = goal
+
+    def get_action(self, observation):
+        """Get action given observation.
+
+        Args:
+            observation (np.ndarray): Observation from the environment.
```
avnishn, line 31: please document shapes. Variables to describe the shape can be found in a table under garage/_dtypes.py.

examples/torch/bc_point_deterministic_policy.py
```diff
@@ -0,0 +1,82 @@
...
+
+        Returns:
+            tuple:
+                * np.ndarray: Action.
+                * dict[str, np.ndarray]: Agent info (empty).
+
+        """
+        return self.goal - observation[:2], {}
+
+    def get_actions(self, observations):
```
avnishn, line 41: please document shapes. Variables to describe the shape can be found in a table under garage/_dtypes.py

src/garage/_dtypes.py
```diff
@@ -919,3 +921,28 @@ def from_time_step_list(cls, env_spec, ts_samples):
         ]
 
         return TimeStepBatch.concatenate(*ts_batches)
+
+    @classmethod
+    def from_trajectory_batch(cls, batch):
+        """Construct a TimeStepBatch from a TrajectoryBatch.
+
+        Args:
+            batch (TrajectoryBatch): TrajectoryBatch to convert.
```
avnishn, line 930: please document shapes

examples/torch/bc_point.py
```diff
@@ -0,0 +1,86 @@
...
+import numpy as np
+
+from garage import wrap_experiment
+from garage.envs import GarageEnv, PointEnv
+from garage.experiment import LocalRunner
+from garage.torch.algos import BC
+from garage.torch.policies import GaussianMLPPolicy, Policy
+
+
+class OptimalPolicy(Policy):
```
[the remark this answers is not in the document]
ryanjulian: examples is not on the python path (all files are self-contained scripts), so no doesn't work. also having a library of utilities within examples is just strange. if it's so useful, why isn't it in garage?

i think this solution is fine.

examples/torch/bc_point.py
ryanjulian: alternatively, the entire body of this docstring can just refer to the policy interface.

src/garage/_dtypes.py
ryanjulian: what shapes? the dtype is well-defined. that's why we defined it! to stop typing shapes all the time.

PR #1608, opened 2020-06-23
Don't copy the eval env every epoch

krzentner:
The behavioral cloning PR is on top of this one.


krzentner:
It's a pretty easy backport, so we probably should.

ryanjulian:
@Mergifyio rebase

PR #1609, opened 2020-06-24
Update CHANGELOG.md

ahtsan:


ahtsan:
Should I also make this change in `master`?

ryanjulian:
Yeah, I generally try to keep the changelog in master updated for all the releases, even minor ones.

CHANGELOG.md
```diff
@@ -4,6 +4,26 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 2019.10.2
+
+### Added
```
ryanjulian, line 9: This looks great.

Changed/Added/Fixed all refer to features and APIs (which are frozen for minor releases), so these should all go under "Fixed" (because they were committed to fix some problem)

PR #1614, opened 2020-06-24
tf.Policy inherit from tf.Model

ahtsan:
This is a proposed change that would make it easier to implement a new policy.

Basically, tf.Model inherits from tf.Module and tf.Policy inherits from tf.Model. This PR
also removed usage of `vectorized`, which is no longer needed.

### DIFF
--- a/src/garage/tf/models/model.py
+++ b/src/garage/tf/models/model.py
@@ -5,6 +5,8 @@ import warnings
 
 import tensorflow as tf
 
+from garage.tf.models.module import Module
+
 
 class BaseModel(abc.ABC):
     """Interface-only abstract class for models.
@@ -130,7 +132,7 @@ class Network:
         return self._outputs
 
 
-class Model(BaseModel):
+class Model(BaseModel, Module):
     r"""Model class for TensorFlow.
 
     A TfModel only contains the structure/configuration of the underlying
@@ -180,14 +182,14 @@ class Model(BaseModel):
     """
 
     def __init__(self, name):
-        super().__init__()
-        self._name = name or type(self).__name__  # name default to class
+        super().__init__(name or type(self).__name__)
+        # self._name = name or type(self).__name__  # name default to class
         self._networks = {}
         self._default_parameters = None
         self._variable_scope = None
 
     # pylint: disable=protected-access, assignment-from-no-return
-    def build(self, *inputs, name=None):
+    def _build_network(self, *inputs, name=None):
         """Build a Network with the given input(s).
 
         ***
@@ -223,7 +225,7 @@ class Model(BaseModel):
                 with tf.name_scope(name=network_name):
                     network = Network()
                     network._inputs = inputs
-                    network._outputs = self._build(*inputs, name)
+                    network._outputs = self._build_model(*inputs, name)
                 variables = self._get_variables().values()
                 tf.compat.v1.get_default_session().run(
                     tf.compat.v1.variables_initializer(variables))
@@ -239,7 +241,7 @@ class Model(BaseModel):
                 with tf.name_scope(name=network_name):
                     network = Network()
                     network._inputs = inputs
-                    network._outputs = self._build(*inputs, name)
+                    network._outputs = self._build_model(*inputs, name)
         custom_in_spec = self.network_input_spec()
         custom_out_spec = self.network_output_spec()
         in_spec = ['input', 'inputs']
@@ -275,10 +277,10 @@ class Model(BaseModel):
 
         return out_network
 
-    def _build(self, *inputs, name=None):
-        """Output of the model given input placeholder(s).
+    def _build_model(self, *inputs, name=None):
+        """Build this model given input placeholder(s).
 
-        User should implement _build() inside their subclassed model,
+        User should implement _build_model() inside their subclassed model,
         and construct the computation graphs in this function.
 
         Args:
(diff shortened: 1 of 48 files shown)


src/garage/tf/models/model.py
```diff
@@ -180,14 +182,14 @@ class Model(BaseModel):
     """
 
     def __init__(self, name):
-        super().__init__()
-        self._name = name or type(self).__name__  # name default to class
+        super().__init__(name or type(self).__name__)
+        # self._name = name or type(self).__name__  # name default to class
```
ryanjulian, line 186: did you forget to remove?

src/garage/tf/policies/categorical_mlp_policy.py
```diff
@@ -12,7 +12,7 @@
 from garage.tf.policies.policy import StochasticPolicy
 
 
-class CategoricalMLPPolicy(StochasticPolicy):
+class CategoricalMLPPolicy(CategoricalMLPModel, StochasticPolicy):
```
ryanjulian, line 15: what purpose does inheriting from StochasticPolicy serve here?

ahtsan: Complying with the Policy interface and indicating this policy has a `distribution` (from `StochasticPolicy`)

src/garage/tf/models/model.py
ahtsan: good catch, will remove

src/garage/tf/policies/categorical_mlp_policy.py
ryanjulian: Maybe instead we should move this up into StochasticModel? Then `class CategoricalMLPModel(StochasticModel)` and `class CategoricalMLPPolicy(CategorialMLPModel)`?

I think this is close to what @krzentner is trying to do in the torch tree, so you *only* have to implement the model to make a policy less boilerplate)

ahtsan: Totally agree, will fix

ahtsan: `CategoricalMLPModel` now inherits from `MLPModel`. In tf tree, it feels like appending `Stochastic` to a pure interface (which is `tf.Policy`) is simpler?

ahtsan: Otherwise, we will have something like `CategoricalMLPModel(MLPModel, Stochastic)`

ahtsan: or we can have another abc “StochasticMLPModel”

ahtsan: After a second thought, I think we could get rid of `StochasticPolicy` and cache the distribution returned when calling `model.build()`. Eventually, a model does not own a distribution object -- it returns one when you call `.build()`/`.forward()`

ryanjulian: it will probably take multiple passes to make this work really well. for now, is suggest going with whatever gets you a simple linear inheritance pattern. i think the one i propose above (StochasticModel/CategoricalMLPModel) leads to the least number of classes in the interim. if an algorithm wants to know if something is stochastic, it can still ask `isinstance(policy, StochasticModel)` which is IMO a step forward, because the policy is now a model.

i would very much like to get rid of StochasticPolicy.

multiple inheritance (i.e. `CategoricalMLPModel(MLPModel, Stochastic)`) can get really hard to understand in python. the pattern you are describing is called the mixins pattern, but we've chosen not to use it because it makes it difficult for users to extend existing primitives, and also difficult to test.

in the torch tree, KR is working towards a future where to define a new Policy/QFunction/etc. you *only* have to define the model (i.e. `__init__` and `torch.nn.Module.forward()`, see #1528 ). I think we should try to start moving in that direction herein TF-land, too.

ahtsan: In order to move towards `build-by-call` and only have to define a model for a policy, we shouldn't give policies/model default attributes, i.e. distribution. Then we can safely remove `StochasticPolicy` and when someone wants access to the distribution, they should call `.build()` and cache the network returned.

ryanjulian: Great point!

ryanjulian: So we should aim to remove stochastic entirely?

ahtsan: I was thinking to abandon the use of `.distribution` at all. When someone wants a symbolic distribution, they can get it by calling `.build()` and cache the return.

ryanjulian: Yes, I agree! I'm just pointing out that means deterministic policies will have a different return signature from stochastic ones.

ahtsan:
@Mergifyio rebase

ahtsan:
@ryanjulian @krzentner Ready for review

ahtsan:
> I don't really understand why this design is advantageous to the way things are now, but I trust Anson to know what's best for this part of the code.

Basically here's what I visualize: In tf tree, all primitives eventually become a `tf.Model`, i.e. policies, q-functions and baselines. Once we are able to figure out the proper interface/abstract class for each, we would be able to only implement a model for a new primitive.

PR #1617, opened 2020-06-25
Fix a typo to allow evaluating algos deterministically

maciejwolczyk:
When running some experiments with SAC I have discovered that my algorithm does not act deterministically during the evaluation (i.e. the action is sampled from the policy distribution instead of taking the mean/mode of the distribution). The code for [obtaining evaluation samples](https://github.com/rlworkgroup/garage/blob/master/src/garage/np/_functions.py#L32) uses the `rollout` function from [sampler.utils](https://github.com/rlworkgroup/garage/blob/master/src/garage/sampler/utils.py#L10) with the argument `deterministic=True`.

The rollout function is then supposed to look into `agent_info` dictionary and use the `mean` value stored there. Unfortunately, currently in the code it looks into the `agent_infos` (with `s` at the end), which is a list containing `agent_info` dictionaries and as such obviously does not contain the `mean` key. This means that the stochastic, sampled action is used instead. My pull request solves this issue by fixing the typo.

Technical sidenote - maybe there should be an exception raised if `deterministic=True` and there is no `mean` key in the dict?

### DIFF
--- a/tests/fixtures/policies/dummy_policy.py
+++ b/tests/fixtures/policies/dummy_policy.py
@@ -17,6 +17,7 @@ class DummyPolicy(Policy):
             self,
             env_spec,
     ):
+        # pylint: disable=super-init-not-called
         self._env_spec = env_spec
         self._param = []
         self._param_values = np.random.uniform(-1, 1, 1000)
@@ -32,7 +33,7 @@ class DummyPolicy(Policy):
             dict: Distribution parameters.
 
         """
-        return self.action_space.sample(), dict(dummy='dummy')
+        return self.action_space.sample(), dict(dummy='dummy', mean=0.)
 
     def get_actions(self, observations):
         """Get multiple actions from this policy for the input observations.
--- a/tests/fixtures/policies/dummy_policy.py
+++ b/tests/fixtures/policies/dummy_policy.py
@@ -13,10 +13,7 @@ class DummyPolicy(Policy):
 
     """
 
-    def __init__(
-            self,
-            env_spec,
-    ):
+    def __init__(self, env_spec):
         # pylint: disable=super-init-not-called
         self._env_spec = env_spec
         self._param = []
--- a/src/garage/sampler/utils.py
+++ b/src/garage/sampler/utils.py
@@ -61,7 +61,7 @@ def rollout(env,
     while path_length < (max_path_length or np.inf):
         o = env.observation_space.flatten(o)
         a, agent_info = agent.get_action(o)
-        if deterministic and 'mean' in agent_infos:
+        if deterministic and 'mean' in agent_info:
             a = agent_info['mean']
         next_o, r, d, env_info = env.step(a)
         observations.append(o)
--- a/tests/garage/sampler/test_utils.py
+++ b/tests/garage/sampler/test_utils.py
@@ -32,6 +32,13 @@ class TestRollout:
         assert path['observations'][0].shape == (16, )
         assert path['actions'][0].shape == (2, 2)
 
+    def test_deterministic_action(self):
+        path = utils.rollout(self.env,
+                             self.policy,
+                             max_path_length=5,
+                             deterministic=True)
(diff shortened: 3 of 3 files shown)


krzentner:
Oh, I thought we had a test case for this code path. I guess not.

Honestly, looking for mean in the first place is kinda a hack. It only really makes sense for gaussian policies. I've wondered for some time if we should add a argument to `get_action`/`get_actions` that makes the actions deterministic.

avnishn:
https://github.com/rlworkgroup/garage/blob/master/tests/garage/sampler/test_utils.py

ryanjulian:
@Mergifyio rebase

ryanjulian:
@maciejwolczyk thanks so much for the PR!

Can you add a test which would have detected this bug (as suggested by @avnishn )?

After that, I think it's ready to merge.

maciejwolczyk:
I'm glad I could help and sorry for the delay!

Is this test okay? I've run the test on the old code with the typo and it does fail, and of course it should pass after the fix.

tests/fixtures/policies/dummy_policy.py
```diff
@@ -32,7 +33,7 @@ def get_action(self, observation):
             dict: Distribution parameters.
 
         """
-        return self.action_space.sample(), dict(dummy='dummy')
+        return self.action_space.sample(), dict(dummy='dummy', mean=0.)
```
ryanjulian, line 36: don't you need to call `super().__init__(...)` for `self.action_space` exist in this class?

maciejwolczyk: I guess the class just inherits the `action_space` property, which uses `self.env_spec`, which is defined for the `DummyPolicy` so everything works. The `__init__` for the base Policy is blank, so it doesn't do anything interesting. 
The CI passes, but if you want, I can add the `super()` call.

ryanjulian:
@Mergifyio rebase

ryanjulian:
@Mergifyio rebase

ryanjulian:
@Mergifyio rebase

ryanjulian:
@maciejwolczyk it looks like the CI is failing due to a formatting issue. Can you correct it and re-upload?

Note that this should be handled for you if you setup pre-commit, as detailed in [CONTRIBUTING.md](https://github.com/rlworkgroup/garage/blob/master/CONTRIBUTING.md)

maciejwolczyk:
That's a bit weird, earlier I've checked the commits with the pre-commit hooks enabled and also the CI passed for the first time... I guess maybe the failure now is because the [pre-commit script changed recently](https://github.com/rlworkgroup/garage/commit/d9aa7e10a29922eeb6e51aa587503432d47027be#diff-82c3b36709926de3a1b27bea0d8258cc)?

I've pushed a commit which passes the pre-commit scripts. It was actually detecting a formatting error in lines that I haven't touched, so I guess it just checks the whole file. Anyway, I hope it should pass now.

maciejwolczyk:
It's failing now, but on a later stage (normal tests) and the failing test is `test_time_limit_env` for the BulletEnv. I'm not sure if this issue is connected with the current pull request, I've checked that some of the [CI for different PRs](https://travis-ci.com/github/rlworkgroup/garage/builds/174306469) fail on this test as well.

AiRuiChen:
> It's failing now, but on a later stage (normal tests) and the failing test is `test_time_limit_env` for the BulletEnv. I'm not sure if this issue is connected with the current pull request, I've checked that some of the [CI for different PRs](https://travis-ci.com/github/rlworkgroup/garage/builds/174306469) fail on this test as well.

Hi, just sent [#1710 ](https://github.com/rlworkgroup/garage/pull/1710) to fix this failing test. Once this is merged please try rebasing the latest master branch and check CI again.

maciejwolczyk:
@Mergifyio rebase

ryanjulian:
@Mergifyio rebase

ryanjulian:

ryanjulian:

PR #1619, opened 2020-06-25
Only build Docker container for deps changes

ryanjulian:
Changes TravisCI to only build the Docker container for a PR or push if
it contains changes likely to invalidate the Docker container.

The container always builds on nightly builds or API builds.


ryanjulian:
@Mergifyio rebase

PR #1626, opened 2020-06-26
Add FragmentWorker for Off-Policy RL

krzentner:
This is a sampler.Worker which collects fragments of trajectories,
allowing off-policy RL algorithms to update with partial trajectories.

### DIFF
--- a/src/garage/sampler/fragment_worker.py
+++ b/src/garage/sampler/fragment_worker.py
@@ -1,0 +1,156 @@
+"""Worker that "vectorizes" environments."""
+import copy
+
+import numpy as np
+
+from garage import TrajectoryBatch
+from garage.sampler import _apply_env_update, InProgressTrajectory
+from garage.sampler.default_worker import DefaultWorker
+
+
+class FragmentWorker(DefaultWorker):
+    """Vectorized Worker that collects partial trajectories.
+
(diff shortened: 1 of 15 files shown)


src/garage/sampler/fragment_worker.py
```diff
@@ -0,0 +1,156 @@
...
+        """Take a single time-step in the current rollout.
+
+        Returns:
+            bool: True iff at least one of the paths was completed.
+
+        """
+        prev_obs = np.asarray([frag.last_obs for frag in self._fragments])
+        actions, agent_infos = self.agent.get_actions(prev_obs)
+        completes = [False] * len(self._envs)
+        for i, action in enumerate(actions):
```
avnishn, line 104: why iterate over the number of actions? will this not always be 1 action?

avnishn: @krzentner bump

ryanjulian: as the PR description says, it's vectorized.

ryanjulian:
@Mergifyio rebase

krzentner:
There are still some test failures with this change (specifically, the off-policy TF algos report low performance in tests, despite working normally when I run examples).

PR #1629, opened 2020-06-26
Delete .mdlrc

ryanjulian:
Let's try the strictest version first.


irisliucy:
The reason for the markdown style file is that some rules don't really make sense. For example, line length = 80, some url / code block will exceed this number and it's a pain to manually correct each line for each markdown file. 

But we can try first.

PR #1633, opened 2020-06-26
Skip bullet envs in test_base

AiRuiChen:
Fix broken CI by skipping bullet environment checks in `test_base.py`.

Now `GarageEnv` can't handle bullet environment pickle logic, so skip bullet environment related tests. Bullet environments are already tested in [`test_bullet_env.py`](https://github.com/rlworkgroup/garage/blob/master/tests/garage/envs/bullet/test_bullet_env.py).

This is only a temporary fix. Future PRs will make GarageEnv silently wrap bullet environments with BulletEnv, and make GarageEnv works for all gym.make() instances.

Note: `garage.env.bullet` is imported to _explicitly_ add bullet environments to gym registry. Without this import, bullet environments won't be added to gym registry unless [`test_bullet_env.py`](https://github.com/rlworkgroup/garage/blob/master/tests/garage/envs/bullet/test_bullet_env.py#L5) is tested at the same time. This ensures consistent test input of `test_base.py`.

### DIFF
--- a/src/garage/envs/bullet/__init__.py
+++ b/src/garage/envs/bullet/__init__.py
@@ -11,3 +11,20 @@ except Exception as e:
 from garage.envs.bullet.bullet_env import BulletEnv
 
 __all__ = ['BulletEnv']
+
+
+def get_bullet_env_list():
+    """Return a complete list of Bullet Gym environments.
+
+    Returns:
+        list: a list of bullet environment names (str)
+    """
+    envs = [env.replace('- ', '') for env in pybullet_envs.getList()]
+    # Hardcoded missing environment names from pybullet_envs.getList()
+    envs.extend([
+        'MinitaurExtendedEnv-v0', 'MinitaurReactiveEnv-v0',
+        'MinitaurBallGymEnv-v0', 'MinitaurTrottingEnv-v0',
+        'MinitaurStandGymEnv-v0', 'MinitaurAlternatingLegsEnv-v0',
+        'MinitaurFourLegStandEnv-v0', 'KukaDiverseObjectGrasping-v0'
+    ])
+    return envs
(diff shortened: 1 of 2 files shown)


tests/garage/tf/envs/test_base.py
```diff
@@ -17,18 +18,30 @@ def test_is_pickleable(self):
     @pytest.mark.nightly
     @pytest.mark.parametrize('spec', list(gym.envs.registry.all()))
     def test_all_gym_envs(self, spec):
+        bullet_envs = [
+            env.replace('- ', '') for env in pybullet_envs.getList()
```
ryanjulian, line 22: why not remove them from the parametrize list in exactly the same fashion?

ryanjulian:
I'm not sure I understand how this solves the build failure, which was fundamentally caused by an ImportError from pybullet code, IIRC

src/garage/envs/bullet/__init__.py
```diff
@@ -11,3 +11,22 @@
...
+
+def get_bullet_env_list():
+    """Return a complete list of Bullet Gym environments.
+
+    Returns:
+        list: a list of bullet environment names (str)
+    """
+    envs = [env.replace('- ', '') for env in pybullet_envs.getList()]
+    # Hardcoded missing environment names from pybullet_envs.getList()
+    envs.append('MinitaurExtendedEnv-v0')
```
ryanjulian, line 24: you could use `.extend`, rather than being repetitive

ryanjulian:
@Mergifyio rebase

AiRuiChen:
[#1684 ](https://github.com/rlworkgroup/garage/pull/1684) allows `GarageEnv` to handle bullet environments as well -> no need to skip bullet environments in the test.

PR #1635, opened 2020-06-26
Don't push to latest tag on DockerHub

gitanshu:
This commit prevents this release branch from using
garage-ci:latest on DockerHub, since that tag is used
by PRs on master branch, and images from release branches
can negatively affect docker cache.

Also, backports #1576

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -81,8 +81,8 @@ before_install:
 
 install:
   # Pull cached docker image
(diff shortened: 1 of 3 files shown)


ryanjulian:
Does this also apply in some way to `release-2019.10`?

gitanshu:
> Does this also apply in some way to `release-2019.10`?

Yes, working on it

Issue #1537, opened 2020-06-09
Docs page "Save, load, and resume experiments"


ryanjulian:
See #1426

Issue #1535, opened 2020-06-09
Docs page "Usage Guide > Use a pre-trained network to start a new experiment"


ryanjulian:
See #1426

PR #1642, opened 2020-06-27
Documentation on loading a trained policy

### LINKED ISSUE #1534: Docs page "Load and use a trained policy"
See #1426

haydenshively:
It's been a little unclear what the scope of this usage guide should be, especially considering the similarities between #1534, #1535, and #1537. Based on Slack conversations, I think this covers the right material but please correct me if I'm wrong.

Should close #1534

### DIFF
--- a/docs/user/reuse_garage_policy.md
+++ b/docs/user/reuse_garage_policy.md
@@ -1,0 +1,83 @@
+# Load and Use a Trained Policy
+
+In this section you will learn how to extract a trained policy from an experiment
+snapshot, as well as how to evaluate that policy in an environment.
+
+## Obtaining an experiment snapshot
+
+Please refer to [this page](save_load_resume_exp.md) for information on how to save
+an experiment snapshot. The snapshot contains data such as:
+
+- The runner's `setup_args` and `train_args`
+  - Random seed
+  - Batch size
+  - Number of epochs
+  - And [more](https://github.com/rlworkgroup/garage/blob/175ac4c90a408e2314d91cdbe95e419b183e0684/src/garage/experiment/local_runner.py#L393)
+- The experiment's `stats`
+- The environment
+- The algorithm **(which includes the policy we want to evaluate)**
+
+## Extracting a trained policy from a snapshot
+
+To extract the trained policy from a saved experiment, you only need a few lines
+of code:
+
+```python
+# Load the policy
+from garage.experiment import Snapshotter
+snapshotter = Snapshotter()
+data = snapshotter.load('path/to/snapshot/dir')
+policy = data['algo'].policy
+
+# You can also access other components of the experiment
+env = data['env']
+```
+
+This code makes use of a Garage `Snapshotter` instance. It calls
+cloudpickle behind the scenes, and should continue to work even if we
+change how we pickle (we used to use joblib, for example).
+
+## Applying the policy to an environment
+
+In order to use your newly-loaded trained policy, you first have to make sure that
+the shapes of its observation and action spaces match those of the target environment.
+An easy way of doing this is run the policy in the same environment in which it
+was trained.
+
+Once you have an environment initialized, the basic idea is this:
+
+```python
+steps, max_steps = 0, 150
+done = False
+obs = env.reset()  # The initial observation
+policy.reset()
+
+while steps < max_steps and not done:
+    obs, rew, done, _ = env.step(policy.get_action(obs))
+    env.render()  # Render the environment to see what's going on (optional)
+    steps += 1
+
+env.close()
+```
+
+This logic is bundled up in a more robust way in the `rollout()` function from
+`garage.sampler.utils`. Let's bring everything together:
+
+```python
+# Load the policy and the env in which it was trained
+from garage.experiment import Snapshotter
+snapshotter = Snapshotter()
+data = snapshotter.load('path/to/snapshot/dir')
+policy = data['algo'].policy
+env = data['env']
+
+# See what the trained policy can accomplish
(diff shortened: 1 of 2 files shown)


docs/user/reuse_garage_policy.md
```diff
@@ -0,0 +1,163 @@
+# Reusing a policy trained with Garage
```
ryanjulian, line 1: This will become the table-of-contents entry, so keep it terse

"Load and Use a Trained Policy"

docs/user/reuse_garage_policy.md
```diff
@@ -0,0 +1,163 @@
...
+
+In this section you will learn how to save and load snapshots of your experiment
+while training. We will also show you how to extract a trained policy from a
+snapshot.
+
+## Saving an experiment snapshot
```
ryanjulian, line 7: Just refer back to the experiment documentation for how to get one of these. No need to recapitulate it here. This doc can assume you already have run an experiment.

docs/user/reuse_garage_policy.md
```diff
@@ -0,0 +1,163 @@
...
+
+
+vpg_pendulum()
+```
+
+By default, the data is saved to `./data/local/experiment`, but you can
+specify your own directory by setting `snapshot_dir` when instantiating
+the `LocalRunner`.
+
+## Resuming an experiment
```
ryanjulian, line 92: this is a different documentation page. #1537

docs/user/reuse_garage_policy.md
```diff
@@ -0,0 +1,163 @@
...
+    # 1. resume immediately using existing parameters
+    runner.resume()
+    # 2. or modify training parameters
+    runner.resume(n_epochs=20)
+
+
+resume_experiment()
+```
+
+## Extracting a trained policy from a snapshot
```
ryanjulian, line 130: this is the topic of this document.

docs/user/reuse_garage_policy.md
```diff
@@ -0,0 +1,163 @@
...
+
+If instead you wish to extract the trained policy from a saved experiment,
```
ryanjulian, line 132: you should only present one way of doing things.

docs/user/reuse_garage_policy.md
```diff
@@ -0,0 +1,163 @@
...
+env = data['env']
+```
+
+Alternatively, you can use a Garage `Snapshotter` instance. It will handle
+the calls to cloudpickle for you, and should continue to work even if we
+change how we pickle (we used to use joblib, for example):
+
+```
+# Load the policy
+from garage.experiment import SnapshotConfig, Snapshotter
```
ryanjulian, line 151: i think you should stick with this one for now, for the exact reason you note (we could change the snapshot format).

docs/user/reuse_garage_policy.md
```diff
@@ -0,0 +1,163 @@
...
+snapshotter = Snapshotter()
+data = snapshotter.load('path/to/snapshot/file')
```
ryanjulian, line 153: is this a path to a directory or a path to a file?

docs/user/reuse_garage_policy.md
```diff
@@ -0,0 +1,163 @@
+# Reusing a policy trained with Garage
+
+In this section you will learn how to save and load snapshots of your experiment
+while training. We will also show you how to extract a trained policy from a
+snapshot.
+
+## Saving an experiment snapshot
+
+As you've seen in previous guides, experiment activities are governed by a
```
ryanjulian, line 9: if you refer to other docs pages, you should always link to them.

haydenshively: I don't think the one I need to link to is on master yet, so I'll come back to this PR later in the week.

docs/user/reuse_garage_policy.md
ryanjulian: specifically, how to load a trained policy and evaluate it, without updating it

docs/user/reuse_garage_policy.md
```diff
@@ -0,0 +1,54 @@
...
+By default, the data is saved to `./data/local/experiment`, but you can
+specify your own directory by setting `snapshot_dir` when instantiating
+the `LocalRunner`.
+
+## Extracting a trained policy from a snapshot
+
+If instead you wish to extract the trained policy from a saved experiment,
+the process is even simpler. You can do it in a few lines like this:
+
+```
```
ryanjulian, line 36: this code block should specify `py` or `python`

docs/user/reuse_garage_policy.md
```diff
@@ -0,0 +1,54 @@
+# Load and Use a Trained Policy
+
+In this section you will learn how to save and load snapshots of your experiment
```
ryanjulian, line 3: Saving/loading snapshots is covered by a different tutorial. Please keep the focus here singular.

docs/user/reuse_garage_policy.md
```diff
@@ -0,0 +1,54 @@
...
+
+If instead you wish to extract the trained policy from a saved experiment,
+the process is even simpler. You can do it in a few lines like this:
+
+```python
+# Load the policy
+from garage.experiment import SnapshotConfig, Snapshotter
+snapshotter = Snapshotter()
+data = snapshotter.load('path/to/snapshot/dir')
+policy = data['algo'].policy
```
ryanjulian, line 41: this code only extracts the policy -- it doesn't make use of it.

can you show the user how to run the policy in an environment?

ryanjulian:
@Mergifyio rebase

PR #1663, opened 2020-06-29
Better API documentation

ryanjulian:
Adds sphinx-autoapi and custom templates for better API documentation.


ryanjulian:
@Mergifyio rebase

PR #1665, opened 2020-06-30
Fail the CI when codecov upload fails

ryanjulian:


ryanjulian:
@krzentner The problem is generally with upload (the script curl is executing), but it certainly doesn't hurt.

PR #1684, opened 2020-07-01
GarageEnv wraps BulletEnv

AiRuiChen:
1. Make `GarageEnv` capable of handling all environments created by `gym.make()`.
Now `GarageEnv.__new__()` checks the input environment type and returns a `BulletEnv` if the environment is bullet-based.
User can use `GarageEnv` to work with bullet-based environments, instead of importing `BulletEnv` -> Increases user-interface simplicity.

2. Change `BulletEnv` pickle appraoch
A new environment should be created via `gym.make()` rather than its class constructor; otherwise the environment won't be wrapped with `TimeLimit`.

### DIFF
--- a/src/garage/envs/bullet/__init__.py
+++ b/src/garage/envs/bullet/__init__.py
@@ -11,3 +11,36 @@ except Exception as e:
 from garage.envs.bullet.bullet_env import BulletEnv
 
 __all__ = ['BulletEnv']
+
+
+def _get_bullet_env_list():
+    """Return a complete list of Bullet Gym environments.
+
+    Returns:
+        list: a list of bullet environment id (str)
+    """
+    envs = [env.replace('- ', '') for env in pybullet_envs.getList()]
+    # Hardcoded missing environment names from pybullet_envs.getList()
+    envs.extend([
+        'MinitaurExtendedEnv-v0', 'MinitaurReactiveEnv-v0',
+        'MinitaurBallGymEnv-v0', 'MinitaurTrottingEnv-v0',
+        'MinitaurStandGymEnv-v0', 'MinitaurAlternatingLegsEnv-v0',
+        'MinitaurFourLegStandEnv-v0', 'KukaDiverseObjectGrasping-v0'
+    ])
+    return envs
+
+
+def _get_unsupported_env_list():
+    """Return a list of unsupported Bullet Gym environments.
+
+    See https://github.com/rlworkgroup/garage/issues/1668
+
+    Returns:
+        list: a list of bullet environment id (str)
+    """
+    return [
+        'MinitaurExtendedEnv-v0', 'MinitaurReactiveEnv-v0',
+        'MinitaurBallGymEnv-v0', 'MinitaurTrottingEnv-v0',
+        'MinitaurStandGymEnv-v0', 'MinitaurAlternatingLegsEnv-v0',
+        'MinitaurFourLegStandEnv-v0', 'KukaDiverseObjectGrasping-v0'
+    ]
--- a/src/garage/envs/bullet/bullet_env.py
+++ b/src/garage/envs/bullet/bullet_env.py
@@ -1,15 +1,125 @@
 """Wrappers for py_bullet environments."""
 import inspect
 
+import akro
+import gym
 from pybullet_envs.bullet.minitaur_duck_gym_env import MinitaurBulletDuckEnv
 from pybullet_envs.bullet.minitaur_gym_env import MinitaurBulletEnv
 from pybullet_envs.env_bases import MJCFBaseBulletEnv
 
-from garage.envs import GarageEnv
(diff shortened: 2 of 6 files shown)


src/garage/envs/bullet/__init__.py
```diff
@@ -10,4 +10,37 @@
 
 from garage.envs.bullet.bullet_env import BulletEnv
 
-__all__ = ['BulletEnv']
+__all__ = ['BulletEnv', 'get_bullet_env_list', 'get_unsupported_env_list']
+
+
+def get_bullet_env_list():
```
ryanjulian, line 16: can you make these private?

src/garage/envs/bullet/__init__.py
```diff
@@ -10,4 +10,37 @@
 
 from garage.envs.bullet.bullet_env import BulletEnv
 
-__all__ = ['BulletEnv']
+__all__ = ['BulletEnv', 'get_bullet_env_list', 'get_unsupported_env_list']
```
ryanjulian, line 13: no need to export get* to the API -- users are not expected to use these

src/garage/envs/bullet/bullet_env.py
```diff
@@ -1,16 +1,126 @@
...
+        return self.env.reset(**kwargs)
+
+    def step(self, action):
+        """Call step on wrapped env.
+
+        This method is necessary to suppress a deprecated warning
+        thrown by gym.Wrapper.
+
+        Args:
+            action (object): An action provided by the agent.
```
ryanjulian, line 99: this is an np.ndarry

src/garage/envs/bullet/bullet_env.py
```diff
@@ -1,16 +1,126 @@
...
+
+        Returns:
+            object: Agent's observation of the current environment
```
ryanjulian, line 102: this is an np.ndarray

src/garage/envs/bullet/bullet_env.py
```diff
@@ -1,16 +1,126 @@
...
+            float : Amount of reward returned after previous action
+            bool : Whether the episode has ended, in which case further step()
```
ryanjulian, line 104: i think pre-commit will catch this, but the form is `bool: Whether...` (i.e. no spaces after the colon)

formatting is important so that the API doc parser can work.

src/garage/envs/bullet/bullet_env.py
```diff
@@ -1,16 +1,126 @@
...
+            kwargs: Keyword args
+
+        Returns:
+            object: The initial observation.
+
+        """
+        return self.env.reset(**kwargs)
+
+    def step(self, action):
+        """Call step on wrapped env.
```
ryanjulian, line 93: note that you should be able to simply use "See `:class:`gym.Wrapper`" or similar

tests/garage/envs/bullet/test_bullet_env.py
```diff
@@ -88,3 +79,5 @@ def test_pickle_creates_new_server(env_ids):
                 # Some environments have _p as the pybullet module, and they
                 # don't store client id, so can't check here
                 pass
+        for env in envs:
```
ryanjulian, line 82: i think pre-commit with catch this, but you should always leave a trailing line after a block

src/garage/envs/bullet/bullet_env.py
```diff
@@ -1,16 +1,126 @@
...
+
+        # Needed for deserialization
+        self._env = env
+        self._env_name = env_name
+
+        super().__init__(env)
+        self.action_space = akro.from_gym(self.env.action_space)
+        self.observation_space = akro.from_gym(self.env.observation_space,
+                                               is_image=is_image)
+        self.__spec = EnvSpec(action_space=self.action_space,
```
ryanjulian, line 48: i think you only need a single underscore? why double?

AiRuiChen: Copied from garage_env.py. Found that `__spec` is only accessed as a private variable inside the class. Will change this in `BulletEnv` and `GarageEnv`.

src/garage/envs/bullet/__init__.py
AiRuiChen: Do you mean move them into BulletEnv()? 

We can have something like 
`@classmethod 
BulletEnv.check_env_supporeted()`
and 
`@classmethod 
BulletEnv.get_bullet_env_list()`

note:
`get_bullet_env_list` can be removed once `pybullet` is upgraded to the next release (can call `pybullet_envs.getList()` instead).

krzentner: I think Ryan just means you should put an underscore at the start of the function name (e.g. `_get_bullet_env_list`). This indicates outside users of garage shouldn't call this function.

Issue #1630, opened 2020-06-26
Baselines in tf tree should inherit from garage.tf.Model


ahtsan:
- [x] GaussianCNNBaseline

- [ ] GaussianMLPBaseline

- [x] ContinuousMLPBaseline

ryanjulian:
Why make this refactor when we can just spend time deleting regressors?

ahtsan:
Oh, I meant Baseline in this case.

PR #1705, opened 2020-07-03
Refactor GaussianMLPBaseline

yeukfu:
#1630 
Below are the results of benchmarks with this [experiment](https://github.com/rlworkgroup/garage/blob/master/benchmarks/src/garage_benchmarks/experiments/baselines/gaussian_mlp_baseline.py).
- 5 same seeds:
HalfCheetah-v2
Hopper-v2

- 3 different seeds:
HalfCheetah-v2
Hopper-v2
InvertedDoublePendulum-v2
InvertedPendulum-v2
Swimmer-v2
Walker2d-v2


yeukfu:
> Can you also benchmark for this change?

Sure, will do.

yeukfu:
@ahtsan The results of benchmarks are ready in the first comment.

ryanjulian:
@yeukfu how many seeds do the benchmarks have?

yeukfu:
> @yeukfu how many seeds do the benchmarks have?

3 seeds.

ryanjulian:
Can you run 5 or 10 for Hopper and HalfCheetah? I think it is probably seed-dependence, but I want to double-check.

yeukfu:
> Can you run 5 or 10 for Hopper and HalfCheetah? I think it is probably seed-dependence, but I want to double-check.

Do I need to run with the same seeds? The benchmark shown use different seeds for the two baselines.

ryanjulian:
@yeukfu it's essential that you use the same seed for both baselines, so that they experience the same sequence of random numbers. that is "run n seeds" means "choose seeds i = 1...n, then run n experiments. in each experiment, set the seed of BOTH baselines to i"

if you didn't do that before, i think it's sufficient to run 5 seeds for both HalfCheetah and Hopper.

yeukfu:
> @yeukfu it's essential that you use the same seed for both baselines, so that they experience the same sequence of random numbers. that is "run n seeds" means "choose seeds i = 1...n, then run n experiments. in each experiment, set the seed of BOTH baselines to i"
> 
> if you didn't do that before, i think it's sufficient to run 5 seeds for both HalfCheetah and Hopper.

Got it. I Have add the new results.

ryanjulian:
Looks good to me! Thanks for the updated plots!

ryanjulian:
@Mergifyio rebase

Issue #1729, opened 2020-07-08
Importing from garage.envs doesn't work without installing garage[bullet]


gitanshu:
Since bullet is optional for installing garage, importing from garage.envs should work without installing pybullet, for code that doesn't need it. 

```
$ pytest tests/garage/envs/test_normalized_env.py              

_________________ ERROR collecting tests/garage/envs/test_normalized_env.py______________________
ImportError while importing test module '/home/gitanshu/resl/garage/tests/garage/envs/test_normalized_env.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
src/garage/envs/bullet/__init__.py:6: in <module>
    import pybullet_envs  # noqa: F401
E   ModuleNotFoundError: No module named 'pybullet_envs'

During handling of the above exception, another exception occurred:
tests/garage/envs/test_normalized_env.py:5: in <module>
    from garage.envs import PointEnv
src/garage/envs/__init__.py:4: in <module>
    from garage.envs.garage_env import GarageEnv
src/garage/envs/garage_env.py:9: in <module>
    from garage.envs.bullet import _get_bullet_env_list, BulletEnv
src/garage/envs/bullet/__init__.py:8: in <module>
    raise ImportError('To use garage\'s bullet wrappers, please install '
E   ImportError: To use garage's bullet wrappers, please install garage[bullet]
```

PR #1730, opened 2020-07-08
Conditionally imports BulletEnv

AiRuiChen:
Closes [#1729 ](https://github.com/rlworkgroup/garage/issues/1729)

### DIFF
--- a/src/garage/envs/garage_env.py
+++ b/src/garage/envs/garage_env.py
@@ -6,7 +6,6 @@ import akro
 import gym
 from gym.wrappers.time_limit import TimeLimit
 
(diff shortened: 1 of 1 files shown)


src/garage/envs/garage_env.py
```diff
@@ -55,7 +54,12 @@ def __new__(cls, *args, **kwargs):
         Returns:
              garage.envs.bullet.BulletEnv: if the environment is a bullet-based
                 environment. Else returns a garage.envs.GarageEnv
+
+        Raises:
+            ImportError: Raise if the input environment is a bullet
+            environment but garage[bullet] is installed.
```
krzentner, line 60: Should be "is not installed."

PR #1745, opened 2020-07-09
Relax last_avg_ret to deflaky test_td3

### LINKED ISSUE #1732: test_td3_pedulum is flaky
See https://travis-ci.com/github/rlworkgroup/garage/builds/174753960

yeukfu:
Close #1732.

### DIFF
--- a/tests/garage/tf/algos/test_td3.py
+++ b/tests/garage/tf/algos/test_td3.py
@@ -20,7 +20,6 @@ class TestTD3(TfGraphTestCase):
     """Tests for TD3 algo."""
 
     @pytest.mark.mujoco_long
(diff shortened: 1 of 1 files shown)


yeukfu:
> I know it's not perfect, but can you also add a call to `deterministic.set_seed`?

The class `TestTD3` inherits from `TfGraphTestCase`  and in [`setup_method()`](https://github.com/rlworkgroup/garage/blob/80b99b8f99fbc8ad559846134b8b94182cf85903/tests/fixtures/fixtures.py#L52) of `TfGraphTestCase`, `deterministic.set_seed` has been called. I think we don't need to call it again here?

Issue #1260, opened 2020-03-27
deterministic.set_seed does not guarantee full determinism in TensorFlow


ahtsan:
Determinism in TensorFlow depends on two seeds: Graph seed and op seed. Currently `deterministic.set_seed()` only set the graph seed, which makes the random sequences generated by all ops be repeatable across sessions, but not when sequences of ops changed.

Setting both graph_seed and op_seed will guarantee full determinism, i.e. calling 
[https://www.tensorflow.org/api_docs/python/tf/compat/v1/set_random_seed](set_random_seed) and applying a seed when calling any tf.ops (e.g. `tf.random.uniform([1], seed=1`).

ryanjulian:
@ahtsan can you estimate how big a change we would need to fix this?

ahtsan:
Basically providing a seed to all tf.op -- shouldn't take long

PR #1757, opened 2020-07-11
Set seed for all ops in TF tree

ahtsan:
This PR set `SeedStream` for the following ops in order to achieve determinism in TF tree:
* tf.random.normal()
* tfp.distribution.sample()
* tf.initializers.glorot_uniform()


ahtsan:
Close #1260

src/garage/experiment/deterministic.py
```diff
@@ -4,6 +4,7 @@
 import warnings
 
 import numpy as np
+import tensorflow_probability as tfp
```
ryanjulian, line 7: please make sure this doesn't crash garage if TF isn't installed in the environment. we are working towards making TF optional.

tests/garage/experiment/test_deterministic.py
```diff
@@ -38,6 +38,34 @@ def test_deterministic_tensorflow():
     assert np.allclose(rand_tensor, deterministic_tensor)
 
 
+def test_deterministic_tfp_seed_stream():
```
ryanjulian, line 41: this test is an okay unit tests but i think its helpfulness is very narrow for what's most important to test: whether we have achieved (a minimum level of) determinism.

can you also add an integration test which verifies that TF experiments are deterministic? rather than interacting with `get_tf_seed_stream()` directly, it would instead run an algorithm+environment (for very few steps) and ensure it's getting the same results.

Note that I don't think we've fully covered determinism in MuJoCo (or gym or Bullet or dm_control or Metaworld). Writing this test might tell us where our uncovered sources of randomness are. If you find them, we can make issues for them and address them later.

ryanjulian:
@ahtsan is there anything blocking this PR?

ahtsan:
Actually yes, DQN integration test is failing. I haven’t been able to get time to look into that.

ryanjulian:

ryanjulian:

PR #1772, opened 2020-07-16
Remove old off-policy logging code

krzentner:
This is a precursor to merging `FragmentWorker`, since it breaks this code, and we rely on it for testing.

I don't know why the magic numbers in the DQN tests need to be decreased, after a few days of digging. This change only affects the logging, so the true performance shouldn't have decreased.


krzentner:
@ryanjulian Is there any way to re-trigger the doc build? It looks like it timed out.

ryanjulian:
@Mergifyio rebase

krzentner:
Oh, apparently this change also Fixes #1773. I wasn't trying to include any behavior changes.

Issue #1239, opened 2020-03-21
Remove setter methods in EnvSpec


lywong92:
Remove setter methods in `EnvSpec` because `action_space` and `observation_space` should be immutable. Several existing tests set `action_space` and `observation_space`, and these tests should be corrected as well.

PR #1775, opened 2020-07-16
Define garage.Environment API

AiRuiChen:
This is a proposed API for all garage environments.

Design decisions to note:

1. `reset()` and `step()` returns a `TimeStep`. 
2. `TimeStep` has the attribute `step_type`, which can be `FIRST`, `MID`, `LAST`, `TIMEOUT`.
3. `reset()` starts a new sequence, instead of just returning the initial observation as in the Gym API.
4. Separate `visualize()` (the `human` mode) from gym.Env `render()` API.
The key difference is `visualize()` _displays_ the environment and returns **nothing**, whereas `render()` _returns_ something.
5. `observation` can be in any format as long as it conforms to `env_spec.observation_space`. 
6. Update `TimeStep` documentation to make it clear that `TimeStep` represents a SARS tuple.

Note that modifications of `TimeStep` serves a demo purpose to show how this new API works. Changes for `TimeStep` will be removed once this PR is approved (i.e. ultimately this PR should only introduce `src/garage/envs/environment.py`). I will create a new PR for `TimeStep` with the new definition and implementations.


src/garage/envs/environment.py
```diff
@@ -0,0 +1,176 @@
...
+
+    @abc.abstractmethod
+    def render(self, mode):
+        """Renders the environment.
+
+        The set of supported modes varies per environment. (And some
+        environments do not support rendering at all.) By convention,
+        if mode is:
+
+        * display: render to the current display or terminal and
```
krzentner, line 109: I definitely think it doesn't make sense to have this be part of the same method as the other options (even though gym does it this way).

src/garage/envs/environment.py
```diff
@@ -0,0 +1,176 @@
+"""Base Garage Environment API."""
+
+import abc
+
+
+class Environment(abc.ABC):
```
krzentner, line 6: Somewhere there should be a reference to the pickle protocol, and that it should work for this interface.

src/garage/_dtypes.py
```diff
@@ -428,22 +431,29 @@ def next_observations(self):
             ]))
 
 
+class StepType(enum.IntEnum):
+    """Defines the status of a `TimeStep` within a sequence."""
+    # Denotes the first `TimeStep` in a sequence.
+    FIRST = 0
+    # Denotes any `TimeStep` in a sequence that is not FIRST or LAST.
+    MID = 1
+    # Denotes the last `TimeStep` in a sequence.
```
ryanjulian, line 440: this comment is not actually true because TIMEOUT is also a last TimeStep

src/garage/_dtypes.py
```diff
@@ -428,22 +431,29 @@ def next_observations(self):
...
+    LAST = 2
```
ryanjulian, line 441: If you're going to have `TIMEOUT` then this should be named `TERMINALS` to distinguish it from `TIMEOUT`.

`TIMEOUT` is also a `LAST` step. This could actually accidentally make the API harder to use. I'm not sure it's bad (it forces you to make sure you're getting what you asked for), but it might also introduce a sharp edge.

Imagine I want to retrieve all the last observations. Without `TIMEOUT` I would do

```python
last_obs = tsb.next_observations[tsb.step_types == StepType.TERMINAL]
```

With `TIMEOUT` you need
```python
last_obs = tsb.next_observations[tsb.step_types == StepType.TERMINAL | tsb.step_types == StepType.TIMEOUT]
```

**Ways of dealing with this**
* Do nothing

* Use a different field for indicating timeout
    ```py
    last_obs = tsb.next_observations[tsb.step_types == StepType.LAST]  # includes TERMINAL and TIMEOUT
    # timeout_obs = last_obs[tsb.timeouts]  <-- oof, you actually can't do this
    timeout_obs = tsb.next_observations[tsb.timeouts]  # <-- okay this should work fine  # just timeouts
    ```

* Leave it as is, but add a helper function

    ```py
    def get_last_obs(self):
        return self.next_observations[self.step_types == StepType.TERMINAL | self.step_types == StepType.TIMEOUT]

* Switch up the Enum to make TERMINAL the special case

    ```python
    last_obs = tsb.next_observations[tsb.step_types == StepType.LAST]  # includes TERMINAL and TIMEOUT
    terminal_obs = tsb.next_observations[tsb.terminals]  # just terminals
    ```

**How to decide:**
Try refactoring functions like these using different options for this design, and see which one leads to more intuitive/less error-prone code. Note that in the off-policy algorithms, terminals means specifically *not* a timeout step, whereas `paths_to_tensors` (for the on-policy algorithms) doesn't care.

https://github.com/rlworkgroup/garage/blob/f5e8cbc21c5a78e2c2996e256f0f2bd5b6930f90/src/garage/torch/algos/sac.py#L370
https://github.com/rlworkgroup/garage/blob/f5e8cbc21c5a78e2c2996e256f0f2bd5b6930f90/src/garage/tf/algos/ddpg.py#L370
https://github.com/rlworkgroup/garage/blob/f5e8cbc21c5a78e2c2996e256f0f2bd5b6930f90/src/garage/tf/_functions.py#L81
https://github.com/rlworkgroup/garage/blob/f5e8cbc21c5a78e2c2996e256f0f2bd5b6930f90/src/garage/tf/algos/dqn.py#L154

src/garage/envs/environment.py
```diff
@@ -0,0 +1,189 @@
...
+    """The main API for garage environments.
+
+    The main public API methods are:
```
ryanjulian, line 9: make sure these blocks all render nicely on ReadTheDocs

src/garage/envs/environment.py
```diff
@@ -0,0 +1,189 @@
...
+        contain attributes that are not pickle-able (e.g. a client-server
+        connection). In such cases, override `__getstate__()` and
+        `__setstate()__` to add your custom pickle logic.
+
+    Note: a default reward range set to [-inf,+inf] already exists. Set it if
+        you want a narrower range.
+
+    """
+
+    # Set this in SOME subclasses
```
ryanjulian, line 35: ??? which ones

src/garage/envs/environment.py
```diff
@@ -0,0 +1,189 @@
...
+    metadata = {'render.modes': []}
+    reward_range = (-float('inf'), float('inf'))
+    spec = None
```
ryanjulian, line 38: what is this?

src/garage/envs/environment.py
```diff
@@ -0,0 +1,189 @@
...
+
+    # Set these in ALL subclasses
+    action_space = None
```
ryanjulian, line 41: make these an abstract property, so that they are immutable

src/garage/envs/environment.py
```diff
@@ -0,0 +1,189 @@
...
+        connection). In such cases, override `__getstate__()` and
+        `__setstate()__` to add your custom pickle logic.
+
+    Note: a default reward range set to [-inf,+inf] already exists. Set it if
+        you want a narrower range.
+
+    """
+
+    # Set this in SOME subclasses
+    metadata = {'render.modes': []}
```
ryanjulian, line 36: why have this mutable metadata field? will this ever contain keys other than `render.modes`? Why not instead just have an abstract property `render_modes`

src/garage/envs/environment.py
```diff
@@ -0,0 +1,189 @@
...
+            def render(self, mode):
+                if mode == 'rgb_array':
+                    return np.array(...) # return RGB frame suitable for video
+                elif mode == 'ansi':
+                    ... # return text output
+                else:
+                    super(MyEnv, self).render(mode=mode) # just raise an
+                    exception
+        """
+        raise NotImplementedError
```
ryanjulian, line 147: this is not necessary in ABCs

src/garage/envs/environment.py
```diff
@@ -0,0 +1,189 @@
...
+        Args:
+            mode (str): the mode to render with. The string must be present in
+            self.metadata['render.modes'].
+
+        Raises:
+            NotImplementedError: If the mode is not implemented.
+
+        For example:
+
+        class MyEnv(Env):
```
ryanjulian, line 135: make sure this renders nicely on ReadTheDocs

src/garage/envs/environment.py
```diff
@@ -0,0 +1,189 @@
...
+
+        Returns:
+            str: String representation of the environment.
+        """
+        if self.spec is None:
+            return '<{} instance>'.format(type(self).__name__)
+        else:
+            return '<{}<{}>>'.format(type(self).__name__, self.spec.id)
+
+    def __enter__(self):
```
ryanjulian, line 175: why???

src/garage/envs/environment.py
```diff
@@ -0,0 +1,189 @@
...
+        """Closes the environment.
+
+        Override this function in your subclass to perform any necessary
+        cleanup.
+
+        Environments will automatically close() themselves when
+        garbage collected or when the program exits.
+        """
+
+    def __str__(self):
```
ryanjulian, line 164: is this really better than the default repr?

src/garage/envs/environment.py
```diff
@@ -0,0 +1,189 @@
...
+        """Displays the environment visualization."""
+
+    @abc.abstractmethod
+    def close(self):
+        """Closes the environment.
+
+        Override this function in your subclass to perform any necessary
+        cleanup.
+
+        Environments will automatically close() themselves when
```
ryanjulian, line 160: i don't see any features in this API ensuring this will get called on destruct (e.g. an overload of `__del__`)

if you want the implementer to ensure that close() is called on destruct, you need to say that. 

in general Python will try to clean up, but you can definitely cause huge memory leaks in MuJoCo by not close()ing your environments (especially if you launch viewers for them).

does this also close any windows launched by `visualize()`?

src/garage/envs/environment.py
```diff
@@ -0,0 +1,189 @@
...
+                elif mode == 'ansi':
+                    ... # return text output
+                else:
+                    super(MyEnv, self).render(mode=mode) # just raise an
+                    exception
+        """
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def visualize(self):
```
ryanjulian, line 150: do i call this once or may times?

src/garage/envs/environment.py
```diff
@@ -0,0 +1,189 @@
...
+        close
+
+    And set the following attributes:
+
+        action_space: The Space object corresponding to valid actions
+        observation_space: The Space object corresponding to valid observations
+        reward_range: A tuple corresponding to the min and max possible rewards
+
+    Make sure your environment is pickle-able:
+        Garage pickles the environment via the `pickle` or `cloudpickle` module
```
ryanjulian, line 24: it uses cloudpickle only

src/garage/envs/environment.py
```diff
@@ -0,0 +1,189 @@
...
+    # Set these in ALL subclasses
+    action_space = None
+    observation_space = None
+
+    @abc.abstractmethod
+    def reset(self):
+        """Resets the environment and returns the first `TimeStep` of sequence.
+
+        Returns:
+            garage.TimeStep: The first TimeStep. Attributes:
```
ryanjulian, line 49: you don't need to document the attributes of predefined data types -- you can just give the type itself. That's one of the benefits of defining ADTs :).

src/garage/envs/environment.py
```diff
@@ -0,0 +1,189 @@
...
+        This method will also start a new sequence if called after the
+        environment has been constructed and `reset` has not been called.
+        Again, in this case `action` will be ignored.
+
+        Args:
+            action (np.ndarray): A NumPy array, or a nested dict, list or tuple
+                of arrays corresponding to `action_spec()`.
+
+        Returns:
+            garage.TimeStep: The TimeStep after the action. Attributes:
```
ryanjulian, line 85: No need to summarize all these fields

src/garage/envs/environment.py
```diff
@@ -0,0 +1,189 @@
+"""Base Garage Environment API."""
+
+import abc
+
+
+class Environment(abc.ABC):
```
ryanjulian, line 6: how does it look if you, for instance, try to rewrite DmControlEnv to match this interface?

src/garage/_dtypes.py
```diff
@@ -428,22 +431,29 @@ def next_observations(self):
             ]))
 
 
+class StepType(enum.IntEnum):
```
ryanjulian, line 434: when you introduce this, you should also modify `TimeStepBatch` and `TrajectoryBatch` to use it.

src/garage/envs/environment.py
AiRuiChen: For garage, only `spec` from `gym.Env` is used. In fact `GarageEnv` provides its own way of returning `spec`. I will only keep `action_space`, `observation_space`, `spec` and `render_modes` as the abstract attributes for this API.

src/garage/envs/environment.py
AiRuiChen: Yea in garage we actually don't do `with GarageEnv() as env:`. We use `with LocalTFRunner(ctxt) as runner:` and pass the environment. This is in fact not necessary - I will remove it.

src/garage/envs/environment.py
AiRuiChen: Yes it should close windows. I chose to have a `__del__` here and update the docstring of `close()`.

src/garage/envs/environment.py
AiRuiChen: I think we can call this once, and let reset() handles update of the visualization. 

Copied workflow from the doc string:
```
env.reset() # reset the environment
env.visualize() # open a window for visualization
act_space = env.action_space
action = act_space.sample() # sample a random action
env.step(action) # step the environment
# the display should be updated as well
env.close() # close the environment and the visualization
```

src/garage/_dtypes.py
AiRuiChen: Thanks for this very helpful comment. Since we are certain that `reset()` and `step()` should return `TimeStep`, I will create a new PR just for TimeStep as well as the implementations, and see how different choices work. 

This PR now only introduces src/garage/environment.py

src/garage/envs/__init__.py
```diff
@@ -13,6 +14,7 @@
     'GarageEnv',
     'Step',
     'EnvSpec',
+    'Environment',
```
ryanjulian, line 17: i think this should actually be defined in `src/garage/_environment.py` and exported as `garage.Environment`

src/garage/envs/environment.py
```diff
@@ -0,0 +1,188 @@
...
+    """The main API for garage environments.
+
+    The main public API methods are:
```
ryanjulian, line 9: instead of saying "main", I recommend you make this statement unqualified "The public API methods are..." -- to me, "main" implies that there might be more, which makes me anxious as someone who wants to make sure I implement the full API.

src/garage/_dtypes.py
ryanjulian: Did you do this exercise? What were the results? I think it's a real design concern.

src/garage/envs/environment.py
ryanjulian: I think a minimal example of the rollout loop like this one would be extremely helpful in the docstrings here. It is the most powerful way of showing people how your API is meant to be used.

src/garage/envs/environment.py
ryanjulian: i think you should do this exercise before merging this PR. We can only get so far speculating how to design an API -- the real proof is trying to implement it and use it.

src/garage/envs/environment.py
```diff
@@ -0,0 +1,188 @@
...
+        """
+
+    @property
+    @abc.abstractmethod
+    def observation_space(self):
+        """Return the observation space specification.
+
+        Some environments do not provide their observation space
+        specifications as type akro.Space. In such cases, subclasses should
+        silently convert the environment's observation_space to akro.Space.
```
ryanjulian, line 52: "silently" converts no longer appies -- the types here are akro.Space, full stop.

src/garage/envs/environment.py
```diff
@@ -0,0 +1,188 @@
...
+
+        Returns:
+            akro.Space: The envionrment specification.
```
ryanjulian, line 55: spelling

src/garage/envs/environment.py
```diff
@@ -0,0 +1,188 @@
...
+    @abc.abstractmethod
+    def action_space(self):
+        """Return the action space specification.
+
+        Some environments do not provide their action space specifications as
+        type akro.Space. In such cases, subclasses should silently convert
+        the environment's action_space to akro.Space.
+
+        Returns:
+            akro.Space: The envionrment specification.
```
ryanjulian, line 41: spelling

src/garage/envs/environment.py
```diff
@@ -0,0 +1,188 @@
...
+        """
+
+    @property
+    @abc.abstractmethod
+    def spec(self):
+        """Return the environment specification.
+
+        Returns:
+            garage.envs.env_spec.EnvSpec: The envionrment specification.
```
ryanjulian, line 64: if this type isn't available as garage.EnvSpec, let's make it available there.

src/garage/envs/environment.py
```diff
@@ -0,0 +1,188 @@
...
+                `self.render_modes`.
+
+        """
+
+    @abc.abstractmethod
+    def visualize(self):
+        """Creates a visualization of the environment.
+
+        This function should be called only once after reset() to set up the
+        visualization display. Update of the visualization should be done in
```
ryanjulian, line 156: i think you should make very explicit here that calling "close()" should deallocate any resources and close any windows created by calling this. 

you may consider providing a way of closing the visualization without closing the environment. 

@krzentner @ahtsan WDYT?

src/garage/envs/environment.py
```diff
@@ -0,0 +1,188 @@
...
+            class MyEnv(Env):
+                def render_modes(self):
+                    return ['rgb_array', 'ansi']
+
+                def render(self, mode):
+                    if mode == 'rgb_array':
+                        return np.array(...) # return RGB frame for video
+                    elif mode == 'ansi':
+                        ... # return text output
+                    else:
```
ryanjulian, line 141: i'm not sure what the else case here makes clear your intention?

I think the idiomatic thing to do here (and what we should encourage implementers to do) is explicitly raise a `ValueError` which says (1) which mode you got and (2) which modes you support.

src/garage/envs/environment.py
```diff
@@ -0,0 +1,188 @@
...
+    """The main API for garage environments.
+
+    The main public API methods are:
+        * step()
+        * reset()
+        * render()
+        * visualize()
+        * close()
+
+    Set the following attributes:
```
ryanjulian, line 16: please make these properties rather than attributes. that way they are immutable.

src/garage/envs/environment.py
```diff
@@ -0,0 +1,188 @@
...
+        * action_space: The action space specification
+        * observation_space: The observation space specification
+        * spec: The environment specifications
+        * render_modes: The list of supported render modes
+
+    Make sure your environment is pickle-able:
```
ryanjulian, line 22: does this actually render well in the docs? i think you might need to adjust the markup here.

typically in Python docs this kind of info is presented as an admonition. Example: https://recommonmark.readthedocs.io/en/latest/auto_structify.html#recommonmark.transform.AutoStructify

src/garage/envs/environment.py
```diff
@@ -0,0 +1,188 @@
...
+        Garage pickles the environment via the `cloudpickle` module
+        to save snapshots of the experiment. However, some environments may
+        contain attributes that are not pickle-able (e.g. a client-server
+        connection). In such cases, override `__getstate__()` and
+        `__setstate()__` to add your custom pickle logic.
```
ryanjulian, line 27: `__setstate__()` and `__getstate__()`

ryanjulian: you might also mention this EzPickle example as a lightweight way of achieving the pickle functionality most environment implementers want: https://github.com/openai/gym/blob/master/gym/utils/ezpickle.py

src/garage/envs/environment.py
AiRuiChen: I think "closing the visualization without closing the environment" is definitely a common use case. However, I am not sure if such functionality should be placed in the main API. 

For instance if an environment opens many visualization windows, then something like `close_visualization()` will need to determine which window to close. It may also need extra parameters.

These details can be left to subclasses implementations. In my opinion the main API should do less and simply say that close() will close _all_ visualizations created by `visualize()`.

ryanjulian: that's fine with me. it can always be revised.

in the case of windows, a user can just click [x] :)

src/garage/envs/environment.py
ryanjulian: for instance, you can look at the metaworld README: https://github.com/rlworkgroup/metaworld

src/garage/envs/environment.py
AiRuiChen: I can't get this to render nicely on ReadTheDocs. I tried https://recommonmark.readthedocs.io/en/latest/auto_structify.html#codeblock-extensions but it is not working. Any examples?

src/garage/envs/environment.py
AiRuiChen: Please see my latest commit. The code only serves as a code snippet example as changes from TimeStep are not included in the PR. 

This is a great exercise as I realized it makes more sense to let the environment know `max_n_steps` so it can produce TIMEOUT time step directly. 

Pls let me know your thoughts;)

src/garage/envs/dm_control/dm_control_env.py
```diff
@@ -1,53 +1,115 @@
+# pylint: skip-file
```
ryanjulian, line 1: please make this file conforming before submit

src/garage/_environment.py
```diff
@@ -0,0 +1,243 @@
...
+    +=======================+=================================================+
+    | action_space          | The action space specification                  |
+    +-----------------------+-------------------------------------------------+
+    | observation_space     | The observation space specification             |
+    +-----------------------+-------------------------------------------------+
+    | spec                  | The environment specifications                  |
+    +-----------------------+-------------------------------------------------+
+    | render_modes          | The list of supported render modes              |
+    +-----------------------+-------------------------------------------------+
+    | max_n_steps           | The maximum number of steps (i.e. time limit)   |
```
ryanjulian, line 38: perhaps `max_episode_steps` (as in gym)?

src/garage/_environment.py
```diff
@@ -0,0 +1,247 @@
...
+        """Return the environment specification.
+
+        Returns:
+            garage.EnvSpec: The environment specification.
+
+        """
+
+    @property
+    @abc.abstractmethod
+    def max_episode_steps(self):
```
krzentner, line 95: I wonder if this should be part of `EnvSpec` instead? It's often needed in similar places.

src/garage/envs/environment.py
ryanjulian: looks great! can you include a rewrite of PointEnv and/or GridWorldEnv?

src/garage/_environment.py
```diff
@@ -0,0 +1,247 @@
...
+        `step()`.
+
+        Calling `close()` will deallocate any resources and close any
+        windows created by `visualize()`. If `close()` is not explicitly
+        called, the visualization will be closed when the environment is
+        destructed (i.e. garbage collected).
+
+        A typical workflow:
+
+            ts = env.reset()  # reset the environment
```
ryanjulian, line 220: you need to enclose these in code fences:

````
```py
f = a + b
```
````

File: src/garage/_environment.py (hunk 34 above)
Comment (ryanjulian) [reply src/garage/_environment.py:95]: i agree -- it's an important part of the EnvSpec because it's something you are likely to need to know to construct a function approximator associated with this environment.

**Comment by ryanjulian** [comment]
This is looking really great.

Please seek out input from others on the remaining issues and report back on your conclusions and with an updated PR(s) integrating those conclusions:
* `max_path_length` in `EnvSpec` or not? -- take a look around the codebase for where EnvSpec and GarageEnv appear and see how frequently it's along with `max_path_length`. If it's a lot, probably make it part of `EnvSpec`.
* Naming convention for episodes/paths/trajectories and steps/transitions and associated names (e.g. `max_episode_steps`? `max_trajectory_transitions`? etc.) -- come up with a few consistent name sets (e.g. `Transition`, `Trajectory`, `max_trajectory_transitions`, etc.) and ask the channel
* How does the use of an `Enum` for `TERMINAL`/`TIMEOUT` affect the off-policy algorithms? (ask @avnishn @maliesa96 @ahtsan @krzentner )

Please also add
* An implementation of rollout (even just as a test)
* Convert one or more environments (perhaps GridWorldEnv or PointEnv). This could also appear only as a test if you want to avoid refactors.

Generate issues for:
* Updating the samplers/rollout functions to use the new API
* Moving GarageEnv to GymEnv and making it implement this API
* Changing API/type names to match the naming convention you settle on
* Moving the remainder of the Environment wrappers to use the new Environment API

File: src/garage/_environment.py [hunk 36]
```diff
@@ -0,0 +1,247 @@
...
+        * ansi: Return a string (str) or StringIO.StringIO containing a
+          terminal-style text representation. The text can include newlines
+          and ANSI escape sequences (e.g. for colors).
+
+        Make sure that your class's `render_modes` includes the list of
+        supported modes.
+
+        For example:
+
+            class MyEnv(Env):
```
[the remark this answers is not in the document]
AiRuiChen: Good catch!

src/garage/_environment.py
AiRuiChen: I found that this doesn't work. But this works:
````
```
.. code-block:: python

    f = a + b 
```
````

File: src/garage/_dtypes.py (hunk 4 above)
Comment (AiRuiChen) [reply src/garage/_dtypes.py]: See https://github.com/rlworkgroup/garage/pull/1786 for refactoring implementation.

**Comment by AiRuiChen** [comment]
1. Naming convention: `max_episode_length`
Copying comments from @ryanjulian :

> after taking a look a few RL libraries, i think episode is the most agnostic choice. you can derive a trajectory from an episode but not the other way around. there's also no reason a trajectory need be complete. i recommend we use episodes primarily so that trajectory-oriented algorithms don't suffer a name collision. I
 advocate for episode , max_episode_steps, and TimeStep, which implies also TrajectoryBatch --> EpisodeBatch and PathBuffer --> EpisodeBuffer

Key point: 

> You can derive a trajectory from an episode but not the other way around

2. `EnvSpec` is frequently passed along with `max_path_length` into algorithms. Decision: `max_episode_length` goes in `EnvSpec`. 
3. [#1786  ](https://github.com/rlworkgroup/garage/pull/1786) will focus on `TimeStep` enum choice and implementation. This PR only introduces the new API.
4. Rewrite of `DmControlEnv`, `GridWorldEnv`, `PointEnv` will be done in a separate PR after [#1786  ](https://github.com/rlworkgroup/garage/pull/1786) is merged. Otherwise definition of TimeStep is not done and environment can't return `TimeStep`.

File: src/garage/envs/dm_control/dm_control_env.py [hunk 37]
```diff
@@ -9,32 +11,69 @@
...
-    """
-    Binding for `dm_control <https://arxiv.org/pdf/1801.00690.pdf>`_
-    """
+    """Binding for `dm_control <https://arxiv.org/pdf/1801.00690.pdf>`."""
 
     def __init__(self, env, name=None):
+        """Create a DmControlEnv.
+
+        Args:
+            env (dm_control.suit.Task): The wrapped dm_control environment.
```
ryanjulian, line 20: suite?

src/garage/_environment.py
```diff
@@ -0,0 +1,226 @@
...
+            See render() for a list of modes.
+
+        """
+
+    @abc.abstractmethod
+    def reset(self):
+        """Resets the environment and returns the first `TimeStep` of sequence.
+
+        Returns:
+            garage.TimeStep: The first time step. Note that this `TimeStep`
```
ryanjulian, line 122: generally in the case of garage types you can omit the `garage.foo.bar` stem. For external code, include the full package path.

src/garage/envs/dm_control/dm_control_env.py
```diff
@@ -49,17 +88,32 @@ def render(self, mode='human'):
...
+            observation (np.ndarray): the observation
+
+        Returns:
+            np.ndarray: the flattened dimension of observation.
+        """
         return np.sum(int(np.prod(v.shape)) for k, v in observation.items())
 
     @property
     def action_space(self):
+        """Return action_space.
```
ryanjulian, line 112: see my previous comment about how to document properties (i.e. like attributes)

src/garage/envs/dm_control/dm_control_env.py
```diff
@@ -49,17 +88,32 @@ def render(self, mode='human'):
...
 
     def close(self):
+        """Close the environment."""
         if self._viewer:
             self._viewer.close()
         self._env.close()
         self._viewer = None
         self._env = None
 
+    # pylint: disable=no-self-use
```
ryanjulian, line 98: why not just make it helper in this module?

src/garage/envs/env_spec.py
```diff
@@ -54,3 +59,24 @@ def observation_space(self, observation_space):
...
+    def max_episode_steps(self):
+        """Get max episode steps.
+
+        Returns:
+            int: The maximum number of steps that an episode
+
+        """
+        return self.max_episode_steps
+
+    @max_episode_steps.setter
```
ryanjulian, line 73: please don't allow for setters. These should be immutable. We have #1239 open for the other setters in this file (we can probably remove them easily after you're done with this new API).

src/garage/envs/env_spec.py
```diff
@@ -9,10 +9,15 @@ class EnvSpec(InOutSpec):
...
         action_space (akro.Space): The action space of the env.
+        max_episode_steps (int): The maximum number of steps that an episode.
 
     """
 
-    def __init__(self, observation_space, action_space):
+    def __init__(self,
+                 observation_space,
+                 action_space,
+                 max_episode_steps=None):
```
ryanjulian, line 19: use `math.inf` instead so that comparisons never have to check for `None`

src/garage/_environment.py
```diff
@@ -0,0 +1,226 @@
...
+        the environment's action_space to `akro.Space`.
+
+        Returns:
+            akro.Space: The environment specification.
+
+        """
+
+    @property
+    @abc.abstractmethod
+    def observation_space(self):
```
ryanjulian, line 85: see my previous comment about how to document properties (i.e. as if they are attributes)

src/garage/_environment.py
```diff
@@ -0,0 +1,226 @@
...
+        """Return the observation space specification.
+
+        Some environments do not provide their observation space
+        specifications as type `akro.Space`. In such cases, subclasses should
+        silently convert the environment's observation_space to `akro.Space`.
+
+        Returns:
+            akro.Space: The environment specification.
```
ryanjulian, line 93: this isn't an env spec?

AiRuiChen: The observation space is akro.Space. And spec is EnvSpec?

ryanjulian: ...but the `observation_space` is not `The environment specification`?

By the way, I don't think you need to say anything about "silently convert" -- this is the case with all APIs.

AiRuiChen: Ah I see. Will change to observation space specification.

src/garage/_environment.py
```diff
@@ -0,0 +1,201 @@
...
+
+    @abc.abstractmethod
+    def step(self, action):
+        """Steps the environment with the action and returns a `TimeStep`.
+
+        If the environment returned the last `TimeStep` of a sequence (either
+        of type TERMINAL or TIMEOUT) at the previous step, this call to
+        `step()` will start a new sequence and `action` will be ignored.
+
+        This method will also start a new sequence if called after the
```
ryanjulian, line 111: i don't agree with this stance. Instead, I think behavior if `reset()` has not been called should be undefined or lead to an exception.

src/garage/_environment.py
```diff
@@ -0,0 +1,201 @@
...
+        `step_type==StepType.TIMEOUT`.
+
+        If possible, update the visualization display as well.
+
+        Args:
+            action (np.ndarray): A NumPy array, or a nested dict, list or tuple
+                of arrays corresponding to `action_space`.
+
+        Returns:
+            TimeStep: The time step after the action.
```
ryanjulian, line 126: it is the time step resulting from the action

src/garage/_environment.py
```diff
@@ -0,0 +1,201 @@
...
+        Again, in this case `action` will be ignored.
+
+        If `spec.max_episode_length` is reached after applying the action
+        and the sequence is not done, `step()` should return a `TimeStep` with
+        `step_type==StepType.TIMEOUT`.
+
+        If possible, update the visualization display as well.
+
+        Args:
+            action (np.ndarray): A NumPy array, or a nested dict, list or tuple
```
ryanjulian, line 122: perhaps just type this `object` for now.

src/garage/_environment.py
```diff
@@ -0,0 +1,201 @@
...
+                of arrays corresponding to `action_space`.
```
ryanjulian, line 123: conforming to action_space

src/garage/_environment.py
```diff
@@ -0,0 +1,201 @@
...
+        If the environment returned the last `TimeStep` of a sequence (either
+        of type TERMINAL or TIMEOUT) at the previous step, this call to
+        `step()` will start a new sequence and `action` will be ignored.
+
+        This method will also start a new sequence if called after the
+        environment has been constructed and `reset()` has not been called.
+        Again, in this case `action` will be ignored.
+
+        If `spec.max_episode_length` is reached after applying the action
+        and the sequence is not done, `step()` should return a `TimeStep` with
```
ryanjulian, line 116: and the environment has not terminated the episode

src/garage/_environment.py
```diff
@@ -0,0 +1,201 @@
...
+        """
+
+    @abc.abstractmethod
+    def render(self, mode):
+        """Renders the environment.
+
+        The set of supported modes varies per environment. By convention,
+        if mode is:
+
+        * rgb_array: Return an `numpy.ndarray` with shape (x, y, 3),
```
ryanjulian, line 137: what is the type? usually uint8 i believe

src/garage/envs/dm_control/dm_control_env.py
```diff
@@ -8,33 +10,82 @@
...
                    name='{}.{}'.format(domain_name, task_name))
 
     def step(self, action):
+        """Step the environment.
+
+        Args:
+            action (np.ndarray): input action
+
+        Returns:
+            Step: The time step after applying this action.
```
ryanjulian, line 60: TimeStep?

src/garage/envs/dm_control/dm_control_env.py
```diff
@@ -8,33 +10,82 @@
...
     @classmethod
     def from_suite(cls, domain_name, task_name):
+        """Create a DmControl task given the domain name and task name.
+
+        Args:
+            domain_name (str): Domain name
+            task_name (str): Task name
+
+        Return:
+            dm_control.suit.Task: the dm_control task environment
```
ryanjulian, line 48: it returns an `Environment` (the garage type)

src/garage/envs/dm_control/dm_control_env.py
```diff
@@ -8,33 +10,82 @@
...
         return Step(
             flatten_observation(time_step.observation)['observations'],
             time_step.reward, time_step.step_type == StepType.LAST,
             **time_step.observation)
 
     def reset(self):
+        """Reset the environment.
+
+        Returns:
+            Step: The first time step.
```
ryanjulian, line 72: TimeStep -- this is your own interface!

src/garage/envs/dm_control/dm_control_env.py
```diff
@@ -8,33 +10,82 @@
...
+        """Render the environment.
+
+        Args:
+            mode (str): render mode.
+
+        Returns:
+            np.ndarray: if mode is 'rgb_array', else return None.
+
+        Raises:
+            NotImplementedError: if mode is not supported.
```
ryanjulian, line 87: this is a ValueError. see definitions here: https://docs.python.org/3/library/exceptions.html#ValueError

src/garage/envs/dm_control/dm_control_env.py
```diff
@@ -49,17 +100,16 @@ def render(self, mode='human'):
...
         self._env.close()
         self._viewer = None
         self._env = None
 
-    def _flat_shape(self, observation):
-        return np.sum(int(np.prod(v.shape)) for k, v in observation.items())
-
     @property
     def action_space(self):
+        """gym.Space: the action space specification."""
```
ryanjulian, line 112: it returns akro.Space!

src/garage/envs/dm_control/dm_control_env.py
```diff
@@ -8,33 +10,82 @@
...
+    Args:
+        observation (np.ndarray): the observation
+
+    Returns:
+        np.ndarray: the flattened dimension of observation.
     """
+    return np.sum(int(np.prod(v.shape)) for k, v in observation.items())
+
+
+class DmControlEnv(gym.Env):
```
ryanjulian, line 25: why not just make it garage.Environment if you're going to touch so much code?

src/garage/envs/env_spec.py
```diff
@@ -9,10 +11,15 @@ class EnvSpec(InOutSpec):
     Args:
         observation_space (akro.Space): The observation space of the env.
         action_space (akro.Space): The action space of the env.
+        max_episode_length (int): The maximum number of steps that an episode.
```
ryanjulian, line 14: incomplete sentence?

src/garage/envs/dm_control/dm_control_env.py
AiRuiChen: I can't because changes of TimeStep are not merged yet. I touched the code for exercise and later reverted the changes. I have to fix the style to commit it.

I'd prefer to leave dm_control_env as it is, then rewrite it with the new API in a new PR.

src/garage/_environment.py
AiRuiChen: I agree. It should raise a `RuntimeError`.

gitanshu:
@Mergifyio rebase

gitanshu:
@Mergifyio rebase

src/garage/envs/dm_control/dm_control_env.py
```diff
@@ -8,33 +10,82 @@
...
             **time_step.observation)
 
     def reset(self):
+        """Reset the environment.
+
+        Returns:
+            Step: The first time step.
+        """
         time_step = self._env.reset()
         return flatten_observation(time_step.observation)['observations']
```
krzentner, line 75: I think there needs to be a dict here too.

AiRuiChen: I will rewrite DmControlEnv as well as other environments in a new PR. This file is in the commit because I once made experimental changes here. But it is basically the same as the one in master, with style/documentation fixes.

AiRuiChen:
@Mergifyio rebase

PR #1779, opened 2020-07-18
Refactor tf.q_function

### LINKED ISSUE #1632: QFunctions in tf tree should inherit from garage.tf.Model
- [ ] DiscreteCNNQFunction
- [ ] DiscreteMLPQFunction
- [ ] ContinuousCNNQFunction
- [ ] ContinuousMLPQFunction

yeukfu:
This should close #1632.


src/garage/tf/q_functions/continuous_cnn_q_function.py
```diff
@@ -154,9 +153,8 @@ def _initialize(self):
                                               (None, ) + self._obs_dim,
                                               name='state')
             augmented_obs_ph = obs_ph
-        with tf.compat.v1.variable_scope(self.name) as vs:
-            self._variable_scope = vs
-            outputs = self.model.build(augmented_obs_ph, action_ph).outputs
+        print(action_ph)
```
ryanjulian, line 156: remove?

src/garage/tf/q_functions/q_function.py
```diff
@@ -3,17 +3,9 @@
...
-
-    """
-
-    def __init__(self, name):
-        self.name = name or type(self).__name__
-        self._variable_scope = None
+    """Q-function base class without Parameterzied."""
 
+    @abc.abstractmethod
     def get_qval_sym(self, *input_phs):
```
ahtsan, line 9: Do we still need this function in the interface? Isn’t it equal to .build()?

tests/garage/tf/algos/test_td3.py
```diff
@@ -71,4 +71,4 @@ def test_td3_pendulum(self):
 
             runner.setup(algo, env, sampler_cls=LocalSampler)
             last_avg_ret = runner.train(n_epochs=10, batch_size=250)
-            assert last_avg_ret > 300
+            assert last_avg_ret > 100
```
ahtsan, line 74: Could it mean something is not working? What’s the average return before and after the change?

src/garage/tf/q_functions/q_function.py
yeukfu: Yes, it is equal to build(). It seems that we can completely remove the QFunction interface? As the remaining method is clone().

ahtsan: Sounds good to me

tests/garage/tf/algos/test_td3.py
yeukfu: I run 20 tests respectively in my local machine. The `last_avg_ret` is:
- Before change: 
mean = 425.04948958342857
std = 69.67050410993383
- After change:
mean = 430.92078589330924
std = 82.32598708139051
So I think this is not because this change.

ahtsan: It seems really flaky. Is there already an issue for that? Can you set it to 200?

yeukfu: In [last build](https://travis-ci.com/github/rlworkgroup/garage/jobs/362404060), it fails. The `last_avg_ret` is 151.178

PR #1786, opened 2020-07-22
Add StepType and remove terminal

AiRuiChen:
This PR comes with [#1775 ](https://github.com/rlworkgroup/garage/pull/1775)
1. Add Enum `StepType`
2. Remove `terminal` from time step, since it can be inferred from `StepType.TERMINAL`. In the new API, `reset()` and `step()` returns a `TimeStep`. Downstream consumers of `TimeStep` will need to check `TimeStep.step_type` for terminal. 
3. Similarly, `terminals` is removed from `TimeStepBatch` and `TrajectoryBatch`.
4. Consumers can tell if a `TimeStep` is a `TERMINAL`, `TIMEOUT` or simply the last (i.e. `TERMINAL` or `TIMEOUT) with its helper functions.

Will add usage implementation & test once the changes of TimeStep itself are approved.

### DIFF
--- a/src/garage/_dtypes.py
+++ b/src/garage/_dtypes.py
@@ -1,10 +1,13 @@
 """Data types for agent-based learning."""
 import collections
+import enum
 
 import akro
 import numpy as np
 
 from garage.misc import tensor_utils
+
+# pylint: disable=too-many-lines
 
 
 class TrajectoryBatch(
@@ -14,9 +17,9 @@ class TrajectoryBatch(
             'last_observations',
             'actions',
             'rewards',
-            'terminals',
             'env_infos',
             'agent_infos',
+            'step_types',
             'lengths',
         ])):
     # pylint: disable=missing-return-doc, missing-return-type-doc, missing-param-doc, missing-type-doc  # noqa: E501
@@ -61,9 +64,6 @@ class TrajectoryBatch(
         rewards (numpy.ndarray): A numpy array of shape
             :math:`(N \bullet [T])` containing the rewards for all time steps
             in this batch.
-        terminals (numpy.ndarray): A boolean numpy array of shape
-            :math:`(N \bullet [T])` containing the termination signals for all
-            time steps in this batch.
         env_infos (dict): A dict of numpy arrays arbitrary environment state
             information. Each value of this dict should be a numpy array of
             shape :math:`(N \bullet [T])` or :math:`(N \bullet [T], S^*)`.
@@ -71,6 +71,9 @@ class TrajectoryBatch(
             state information. Each value of this dict should be a numpy array
             of shape :math:`(N \bullet [T])` or :math:`(N \bullet [T], S^*)`.
             For example, this may contain the hidden states from an RNN policy.
+        step_types (numpy.ndarray): A numpy array of `StepType with shape
+            :math:`(N,)` containing the time step types for all transitions in
+            this batch.
         lengths (numpy.ndarray): An integer numpy array of shape :math:`(N,)`
             containing the length of each trajectory in this batch. This may be
             used to reconstruct the individual trajectories.
@@ -83,7 +86,7 @@ class TrajectoryBatch(
     __slots__ = ()
 
     def __new__(cls, env_spec, observations, last_observations, actions,
-                rewards, terminals, env_infos, agent_infos,
+                rewards, env_infos, agent_infos, step_types,
                 lengths):  # noqa: D102
         # pylint: disable=too-many-branches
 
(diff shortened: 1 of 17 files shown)


src/garage/_dtypes.py
```diff
@@ -532,19 +574,59 @@ def __new__(cls, env_spec, observation, action, reward, next_observation,
...
+
+    def mid(self):
+        """Checks if the `TimeStep` is the mid one in a sequence.
+
+        Returns:
+            bool: True if it is a MID `TimeStep`.
+        """
+        return self.step_type is StepType.MID
+
+    def terminal(self):
```
ryanjulian, line 606: these should be properties

you can document a property like an attribute, e.g.
```python
@property
def terminal(self):
    """ bool: Whether this`TimeStep` is associated with a termination condition."""
    return self.step_type is StepType.TERMINAL
```

src/garage/_dtypes.py
```diff
@@ -428,22 +437,42 @@ def next_observations(self):
...
+    Suppose max_path_length = 5:
+    * A success sequence terminated at step 4 will look like:
+        FIRST, MID, MID, TERMINAL
+    * A success sequence terminated at step 5 will look like:
+        FIRST, MID, MID, MID, TERMINAL
+    * An unsuccessful sequence truncated by time limit will look like:
+        FIRST, MID, MID, MID, TIMEOUT
+    """
+    # Denotes the first `TimeStep` in a sequence.
+    FIRST = 0
```
ryanjulian, line 455: RESET? INITIAL?

src/garage/_dtypes.py
```diff
@@ -428,22 +437,42 @@ def next_observations(self):
...
+    # Denotes any `TimeStep` in the middle of a sequence (i.e. not the first or
+    # last one).
+    MID = 1
```
ryanjulian, line 458: INTERMEDIATE?

src/garage/_dtypes.py
```diff
@@ -227,9 +219,20 @@ def __new__(cls, env_spec, observations, last_observations, actions,
                     'length {}, but got key {} with batch size {} instead.'.
                     format(inferred_batch_size, key, val.shape[0]))
 
+        # step_types
```
ryanjulian, line 222: can you check that `lengths` is consistent with `step_types`?

src/garage/_dtypes.py
AiRuiChen: Many libraries use the FIRST, MID, LAST pattern (such as [`dm_control`](https://github.com/deepmind/dm_env/blob/master/dm_env/_environment.py#L71) and `tensorflow`). I would advocate for keeping FIRST and MID so the naming is consistent. 

We also provide `last` as a property for `TimeStep`, so `garage.TimeStep` can be used in a similar way like external TimeStep. Depends on user's need they can decide whether they want to differentiate `TERMINAL` and `TIMEOUT` or not.

ryanjulian: dm_control/dm_eng and tf_agents are written by approximately same group of people (really two groups of people who work together a lot) so it's unsurprising that they use the same naming convention. Basically, I don't think that's much evidence for a concensus.

FIRST and MID make sense in the context of LAST. Once you remove LAST and replace it with TERMINAL and TIMEOUT, the naming convention becomes more surprising.

We should strive to use what we think will be least surprising to users, not cargo-culting other libraries.

AiRuiChen: Let's discuss this in standup maybe?

src/garage/_dtypes.py
AiRuiChen: If I understand your question correctly - `inferred_batch_size` is from `lengths` and this is checking if the length of `step_types` is consistent with `lengths`.

src/garage/_dtypes.py
krzentner: So the main takeaway from that standup was that FIRST is probably a bad idea anyways, because those steps act very differently from the other steps, and almost all algorithms would want to ignore those steps.

AiRuiChen: Since we decided `env.reset()` shouldn't return `TimeStep` (because then we would end up with timesteps that don't contain actions / aren't actually a step in the underlying MDP), I will remove `FIRST` from the enum.

AiRuiChen: Thoughts gathered from standup: `FIRST` should denote the first MDP time step, which is returned from the first `step()`. In other words `step()` can return step type `FIRST`/`MID`/`TERMINAL`/`TIMEOUT`.

src/garage/tf/algos/reps.py
```diff
@@ -6,14 +6,17 @@
...
-from garage import _Default, log_performance, make_optimizer, TrajectoryBatch
+from garage import _Default, log_performance, make_optimizer
 from garage.np.algos import RLAlgorithm
 from garage.sampler import RaySampler
 from garage.tf import paths_to_tensors
 from garage.tf.misc import tensor_utils
 from garage.tf.misc.tensor_utils import flatten_inputs, graph_inputs
 from garage.tf.optimizers import LbfgsOptimizer
 
+# skip isort because isort result conflicts with yaqf auto-format
```
ryanjulian, line 17: the correct solution here is to disable YAPF, not isort. see several files in this same PR for an example.

unfortunately, whitespace matters.

example:
```py
"""My cool module."""
# yapf: disable
import os
import sys

import numpy as np

from garage import StepType
                                                              # <-- note the newline here
# yapf: enable
```

ryanjulian:
Please address last comments before submit

ryanjulian:
IMO this is ready to merge once my latest comment is addressed

AiRuiChen:
> IMO this is ready to merge once my latest comment is addressed

Looks like there's something wrong with the merge queue. Can you help force merge this?

ryanjulian:
@Mergifyio refresh

PR #1787, opened 2020-07-22
a policy simulation script that works with snapshotted policies

chelseas:
The examples/sim_policy.py script didn't work with the torch training example script I ran, so I created a sim_policy script using code from the docs that works for snapshotted policies.

### DIFF
--- a/examples/simulate_policy.py
+++ b/examples/simulate_policy.py
@@ -1,0 +1,19 @@
+# Load the policy and the env in which it was trained
+from garage.experiment import Snapshotter
+import tensorflow as tf # optional, only for TensorFlow as we need a tf.Session
+import fire
+
+def run_rollouts(file, animated=True):
+    snapshotter = Snapshotter()
+    with tf.compat.v1.Session(): # optional, only for TensorFlow
+        data = snapshotter.load(file)
+    policy = data['algo'].policy
+    env = data['env']
+
+    # See what the trained policy can accomplish
+    from garage.sampler.utils import rollout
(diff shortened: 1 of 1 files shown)


ryanjulian:
@chelseas thank you so much for the PR! We'll take a look ASAP.

Did you find the existing example is broken?

chelseas:
Yea the sim_policy.py script didn't work when I ran the pytorch PPO+cartpole example, so I wrote this new one using the docs

chelseas:
The code, which largely comes from here:
https://garage.readthedocs.io/en/latest/user/reuse_garage_policy.html

Fails with this error when using a tf policy:
```
RuntimeError: Attempted to use a closed Session.
```

when I change things to:
```
with tf.compat.v1.Session() as sess: # optional, only for TensorFlow
    data = snapshotter.load(file)
```
I get this error:
```
ValueError: Variable policy/mlp/hidden_0/kernel already exists, disallowed. Did you mean to set reuse=True or reuse=tf.AUTO_REUSE in VarScope?
```

Perhaps the person who wrote the snapshotter would better be able to debug why this is happening?

examples/simulate_policy.py
```diff
@@ -0,0 +1,19 @@
+# Load the policy and the env in which it was trained
+from garage.experiment import Snapshotter
+import tensorflow as tf # optional, only for TensorFlow as we need a tf.Session
+import fire
```
ryanjulian, line 4: can you use click, which we already have as a dependency?

PR #1791, opened 2020-07-24
Fix qf_derived_policy variable_scope

ahtsan:
Currently qf_derived_policy has an empty variable_scope, which makes `get_params()` returns an empty list.
This PR fixes the issue by assigning the underlying QFunction variable_scope to its variable_scope.

### DIFF
--- a/src/garage/tf/policies/discrete_qf_derived_policy.py
+++ b/src/garage/tf/policies/discrete_qf_derived_policy.py
@@ -30,7 +30,6 @@ class DiscreteQfDerivedPolicy(Module, Policy):
 
     def _initialize(self):
         # pylint: disable=protected-access
-        self._variable_scope = self._qf._variable_scope
         self._f_qval = tf.compat.v1.get_default_session().make_callable(
             self._qf.q_vals, feed_list=[self._qf.input])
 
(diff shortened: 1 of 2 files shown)


src/garage/tf/policies/discrete_qf_derived_policy.py
```diff
@@ -29,10 +29,10 @@ def __init__(self, env_spec, qf, name='DiscreteQfDerivedPolicy'):
         self._initialize()
 
     def _initialize(self):
-        with tf.compat.v1.variable_scope(self.name, reuse=False) as vs:
-            self._variable_scope = vs
-            self._f_qval = tf.compat.v1.get_default_session().make_callable(
-                self._qf.q_vals, feed_list=[self._qf.input])
+        # pylint: disable=protected-access
+        self._variable_scope = self._qf._variable_scope
```
ryanjulian, line 33: maybe we should just add a variable_scope property to `Model`?

ryanjulian, line 33: i think the more idiomatic thing to do is override get_params, because this is a special case, right?

ahtsan: It also needs to override other functions which are also based on `_variable_scope` since `_variable_scope` is what those functions are really about. But to make sure all functions work, regardless of whether it depends on `_variable_scope`, it makes sense to override them explicitly.

ryanjulian: But why do `_qf` and `DiscreteQFDerivedPolicy` need to share a variable scope?

Can't we have `DiscreteQFDerivedPolicy` just overwrite all of the methods which use a variable scope to instead pass info from the inner Q-function? Isn't that the pattern we would use anywhere else (e.g. an environment wrapper)?

ahtsan: Right

ahtsan:
@Mergifyio rebase

ahtsan:
@Mergifyio rebase

ahtsan:

ahtsan:

PR #1807, opened 2020-07-26
Use cudnn7-runtime for run-nvidia

ryanjulian:
This is needed to make TF work in Docker+CUDA

TODO
* [x] Change docker/hooks
* [x] Update docs
* [ ] Document testing procedure which would have caught this? (#1829 opened for this)


ryanjulian:

PR #1812, opened 2020-07-26
Flatten observations in the policy if needed

avnishn:
The current behavior of garage is for samplers,
algorithms, and users to flatten their observations
before feeding them into policies. This is because
garage policies flatten their observation spaces
and expect observations from that flattened observation
space.

The proposed change is to flatten observations inside of
the policies if needed. Observations should be flattened
only if they arent numpy arrays, because numpy arrays are
the type of observation that would be returned by a flattened
observation box space.

This makes ddpg-fetch-her work, however some additional changes
were needed to make it work. For example, the timelimit terminated
fetch environment only inserts timelimit truncated signals
if the environment has reached a timelimit. This doesn't work
when considering trajectory batches, which require that the
same number of keys are in every observation in a trajectory.
To get around this, if there is no time limit termination,
a time limit termination False signal is inserted.

Something that I need help on is how to document observations when they aren't 
numpy arrays. For example, if they come from dict spaces, they are dictionary observations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -211,7 +211,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `garage.np.OffPolicyRLAlgorithm` ([#1552](https://github.com/rlworkgroup/garage/pull/1552))
 
 ### Fixed
-- Bug where `GarageEnv` did not pickle ([#1029](https://github.com/rlworkgroup/garage/pull/1029))
+- Bug where `GymEnv` did not pickle ([#1029](https://github.com/rlworkgroup/garage/pull/1029))
 - Bug where `VecEnvExecutor` conflated terminal state and time limit signal (
   [#1178](https://github.com/rlworkgroup/garage/pull/1178),
   [#1570](https://github.com/rlworkgroup/garage/pull/1570))
--- a/docs/user/implement_algo.md
+++ b/docs/user/implement_algo.md
@@ -84,13 +84,21 @@ We'll also want an environment to test our algorithm with.
 
 ```py
 from garage import wrap_experiment
+<<<<<<< HEAD
 from garage.envs import PointEnv
+=======
+from garage.envs import PointEnv, GymEnv
+>>>>>>> a1877525... Flatten observations if necessary
 from garage.experiment import LocalRunner
 
 @wrap_experiment
 def debug_my_algorithm(ctxt):
     runner = LocalRunner(ctxt)
+<<<<<<< HEAD
     env = PointEnv()
+=======
+    env = GymEnv(PointEnv())
+>>>>>>> a1877525... Flatten observations if necessary
     algo = MyAlgorithm()
     runner.setup(algo, env)
     runner.train(n_epochs=3)
@@ -159,7 +167,11 @@ class SimpleVPG:
             samples = runner.obtain_samples(epoch)
 
 from garage import wrap_experiment
+<<<<<<< HEAD
 from garage.envs import PointEnv
+=======
+from garage.envs import PointEnv, GymEnv
+>>>>>>> a1877525... Flatten observations if necessary
 from garage.experiment import LocalRunner
 from garage.experiment.deterministic import set_seed
 from garage.torch.policies import GaussianMLPPolicy
@@ -168,7 +180,11 @@ from garage.torch.policies import GaussianMLPPolicy
 def debug_my_algorithm(ctxt):
     set_seed(100)
     runner = LocalRunner(ctxt)
+<<<<<<< HEAD
     env = PointEnv()
+=======
+    env = GymEnv(PointEnv())
+>>>>>>> a1877525... Flatten observations if necessary
     policy = GaussianMLPPolicy(env.spec)
     algo = SimpleVPG(policy)
     runner.setup(algo, env)
@@ -306,7 +322,11 @@ policy when training, you can solve an Gym environment, for example
 def tutorial_vpg(ctxt=None):
     set_seed(100)
     runner = LocalRunner(ctxt)
+<<<<<<< HEAD
     env = GymEnv('LunarLanderContinuous-v2')
+=======
+    env = GymEnv(env_name='LunarLanderContinuous-v2')
+>>>>>>> a1877525... Flatten observations if necessary
     policy = GaussianMLPPolicy(env.spec)
     algo = SimpleVPG(env.spec, policy)
     runner.setup(algo, env)
@@ -322,7 +342,11 @@ except for the replacement of `LocalRunner` with `LocalTFRunner`.
 ```py
 ...
 from garage import wrap_experiment
+<<<<<<< HEAD
 from garage.envs import PointEnv
+=======
+from garage.envs import PointEnv, GymEnv
+>>>>>>> a1877525... Flatten observations if necessary
 from garage.experiment import LocalTFRunner
 from garage.experiment.deterministic import set_seed
 from garage.tf.policies import GaussianMLPPolicy
@@ -331,7 +355,11 @@ from garage.tf.policies import GaussianMLPPolicy
 def tutorial_vpg(ctxt=None):
     set_seed(100)
     with LocalTFRunner(ctxt) as runner:
+<<<<<<< HEAD
         env = PointEnv()
+=======
+        env = GymEnv(PointEnv())
+>>>>>>> a1877525... Flatten observations if necessary
         policy = GaussianMLPPolicy(env.spec)
         algo = SimpleVPG(env.spec, policy)
         runner.setup(algo, env)
@@ -496,7 +524,11 @@ from garage.tf.policies import CategoricalMLPPolicy
 def tutorial_cem(ctxt=None):
     set_seed(100)
     with LocalTFRunner(ctxt) as runner:
+<<<<<<< HEAD
         env = GymEnv('CartPole-v1')
+=======
+        env = GymEnv(env_name='CartPole-v1')
+>>>>>>> a1877525... Flatten observations if necessary
         policy = CategoricalMLPPolicy(env.spec)
         algo = SimpleCEM(env.spec, policy)
         runner.setup(algo, env)
--- a/docs/user/meta_multi_task_rl_exp.md
+++ b/docs/user/meta_multi_task_rl_exp.md
@@ -50,7 +50,11 @@ def te_ppo_ml1_push(ctxt, seed, n_epochs, batch_size_per_task):
 
     """
     set_seed(seed)
+<<<<<<< HEAD
     envs = [normalize(GymEnv(ML1.get_train_tasks('push-v1')))]
+=======
+    envs = [GymEnv(normalize(ML1.get_train_tasks('push-v1')))]
+>>>>>>> a1877525... Flatten observations if necessary
     env = MultiEnvWrapper(envs, mode='del-onehot')
 
     latent_length = 2
@@ -184,7 +188,11 @@ def te_ppo_mt50(ctxt, seed, n_epochs, batch_size_per_task):
     """
     set_seed(seed)
     tasks = MT50.get_train_tasks().all_task_names
+<<<<<<< HEAD
     envs = [GymEnv(normalize(MT50.from_task(task))) for task in tasks]
+=======
+    envs = [normalize(GymEnv(MT50.from_task(task))) for task in tasks]
+>>>>>>> a1877525... Flatten observations if necessary
     env = MultiEnvWrapper(envs,
                           sample_strategy=round_robin_strategy,
                           mode='del-onehot')
--- a/docs/user/training_a_policy.md
+++ b/docs/user/training_a_policy.md
@@ -47,7 +47,11 @@ like [this](implement_env). In this example, we choose `CartPole-V1`
 environment.
 
 ```py
+<<<<<<< HEAD
 env = GymEnv('CartPole-v1')
+=======
+env = GymEnv(env_name='CartPole-v1')
+>>>>>>> a1877525... Flatten observations if necessary
 ```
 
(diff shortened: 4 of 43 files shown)


maliesa96:
CNN policies expected flattened observations and unflattened them if necessary. Was this feature removed?

ryanjulian:
> For example, the timelimit terminated
> fetch environment only inserts timelimit truncated signals
> if the environment has reached a timelimit.

Can you split this fix into a different PR?

src/garage/np/policies/policy_utils.py
```diff
@@ -0,0 +1,40 @@
+"""Module for utility functions commonly used by policies."""
```
ryanjulian, line 1: please do not introduce new `_utils` modules. This should go in garage.np and src/garage/np/_functions.py

src/garage/np/policies/policy_utils.py
```diff
@@ -0,0 +1,40 @@
...
+        observation_space(akro.Space): The observation space that
+            observations comes from.
+        observations(observation_space.dtype): The observations to be flattened
+            if necessary.
+
+    Returns:
+        np.ndarray: Flattened observations.
+    """
+    if isinstance(observations, list):
+        if isinstance(observations[0], np.ndarray):
```
ryanjulian, line 36: does this really tell me whether it's flat or not? what if the ndarray has more than 2 dimensions?

avnishn:
@maliesa96 no the feature was not removed. The issue is that there are observations that come from various spaces (box, dict, etc), but garage was really built around box spaces as they are the most common. The change in this pr will not change the current behavior of CNN policies.

src/garage/np/policies/policy_utils.py
avnishn: Hmm. If the ndarray has more than 2 dimensions, we don't flatten it. There are still experiments that use 2d ndarray observations. We wouldn't want to flatten those because then they wouldn't match the flattened observation space. The reason why I added this if statement is because though it is documented that a user should pass numpy array observations, there were tests that passed lists of numpy arrays to the function `get_actions`

src/garage/tf/policies/categorical_cnn_policy.py
```diff
@@ -152,6 +153,8 @@ def get_action(self, observation):
             dict(numpy.ndarray): Distribution parameters.
 
         """
+        observation = flatten_if_unflattened(self.observation_space,
```
yeukfu, line 156: Does the doctring of args need to change?

krzentner:
> The current behavior of garage is for samplers,
> algorithms, and users to flatten their observations
> before feeding them into policies. This is because
> garage policies flatten their observation spaces
> and expect observations from that flattened observation
> space.

Honestly, this isn't even consistently true. As far as I can tell, the current behavior is that `get_action` and `get_actions` are *usually* given unflattened observations, and in PyTorch `forward` is given flattened observations. In TensorFlow we construct placeholders for unflattened observations, but don't feed them correctly except for images.

src/garage/_functions.py
```diff
@@ -159,3 +159,43 @@ def log_performance(itr, batch, discount, prefix='Evaluation'):
...
+    Args:
+        observation_space(akro.Space): The observation space that
+            observation comes from.
+        observation(observation_space.dtype): The observations to be flattened
+            if necessary.
+
+    Returns:
+        np.ndarray: Flattened observations.
+    """
+    if not isinstance(observation, np.ndarray):
```
krzentner, line 176: I think this should also check that `len(observation.shape) == 1`

src/garage/_functions.py
```diff
@@ -159,3 +159,43 @@ def log_performance(itr, batch, discount, prefix='Evaluation'):
...
+            if necessary.
+
+    Returns:
+        np.ndarray: Flattened observations.
+    """
+    # if a list of observations is passed, then
+    # don't flatten them if the individual observations
+    # are numpy arrays
+    if isinstance(observations, list):
+        if isinstance(observations[0], np.ndarray):
```
krzentner, line 197: Same here.

src/garage/_functions.py
```diff
@@ -159,3 +159,43 @@ def log_performance(itr, batch, discount, prefix='Evaluation'):
...
+            return observations
+    if not isinstance(observations, np.ndarray):
```
krzentner, line 199: Should probably check that `len(observation.shape) == 2`.

src/garage/tf/policies/categorical_cnn_policy.py
krzentner, line 156: This is really confusing. Basically, the code here looks like it flattens images, but the code for doing it won't work correctly (because unflattened images are still `np.ndarray`s), and then the next line requires unflattened images. I think the right thing to do here is just to not flatten the observations.

tests/garage/sampler/test_utils.py
```diff
@@ -23,11 +23,6 @@ def test_max_path_length(self):
         assert path['agent_infos']['dummy'].shape[0] == 3
         assert path['env_infos']['dummy'].shape[0] == 3
 
-    def test_does_flatten(self):
```
krzentner, line 26: Wait, shouldn't this test still pass? Or maybe it's named wrong.

src/garage/_functions.py
```diff
@@ -159,3 +159,43 @@ def log_performance(itr, batch, discount, prefix='Evaluation'):
             tabular.record('SuccessRate', np.mean(success))
 
     return undiscounted_returns
+
+
+def flatten_if_unflattened(observation_space, observation):
+    """Flattens observation to a numpy.ndarray if it is not.
```
haydenshively, line 165: This isn't very clear, at least to me. In numpy-speak, flattened would mean `ndarray.ndim == 1`, but it looks like all this function does is ensure `isinstance(observation, np.ndarray)`. Basically, the `ndim` of the returned array is unconstrained, so "flattened" doesn't really mean what it usually means.

tests/garage/sampler/test_utils.py
avnishn: no this test checks to see that the sampler is flattening observations. After a discussion we had, the conclusion we came to was that the sampler shouldn't be doing any flattening or unflattening.

avnishn:
perhaps flatten is the wrong word to use here. As @haydenshively pointed out, 

> flattened would mean ndarray.ndim == 1, but it looks like all this function does is ensure isinstance(observation, np.ndarray)

Thats not the goal of this change. The goal is to verify that the observations are numpy arrays before feeding them into the policies.

Also @krzentner to answer your question, yes this does fix the ddpg-fetchreach-her example.

src/garage/_functions.py
```diff
@@ -159,3 +159,43 @@ def log_performance(itr, batch, discount, prefix='Evaluation'):
...
+    Args:
+        observation_space(akro.Space): The observation space that
+            observations comes from.
+        observations(observation_space.dtype): The observations to be flattened
+            if necessary.
+
+    Returns:
+        np.ndarray: Flattened observations.
+    """
+    # if a list of observations is passed, then
```
ryanjulian, line 193: why not just make this one function which checks the input type?

avnishn:
@mergifyio rebase

src/garage/tf/policies/categorical_cnn_policy.py
```diff
@@ -152,6 +153,7 @@ def get_action(self, observation):
             dict(numpy.ndarray): Distribution parameters.
 
         """
+        observation = convert_to_numpy(self.observation_space, observation)
```
krzentner, line 156: These really shouldn't be necessary for this policy, since if your input isn't a valid image, flattening it certainly won't make it into one.

src/garage/_functions.py
```diff
@@ -163,3 +163,43 @@ def log_performance(itr, batch, discount, prefix='Evaluation'):
...
+
+    Returns:
+        np.ndarray: Flattened observations.
+    """
+    if not isinstance(observation, np.ndarray):
+        observation = (observation_space.flatten(observation))
+    return observation
+
+
+def convert_n_to_numpy(observation_space, observations):
```
ryanjulian, line 185: move these to garage.np

src/garage/tf/policies/continuous_mlp_policy.py
```diff
@@ -122,9 +123,9 @@ def get_action(self, observation):
             dict: Empty dict since this policy does not model a distribution.
 
         """
-        action = self._f_prob([observation])
-        action = self.action_space.unflatten(action)
-        return action, dict()
+        actions, agent_infos = self.get_actions(np.array([observation]))
```
ryanjulian, line 126: i thought you already guaranteed the observation is an np.ndarray?

src/garage/tf/policies/discrete_qf_derived_policy.py
```diff
@@ -50,10 +50,8 @@ def get_action(self, observation):
                 len(self.env_spec.observation_space.shape):
             observation = self.env_spec.observation_space.unflatten(
                 observation)
-        q_vals = self._f_qval([observation])
-        opt_action = np.argmax(q_vals)
-
-        return opt_action, dict()
+        opt_actions, agent_infos = self.get_actions(np.array([observation]))
```
ryanjulian, line 53: why always the cast to np.ndarray?

tests/garage/sampler/test_utils.py
ryanjulian: perhaps you should just modify it to `test_does_not_flatten` instead

src/garage/tf/policies/discrete_qf_derived_policy.py
avnishn: The method `get_actions` asks for a numpy array. I thought I would stay true to the documentation where possible.

I'll probably have to update the docs to allow for the union of all the different types of akro observations.

src/garage/tf/policies/continuous_mlp_policy.py
avnishn: I will need to update the documentation to support the different observation types of akro.

src/garage/tf/policies/discrete_qf_derived_policy.py
ryanjulian: ...but why would observation not be a numpy array after flattening? wasn't that the point?

tests/garage/replay_buffer/test_her_replay_buffer.py
```diff
@@ -11,7 +12,8 @@
 class TestHerReplayBuffer:
 
     def setup_method(self):
-        self.env = DummyDictEnv()
+        self.env = GarageEnv(DummyDictEnv())
+        # import ipdb; ipdb.set_trace()
```
ryanjulian, line 16: remove

CHANGELOG.md
```diff
@@ -211,7 +211,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `garage.np.OffPolicyRLAlgorithm` ([#1552](https://github.com/rlworkgroup/garage/pull/1552))
 
 ### Fixed
-- Bug where `GarageEnv` did not pickle ([#1029](https://github.com/rlworkgroup/garage/pull/1029))
+- Bug where `GymEnv` did not pickle ([#1029](https://github.com/rlworkgroup/garage/pull/1029))
```
yeukfu, line 214: I think we doesn't need to change this name as it is in a log.

docs/user/implement_algo.md
```diff
@@ -84,13 +84,21 @@ We'll also want an environment to test our algorithm with.
 
 ```py
+<<<<<<< HEAD
```
yeukfu, line 87: should remove?

docs/user/implement_algo.md
```diff
@@ -84,13 +84,21 @@ We'll also want an environment to test our algorithm with.
...
 from garage.envs import PointEnv
+=======
+from garage.envs import PointEnv, GymEnv
+>>>>>>> a1877525... Flatten observations if necessary
 from garage.experiment import LocalRunner
 
 @wrap_experiment
 def debug_my_algorithm(ctxt):
     runner = LocalRunner(ctxt)
+<<<<<<< HEAD
```
yeukfu, line 97: should remove?

docs/user/meta_multi_task_rl_exp.md
```diff
@@ -50,7 +50,11 @@ def te_ppo_ml1_push(ctxt, seed, n_epochs, batch_size_per_task):
 
     """
     set_seed(seed)
+<<<<<<< HEAD
```
yeukfu, line 53: should remove?

docs/user/training_a_policy.md
```diff
@@ -47,7 +47,11 @@ like [this](implement_env). In this example, we choose `CartPole-V1`
 environment.
 
 ```py
+<<<<<<< HEAD
```
yeukfu, line 50: should remove?

docs/user/use_pretrained_network_to_start_new_experiment.md
```diff
@@ -75,6 +75,7 @@ import gym
 
 from garage import wrap_experiment
 from garage.envs import GymEnv
+<<<<<<< HEAD
```
yeukfu, line 78: the same as above

examples/jupyter/custom_env.ipynb
```diff
@@ -359,7 +359,11 @@
     "        return obs\n",
     "      \n",
     "env = NpWrapper(env)\n",
+<<<<<<< HEAD
```
yeukfu, line 362: the same as above

yeukfu:
@avnishn Could you please review my comments?

avnishn:
Yup sorry about that I didn't catch it in the tests thanks! Will fix rn.

ryanjulian:

PR #1820, opened 2020-07-28
Pin TF and torch to workaound #1819 & #1823

### LINKED ISSUE #1819: TF 2.3.0 incompatibility
TF 2.3.0 was released yesterday, and seems to be incompatible with TFP <0.11.0 and breaks imports (https://travis-ci.com/github/rlworkgroup/garage/jobs/365922927#L3061). We pin TFP to <=0.10.0 in the first place to resolve cloudpickle version mismatch (https://github.com/rlworkgroup/garage/issues/1758). Since TFP 0.11.0 still pins cloudpickle to 1.3.0 while gym needs 1.2.x, unpinning TFP won't work. So for now, TF needs to be pinned to <2.3.0

gitanshu:
#1819 & #1823

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,8 +28,8 @@ REQUIRED = [
     'setproctitle>=1.0',
     'tensorflow>=1.14,<2.3.0',
     'tensorflow-probability<=0.10.0',
-    'torch>=1.0.0,!=1.5.0',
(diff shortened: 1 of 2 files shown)


krzentner:
I'm going to submit this manually, since everything else in the merge queue is going to fail until it's in.

PR #1844, opened 2020-08-03
Normalize image observation of input in GaussianCNNBaseline.

### LINKED ISSUE #1634: GaussianCNNBaseline should normalize image observations
Similar to `CategoricalCNNPolicy`, `GaussianCNNBaseline` should normalize observations if observation space is `akro.Image`.

yeukfu:
This should close #1634.

### DIFF
--- a/src/garage/tf/baselines/gaussian_cnn_baseline.py
+++ b/src/garage/tf/baselines/gaussian_cnn_baseline.py
@@ -198,14 +198,11 @@ class GaussianCNNBaseline(GaussianCNNBaselineModel, Baseline):
         self._initialize()
 
     def _initialize(self):
-        if isinstance(self._env_spec.observation_space, akro.Image):
-            input_var = tf.compat.v1.placeholder(tf.uint8,
-                                                 shape=(None, ) +
-                                                 self._input_shape)
-        else:
-            input_var = tf.compat.v1.placeholder(tf.float32,
-                                                 shape=(None, ) +
-                                                 self._input_shape)
+        input_var = tf.compat.v1.placeholder(tf.float32,
+                                             shape=(None, ) +
+                                             self._input_shape)
+        if isinstance(self.env_spec.observation_space, akro.Image):
+            input_var = tf.cast(input_var, tf.float32) / 255.0
 
         ys_var = tf.compat.v1.placeholder(dtype=tf.float32,
                                           name='ys',
--- a/src/garage/tf/baselines/gaussian_cnn_baseline.py
+++ b/src/garage/tf/baselines/gaussian_cnn_baseline.py
@@ -6,7 +6,6 @@ import tensorflow as tf
 
 from garage import make_optimizer
 from garage.experiment import deterministic
-from garage.misc.tensor_utils import normalize_pixel_batch
 from garage.np.baselines.baseline import Baseline
 from garage.tf.baselines.gaussian_cnn_baseline_model import (
     GaussianCNNBaselineModel)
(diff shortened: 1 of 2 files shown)


src/garage/tf/baselines/gaussian_cnn_baseline.py
```diff
@@ -202,7 +201,9 @@ def _initialize(self):
         input_var = tf.compat.v1.placeholder(tf.float32,
                                              shape=(None, ) +
                                              self._input_shape)
-
+        if isinstance(self.env_spec.observation_space, akro.Image):
+            input_var = tf.cast(input_var, tf.float32)
```
ryanjulian, line 205: CNNs use uint8 placeholders? This will help save a lot of memory.

src/garage/tf/baselines/gaussian_cnn_baseline.py
```diff
@@ -199,9 +198,15 @@ def __init__(self,
...
 
     def _initialize(self):
-        input_var = tf.compat.v1.placeholder(tf.float32,
-                                             shape=(None, ) +
-                                             self._input_shape)
+        if isinstance(self._env_spec.observation_space, akro.Image):
+            input_var = tf.compat.v1.placeholder(tf.uint8,
+                                                 shape=(None, ) +
+                                                 self._input_shape)
+            input_var = tf.cast(input_var, tf.float32) / 255.0
```
ryanjulian, line 205: this defeats the purpose of using a uint8 placeholder -- you want only uint8 to feedforward through the network

yeukfu: Got it.

ryanjulian:
@maliesa96 can you review this PR?

src/garage/tf/baselines/gaussian_cnn_baseline.py
maliesa96: @ryanjulian we can't normalize pixel values to between 0 and 1 if we don't cast to a float though. AFAIK using uint8 as the input placeholder dtype optimizes memory usage before data is passed to the CNN, not while feeding forward?

maliesa96:
Is the intention here to normalize pixel values to 0-1 and mimic `CategoricalCNNPolicy` or to normalize using the mean/std? If you're normalizing values to be between 0 and 1 you should probably only do so if `self._normalize_inputs` is true

src/garage/tf/baselines/gaussian_cnn_baseline.py
ahtsan: Right, uint8 works in replay buffer where no normalization is required.

ahtsan: if observation_space is not Image, uint8 could still save memory so we should use uint8 by default, and only cast to float and normalize if observation_space is Image?

ryanjulian:
@maliesa96 the intent is to be like `CategoricalCNNPolicy`/`GaussianCNNPolicy`

src/garage/tf/baselines/gaussian_cnn_baseline.py
```diff
@@ -199,9 +198,14 @@ def __init__(self,
         self._initialize()
 
     def _initialize(self):
-        input_var = tf.compat.v1.placeholder(tf.float32,
-                                             shape=(None, ) +
-                                             self._input_shape)
+        if isinstance(self._env_spec.observation_space, akro.Image):
+            input_var = tf.compat.v1.placeholder(tf.uint8,
```
ryanjulian, line 202: sorry i take this back as i wasn't thinking very clearly. you should follow what `CategoriaCNNPolicy` does (i.e. use float32): https://github.com/rlworkgroup/garage/blob/master/src/garage/tf/policies/categorical_cnn_policy.py#L123

I think your original version was fine.

src/garage/tf/baselines/gaussian_cnn_baseline.py
maliesa96: The primitive then wouldn't support float data out of the box though. We could just add the dtype as a param in the constructor with uint8 as the default.

yeukfu:
@Mergifyio rebase

yeukfu:
@Mergifyio rebase

PR #1853, opened 2020-08-04
Use eval_max_episode lengths in evalutation

avnishn:
This commit fixes a bug where when obtaining
eval trajectories with sac, the max_eval_episode_length
isn't actually used by the algorithm or anywhere.

### DIFF
--- a/src/garage/torch/algos/sac.py
+++ b/src/garage/torch/algos/sac.py
@@ -45,7 +45,7 @@ class SAC(RLAlgorithm):
             environment that the agent is being trained in. Usually accessable
             by calling env.spec.
         max_episode_length (int): Max path length of the algorithm.
-        max_eval_path_length (int or None): Maximum length of paths used for
+        max_eval_episode_length (int or None): Maximum length of paths used for
             off-policy evaluation. If None, defaults to `max_episode_length`.
         gradient_steps_per_itr (int): Number of optimization steps that should
         max_episode_length(int): Max path length of the environment.
@@ -94,7 +94,7 @@ class SAC(RLAlgorithm):
         replay_buffer,
         *,  # Everything after this is numbers.
         max_episode_length,
-        max_eval_path_length=None,
+        max_eval_episode_length=None,
         gradient_steps_per_itr,
         fixed_alpha=None,
         target_entropy=None,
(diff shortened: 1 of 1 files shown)


src/garage/torch/algos/sac.py
```diff
@@ -45,7 +45,7 @@ class SAC(RLAlgorithm):
             environment that the agent is being trained in. Usually accessable
             by calling env.spec.
         max_episode_length (int): Max path length of the algorithm.
```
ryanjulian, line 47: this description doesn't make much sense -- can you make it longer?

src/garage/torch/algos/sac.py
```diff
@@ -94,7 +94,7 @@ def __init__(
         replay_buffer,
         *,  # Everything after this is numbers.
         max_episode_length,
-        max_eval_path_length=None,
+        max_eval_episode_length=None,
```
ryanjulian, line 97:
```suggestion
        max_episode_length_eval=None,
```

src/garage/torch/algos/sac.py
```diff
@@ -94,7 +94,7 @@ def __init__(
         replay_buffer,
         *,  # Everything after this is numbers.
         max_episode_length,
```
ryanjulian, line 96:
```suggestion
        max_episode_length_explore,
```

src/garage/torch/algos/sac.py
ryanjulian: For both of these, please explain what happens if the environment has a `max_episode_length` which is less than these numbers

src/garage/torch/algos/sac.py
ryanjulian: if i don't set it, does it use the `max_episode_length` from the environment?

avnishn: no, it is default 1000. I had done this when we were only doing apples to apples comparisons with the paper. A less confusing behavior would be do use `max_episode_length_explore`

ryanjulian: i agree with that as long as `max_episode_length_explore` defaults to `env.spec.max_episode_length`

In general I think people shouldn't interact with these unless they have very specific needs (e.g. your baseline comparisons), e.g. this wouldn't appear in an example.

Issue #1866, opened 2020-08-07
`from garage import wrap_experiment` throws TypeError: an integer is required (got type bytes)


JustoA:
I recently installed a brand new Anaconda 3, python 3.8 environment and then installed `gym`, `garage`, and `gym-retro` using the following command:
`pip3 install gym garage gym-retro`
Installation went off without a hitch, and I double checked with `
pip3 check`:
> No broken requirements found.

However, when I try to run the "Run Experiments" example, I'm met with the following error:

```
Traceback (most recent call last):
  File "/home/jutin/PycharmProjects/newML/test-garage.py", line 1, in <module>
    from garage import wrap_experiment
  File "/home/jutin/anaconda3/envs/newML/lib/python3.8/site-packages/garage/__init__.py", line 10, in <module>
    from garage.experiment.experiment import wrap_experiment
  File "/home/jutin/anaconda3/envs/newML/lib/python3.8/site-packages/garage/experiment/__init__.py", line 4, in <module>
    from garage.experiment.local_runner import LocalRunner
  File "/home/jutin/anaconda3/envs/newML/lib/python3.8/site-packages/garage/experiment/local_runner.py", line 11, in <module>
    from garage.experiment.snapshotter import Snapshotter
  File "/home/jutin/anaconda3/envs/newML/lib/python3.8/site-packages/garage/experiment/snapshotter.py", line 8, in <module>
    import joblib
  File "/home/jutin/anaconda3/envs/newML/lib/python3.8/site-packages/joblib/__init__.py", line 119, in <module>
    from .parallel import Parallel
  File "/home/jutin/anaconda3/envs/newML/lib/python3.8/site-packages/joblib/parallel.py", line 32, in <module>
    from .externals.cloudpickle import dumps, loads
  File "/home/jutin/anaconda3/envs/newML/lib/python3.8/site-packages/joblib/externals/cloudpickle/__init__.py", line 3, in <module>
    from .cloudpickle import *
  File "/home/jutin/anaconda3/envs/newML/lib/python3.8/site-packages/joblib/externals/cloudpickle/cloudpickle.py", line 151, in <module>
    _cell_set_template_code = _make_cell_set_template_code()
  File "/home/jutin/anaconda3/envs/newML/lib/python3.8/site-packages/joblib/externals/cloudpickle/cloudpickle.py", line 132, in _make_cell_set_template_code
    return types.CodeType(
TypeError: an integer is required (got type bytes)
 ```

[...]

PR #1871, opened 2020-08-07
Print frozen deps in the CI

ryanjulian:
This will make it easy to find a frozen set of dependencies for each release branch, e.g. in case `pip` installs are failing.


ryanjulian:
ref: #1866

ryanjulian:
@Mergifyio rebase

PR #1826, opened 2020-07-30
Request review from contributors only once

gitanshu:

PR #1872, opened 2020-08-07
Test commit to see if #1826 works

gitanshu:
TESTING MERGIFY


gitanshu:
Wasn't working immediately but seems to be working now :confused: :shrug:

PR #1875, opened 2020-08-10
Update implement env doc

AiRuiChen:
Closes [#1672 ](https://github.com/rlworkgroup/garage/issues/1672)


docs/user/implement_env.rst
```diff
@@ -1,50 +1,68 @@
 .. _implement_mdp:
 
-===========================
-Adding a New Environment
-===========================
+================================
+Working with Garage Environments
```
ryanjulian, line 4: Please keep the title 'Add a New Environment' because this is a tutorial

If you want to write the reference guide for environments, that's a different article.

docs/user/implement_env.rst
```diff
@@ -1,50 +1,68 @@
...
+* :code:`observation_space`: The observation space specification.
+* :code:`spec`: The environment specifications.
+* :code:`render_modes`: The list of supported render modes.
 
-    myEnv = GarageEnv(gym.make('CartPole-v1'))  # Wraps a CartPole-v1 environment
+You should expect to use all environments in garage with this API.
 
 +++++++++++++++++++++++++++
-Add a Environment Wrapper
+Add an Environment Wrapper
```
ryanjulian, line 30: Usually environment wrapper means something different -- it's a class you wrap around environment to change something like the observation, action, etc.

This section should be more like "Add Support for an Existing Environment Library"

docs/user/implement_env.rst
```diff
@@ -1,50 +1,68 @@
...
 
-Available environment wrappers in garage are:
+Here is an handy To-do list for adding a new environment wrapper:
 
-* :code:`garage.envs.dm_control.DmControlEnv`, which supports Deepmind's :code:`dm_control` API.
-* :code:`garage.envs.bullet.BulletEnv`, which supports the :code:`pybullet` API.
+* **Inherit** the environment wrapper from :code:`Environment`.
 
-To-do list for adding a new environment wrapper:
+* **Specify** attributes: :code:`observation_space`, :code:`action_space`, :code:`spec`, and :code:`render_modes`
```
ryanjulian, line 50: I think specify and implement here are the same thing

docs/user/implement_env.rst
```diff
@@ -1,50 +1,68 @@
 .. _implement_mdp:
```
ryanjulian, line 1: Please convert this file to Markdown

AiRuiChen:
@Mergifyio rebase

docs/user/implement_env.md
```diff
@@ -0,0 +1,308 @@
+# Adding a New Environment
+
+Garage uses the [`Environment`](https://github.com/rlworkgroup/garage/blob/master/src/garage/_environment.py)
+API for all RL environments.
+
+The public API methods of :`Environment` are:
+
+* `reset()`: Resets the environment. Return the first observation, and the episode-level information.
```
ryanjulian, line 8: Please use tables for these sections as you did in the docstring. You can refer directly to the class reference using e.g.```:meth:`~Environment.reset` ``` which should render as "reset" with a link to the API docs.

docs/user/implement_env.md
```diff
@@ -0,0 +1,308 @@
...
+* `spec`: The environment specifications.
+* `render_modes`: The list of supported render modes.
+
+You should expect to use all environments in garage with this API.
+
+## Add Support for an Existing Environment Library
+
+Luckily, Garage already came with a variety of external environment wrappers:
+
+* `garage.envs.GymEnv`, which supports the `gym` API.
```
ryanjulian, line 30: I think class names are sufficient here rather than fully-qualified module paths. You can refer to these using ``` :class:`GymEnv` ``` (for instance) to link to the API reference.

docs/user/implement_env.md
```diff
@@ -0,0 +1,308 @@
...
+    can be wrapped with `garage.envs.GymEnv` as well. In such case,
+    `garage.envs.GymEnv` will detect the bullet-based environment and returns
+    a `garage.envs.bullet.BulletEnv`.
+
+If you would like to add an existing environment to garage, you will probably
+need an environment wrapper to handle environment-specific logic.
+
+Here is an handy To-do list for adding a new environment wrapper:
+
+* **Inherit** the environment wrapper from `Environment`.
```
ryanjulian, line 45: This is never required in Python because of duck typing. They merely need to implement the interface methods.

docs/user/implement_env.md
```diff
@@ -0,0 +1,308 @@
...
+    If you would like to contribute your own environment back to garage,
+    make sure to add unit tests under the directory `tests/garage/envs`.
+    You can find examples of environment tests here as well.
+
+
+## Implement a New Environment
+
+In the rest of this section, we will walk through an example of implementing a
+point robot environment using our framework. A more complete version of this
+environment is available as `garage.envs.PointEnv`.
```
ryanjulian, line 73:
```suggestion
environment is available as `PointEnv`.
```

docs/user/implement_env.md
AiRuiChen: I don't think sphnix will create a link to the API in markdown file with :class:`GymEnv`. Please check the read the docs result.

Seems like this only works in docstrings.

docs/user/implement_env.md
AiRuiChen: :meth:`~Environment.reset` doesn't seem to work, even in code strings. Eg :method:`~LocalRunner.step_epochs()` in the `train` docstring doesn't link to the API reference in the latest read the doc [page](https://garage.readthedocs.io/en/latest/_autoapi/garage/torch/algos/index.html#id1).

And sphix doesn't come with support of markdown table (https://github.com/readthedocs/readthedocs.org/issues/4466). Though there seem to be some workarounds but the setup could be a pain. We can either:
1. stick to bullet points 2. Try set up markdown table support (https://pypi.org/project/sphinx-markdown-tables/).

What do you think?

docs/user/implement_env.md
```diff
@@ -0,0 +1,312 @@
+# Adding a New Environment
+
+Garage uses the :class:`Environment` API for all RL environments.
```
yeukfu, line 3: :class:`Environment` seems not work with markdown.

docs/user/implement_env.md
```diff
@@ -0,0 +1,312 @@
...
+
+The public API methods of :class:`Environment` are:
+
+| Functions                      | Description                          |
```
yeukfu, line 7: I think maybe rst table is needed here, as the markdown table doesn't show well in the [generated html](https://garage--1875.org.readthedocs.build/en/1875/user/implement_env.html).

docs/user/implement_env.md
```diff
@@ -0,0 +1,312 @@
...
+| ------------------------------ | ------------------------------------ |
+| :meth:`~Environment.reset`     | Resets the environment. Return the first observation, and the episode-level information. |
```
yeukfu, line 9: doesn't work

docs/user/implement_env.md
maliesa96: ``:meth: `~LocalRunner.step_epochs()` `` works for me in `train()`

docs/user/implement_env.md
```diff
@@ -0,0 +1,312 @@
...
+
+## Implement a New Environment
+
+In the rest of this section, we will walk through an example of implementing a
+point robot environment using our framework. A more complete version of this
+environment is available as :class:`PointEnv`.
+
+We will implement a simple environment with 2D observations and 2D actions. The
+goal is to control a point robot in 2D to move it to the origin. We receive
+position of a point robot in the 2D plane :math:`(x, y) \in \mathbb{R}^2`.
```
krzentner, line 81: These `:math:` don't seem to work either.

krzentner:
I'm not really sure what's needed to make all of the formatting statements work with this change. Honestly, it feels like it might just be simpler to keep the file in rst? I don't have a strong opinion either way though.

docs/user/implement_env.md
AiRuiChen: I see. Yea this seems to only work in docstrings but not in rst/md files.

docs/user/implement_env.md
AiRuiChen: fixed with `eval_rst`

AiRuiChen:
> I'm not really sure what's needed to make all of the formatting statements work with this change. Honestly, it feels like it might just be simpler to keep the file in rst? I don't have a strong opinion either way though.

I decided to keep this in .md for doc file format consistency. Solved formatting problems via `eval_rst`. However stuffs like :class: or :meth: seem to only work in docstrings. 

@krzentner @yeukfu @ryanjulian Please take another look, thanks!

ryanjulian:
:class:/:meth: directives work in rst and not markdown. Docstrings and `eval_rst` are written in rst, which is why they work there.

PR #1879, opened 2020-08-10
Unpin TensorFlow

### LINKED ISSUE #1819: TF 2.3.0 incompatibility

yeukfu:
This should close #1819.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,6 @@ REQUIRED = [
     f'gym[atari,box2d,classic_control]=={GYM_VERSION}',
     'numpy>=1.14.5',
     'psutil',
-    'pyglet>=1.4.0,<=1.5.0',
     'python-dateutil',
     'ray',
     'scikit-image',
--- a/setup.py
+++ b/setup.py
@@ -4,30 +4,27 @@ import os
 from setuptools import find_packages, setup
 
 GARAGE_GH_TOKEN = os.environ.get('GARAGE_GH_TOKEN') or 'git'
-GYM_VERSION = '0.15.4'
+GYM_VERSION = '0.17.2'
 
 # Required dependencies
 REQUIRED = [
     # Please keep alphabetized
     'akro',
     'click>=2.0',
-    'cloudpickle<1.5',
+    'cloudpickle==1.3',
     'cma==2.7.0',
     'dowel==0.0.3',
     f'gym[atari,box2d,classic_control]=={GYM_VERSION}',
     'numpy>=1.14.5',
     'psutil',
-    # Pyglet 1.4.0 introduces some api change which breaks some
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -4,30 +4,27 @@
...
     'cma==2.7.0',
     'dowel==0.0.3',
     f'gym[atari,box2d,classic_control]=={GYM_VERSION}',
     'numpy>=1.14.5',
     'psutil',
-    # Pyglet 1.4.0 introduces some api change which breaks some
-    # gym environments
-    # See: https://github.com/openai/gym/issues/1588
-    'pyglet<1.4.0,>=1.3.0',
+    'pyglet>=1.4.0,<=1.5.0',
```
gitanshu, line 20: I think we can remove this as a direct dependency if we are upgrading gym

setup.py
```diff
@@ -4,30 +4,27 @@
...
-GYM_VERSION = '0.15.4'
+GYM_VERSION = '0.17.2'
 
 # Required dependencies
 REQUIRED = [
     # Please keep alphabetized
     'akro',
     'click>=2.0',
-    'cloudpickle<1.5',
+    'cloudpickle==1.3',
```
gitanshu, line 14: Can we do `>=1.3` instead? This seems too restrictive

setup.py
yeukfu: I copy this form gym's [setup.py](https://github.com/openai/gym/blob/6df1b994bae791667a556e193d2a215b8a1e397a/setup.py#L31)

setup.py
yeukfu: It seems that tensorflow-probability 0.11 pin cloudpickle to 1.3.

gitanshu: Ahh right. It's set to >=1.3 on master, but that's unreleased yet. Can you open an issue to track this?

yeukfu: Sure. So currently should we pin cloudpickle to 1.3 or >=1.3?

gitanshu: I guesswe can leave it at ==1.3 for now

ryanjulian:
@Mergifyio rebase

yeukfu:

Issue #1556, opened 2020-06-11
torch/TD3


ryanjulian:
See https://arxiv.org/abs/1802.09477

PR #1890, opened 2020-08-11
TD3 implementation in pytorch 

irisliucy:
TD3 implementation in Pytorch addressing #1556.


src/garage/torch/algos/td3.py
```diff
@@ -0,0 +1,286 @@
...
+                 policy_noise_clip=0.5,
+                 clip_return=np.inf,
+                 policy_lr=_Default(1e-4),
+                 qf_lr=_Default(1e-3),
+                 policy_optimizer=torch.optim.Adam,
+                 qf_optimizer=torch.optim.Adam,
+                 steps_per_epoch=20,
+                 n_train_steps=50):
+
+        self.env_spec = env_spec
```
ryanjulian, line 77: can this be private?

src/garage/torch/algos/td3.py
```diff
@@ -0,0 +1,286 @@
...
+        self._n_train_steps = n_train_steps
+        self._update_actor_interval = update_actor_interval
+        self._steps_per_epoch = steps_per_epoch
+        self._episode_policy_losses = []
+        self._episode_qf_losses = []
+        self._epoch_ys = []
+        self._epoch_qs = []
+        self._eval_env = None
+
+        self.replay_buffer = replay_buffer
```
ryanjulian, line 97: please make all of these private

src/garage/torch/algos/td3.py
```diff
@@ -0,0 +1,286 @@
...
+                                              lr=qf_lr)
+        self._itr = 0
+
+    def train(self, runner):
+        """Obtain samplers and start actual training for each epoch.
+
+        Args:
+            runner (LocalRunner): LocalRunner is passed to give algorithm
+                the access to runner.step_epochs(), which provides services
+                such as snapshotting and sampler control.
```
ryanjulian, line 122:
```suggestion
            runner (LocalRunner): Experiment runner, which provides services
                such as snapshotting and sampler control.
```

src/garage/torch/algos/td3.py
```diff
@@ -0,0 +1,286 @@
...
+
+        Returns:
+            float: The average return in last epoch cycle.
+
+        """
+        if not self._eval_env:
+            self._eval_env = runner.get_env_copy()
+        last_returns = [float('nan')]
```
ryanjulian, line 130: probably just use `None`?

src/garage/torch/algos/td3.py
```diff
@@ -0,0 +1,286 @@
...
+                tabular.record('QFunction/AverageQ', np.mean(self._epoch_qs))
+                tabular.record('QFunction/MaxQ', np.max(self._epoch_qs))
+                tabular.record('QFunction/AverageAbsQ',
+                               np.mean(np.abs(self._epoch_qs)))
+                tabular.record('QFunction/AverageY', np.mean(self._epoch_ys))
+                tabular.record('QFunction/MaxY', np.max(self._epoch_ys))
+                tabular.record('QFunction/AverageAbsY',
+                               np.mean(np.abs(self._epoch_ys)))
+
+    def optimize_policy(self, samples_data):
```
ryanjulian, line 194:
```suggestion
    def _optimize_policy(self, samples_data):
```

src/garage/torch/algos/td3.py
```diff
@@ -0,0 +1,286 @@
...
+                    eval_samples = obtain_evaluation_samples(
+                        self.policy, self._eval_env)
+                    last_returns = log_performance(runner.step_itr,
+                                                   eval_samples,
+                                                   discount=self._discount)
+                runner.step_itr += 1
+
+        return np.mean(last_returns)
+
+    def train_once(self, itr, trajectories):
```
ryanjulian, line 149:
```suggestion
    def _train_once(self, itr, trajectories):
```

src/garage/torch/algos/td3.py
```diff
@@ -0,0 +1,286 @@
...
+        self._epoch_qs = []
+        self._eval_env = None
+
+        self.replay_buffer = replay_buffer
+        self.policy = policy
+        self.qf_1 = qf
+        self.qf_2 = qf2
+        self.target_policy = copy.deepcopy(self.policy)
+        self.target_qf_1 = copy.deepcopy(self.qf_1)
+        self.target_qf_2 = copy.deepcopy(self.qf_2)
```
ryanjulian, line 103:
```suggestion
        self._replay_buffer = replay_buffer
        self._policy = policy
        self._qf_1 = qf
        self._qf_2 = qf2
        self._target_policy = copy.deepcopy(self.policy)
        self._target_qf_1 = copy.deepcopy(self.qf_1)
        self._target_qf_2 = copy.deepcopy(self.qf_2)
```

src/garage/torch/algos/td3.py
ryanjulian, line 77:
```suggestion
        self._env_spec = env_spec
```

src/garage/torch/algos/td3.py
```diff
@@ -0,0 +1,286 @@
...
+            actor_loss.backward()
+            self._policy_optimizer.step()
+
+            # update target networks
+            self.update_network_parameters()
+
+        return (critic_loss.detach(), target_Q, current_Q.detach(),
+                actor_loss.detach())
+
+    def update_network_parameters(self):
```
ryanjulian, line 271:
```suggestion
    def _update_network_parameters(self):
```

src/garage/torch/algos/td3.py
```diff
@@ -0,0 +1,286 @@
...
+        steps_per_epoch (int): Number of train_once calls per epoch.
+        n_train_steps (int): Number of training steps.
+
+    """
+
+    def __init__(self,
+                 env_spec,
+                 policy,
+                 qf,
+                 qf2,
```
ryanjulian, line 58:
```suggestion
                 qf1,
                 qf2,
```

src/garage/torch/algos/td3.py
```diff
@@ -0,0 +1,286 @@
...
+from garage.torch import dict_np_to_torch, torch_to_np
+
+
+class TD3(RLAlgorithm):
+    """Implementation of TD3.
+
+    Based on https://arxiv.org/pdf/1802.09477.pdf.
+
+    Args:
+        env_spec (garage.envs.EnvSpec): Environment.
```
ryanjulian, line 20:
```suggestion
        env_spec (EnvSpec): Environment specification.
```

src/garage/torch/algos/td3.py
```diff
@@ -0,0 +1,286 @@
...
+        policy (garage.torch.policies.Policy): Policy (actor network).
+        qf (garage.torch.q_functions.QFunction): Q-function (critic network).
+        qf2 (garage.torch.q_functions.QFunction): Q function (critic network).
+        replay_buffer (garage.replay_buffer.ReplayBuffer): Replay buffer.
```
ryanjulian, line 24:
```suggestion
        replay_buffer (ReplayBuffer): Replay buffer.
```

src/garage/torch/algos/td3.py
```diff
@@ -0,0 +1,286 @@
...
+            for training policy network. This can be an optimizer type such as
+            `torch.optim.Adam` or a tuple of type and dictionary, where
+            dictionary contains arguments to initialize the optimizer
+            e.g. `(torch.optim.Adam, {'lr' : 1e-3})`.
+        qf_optimizer (Union[type, tuple[type, dict]]): Type of optimizer
+            for training Q-value network. This can be an optimizer type such
+            as `torch.optim.Adam` or a tuple of type and dictionary, where
+            dictionary contains arguments to initialize the optimizer
+            e.g. `(torch.optim.Adam, {'lr' : 1e-3})`.
+        steps_per_epoch (int): Number of train_once calls per epoch.
```
ryanjulian, line 49: `train_once` is private.

Please specify this as `n_grad_steps_per_env_step` -- i.e. the number of gradient steps taken per environment step sampled.

ryanjulian, line 49:
```suggestion
        grad_steps_per_epoch (int): Number of gradient steps per evaluation epoch.
```

src/garage/torch/algos/td3.py
```diff
@@ -0,0 +1,286 @@
...
+        n_train_steps (int): Number of training steps.
```
ryanjulian, line 50:
```suggestion
        grad_steps_per_env_step (int): Number of gradient steps taken per environment step sampled.
```

Please update the code to reflect this!

src/garage/torch/algos/td3.py
```diff
@@ -0,0 +1,286 @@
...
+                runner.step_path = runner.obtain_trajectories(runner.step_itr)
+                self.train_once(runner.step_itr, runner.step_path)
+                if (cycle == 0 and self.replay_buffer.n_transitions_stored >=
+                        self._min_buffer_size):
+                    runner.enable_logging = True
+                    eval_samples = obtain_evaluation_samples(
+                        self.policy, self._eval_env)
+                    last_returns = log_performance(runner.step_itr,
+                                                   eval_samples,
+                                                   discount=self._discount)
```
ryanjulian, line 144:
```suggestion
                    eval_eps = obtain_evaluation_samples(
                        self.policy, self._eval_env)
                    last_returns = log_performance(runner.step_itr,
                                                   eval_eps,
                                                   discount=self._discount)
```

src/garage/torch/algos/td3.py
ryanjulian, line 149:
```suggestion
    def train_once(self, itr, episodes):
```

src/garage/torch/algos/td3.py
```diff
@@ -0,0 +1,286 @@
...
+                runner.step_itr += 1
+
+        return np.mean(last_returns)
+
+    def train_once(self, itr, trajectories):
+        """Perform one iteration of training.
+
+        Args:
+            itr (int): Iteration number.
+            trajectories (TrajectoryBatch): Batch of trajectories.
```
ryanjulian, line 154:
```suggestion
            episodes (TrajectoryBatch): Batch of episodes.
```

src/garage/torch/algos/td3.py
```diff
@@ -0,0 +1,286 @@
...
+
+        """
+        self.replay_buffer.add_trajectory_batch(trajectories)
```
ryanjulian, line 157: please move this into `train`

src/garage/torch/algos/td3.py
ryanjulian, line 194: please make this operate on `TimeStepBatch` to make future refactoring easier.

ryanjulian (commented):
Looking good!

Don't forget to add:
- [x] Examples
- [x] Benchmark file in benchmarks/
- [x] Documentation page (`Algorithms and Methods / Off-Policy Continuous Action Space / TD3 (PyTorch)`)
- [x] GPU support (see SAC and PEARL for examples)
- [x] A pickle implementation (`__setstate__`/`__getstate__`)
- [x] Tests

How do the benchmark results look?

irisliucy:
Experiment result with TD3 on 'InvertedDoublePendulum-v2': https://tensorboard.dev/experiment/EfZ04qxyTLO5qjBx57rp0g/#scalars

last tested: 4:55pm August 11, 2020

.pylintrc
```diff
@@ -22,6 +22,8 @@ disable =
     invalid-name,
     line-too-long,
     wrong-import-order,
+    # Class must have 2 public methods seems arbitary
+    too-few-public-methods,
```
ryanjulian, line 26: LGTM

avnishn:
@irisliucy please convert this to a draft pr to avoid triggering the CI until it is ready to merge.

avnishn:
Can you please also look to TF agents as an example of the tests that would be important to have before merging? @irisliucy

src/garage/torch/algos/td3.py
```diff
@@ -0,0 +1,455 @@
...
+            TrajectoryBatch: Evaluation trajectories, representing the best
+                current performance of the algorithm.
+
+        """
+        paths = []
+
+        for _ in range(self._num_evaluation_trajectories):
+            observations, actions, rewards, dones, agent_infos, env_infos = \
+                [], [], [], [], [], []
+            obs, path_length, episode_reward, agenet_info = \
```
avnishn, line 366: agent

irisliucy:
Benchmarking of TD3-garage in Pytorch compared to OpenAI Spinningup (Last Updated on September 02, 10:02am):

HalfCheetah-v2
Hopper-v2
InvertedDoublePendulum-v2
Walker2d-v2

irisliucy:
Comparison of TD3 Pytorch and OpenAI Spinningup on 'Hopper-v2', 'Walker-v2', 'Swimmer-v2' for 5 seeds (updated on Sep 10):
Hopper-v2
Swimmer-v2
Walker2d-v2

src/garage/torch/algos/td3.py
```diff
@@ -0,0 +1,429 @@
...
+        ]
+
+    def to(self, device=None):
+        """Put all the networks within the model on device.
+
+        Args:
+            device (str): ID of GPU or CPU.
+
+        """
+        if torch.cuda.is_available():
```
krzentner, line 422: I don't think the other algorithms' `.to()` methods do this. Can you move this logic to create a function such as `garage.torch.prefer_gpu()`

src/garage/torch/algos/td3.py
```diff
@@ -0,0 +1,429 @@
...
+        """
+        return obtain_evaluation_episodes(
+            self.exploration_policy,
+            self._eval_env,
+            num_eps=self._num_evaluation_episodes,
+            deterministic=False)
+
+    def _update_network_parameters(self):
+        """Update parameters in actor network and critic networks."""
+        for target_param, param in zip(self._target_qf_1.parameters(),
```
krzentner, line 372: I think I've seen this same loop in multiple algorithms now. Can you pull it out into a function like `garage.torch.soft_update_model(target_model, source_model, tau)`? You don't need to refactor the other algorithms.

src/garage/torch/algos/td3.py
```diff
@@ -0,0 +1,429 @@
...
+
+    def __getstate__(self):
+        """Object.__getstate__.
+
+        Returns:
+            dict: the state to be pickled for the instance.
+
+        """
+        data = self.__dict__.copy()
+        del data['_replay_buffer']
```
krzentner, line 164: Please don't do this, since it makes the results of `resume` really difficult to interpret. Just set the `snapshot_gap` (and `snapshot_mode`) parameters to `wrap_experiment`.

src/garage/torch/algos/td3.py
```diff
@@ -0,0 +1,429 @@
...
+                                    eval_eps,
+                                    discount=self._discount,
+                                    prefix='Training')
+                    last_returns = log_performance(trainer.step_itr,
+                                                   eval_eps,
+                                                   discount=self._discount,
+                                                   prefix='Evaluation')
+                trainer.step_itr += 1
+
+        return np.mean(last_returns) if last_returns else 0
```
krzentner, line 241: I still don't think we should return anything from `.train()`, since it's not statistically meaningful, and tests using it tend to be flaky.

src/garage/torch/algos/td3.py
```diff
@@ -0,0 +1,429 @@
...
+
+    def _train_once(self, itr):
+        """Perform one iteration of training.
+
+        Args:
+            itr (int): Iteration number.
+
+        """
+        for grad_step_timer in range(self._grad_steps_per_env_step):
```
krzentner, line 250: How long does this loop take to run in total? If it takes more than about 5 seconds, can you add a `click.progressbar` to it? There's a few other places in the code base that use it already.

src/garage/torch/algos/td3.py
```diff
@@ -0,0 +1,429 @@
...
+
+    def __init__(
+        self,
+        env_spec,
+        policy,
+        qf1,
+        qf2,
+        replay_buffer,
+        grad_steps_per_env_step,
+        exploration_policy=None,
```
krzentner, line 81: Does this algorithm work at all if `exploration_policy` isn't passed? If it's actually required, don't give it a default value.

tests/garage/torch/algos/test_td3.py
```diff
@@ -0,0 +1,57 @@
...
+                  qf2=qf2,
+                  replay_buffer=replay_buffer,
+                  steps_per_epoch=20,
+                  max_episode_length=1000,
+                  grad_steps_per_env_step=1,
+                  discount=0.99)
+
+        trainer.setup(td3, env, sampler_cls=LocalSampler)
+        last_avg_ret = trainer.train(n_epochs=10, batch_size=100)
+        assert last_avg_ret > 0
```
krzentner, line 55: It would probably be better for this test to run fewer evaluation steps on the intermediate policies (i.e. pass `num_evaluation_episodes=1` to `TD3`, and then run a big evaluation on the end on `td3.exploration_policy` (or `td3.policy`, it's not completely clear to me what's going on here).

tests/garage/torch/algos/test_td3.py
```diff
@@ -0,0 +1,57 @@
...
+                                     hidden_sizes=[256, 256],
+                                     hidden_nonlinearity=F.relu)
+
+        qf2 = ContinuousMLPQFunction(env_spec=env.spec,
+                                     hidden_sizes=[256, 256],
+                                     hidden_nonlinearity=F.relu)
+
+        replay_buffer = PathBuffer(capacity_in_transitions=int(1e6))
+
+        td3 = TD3(env_spec=env.spec,
```
krzentner, line 43: How does this call work without passing `exploration_policy`? Isn't it needed to run evaluations? Does this run any evaluations?

src/garage/torch/algos/td3.py
```diff
@@ -0,0 +1,429 @@
...
+                # Get action randomly from environment within warm-up steps.
+                # Afterwards, get action from policy.
+                if self._uniform_random_policy and \
+                        trainer.step_itr < self._start_steps:
+                    trainer.step_path = trainer.obtain_episodes(
+                        trainer.step_itr,
+                        agent_update=self._uniform_random_policy)
+                else:
+                    trainer.step_path = trainer.obtain_episodes(
+                        trainer.step_itr, agent_update=self.exploration_policy)
```
krzentner, line 219: It's pretty confusing that `self.exploration_policy` can be `None`, in which case this call actually causes `self.policy` to be used. Can you make the logic here explicit so we don't accidentally break it when we refactor?

krzentner (commented):
This is looking really good. I have several smaller questions and comments, but nothing that should affect the performance of the algorithm

irisliucy:
TD3 3M steps benchmark (walker2d-v2)
Walker2d-v2

src/garage/torch/algos/td3.py
irisliucy: Each `grad_step_timer` will take around 2.5e-6 sec. An example of `grad_steps_per_env_step = 50 ` will take ~ 0.25 sec for the whole loop.

src/garage/torch/algos/td3.py
irisliucy: @krzentner Do you mean by removing the line that set state by making a copy of `self.__dict__` (because it can be slow?? so it's better to use `snapshot_gap` instead)?

krzentner: I actually don't think you need `__setstate__` and `__getstate__` at all. It's true that pickling the replay buffer is slow, but that's not a big problem if we snapshot only occasionally.

irisliucy: Got it, I agreed and it's fixed.

irisliucy:
@Mergifyio rebase

examples/torch/td3_halfcheetah.py
```diff
@@ -0,0 +1,86 @@
...
+
+    Args:
+        ctxt (garage.experiment.ExperimentContext): The experiment
+            configuration used by LocalRunner to create the snapshotter.
+        seed (int): Used to seed the random number generator to produce
+            determinism.
+
+    """
+    set_seed(seed)
+    with TFTrainer(ctxt) as trainer:
```
krzentner, line 31: You shouldn't need to use `TFTrainer` in a torch algorithm, right? Please use `Trainer` instead to avoid confusing readers.

krzentner (approved):
Looks good! I'm glad to have another good off-policy algorithm in garage.

I have one small nit with one of the examples, but feel free to submit as soon as you address that and pass pre-commit.

src/garage/torch/algos/td3.py
```diff
@@ -0,0 +1,386 @@
...
+        Returns:
+            TrajectoryBatch: Evaluation trajectories, representing the best
+                current performance of the algorithm.
+
+        """
+        return obtain_evaluation_episodes(
+            self.exploration_policy,
+            self._eval_env,
+            num_eps=self._num_evaluation_episodes,
+            deterministic=False)
```
avnishn, line 341: please make this a command line argument like in SAC

src/garage/torch/algos/td3.py
```diff
@@ -0,0 +1,386 @@
...
+            recorded.
+
+        Returns:
+            TrajectoryBatch: Evaluation trajectories, representing the best
+                current performance of the algorithm.
+
+        """
+        return obtain_evaluation_episodes(
+            self.exploration_policy,
+            self._eval_env,
```
avnishn, line 339: I'm pretty sure that `obtain_evaluation_episodes` doesn't look for max_episode_length inside your eval env. You need to instead specify the max episode length as an argument to this function, otherwise it will default to 1000. This is probably working for gym environments, but won't work for environments like metaworld environments.

avnishn (approved):
take a look at this set of TD3 tests. I would ideally want [these](https://github.com/tensorflow/agents/blob/0239c360a317bee06e2e7a04d6caffb173af0e81/tf_agents/agents/td3/td3_agent_test.py#L183) tests. These are really high quality and super protect against any regressions. What you have here are integration tests that test if all your components are working together, but they don't reveal what components have broken if they break.

What you want is unit tests, which check to see whether changes to dependencies or other changes that contributors have made break the way that your losses are computed, or how your policy inputs are preprocessed.

You can even do a golden values test, in which you compute what the output of your lossess, policy, etc, are given a certain set of inputs, and then preserve those outputs of those primitives and check to make sure that your components are outputting the same thing every time. 

You'll want to use environments like gridworld which we have in garage, etc.

I'm going to approve this anyways, but I'd really like to see this before merged if possible.

irisliucy:
@Mergifyio rebase

krzentner:
@Mergifyio rebase

PR #1893, opened 2020-08-13
Documentation on TD3

irisliucy:
This is a doc on TD3 #1676


irisliucy:
Benchmark Results (TF) and Tf logo are added to the document.

krzentner (approved):
I think this is acceptable documentation for TF TD3. Is there some way we can make the PyTorch and Tensorflow logos not clickable? Since they are clickable, I expected clicking on them to take me to the algorithm source, but instead it just shows the image.

irisliucy:
@krzentner I updated the TD3 tf and torch logo to be non-clickable. After the torch/td3 is merged, I will add in the documentation for the torch td3 to another PR.

Issue #1896, opened 2020-08-13
Nightly build is broken by Environment


ryanjulian:
https://travis-ci.com/github/rlworkgroup/garage/jobs/371516522

PR #1897, opened 2020-08-13
Fix EnvStep comparison and tests

AiRuiChen:
1. Add EnvSpec Comparison logic so it works with `roundtrip.spec == env.spec` evaluation.
2. Fix GymEnv's get time limit logic - `gym.TimeLimit` doesn't have `spec` populated, so in this case we should look at `_max_episode_steps` directly.
3. Fix some example usages.

Closes [#1896 ](https://github.com/rlworkgroup/garage/issues/1896)

### DIFF
--- a/src/garage/_environment.py
+++ b/src/garage/_environment.py
@@ -74,6 +74,19 @@ class EnvSpec(InOutSpec):
         """
         return self._max_episode_length
 
+    def __eq__(self, other):
+        """See :meth:`object.__eq__`.
+
+        Args:
+            other (EnvSpec): :class:`~EnvSpec` to compare with.
+
+        Returns:
+            bool: Whether these :class:`~EnvSpec` instances are equal.
+        """
+        return (self.observation_space == other.observation_space
+                and self.action_space == other.action_space
+                and self.max_episode_length == other.max_episode_length)
+
 
 class EnvStep(
         collections.namedtuple('EnvStep', [
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -1,6 +1,5 @@
 """helper functions for tests and benchmarks."""
-import pickle
-
+import cloudpickle
 import numpy as np
 import pytest
 
@@ -33,10 +32,10 @@ def step_env_with_gym_quirks(env,
                              n=10,
                              visualize=True,
                              serialize_env=False):
-    """Step env gym helper.
+    """Step env helper.
 
     Args:
-        env (GymEnv): Input environment.
+        env (Environment): Input environment.
         spec (EnvSpec): The environment specification.
         n (int): Steps.
         visualize (bool): Whether to visualize the environment.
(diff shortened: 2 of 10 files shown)


tests/helpers.py
```diff
@@ -45,8 +44,15 @@ def step_env_with_gym_quirks(env,
     """
     if serialize_env:
         # Roundtrip serialization
-        round_trip = pickle.loads(pickle.dumps(env))
-        assert round_trip.env.spec == env.env.spec
+        round_trip = cloudpickle.loads(cloudpickle.dumps(env))
+        # print (round_trip.spec.observation_space,
```
yeukfu, line 48: should remove?

src/garage/_environment.py
```diff
@@ -74,6 +74,20 @@ def max_episode_length(self):
...
         return self._max_episode_length
 
+    def __eq__(self, other):
+        """See object.__eq__.
+
+        Args:
+            other (EnvSpec): the EnvSpec to compare with.
+
+        Returns:
+            True: if the attributes of EnvSpec are the same.
```
ryanjulian, line 84:
```suggestion
            bool: Whether these :class:`~EnvSpec` instances are equal.
```

src/garage/_environment.py
```diff
@@ -74,6 +74,20 @@ def max_episode_length(self):
         """
         return self._max_episode_length
 
+    def __eq__(self, other):
+        """See object.__eq__.
```
ryanjulian, line 78:
```suggestion
        """See :meth:`object.__eq__`.
```

src/garage/_environment.py
```diff
@@ -74,6 +74,20 @@ def max_episode_length(self):
...
+
+        Args:
+            other (EnvSpec): the EnvSpec to compare with.
```
ryanjulian, line 81:
```suggestion
            other (EnvSpec): :class:`~EnvSpec` to compare with.
```

src/garage/_environment.py
```diff
@@ -74,6 +74,20 @@ def max_episode_length(self):
...
+
+        Returns:
+            True: if the attributes of EnvSpec are the same.
+        """
+        return ((self.observation_space, self.action_space,
+                 self.max_episode_length) == (other.observation_space,
+                                              other.action_space,
+                                              other.max_episode_length))
```
ryanjulian, line 89:
```suggestion
        return (
            self.observation_space == other.observation_space and
            self.action_space == other.action_space and
            self.max_episode_length == other.max_episode_length
        )
```

I'm surprised this isn't handled by namedtuple

tests/helpers.py
ryanjulian: 👍

AiRuiChen: good catch!

src/garage/_environment.py
AiRuiChen: i think the default `__eq__` checks if they are the exact same instance (via looking at the address?). `==` wouldn't return true for `EnvSpec` with same values without this magic function change.

AiRuiChen:
> Does this fix #1896?

It's passing the failed tests locally on my end.

Issue #1906, opened 2020-08-16
rollout doesn't stop collecting samples if environment has a timeout


avnishn:
Rollout currently stops collecting steps on terminal steps. It should also stop collecting steps on a timeout.

maliesa96:
@avnishn is this resolved?

PR #1907, opened 2020-08-16
fix rollout episode termination logic

avnishn:
addresses #1906

### DIFF
--- a/src/garage/sampler/utils.py
+++ b/src/garage/sampler/utils.py
@@ -63,7 +63,7 @@ def rollout(env,
         observations.append(last_obs)
         agent_infos.append(agent_info)
         episode_length += 1
-        if es.terminal:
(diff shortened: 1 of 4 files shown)


src/garage/sampler/utils.py
```diff
@@ -63,7 +63,7 @@ def rollout(env,
         observations.append(last_obs)
         agent_infos.append(agent_info)
         episode_length += 1
-        if es.terminal:
+        if es.terminal or es.timeout:
```
AiRuiChen, line 66: you can use `if es.last:`

AiRuiChen:
@avnishn thanks for catching this. I found similar errors regarding timeout logic, and I have pushed a commit to fix them.
Maybe we can rename this PR as something like "Fix sampler episode termination logic"?

PR #1915, opened 2020-08-17
Pin gym version in benchmarks package

gitanshu:
openai/baselines depends on gym<16.0 while it is pinned
to 0.17.2 in garage. Installing baselines over garage
leads to installation of gym 0.15.7 which isn't compatible
with cloudpickle 1.3. We do this in our Dockerfile.

This is a temporary fix until we can remove gym dependency
in the main garage package

This might break baselines but since we are not actively
using that, we can deal with that later


gitanshu:
@Mergifyio rebase

ryanjulian:
Should this be backported?

gitanshu:
> Should this be backported?

Just checked that we backported pinning gym in garage package, so yeah we should backport this

gitanshu:
@mergifyio rebase

gitanshu:
@mergifyio rebase

gitanshu:
@Mergifyio rebase

PR #1922, opened 2020-08-17
Add PCGrad in tf

yeukfu:
The first draft of PCGrad with tf. The optimizer is from the author (https://github.com/tianheyu927/PCGrad). Do the following things to use the optimizer.

 - Input a list of losses Tensor to the optimizer instead of one Tensor.  The length of this list is the length of tasks we want to train simultaneously.
 - Add a custom worker to make sure every `train_once` we have trajectories from all the tasks.

Questions and improvements.

 - The current version is not good. Because if the length of tasks is large, for example MT50 in metaworld, the list of losses will be very big and there are too many trajectories in one train step, which make training very slow.
 - A better version in my opinion is that the PCGrad optimizer can optimize a list of losses with various length, such that we don't have to sample all the tasks in one train step. 
 - I have try some like below, but it stuck on my local machine:
```
        # Compute gradient projections.
        def proj_grad(grad_task):
            # original fixed length of tasks
            # for k in range(num_tasks):
            #     inner_product = tf.reduce_sum(grad_task*grads_task[k])
            #     proj_direction = inner_product / tf.reduce_sum(grads_task[k]*grads_task[k])
            #     grad_task = grad_task - tf.minimum(proj_direction, 0.) * grads_task[k]
            # return grad_task

            # dynamic length of tasks
            k = tf.constant(0)
            while_condition = lambda k, grad_task, grads_task: tf.less(k, tf.size(loss))
            def body(k, grad_task, grads_task):
                inner_product = tf.reduce_sum(grad_task*grads_task[k])
                proj_direction = inner_product / tf.reduce_sum(grads_task[k]*grads_task[k])
                grad_task = grad_task - tf.minimum(proj_direction, 0.) * grads_task[k]
                tf.add(k, 1)
                return k, grad_task, grads_task
            k, grad_task, _ = tf.while_loop(while_condition, body, [k, grad_task, grads_task])
            return grad_task

        proj_grads_flatten = tf.map_fn(proj_grad, grads_task)
```

### DIFF
--- a/examples/tf/pcgrad_ppo_metaworld_mt10.py
+++ b/examples/tf/pcgrad_ppo_metaworld_mt10.py
@@ -1,0 +1,75 @@
+#!/usr/bin/env python3
+"""This is an example to train multiple tasks with PCGrad-PPO algorithm."""
+from metaworld.benchmarks import MT10
+import tensorflow as tf
+
+from garage import wrap_experiment
+from garage.envs import GymEnv, normalize
+from garage.envs.multi_env_wrapper import MultiEnvWrapper, round_robin_strategy
+from garage.experiment import LocalTFRunner
+from garage.experiment.deterministic import set_seed
+from garage.tf.algos import PPO
+from garage.tf.algos.pcgrad import PCGradWorker
+from garage.tf.baselines import GaussianMLPBaseline
+from garage.tf.optimizers import PCGradOptimizer
+from garage.tf.policies import GaussianMLPPolicy
+
+
+@wrap_experiment
+def pcgrad_ppo_metaworld_mt10(ctxt=None, seed=1):
+    """Train PPO on MT10 environment with PCGrad optimizer.
+
+    Args:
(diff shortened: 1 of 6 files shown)


ryanjulian:
@yeukfu can you show us some benchmark results on MT10?

yeukfu:
The following is the benchmark results on MT10.  TF PCGrad ppo launcher is [this](https://github.com/rlworkgroup/garage/blob/pcgrad_benchmark/examples/tf/pcgrad_ppo_metaworld_mt10.py), while torch mtppo launcher is [this](https://github.com/rlworkgroup/garage/blob/pcgrad_benchmark/examples/torch/mtppo_metaworld_mt10.py).

From the results we can see that the current version cannot solve MT10. 

Besides, the current sample strategy is sample 450 env steps for each task and randomly choose 150 steps for optimization. When I increased env steps of each task from 450 to 600, I got nan error and could not train.

image
image

PR #1927, opened 2020-08-18
Return unscaled action in normalize wrapper

### LINKED ISSUE #1917: On policy algos stop learning midway
Avg return either drops dramatically or the run stops completely due to NaN errors. Could affect off policy as well.

maliesa96:
closes #1917

### DIFF
--- a/src/garage/envs/normalized_env.py
+++ b/src/garage/envs/normalized_env.py
@@ -109,7 +109,7 @@ class NormalizedEnv(Wrapper):
             reward = self._apply_normalize_reward(reward)
 
         return EnvStep(env_spec=es.env_spec,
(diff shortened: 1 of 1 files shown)


maliesa96:
@avnishn its a bit lengthy to put in a commit message, but essentially by returning `es.action` rather than `action` we end up training the policy on actions it never took. The policy should never see or be aware of scaled actions, it should be trained on actions that it outputted.

avnishn:
@maliesa96 thanks for the explanation. I still maintain that you should add this to the body of the commit message that way we can quickly understand the change in the future :)

maliesa96:
@avnishn Sure, done.

PR #1945, opened 2020-08-19
Fix test dependency on pre-commit

ziyiwu9494:
Pre-commit dependency meant that tests don't run on master

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,7 +78,7 @@ jobs:
   normal_test:
     name: Normal Tests
     runs-on: ubuntu-latest
-    if: github.event_name != 'pull_request' || steps.check_pre_commit.conclusion == success
+    if: github.event_name != 'pull_request' || steps.check_pre_commit.conclusion == 'success'
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


ryanjulian:
@Mergifyio rebase

.github/workflows/ci.yml
```diff
@@ -60,6 +59,7 @@ jobs:
     - name: Tag docker image
       run: docker tag docker.pkg.github.com/${OWNER}/${DOCKER_CACHE_REPO}/${DOCKER_TAG} ${DOCKER_TAG}
     - name: Check pre-commit
+      if: github.event_name != 'pull_request'
```
gitanshu, line 62: I think we should add this to all the steps of the pre-commit job. That'll save time. Sadly github actions doesn't yet support skipping all the steps together, so it has to be in every step

ryanjulian:
@Mergifyio rebase

.github/workflows/ci.yml
ryanjulian: do you want @ziyiwu9494 to change this before I merge it?

scripts/travisci/check_precommit.sh
```diff
@@ -1,9 +1,9 @@
 #!/usr/bin/env bash
 status=0
 
-echo "Checking commit range ${TRAVIS_COMMIT_RANGE}"
-SOURCE="${TRAVIS_COMMIT_RANGE%...*}"
-ORIGIN="${TRAVIS_COMMIT_RANGE#*...}"
+echo "Checking commit range ${COMMIT_RANGE}"
+SOURCE="${COMMIT_RANGE%...*}"
```
ryanjulian, line 5: doesn't this break travis?

ziyiwu9494: oh wait, I forgot we still had it. Would it be better if we changed the environment variable in github actions for the time being?

PR #1968, opened 2020-08-25
Allow deterministic=False for off-policy

irisliucy:
This PR allows off-policies like SAC and TD3 to have a non-deterministic rollout. I updated the `obtain_evaluation_episodes` (previously known as `obtain_evaluation_samples`) which now accept also a boolean argument `deterministic` and by default it's true.

### DIFF
--- a/src/garage/np/_functions.py
+++ b/src/garage/np/_functions.py
@@ -32,7 +32,8 @@ def samples_to_tensors(paths):
 def obtain_evaluation_episodes(policy,
                                env,
                                max_episode_length=1000,
-                               num_eps=100):
+                               num_eps=100,
+                               deterministic=True):
     """Sample the policy for num_eps episodes and return average values.
 
     Args:
@@ -41,6 +42,8 @@ def obtain_evaluation_episodes(policy,
         max_episode_length (int): Maximum episode length. The episode will
             truncated when length of episode reaches max_episode_length.
         num_eps (int): Number of episodes.
+        deterministic (bool): Whether the a deterministic approach is used
+            in rollout.
 
     Returns:
         EpisodeBatch: Evaluation episodes, representing the best current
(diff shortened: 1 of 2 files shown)


src/garage/np/_functions.py
```diff
@@ -41,6 +42,8 @@ def obtain_evaluation_episodes(policy,
         max_episode_length (int): Maximum episode length. The episode will
             truncated when length of episode reaches max_episode_length.
         num_eps (int): Number of episodes.
+        deterministic (bool): Whether the a deterministic approach is used
```
avnishn, line 45: Typo, please correct before merging. 

Could be better phrased as: when using stochastic policies, whether random true if deterministic sampling (taking the mean of the distribution outputted by the stochastic policy) is done and false if random sampling is done (sampling from the outputted distribution).

src/garage/torch/algos/sac.py
```diff
@@ -462,7 +462,10 @@ def _evaluate_policy(self, epoch):
 
         """
         eval_episodes = obtain_evaluation_episodes(
-            self.policy, self._eval_env, num_eps=self._num_evaluation_episodes)
+            self.policy,
+            self._eval_env,
+            num_eps=self._num_evaluation_episodes,
```
avnishn, line 467: This will end up changing the output of the benchmark. While this is fine, I think that the algorithm implementer/benchmarker should be able to pass whether they want deterministic evaluation sampling or not.

irisliucy: I update the algorithm interface that now accepts `deterministic_eval_sampling`, a boolean argument.

ryanjulian: Hmm I don't think a parameter is the correct thing to do here. A stochastic algorithm should be evaluated with stochastic samples.

Are there prior works people need to reproduce which evaluate SAC using deterministic samples?

PR #1969, opened 2020-08-25
Support for new metaworld API

krzentner:
This change is complete in the sense that I believe all major difficulties have been addressed.

However, there are probably still several small bugs, and the expected performance of this change is unknown,


krzentner:
Here's the major things we need to check for this change. Feel free to mention if you think more are needed:
 1. Confirm that metaworld environments aren't being repeatedly constructed (and closed) in running any of the benchmarks.
 2. Confirm that the meta-RL algorithms (RL2-PPO, MAML-TRPO, PEARL) still work well on HalfCheetahDir
 3. Check that the meta-RL algorithms work on ML1-reach.
 4. Check that RL2-PPO and MAML-TRPO still work about as expected on ML10

examples/tf/rl2_ppo_metaworld_ml10.py
```diff
@@ -24,29 +25,32 @@
 @wrap_experiment
 def rl2_ppo_metaworld_ml10(ctxt, seed, max_episode_length, meta_batch_size,
                            n_epochs, episode_per_task):
-    """Train PPO with ML10 environment.
+    """Train RL2 PPO with ML10 environment.
 
     Args:
-        ctxt (ExperimentContext): The experiment configuration used by
```
ryanjulian, line 30: do you mind retaining the updated docstrings?

examples/torch/mtppo_metaworld_mt1_push.py
```diff
@@ -2,13 +2,15 @@
...
 # pylint: disable=no-value-for-parameter
 import click
-import metaworld.benchmarks as mwb
+import metaworld
 import torch
 
 from garage import wrap_experiment
-from garage.envs import GymEnv, normalize
+from garage.envs import normalize
+from garage.envs.multi_env_wrapper import MultiEnvWrapper, round_robin_strategy
```
ryanjulian, line 10: is this not available in the `garage.envs` namespace?

krzentner:
I'm currently checking that PEARL still works well with `HalfCheetahDir`.

examples/tf/te_ppo_metaworld_mt50.py
```diff
@@ -33,18 +35,25 @@ def te_ppo_mt50(ctxt, seed, n_epochs, batch_size_per_task):
...
     set_seed(seed)
-    tasks = MT50.get_train_tasks().all_task_names
-    envs = [normalize(GymEnv(MT50.from_task(task))) for task in tasks]
+    mt50 = metaworld.MT50()
+    task_sampler = MetaWorldTaskSampler(mt50,
+                                        'train',
+                                        lambda env, _: normalize(env),
+                                        add_env_onehot=False)
+    assert n_tasks % 50 == 0
+    assert n_tasks <= 2500
```
haydenshively, line 48: ?Should the docstring indicate there's a maximum value for `n_tasks`?

adibellathur:
> Here's the major things we need to check for this change. Feel free to mention if you think more are needed:
> 
> 1. Confirm that metaworld environments aren't being repeatedly constructed (and closed) in running any of the benchmarks.
> ...

Quick Update on the checks. Currently `rl2_ppo` works as expected,`te_ppo` and `mtppo` creates new environments what appears to be every time the env is sampled (it creates hundreds of envs before epoch 1), and `maml_trpo` creates new test environments every epoch but keeps the same train envs for the whole experiment. I have a list of the envs created as text files in a folder here if you would like to see the details: https://drive.google.com/drive/folders/1bl5VMK86gKlpTHTMKNE8URJ9Bu7Iae2G?usp=sharing

krzentner:
After nearly four weeks of work, I think this branch is now reasonable to merge (at least once we fix whatever's breaking the doc build).
We should still confirm that RL2 and MAML ML10 perform as well as our prior runs indicate, but I think we should merge this branch before we get those results in, since people are already using it for ICLR, and it keeps bit-rotting rapidly outside of master.

ryanjulian (approved):

PR #1971, opened 2020-08-26
Update algorithm interface to accept deterministic eval sampling

irisliucy:
This PR is related to #1968, fixed the docstring and updated the algorithm interface to accept `deterministic_eval_sampling`

### DIFF
--- a/src/garage/np/_functions.py
+++ b/src/garage/np/_functions.py
@@ -42,8 +42,8 @@ def obtain_evaluation_episodes(policy,
         max_episode_length (int): Maximum episode length. The episode will
             truncated when length of episode reaches max_episode_length.
         num_eps (int): Number of episodes.
-        deterministic (bool): Whether the a deterministic approach is used
-            in rollout.
+        deterministic (bool): Whether deterministic sampliing is used in
+            evaluation.
 
     Returns:
         EpisodeBatch: Evaluation episodes, representing the best current
--- a/src/garage/tf/algos/ddpg.py
+++ b/src/garage/tf/algos/ddpg.py
@@ -56,6 +56,11 @@ class DDPG(RLAlgorithm):
         max_action (float): Maximum action magnitude.
         reward_scale (float): Reward scale.
         name (str): Name of the algorithm shown in computation graph.
+        deterministic_eval_sampling (bool) : When using stochastic policies,
+            whether true if deterministic sampling (taking the mean of the
+            distribution output by the stochastic policy) is done and
+            false if random sampling is done (sampling from the output
+            distribution).
 
     """
 
@@ -83,6 +88,7 @@ class DDPG(RLAlgorithm):
             qf_lr=_Default(1e-3),
             clip_pos_returns=False,
             clip_return=np.inf,
+            deterministic_eval_sampling=True,
             max_action=None,
             reward_scale=1.,
             name='DDPG'):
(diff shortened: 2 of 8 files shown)


src/garage/tf/algos/ddpg.py
```diff
@@ -56,6 +56,11 @@ class DDPG(RLAlgorithm):
         max_action (float): Maximum action magnitude.
         reward_scale (float): Reward scale.
         name (str): Name of the algorithm shown in computation graph.
+        deterministic_eval_sampling (bool) : When using stochastic policies,
```
ryanjulian, line 59: can you change this to `eval_deterministic`?

ryanjulian:
I don't think this is necessary for TD3 or DDPG, which both only train deterministic policies. It only comes up for SAC.

Does the SAC paper use deterministic evaluation sampling?

irisliucy:
The original SAC paper suggests using deterministic sampling for policy evaluation. So I think in SAC, `deterministic = True` should be passed to `obtain_sample_episodes` in `_evaluate_policy`. Screenshot 2020-08-27 at 3 09 35 PM
@krzentner @avnishn  @ryanjulian

src/garage/np/_functions.py
```diff
@@ -42,8 +42,8 @@ def obtain_evaluation_episodes(policy,
         max_episode_length (int): Maximum episode length. The episode will
             truncated when length of episode reaches max_episode_length.
         num_eps (int): Number of episodes.
-        deterministic (bool): Whether the a deterministic approach is used
-            in rollout.
+        deterministic (bool): Whether deterministic sampliing is used in
```
maliesa96, line 45:
```suggestion
        deterministic (bool): Whether deterministic sampling is used in
```

src/garage/tf/algos/dqn.py
```diff
@@ -47,6 +47,11 @@ class DQN(RLAlgorithm):
         double_q (bool): Bool for using double q-network.
         reward_scale (float): Reward scale.
         name (str): Name of the algorithm.
+        deterministic_eval_sampling (bool) : When using stochastic policies,
+            whether true if deterministic sampling (taking the mean of the
+            distribution output by the stochastic policy) is done and
+            false if random sampling is done (sampling from the output
+            distribution).
```
maliesa96, line 54: DQN can't train a stochastic policy either, so no need for this here.

krzentner:
> I don't think this is necessary for TD3 or DDPG, which both only train deterministic policies. It only comes up for SAC.

Wait, the TD3 paper explicitly says to use noise when evaluating, so we should definitely have this option there, right? May as well also have it for DDPG, for consistency.

ryanjulian:
Okay -- I think I misunderstood the purpose. Couldn't this all be simpler if we allowed the user to specify a different `eval_policy`?

krzentner:
Yeah, I also think this change would probably overall be simpler / more clear / more general if we had a concept of `eval_policy`.

One concern I have with it as is that if I pass an `exploration_policy` and set `eval_deterministic` to `False`, I would expect the `exploration_policy` to be used in evaluation, but it won't be.

avnishn:
@irisliucy what is the state of this pr? I just merged #2003 which does this, but only for SAC and MTSAC.

krzentner (commented):
So I'm still not super clear on *exactly* what this parameter should do. For stochastic policies the meaning is fairly clear, but for algorithms with exploration policies, does setting this parameter to `False` cause the exploration policy to be used, or does the parameter do nothing for those algorithms?

If the parameter does nothing for an algorithm, we should definitely remove it.

avnishn:
I'm going to close this since I believe that it was solved with #2003

PR #1975, opened 2020-08-27
Various MTSAC bug fixes

avnishn:
fixes to Examples to use the correct num_tasks

fixes to max_episode_length_eval being used by the algorithm

Co-authored-by: Tianhong Dai <tianhongdai914@gmail.com>

### DIFF
--- a/examples/torch/mtsac_metaworld_ml1_pick_place.py
+++ b/examples/torch/mtsac_metaworld_ml1_pick_place.py
@@ -84,24 +84,30 @@ def mtsac_metaworld_ml1_pick_place(ctxt=None, seed=1, _gpu=None):
     epochs = timesteps // batch_size
     epoch_cycles = epochs // num_evaluation_points
     epochs = epochs // epoch_cycles
-    mtsac = MTSAC(policy=policy,
-                  qf1=qf1,
-                  qf2=qf2,
-                  gradient_steps_per_itr=150,
-                  max_episode_length=150,
-                  eval_env=ml1_test_envs,
-                  env_spec=ml1_train_envs.spec,
-                  num_tasks=50,
-                  steps_per_epoch=epoch_cycles,
-                  replay_buffer=replay_buffer,
-                  min_buffer_size=1500,
-                  target_update_tau=5e-3,
-                  discount=0.99,
-                  buffer_batch_size=1280)
+    mtsac = MTSAC(
+        policy=policy,
+        qf1=qf1,
+        qf2=qf2,
+        gradient_steps_per_itr=150,
+        max_episode_length=150,
+        max_episode_length_eval=150,
+        eval_env=ml1_test_envs,
+        env_spec=ml1_train_envs.spec,
+        num_tasks=50,
+        steps_per_epoch=epoch_cycles,
+        replay_buffer=replay_buffer,
+        min_buffer_size=1500,
+        target_update_tau=5e-3,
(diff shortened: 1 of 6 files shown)


avnishn:
@tianhongdai

TianhongDai:
> @TianhongDai

@avnishn Thanks Avnishn! I think I forget to read contribution guidelines when submit the PR, sorry for making so much troubles for you.

examples/torch/mtsac_metaworld_ml1_pick_place.py
```diff
@@ -84,24 +84,30 @@ def mtsac_metaworld_ml1_pick_place(ctxt=None, seed=1, _gpu=None):
...
+        qf1=qf1,
+        qf2=qf2,
+        gradient_steps_per_itr=150,
+        max_episode_length=150,
+        max_episode_length_eval=150,
+        eval_env=ml1_test_envs,
+        env_spec=ml1_train_envs.spec,
+        num_tasks=50,
+        steps_per_epoch=epoch_cycles,
+        replay_buffer=replay_buffer,
```
ryanjulian, line 98: please only call args as args and kwargs as kwargs.

tests/garage/torch/algos/test_mtsac.py
```diff
@@ -70,6 +70,61 @@ def test_mtsac_get_log_alpha(monkeypatch):
...
+                  target_update_tau=5e-3,
+                  discount=0.99,
+                  buffer_batch_size=buffer_batch_size)
+    monkeypatch.setattr(mtsac, '_log_alpha', torch.Tensor([1., 2.]))
+    error_string = ('The number of tasks in the environment does '
+                    'not match self._num_tasks. Are you sure that you passed '
+                    'The correct number of tasks?')
+    obs = torch.Tensor([env.reset()[0]] * buffer_batch_size)
+    with pytest.raises(ValueError, match=error_string):
+        mtsac._get_log_alpha(dict(observation=obs))
```
ryanjulian, line 125: no need to test a private method?

ryanjulian:
Please reference the issues you're fixing

tests/garage/torch/algos/test_mtsac.py
avnishn: This is true, however, the tests are in place to verify the correctness of this implementation. I feel more comfortable having the tests than not. At the same time, it makes no sense to have this as a publicly exposed field because it has no use outside of the algorithm.

examples/torch/mtsac_metaworld_mt10.py
```diff
@@ -95,7 +96,10 @@ def mtsac_metaworld_mt10(ctxt=None, seed=1, _gpu=None):
     if _gpu is not None:
         set_gpu_mode(True, _gpu)
     mtsac.to()
-    runner.setup(algo=mtsac, env=mt10_train_envs, sampler_cls=LocalSampler)
+    runner.setup(algo=mtsac,
+                 env=mt10_train_envs,
+                 sampler_cls=LocalSampler,
+                 n_workers=1)
```
maliesa96, line 102: Why limit the number of workers here, and can't we use ray?

avnishn:
@maliesa96 we can't. Tldr; using the ray sampler with the old metaworld envs uses more memory than we have on lab machines.

maliesa96:
Damn alright, LGTM then.

Issue #1868, opened 2020-08-07
Put active_task_one_hot in GridWorldEnv episode info


AiRuiChen:
With [#1845 ](https://github.com/rlworkgroup/garage/pull/1845), we support returning episode_info from an environment at `reset()`. We should put `active_task_one_hot` in `GridWorldEnv`'s episode_info.

PR #1978, opened 2020-08-28
Include one_hot_active_task in MultiEnvWrapper

nicolengsy:
To resolve issue #1868

### DIFF
--- a/src/garage/envs/multi_env_wrapper.py
+++ b/src/garage/envs/multi_env_wrapper.py
@@ -187,7 +187,7 @@ class MultiEnvWrapper(Wrapper):
         if self._mode == 'vanilla':
             pass
         elif self._mode == 'add-onehot':
-            obs = np.concatenate([obs, self._active_task_one_hot()])
+            episode_info["active_task_one_hot"] = self._active_task_one_hot()
         else:  # self._mode == 'del-onehot'
             obs = obs[:-self._num_tasks]
 
--- a/tests/garage/envs/test_multi_env_wrapper.py
+++ b/tests/garage/envs/test_multi_env_wrapper.py
@@ -136,15 +136,26 @@ class TestMultiEnvWrapper:
         mt_env = self._init_multi_env_wrapper(
             envs, sample_strategy=round_robin_strategy)
 
-        obs, _ = mt_env.reset()
-        assert (obs[-2:] == np.array([1., 0.])).all()
+        obs, episode_info = mt_env.reset()
+        assert (episode_info['active_task_one_hot'] == np.array([1., 0.])).all()
         obs = mt_env.step(1).observation
         assert (obs[-2:] == np.array([1., 0.])).all()
 
-        obs, _ = mt_env.reset()
-        assert (obs[-2:] == np.array([0., 1.])).all()
+        obs, episode_info = mt_env.reset()
+        assert (episode_info['active_task_one_hot'] == np.array([0., 1.])).all()
         obs = mt_env.step(1).observation
         assert (obs[-2:] == np.array([0., 1.])).all()
+
+
+    def test_one_hot_episode_info(self):
+        """test if episode info is correct"""
+        envs = ['CartPole-v0', 'CartPole-v0']
+        mt_env = self._init_multi_env_wrapper(
+            envs, sample_strategy=round_robin_strategy)
+        
+        _, episode_info = mt_env.reset()
+        assert (episode_info['active_task_one_hot'] == np.array([1., 0.])).all()
+
 
     def test_visualization(self):
         envs = ['CartPole-v0', 'CartPole-v1']


src/garage/envs/multi_env_wrapper.py
```diff
@@ -187,7 +187,7 @@ def reset(self):
         if self._mode == 'vanilla':
             pass
         elif self._mode == 'add-onehot':
-            obs = np.concatenate([obs, self._active_task_one_hot()])
+            episode_info["active_task_one_hot"] = self._active_task_one_hot()
```
ryanjulian, line 190: did you install pre-commit? we only use single quotes

ryanjulian, line 190:
```suggestion
            episode_info['task_one_hot'] = self._active_task_one_hot()
```

irisliucy:
@nicolengsy Good job! Look like @ryanjulian was right, some tests in `test_mtsac.py` are broken. In particular, I think you need to change `torch.Tensor([env.reset()[0]]* buffer_batch_size)` to `torch.Tensor([np.concatenate([env.reset()[0], env.reset()[1]['task_one_hot']])] * buffer_batch_size)` under `test_mtsac_get_log_alpha()` and maybe some other things too.

nicolengsy:
Thanks, made the requested changes! Didn't include it in this PR, but moving forward wondering if we'd want to get rid of the `mode` argument when initializing MultiEnvWrapper, since it was previously used to control whether one_hot info was included in observations, which won't happen anymore.

src/garage/envs/multi_env_wrapper.py
```diff
@@ -106,17 +106,7 @@ def observation_space(self):
...
 
         """
-        if self._mode == 'vanilla':
-            return self._env.observation_space
-        elif self._mode == 'add-onehot':
-            task_lb, task_ub = self.task_space.bounds
-            env_lb, env_ub = self._env.observation_space.bounds
-            return akro.Box(np.concatenate([env_lb, task_lb]),
-                            np.concatenate([env_ub, task_ub]))
-        else:  # self._mode == 'del-onehot'
```
yeukfu, line 116: If 'del-onehot' mode is kept. I think we should still process it.

src/garage/envs/multi_env_wrapper.py
```diff
@@ -204,13 +194,7 @@ def step(self, action):
 
         """
         es = self._env.step(action)
-
-        if self._mode == 'add-onehot':
-            obs = np.concatenate([es.observation, self._active_task_one_hot()])
-        elif self._mode == 'del-onehot':
```
yeukfu, line 210: If 'del-onehot' mode is kept. We still need to remove these from obs?

yeukfu (approved):
Looks good to me. I think you should also update the 'mode' docstring.

krzentner:
Wait, I don't understand this change at all. Why are we just removing onehot support from `MultiEnvWrapper`? That doesn't look like what the original bug report was describing.

nicolengsy:
@krzentner I've removed one-hot from the `observations`, and moved it into `episode_info` instead (see line 180 in `multi_env_wrapper.py`). I deleted checks for the mode (i.e. `add-onehot`), since one-hot is always included in `episode_info` now. I should refactor the mode out of all the files to be consistent, which I can do within this PR as well.

Was this not what Eric asked for in #1868 ?

src/garage/envs/multi_env_wrapper.py
AiRuiChen: Is mode `del-onehot` ever used? IIUC only MultiEnvWrapper will populate the active task index in observation. After this PR, the index goes into `episode_info`. So I believe there is no need to remove task index from `obs` (aka mode `del-onehot`), and we can simply do `obs = es.observation` here

src/garage/envs/multi_env_wrapper.py
```diff
@@ -187,7 +177,7 @@ def reset(self):
         if self._mode == 'vanilla':
             pass
         elif self._mode == 'add-onehot':
-            obs = np.concatenate([obs, self._active_task_one_hot()])
+            episode_info['task_one_hot'] = self._active_task_one_hot()
         else:  # self._mode == 'del-onehot'
             obs = obs[:-self._num_tasks]
```
AiRuiChen, line 182: I'd suggest to remove `mode` from MultiEnvWrapper in this PR. The active task index should *always* be in `episode_info`, we don't need a mode to determine that. 
Once you done that you can remove the if conditions here and simply do `episode_info['task_one_hot'] = self._active_task_one_hot()`.

AiRuiChen:
> Wait, I don't understand this change at all. Why are we just removing onehot support from `MultiEnvWrapper`? That doesn't look like what the original bug report was describing.

@krzentner The issue description was wrong and I just updated it. I meant to put `MultiEnvWrapper`'s active task index in `episode_info`, instead of concatenating it to `obs`. 

This PR does the described functionality. But I think what's implied in that issue is: we should also make all downstream algos to consume active task index from `episode_info`, not from`obs`. @nicolengsy This future change can go into another PR. It's fine now to manually plug active task index back to `obs` before feeding into algos (like what you did in test_mtsac.py).

krzentner:
This sounds fairly complex to achieve, since it will require all of those multi-task algorithms to somehow change the shape of their policies, even though the environments environment spec doesn't match. Currently only PEARL does that, and it's one of the things that make it more difficult to use.

I think we should drop this issue from the release, since it seems unlikely we can achieve this in the time remaining.
@ryanjulian What do you think?

PR #1979, opened 2020-08-28
Deflake test_multiprocessing_sampler

### LINKED ISSUE #1824: test_pickle in test_multiprocessing_sampler.py is flaky
I haven't been able to repro this on local but it's very flaky on the CI (https://travis-ci.com/github/rlworkgroup/garage/jobs/366008841#L2240)

yeukfu:
This should close #1824. 

`test_update_envs_env_update` in `test_multiprocessing_sampler`  is flaky because the number of samples is small so that each worker only rollouts by once, which makes the variance of collected rewards could be very low.

### DIFF
--- a/tests/garage/sampler/test_multiprocessing_sampler.py
+++ b/tests/garage/sampler/test_multiprocessing_sampler.py
@@ -50,7 +50,6 @@ def test_obtain_samples():
     env.close()
 
 
-@pytest.mark.flaky
 @pytest.mark.timeout(10)
 def test_update_envs_env_update():
     max_episode_length = 16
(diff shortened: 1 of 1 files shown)


tests/garage/sampler/test_multiprocessing_sampler.py
```diff
@@ -67,16 +66,16 @@ def test_update_envs_env_update():
...
                                       np.asarray(policy.get_param_values()),
                                       env_update=tasks.sample(n_workers))
     mean_rewards = []
     goals = []
     for eps in episodes.split():
         mean_rewards.append(eps.rewards.mean())
         goals.append(eps.env_infos['task'][0]['goal'])
-    assert np.var(mean_rewards) > 0
-    assert np.var(goals) > 0
+    assert np.var(mean_rewards) > 1e-3
```
krzentner, line 77: Doesn't this make the test more likely to fail? It should still be fine, I guess.

yeukfu: I ran it for 100 times and the lowest variance is greater than 1e-2, so I think this should be fine.

yeukfu:
@Mergifyio rebase

PR #1985, opened 2020-08-28
Make gym dependency optional

### LINKED ISSUE #1276: Remove gym dependency

haydenshively:
Should close #1276. I'll put up a PR that removes gym from akro soon

Note that gym is still present in the `garage.envs` folder and many tests

### DIFF
--- a/src/garage/tf/plotter/plotter.py
+++ b/src/garage/tf/plotter/plotter.py
@@ -30,11 +30,11 @@ class Plotter:
     Usually, this class is used by sending plot=True to LocalRunner.train().
 
     Args:
-        env (gym.Env): Environment from which to visualize episodes. This will
-        be used without copying in the current process but in a separate
-        thread, so it should be given a unique copy (in particular, do not pass
-        the training environment here, then try to pickle it, or you will
-        occasionally get crashes).
+        env (garage.Environment): Environment from which to visualize episodes.
+            This will be used without copying in the current process but in a
+            separate thread, so it should be given a unique copy (in
+            particular, do not pass the training environment here, then try to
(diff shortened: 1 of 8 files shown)


zequnyu:
Will this unpin gym for [benchmarking](https://github.com/rlworkgroup/garage/pull/1915)?

haydenshively:
> Will this unpin gym for [benchmarking](https://github.com/rlworkgroup/garage/pull/1915)?

Unfortunately no, I don't think it will. The benchmarking suite looks to be heavily reliant on `GymEnv`, which in turn requires gym (as a now-optional garage dependency). Assuming we/users want that optional dependency, the version requirements haven't changed.

That said, I think it would be fairly straightforward to extricate direct usages of `gym` from `GymEnv` by assuming that the `env` passed to init is a fully-formed gym env rather than a string. If we do that then our ties to gym might loosen enough to unpin it.

src/garage/tf/plotter/plotter.py
```diff
@@ -30,11 +30,11 @@ class Plotter:
     Usually, this class is used by sending plot=True to LocalRunner.train().
 
     Args:
-        env (gym.Env): Environment from which to visualize episodes. This will
-        be used without copying in the current process but in a separate
-        thread, so it should be given a unique copy (in particular, do not pass
-        the training environment here, then try to pickle it, or you will
-        occasionally get crashes).
+        env (garage.Environment): Environment from which to visualize episodes.
```
ryanjulian, line 33:
```suggestion
        env (Environment): Environment from which to visualize episodes.
```

ryanjulian:
@Mergifyio rebase

PR #1986, opened 2020-08-28
Fix testing env pickleable

yeukfu:
Fix comparing env spec by only comparing the name of the env. This is a temporary solution. In ideal solution, we should also compare the action_space, observation_space, max_episode_steps. However, one environment, `KellyCoinflipGeneralized-v0` will generate a different actions space each time constructed by `gym.make()`.

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -43,7 +43,7 @@ def step_env_with_gym_quirks(env, spec, n=10, render=True,
     if serialize_env:
         # Roundtrip serialization
         round_trip = pickle.loads(pickle.dumps(env))
-        assert round_trip.env.spec == env.env.spec
+        assert round_trip.env.spec.id == env.env.spec.id
         env = round_trip
 
     env.reset()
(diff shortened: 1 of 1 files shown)


AiRuiChen:
This looks good. 

See https://github.com/rlworkgroup/garage/blob/master/src/garage/_environment.py#L77 for an example of the solution you mentioned. Maybe we should just do that in this PR?

yeukfu:
> This looks good.
> 
> See https://github.com/rlworkgroup/garage/blob/master/src/garage/_environment.py#L77 for an example of the solution you mentioned. Maybe we should just do that in this PR?

This is very helpful for comparing two `EnvSpec`s. But I think the `KellyCoinflipGeneralized-v0`'s problem will still exist if we unpickle the env by env_name.

AiRuiChen:
> > This looks good.
> > See https://github.com/rlworkgroup/garage/blob/master/src/garage/_environment.py#L77 for an example of the solution you mentioned. Maybe we should just do that in this PR?
> 
> This is very helpful for comparing two `EnvSpec`s. But I think the `KellyCoinflipGeneralized-v0`'s problem will still exist if we unpickle the env by env_name.

We can also just skip that environment in the test. I don't have a preference though. If we are going for a long-term solution, then maybe implement `__eq__` sounds better. 

Btw, I didn't encounter a problem with this environment when I modified `test_gym_base.py`. I think this error will go away with the new version of `EnvSpec` and environment.

PR #1993, opened 2020-09-01
Add torch DiscreteQFDerivedPolicy and DiscreteMLPQFunction

maliesa96:


src/garage/torch/policies/discrete_qf_derived_policy.py
```diff
@@ -0,0 +1,72 @@
...
+    The action returned is the one that yields the highest Q value for
+    a given state, as determined by the supplied Q function.
+
+    Args:
+        qf (object): Q network.
+        env_spec (EnvSpec): Environment specification.
+        name (str): Name of this policy.
+    """
+
+    def __init__(self, qf, env_spec, name='DiscreteQFDerivedPolicy'):
```
ryanjulian, line 22: i don't think name is needed for torch?

src/garage/torch/policies/discrete_qf_derived_policy.py
```diff
@@ -0,0 +1,72 @@
...
+            observations (np.ndarray): Batch of observations, should
+                have shape :math:`(N, O)`.
+
+        Returns:
+            tuple:
+                * torch.Tensor: Predicted actions. Tensor has shape
+                    :math:`(N, A)`.
+                * dict: Empty since this policy does not produce a
+                    distribution.
+        """
```
ryanjulian, line 70:
```suggestion
        Returns:
            torch.Tensor: Predicted actions. Tensor has shape
                :math:`(N, A)`.
            dict: Empty since this policy does not produce a
                distribution.
        """
```

src/garage/torch/policies/discrete_qf_derived_policy.py
```diff
@@ -0,0 +1,72 @@
...
+
+        Args:
+            observation (np.ndarray): Observation with shape :math:`(O, )`.
+
+        Returns:
+            tuple:
+                * torch.Tensor: Predicted action with shape :math:`(A, )`.
+                * dict: Empty since this policy does not produce a
+                    distribution.
+        """
```
ryanjulian, line 53:
```suggestion
        Returns:
            torch.Tensor: Predicted action with shape :math:`(A, )`.
            dict: Empty since this policy does not produce a distribution.
        """
```

src/garage/torch/policies/discrete_qf_derived_policy.py
maliesa96: `garage.torch.policy` requires one, though it doesn't look like it's being used anywhere so not sure why.

src/garage/torch/policies/discrete_qf_derived_policy.py
```diff
@@ -0,0 +1,68 @@
...
+            observations(torch.Tensor): Batch of observations of shape
+                :math:`(N, O)`. Observations should be flattened even
+                if they are images as the underlying Q network handles
+                unflattening.
+
+        Returns:
+            torch.Tensor: Batch of actions of shape :math:`(N, A)`
+        """
+        qs = self._qf(observations)
+        return torch.argmax(qs, dim=1)
```
ryanjulian, line 41: do you need to differentiate through this? if not, it could just be in garage.np.policies

src/garage/torch/policies/discrete_qf_derived_policy.py
```diff
@@ -0,0 +1,68 @@
...
+
+This policy chooses the action that yields to the largest Q-value.
+"""
+import numpy as np
+import torch
+
+from garage.torch.policies.policy import Policy
+
+
+class DiscreteQFDerivedPolicy(Policy):
```
ryanjulian, line 11: DiscreteQFArgmaxPolicy?

src/garage/torch/policies/discrete_qf_derived_policy.py
maliesa96: No, but I kept it under `torch.policies` because it subclasses `garage.torch.policy` to access the default methods defined there.

src/garage/torch/policies/discrete_qf_derived_policy.py
maliesa96: Hmm sure, though I named it after the tf implementation (also `DiscreteQFDerivedPolicy`). Should both be renamed?

ryanjulian: Yeah

src/garage/torch/policies/discrete_qf_derived_policy.py
ryanjulian: hmm examples?

ryanjulian: in any case, probably best addressed when we have a proper PolicyWrapper class

tests/garage/torch/policies/test_discrete_qf_derived_policy.py
```diff
@@ -0,0 +1,73 @@
...
+import torch
+
+from garage.envs import GymEnv
+from garage.torch.policies import DiscreteQFDerivedPolicy
+from garage.torch.q_functions import DiscreteMLPQFunction
+
+from tests.fixtures.envs.dummy import DummyBoxEnv
+
+
+class TestDiscreteQFDerivedPolicy:
```
ryanjulian, line 14: do these need to use xunit style? we've been trying to use pytest style lately instead

tests/garage/torch/q_functions/test_discrete_mlp_q_function.py
```diff
@@ -0,0 +1,82 @@
...
+import torch
+from torch import nn
+
+from garage.envs import GymEnv
+from garage.torch.q_functions import DiscreteMLPQFunction
+
+from tests.fixtures.envs.dummy import DummyBoxEnv
+
+
+class TestDiscreteMLPQFunction:
```
ryanjulian, line 14: do these tests need xunit style?

src/garage/torch/policies/discrete_qf_derived_policy.py
maliesa96: `policy.get_param_values()` and `policy.set_param_values()`, both of which are used by the runner/sampler.

ryanjulian: good to know -- thanks!

irisliucy (approved):

PR #1995, opened 2020-09-02
Backport #1812 to release 2020.06

avnishn:
* Flatten observations if necessary

This commit takes care to convert observations that belong to
spaces that aren't box spaces into observations that would
come from box spaces via akro observation space flattening.

It is really a temporary change. The better solution would be
to modify our GymEnv/GarageEnv and BulletEnv Wrappers to flatten
observations spaces that aren't box spaces.

This is a smaller change, however, the HER replay buffer would
need to be rewritten, which is a much larger change/refactor.

I have refrained from modifying the primitives that belong to
multi task and metalearning algorithms specifically.

Additionally, this change updates usages of ray to reflect
the update in parameter names from ray 0.8.6->0.8.7

### DIFF
--- a/docs/user/meta_multi_task_rl_exp.md
+++ b/docs/user/meta_multi_task_rl_exp.md
@@ -1,0 +1,321 @@
+# Run Meta-/Multi-Task RL Experiments
+
+This guide will walk you through how to run meta-/ multi-task RL experiments in garage.
+
+Similar to running all other experiments in garage, running meta-/ multi-task RL experiments generally involves steps such as:
+
+- Defining the experiement with the `wrap_experiment` decorator
+- Constructing a `LocalRunner`
+- Constructing an environment
+- Constructing policy/ algorithm object(s)
+
+In meta-/multi-task RL experiment, it revolves around solving multiple tasks and hence the construction of multiple/specific environment(s). Belows are a few environment wrappers commonly used in garage for meta-/multi-task RL learning:
+
+- `MultiEnvWrapper`: a wrapper for mulitiple environments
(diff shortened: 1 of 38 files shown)


avnishn:
I think I may have accidentally added some docs from after the release, but I'm not sure if this matters.

avnishn:
@gitanshu @ryanjulian, the CI fails because its using a docker container that has been built with an older version of ray. The newest version of ray changes the parameter name `use_webui` to `use_dashboard` for the function `ray.init`. We actually never pinned ray in 2020.06, so I figured that I should also update this parameter name in my backport, however the CI seems to be using this old ray. Can you please advise on next steps.

ryanjulian:
The CI builds the docker container from scratch: https://travis-ci.com/github/rlworkgroup/garage/jobs/380405992

Could it be that the new version of ray does not support Python 3.5, or that some other dependency is holding ray back?

ryanjulian:
For a release branch, it's probably easier just to pin ray down to your preferred version, i.e.e `ray<=0.8.6`

docs/user/meta_multi_task_rl_exp.md
```diff
@@ -0,0 +1,321 @@
+# Run Meta-/Multi-Task RL Experiments
```
ryanjulian, line 1: please do not update the documentation

ryanjulian (commented):
this has tons of merge conflicts and changes to unrelated files. please let me know when it's ready for recview.

avnishn:
Please see #1997. I had issues with this branch because branches that start with release- are treated as protected branches.

PR #1997, opened 2020-09-02
Backport #1812 to 2020.06

avnishn:
Backport PR #1812 to the 2020.06 release.
    
By flattening observations inside of policies
dict observations that need their structure to be
retained, in cases such as with the gym fetch reach-env
can be used with garage algos.


ryanjulian (approved):

src/garage/tf/policies/gaussian_mlp_task_embedding_policy.py
```diff
@@ -237,8 +238,8 @@ def get_action(self, observation):
                     A is the dimension of action.
 
         """
-        obs, task = self.split_augmented_observation(observation)
-        return self.get_action_given_task(obs, task)
+        actions, agent_infos = self.get_actions([observation])
```
ryanjulian, line 241: wait, is this right?

avnishn: is there something about it that would be wrong for task embedding policies? It passes all of the current tests. @naeioi could you chime in here?

ryanjulian: `self.get_action_given_task(obs, task)` is a pretty essential part of task embeddings...

src/garage/torch/policies/stochastic_policy.py
```diff
@@ -55,6 +57,24 @@ def get_actions(self, observations):
                         values of the distribution.
 
         """
+        if not isinstance(observations[0], np.ndarray) and not isinstance(
+                observations[0], torch.Tensor):
+            observations = self._env_spec.observation_space.flatten_n(
+                observations)
+
+        # frequently users like to pass lists of torch tensors or lists of
```
ryanjulian, line 65: is there a reason you added this feature? i thought this was a backport

src/garage/tf/policies/gaussian_mlp_task_embedding_policy.py
avnishn: I see

src/garage/torch/policies/stochastic_policy.py
ryanjulian: but this wasn't necessary to pass the torch tests? if you are going to add the feature to torch, please also add the tests to torch.

src/garage/tf/policies/gaussian_mlp_task_embedding_policy.py
avnishn: I reverted this change and things are back to their original behavior.

ryanjulian:
Fixes #1913

krzentner (approved):
This looks good. The few uncovered lines are covered by tests on master, so I think we can merge them.

Issue #1991, opened 2020-09-01
Codecov reports not working from GH Actions


ryanjulian:
See https://github.com/rlworkgroup/garage/pull/1811/checks?check_run_id=1058482191#step:6:168

PR #1999, opened 2020-09-03
Fix Codecov Bug On Master

ziyiwu9494:
test fix for #1991

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -8,6 +8,7 @@ coverage:
 codecov:
   ci:
     - "travis-ci.com"
+    - "https://github.com/features/actions"
   notify:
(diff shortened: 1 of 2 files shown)


.codecov.yml
```diff
@@ -8,6 +8,7 @@ coverage:
 codecov:
   ci:
     - "travis-ci.com"
+    - "https://github.com/features/actions"
```
ryanjulian, line 11: I have never gotten this to work reliably within docker, which is why we pass the token directly in TravisCI

https://docs.codecov.io/docs/about-the-codecov-bash-uploader

https://github.com/rlworkgroup/garage/blob/master/Makefile#L176

Issue #1045, opened 2019-11-19
Move public utils functions into module context


ryanjulian:
Some guiding principles
* Public utilities should be available in the relevant package without importing `utils` module (e.g. `from garage.torch.algos import compute_advantages` rather than `from garage.torch.algos.loss_function_utils import compute_advantages`
* Private utilities may also be made available this way
* Utilities should be placed in the smallest package scope which contains all of their users
* A package may have more than one `utils` module if this makes testing and organization make more sense, but all public utils should end up in the package scope
* Utils modules should have _private names (e.g. `garage/torch/_utils.py`) to discourage users from importing them directly.

See comments on #1030 for how to do this in `__init__.py` so that you don't end up with circular dependencies.

TODO
* [x] `garage/misc/tensor_utils.py`
* [x] `garage/torch/utils.py`
* [x] `garage/sampler/utils.py`
* [x] `garage/experiment/nb_utils.py`
* [x] `garage/tf/misc/tensor_utils.py`
* [x] `garage/tf/optimizers/utils.py`
* [x] `garage/torch/algos/loss_function_utils.py`

PR #2020, opened 2020-09-03
Remove unused garage.sampler.utils module

ryanjulian:


ryanjulian:
Ref #1045

avnishn (approved):
didn't the rollout function used to be in here?

Issue #1399, opened 2020-05-07
Sampler's don't pickle (correctly)


krzentner:
Currently, samplers generally don't implement the pickle interface. Furthermore, the one sampler that does (`LocalSampler`) assumes that there's no state contained in its workers that wasn't present when the sampler was constructed. This breaks in the following circumstances:
 - The environment was updated after the experiment was started, and the same environment update isn't replicated on resume.
 - The agent / policy has additional state besides its parameters. For example, exploration policies may have internal state that persists across epochs.

Historically, we assumed that such state wasn't possible, and that the runner would reconstruct the sampler in an idempotent way. However, adding pickling support to the existing samplers would not be particularly onerous, so we should just do that, since it would ensure resume works correctly.

krzentner:
Note that for the case of exploration strategies with `LocalSampler`, the setup "currently" works correctly because the exploration strategy never gets pickled in the first place, and is saved as a field on the algorithm.

krzentner:
This issue will become critical once we move the algorithms that use exploration policies to use normal samplers. For the 2020.06 release, I think fixing this issue would introduce more complexity than we would like.

yeukfu:
This [example](https://github.com/rlworkgroup/garage/commit/9cf85deb52790279afcf8804ea02b0e80c34bea8) adds an `update` method to exploration_policy to give worker the knowledge of new_trajectory when resuming. I have two questions about it.

- For those exploration policies that need to keep their internal states (e.g. `AddOrnsteinUhlenbeckNoise`), it seems that it cannot resume their states because `trajectory_batch` does not contain the states. In this situation, can we add these internal states into `agent_info`?

- In this `AddExpertActions` [example](https://github.com/rlworkgroup/garage/commit/f510529850dac2bc9fd54eae823c835cf221d62f#diff-32cc0a9062698c25476e68fdd12d4c46R56), the policy maintains a `total_steps_so_far`. If there are multiple workers, should every worker only count env steps rollout by theirselves?

PR #2023, opened 2020-09-04
Pickling the exploration policy

yeukfu:
Add an `update` method to exploration policy to pickle and resume their internal states. This should close https://github.com/rlworkgroup/garage/issues/1399.
Ref: https://github.com/rlworkgroup/garage/commit/9cf85deb52790279afcf8804ea02b0e80c34bea8

### DIFF
--- a/src/garage/np/exploration_policies/add_gaussian_noise.py
+++ b/src/garage/np/exploration_policies/add_gaussian_noise.py
@@ -1,5 +1,6 @@
 """Gaussian exploration strategy."""
 import akro
+from dowel import tabular
 import numpy as np
 
 from garage.np.exploration_policies.exploration_policy import ExplorationPolicy
@@ -100,6 +101,7 @@ class AddGaussianNoise(ExplorationPolicy):
         self._total_env_steps = (self._last_total_env_steps +
                                  np.sum(episode_batch.lengths))
         self._last_total_env_steps = self._total_env_steps
+        tabular.record('AddGaussianNoise/Sigma', self._sigma())
 
     def get_param_values(self):
         """Get parameter values.
--- a/src/garage/np/exploration_policies/add_gaussian_noise.py
+++ b/src/garage/np/exploration_policies/add_gaussian_noise.py
@@ -11,40 +11,34 @@ class AddGaussianNoise(ExplorationPolicy):
     Args:
         env_spec (EnvSpec): Environment spec to explore.
         policy (garage.Policy): Policy to wrap.
+        total_timesteps (int): Total steps in the training, equivalent to
+            max_episode_length * n_epochs.
         max_sigma (float): Action noise standard deviation at the start of
             exploration.
         min_sigma (float): Action noise standard deviation at the end of the
             decay period.
-        decay_period (int): Number of episodes over which to linearly decay
-            sigma from max_sigma to min_sigma.
+        decay_ratio (float): Fraction of total steps for epsilon decay.
 
     """
 
     def __init__(self,
                  env_spec,
                  policy,
+                 total_timesteps,
                  max_sigma=1.0,
                  min_sigma=0.1,
-                 decay_period=1000000):
+                 decay_ratio=1.0):
         assert isinstance(env_spec.action_space, akro.Box)
         assert len(env_spec.action_space.shape) == 1
         super().__init__(policy)
         self._max_sigma = max_sigma
         self._min_sigma = min_sigma
-        self._decay_period = decay_period
+        self._decay_period = int(total_timesteps * decay_ratio)
         self._action_space = env_spec.action_space
-        self._iteration = 0
-
-    def reset(self, dones=None):
-        """Reset the state of the exploration.
-
-        Args:
-            dones (List[bool] or numpy.ndarray or None): Which vectorization
-                states to reset.
-
-        """
-        self._iteration += 1
-        super().reset(dones)
+        self._decrement = (self._max_sigma -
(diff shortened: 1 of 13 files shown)


ryanjulian (approved):
I'm not 100% happy with this design, but I think @krzentner will clean a lot of this up when she formalizes exploration policies in `PolicyWrapper`.

src/garage/experiment/local_runner.py
```diff
@@ -321,7 +321,12 @@ def obtain_episodes(self,
                              ' or pass `batch_size` to runner.obtain_samples.')
         episodes = None
         if agent_update is None:
-            agent_update = self._algo.policy.get_param_values()
+            policy = getattr(self._algo, 'exploration_policy', None)
```
avnishn, line 324: is the reason we are able to do this because exploration_policy is a wrapper around a policy?

avnishn (approved):
other than my one confusion, this lgtm.

src/garage/experiment/local_runner.py
yeukfu: yes

krzentner: Yeah, basically exploration policies are just policy wrappers, so to just works. I thought we already did this, but I guess I was wrong.

src/garage/np/exploration_policies/add_gaussian_noise.py
```diff
@@ -81,3 +82,36 @@ def get_actions(self, observations):
...
+        }
+
+    def set_param_values(self, params):
+        """Set param values.
+
+        Args:
+            params (np.ndarray): A numpy array of parameter values.
+
+        """
+        self._episode_num = params['episode_num']
```
krzentner, line 116: Please merge `episode_num` and `iteration`. Since you don't set `iteration` here, this change is probably broken. Please write a test to confirm that the noise level is actually changed by `set_param_values`

src/garage/np/exploration_policies/epsilon_greedy_policy.py
```diff
@@ -86,3 +87,40 @@ def get_actions(self, observations):
...
+        }
+
+    def set_param_values(self, params):
+        """Set param values.
+
+        Args:
+            params (np.ndarray): A numpy array of parameter values.
+
+        """
+        self._total_env_steps = params['total_env_steps']
```
krzentner, line 125: Because you don't set `epsilon`, this one is probably also broken.

krzentner (commented):
Hey, thanks for taking on this change. This is definitely the right approach overall, but I think you've made some small mistakes on the specifics. Because `update` is never called in the sampler, `set_param_values` needs to change the field that controls the noise level. Since this was definitely an easy mistake to make, please write some tests that `set_param_values` (without calling `update`) can change the action noise.

yeukfu:
@krzentner It is ready for review.

src/garage/np/exploration_policies/epsilon_greedy_policy.py
```diff
@@ -42,9 +42,10 @@ def __init__(self,
         self._min_epsilon = min_epsilon
         self._decay_period = int(total_timesteps * decay_ratio)
         self._action_space = env_spec.action_space
-        self._epsilon = self._max_epsilon
         self._decrement = (self._max_epsilon -
                            self._min_epsilon) / self._decay_period
+        self._total_env_steps = 0
+        self._last_total_env_steps = 0
```
krzentner, line 48: Is this field needed? It only seems to be set, not used.

src/garage/np/exploration_policies/epsilon_greedy_policy.py
```diff
@@ -77,12 +77,54 @@ def get_actions(self, observations):
...
+        """Update the exploration policy using a batch of trajectories.
+
+        Args:
+            episode_batch (EpisodeBatch): A batch of trajectories which
+                were sampled with this policy active.
+
+        """
+        self._total_env_steps = (self._last_total_env_steps +
+                                 np.sum(episode_batch.lengths))
+        self._last_total_env_steps = self._total_env_steps
```
krzentner, line 107: Can you log `EpsilonGreedyPolicy/Epsilon` here?

src/garage/np/exploration_policies/add_gaussian_noise.py
```diff
@@ -76,8 +70,56 @@ def get_actions(self, observations):
...
+        """Update the exploration policy using a batch of trajectories.
+
+        Args:
+            episode_batch (EpisodeBatch): A batch of trajectories which
+                were sampled with this policy active.
+
+        """
+        self._total_env_steps = (self._last_total_env_steps +
+                                 np.sum(episode_batch.lengths))
+        self._last_total_env_steps = self._total_env_steps
```
krzentner, line 102: Can you log `AddGaussianNoise/Sigma` here?

krzentner (approved):
Looks pretty good. There's a few small extra things that would be nice to have.

Please feel free to submit without any further review after you've addressed the comments.

src/garage/np/exploration_policies/epsilon_greedy_policy.py
yeukfu: I use it in `update` method. Every time we update the `total_env_step`, we only need add the steps of new batch into this `last_total_env_steps`, so that we can update the `total_env_step` each step in one batch and do not infect the cumulative env steps.

ryanjulian:

ryanjulian:

Issue #1785, opened 2020-07-21
Summary file with experiment parameters


chelseas:
This is a feature request.
I was playing around with rl_coach and it automatically generates a json file that contains all of an experiment's parameters. It's really helpful for repeatability and record keeping. Would love if garage had something like this too.

```
data_store: 
None
name: 
simple_rl_graph
environments: 
0: <rl_coach.environments.gym_environment.GymEnvironment object at 0x7f5cb7e30710>
q
level_managers: 
0: <rl_coach.level_manager.LevelManager object at 0x7f5cb7de0eb8>

_phase: 
RunPhase.UNDEFINED
reset_required: 
False
env_params: 
"GymVectorEnvironment" {
    "additional_simulator_parameters": {},
    "custom_reward_threshold": null,
    "default_input_filter": {
        "_observation_filters": {},
        "_reward_filters": {},
        "i_am_a_reference_filter": false,
        "name": "no_input_filter",
        "__class__": "NoInputFilter"
    },
    "default_output_filter": {
        "_action_filters": {},
        "i_am_a_reference_filter": false,
        "name": null,
        "__class__": "NoOutputFilter"
    },
    "experiment_path": "./experiments/bob213/21_07_2020-14_29",
    "frame_skip": 1,
    "human_control": false,
    "level": "CartPole-v0",
    "max_over_num_frames": 1,
    "observation_space_type": null,
    "random_initialization_steps": 0,
    "seed": null,
    "target_success_rate": 1.0
}

agent_params: 
"PolicyGradientsAgentParameters" {
    "algorithm": {
        "act_for_full_episodes": false,
        "apply_gradients_every_x_episodes": 5,
        "beta_entropy": 0,
        "discount": 0.99,
        "distributed_coach_synchronization_type": null,
        "heatup_using_network_decisions": false,
        "in_action_space": null,
        "load_memory_from_file_path": null,
        "n_step": -1,
        "num_consecutive_playing_steps": {
            "_num_steps": 1,
            "__class__": "EnvironmentSteps"
        },
        "num_consecutive_training_steps": 1,
        "num_steps_between_copying_online_weights_to_target": {
            "_num_steps": 0,
            "__class__": "TrainingSteps"
        },
        "num_steps_between_gradient_updates": 20000,
        "override_episode_rewards_with_the_last_transition_reward": false,
        "policy_gradient_rescaler": {
            "_value_": 3,
            "__objclass__": "<enum 'PolicyGradientRescaler'>",
[...]

PR #2024, opened 2020-09-04
Log all experiment start info to experiment.json

krzentner:
This is an initial shot at addressing #1785. It already helped me find a bug, so I think it's pretty worthwhile.

### DIFF
--- a/src/garage/experiment/experiment.py
+++ b/src/garage/experiment/experiment.py
@@ -10,10 +10,12 @@ import os
 import pathlib
 import subprocess
 import warnings
+import weakref
 
 import dateutil.tz
 import dowel
 from dowel import logger
+import numpy as np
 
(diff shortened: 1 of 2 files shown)


krzentner:
Example resulting file in this gist: https://gist.github.com/krzentner/1b30f48286377908751da511f266544f

ryanjulian (approved):
Sample output?

ryanjulian:
Perhaps we should just put this all into variant.json? Having one seems better than 2, if there aren't compelling reasons to separate them.

maliesa96:
Is the above example still valid?

maliesa96 (approved):

ryanjulian (approved):
LGTM -- perhaps this should go into experiment.json or trainer.json?

PR #1908, opened 2020-08-16
Use max episode length eval in SAC evaluation

avnishn:
closes #1853

PR #1905, opened 2020-08-16
Update mtsac_metaworld_mt50.py

avnishn:
update the max_episode length of this example.

PR #2029, opened 2020-09-08
Backport #1975, #1908, and #1905; Fix #2028

### LINKED ISSUE #2028: MTSAC not learning
@avnishn 
Hi! Thank you for developing and maintaining this great repo. It makes developing and benchmarking models so much easier.

I am trying to run the baseline MTSAC on metaworld mt10 and mt50. However, it seems that the model is not learning anything (the success rate and return for each task are not changing much). For mt10, the success rate remains 0 at 3.6e+06 env steps / iteration 90 (with an average return of 3.08112). I'm observing similar things for mt50. Have you run into similar problems and do you have any suggestions on how I can approach those problems?

I am using the `2020-06` release of garage (installed from pip), with the correct metaworld version (the one provided in `setup.py`. The example scripts I'm using are from the 2020-06 release branch of the repo. The only thing I changed in the code was in `python3.6/site-packages/garage/torch/algos/sac.py` line 202, where I changed the `is` in the assert to a `==`, due to an error at the runtime of `mtsac_metaworld_mt10.py`, which should not affect the algorithm's performance. I also tested the `sac_half_cheetah_batch.py`, which trains normally as the return goes to a few thousand after some iterations.

Thanks a lot for your help! Please let me know if there's anything unclear in my description, or if you need any other information about this problem.

avnishn:
There was an issue in 2020.06 where if an environment was created using garage env, then that environment was wrapped with multienv wrapper, and then garage env wrapped that multienv wrapper, that there would be errors with how the done signal is modified. Now, this issue has been fixed. 

[Here is a TBdev link](https://tensorboard.dev/experiment/ixc7yI7lQzunyaDDEtLRLw/#scalars) to an mtsac link run with the fix.

This issue is the cause for #2028


ryanjulian:
It looks like you have some serious problems with your branch.

Does this need to be backported to 2020.06?

avnishn:
> It looks like you have some serious problems with your branch.
> 
> Does this need to be backported to 2020.06?

yeah my mistake I had accidentally targeted master instead of `release-2020.06`

yeukfu (approved):

examples/torch/mtsac_metaworld_mt50.py
```diff
@@ -84,10 +84,11 @@ def mtsac_metaworld_mt50(ctxt=None, seed=1, use_gpu=False, _gpu=0):
                   qf1=qf1,
                   qf2=qf2,
                   gradient_steps_per_itr=150,
-                  max_path_length=250,
+                  max_path_length=150,
+                  max_eval_path_length=150,
```
ryanjulian, line 88: 200 here?

ryanjulian (approved):
Can you set the max path lengths to 200 (or do that in a different PR + backport?)

avnishn:
> Can you set the max path lengths to 200 (or do that in a different PR + backport?)

I think we should instead focus on getting out our garage examples using the new metaworld api because this would also require making a backport for metaworld, and we don't want people using an older metaworld version to begin with. 
 

I think one thing that is related to this that I will do is change the parameters of our mtppo and Mttrpo. I think they both use max path lengths of 128, mainly because we didn't understand the effect that this would have.

ryanjulian (dismissed):

src/garage/envs/garage_env.py
```diff
@@ -162,7 +162,8 @@ def step(self, action):
         # 'GarageEnv.TimeLimitTerminated'
         if 'TimeLimit.truncated' in info:
             info['GarageEnv.TimeLimitTerminated'] = done  # done = True always
-            done = not info['TimeLimit.truncated']
+            if info['TimeLimit.truncated']:
```
ryanjulian, line 165: i don't understand why this change is needed.

if `info['TimeLimit.truncated'] = False` and `done = True` then the environment terminated before the horizon limit and we should pass along `done = True`, i.e. `done = done and not info['TimeLimit.truncated']`

i don't think this configuration can actually appear in the current implementation of gym (which only ever populates `info['TimeLimit.truncated'] = True` at the horizon limit and omits it otherwise, but it is logically valid.

src/garage/torch/algos/mtsac.py
```diff
@@ -93,8 +97,9 @@ def __init__(
             reward_scale=1.0,
             optimizer=torch.optim.Adam,
             steps_per_epoch=1,
-            num_evaluation_trajectories=5,
-    ):
+            # yapf: disable
```
ryanjulian, line 100: why?

avnishn: The CI was complaining that the placement of the `):` characters. It called it invalid syntax whenever YAPF would place it automatically.

ryanjulian:
If you are going to fix a ton of issues in one PR, please document all of them extensively in the commit message. Right now, the PR title is misleading.

src/garage/envs/garage_env.py
avnishn: > If you are going to fix a ton of issues in one PR, please document all of them extensively in the commit message. Right now, the PR title is misleading.

Sorry I'll take care of that.

avnishn:
> If you are going to fix a ton of issues in one PR, please document all of them extensively in the commit message. Right now, the PR title is misleading.

Sorry I'll take care of that.

src/garage/envs/garage_env.py
avnishn: @ryanjulian afaict this isn't the case.

The gym time limit truncated wrapper will only ever populate a info['TimeLimit.truncated'] = False iff the environment has been stepped past its max episode length AND done = True.

Another way of phrasing this is that timelimit truncated envs only ever return info['TimeLimit.truncated'] = True when the env has been stepped past its max episode length AND done = False.

See this line inside the timelimit gym wrapper for where this logic is implemented.
https://github.com/openai/gym/blob/abb815c871447a14532558772739ccb88deee109/gym/wrappers/time_limit.py#L19

ryanjulian: simply unwinding the logic from the code you linked, recovering the original value of `done`, looks like this:

```python
if 'TimeLimit.truncated' in info:
    info['GarageEnv.TimeLimitTerminated'] = True
    done = not info['TimeLimit.truncated']
else:
    info['GarageEnv.TimeLimitTerminated'] = False
    info['TimeLimited.truncated'] = False
```

Why would you change it to be unconditional of the value of `info['TimeLimit.truncated']`? Nothing in the code you linked implies that `done` MUST be `False` when `self._elapsed_steps >= self._max_episode_steps`

src/garage/torch/algos/mtsac.py
ryanjulian: just place `# yapf: disable` on line 101 to prevent yapf from messing with the entire args block.

krzentner:
Please address Ryan's comments, but otherwise this change looks good to me.

src/garage/envs/garage_env.py
avnishn: @ryanjulian well a problem comes if we wrap an environment this way: `GarageEnv(MultiEnvwrapper([GarageEnv(env)]))`, which is what happens in all of our MT examples.

Assuming an environment hasn't hit a timelimit here's what ends up happening is the following:

The environment gets stepped and inside the `MultiEnvWrapper` the first `GarageEnv` wrapper adds the info entry `info['TimeLimit.truncated'] = False`. Then the outer `GarageEnv` wrapper evaluates the info that is returned by the inner `GarageEnv` wrapper, which has info with the entry `info['TimeLimit.truncated'] = False`. The outer `GarageEnv` Wrapper now enters the if statement that you linked above:
```python
if 'TimeLimit.truncated' in info:
    info['GarageEnv.TimeLimitTerminated'] = True
    done = not info['TimeLimit.truncated']
```
it now sets `info['GarageEnv.TimeLimitTerminated'] = True` and `done = not False` which is the incorrect behavior. `Done` should only be modified when `info['TimeLimit.truncated'] = True` never when it is `False`.

That is the logic behind why I made this change.

ryanjulian: isn't the bug then that it was wrapped twice with `GarageEnv`? what should make me believe this won't break the non-pathological case where there's only one `GarageEnv`?

ryanjulian: why not just detect double-wrapping and then vomit an error?

avnishn: > isn't the bug then that it was wrapped twice with `GarageEnv`?

I thought this was a valid use case, which is why I wrote `MultiEnvWrapper` the way that I did with nested `GarageEnvs`

> what should make me believe this won't break the non-pathological case where there's only one `GarageEnv`

There are 3 input cases here afaict if we assume that we are handling only the output of the `TimeLimitWrapper`:

1) `'TimeLimit.truncated' = True and done = True`
	
	we enter the proposed if statement `if info['TimeLimit.truncated']:` and set `done = False`
2) `'TimeLimit.truncated' = False and done = True`

	we don't enter the newly proposed if statement, and done is not modified.
3) `'TimeLimit.truncated'` isn't in `env_infos`.
	we don't enter the newly proposed if statement, and done is not modified.

ryanjulian: If it's a valid use case, it's extremely strange and surprising to users -- why would I wrap the same thing twice? Is wrapping twice strictly necessary in the first place?

In any case, it's apparently *not* valid to wrap `GarageEnv` twice, because apparently `GarageEnv` isn't idempotent. So either you need to make `GarageEnv` idempotent, or you need to detect double-wrapping and throw an error.

Here's the truth table as I read it:
| `'TimeLimited.truncated' in info` | `info['TimeLimit.truncated']` | done  | done = |
|---|---|---|---|
| `False` | N/A | `False` | `False` |
| `False` | N/A | `True` | `True` |
| `True` | `False` | `False` | `True` |
| `True` | `False` | `True` | `True` |
| `True` | `True` | `False` | `False` |
| `True` | `True` | `True` | `False` |

As long as your backport implements this truth table for arbitrary nestings of `GarageEnv` (or only one wrapping of `GarageEnv`, and throws an error if you do it twice), I am fine with it.

ryanjulian: If it's a valid use case, it's really strange. What about the definition of `GarageEnv` makes it seem like it would be intended to be used twice? That would be really surprising to any user.

In any case, it's apparently *not* a valid use case, because `GarageEnv` is apparently not idempotent. So you *either* need to make `GarageEnv` idempotent, *or* you can leave it non-idempotent, but detect double-wrapping and throw an error.

Here's the truth table as I see it:
| `'TimeLimit.truncated' in info` | `info['TimeLimit.truncated']` | `done` | `done = ` |
|---|---|---|---|
| `False` | N/A | `False` | `False` |
| `False` | N/A | `True` | `True` |
| `True` | `False` | `False` | `True` |
| `True` | `False` | `True` | `True` |
| `True` | `True` | `False` | `False` |
| `True` | `True` | `True` | `False` |

As long as your fix implements this truth table for arbitrary nestings of GarageEnv (or only one wrapping, but throws errors on double-wrapping), I'm happy.

Please prove it with a test, since obviously this is getting pretty subtle.

avnishn: gotcha, yes it implements this change. Let me go ahead and implement a test that checks this on master and on this release branch.

zequnyu (approved):

avnishn:
@ryanjulian I investigated whether the GarageEnv Idempotent fix would need to be ported to master, and it doesn't. GymEnv can only wrap envs that implement gym.env, and Eric rewrote Multienv wrapper to no longer use GymEnv internally.

PR #1845, opened 2020-08-04
Use garage environment

AiRuiChen:
1. Use `garage.Environment` for all garage environments.
2. Rename `GarageEnv` as `GymEnv` - it should only handle gym-related environments.
3. Define `NormalizedEnv` or `normalize()` usage - it should only wrap environments that implement `garage.Environment`.
4. Move `EnvSpec` to `_environment.py`
5. Create `EnvStep` in `_environment.py` and let `Environment.step()` return this dtype.

PR #2031, opened 2020-09-09
Add episode_info to TimeStep, TimeStepBatch, and EpisodeBatch

nicolengsy:
Fixing issue #1845, allowing for `episode_info` (returned by `env.reset()`) to be used/passed around in these dtypes.

### DIFF
--- a/src/garage/_dtypes.py
+++ b/src/garage/_dtypes.py
@@ -5,7 +5,8 @@ import enum
 import akro
 import numpy as np
 
-from garage.np import concat_tensor_dict_list, slice_nested_dict
+from garage.np import (concat_tensor_dict_list, slice_nested_dict,
+                       stack_tensor_dict_list)
 
 # pylint: disable=too-many-lines
 
@@ -13,6 +14,7 @@ from garage.np import concat_tensor_dict_list, slice_nested_dict
 class EpisodeBatch(
         collections.namedtuple('EpisodeBatch', [
             'env_spec',
+            'episode_infos',
             'observations',
             'last_observations',
             'actions',
@@ -49,6 +51,11 @@ class EpisodeBatch(
     Attributes:
         env_spec (EnvSpec): Specification for the environment from
             which this data was sampled.
+        episode_infos (dict[str, np.ndarray]): A dict of numpy arrays
+            containing the episode-level information of each episode. Each
+            value of this dict should be a numpy array of shape :math:`(N,
+            S^*)`. For example, in goal-conditioned reinforcement learning this
+            could contain the goal state for each episode.
         observations (numpy.ndarray): A numpy array of shape
             :math:`(N \bullet [T], O^*)` containing the (possibly
             multi-dimensional) observations for all time steps in this batch.
@@ -64,13 +71,15 @@ class EpisodeBatch(
         rewards (numpy.ndarray): A numpy array of shape
             :math:`(N \bullet [T])` containing the rewards for all time steps
             in this batch.
-        env_infos (dict): A dict of numpy arrays arbitrary environment state
-            information. Each value of this dict should be a numpy array of
-            shape :math:`(N \bullet [T])` or :math:`(N \bullet [T], S^*)`.
-        agent_infos (numpy.ndarray): A dict of numpy arrays arbitrary agent
-            state information. Each value of this dict should be a numpy array
-            of shape :math:`(N \bullet [T])` or :math:`(N \bullet [T], S^*)`.
-            For example, this may contain the hidden states from an RNN policy.
+        env_infos (dict[str, np.ndarray]): A dict of numpy arrays arbitrary
+            environment state information. Each value of this dict should be
+            a numpy array of shape :math:`(N \bullet [T])` or :math:`(N \bullet
+            [T], S^*)`.
+        agent_infos (dict[str, np.ndarray]): A dict of numpy arrays arbitrary
+            agent state information. Each value of this dict should be a numpy
+            array of shape :math:`(N \bullet [T])` or :math:`(N \bullet [T],
+            S^*)`.  For example, this may contain the hidden states from an RNN
+            policy.
         step_types (numpy.ndarray): A numpy array of `StepType with shape
             :math:`(N,)` containing the time step types for all transitions in
             this batch.
@@ -85,8 +94,8 @@ class EpisodeBatch(
     """
     __slots__ = ()
 
-    def __new__(cls, env_spec, observations, last_observations, actions,
-                rewards, env_infos, agent_infos, step_types,
+    def __new__(cls, env_spec, episode_infos, observations, last_observations,
+                actions, rewards, env_infos, agent_infos, step_types,
                 lengths):  # noqa: D102
         # pylint: disable=too-many-branches
 
@@ -230,9 +239,23 @@ class EpisodeBatch(
                 'step_types tensor must be dtype `StepType`, but got tensor '
                 'of dtype {} instead.'.format(step_types.dtype))
 
-        return super().__new__(EpisodeBatch, env_spec, observations,
-                               last_observations, actions, rewards, env_infos,
-                               agent_infos, step_types, lengths)
+        # episode_infos
+        for key, val in episode_infos.items():
+            if not isinstance(val, (dict, np.ndarray)):
+                raise ValueError(
+                    'Each entry in episode_infos must be a numpy array,'
+                    'but got key {} with value type {} instead.'.format(
+                        key, type(val)))
+            if (isinstance(val, np.ndarray) and val.shape[0] != len(lengths)):
+                raise ValueError(
+                    'Each entry in episode_infos must have a batch dimension '
+                    'of length {}, but got key {} with batch size {} instead.'.
+                    format(len(lengths), key, val.shape[0]))
+
+        return super().__new__(EpisodeBatch, env_spec, episode_infos,
+                               observations, last_observations, actions,
+                               rewards, env_infos, agent_infos, step_types,
+                               lengths)
 
     @classmethod
     def concatenate(cls, *batches):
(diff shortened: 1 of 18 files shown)


src/garage/_dtypes.py
```diff
@@ -67,7 +68,7 @@ class EpisodeBatch(
         env_infos (dict): A dict of numpy arrays arbitrary environment state
             information. Each value of this dict should be a numpy array of
             shape :math:`(N \bullet [T])` or :math:`(N \bullet [T], S^*)`.
-        agent_infos (numpy.ndarray): A dict of numpy arrays arbitrary agent
+        agent_infos (dict): A dict of numpy arrays arbitrary agent
```
ryanjulian, line 71:
```suggestion
        agent_infos (dict[str, np.ndarray]): A dict of numpy arrays arbitrary agent
```

src/garage/_dtypes.py
```diff
@@ -77,6 +78,8 @@ class EpisodeBatch(
         lengths (numpy.ndarray): An integer numpy array of shape :math:`(N,)`
             containing the length of each episode in this batch. This may be
             used to reconstruct the individual episodes.
+        episode_infos (dict): A dict of numpy arrays containing the episode-level
```
ryanjulian, line 81:
```suggestion
        episode_infos (dict[str, np.ndarray]): A dict of numpy arrays containing the episode-level
```

ryanjulian, line 81: i think you should list this as the first member/argument, since logically its existence precedes everything else

src/garage/_dtypes.py
```diff
@@ -77,6 +78,8 @@ class EpisodeBatch(
...
+            information of each episode.
```
ryanjulian, line 82: "For instance, in goal-conditioned reinforcement learning this could contain the goal state for each episode."

src/garage/_dtypes.py
ryanjulian: please add shape information to the documentation. These need to have shape :math:`(N, S^*)` (i.e. one for each episode with arbitrary shape of an individual sample.

src/garage/_dtypes.py
```diff
@@ -230,9 +233,25 @@ def __new__(cls, env_spec, observations, last_observations, actions,
...
+        # episode_infos
+        for key, val in episode_infos.items():
+            if not isinstance(val, (dict, np.ndarray)):
+                raise ValueError(
+                    'Each entry in episode_infos must be a numpy array or '
+                    'dictionary, but got key {} with value type {} instead.'
+                    'instead'.format(key, type(val)))
+
+            if (isinstance(val, np.ndarray)
+                    and val.shape[0] != inferred_batch_size):
```
ryanjulian, line 245: this is incorrect -- inferred_batch_size is the number of time steps, but we need "number of episodes" episode_infos instead (i.e. `num_episodes = len(lengths)`)

src/garage/_dtypes.py
```diff
@@ -521,7 +555,7 @@ class TimeStep(
         env_spec (EnvSpec): Specification for the environment from which this
             data was sampled.
         observation (numpy.ndarray): A numpy array of shape :math:`(O^*)`
-            containing the observation for the this time step in the
+            containing the observation for this time step in the
             environment. These must conform to
             :obj:`EnvStep.observation_space`.
```
ryanjulian, line 560:
```suggestion
            :attr:`EnvStep.observation_space`.
```

src/garage/_dtypes.py
```diff
@@ -548,6 +582,8 @@ class TimeStep(
         step_type (StepType): a :class:`~StepType` enum value. Can be one of
             :attribute:`~StepType.FIRST`, :attribute:`~StepType.MID`,
             :attribute:`~StepType.TERMINAL`, or :attribute:`~StepType.TIMEOUT`.
+        episode_info (dict): A dict of episode-level information.
```
ryanjulian, line 585: as with EpisodeBatch, please make this the first member/argument, and include the shape information. using the terminology from above, it would a dict of numpy arrays of shape `(S*^,)` (there is no batch dimension, because this is a single time step)

src/garage/_dtypes.py
```diff
@@ -641,7 +678,8 @@ def output_space(self):
 class TimeStepBatch(
         collections.namedtuple('TimeStepBatch', [
             'env_spec', 'observations', 'actions', 'rewards',
-            'next_observations', 'env_infos', 'agent_infos', 'step_types'
+            'next_observations', 'env_infos', 'agent_infos', 'step_types', 
+            'episode_infos'
```
ryanjulian, line 682: as above, please make this the first member

src/garage/_dtypes.py
```diff
@@ -669,6 +707,10 @@ class TimeStepBatch(
         step_types (numpy.ndarray): A numpy array of `StepType with shape (
             batch_size,) containing the time step types for all transitions in
             this batch.
+        episode_infos (dict): A dict of episode-level information.
+            It contains information of he entire episode， which could be
+            needed to determine the first action (e.g. in the case of
+            goal-conditioned or MTRL.)
```
ryanjulian, line 713:
```suggestion
        episode_infos (dict): A dict of episode-level information.
            It contains information of he entire episode, which could be
            needed to determine the first action (e.g. in the case of
            goal-conditioned or MTRL it could contain the goal or task for this episode, respectively).
```

src/garage/_dtypes.py
```diff
@@ -807,10 +849,26 @@ def __new__(cls, env_spec, observations, actions, rewards,
...
+        # episode_infos
+        for key, val in episode_infos.items():
+            if not isinstance(val, (dict, np.ndarray)):
+                raise ValueError(
+                    'Each entry in episode_infos must be a numpy array or '
+                    'dictionary, but got key {} with value type {} instead.'
+                    'instead'.format(key, type(val)))
+
+            if (isinstance(val, np.ndarray)
+                    and val.shape[0] != inferred_batch_size):
```
ryanjulian, line 862: this is correct in the case of TimeStepBatch (here), but not in the base of EpisodeBatch (above).

src/garage/_dtypes.py
```diff
@@ -669,6 +707,10 @@ class TimeStepBatch(
         step_types (numpy.ndarray): A numpy array of `StepType with shape (
             batch_size,) containing the time step types for all transitions in
             this batch.
+        episode_infos (dict): A dict of episode-level information.
```
ryanjulian, line 710: please include the shape information and validate below. for a TimeStepBatch, we need episode_info for each TimeStep, because the TimeSteps don't need to share a "parent" episode (i.e. unlike EpisodeBatch, we will need a copy for each TimeStep).

using the terminology from above, this is :math:`(N, S^*)`

src/garage/_dtypes.py
```diff
@@ -967,7 +1041,10 @@ def from_time_step_list(cls, env_spec, ts_samples):
                 * step_types (numpy.ndarray): A numpy array of `StepType with
                 shape (batch_size,) containing the time step types for all
                     transitions in this batch.
-
+                * episode_infos (dict): The episode-level information.
```
ryanjulian, line 1044: please include shape information

ryanjulian, line 1044:
```suggestion
                * episode_infos (dict[str, np.ndarray]): The episode-level information.
```

ryanjulian (commented):
looking good. please see my comments about leaving shape documentation and validation. i think you should also make episode_infos the first member of each collective type, since it has the "broadest" scope (i.e. it is the first data point to appear, chronologically, in any sampling process).

ryanjulian:
@nicolengsy heads up that usually reviewers ignore PRs listed as "draft" -- i recall you said this was ready in stand-up today so I'm taking a look now.

avnishn (approved):
I'm not sure if I see anywhere where you test inputting invalid data into your episode infos. This is a small but crucial detail that can lead to bugs being produced later.

Though your test coverage is high, I recommend writing a small test to cover this datatype rule.

src/garage/_dtypes.py
```diff
@@ -230,9 +236,23 @@ def __new__(cls, env_spec, observations, last_observations, actions,
...
                 'of dtype {} instead.'.format(step_types.dtype))
 
-        return super().__new__(EpisodeBatch, env_spec, observations,
-                               last_observations, actions, rewards, env_infos,
-                               agent_infos, step_types, lengths)
+        # episode_infos
+        for key, val in episode_infos.items():
+            if not isinstance(val, (dict, np.ndarray)):
+                raise ValueError(
+                    'Each entry in episode_infos must be a numpy array or '
```
ryanjulian, line 243: according to the docstring the items can only be numpy arrays -- why allow dicts?

src/garage/_dtypes.py
```diff
@@ -807,10 +854,26 @@ def __new__(cls, env_spec, observations, actions, rewards,
                     'dimension of '
                     'length {}, but got key {} with batch size {} instead.'.
                     format(inferred_batch_size, key, val.shape[0]))
+        # episode_infos
+        for key, val in episode_infos.items():
+            if not isinstance(val, (dict, np.ndarray)):
+                raise ValueError(
+                    'Each entry in episode_infos must be a numpy array or '
+                    'dictionary, but got key {} with value type {} instead.'
```
ryanjulian, line 862: docstring says it must be a numpy array

src/garage/sampler/_dtypes.py
```diff
@@ -81,6 +81,7 @@ def to_batch(self):
                                                   dtype=StepType),
                             env_infos=env_infos,
                             agent_infos=agent_infos,
+                            episode_infos=dict(),
```
ryanjulian, line 84: you need to modify this class to track episode_infos as well -- it is used to create EpisodeBatches in the sampler. If they aren't captured here, they will be lost.

krzentner (approved):
This looks good. It touches a decent amount of the code base, so let's merge it. (Also, see my one small comment.)

PR #2034, opened 2020-09-11
Save git hash for auto benchmarking

zequnyu:


ryanjulian (approved):
I think KR already has a utility for this in wrap_experiment. You should check

haydenshively (approved):

Issue #1961, opened 2020-08-24
Ray tests are flaking on GH Actions


ryanjulian:
See https://github.com/rlworkgroup/garage/pull/1959/checks?check_run_id=1016422383#step:6:111

avnishn:
I think I might need more information, the checks on this don't reveal a bug, and I'm unsure how to view the logs from gH actions builds vs travis builds

ryanjulian:
Here's another example https://github.com/rlworkgroup/garage/runs/1016409427?check_suite_focus=true#step:6:121

PR #2036, opened 2020-09-11
Disable Ray Dashboard

avnishn:
Currently, when we test examples that used the Ray Sampler,
which is every on policy rl algorithm, we shut down the example
after 1 epoch. Because of this, ray.shutdown() is never called
which is necessary for closing ray features such as the ray
dashboard. The dashboard by default tries to attach to the same
port, and so because it hasn't been properly freed after the
example is prematurely shutdown, sometimes the ray sampler
fails to properly init when a new example is started.

To work around this, I have disabled the behavior where the
ray sampler activates its ray dashboard. Users can renable
the dashboard by initing ray inside of their launchers,
like we do in our trpo_swimmer_ray_sampler.py example.

See #1961 for more information


gitanshu (approved):
Sounds good to me

krzentner (approved):

ziyiwu9494 (approved):

PR #2037, opened 2020-09-12
Add RL2 doc

yeukfu:
close https://github.com/rlworkgroup/garage/issues/2013.

### DIFF
--- a/docs/user/algo_rl2.md
+++ b/docs/user/algo_rl2.md
@@ -19,8 +19,6 @@
      - `garage.tf.algos.RL2 <../_autoapi/garage/tf/algos/index.html#garage.tf.algos.RL2>`_
    * - **Code**
      - `garage/tf/algos/rl2.py <https://github.com/rlworkgroup/garage/blob/master/src/garage/tf/algos/rl2.py>`_
-   * - **Examples**
-     - `examples <algo_rl2.html#examples>`_
 ```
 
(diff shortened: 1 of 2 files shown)


docs/user/algo_rl2.md
```diff
@@ -0,0 +1,52 @@
...
+     - `examples <algo_rl2.html#examples>`_
+```
+
+When sampling for RL<sup>2</sup>, there are more than one environments to be
+sampled from. In the original implementation, within each trial, all episodes
+sampled will be concatenated into one single episode, and fed to the
+inner algorithm. Thus, returns and advantages are calculated across the
+episode.
+
+User should not instantiate RL2 directly. Currently garage supports [PPO](https://github.com/rlworkgroup/garage/blob/master/src/garage/tf/algos/rl2ppo.py)
```
ryanjulian, line 32: this is confusing on its own -- maybe you should add a subsection on RL2PPO and RL2TRPO

docs/user/algo_rl2.md
```diff
@@ -0,0 +1,52 @@
...
+and [TRPO](https://github.com/rlworkgroup/garage/blob/master/src/garage/tf/algos/rl2trpo.py)
+as the inner algorithm.
+
+## Examples
```
ryanjulian, line 36: can you include more examples? you can describe each example with a title

ryanjulian (approved):

haydenshively (approved):

PR #2040, opened 2020-09-14
Refactor away TFTrainer

zequnyu:
#1501.


src/garage/trainer.py
```diff
@@ -155,6 +159,20 @@ def __init__(self, snapshot_config):
         self._worker_class = None
         self._worker_args = None
 
+        # only used for TensorFlow algorithms
+        self._tf = tf
```
ryanjulian, line 163: rather than having a boolean, can you just make a session if you find tensorflow in sys.modules?

src/garage/trainer.py
```diff
@@ -155,6 +160,20 @@ def __init__(self, snapshot_config):
         self._worker_class = None
         self._worker_args = None
 
+        # only used for TensorFlow algorithms
+        self._tf = 'tensorflow' in sys.modules
```
krzentner, line 164: I *really* don't like this approach. The existence of a tensorflow session in a python process permanently alters the properties of the process. Looking at a launcher, it should be immediately clear if and when a session is created. Here, a session will almost always be created (even when using PyTorch), because we've inevitably tried to import `tf` somewhere.

PR #2043, opened 2020-09-14
Remove unnecessary list comprehensions from VPG

haydenshively:
Part of #1110

The `train_once` function now takes an `EpisodeBatch` rather than an `EpisodeBatch` that's been converted to a `list(dict)` with `.to_list()`. Unfortunately, the algorithm requires the observations/rewards/returns to be padded to `max_episode_length` for certain operations, so I had to use `.to_list()` inside the algo.

Also fixed a small `__init__` arg bug in `RaySampler`

### DIFF
--- a/src/garage/torch/algos/vpg.py
+++ b/src/garage/torch/algos/vpg.py
@@ -7,7 +7,7 @@ import numpy as np
 import torch
 import torch.nn.functional as F
 
-from garage import EpisodeBatch, log_performance
+from garage import log_performance
 from garage.np import discount_cumsum
 from garage.np.algos import RLAlgorithm
 from garage.sampler import RaySampler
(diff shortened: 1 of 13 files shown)


src/garage/torch/algos/vpg.py
```diff
@@ -447,13 +444,13 @@ def _compute_objective(self, advantages, obs, actions, rewards):
 
         return log_likelihoods * advantages
 
-    def _process_samples(self, paths):
+    def _process_samples(self, eps):
```
ryanjulian, line 447: is there any way to eliminate this method? that's kind of the idea behind EpisodeBatch

src/garage/sampler/ray_sampler.py
```diff
@@ -30,9 +30,7 @@ class RaySampler(Sampler):
     def __init__(self, worker_factory, agents, envs):
         # pylint: disable=super-init-not-called
         if not ray.is_initialized():
-            ray.init(log_to_driver=False,
-                     ignore_reinit_error=True,
-                     include_dashboard=False)
+            ray.init(log_to_driver=False, ignore_reinit_error=True)
```
krzentner, line 33: Please revert these lines. `include_dashboard=False` is intentional

src/garage/torch/algos/vpg.py
krzentner: The main thing this method does is pad out paths so that we can compute advantages in a differentiable way. I'm not sure that's so important, but it's why `EpisodeBatch` alone does not suffice.

krzentner (approved):
This change looks like a net improvement, so I say merge it. I'm a bit surprised that the main code path of `garage.torch.algos.vpg` doesn't appear to have test coverage.

Also, please note my comment on `RaySampler`.

krzentner (approved):
This is a definite improvement. Please merge it once the CI passes.

krzentner:
It looks like this broke MAML, since it calls `inner_algo._process_samples`. Feel free to copy whatever logic you need into `MAML._process_samples`

yeukfu (approved):

PR #2044, opened 2020-09-14
Use EpisodeBatch in Torch SAC

haydenshively:
Part of #1110

I think this really should use `TimeStepBatch`, but as far as I can tell, the `trainer` doesn't have a way of retrieving one of those. It only deals in `EpisodeBatch` or the `list(dict)` type that's scattered everywhere.

### DIFF
--- a/src/garage/torch/algos/sac.py
+++ b/src/garage/torch/algos/sac.py
@@ -8,7 +8,7 @@ import numpy as np
 import torch
 import torch.nn.functional as F
 
-from garage import log_performance, obtain_evaluation_episodes, StepType
+from garage import log_performance, obtain_evaluation_episodes
 from garage.np.algos import RLAlgorithm
 from garage.sampler import FragmentWorker, RaySampler
 from garage.torch import dict_np_to_torch, global_device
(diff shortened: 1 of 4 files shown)


src/garage/torch/algos/sac.py
```diff
@@ -217,12 +209,12 @@ def train(self, trainer):
 
         return np.mean(last_return)
 
-    def train_once(self, itr=None, paths=None):
+    def train_once(self, itr=None, eps=None):
```
ryanjulian, line 212: please make this _private

src/garage/torch/algos/sac.py
```diff
@@ -217,12 +209,12 @@ def train(self, trainer):
...
         """Complete 1 training iteration of SAC.
 
         Args:
             itr (int): Iteration number. This argument is deprecated.
-            paths (list[dict]): A list of collected paths.
+            eps (EpisodeBatch): An EpisodeBatch.
```
ryanjulian, line 217: this needs to take a TimeStepBatch -- you should get it from the replay buffer.

src/garage/torch/algos/sac.py
```diff
@@ -316,17 +308,17 @@ def _actor_objective(self, samples_data, new_actions, log_pi_new_actions):
         Note:
             samples_data's entries should be torch.Tensor's with the following
             shapes:
-                observation: :math:`(N, O^*)`
-                action: :math:`(N, A^*)`
-                reward: :math:`(N, 1)`
-                terminal: :math:`(N, 1)`
-                next_observation: :math:`(N, O^*)`
+                observations: :math:`(N, O^*)`
```
ryanjulian, line 311: a goal in this PR should be to remove samples_data and use TimeStepBatch

krzentner: In this case the input is a `torch.Tensor`, so `TimeStepBatch` can't be used. However, looking at the definition of this method, `SAC._get_log_alpha`, and the override `MTSAC._get_log_alpha`, only the observations are ever used. It would probably be best to refactor all three methods to only take the `observations` tensor.

krzentner (approved):
This is definitely a step in the right direction, but it looks like there's a bunch of small improvements that can be made.

I'm fine with making those improvements in this PR or a later one.

ryanjulian (approved):

PR #2070, opened 2020-09-19
Use TimeStepBatch in off policy algos in tf

yeukfu:
Address part of https://github.com/rlworkgroup/garage/issues/1110.

### DIFF
--- a/src/garage/_dtypes.py
+++ b/src/garage/_dtypes.py
@@ -940,7 +940,8 @@ class TimeStepBatch(
             })
         return samples
 
-    def get_terminals(self):
+    @property
+    def terminals(self):
         """Get an array of boolean indicating ternianal information.
 
         Returns:
--- a/src/garage/_dtypes.py
+++ b/src/garage/_dtypes.py
@@ -655,8 +655,7 @@ class TimeStepBatch(
         actions (numpy.ndarray): Non-flattened array of actions. Should
             have shape (batch_size, S^*) (the unflattened action space of the
             current environment).
-        rewards (numpy.ndarray): Array of rewards of shape (batch_size,) (1D
-            array of length batch_size).
+        rewards (numpy.ndarray): Array of rewards of shape (batch_size, 1).
         next_observation (numpy.ndarray): Non-flattened array of next
             observations. Has shape (batch_size, S^*). next_observations[i] was
             observed by the agent after taking actions[i].
@@ -759,6 +758,12 @@ class TimeStepBatch(
                 'Expected batch dimension of actions to be length {}, but got '
                 'length {} instead.'.format(inferred_batch_size,
                                             actions.shape[0]))
+
+        # rewards
+        if rewards.shape != (inferred_batch_size, 1):
+            raise ValueError(
+                'Rewards tensor must have shape {}, but got shape {} '
+                'instead.'.format((inferred_batch_size, 1), rewards.shape))
 
         # step_types
         if step_types.shape[0] != inferred_batch_size:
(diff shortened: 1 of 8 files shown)


src/garage/tf/algos/ddpg.py
```diff
@@ -352,14 +353,15 @@ def _optimize_policy(self):
...
-        observations = transitions['observations']
-        next_observations = transitions['next_observations']
-        rewards = transitions['rewards'].reshape(-1, 1)
-        actions = transitions['actions']
-        terminals = transitions['terminals'].reshape(-1, 1)
+        observations = timesteps.observations
+        rewards = timesteps.rewards.reshape(-1, 1)
+        actions = timesteps.actions
+        next_observations = timesteps.next_observations
+        terminals = np.array([[s == StepType.TERMINAL]
```
ryanjulian, line 363: consider adding a helper function to TimeStepBatch for this

src/garage/tf/algos/ddpg.py
```diff
@@ -352,14 +353,15 @@ def _optimize_policy(self):
...
+        timesteps = self._replay_buffer.sample_timesteps(
             self._buffer_batch_size)
 
-        observations = transitions['observations']
-        next_observations = transitions['next_observations']
-        rewards = transitions['rewards'].reshape(-1, 1)
-        actions = transitions['actions']
-        terminals = transitions['terminals'].reshape(-1, 1)
+        observations = timesteps.observations
+        rewards = timesteps.rewards.reshape(-1, 1)
```
ryanjulian, line 360: hmm why does the shape differ from what's needed? you would expect TimeStepBatch to already be perfect for vanilla off-policy RL, so maybe we should change the internal array shape

src/garage/tf/algos/td3.py
```diff
@@ -376,14 +377,15 @@ def _optimize_policy(self, itr):
...
-        rewards = transitions['rewards']
-        actions = transitions['actions']
-        next_observations = transitions['next_observations']
-        terminals = transitions['terminals']
+        observations = timesteps.observations
+        rewards = timesteps.rewards.reshape(-1, 1)
+        actions = timesteps.actions
+        next_observations = timesteps.next_observations
+        terminals = np.array([[s == StepType.TERMINAL]
+                              for s in timesteps.step_types])
```
ryanjulian, line 388: is there a reason you unpack all of these int aliases? IMO the most idiomatic usage would be to pass the entire TimeStepBatch data structure to TF (i.e. to self._train_qf). At least that's the idea.

src/garage/_dtypes.py
```diff
@@ -935,6 +940,16 @@ def to_time_step_list(self):
             })
         return samples
 
+    def get_terminals(self):
```
krzentner, line 943: I don't think you need `get_` here, it's perfectly clean and nicer looking to just have this method named `.terminals`. It could also reasonably be a property, but I don't have a strong opinion about that.

krzentner (approved):
Looks good. Glad to see types getting clearer throughout the code base.

I have one small nit on the method name `get_terminals`, which should *probably* just be a property called `.terminals`.

src/garage/_dtypes.py
```diff
@@ -760,6 +759,12 @@ def __new__(cls, env_spec, observations, actions, rewards,
                 'length {} instead.'.format(inferred_batch_size,
                                             actions.shape[0]))
 
+        # rewards
+        if rewards.shape != (inferred_batch_size, 1):
+            raise ValueError(
```
maliesa96, line 764: Can you please test this?

maliesa96:
LGTM other than a small nitpick

maliesa96 (approved):

PR #2076, opened 2020-09-22
Add torch DQN

maliesa96:
This also adds several smaller features:
 
- `torch/examples/watch_atari.py`: use a trained agent to play atari.
- snapshotting: `snapshot_mode='gap_overwrite'` will save a snapshot every `snapshot_gap` but will overwrite the previous one. Also added error handling for invalid argument combinations (eg. `'snapshot_mode='last'` and `snapshot_gap=10`)
 - `torch/examples/dqn_atari.py`: train on atari environments.


maliesa96:
Benchmarks:
image (1)
with 95% confidence interval (rather than std dev):
subplot

src/garage/_functions.py
```diff
@@ -68,7 +69,7 @@ def rollout(env,
             *,
             max_episode_length=np.inf,
             animated=False,
-            speedup=1,
+            sleep=None,
```
ryanjulian, line 72: `pause_per_frame`?

examples/sim_policy.py
```diff
@@ -53,7 +53,6 @@ def query_yes_no(question, default='yes'):
                         type=int,
                         default=1000,
                         help='Max length of episode')
-    parser.add_argument('--speedup', type=float, default=1, help='Speedup')
```
avnishn, line 56: why are you removing this?

src/garage/torch/algos/dqn.py
```diff
@@ -0,0 +1,268 @@
...
+            trainer (Trainer): Experiment trainer.
+
+        Returns:
+            float: The average return in last epoch cycle.
+
+        """
+        if not self._eval_env:
+            self._eval_env = trainer.get_env_copy()
+        last_returns = [float('nan')]
+        trainer.enable_logging = False
```
avnishn, line 130: what is the purpose of this flag? Just curious as I haven't seen it in other algorithms.

src/garage/torch/algos/dqn.py
```diff
@@ -0,0 +1,268 @@
...
+                if (cycle == 0 and self.replay_buffer.n_transitions_stored >=
+                        self._min_buffer_size):
+                    trainer.enable_logging = True
+                    logger.log('Evaluating policy')
+
+                    epsilon_before = self.exploration_policy.epsilon
+                    eval_eps = obtain_evaluation_episodes(
+                        self.exploration_policy,
+                        self._eval_env,
+                        num_eps=10,
```
avnishn, line 145: This needs to be a hyperparameter of your algorithm.

avnishn, line 145: can you also provide the parameter to your algorithm for deterministic/nondeterministic evaluation?

src/garage/torch/algos/dqn.py
```diff
@@ -0,0 +1,268 @@
...
+                        max_episode_length=self._max_episode_length_eval)
+                    self.exploration_policy.epsilon = epsilon_before
+
+                    last_returns = log_performance(trainer.step_itr,
+                                                   eval_eps,
+                                                   discount=self._discount)
+                    self._episode_reward_mean.extend(last_returns)
+                    tabular.record('Evaluation/100EpRewardMean',
```
avnishn, line 153: why is it called 100EpRewardMean? 100 sounds like some type of hyperparam.

src/garage/torch/algos/dqn.py
```diff
@@ -0,0 +1,268 @@
...
+                    self._buffer_batch_size)
+                samples['rewards'] *= self._reward_scale
+                qf_loss, y, q = tuple(v.cpu().numpy()
+                                      for v in self.optimize_qf(samples))
+
+                self._episode_qf_losses.append(qf_loss)
+                self._epoch_ys.append(y)
+                self._epoch_qs.append(q)
+
+        if itr % self._steps_per_epoch == 0:
```
avnishn, line 184: ideally, can you break this out into a separate logging function?

src/garage/torch/algos/dqn.py
```diff
@@ -0,0 +1,268 @@
...
+    def to(self, device=None):
+        """Put all the networks within the model on device.
+
+        Args:
+            device (str): ID of GPU or CPU.
+
+        """
+        if device is None:
+            device = global_device()
+        torch.set_default_tensor_type(torch.cuda.FloatTensor)
```
avnishn, line 265: does this work on CPU only devices? Can you confirm this? It makes more sense to use our set_gpu function than this line over here.

tests/garage/experiment/test_snapshotter.py
```diff
@@ -47,6 +48,17 @@ def test_snapshotter(self, mode, files):
 
     def test_invalid_snapshot_mode(self):
         with pytest.raises(ValueError):
-            snapshotter = Snapshotter(
-                snapshot_dir=self.temp_dir.name, snapshot_mode='invalid')
+            snapshotter = Snapshotter(snapshot_dir=self.temp_dir.name,
+                                      snapshot_mode='invalid')
             snapshotter.save_itr_params(2, {'testparam': 'invalid'})
+
+    def test_conflicting_params(self):
```
avnishn, line 55: thanks for adding this!

avnishn: perhaps it belongs in a separate pr, although its a small change.

make sure to add it to your commit message.

src/garage/torch/algos/dqn.py
```diff
@@ -0,0 +1,268 @@
...
+        if self._clip_reward is not None:
+            rewards_clipped = torch.clamp(rewards, -1 * self._clip_reward,
+                                          self._clip_reward)
+        y_target = (rewards_clipped +
+                    (1.0 - terminals) * self._discount * best_qvals)
+        y_target = y_target.squeeze(1)
+
+        # optimize qf
+        qvals = self._qf(inputs)
+        selected_qs = torch.sum(qvals * actions, axis=1)
```
avnishn, line 242: Ideally, can you break out your losses into separate functions, and write tests for them to ensure that they always maintain the same values given the same inputs. Those types of tests are the strongest for protecting against performance regressions.

avnishn (commented):
Left multiple comments on ways in which this PR can be improved. Some need to be addressed, some are nice to haves. Good job!

src/garage/torch/algos/dqn.py
```diff
@@ -0,0 +1,268 @@
...
+                trainer.step_path = trainer.obtain_episodes(trainer.step_itr)
+                self.train_once(trainer.step_itr, trainer.step_path)
+                if (cycle == 0 and self.replay_buffer.n_transitions_stored >=
+                        self._min_buffer_size):
+                    trainer.enable_logging = True
+                    logger.log('Evaluating policy')
+
+                    epsilon_before = self.exploration_policy.epsilon
+                    eval_eps = obtain_evaluation_episodes(
+                        self.exploration_policy,
```
yeukfu, line 143: Curious why use exploration_policy for evaluating. It seems that other algorithms use policy here.

src/garage/torch/algos/dqn.py
```diff
@@ -0,0 +1,268 @@
...
+
+        """
+        if not self._eval_env:
+            self._eval_env = trainer.get_env_copy()
+        last_returns = [float('nan')]
+        trainer.enable_logging = False
+
+        for _ in trainer.step_epochs():
+            for cycle in range(self._steps_per_epoch):
+                trainer.step_path = trainer.obtain_episodes(trainer.step_itr)
```
DennisCraandijk, line 134: Thanks for this implementation! I tried it out and noticed the exploration policy wasn't logging since the update method isn't called. In the TF implementation this line is followed by:

```python              
if hasattr(self.exploration_policy, 'update'):
   self.exploration_policy.update(trainer.step_path)
```

examples/sim_policy.py
maliesa96: It wasnt being used in `rollout` so it had no effect.

src/garage/torch/algos/dqn.py
maliesa96: An example would be if you want to train and evaluate using different reward scales. For example, the `EpisodicLife` wrapper makes it so that episodes end after the agent loses a life, not until all lives are lost. This is fine for training, but if we wanted to report our episodic reward to be the total reward accumulated when **all** lives were lost, we would have to pass a clone of the environment to `eval_env` with the `EpisodicLife` wrapper removed. This would be helpful when comparing an agent's score to that of a human player, which is what deepmind does.

src/garage/torch/algos/dqn.py
maliesa96: Depends on the algorithm and policy. All the DQN implementations evaluate with exploration noise because it actually helps the agent. There are certain edge cases where a deterministic agent can get stuck in an infinite loop if noise isn't added. See the beginning of this clip https://drive.google.com/file/d/1c3BQU5t9gUhWnS-ymzgJFP8R9tZCR_X0/view?usp=sharing

src/garage/torch/algos/dqn.py
maliesa96: Ah nice catch thank you!

src/garage/torch/algos/dqn.py
maliesa96: I needed both `set_gpu` (which is called in the launcher) and setting the default tensor type. Confirmed it works on CPU only devices.

src/garage/torch/algos/dqn.py
maliesa96: Not really, it is just the average reward smoothed over 100 episodes. Its useful when comparing to other libraries because most of them report this metric.

src/garage/torch/algos/dqn.py
maliesa96: Sure, done.

src/garage/torch/algos/dqn.py
maliesa96: Done.

tests/garage/experiment/test_snapshotter.py
maliesa96: No prob, added to the commit msg.

src/garage/torch/algos/dqn.py
yeukfu: Thanks for clarity. Should we also make this change to tf/DQN?

src/garage/torch/algos/dqn.py
```diff
@@ -0,0 +1,268 @@
...
+                self.train_once(trainer.step_itr, trainer.step_path)
+                if (cycle == 0 and self.replay_buffer.n_transitions_stored >=
+                        self._min_buffer_size):
+                    trainer.enable_logging = True
+                    logger.log('Evaluating policy')
+
+                    epsilon_before = self.exploration_policy.epsilon
```
yeukfu, line 141:
```suggestion
                    params_before = self.exploration_policy.get_param_values()
```

src/garage/torch/algos/dqn.py
```diff
@@ -0,0 +1,268 @@
...
+                    eval_eps = obtain_evaluation_episodes(
+                        self.exploration_policy,
+                        self._eval_env,
+                        num_eps=10,
+                        max_episode_length=self._max_episode_length_eval)
+                    self.exploration_policy.epsilon = epsilon_before
```
yeukfu, line 147:
```suggestion
                    self.exploration_policy.set_param_values(params_before)
```

src/garage/np/exploration_policies/epsilon_greedy_policy.py
```diff
@@ -48,6 +48,15 @@ def __init__(self,
         self._total_env_steps = 0
         self._last_total_env_steps = 0
 
+    @property
+    def epsilon(self):
+        """Float: the instantaneous level of exploration noise."""
+        return self._epsilon
+
+    @epsilon.setter
```
yeukfu, line 56: I doubt this can set the epsilon, as the epsilon here is inferred by `self._total_env_steps`. So I recommend to set the total_env_steps or use `get_param_values` and `set_param_values`.

src/garage/torch/algos/dqn.py
maliesa96: Tbh I'd rather keep the loss computation in the optimize function, it makes it much simpler to read. Other algos (VPG comes to mind) have multiple functions that compute various parts of the loss and I find that harder to follow. I agree that the loss should be tested though so I added a test to make sure the computed loss is what is expected.

src/garage/np/exploration_policies/epsilon_greedy_policy.py
maliesa96: Ah you're right, thanks! I'll fix that.

src/garage/torch/algos/dqn.py
maliesa96: Probably yes, but there's a bug in tf DQN right now. I'll add this change to that bug fix PR once its up.

maliesa96:
@ryanjulian @avnishn this is ready for another look!

examples/torch/dqn_atari.py
```diff
@@ -0,0 +1,195 @@
...
+                   clip_gradient=10)
+
+
+@click.command()
+@click.argument('env', type=str)
+@click.option('--seed', default=24)
+@click.option('--n', type=int, default=psutil.cpu_count(logical=False))
+@click.option('--buffer_size', type=int, default=None)
+@click.option('--max_episode_length', type=int, default=None)
+def main(env=None,
```
ryanjulian, line 58: this doesn't seem like canonical usage of wrap_experiment -- there a reason you set it up like this?

@krzentner PTAL

examples/torch/dqn_atari.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+# pylint: disable=unused-argument
+@wrap_experiment(snapshot_mode='gap_overwrite', snapshot_gap=30)
+def dqn_atari(ctxt=None,
+              env=None,
+              seed=24,
+              n_workers=psutil.cpu_count(logical=False),
+              max_episode_length=None,
+              **kwargs):
```
ryanjulian, line 102: please do not use **kwargs in experiment launchers. it circumvents the ideas being wrap_experiment

examples/torch/dqn_atari.py
```diff
@@ -0,0 +1,195 @@
...
+from garage.np.exploration_policies import EpsilonGreedyPolicy
+from garage.replay_buffer import PathBuffer
+from garage.sampler import FragmentWorker, LocalSampler
+from garage.torch import set_gpu_mode
+from garage.torch.algos import DQN
+from garage.torch.policies import DiscreteQFArgmaxPolicy
+from garage.torch.q_functions import DiscreteCNNQFunction
+from garage.trainer import Trainer
+
+hyperparams = dict(n_epochs=500,
```
ryanjulian, line 32: can you just make these the default parameters to your experiment function?

src/garage/torch/algos/dqn.py
```diff
@@ -0,0 +1,286 @@
...
+
+    Args:
+        env_spec (EnvSpec): Environment specification.
+        policy (garage.torch.policies.Policy): Policy. For DQN, this is a
+            policy that performs the action that yields the highest Q value.
+        qf (object): Q-value network.
+        replay_buffer (ReplayBuffer): Replay buffer.
+        steps_per_epoch (int): Number of train_once calls per epoch.
+        n_train_steps (int): Training steps.
+        eval_env (gym.Env): Evaluation environment. If None, a copy of the main
```
ryanjulian, line 34: `Environment` not `gym.Env`

src/garage/torch/algos/dqn.py
```diff
@@ -0,0 +1,286 @@
...
+    method to optimize the policy and Q-function prediction. It uses a
+    supervised method to update the critic network and policy gradient to
+    update the actor network. And there are exploration strategy, replay
+    buffer and target networks involved to stabilize the training process.
+
+    Args:
+        env_spec (EnvSpec): Environment specification.
+        policy (garage.torch.policies.Policy): Policy. For DQN, this is a
+            policy that performs the action that yields the highest Q value.
+        qf (object): Q-value network.
```
ryanjulian, line 30: i think this pretty clearly needs to be an `nn.Module`?

src/garage/torch/algos/dqn.py
```diff
@@ -0,0 +1,286 @@
...
+
+from garage import _Default, log_performance, make_optimizer
+from garage._functions import obtain_evaluation_episodes
+from garage.np.algos import RLAlgorithm
+from garage.sampler import FragmentWorker
+from garage.torch import dict_np_to_torch, global_device
+
+
+class DQN(RLAlgorithm):
+    """A DDPG model implemented with PyTorch.
```
ryanjulian, line 18: please update this documentation

src/garage/torch/algos/dqn.py
```diff
@@ -0,0 +1,286 @@
...
+import torch.nn.functional as F
+
+from garage import _Default, log_performance, make_optimizer
+from garage._functions import obtain_evaluation_episodes
+from garage.np.algos import RLAlgorithm
+from garage.sampler import FragmentWorker
+from garage.torch import dict_np_to_torch, global_device
+
+
+class DQN(RLAlgorithm):
```
ryanjulian, line 17: does this match the hparams used by TF DQN?

src/garage/torch/algos/dqn.py
```diff
@@ -0,0 +1,286 @@
...
+        self.replay_buffer = replay_buffer
+        self.policy = policy
+        self.exploration_policy = exploration_policy
+
+        self._target_qf = copy.deepcopy(self._qf)
+        self._qf_optimizer = make_optimizer(qf_optimizer,
+                                            module=self._qf,
+                                            lr=qf_lr)
+        self._eval_env = eval_env
+        self.worker_cls = FragmentWorker
```
ryanjulian, line 122: please make this a class member if it's not changed by `__init__`

src/garage/torch/algos/dqn.py
```diff
@@ -0,0 +1,286 @@
...
+                                                   eval_eps,
+                                                   discount=self._discount)
+                    self._episode_reward_mean.extend(last_returns)
+                    tabular.record('Evaluation/100EpRewardMean',
+                                   np.mean(self._episode_reward_mean))
+                trainer.step_itr += 1
+
+        return np.mean(last_returns)
+
+    def train_once(self, itr, episodes):
```
ryanjulian, line 170: please make this private

src/garage/torch/algos/dqn.py
```diff
@@ -0,0 +1,286 @@
...
+                self._epoch_ys.append(y)
+                self._epoch_qs.append(q)
+
+        if itr % self._steps_per_epoch == 0:
+            self.log_eval_results(epoch)
+
+        if itr % self._target_update_freq == 0:
+            self._target_qf = copy.deepcopy(self._qf)
+
+    def log_eval_results(self, epoch):
```
ryanjulian, line 201: please make this private

src/garage/torch/algos/dqn.py
```diff
@@ -0,0 +1,286 @@
...
+            tabular.record('QFunction/AverageQ', np.mean(self._epoch_qs))
+            tabular.record('QFunction/MaxQ', np.max(self._epoch_qs))
+            tabular.record('QFunction/AverageAbsQ',
+                           np.mean(np.abs(self._epoch_qs)))
+            tabular.record('QFunction/AverageY', np.mean(self._epoch_ys))
+            tabular.record('QFunction/MaxY', np.max(self._epoch_ys))
+            tabular.record('QFunction/AverageAbsY',
+                           np.mean(np.abs(self._epoch_ys)))
+
+    def optimize_qf(self, samples_data):
```
ryanjulian, line 222: please use `TimeStepBatch` as the input to the loss function.

src/garage/torch/policies/discrete_qf_argmax_policy.py
```diff
@@ -65,4 +65,4 @@ def get_actions(self, observations):
             dict: Empty since this policy does not produce a distribution.
         """
         with torch.no_grad():
-            return self(torch.Tensor(observations)).numpy(), dict()
+            return self(torch.Tensor(observations)).cpu().numpy(), dict()
```
ryanjulian, line 68: how do you i know i'm sampling on CPU?

src/garage/torch/q_functions/discrete_cnn_q_function.py
```diff
@@ -110,3 +111,21 @@ def __init__(self,
                          output_b_init=output_b_init,
                          layer_normalization=layer_normalization,
                          is_image=is_image)
+
+    # pylint: disable=arguments-differ
+    def forward(self, observations):
+        """Return Q-value(s).
+
+        Args:
+            observations (np.ndarray): observations of shape `input_shape`.
```
ryanjulian, line 120: where can the user reference `input_shape`?

src/garage/torch/q_functions/discrete_cnn_q_function.py
```diff
@@ -73,8 +73,8 @@ def __init__(self,
                  strides,
                  minibatch_size,
```
ryanjulian, line 74: this is a property of the optimizer and not the model -- why is this needed?

src/garage/torch/q_functions/discrete_cnn_q_function.py
```diff
@@ -110,3 +111,21 @@ def __init__(self,
...
+    def forward(self, observations):
+        """Return Q-value(s).
+
+        Args:
+            observations (np.ndarray): observations of shape `input_shape`.
+
+        Returns:
+            torch.Tensor: Output value
+        """
+        if observations.shape != self._env_spec.observation_space.shape:
```
ryanjulian, line 125: just make this an error?

examples/torch/dqn_atari.py
krzentner: This is kind of a weird setup, especially with the `hyperparams` dictionary, but it's something I could reasonably see myself writing. I also think having a general dqn atari example like this is valuable (instead of an example for a single atari environment, or many atari environment examples). That kinda necessitates some of this complexity, so it's fine.

tests/garage/torch/algos/test_dqn.py
```diff
@@ -0,0 +1,109 @@
+"""Test DQN performance on cartpole."""
```
ryanjulian, line 1: can you add a test which ensures that you can resume the algorithm from a snapshot?

src/garage/torch/policies/discrete_qf_argmax_policy.py
maliesa96: Well regardless of where the model is the output tensor would need to be moved to the CPU to convert it to a numpy array (the `numpy()` call isn't supported for cuda tensors.

src/garage/torch/q_functions/discrete_cnn_q_function.py
maliesa96: I don't think this should be an error, this just unflattens a flattened batch of observations. CNNs specifically need them to be unflattened (while other primitives don't necessarily) so I think it makes sense to do this here.

examples/torch/dqn_atari.py
maliesa96: It's just easier to run various atari experiments using one launcher. The extra click arguments are just so that tests can set a smaller buffer size and max episode length.

examples/torch/dqn_atari.py
maliesa96: So the idea behind this and the `hyperparams` dict is that it makes life easier when comparing experiments. Passing the hyperparms as a dict to the launcher function logs them into`variant.json` which is helpful later when debugging runs and comparing hyperparams.

src/garage/torch/algos/dqn.py
maliesa96: It does yes, with a couple of extra ones for evaluation.

examples/torch/dqn_atari.py
maliesa96: Please see my other comment, though I'm happy to do this if you prefer.

src/garage/torch/q_functions/discrete_cnn_q_function.py
maliesa96: Oh I initially put that there because `CNNModule` needs to be given the input shape, but it actually just ignores the batch dimension so this turned out to be unnecessary. Fixed.

src/garage/torch/q_functions/discrete_cnn_q_function.py
maliesa96: Hmm you're right, fixed.

tests/garage/torch/algos/test_dqn.py
maliesa96: done.

src/garage/torch/algos/dqn.py
```diff
@@ -0,0 +1,284 @@
...
+            float: The average return in last epoch cycle.
+
+        """
+        if not self._eval_env:
+            self._eval_env = trainer.get_env_copy()
+        last_returns = [float('nan')]
+        trainer.enable_logging = False
+
+        for _ in trainer.step_epochs():
+            for cycle in range(self._steps_per_epoch):
```
DennisCraandijk, line 139: @maliesa96 I have a question. What is the added value of `steps_per_epoch`? The trainer provides epochs to iterate over and nothing extra happens by also iterating over `steps_per_epoch`. So couldn't you just leave it out?

maliesa96: You're right, `steps_per_epoch` separates evaluation and training, the idea being if you want to evaluate less frequently you would increase `steps_per_epoch`. I moved it out of the loop because I agree, it does make more sense and there wasn't a clear reason why it was inside the loop in the first place.

ryanjulian: Is this consistent with how `steps_per_epoch` is defined in other off-policy algos?

maliesa96: it is.

src/garage/torch/algos/dqn.py
ryanjulian: Can you explain more?  I'm not sure I understand why you'd want to disable logging in this use case.

What you describe (passing the env you actually want evaluated to eval_env) is the garage way of achieving what you're describing. The intention is that the user will pass that modified env in the launcher (rather than the algo magically doing it).

src/garage/torch/algos/dqn.py
```diff
@@ -0,0 +1,285 @@
...
+        Args:
+            timesteps (TimeStepBatch): Processed batch data.
+
+        Returns:
+            qval_loss: Loss of Q-value predicted by the Q-network.
+            ys: y_s.
+            qval: Q-value predicted by the Q-network.
+
+        """
+        observations = np_to_torch(timesteps.observations)
```
ryanjulian, line 233: can you file an issue to add TimeStepBatch.torch()?

src/garage/torch/algos/dqn.py
ryanjulian: as long as the important part fits on ~1 screen this is ok with me. if it gets bigger, you should split it up.

src/garage/torch/policies/discrete_qf_argmax_policy.py
ryanjulian: @krzentner can you comment here? not sure what to do in torch

src/garage/torch/q_functions/discrete_cnn_q_function.py
ryanjulian: @krzentner can you weigh in? I think that it's always an error to pass an observation which doesn't match env_spec.observation_space.

src/garage/torch/policies/discrete_qf_argmax_policy.py
krzentner: Yeah, we expect policies to always output numpy arrays. However, instead of using `torch.Tensor`, you should probably use `garage.torch.np_to_torch`.

src/garage/torch/algos/dqn.py
maliesa96: done

src/garage/torch/algos/dqn.py
maliesa96: Ah, I was referring to `eval_env`. I disable logging initially to avoid logging before the warm up phase ends, ie. when the replay buffer contains `min_buffer_size` steps. All other off policy algos in garage do this. Not doing so will produce logs for the initial epochs where no training occurred, and also causes dowel to spit out a long warning about dynamically changing the logging keys. Ex:
```
2020-10-19 15:16:25 | [dqn_atari] Obtaining samples...
2020-10-19 15:16:28 | [dqn_atari] epoch #0 | Training finished
2020-10-19 15:16:50 | [dqn_atari] epoch #0 | Saving snapshot...
2020-10-19 15:16:50 | [dqn_atari] epoch #0 | Saved
2020-10-19 15:16:50 | [dqn_atari] epoch #0 | Time 24.99 s
2020-10-19 15:16:50 | [dqn_atari] epoch #0 | EpochTime 24.99 s
---------------------------  ------------
EpsilonGreedyPolicy/Epsilon      0.980042
TotalEnvSteps                10080
---------------------------  ------------
2020-10-19 15:16:50 | [dqn_atari] epoch #1 | Evaluating policy
2020-10-19 15:17:13 | [dqn_atari] epoch #1 | Training finished
2020-10-19 15:17:45 | [dqn_atari] epoch #1 | Saving snapshot...
2020-10-19 15:17:45 | [dqn_atari] epoch #1 | Saved
2020-10-19 15:17:45 | [dqn_atari] epoch #1 | Time 79.59 s
2020-10-19 15:17:45 | [dqn_atari] epoch #1 | EpochTime 54.60 s
/home/mishari/miniconda3/envs/garage/lib/python3.7/site-packages/dowel/logger.py:207: CsvOutputWarning: Inconsistent TabularInput keys detected. CsvOutput keys: {'EpsilonGreedyPolicy/Epsilon', 'TotalEnvSteps'}. TabularInput keys: {'Evaluation/StdReturn', 'QFunction/AverageY', 'Evaluation/AverageDiscountedReturn', 'Evaluation/MaxReturn', 'QFunction/AverageAbsQ', 'Evaluation/Iteration', 'Evaluation/100EpRewardMean', 'Evaluation/TerminationRate', 'Evaluation/AverageReturn', 'QFunction/MaxQ', 'QFunction/AverageQFunctionLoss', 'Evaluation/MinReturn', 'EpsilonGreedyPolicy/Epsilon', 'Epoch', 'TotalEnvSteps', 'Evaluation/NumEpisodes', 'QFunction/AverageAbsY', 'QFunction/AverageQ', 'QFunction/MaxY'}. Did you change key sets after your first logger.log(TabularInput)?
  output.record(data, prefix=self._prefix_str)
----------------------------------  -------------
Epoch                                   1
EpsilonGreedyPolicy/Epsilon             0.960083
Evaluation/100EpRewardMean            -20.3
Evaluation/AverageDiscountedReturn     -1.56143
Evaluation/AverageReturn              -20.3
Evaluation/Iteration                   20
Evaluation/MaxReturn                  -18
Evaluation/MinReturn                  -21
Evaluation/NumEpisodes                 10
Evaluation/StdReturn                    1.00499
Evaluation/TerminationRate              1
QFunction/AverageAbsQ                   0.869504
QFunction/AverageAbsY                   0.873958
QFunction/AverageQ                      0.866967
QFunction/AverageQFunctionLoss          0.0127423
QFunction/AverageY                      0.871055
QFunction/MaxQ                          1.53793
QFunction/MaxY                          1.86628
TotalEnvSteps                       20160
----------------------------------  -------------
```

src/garage/torch/algos/dqn.py
```diff
@@ -0,0 +1,285 @@
...
+
+                last_returns = log_performance(trainer.step_itr,
+                                               eval_eps,
+                                               discount=self._discount)
+                self._episode_reward_mean.extend(last_returns)
+                tabular.record('Evaluation/100EpRewardMean',
+                               np.mean(self._episode_reward_mean))
+
+            for _ in range(self._steps_per_epoch):
+                trainer.step_path = trainer.obtain_episodes(trainer.step_itr)
```
avnishn, line 161: over here you can pass the number of timesteps that you want to trainer.obtain_episodes. instead of enabling and disabling logging because you're filling the replay buffer, instead get the number of steps that you need in the first training epoch here.

src/garage/torch/algos/dqn.py
avnishn: you can get around this by passing the number of timesteps that you need to fill the replay buffer to obtain episodes.

avnishn (approved):
please merge and then add a golden values test like I mentioned at standup w.r.t the TD3 PR. 

having a golden values test on your loss functions makes it much easier to debug this implementation if it has performance regressions.

src/garage/torch/algos/dqn.py
ryanjulian: this makes sense to me as the solution most consistent with the sampler API. for instance it allows a parallel sampler to farm out the initial sampling efficiently to workers.

src/garage/torch/q_functions/discrete_cnn_q_function.py
krzentner: As I remember it, last time we discussed this we concluded the opposite: that `forward` would always take flattened observations, since it made everything except CNNs simpler. This logic should probably really live in akro eventually, but for the foreseeable future this is a more practical solution.

src/garage/torch/algos/dqn.py
maliesa96: just curious, what would be the benefit of doing this over disabling logging? I imagine it would have to be something like:

```python
if self._min_buffer_size - self.replay_buffer.n_transitions_stored > 0:
      trainer.step_path = trainer.obtain_episodes(trainer.step_itr, self._min_buffer_size - self.replay_buffer.n_transitions_stored)
else:
      trainer.step_path = trainer.obtain_episodes(trainer.step_itr)
```

Seems simpler to just disable and re-enable logging?

krzentner: Conceptually it seems a little simpler to everyone else to just call the sampler once up-front, because disable logging is kinda weird.

maliesa96: Sounds good.

ryanjulian (approved):

PR #2078, opened 2020-09-23
Add deployments to GitHub Actions

ziyiwu9494:
Test deployment action to PYPI. This is using a test PYPI account currently


avnishn (approved):

krzentner (approved):
This looks vaguely correct to me, although it appears it does not work

.github/workflows/ci.yml
```diff
@@ -374,6 +376,31 @@ jobs:
...
+#    if: startsWith(github.ref, 'ref/tags/*')
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Set up Python 3.7
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.7
```
gitanshu, line 392: I am not sure what difference python version makes here since all we're doing is generate an sdist, but since we use 3.6 as min version everywhere, I think we should do that here too

.github/workflows/ci.yml
```diff
@@ -374,6 +380,32 @@ jobs:
           pipenv run pylint --disable=all --enable=import-error garage"
 
 
+  deploy_to_PYPI:
```
ryanjulian, line 383:
```suggestion
  deploy_to_pypiI:
```

.github/workflows/ci.yml
```diff
@@ -13,9 +13,15 @@ on:
   schedule:
     - cron: '0 9 * * *'
   push:
-    branches: [master]
+    branches:
+      - master
+      - 'release-.*'
+    tags:
+      - 'v[1-9][0-9]*.[0-9]+.[0-9]+'
```
ryanjulian, line 20: can you use the release tag regex from TravisCI?

`v([1-9][0-9]*!)?(0|[1-9][0-9]*)(\.([0-9]*))*((a|b|rc)(0|[1-9][0-9]*))?(\.post(0|[1-9][0-9]*))?(\.dev(0|[1-9][0-9]*))?`

Based on my reading of https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet it should be compatible

.github/workflows/ci.yml
```diff
@@ -374,6 +380,32 @@ jobs:
           pipenv run pylint --disable=all --enable=import-error garage"
 
 
+  deploy_to_PYPI:
+    name: Deploy to PYPI
+    runs-on: ubuntu-latest
+    needs: [normal_test, large_test, mujoco_test, mujoco_test_long]
+    if: startsWith(github.ref, 'refs/tags/v')
```
ryanjulian, line 387: Please make sure this *only* matches `refs/tags/` which correspond with the tags regex above!

.github/workflows/ci.yml
```diff
@@ -374,6 +380,32 @@ jobs:
...
+      - name: Set up Python 3.6
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.6
+      - name: Update Version
+        run: echo ${GITHUB_REF##*/} > VERSION
+      - name: Setup Distribution
+        run: >-
+          python setup.py sdist
+      - name: Deploy to PYPI
```
ryanjulian, line 402:
```suggestion
      - name: Deploy to PyPI
```

.github/workflows/ci.yml
ziyiwu9494: unfortunately there's no way to do regex matching an if statement. I'm assuming that if a tag runs, it will be one that matches the trigger, which is a correct regex

setup.py
```diff
@@ -92,7 +92,7 @@
     VERSION = v.read().strip()
 
 setup(
-    name='garage',
+    name='garage_test',
```
ryanjulian, line 95: please correct

ryanjulian (approved):

PR #2081, opened 2020-09-28
Use episode batch in tf/npo

yeukfu:
Partly address https://github.com/rlworkgroup/garage/issues/1110.

### DIFF
--- a/src/garage/tf/algos/npo.py
+++ b/src/garage/tf/algos/npo.py
@@ -7,8 +7,8 @@ from dowel import logger, tabular
 import numpy as np
 import tensorflow as tf
 
-from garage import EpisodeBatch, log_performance, make_optimizer, StepType
-from garage.np import explained_variance_1d, pad_tensor_n
+from garage import log_performance, make_optimizer
+from garage.np import explained_variance_1d
 from garage.np.algos import RLAlgorithm
 from garage.sampler import RaySampler
 from garage.tf import (center_advs,
@@ -17,7 +17,6 @@ from garage.tf import (center_advs,
                        discounted_returns,
                        flatten_inputs,
                        graph_inputs,
-                       paths_to_tensors,
                        positive_advs)
 from garage.tf.optimizers import LBFGSOptimizer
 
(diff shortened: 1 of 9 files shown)


src/garage/tf/algos/npo.py
```diff
@@ -474,53 +442,64 @@ def _build_entropy_term(self, i):
...
+        aug_returns = pad_tensor_n(aug_returns, self.max_episode_length)
 
         # Fit baseline
         logger.log('Fitting baseline...')
         self._baseline.fit(paths)
+        return aug_returns
 
-    def _policy_opt_input_values(self, samples_data):
-        """Map episode samples to the policy optimizer inputs.
+    def _process_episodes_batch(self, episodes):
```
nicolengsy, line 479: Rename to `_process_episode_batch` to keep naming consistent

nicolengsy (approved):

src/garage/tf/algos/npo.py
```diff
@@ -474,53 +442,64 @@ def _build_entropy_term(self, i):
...
+                baselines.
 
         """
+        paths = episodes.to_list()
+        for i, _ in enumerate(paths):
+            paths[i]['actions'] = self._env_spec.action_space.flatten_n(
+                paths[i]['actions'])
+
+        baselines = [self._baseline.predict(path) for path in paths]
+        samples_data = paths_to_tensors(paths, self.max_episode_length,
```
ryanjulian, line 499: is there any way we can pass EpisodeBatch + a few other tensors to TensorFlow? A good detector that #1110 is done is "we can delete all usages of `paths_to_tensors`"

yeukfu: I think if padding the tensor is necessary, we must split the EpisodeBatch.

yeukfu: Or we can do the padding operation in TensorFlow.

src/garage/tf/algos/npo.py
```diff
@@ -177,76 +177,45 @@ def train(self, trainer):
         last_return = None
 
         for _ in trainer.step_epochs():
-            trainer.step_path = trainer.obtain_samples(trainer.step_itr)
-            last_return = self._train_once(trainer.step_itr, trainer.step_path)
+            episodes = trainer.obtain_episodes(trainer.step_itr)
```
krzentner, line 180: You should still save them to `trainer.step_path`, even though they're episodes. This is needed for `save_paths` to work.

krzentner (commented):
I think this change is an improvement. Honestly, I'm not sure how much of this refactoring we want to do for TF algorithms.

src/garage/tf/algos/npo.py
yeukfu: @ryanjulian Can we do the padding operation in EpisodeBatch. For example, `episode_batch.padded_list()`. I have tried to do all things in tensorflow, but it is difficult as the padding size is a variable.

krzentner (approved):
Nice! Looks good to me.

PR #2082, opened 2020-09-30
Change env_infos from float to nparray

### LINKED ISSUE #1215: Shape of env_info of HalfCheetah* environments not compatible with TrajectoryBatch
A trajectory of `HalfCheetahVel` and `HalfCheetahDir` should have an `env_info` dict of numpy array but they do not. The `task` in their `env_info` is instead a dict. 

`env_info` is assumed by `TrajectoryBatch` to be a dict of numpy array. 

This is a bug that will break an algorithm when it uses `HalfCheetah` together with `TrajectoryBatch`.

nicolengsy:
To close #1215, making sure that the infos returned by HalfCheetahDirEnv / HalfCheetahVelEnv's `step()` function is a dict of np arrays instead of a dict of floats.

### DIFF
--- a/src/garage/envs/mujoco/half_cheetah_dir_env.py
+++ b/src/garage/envs/mujoco/half_cheetah_dir_env.py
@@ -76,10 +76,10 @@ class HalfCheetahDirEnv(HalfCheetahEnvMetaBase):
             task_name = 'backward'
         else:
             raise ValueError('task direction should be 1. or -1.')
-        infos = dict(reward_forward=forward_reward,
-                     reward_ctrl=-ctrl_cost,
-                     task_dir=self._task['direction'],
-                     task_name=task_name)
+        infos = dict(reward_forward=np.asarray([forward_reward]),
+                     reward_ctrl=np.asarray([-ctrl_cost]),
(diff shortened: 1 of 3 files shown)


yeukfu (approved):
Should we add a test for this?

krzentner (approved):

yeukfu (approved):
Looks good to me!

src/garage/envs/mujoco/half_cheetah_dir_env.py
```diff
@@ -76,10 +76,10 @@ def step(self, action):
...
         else:
             raise ValueError('task direction should be 1. or -1.')
-        infos = dict(reward_forward=forward_reward,
-                     reward_ctrl=-ctrl_cost,
-                     task_dir=self._task['direction'],
-                     task_name=task_name)
+        infos = dict(reward_forward=np.asarray([forward_reward]),
+                     reward_ctrl=np.asarray([-ctrl_cost]),
+                     task_dir=np.asarray([self._task['direction']]),
+                     task_name=np.asarray([task_name]))
```
krzentner, line 82: BTW, you can't make an `np.ndarray` of the `task_name`. That will cause things to crash later.

PR #2083, opened 2020-09-30
Check in GymEnv for env_info consistency

### LINKED ISSUE #1841: Check that env_info is consistent in Environment
Currently, if an environment outputs inconsistent env_info's, it results in a bizarre error (for example, #1797 ). GarageEnv should check that the underlying environment always outputs the same env_infos with the same shapes, and produce an error early if it does not.

nicolengsy:
To close #1841. Checks in GymEnv's `step()` that the underlying environment outputs the same env_info keys and types of values.


src/garage/envs/gym_env.py
```diff
@@ -212,6 +216,17 @@ def step(self, action):
...
+        # check that env_infos are consistent
+        if not self._env_info:
+            self._env_info = {k: type(info[k]) for k in info}
+        elif self._env_info.keys() != info.keys():
+            raise RuntimeError('GymEnv outputs inconsistent env_info keys.')
+        else:
+            info_types = {k: type(info[k]) for k in info}
+            if self._env_info != info_types:
+                raise RuntimeError(
+                    'GymEnv outputs inconsistent env_info shapes.')
```
krzentner, line 228: This should say "inconsistent env_info types", I think. I'm not actually sure if we should have this check at all though, since outputting `None` some epochs might be an okay thing to do.

krzentner (commented):
This looks pretty good, but please write a test to make sure those `!=` signs do what you intend. In particular, does this work if there are multiple env infos that are in different key orders?

src/garage/envs/gym_env.py
```diff
@@ -205,13 +209,22 @@ def step(self, action):
...
+                env_info keys.
 
         """
         if self._step_cnt is None:
             raise RuntimeError('reset() must be called before step()!')
 
         observation, reward, done, info = self._env.step(action)
 
+        # check that env_infos are consistent
+        print(info)
```
yeukfu, line 222: Should remove this?

tests/garage/envs/test_gym_env.py
```diff
@@ -139,3 +139,13 @@ def test_done_resets_step_cnt():
...
+
+def test_inconsistent_env_infos():
+    env = GymEnv('MountainCar-v0')
+    env.reset()
+    env._env_info = {'k1': 'v1', 'k2': 'v2'}
+    with pytest.raises(RuntimeError,
+                    match='GymEnv outputs inconsistent env_info keys.'):
+        env.step(env.action_space.sample())
+    # check that order of keys don't matter for equality
+    assert env._env_info.keys() == {'k2': 'v2', 'k1': 'v1'}.keys()
```
yeukfu, line 151: No new line at the end of the file. You can install the pre-commit check locally. For example, `conda activate garage` `pre-commit uninstall` `pre-commit install -t pre-commit` `pre-commit install -t pre-push` `pre-commit install -t commit-msg`

yeukfu (approved):

krzentner (commented):
Please address Ruofu's comments, but besides what he mentioned this looks good to go for me.

tests/garage/envs/test_gym_env.py
nicolengsy: Thanks!

maliesa96 (approved):

krzentner (approved):

src/garage/envs/gym_env.py
```diff
@@ -205,13 +209,21 @@ def step(self, action):
...
+            RuntimeError: if underlying environment outputs inconsistent
+                env_info keys.
 
         """
         if self._step_cnt is None:
             raise RuntimeError('reset() must be called before step()!')
 
         observation, reward, done, info = self._env.step(action)
 
+        # check that env_infos are consistent
```
krzentner, line 221: Unfortunately, HalfCheetah is broken, so this change actually needs to go *after* the `TimeLimit.truncated` logic below. That should let you merge this change.

nicolengsy:
Note: I had to push with `--no-verify` to bypass a metaworld ImportError that I haven't been able to resolve on my computer, but it was the only error preventing the push.

PR #2085, opened 2020-09-30
Make snapshot_config optional in benchmarks

gitanshu:
PR #2072 added an argument to functions decorated with @benchmark
to pass snapshot_config to auto benchmarks but that broke other
benchmarks that didn't need that argument.

This commit cleans it up by passing snapshot_config directly to
iterate_experiments, since trying to pass it from within decorator
function makes for a confusing API


ryanjulian (approved):

zequnyu (approved):

gitanshu:
@mergifyio rebase

PR #2087, opened 2020-10-01
Refactor away garage.tf.plotter

### LINKED ISSUE #1502: Refactor away garage.tf.plotter
This can instead be handled by conditional behavior (or just a unification) in garage.plotter

nicolengsy:
To close #1502. Moved garage.tf.plotter into garage.plotter.

### DIFF
--- a/src/garage/plotter/plotter.py
+++ b/src/garage/plotter/plotter.py
@@ -1,6 +1,7 @@
 """Visualizes a policy running in an environment, in parallel with training."""
 import atexit
 from collections import namedtuple
+import contextlib
 from enum import Enum
 from multiprocessing import JoinableQueue, Process
 import platform
@@ -8,9 +9,19 @@ from threading import Thread
 
 import numpy as np
 
-from garage import rollout
+from garage import rollout as garage_rollout
 
 __all__ = ['Plotter']
+
+
+@contextlib.contextmanager
+def context_mgr():
+    """Context Manager for non-tf plotter.
+
+    Yields:
+        NoneType: None
+    """
+    yield None
 
 
 class Op(Enum):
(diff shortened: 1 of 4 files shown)


src/garage/plotter/plotter.py
```diff
@@ -5,13 +5,19 @@
 from multiprocessing import JoinableQueue, Process
 import platform
 from threading import Thread
+import contextlib
+
+import tensorflow as tf
```
krzentner, line 10: We can't assume that tf is actually installed when running this file, so we should only run this import once `tf_plot` has been set to True (this requires ignoring pylint).

src/garage/plotter/plotter.py
```diff
@@ -134,6 +147,18 @@ def init_plot(self, env, policy):
         """
         if not Plotter.enable:
             return
+
+        self._env = env
+        self._tf_plot = tf_plot
+        self._rollout = rollout
+        if tf_plot:
```
krzentner, line 154: You should be able to put the tensorflow import here (writing it out to a global variable in the process).

src/garage/trainer.py
```diff
@@ -807,11 +807,11 @@ def _start_worker(self):
         self._sampler.start_worker()
         if self._plot:
             # pylint: disable=import-outside-toplevel
-            from garage.tf.plotter import Plotter
-            self._plotter = Plotter(self.get_env_copy(),
+            from garage.plotter import Plotter
+            self._plotter = Plotter()
+            self._plotter.init_plot(self.get_env_copy(),
```
krzentner, line 812: I think you need to set `tf_plot=True` here?

src/garage/plotter/plotter.py
ryanjulian: If TensorFlow is installed, there's virtually no harm in creating a CPU-bound session, even if the user is running PyTorch or NumPy in actuality. Why not eliminate this flag and just create a session if TF is installed?

krzentner: Creating a truly CPU bound session without starting a new process doesn't appear to be possible in TF1. Personally, I'd rather we stop opening TensorFlow sessions in PyTorch experiments if it's easy for us to do so.

PR #2089, opened 2020-10-03
Check observation shape conforms to observation_space

### LINKED ISSUE #1842: Check that the environment produces observations of a shape consistent with its observation space before giving them to a policy during sampling
It's common for environments to produce observations that aren't *technically* in their observation space (i.e. they're outside of the box, but have the right number of dimensions). However, some rare environments, and many environments when they're being developed, produce observations that aren't even of the same shape.

Currently, if an environment produces such an observation, garage crashes in a bizarre location, typically when feeding the observation into a tensor inside of the policy's calling `get_action` method on the first rollout. This is really confusing, since the environment code is nowhere near this code. If GarageEnv produced a sensible message when an environment mis-behaves in this way, tracking down this type of failure would be much easier.

nicolengsy:
To close #1842.


src/garage/envs/gym_env.py
```diff
@@ -242,6 +244,9 @@ def step(self, action):
         if step_type in (StepType.TERMINAL, StepType.TIMEOUT):
             self._step_cnt = None
 
+        if self.spec.observation_space.flat_dim != np.prod(observation.shape):
```
yeukfu, line 247: Just curious why comparing the flat_dim instead of the original shape.

yeukfu (approved):

src/garage/envs/gym_env.py
krzentner: It's pretty common for environments to give `observation_space` as a finite box, then exceed that box. Having said that, this change should probably instead do a check more like the one in `EpisodeBatch`: first check if `self.spec.observation_space.contains(observation)`, and then only check the flat dimensions if that's false (this matters for discrete observations).

krzentner (approved):

krzentner:
The uncovered code in this change is essentially one line and pretty inline with other uncovered code in garage, so I'm merging this.

PR #2097, opened 2020-10-13
Add dueling to torch discrete CNN

### LINKED ISSUE #1490: torch/dueling (D)DQN

maliesa96:
closes #1490

### DIFF
--- a/src/garage/torch/modules/__init__.py
+++ b/src/garage/torch/modules/__init__.py
@@ -12,12 +12,15 @@ from garage.torch.modules.mlp_module import MLPModule
 from garage.torch.modules.multi_headed_mlp_module import MultiHeadedMLPModule
 # DiscreteCNNModule must go after MLPModule
 from garage.torch.modules.discrete_cnn_module import DiscreteCNNModule
+from garage.torch.modules.discrete_dueling_cnn_module import (
+    DiscreteDuelingCNNModule)
 # yapf: enable
 
 __all__ = [
     'CategoricalCNNModule',
     'CNNModule',
     'DiscreteCNNModule',
+    'DiscreteDuelingCNNModule',
     'MLPModule',
     'MultiHeadedMLPModule',
     'GaussianMLPModule',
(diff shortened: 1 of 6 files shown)


src/garage/torch/modules/discrete_cnn_module.py
```diff
@@ -31,6 +31,8 @@ class DiscreteCNNModule(nn.Module):
         hidden_sizes (list[int]): Output dimension of dense layer(s) for
             the MLP for mean. For example, (32, 32) means the MLP consists
             of two hidden layers, each with 32 hidden units.
+        dueling (bool): Whether to use a dueling architecture for the
```
ryanjulian, line 34: can you make this a separate module/QFunction class instead?

ryanjulian:
ETA for benchmarks?

src/garage/torch/modules/discrete_cnn_module.py
maliesa96: I can. It seems a bit overkill to create a separate module though since there will be lots of overlapping code.

maliesa96:
About a week.

maliesa96:
Benchmarks: 
dueling

src/garage/torch/modules/discrete_cnn_module.py
ryanjulian: the parts of this module not hidden behind "if dueling" are basically all boilerplate. It's an antipattern to implement 2 classes in 1 and then switch between their implementations using a bool. 

In Python, we don't do `my_list = Container(1,2,3, list=True)` and `my_tuple = Container(1,2,3, list=False)`, we do `my_list = list(1,2,3)` and `my_tuple = tuple(1,2,3)`.

It's also more honest to the user about what dueling DQN is, which is basically an alternative QF architecture. If you want to do dueling DQN, you construct a dueling Q-function and pass it to the regular-old DQN algorithm.

@krzentner WDYT?

ryanjulian (commented):
LGTM once comments addresses

src/garage/torch/modules/discrete_cnn_module.py
maliesa96: Fair enough, I'll get that done.

haydenshively (approved):
👍🏻

maliesa96:
@ryanjulian this is ready for a final look!

ryanjulian (approved):

PR #2123, opened 2020-10-14
Refactor REPS to use episode batch

yeukfu:
Partially address https://github.com/rlworkgroup/garage/issues/1110. 
@ryanjulian @krzentner It seems that the implementation of REPS doesn't use baseline predictions, and I don't see that as input of tensorflow. Should we remove the bassline from its arguments?

### DIFF
--- a/src/garage/tf/algos/reps.py
+++ b/src/garage/tf/algos/reps.py
@@ -7,18 +7,13 @@ import numpy as np
 import scipy.optimize
 import tensorflow as tf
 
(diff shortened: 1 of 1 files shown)


ryanjulian (approved):

krzentner (approved):

yeukfu:
@Mergifyio rebase

yeukfu:
@Mergifyio rebase

Issue #2120, opened 2020-10-13
Add "must have" features to make non-RL applications practical to write


krzentner:
Hey, so people (including myself) are starting to use garage more for IL (and related topics). I have a short list of "must have" features which I think would make that _much_ better. When I proposed them seperately in the past people have seemed a little hesitant on supporting them, so I figured putting them here in one place with a unified goal might make them more understandable.
 1. Add a `Trainable` class, which `RLAlgorithm` inherits from, and make `Trainer` use that interface (this is already 90% done, the `Trainer` already only *requires* that interface, but it's confusing and un-obvious).
 2. Allow setting the default X-axis with `wrap_experiment`, instead of locking it to `TotalEnvSteps` (pretty necessary for BC, I have a commit for it).
 3. Allow setting `total_env_steps` on `Trainer`, and make `garage.Sampler` have a `total_env_steps` property, which the Sampler is responsible for counting (I already have a commit for this). This allows much more flexibility to the algorithm about what sampling is for training, and what is for evaluation, etc.
 4. Add a `Sampler` class for sampling from a `PathBuffer`. This removes the complicated "source" concept that BC currently has (I also already have a commit for this).

ryanjulian:
These all look great to me and I look forward to reviewing the PRs! 

If you don't think they will all land around the same time, feel free to split this into 4 or more separate issues for tracking purposes.

PR #2125, opened 2020-10-15
Record total_env_steps in samplers

krzentner:
This change makes it practical to use a Sampler that the Trainer isn't aware of.
See issue #2120

### DIFF
--- a/src/garage/sampler/local_sampler.py
+++ b/src/garage/sampler/local_sampler.py
@@ -40,6 +40,7 @@ class LocalSampler(Sampler):
         for worker, agent, env in zip(self._workers, self._agents, self._envs):
             worker.update_agent(agent)
             worker.update_env(env)
+        self.total_env_steps = 0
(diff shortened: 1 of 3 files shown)


src/garage/sampler/local_sampler.py
```diff
@@ -118,6 +119,9 @@ def obtain_samples(self, itr, num_samples, agent_update, env_update=None):
                 batches.append(batch)
                 if completed_samples >= num_samples:
                     return EpisodeBatch.concatenate(*batches)
```
yeukfu, line 121:
```suggestion
```

yeukfu (approved):

ryanjulian (approved):

src/garage/sampler/local_sampler.py
krzentner: Thanks! Good catch.

PR #2127, opened 2020-10-15
Make plotting plot the learner in BC

krzentner:
Before, using BC with an expert source policy caused the expert to be shown in the plotter, which was very confusing. This change fixes that, by making the expert the algorithm's `.exploration_policy` and the learner its `.policy`


ryanjulian (approved):

krzentner:
@Mergifyio rebase

irisliucy (approved):

krzentner:
@mergifyio rebase

PR #2128, opened 2020-10-15
Allow setting x-axis in wrap_experiment

krzentner:
This change allows changing the dowel scalar key that is the default X-axis in tensorboard.
This is essential for easily seeing what's happening in algorithms that don't step through environments, like BC.
See Issue #2120


zequnyu (approved):

ryanjulian (approved):

krzentner:
@Mergifyio rebase

PR #2129, opened 2020-10-15
Make TanhNormal.log_prob(1.0) != -inf

krzentner:
TanhNormal uses an `epsilon` value to handle numerical instabilities. In very rare circumstances (which I've encountered about once after running about 10^8 steps), an action can get sampled at the very edge of the action space (i.e. one dimension is exactly 1.0 or -1.0).
This change makes it so that such actions are only considered exceptionally unlikely, instead of making the log probability negative infinity.

This also means that actions taken by an expert, then clipped into the action space always have a valid log probability.
I believe this formulation changes the rest of the tanh normal distribution as little as is practical. Essentially, when computing the pre-tanh value, instead of using the inverse of tanh, it uses the inverse of an equivalent shaped sigmoid from -1 - epsilon to 1 + epsilon (where epsilon=1e-6 by default).

### DIFF
--- a/src/garage/torch/distributions/tanh_normal.py
+++ b/src/garage/torch/distributions/tanh_normal.py
@@ -49,7 +49,8 @@ class TanhNormal(torch.distributions.Distribution):
         """
         # pylint: disable=arguments-differ
         if pre_tanh_value is None:
-            pre_tanh_value = torch.log((1 + value) / (1 - value)) / 2
+            pre_tanh_value = torch.log(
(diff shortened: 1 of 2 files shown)


krzentner:
The log likelihood of these boundary values under this formulation is about -13.5 (per dimension), which is consistent with what I observed empirically.

avnishn (approved):

krzentner:
@Mergifyio rebase

ryanjulian (approved):

krzentner:
@Mergifyio rebase

PR #2130, opened 2020-10-15
Decrease BC eval steps

krzentner:
Behavioral cloning doesn't normally need to sample, making it practical to run a very large number of iterations. However, right now we still use 100 evaluation episodes in BC, which makes it much slower to run. This decreases the number of evaluations to 10.


ryanjulian (approved):

krzentner:
@Mergifyio rebase

PR #2132, opened 2020-10-15
Fix #2131

### LINKED ISSUE #2131: CI Is failing on `env.visualize()` in TestNormalizedGym
The error involves `pyglet` decoding an image from Gym's `pendulum-v0`:

```
AttributeError: 'Image' object has no attribute 'tostring'
```

avnishn:
There's something specifically wrong with an image used
in rendering pendulum-v0, causing visualize to fail.

I swapped it out with CartPole-v1 for the purposes of
the tests and that was a fix.

### DIFF
--- a/tests/garage/envs/test_normalized_gym.py
+++ b/tests/garage/envs/test_normalized_gym.py
@@ -4,7 +4,7 @@ from garage.envs import GymEnv, normalize
 class TestNormalizedGym:
 
     def setup_method(self):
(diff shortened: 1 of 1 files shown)


maliesa96:
I'm still curious as to why it would suddenly stop working with pendulum but I'm okay with this being the fix.

maliesa96 (approved):

krzentner (approved):
Sure

PR #2137, opened 2020-10-19
Refactor te_npo to use episode batch

yeukfu:
https://github.com/rlworkgroup/garage/issues/1110.

### DIFF
--- a/src/garage/tf/algos/te_npo.py
+++ b/src/garage/tf/algos/te_npo.py
@@ -245,39 +245,41 @@ class TENPO(RLAlgorithm):
         baselines = episodes.pad_to_last(np.concatenate(baselines))
 
         # Process trajectories
-        trajs, traj_infos = self._process_trajectories(episodes)
+        embed_eps, embed_ep_infos = self._process_episodes(episodes)
 
         average_return = np.mean(undiscounted_returns)
 
         logger.log('Optimizing policy...')
-        self._optimize_policy(itr, episodes, baselines, trajs, traj_infos)
+        self._optimize_policy(itr, episodes, baselines, embed_eps,
+                              embed_ep_infos)
 
         return average_return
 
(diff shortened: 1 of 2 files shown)


src/garage/tf/algos/te_npo.py
```diff
@@ -210,118 +210,102 @@ def train(self, trainer):
...
-                * baselines: (numpy.ndarray)
-                * returns: (numpy.ndarray)
-                * valids: (numpy.ndarray)
-                * agent_infos: (dict)
-                * letent_infos: (dict)
-                * env_infos: (dict)
-                * trjectory_infos: (dict)
-                * paths: (list[dict])
+            np.ndarray: Embedding trajectories.
+            dict: Embedding distribution information.
```
krzentner, line 300: Can you put the keys here?

src/garage/tf/algos/te_npo.py
```diff
@@ -210,118 +210,102 @@ def train(self, trainer):
...
-        self._baseline.fit(paths)
+        self._baseline.fit(fit_paths)
 
         self._old_policy.parameters = self.policy.parameters
         self._old_policy.encoder.model.parameters = (
             self.policy.encoder.model.parameters)
         self._old_inference.model.parameters = self._inference.model.parameters
 
-    def _paths_to_tensors(self, paths):
+    def _process_trajectories(self, episodes):
```
krzentner, line 291: Is there a reason this method isn't named `_process_episodes`?

krzentner (approved):
This looks about right to me. I don't know *that* much about TE-NPO, so I think Ryan might have more to say though. There's still a bunch of places which say `trajectories` instead of `episodes`. Feel free to leave any existing names the same, but please use `episodes` (and `eps` instead of `traj`) for new variables.

haydenshively (approved):

ryanjulian (approved):

PR #2138, opened 2020-10-19
Refactor RL2 to use EpisodeBatch

yeukfu:
https://github.com/rlworkgroup/garage/issues/1110.

### DIFF
--- a/src/garage/tf/algos/rl2.py
+++ b/src/garage/tf/algos/rl2.py
@@ -10,12 +10,8 @@ import akro
 from dowel import logger
 import numpy as np
 
-from garage import (EnvSpec,
-                    EnvStep,
-                    EpisodeBatch,
-                    log_multitask_performance,
-                    StepType,
-                    Wrapper)
+from garage import (EnvSpec, EnvStep, EpisodeBatch, log_multitask_performance,
+                    StepType, Wrapper)
 from garage.np.algos import MetaRLAlgorithm
 from garage.sampler import DefaultWorker
 from garage.tf.algos._rl2npo import RL2NPO
--- a/src/garage/tf/algos/rl2.py
+++ b/src/garage/tf/algos/rl2.py
@@ -10,9 +10,12 @@ import akro
 from dowel import logger
 import numpy as np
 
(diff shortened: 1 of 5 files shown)


krzentner (approved):

src/garage/tf/algos/rl2.py
```diff
@@ -483,23 +480,30 @@ def _concatenate_paths(self, paths):
...
         """
-        observations = np.concatenate([path['observations'] for path in paths])
+        env_infos = {
+            k: np.concatenate([b.env_infos[k] for b in episode_list])
+            for k in episode_list[0].env_infos.keys()
+        }
+        agent_infos = {
+            k: np.concatenate([b.agent_infos[k] for b in episode_list])
+            for k in episode_list[0].agent_infos.keys()
+        }
```
krzentner, line 490: I think there might be a utility in tensor utils for concatenating all arrays in a dictionary. If not, it would probably be a good idea to create one.

yeukfu: I find a function called concat_tensor_dict_list. But it seems not suitable to use here. I think the code here is not generous as it is a modification from EpisodeBatch.concatenate, so I will probably just leave it there.

src/garage/tf/algos/rl2.py
```diff
@@ -450,26 +451,22 @@ def _process_samples(self, itr, paths):
...
-        episodes = EpisodeBatch.from_list(self._env_spec, concatenated_paths)
 
-        return episodes, average_return
+        return concatenated_episodes, average_return
 
-    def _concatenate_paths(self, paths):
+    def _concatenate_paths(self, episode_list):
         """Concatenate paths.
 
         The input paths are from different episodes but same task/environment.
```
maliesa96, line 463:
```suggestion
    def _concatenate_episodes(self, episode_list):
        """Concatenate episodes.

        The input list contains samples from different episodes but same task/environment.
```

maliesa96 (approved):

maliesa96:
LGTM, but please see my suggestion. If we're moving away from using `paths` then its best to just replace that term with episodes where applicable to remain consistent.

PR #2140, opened 2020-10-20
Remove include_dashboard arg so that doctests pass

haydenshively:
I also updated docs/requirements.txt so that it works with pip's new dependency resolver, which is enabled on readthedocs (with no apparent way to disable it)

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,3 +1,6 @@
+sphinx-autoapi
+sphinxcontrib-bibtex
+
 akro
 click
 cloudpickle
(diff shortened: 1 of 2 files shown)


avnishn (approved):

yeukfu (approved):

docs/requirements.txt
```diff
@@ -13,11 +17,3 @@ scipy
...
 tensorflow
 tensorflow-probability
-
-# dev dependencies
-matplotlib
-recommonmark
-sphinx
-sphinx-autoapi>=1.4.0
-sphinx_rtd_theme
-sphinxcontrib-bibtex
```
haydenshively, line 23: These deleted dependencies are installed by readthedocs implicitly and for now this is the only way I can get pip to install everything correctly

PR #2143, opened 2020-10-21
Make mujoco test run a few minutes faster

krzentner:
This test suite has run out of time, and it's blocking merging algorithms. This should buy us about six more minutes of time to run new tests.


gitanshu (approved):

ryanjulian (approved):

ryanjulian:
A simple solution here is to run these tests in GH Actions only, since it's only Travis which has the very short limit. We have seen good test stability from GH Actions so I think it's safe to do this for every test job except deployments @ziyiwu9494

krzentner:
The only one I actually disabled here was the PEARL test. If someone can make it run faster or add a tag to make it run in GH Actions that would be great, but this is blocking other changes from being merged, so I'm going to merge it as-is.

Issue #941, opened 2019-10-17
Move input flattening to primitives


ahtsan:
Currently it is done in algorithms, but it makes more sense for primitives to do it.
Reference: https://github.com/rlworkgroup/garage/blob/master/src/garage/tf/algos/batch_polopt.py#L145

irisliucy:
The link is not working. Currently only `npo`, `reps`, `te_npo` in tf is using `flatten_inputs` function, I will go ahead and move this function to be their primitives.

ahtsan:
> @ahtsan The link is not working. Currently only `npo`, `reps`, `te_npo` in tf is using `flatten_inputs` function.
> 
> Is it necessary to make it primitives? Since it seems like a general function to be put in `/tf/_functions.py` because that's what currently happening in pytorch.

I was referring to the flattening of observations (see https://github.com/rlworkgroup/garage/blob/v2019.10.0/src/garage/tf/algos/batch_polopt.py#L145).

Since now different observation space/action space will be responsible for the flattening (e.g. https://github.com/rlworkgroup/garage/blob/a9ae9989261da6aa85e452ace0c9915fe7c90d22/src/garage/tf/algos/npo.py#L293), I think this issue is safe to be closed.

ryanjulian:
Wait, so do all of the primitives now accept unflattened inputs?

ahtsan:
@ryanjulian Thanks for the catch, I thought `observation_space.to_tf_placeholder()` would take care of the flatten but I was wrong.

Basically, primitives should always flatten observations by calling `self.observation_space.flatten / flatten_n(observations)` in `get_action() / get_actions()` except for akro.Image.

avnishn:
An offline discussion led to the conclusion that until we've defined the workflow of garage, it's best if we flatten observations in the sampler and then unflatten them as necessary. I'm not sure if this issue follows with that approach, so I'm going to close it for now.

avnishn:
@yeukfu Note: In MLP modules, don't unflatten `Akro.Images` please flatten them if they aren't flattend.

krzentner:
Yeah, the logic that reads:

```python
if not isinstance(observation, np.ndarray) and not isinstance(
    observation, torch.Tensor):
    observation = self._env_spec.observation_space.flatten(observation)
```

Should probably read:
```python
if not isinstance(observation, np.ndarray) or len(observation.shape) > 1:
    observation = self._env_spec.observation_space.flatten(observation)
```
[...]

PR #2153, opened 2020-10-26
Flatten input in primitives

yeukfu:
This should close https://github.com/rlworkgroup/garage/issues/941.


krzentner (approved):
Thanks for taking care of this! These changes look like a strict improvement to me from how it was before. Having said that, it would be nice at some point to add the logic here to akro instead, so we don't have to keep copy and pasting it over and over again. That can wait for after the release though.

avnishn (approved):
LGTM

PR #2155, opened 2020-10-28
Allow installing only TF or torch version

gitanshu:


ryanjulian (commented):
Do the tests pass under these configurations? Maybe at least run a pylint on both configs (it should detect importerrors)

ryanjulian (approved):

PR #2157, opened 2020-10-28
Blacklist KellyCoinflip from nightly build GymEnv tests

adibellathur:
Fix for nightly build issue: https://github.com/rlworkgroup/garage/runs/1319831555?check_suite_focus=true

KellyCoinflipGeneralized env has tuple observation instead of np.array: https://github.com/openai/gym/blob/master/gym/envs/toy_text/kellycoinflip.py

### DIFF
--- a/tests/garage/tf/envs/test_gym_base.py
+++ b/tests/garage/tf/envs/test_gym_base.py
@@ -19,6 +19,9 @@ class TestGymEnv:
         if spec._env_name.startswith('CarRacing'):
             pytest.skip(
                 'CarRacing-* envs bundled in atari-py 0.2.x don\'t load')
+        if spec._env_name.startswith('KellyCoinflip'):
+            pytest.skip(
+                'KellyCoinflip env has tuple observation, not np.array')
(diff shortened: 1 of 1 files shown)


tests/garage/tf/envs/test_gym_base.py
```diff
@@ -19,6 +19,9 @@ def test_all_gym_envs(self, spec):
         if spec._env_name.startswith('CarRacing'):
             pytest.skip(
                 'CarRacing-* envs bundled in atari-py 0.2.x don\'t load')
+        if spec._env_name.startswith('KellyCoinflip'):
+            pytest.skip(
+                'KellyCoinflip env has tuple observation, not np.array')
```
krzentner, line 24: This explanation isn't *quite* right. The problem isn't that it has tuple observations exactly, it's that those observations aren't contained in the `observation_space`, and also aren't a `Box` (which we have special logic for when it exceeds the `observation_space`).

krzentner (approved):

avnishn (approved):

PR #2159, opened 2020-10-28
Add PER

### LINKED ISSUE #1489: Implement PER

maliesa96:
Closes #1489.

### DIFF
--- a/src/garage/replay_buffer/per_replay_buffer.py
+++ b/src/garage/replay_buffer/per_replay_buffer.py
@@ -1,0 +1,145 @@
+"""Prioritized Experience Replay."""
+
+import numpy as np
+
+from garage import StepType, TimeStepBatch
+from garage.replay_buffer.path_buffer import PathBuffer
+
+
+class PERReplayBuffer(PathBuffer):
+    """Replay buffer for PER (Prioritized Experience Replay).
+
+    PER assigns priorities to transitions in the buffer. Typically
+    these priority of each transition is proportional to the corresponding
+    loss computed at each update step. The priorities are then used to create
+    a probability distribution when sampling such that higher priority
+    transitions are sampled more frequently. For more see
+    https://arxiv.org/abs/1511.05952.
+
+    Args:
+        capacity_in_transitions (int): total size of transitions in the buffer.
+        env_spec (EnvSpec): Environment specification.
+        total_timesteps (int): Total timesteps the experiment will run for.
+            This is used to calculate the beta parameter when sampling.
+        alpha (float): hyperparameter that controls the degree of
+            prioritization. Typically between [0, 1], where 0 corresponds to
+            no prioritization (uniform sampling).
+        beta_init (float): Initial value of beta exponent in importance
+            sampling. Beta is linearly annealed from beta_init to 1
+            over total_timesteps.
+    """
+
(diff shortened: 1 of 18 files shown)


maliesa96:
Benchmarks should be done in a few days.

src/garage/torch/algos/dqn.py
```diff
@@ -188,10 +192,18 @@ def _train_once(self, itr, episodes):
         for _ in range(self._n_train_steps):
             if (self.replay_buffer.n_transitions_stored >=
                     self._min_buffer_size):
-                timesteps = self.replay_buffer.sample_timesteps(
-                    self._buffer_batch_size)
-                qf_loss, y, q = tuple(v.cpu().numpy()
-                                      for v in self._optimize_qf(timesteps))
+                if self._prioritized_replay:
```
ryanjulian, line 195: can you instead make the signature of `ReplayBuffer.sample_timesteps` always return `timesteps, weights, indices`? this way DQN doesn't need to know about the RB implementation

ryanjulian (commented):
How are the benchmark results?

src/garage/torch/algos/dqn.py
maliesa96: The algo needs to be aware of the type of replace buffer used regardless because the priorities of the samples need to be updated via the `update_priorities()`call  if PER is being used. I suppose we could add that to the `ReplayBuffer` class as well and have it not do anything if PER is not being used but that seems like a poor design decision as priorities, weights, and indices are relevant only when using PER. 

Also the algo infers whether PER is being used or not via `isinstance(replay_buffer, PERReplayBuffer)` so the user doesn't have to pass in any extra arguments.

src/garage/replay_buffer/per_replay_buffer.py
```diff
@@ -0,0 +1,145 @@
...
+        path_len = len(path['observations'])
+        self._curr_timestep += path_len
+
+        # find the indices where the path will be stored
+        first_seg, second_seg = self._next_path_segments(path_len)
+
+        # set priorities for new timesteps = max(self._priorities)
+        # or 1 if buffer is empty
+        max_priority = self._priorities.max() or 1.
+        self._priorities[list(first_seg)] = max_priority
```
krzentner, line 142: Can you do `self._priorities[first_seg.start:first_seg.stop]` instead? That's the idiom that's used in the rest of `PathBuffer`.

src/garage/torch/algos/dqn.py
krzentner: Yeah, even if there needs to be some specialized logic, it's still better if the types are consistent. It should be fairly easy to fix the other algorithms.

krzentner (approved):
Nice. This is a pretty clean implementation.

src/garage/torch/algos/dqn.py
maliesa96: Sure, done!

src/garage/replay_buffer/per_replay_buffer.py
maliesa96: fixed.

maliesa96:
Currently rerunning some of the benchmarks (some were interrupted), will report back once they're done.

PR #2166, opened 2020-11-01
Install examples with garage

gitanshu:


ryanjulian (commented):
This LGTM -- can't you also run them like `python -m garage.examples.tf.dqn_pong`?

I think the detailed documentation you addd should probably go in RTFD rather than the README

Also, didn't you forget to move the files? :)

gitanshu:
> This LGTM -- can't you also run them like `python -m garage.examples.tf.dqn_pong`?

Oh right, that works. I hadn't thought of that. I'll add it to docs.

> I think the detailed documentation you addd should probably go in RTFD rather than the README

Yeah. `Quick Start with garage` would be a good place for this I think?

> Also, didn't you forget to move the files? :)

Oh I chose not to move the files and instead created a symlink. I think having examples in the root directory is good for visibility, gives somebody an instant impression that we have examples available.

ryanjulian:
There are issues with symlinks in git repos (don't work on Windows, for example). I think it's best just to move them.

gitanshu:
This only matters while packaging though. For somebody who is working with the source, it shouldn't matter. Do you still think I should move the files?

ryanjulian:
It matters if someone (who is just a user who prefers source) can't download and use the repo on Windows. I think we should just move them.

gitanshu:
@mergifyio rebase

gitanshu:
@mergifyio rebase

gitanshu:
@mergifyio rebase

avnishn (approved):
I need to update some of the Metaworld examples specifically, because they're not entirely set up the way that they should be, but I guess since we're moving to monthly releases I can always slip in on the next release.

@krzentner what do you think?

haydenshively (approved):

ryanjulian (approved):

gitanshu:
@mergifyio refresh

Issue #2154, opened 2020-10-27
Reshape inputs in tf/cnn.py


yeukfu:
Unflatten inputs in tf/cnn.py and let cnn related q_functions accept flattened input.

PR #2168, opened 2020-11-03
Reshape input in tf/cnn

yeukfu:
This should close https://github.com/rlworkgroup/garage/issues/2154.

In tf/cnn.py, unflatten the input if it is flattened.


yeukfu:
@krzentner I have a concern that if we add env_spec into the constructor of tf/cnn_model, it will be not consistent to the torch's module.

krzentner (commented):
I think that's fine. There's already other minor differences between TF and Torch in the model code.

yeukfu:
> I think that's fine. There's already other minor differences between TF and Torch in the model code.

I pass the input_dim to the constructor of cnn models, which is more consistent the current apis I think. Because some constructors have output_dim as an argument.

krzentner (approved):
Nice!

PR #2176, opened 2020-11-15
Improve data types

krzentner:
This change converts all dtypes to use dataclass, and makes EpisodeBatch
a subclass of TimeStepBatch.

This required slightly changing some shapes, to make the types
completely consistent.

I also merged and re-wrote the checking code.
The error messages now have a more consistent pattern of "\<thing\> has
\<property\> but must have \<correct property\> to match \<ground truth\>".

All core dtypes have (including EnvSpec) have now also been made immutable.

### DIFF
--- a/src/garage/_dtypes.py
+++ b/src/garage/_dtypes.py
@@ -1,12 +1,13 @@
 """Data types for agent-based learning."""
-import collections
+from dataclasses import dataclass
 import enum
+from typing import Dict, List
+import warnings
 
-import akro
 import numpy as np
 
-from garage.np import (concat_tensor_dict_list, slice_nested_dict,
-                       stack_tensor_dict_list)
+from garage.np import (concat_tensor_dict_list, pad_batch_array,
+                       slice_nested_dict, stack_tensor_dict_list)
 
 # pylint: disable=too-many-lines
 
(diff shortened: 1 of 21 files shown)


krzentner:
I had to reorder a lot of things in `_dtypes.py`, which makes the total diff much more confusing to read. I've split this change into two commits, the first of which just re-orders `_dtypes.py`. Please use the diff on the second change and the tests to review.

krzentner:
Github makes it somewhat hard to access the per-commit diff: [it's located here](https://github.com/rlworkgroup/garage/commit/89c71d489adf2aa19c855ea18b711a805d9e7c12#diff-5ec4a0d73f35dc5654c36cb4b1b8bbdb6e0517e8dc1c370609c77d15b24db34e)

src/garage/_dtypes.py
```diff
@@ -286,973 +263,822 @@ def concatenate(cls, *batches):
...
-                {k: np.asarray([v[i]])
-                 for (k, v) in self.env_infos.items()},
-                'agent_infos':
-                {k: np.asarray([v[i]])
-                 for (k, v) in self.agent_infos.items()},
-                'step_types':
-                np.asarray([self.step_types[i]])
-            })
-        return samples
+        return pad_batch_array(self.next_observations, self.lengths,
```
yeukfu, line 944: Should be self.step_types.

src/garage/np/_functions.py
```diff
@@ -368,3 +370,34 @@ def slice_nested_dict(dict_or_array, start, stop):
...
+        max_length (int): Defaults to max(lengths) if not provided.
+
+    Returns:
+        np.ndarray: Of shape :math:`(N, max_length, X^*)`
+
+    """
+    assert array.shape[0] == sum(lengths)
+    if max_length is not None and max_length > max(lengths):
+        # This is probably not a good idea, but RL2 already depends on it.
+        warnings.warn('Creating a padded array with longer length than '
```
yeukfu, line 391: I think this `max_length > max(lengths)` should be the normal situation. Don't understand why warning here?

yeukfu (approved):
Nice work!

src/garage/np/_functions.py
krzentner: Good catch. This logic is backwards. I'll add a test.

src/garage/_dtypes.py
krzentner: Good catch.

ryanjulian (approved):

Issue #1756, opened 2020-07-10
Construct sampler in the launcher, not the LocalRunner


ryanjulian:
This is more consistent with garage's API, and would probably make it more clear that the Sampler is its own configurable components.

Everything else in garage is created and configured in the launcher, why not the sampler?

This would also make the symmetry between a Sampler vs a ReplayBuffer vs a Dataset more clear. If, for instance, we are going to do something like `RLAlgorithm.train(Dataset: d)` then you could imagine `d = replay_buffer.to_dataset(batch_size=100)` or `d = sampler.to_dataset(batch_size=100)`:

```python
from garage.replay_buffer import HERReplayBuffer
from garage.sampler import RaySampler, VecWorker, WorkerFactory
from garage.tf.algos import DDPG
# etc...

ray_sampler = RaySampler.from_worker_factory(WorkerFactory(VecWorker))
her_replay_buffer = HERReplayBuffer(capacity_in_transitions=int(1e6))

ddpg = DDPG(
    # ...
    sampler=ray_sampler,
    replay_buffer=her_replay_buffer,
)
```

If you used a dataset-style interface, it could look like thijs
```python
from garage.replay_buffer import HERReplayBuffer
from garage.sampler import RaySampler, VecWorker, WorkerFactory
from garage.tf.algos import DDPG
# etc...

ray_sampler = RaySampler.from_worker_factory(WorkerFactory(VecWorker))
replay_buffer = HERReplayBuffer(capacity_in_transitions=int(1e6))
rb_dataset = replay_buffer.to_dataset(batch_size=1000)

ddpg = DDPG(
    sampler=ray_sampler,
    dataset=rb_dataset,
)
```

Further, if you allowed samplers (or workers) to fill replay buffers, it could make the dataflow even more clear.
If you used a dataset-style interface, it could look like this
```python
from garage.replay_buffer import HERReplayBuffer, SamplerPusher
from garage.sampler import RaySampler, VecWorker, WorkerFactory
from garage.tf.algos import DDPG
# etc...

replay_buffer = HERReplayBuffer(capacity_in_transitions=int(1e6))
ray_sampler = RaySampler.from_worker_factory(WorkerFactory(VecWorker), post_processor=replay_buffer.put)
rb_dataset = replay_buffer.to_dataset(batch_size=1000)

ddpg = DDPG(
    dataset=rb_dataset,
)
```

PR #2182, opened 2020-11-22
Construct sampler in the launcher

yeukfu:
This PR moves the construction of sampler into the launcher instead of the trainer. And algorithms will have a `sampler` field. 

This should close https://github.com/rlworkgroup/garage/issues/1756.

### DIFF
--- a/docs/user/experiments.rst
+++ b/docs/user/experiments.rst
@@ -35,6 +35,7 @@ simple one, :code:`examples/tf/trpo_cartpole.py`, is also pasted below:
   from garage.envs import GymEnv
   from garage.experiment.deterministic import set_seed
   from garage.np.baselines import LinearFeatureBaseline
+  from garage.sampler import RaySampler
   from garage.tf.algos import TRPO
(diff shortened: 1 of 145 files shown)


krzentner:
Wow, this is a huge change! Good work. The only thing I would change is that we should really make it easy to use the sampler constructors instead of needing the user to construct a `WorkerFactory`:

For example:
```python
sampler = LocalSampler(policy, env, n_workers=10, it_tf_worker=True)
```

Given the large size of this change, I would be fine with merging it before changing that.

yeukfu:
> Wow, this is a huge change! Good work. The only thing I would change is that we should really make it easy to use the sampler constructors instead of needing the user to construct a `WorkerFactory`:
> 
> For example:
> 
> ```python
> sampler = LocalSampler(policy, env, n_workers=10, it_tf_worker=True)
> ```
> 
> Given the large size of this change, I would be fine with merging it before changing that.

That's a good idea.

ryanjulian (approved):

krzentner (approved):

PR #2185, opened 2020-11-26
[WiP] Deterministic on- and off-policy sampling

MkuuWaUjinga:
Extend the `Environment` API to support setting environment library specific seeds.

Tasks:
- [ ] Extend interface
- [ ] Set seeds for Gym envs
- [ ] Set seeds for dm_control envs
- [ ] Set seeds for grid_world envs
- [ ] Set seeds for point envs
- [ ] Set seeds for metaworld envs

Open Questions:
- How to ensure determinism of off-policy algorithms?

### DIFF
--- a/src/garage/_environment.py
+++ b/src/garage/_environment.py
@@ -159,6 +159,8 @@ class Environment(abc.ABC):
     +-----------------------+
     | visualize()           |
     +-----------------------+
+    | seed()                |
+    +-----------------------+
     | close()               |
     +-----------------------+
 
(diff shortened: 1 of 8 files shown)


src/garage/_environment.py
```diff
@@ -452,6 +464,13 @@ def visualize(self):
         """Creates a visualization of the wrapped environment."""
         self._env.visualize()
 
+    def seed(self, seed):
```
ryanjulian, line 467: shouldn't this have an implementation?

src/garage/envs/dm_control/dm_control_env.py
```diff
@@ -184,6 +184,13 @@ def visualize(self):
             self._viewer = DmControlViewer(title=title)
             self._viewer.launch(self._env)
 
+    def seed(self, seed):
+        """Sets all environment seeds.
+
+        Args:
+            seed (int): The seed value to set
```
ryanjulian, line 191: please include an implementation

MkuuWaUjinga:
Thanks for the pointers. Addressed everything in the latest commits. I assume GridWorld and PointEnv don't have any seeds at all? 
Furthermore, with the implementation right now every worker has the _same_ environment seed. This means that each worker always samples the same trajectory given a fixed action sequence. I think this is something we need to fix before merging?

PR #2188, opened 2020-12-01
Resolve conflicts between mujoco-py and gym temporarily

yeukfu:
Pip release [20.3](https://github.com/pypa/pip/releases) today, which makes the CI build failed. Here is the [log](https://github.com/rlworkgroup/garage/runs/1476748450?check_suite_focus=true), and the related [issue](https://github.com/pypa/pip/issues/9011).

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -97,7 +97,7 @@ RUN python -m virtualenv $VIRTUAL_ENV
 ENV PATH "$VIRTUAL_ENV/bin:$PATH"
 
 # Prevent pip from complaining about available upgrades inside virtualenv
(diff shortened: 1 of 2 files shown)


ryanjulian:
Unfortunately, the only way to solve this is to fix the build under the new pip -- we have no control over which version of pip people use, and when in doubt we should use the latest.

krzentner (approved):
Can you add a comment to setup.py explaining what's going on here? As is this change is pretty unclear.

yeukfu:
> Can you add a comment to setup.py explaining what's going on here? As is this change is pretty unclear.

Sure

ryanjulian (approved):

PR #2189, opened 2020-12-01
torch.CNNModule rework

krzentner:
This change reworks garage.torch.CNNModule, to make it much more like other torch modules.
In particular, it no longer takes an "input_var," as though it were a tensorflow module.
It also delete CNNCategoricalModule, since it's unneeded, and fixes a major bug in CNNCategoricalPolicy.
It also deletes the TransposeImage environment wrapper, since all CNNs can now handle images in NCHW and NHWC format.

This change does not delete CNNDiscreteModule, but that should be deleted soon as well.


ryanjulian (approved):

irisliucy (approved):

krzentner:
@Mergifyio rebase

krzentner:
This change fixes #2231. It should be ready to merge now. Thanks Iris for fixing several of the pre-commit issues.

Issue #2190, opened 2020-12-03
TRPO Init Type Error


ManavR123:
I am creating a child class that inherits from TRPO, but upon initializing the optimizer, I get 

`TypeError: Failed to convert object of type <class 'list'> to Tensor`

Here is the overall stacktrace:

```python
File "/content/gdrive/My Drive/cs_285_project/deeptutor/policies/LoggedTRPO.py", line 8, in __init__
    super(LoggedTRPO, self).__init__(*args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/garage/tf/algos/trpo.py", line 106, in __init__
    name=name)
File "/usr/local/lib/python3.6/dist-packages/garage/tf/algos/npo.py", line 151, in __init__
    self._init_opt()
  File "/usr/local/lib/python3.6/dist-packages/garage/tf/algos/npo.py", line 158, in _init_opt
    pol_loss, pol_kl = self._build_policy_loss(pol_loss_inputs)
  File "/usr/local/lib/python3.6/dist-packages/garage/tf/algos/npo.py", line 360, in _build_policy_loss
    policy_entropy = self._build_entropy_term(i)
  File "/usr/local/lib/python3.6/dist-packages/garage/tf/algos/npo.py", line 470, in _build_entropy_term
    [-1, self.max_episode_length])
  File "/usr/local/lib/python3.6/dist-packages/tensorflow/python/util/dispatch.py", line 201, in wrapper
    return target(*args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/tensorflow/python/ops/array_ops.py", line 195, in reshape
    result = gen_array_ops.reshape(tensor, shape, name)
  File "/usr/local/lib/python3.6/dist-packages/tensorflow/python/ops/gen_array_ops.py", line 8234, in reshape
    "Reshape", tensor=tensor, shape=shape, name=name)
  File "/usr/local/lib/python3.6/dist-packages/tensorflow/python/framework/op_def_library.py", line 473, in _apply_op_helper
    raise err
  File "/usr/local/lib/python3.6/dist-packages/tensorflow/python/framework/op_def_library.py", line 470, in _apply_op_helper
    preferred_dtype=default_dtype)
  File "/usr/local/lib/python3.6/dist-packages/tensorflow/python/framework/ops.py", line 1499, in convert_to_tensor
    ret = conversion_func(value, dtype=dtype, name=name, as_ref=as_ref)
[...]

PR #2193, opened 2020-12-04
Check If max_episode_length is None is NPO

ManavR123:
Address #2190

I simply add a check in NPO init to see if max_episode_length is None and if it is, I raise a ValueError.

Discussed with @ryanjulian

### DIFF
--- a/src/garage/tf/algos/npo.py
+++ b/src/garage/tf/algos/npo.py
@@ -102,6 +102,8 @@ class NPO(RLAlgorithm):
         self.policy = policy
         self._scope = scope
         self.max_episode_length = env_spec.max_episode_length
+        if self.max_episode_length == None:
+            raise ValueError("max_episode_length must not be None")
         self._env_spec = env_spec
         self._baseline = baseline
         self._discount = discount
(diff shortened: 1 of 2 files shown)


src/garage/tf/algos/npo.py
```diff
@@ -102,6 +102,8 @@ def __init__(self,
         self.policy = policy
         self._scope = scope
         self.max_episode_length = env_spec.max_episode_length
+        if self.max_episode_length == None:
```
ryanjulian, line 105: please place this validation above the long list of property initializations (i.e. first lines of this method), and put a newline after the if statement

src/garage/tf/algos/npo.py
```diff
@@ -102,6 +102,8 @@ def __init__(self,
...
+            raise ValueError("max_episode_length must not be None")
```
ryanjulian, line 106: we use single quotes in this repo. we have an automated tool called pre-commit which will check your PR for small issues like this and tell you how to fix them (or you can check the CI results below, which will also tell you what to fix).

See the guide here: https://garage.readthedocs.io/en/latest/user/preparing_a_pr.html and https://garage.readthedocs.io/en/latest/user/git_workflow.html#passing-the-pre-commit-hooks

krzentner (approved):

ryanjulian (approved):

PR #2194, opened 2020-12-07
Update docs and benchmarks to fit with #2182

yeukfu:
Update docs and benchmarks to fit with the new way of constructing the sampler. See https://github.com/rlworkgroup/garage/pull/2182.

### DIFF
--- a/docs/user/implement_algo.md
+++ b/docs/user/implement_algo.md
@@ -138,20 +138,18 @@ reinforcement learning.
 In the above section, we set up an algorithm, but never actually explored the
 environment at all, as can be seen by `TotalEnvSteps` always being zero.
 
-In order to collect samples from the environment, we can set the `sampler_cls`
-and `policy` fields on our algorithm, and call `trainer.obtain_samples()`. We'll
-also need to seed the random number generators used for the experiment.
+In order to collect samples from the environment, we need to construct a
+`sampler` and set it a field in our algorithm. Then we can call
+`trainer.obtain_samples()` to get samples. We'll also need to seed the random
+number generators used for the experiment.
 
 ```py
-from garage.samplers import LocalSampler
-
 class SimpleVPG:
 
-    sampler_cls = LocalSampler
-
-    def __init__(self, env_spec, policy):
+    def __init__(self, env_spec, policy, sampler):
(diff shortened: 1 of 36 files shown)


docs/user/implement_algo.md
```diff
@@ -198,11 +198,10 @@ from garage.np import discount_cumsum
 
 class SimpleVPG:
 
-    sampler_cls = LocalSampler
-
-    def __init__(self, env_spec, policy):
+    def __init__(self, env_spec, policy, sampler):
         self.env_spec = env_spec
         self.policy = policy
+        self.sampler = sampler
```
krzentner, line 204: Can you make all of the fields here have leading `_`? It demonstrates that they're not part of the policy API.

docs/user/sampling.md
```diff
@@ -40,34 +40,76 @@ on a single CPU. `VecWorker` can compute a batch of actions from a policy
...
+In the above example, we firstly construct a worker factory, which will
+construct 4 workers for the sampler. And the max length of episodes collected
+by these workers will be 100. Noted that for policies with TensorFLow
+framework, we need to set `is_tf_worker` to be `True`. Here we don't choose
+a type of worker explicitly, so it will construct `DefaultWorker` by default.
+
+With the worker factory, then we construct a `LocalSampler` with policies and
+environments that will be used in sampling.
+
+### From parameters
```
krzentner, line 73: Can you actually swap the order of these two sections? It's now much more common to construct a sampler directly.

krzentner (approved):
This looks great. I'm really glad we're getting this API change done completely. I have a few small comments on how the docs could be a little better. Please feel free to merge them after you've addressed them (and gotten a review from someone else).

docs/user/implement_algo.md
yeukfu: But we need to access algo.sampler in the trainer. Would it be appropriate to make this private? And before this change the `sample_cls` field also did't have leading `_`.

ryanjulian (approved):

docs/user/implement_algo.md
ryanjulian: It should generally be a read-only property.

PR #2198, opened 2020-12-09
Make the sampler field protected in algorithms

yeukfu:
This is a patch PR to https://github.com/rlworkgroup/garage/pull/2194.

### DIFF
--- a/src/garage/trainer.py
+++ b/src/garage/trainer.py
@@ -157,8 +157,10 @@ class Trainer:
         self._env = env
 
         self._seed = get_seed()
-        # pylint: disable=protected-access
-        self._sampler = self._algo._sampler
+
+        if hasattr(self._algo, '_sampler'):
+            # pylint: disable=protected-access
+            self._sampler = self._algo._sampler
 
         self._has_setup = True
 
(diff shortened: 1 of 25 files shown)


ryanjulian (approved):

yeukfu:
@krzentner One problem of constructing the default sampler in the algorithm is that the algorithm doesn't have the env instances, which is required to construct samplers. So the default sampler can only be constructed in the trainer, which is our only pattern.

krzentner:
Honestly, we should probably be passing the environment into the algorithm anyways. As a shorter term solution, we could construct the sampler in `train` by calling `trainer.get_env_copy()`.

src/garage/trainer.py
```diff
@@ -145,19 +145,19 @@ def setup(self, algo, env):
...
         """
         self._algo = algo
         self._env = env
 
         self._seed = get_seed()
-        self._sampler = self._algo.sampler
+
+        if hasattr(self._algo, '_sampler'):
+            # pylint: disable=protected-access
+            self._sampler = self._algo._sampler
```
krzentner, line 160: Oh, I see. The idea here was always for algorithms to call `self._sampler` directly, instead of going through the trainer, but clearly that needs more work. This will do for now, I suppose.

krzentner (approved):

PR #2201, opened 2020-12-14
Backport deployment using build matrix

ziyiwu9494:
Solve #2053

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -23,10 +23,10 @@ jobs:
     strategy:
       matrix:
         branch: [master, release-2020.06]
-    env:
-      DOCKER_TAG: ${{ DOCKER_TAG_BASE }}-${{ matrix.branch }}
 
     steps:
+      - name: Create Docker Tag
+        run: DOCKER_TAG=${{ DOCKER_TAG_BASE }}-${{ matrix.branch }}
       - name: Docker info
         run: docker version
       - uses: actions/checkout@v2
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -10,8 +10,11 @@ env:
   MJKEY: ${{ secrets.MJKEY }}
 
 on:
-  schedule:
-    - cron: '0 9 * * *'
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -22,6 +20,7 @@ on:
   pull_request:
     branches:
       - master
```
ryanjulian, line 22: we still get PRs to release branches, so the normal workflow still needs to run

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,405 @@
...
+  GARAGE_GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+  TRAVIS_COMMIT_RANGE: origin/${{ github.base_ref }}...origin/${{ github.head_ref }}
+  DOCKER_TAG_BASE: garage-ci-${{ github.run_id }}
+  OWNER: "rlworkgroup"
+  DOCKER_CACHE_REPO: "garage-ci"
+  MJKEY: ${{ secrets.MJKEY }}
+
+on:
+#  schedule:
```
ryanjulian, line 13: remove?

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,405 @@
...
+#    - cron: '0 9 * * *'
+  pull_request:
```
ryanjulian, line 15: if this is a nightly, why run on PRs?

.github/workflows/ci.yml
ryanjulian: if the two are incompatible, perhaps you also need a separate workflow for release-2020.06 PRs?

.github/workflows/nightly.yml
ziyiwu9494: I'm just using this trigger to test

.github/workflows/ci.yml
ziyiwu9494: This should still run for PRs and deployments

PR #2207, opened 2020-12-20
Split CI for release 2020-06

ziyiwu9494:
Test alternate fix for #2053

### DIFF
--- a/.github/workflows/ci-release-2020.06.yml
+++ b/.github/workflows/ci-release-2020.06.yml
@@ -1,4 +1,4 @@
-name: Garage CI
+name: Garage CI Release-2020.06
 
 env:
   GARAGE_GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -15,10 +15,12 @@ on:
   push:
     branches:
       - master
-      - 'release-.*'
     tags:
       - 'v[1-9][0-9]*.[0-9]+.[0-9]+'
       - 'v[1-9][0-9]*.[0-9]+.[0-9]+rc[0-9]'
+# Make sure to ignore all current release tags
```
ryanjulian, line 21: might as well ignore *all* release tags instead

ryanjulian (approved):
See my 1 comment. LGTM if this is working!

.github/workflows/ci.yml
ziyiwu9494: Do we never deploy any tags from master? I don't have a regex for all possible release tags.

ryanjulian: in practice i think all release tags will match the ones you have above. we might deploy from master one day soon (for nightly builds), but we can implement that later. just get this one working first.

PR #2221, opened 2021-01-23
Add workflow file in release branch

gitanshu:
GitHub requires the workflow file to be present in the target branch for it to trigger events on that branch


gitanshu:
Since the workflow file isn't already present in the release-2020.06 branch, the CI won't work for this on this PR, and hence the PR needs to be force pushed.

ryanjulian (approved):

PR #2229, opened 2021-02-06
Add command for running CI on PRs from forks

gitanshu:
Secret variable are not accessible to the CI if the
pull request repo head is a fork, so that a random
fork can't access the secrets by logging them in
the CI output. `pull_request_target` event has similar
security issues.

This commit adds a workflow that allows somebody with
write access to the repo to trigger the CI workflow with
access to secrets, against the PR, by commenting `/ok-to-test`
on that PR, after reviewing the code to make sure it is safe.

By default, it runs the workflow against the merge commit of
the head of the PR and the target branch. Newer commits have to
be tested by issuing the slash command again. It is also
possible to test a specific commit using
`/ok-to-test sha=<commit_sha>`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,7 @@ env:
   DOCKER_CACHE_REPO: "garage-ci"
   MJKEY: ${{ secrets.MJKEY }}
   CI_USER: rlworkgroupbot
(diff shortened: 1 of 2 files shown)


ryanjulian (approved):
Great work!

haydenshively (approved):

PR #2232, opened 2021-02-11
Add basic CI Release Documentation

ziyiwu9494:
Readme and template for documenting how to create a new CI file for releases. Right now the process must be done manually.


.github/workflows/README.md
```diff
@@ -0,0 +1,28 @@
+# CI
```
ryanjulian, line 1: Can you put these docs into ReadTheDocs (in docs/) instead?

.github/workflows/ci-release-TEMPLATE.yml
```diff
@@ -0,0 +1,410 @@
+name: Garage CI <Release Version>
```
ryanjulian, line 1: I don't think this is the best place for this file, since it will probably get picked up by GH actions. Is this just a copy (mostly) of the existing release files? In that case, not sure we need it.

ziyiwu9494: Hi! I'm reopening this because I don't think we'll be changing the Github Actions workflows any further for some time. The template is closer to a copy of the Release CI but with updates (doctest, etc.). I feel like it would just make the process more foolproof than it would be if there was no file at all that could be made into a new release ci file with the obvious edits.

PR #2234, opened 2021-02-12
Test remove checkout action

ziyiwu9494:
We think that the checkout action isn't actually needed in the CI so we're testing that


.github/workflows/ci-release-2020.06.yml
```diff
@@ -44,45 +43,12 @@ jobs:
...
-          -e MJKEY \
-          --memory 6500m \
-          --memory-swap 6500m \
-          "${DOCKER_TAG}" scripts/travisci/check_precommit.sh
-
-
   normal_test:
     name: Normal Tests
     runs-on: ubuntu-latest
     needs: check_pre_commit
```
gitanshu, line 49: Since we are removing check_pre_commit, I guess we need to remove this too

gitanshu (approved):

.github/workflows/ci-release-2020.06.yml
```diff
@@ -3,19 +3,18 @@ name: Garage CI Release-2020.06
 env:
   GARAGE_GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
   CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
-  TRAVIS_COMMIT_RANGE: origin/${{ github.base_ref }}...origin/${{ github.head_ref }}
+  CI_COMMIT_RANGE: origin/${{ github.base_ref }}...origin/${{ github.head_ref }}
```
gitanshu, line 6: I think we should call it PR_COMMIT_RANGE because this value doesn't make sense for any other triggers, and also is empty probably

ryanjulian (approved):

ryanjulian (approved):

PR #2235, opened 2021-02-15
Fix saving episodes

krzentner:
When we renamed `trainer.step_episodes` to `trainer.step_path`, another related refactor still used `step_path`, resulting in most algorithms failing to save episodes. This change fixes those algorithms.


krzentner:
Please note that I have not tested this change yet.

ryanjulian (approved):

haydenshively (approved):

PR #2240, opened 2021-02-20
Test running ci in containers

ziyiwu9494:
Test if we can simplify the CI a great deal by using the github actions container option

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,26 +97,20 @@ jobs:
     name: Check pre-commit
     runs-on: ubuntu-latest
     needs: build_docker_container
+    container:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -97,26 +97,20 @@ jobs:
     name: Check pre-commit
     runs-on: ubuntu-latest
     needs: build_docker_container
+    container:
+      image: docker.pkg.github.com/${OWNER}/${DOCKER_CACHE_REPO}/${DOCKER_TAG}
+      credentials:
+        username: ${{ CI_USER }}
```
gitanshu, line 103: this probably won't work either, but even if it did, it would have to be 
```suggestion
        username: ${{ env.CI_USER }}
```

PR #2247, opened 2021-03-05
Update max resource usage docs for new EnvSpec

jimzers:
Changed `env.spec.max_episode_length` to `env.spec.max_episode_steps`. This deals with changes to OpenAI Gym's EnvSpec for `gym` version `0.17.2`.


jimzers:
Ignore... Turns out this change only applies to OpenAI Gym's `EnvSpec`. Looked into `.unwrapped.spec.max_episode_steps` but could not figure things out.

Upon further inspection it seems that there is problem with the sampler... Will open an issue. Sorry!

PR #2249, opened 2021-03-08
Fixing #2248

### LINKED ISSUE #2248: test_tanh_normal_expand is failing with torch 1.8
https://github.com/rlworkgroup/garage/runs/2045693049?check_suite_focus=true#step:5:203

irisliucy:
Fix #2248 by disabling `validate_args`.


avnishn:
@irisliucy can you explain why validating the arguments to the constructor solves our issue?

I'm referencing this article:
https://praveen-palanisamy.github.io/blog/2018/04/22/pytorch-distributions-validate-args

but I'm not sure how correct it is or how this would fix our tests.
Also for some reason we fail pre-commit on your changes, and this probably is because some new functions were added to the torch.distribution interface, that are not being implemented.

irisliucy:
@avnishn There was an issue (https://github.com/rlworkgroup/garage/runs/2045693049?check_suite_focus=true#step:5:255) failing github action earlier and @gitanshu also filed a similar issue for torch 1.8. It failed because of `validate_args` argument in `torch.distribution`.

ryanjulian (approved):

avnishn:
@irisliucy is there a way in which we can go to the tests, and then change the offending the parameters. 

Unless there is a good reason to turn off validation of parameters, I think that we shouldn't go in this direction.

ryanjulian:
According to https://pytorch.org/docs/stable/distributions.html validate_args can be slow which is why it can be turned off in the first place, so this isn't clear-cut.

My big question is, why were our args failing in the first place?

PR #2255, opened 2021-03-17
Master updates for Release-2020.03

ziyiwu9494:


gitanshu (approved):

haydenshively (approved):
Structure looks good on this, but I'd rather keep the CI running on 2020.06 for another month or so. It'd be nice to support current version + previous version at all times. Can you copy these files/lines rather than renaming/replacing them?

PR #2259, opened 2021-03-23
garage.torch.CNNModule backport to 2021.03

krzentner:
It now takes an env spec and automatically computes the output size.
It also handles NHWC and NCHW instead of requiring an environment
wrapper.

This change also fixes several issues with the existing usage of CNNs
in pytorch.

This PR is a backport to 2021.03 (which I believe has not been pushed to PYPI yet).


haydenshively (approved):
Looks good. I wish our version of Python had enum support though, so we could have enums for NCHW rather than strings.

krzentner:
> 
> 
> Looks good. I wish our version of Python had enum support though, so we could have enums for NCHW rather than strings.

Ah, we actually had a discussion about this about two years ago. We used enums for some things for a while, but eventually found they weren't flexible enough to be worth it. In particular, we often encountered scenarios where we wanted "sub-enums" (enums with a subset of the options of the original enum, or, alternatively, super-enums with new entries added), but python enums don't have that capability. They also turned out to be less ergonomic in python than using strings, so we decided to just use strings to keep the API consistent.

PR #2260, opened 2021-03-31
Preserve sac log alpha when moving between CPU and GPU

ziyiwu9494:


src/garage/torch/algos/sac.py
```diff
@@ -516,6 +516,17 @@ def to(self, device=None):
             device (str): ID of GPU or CPU.
 
         """
+
+        def optimizer_to(optim, device):
+            for param in optim.state.values():
```
krzentner, line 521: It would probably be better to have a seperate function (in `garage.torch._functions`) that recursively moves any nesting of dictionaries, lists, and tensors.

krzentner (commented):
This looks about right to me. I'm not sure the test is completely correct though, since `sac._log_alpha` should have changed with the second call to `train`. You might need to increase the alpha learning rate for the test to work. Remember that the test should fail before your fix.

src/garage/torch/algos/sac.py
ziyiwu9494: Can optimizer state dicts be infinitely nested? I haven't really be able to find any documentation on it. Regardless, I feel like having a function that just moves optimizers might be easier to understand?

avnishn: Probably not optimizers themselves. 

Perhaps you could add this to our torch optimizer class, wherever we keep the special MAML optimizer.

ziyiwu9494: interestingly enough, the test would fail on the previous code anyways because it would crash due to things partially not being moved. I'm confused as to why changing the learning rate would have made it fail on the previous code if it ran, though.

Issue #2079, opened 2020-09-24
When pickling torch algorithms, move torch modules to CPU before pickling


avnishn:
Our current behavior is to pickle policies and value/q-functions so that they are pickled in their exact state, whether they are on gpu or not. This is bad because it forces users to use work-arounds to open torch pkled experiments on  cpu only devices. A good example of this is when a user trains their policy on a server, but wants to see the policy in action, and they open it on their laptop, but their laptop doesn't have a gpu.

PR #2265, opened 2021-04-15
Add gpu pickleable module to torch

ziyiwu9494:
A potential fix for #2079, by creating an garage.torch.Module wrapper class which implements pickling with moving to cpu on pickle and back to original device on unpickle if the device exists.

### DIFF
--- a/src/garage/torch/_functions.py
+++ b/src/garage/torch/_functions.py
@@ -369,11 +369,11 @@ def state_dict_to(state_dict, device):
     """Move optimizer to a specified device.
 
     Args:
-        state_dict (dict): state dictionary to be moved
+        state_dict (dict): state dictionary to be moved.
         device (str): ID of GPU or CPU.
 
     Returns:
-        dict: state dictionary moved to device
+        dict: state dictionary moved to device.
     """
     for param in state_dict.values():
         if isinstance(param, torch.Tensor):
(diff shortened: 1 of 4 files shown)


src/garage/torch/_functions.py
```diff
@@ -383,6 +383,61 @@ def state_dict_to(state_dict, device):
...
+        """Save the current device of the module before saving module state.
+
+        Returns:
+            dict: State dictionary.
+        """
+        # do we always run experiments on global device?
+        save_from_device = global_device()
+        self.to('cpu')
+        state = self.__dict__.copy()
+        state['device'] = save_from_device
```
krzentner, line 400: Probably should call this 'garage.global_device' or something, so that it definitely doesn't conflict with any sub-field names.

src/garage/torch/_functions.py
```diff
@@ -383,6 +383,61 @@ def state_dict_to(state_dict, device):
...
+        """Restore the module state, moving it back to the original device if possible.
+
+        Args:
+            state (dict): State dictionary.
+
+        """
+        system_device = global_device()
+        save_from_device = state['device']
+        if save_from_device == system_device:
+            module_state_to(state, system_device)
```
krzentner, line 413: I think you can use the other state-dict moving function you wrote here (even though this kinda isn't a state dict).

src/garage/torch/_functions.py
```diff
@@ -383,6 +383,61 @@ def state_dict_to(state_dict, device):
...
+    def __setstate__(self, state):
+        """Restore the module state, moving it back to the original device if possible.
+
+        Args:
+            state (dict): State dictionary.
+
+        """
+        system_device = global_device()
+        save_from_device = state['device']
+        if save_from_device == system_device:
```
krzentner, line 412: I'm not sure if this if statement is really needed. I think the idea here is to pre-move everything, as an optimization, but I'm not sure if that's actually faster. If you're going to do this, please use `timeit.timeit` to measure the performance difference.

krzentner (commented):
This looks pretty good. There's a few minor improvements I've mentioned above. We should also add this new class to `q_functions` and `value_functions` (as well as probably everything in `garage.torch.modules`), but that can happen in another PR.

src/garage/torch/_functions.py
ziyiwu9494: The reason I was doing it this way was because I didn't know if you needed to move any of the other attributes in the dict as well as the module itself, so I'm just moving all internal attributes first and then moving the module itself.

src/garage/torch/_functions.py
ziyiwu9494: Will every movable parameter of a nn.module be a tensor or a dict?

src/garage/torch/_functions.py
krzentner: Oh, I see. `.to()` *should* know how to handle that already (since that's how the module was moved to the device in the first place). If modules need to move something besides the default behavior, they should override `.to()` themselves.

PR #2266, opened 2021-04-20
Fix dimensionality of log_std in GaussianMLPModule

krzentner:


ryanjulian (approved):

src/garage/torch/modules/gaussian_mlp_module.py
```diff
@@ -121,12 +121,13 @@ def __init__(self,
         if self._std_parameterization not in ('exp', 'softplus'):
             raise NotImplementedError
 
-        init_std_param = torch.Tensor([init_std]).log()
+        self._init_std = torch.Tensor([init_std]).log()
+        log_std = torch.Tensor([init_std] * output_dim).log()
```
akashvelu, line 125: don't think a log() is necessary here, since it has already been taken in line 124?

akashvelu (approved):

src/garage/torch/modules/gaussian_mlp_module.py
akashvelu, line 125: don't think a log() is necessary here, since it has already been taken in line 124?

akashvelu (changes requested):

avnishn:
@krzentner ptal at @akashvelu's comments

src/garage/torch/modules/gaussian_mlp_module.py
akashvelu, line 125: don't think a log() is necessary here, since it has already been taken in line 124?

krzentner: Oh, good catch. Thanks!

krzentner: Oh, I see. Both of these lines used `init_std`, so it was actually fine. Thankfully we have tests that cover this line.

krzentner:
The only untested on CI line in the diff is GPU-only, so I'm merging this.

PR #2271, opened 2021-05-04
Change tf trainer import

ziyiwu9494:


src/garage/trainer.py
```diff
@@ -588,10 +581,17 @@ class TFTrainer(Trainer):
             trainer.resume(n_epochs=20)
 
     """
+    # pylint: disable=no-name-in-module
+    if 'tensorflow' in sys.modules:
+        tf = __import__('tensorflow')
```
krzentner, line 586: You can just use `global tf` in `__init__` (instead of creating a field on self). Just initialize `tf = None` at the top of the module with a comment about where and why it gets imported. Also, instead of checking if `tensorflow` is in sys.modules, you can just unconditionally import it in `__init__`, since the import error should be fairly obvious to the user.

krzentner (approved):

haydenshively (approved):

PR #2276, opened 2021-05-11
disable archive launch repo default

avnishn:
disable archive launch repo by default, that way examples inside the examples directory can be run from pip installations/


avnishn:
This will also likely need a backport, which could probably achieved via the mergify bot, given that this change is just a one liner.

krzentner (approved):
Looks good, need to capitalize the commit message :)

haydenshively (approved):

PR #2287, opened 2021-06-14
Merge mw v2 examples

avnishn:
This is a PR in progess ... definitely doesn't pass pre commit, test cov, etc. There are a few files that have been added that shouldn't have been, it would be nice if @krzentner could take a look at the end of the week once I've made some more progress on cleaning up these files.


avnishn:
The MT1 examples don't work at the moment, but everything else is looking good. @krzentner can you ptal?

qureshinomaan:
> The MT1 examples don't work at the moment

Hey @avnishn, what do you mean by this? Is there a bug in MT1 training examples? (eg : mtppo_metaworld_mt1_push.py)

krzentner:
Anyone looking to reproduce the Meta-World results with garage should checkout the code in this PR. The following command should do the trick: `pip install 'git+https://github.com/rlworkgroup/garage.git@merge-mw-v2-examples'`

PR #2322, opened 2022-04-16
Cloudpickle MultiprocessingSampler EnvUpdates

krzentner:


krzentner:
I'm considering this change to be a fix and merging it without review.

PR #2334, opened 2022-12-06
render_modes compatability with gym>0.17.3

ahalev:
As of gym==0.18.0, the metadata key `render.modes` has been renamed to `render_modes`.  As a result, defining a [GymEnv](https://github.com/rlworkgroup/garage/blob/master/src/garage/envs/gym_env.py#L62-L390) fails with newer versions of gym [here](https://github.com/rlworkgroup/garage/blob/master/src/garage/envs/gym_env.py#L147). 

This change supports new versions of gym while retaining support for older version.

### DIFF
--- a/src/garage/envs/gym_env.py
+++ b/src/garage/envs/gym_env.py
@@ -144,7 +144,10 @@ class GymEnv(Environment):
         self._max_episode_length = _get_time_limit(self._env,
                                                    max_episode_length)
 
-        self._render_modes = self._env.metadata['render.modes']
+        try:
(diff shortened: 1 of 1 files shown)


krzentner (approved):

ahalev:
@krzentner are there additional checks I need to pass to merge this PR? Looks like I can't merge to master as I'm not an authorized user.

krzentner:
The CI was really set up for more maintainers than just myself. I can merge this PR manually though.

